PR #741, opened 2021-04-12
use crictl to remove cached images if it's installed

eriksywu:
We should try to converge to use whatever CLI_TOOL has been set during CSE. This also fixes an issue where the container list of a node may contain <none> image lists because we remove pre-cached kubeproxy images via ctr, not crictl. This could be a bug on the upstream side but for now we should use crictl when we can.

Update: we also need to install crictl before cleanUpContainerImages


adilliadil (approved):

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -174,7 +173,7 @@ installKubeletKubectlAndKubeProxy() {
             extractKubeBinaries ${KUBERNETES_VERSION} ${KUBE_BINARY_URL}
         else
             if [[ "$CONTAINER_RUNTIME" == "containerd" ]]; then
-                extractHyperkube "ctr"
+                extractHyperkube "ctr" #TODO: use crictl if installed though in theory after 1.19.x hyperkube is no longer used
```
xuto2, line 176: we already stop baking hyperkube for 1.19. I think we can remove this now

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -67,6 +67,13 @@ fi
 
 configureAdminUser
 
+{{- if NeedsContainerd}}
+# If crictl gets installed then use it as the cri cli instead of ctr
+# crictl is not a critical component so continue with boostrapping if the install fails
+# CLI_TOOL is by default set to "ctr"
+installCrictl && CLI_TOOL="crictl"
+{{- end}}
+
```
xuto2, line 76: does this work for Mariner? maybe exclude that?

xuto2: also if crictl is not installed, it means it's raw ubuntu, do we still need that?

eriksywu: There's never an absolute need for crictl. If the install fails bootstrapping will continue using ctr. The only part that actuall uses ctr/crictl is to clean up the cached images. 
For for raw ubuntu it'll just use ctr b/c installCrictl will fail.

parts/linux/cloud-init/artifacts/cse_install.sh
eriksywu: I will remove the TODO comment in this PR and remove entire hyperkube code for containerd in a separate PR

eriksywu: added an ado item to track removal of hyperkube code for containerd

parts/linux/cloud-init/artifacts/cse_main.sh
xuto2: ok. if there is no image then no need to install crictl. 
I'm thinking about an implicit rule that we install nothing in CSE part to avoid potential egress break and to boost perf. wdyt?  If we do that, we can move the downloadcrictl to tools_install.sh and separate offline and runtime logic further.

xuto2 (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
xuto2: ok I think downloadCrictl is not beign called by CSE, let's move it to tools_install script in next PR

PR #742, opened 2021-04-13
Increase SIG publish timeout to 90m in packer

xuto2:
As sometimes creating sig image will take more than the default value 60m

https://www.packer.io/docs/builders/azure/arm#shared_image_gallery_timeout

### DIFF
--- a/vhdbuilder/packer/vhd-image-builder-gen2.json
+++ b/vhdbuilder/packer/vhd-image-builder-gen2.json
@@ -46,7 +46,8 @@
         "replication_regions": [
           "{{user `location`}}"
         ]
-      }
+      },
+      "shared_image_gallery_timeout": "90m"
     }
   ],
   "provisioners": [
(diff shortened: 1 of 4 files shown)


chengliangli0918 (approved):

chengliangli0918 (commented):
Let's set it to 120m for more buffer time. According to my observation this afternoon, for the same distro build, sometimes the imageversion replication proceeds 3% per minute, sometimes it proceeds 1% per minute.

mainred:
I am afraid shared_image_gallery_timeout is not correctly used in packer for 1.6.1 we are using. https://github.com/hashicorp/packer/pull/10816
The fix will come up in the next packer release. For now, we can simply enlarge `polling_duration_timeout` as a workaround, because polling_duration_timeout overwrites shared_image_gallery_timeout by mistake in packer.

mainred (changes requested):
I am afraid shared_image_gallery_timeout is not correctly used in packer for 1.6.1 we are using. https://github.com/hashicorp/packer/pull/10816
The fix will come up in the next packer release. For now, we can simply enlarge `polling_duration_timeout` as a workaround, because polling_duration_timeout overwrites shared_image_gallery_timeout by mistake in packer.

PR #744, opened 2021-04-13
Log file name which contains version info when downloading the file f…

AbelHu:
Log file name which contains version info when downloading the file from internet


parts/windows/kuberneteswindowsfunctions.ps1
```diff
@@ -58,7 +58,7 @@ function DownloadFileOverHttp {
         }
 
         $ProgressPreference = $oldProgressPreference
-        Write-Log "Downloaded file to $DestinationPath"
+        Write-Log "Downloaded file $fileName to $DestinationPath"
```
mainred, line 61: What about `Downloaded file $Url to $DestinationPath`, as the URL contains the file name, and both variables are function inputs

mainred (approved):
Thanks for your PR, left a nint.

parts/windows/kuberneteswindowsfunctions.ps1
AbelHu: Use $fileName to keep the same format as above `Using cached version of $fileName`. Agree that $Url is better.

PR #753, opened 2021-04-16
start containerd if it's not running

mainred:


parts/windows/windowscontainerdfunc.ps1
```diff
@@ -37,9 +37,8 @@ function RegisterContainerDService {
   if ($null -eq $svc) {
     Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_INSTALLED -ErrorMessage "containerd.exe did not get installed as a service correctly."
   }
-  Start-Service containerd
   if ($svc.Status -ne "Running") {
-    Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_RUNNING -ErrorMessage "containerd service is not running"
+    Start-Service containerd
```
AbelHu, line 41: Do you think that start-service will always succeed? Should we need to check it again?
I think that we should not remove the error when containerd is not running.

mainred: Added an exception check.

parts/windows/windowscontainerdfunc.ps1
```diff
@@ -37,9 +37,13 @@ function RegisterContainerDService {
   if ($null -eq $svc) {
     Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_INSTALLED -ErrorMessage "containerd.exe did not get installed as a service correctly."
   }
-  Start-Service containerd
   if ($svc.Status -ne "Running") {
-    Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_RUNNING -ErrorMessage "containerd service is not running"
+    try {
+      Start-Service containerd
+    }
+    catch {
```
AbelHu, line 44: Does it mean start-service will throw an exception? Could we retry 3 times to start-service and check the status, then return the error if all retires fail?

mainred: start-service throws an exception in some cases. I am not sure we are over-protecting the service. BTW, docker is using the same way to start up, but yeah we may also have issues in starting docker.

AbelHu: I did not see such error with docker, do you?
Please add an error if any exception will be thrown in starting containerd. And `Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_RUNNING -ErrorMessage $_` after checking that containerd is not running

AbelHu: I did not see such error with docker, do you?
Please add an error if any exception will be thrown in starting containerd. And `Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_RUNNING -ErrorMessage $_` after checking that containerd is not running

mainred: done, please take a look.

parts/windows/windowscontainerdfunc.ps1
```diff
@@ -33,14 +33,29 @@ function RegisterContainerDService {
...
+    Write-Log "Starting containerd, current status: $svc.Status"
+    Start-Service containerd
+    $retryCount++
+    Write-Log "Retry $retryCount : Sleep $retryInterval and check containerd status"
+    Sleep $retryInterval
+  } while ($retryCount -lt $maxRetryCount)
+
   if ($svc.Status -ne "Running") {
-    Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_RUNNING -ErrorMessage "containerd service is not running"
+    Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CONTAINERD_NOT_RUNNING -ErrorMessage $_
```
AbelHu, line 56: $_?

AbelHu (approved):

PR #755, opened 2021-04-16
Migrate installing/enabling Chrony from CSE to VHD

anujmaheshwari1:
Moving: disabling time sync services NTP and timesyncd and enabling Chrony
from: CSE
to: VHD (install dependencies)


adilliadil (approved):
LGTM

adilliadil:
@xuto2 & @anujmaheshwari1 reckon we remove the cse installation as well, or check to install it if not exists?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -91,9 +91,15 @@ disableNtpAndTimesyncdInstallChrony() {
     # Disable systemd-timesyncd
```
xuto2, line 91: let's move this method to tools_install_ubuntu.sh

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -52,12 +52,6 @@ fi
 configureHTTPProxyCA
 {{- end}}
 
-{{- if EnableChronyFor1804}}
```
xuto2, line 55: A follow up after this is to remvoe this EnableChronyFor1804 flag from AgentBaker and RP.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
xuto2: might also be able to move the error constant from cse_helper.sh to tools_install_ubutu.sh.

ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
ERR_STOP_OR_DISABLE_NTP_TIMEOUT
ERR_CHRONY_INSTALL_TIMEOUT

xuto2 (approved):

PR #756, opened 2021-04-16
update version to 0.9 for webhook

pragyaaneja:


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -225,14 +225,14 @@ if [[ ${installSGX} == "True" ]]; then
         echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
     done
 
-    SGX_PLUGIN_VERSIONS="0.2"
+    SGX_PLUGIN_VERSIONS="0.4"
```
robbiezhang, line 228: you should include both old and new until the AKS (with new image) is released to all regions.

pragyaaneja: Done.

robbiezhang (approved):

eriksywu (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -225,14 +225,20 @@ if [[ ${installSGX} == "True" ]]; then
         echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
     done
 
-    SGX_PLUGIN_VERSIONS="0.2"
+    SGX_PLUGIN_VERSIONS="
```
eriksywu, line 228: @adilliadil should we include sgx plugins in our vhd component check?

PR #758, opened 2021-04-19
doc: Rename release notes for AKS Windows VHD with 4B

AbelHu:
Rename release notes for AKS Windows VHD with 4B


mainred (approved):
Good catch, thanks. Publishing the release notes PR automatically is in my plan. Hope this comes soon.
BTW, I think the file name should be a copy and paste from the image version like in your aks-rp PR, and that will bring no problem.

PR #766, opened 2021-04-21
Remove {{/* */}} comments from error constants in cse_helpers.sh

anujmaheshwari1:
The {{/* */}} comments following the timeout error constants in cse_helpers.sh cause the constants to not be properly set in the bash environment. If we ever hit one of these constants in the VHD Build, they will throw out an unbound variable error. 

Fixed this using sed to remove the comments before sourcing the files. 

Also moved chrony/ntp/timesyncd related constants to tool_installs_ubuntu.sh


eriksywu (approved):
Thanks Arun!

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -4,6 +4,9 @@ OS=$(sort -r /etc/*-release | gawk 'match($0, /^(ID_LIKE=(coreos)|ID=(.*))$/, a)
 UBUNTU_OS_NAME="UBUNTU"
 MARINER_OS_NAME="MARINER"
 
+sed -i 's/{{\/\*[^*]*\*\/}}//g' /home/packer/provision_source.sh
```
eriksywu, line 7: small nit: add comment to explain what the sed is doing

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -9,6 +9,13 @@ ERR_UA_DETACH=185 {{/* Error to detach UA */}}
 ERR_LINUX_HEADER_INSTALL_TIMEOUT=186 {{/* Timeout to install linux header */}}
 ERR_STRONGSWAN_INSTALL_TIMEOUT=187 {{/* Timeout to install strongswan */}}
 
+ERR_NTP_INSTALL_TIMEOUT=10 {{/*Unable to install NTP */}}
```
xuto2, line 12: tool_installs_ubuntu.sh is not part of cse, so no need to use the {{ }} template syntax. We can just add # before the comment

xuto2 (approved):

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
anujmaheshwari1: should I update the other constants(FIPS related error codes) in tool_installs_ubuntu.sh declared above the ntp/timesyncd/chrony similarly or leave them as it is?

PR #768, opened 2021-04-22
chore: cache signedscripts v0.0.13

AbelHu:
- cache signedscripts v0.0.13 (fix the issue https://github.com/Azure/aks-engine/issues/4351)
- remove signedscripts-v0.0.10.zip since AKS Windows does not use it


mainred (approved):
Thanks

PR #773, opened 2021-04-23
Add retries while fetching Windows Defender signatures

AbelHu:
Please reference more details in https://github.com/kubernetes-sigs/image-builder/pull/553


mainred (approved):

AbelHu:
@mainred It seems like that the change still does not work. I will still merge this PR since there are comments to explain it.

PR #775, opened 2021-04-23
updating omsagent for April release

daweim0:
This is for the April OMS agent release (Container Insights)

pkg/agent/k8s_versions_for_baker_test.go also has a reference to omsagent, but it doesn't look like that file should be updated to have the newest agent (please correct me if I'm wrong)

/assign @ganga1980


alexeldeib:
lgtm

PR #786, opened 2021-04-27
Check whether a folder exists before creating it

### LINKED ISSUE #785: Check whether the folder C:\k\azurecni\netconf exists before creating it
Check whether the folder C:\k\azurecni\netconf exists before creating it

```
2021-04-27T03:49:28.7639517+00:00: Installing Azure VNet plugins


    Directory: C:\k\azurecni


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----        4/27/2021   3:49 AM                bin
mkdir : An item with the specified name C:\k\azurecni\netconf already exists.
At C:\AzureData\windows\windowsazurecnifunc.ps1:32 char:5
+     mkdir $AzureCNIConfDir
+     ~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceExists: (C:\k\azurecni\netconf:String) [New-Item], IOException
    + FullyQualifiedErrorId : DirectoryExist,Microsoft.PowerShell.Commands.NewItemCommand

```

AbelHu:
fix #785


parts/windows/windowscsehelper.ps1
```diff
@@ -26,6 +26,23 @@ function Postpone-RestartComputer
...
+    Param(
+        [Parameter(Mandatory=$true)][string]
+        $Folder,
+        [Parameter(Mandatory=$false)][string]
+        $FolderUsage = "general purpose"
+    )
+    
+    if (-Not (Test-Path $Folder)) {
+        Write-Log "Create folder $Folder for $FolderUsage"
+        New-Item -ItemType Directory -Path $Folder > $null
```
mainred, line 40: Will this also be able to make parent directories as needed, like `make -p`

pkg/templates/templates_generated.go
```diff
@@ -6865,6 +6860,23 @@ function Postpone-RestartComputer
     Write-Log "Created an one-time task to restart the VM"
 }
 
+function Create-Directory
+{
+    Param(
+        [Parameter(Mandatory=$true)][string]
+        $Folder,
```
karataliu, line 6867: not blocking.. Directory and Folder are two marks referring to same thing.

If the function name is already Create-Directory, the para could be 'Name' and 'Usage'

parts/windows/windowscontainerdfunc.ps1
```diff
@@ -126,7 +126,7 @@ function Enable-Logging {
     $logs = Join-path $pwd.drive.Root logs
     Write-Log "Containerd hyperv logging enabled; temp location $logs"
     $diag = Join-Path $global:ContainerdInstallLocation diag.ps1
-    mkdir -Force $logs
```
mainred, line 129: Does `mkdir -Force` equal our new function `Create-Directory`? because it will create the folder anyway.

parts/windows/windowscsehelper.ps1
AbelHu: Just do the same function as old command `mkdir` and `New-Item -ItemType Directory` do. Do not introduce new functionality.

parts/windows/windowscontainerdfunc.ps1
AbelHu: mkdir -Force can do the same thing. Just keep them in a same format with using a same function.

AbelHu: And share the same code and format for log.

AbelHu: mkdir is a function which calls New-Item to create directories specifically. I prefer to use New-Item in PowerShell script

parts/windows/windowscsehelper.ps1
mainred: `New-Item -ItemType Directory -Path $Folder > $null` is able to make parent directories as needed per my test.

mainred (approved):
LGTM from my side.

pkg/templates/templates_generated.go
AbelHu: @karataliu Rename to "FullPath" since it is a full path but not a name.

feiskyer (approved):

karataliu (approved):

PR #745, opened 2021-04-13
chore: update containerd package

mainred:

PR #788, opened 2021-04-28
20210331 hotfix for Windows containerd

mainred:


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -15,7 +15,7 @@ $ErrorActionPreference = "Stop"
 
 filter Timestamp { "$(Get-Date -Format o): $_" }
 
-$global:containerdPackageUrl = "https://mobyartifacts.azureedge.net/moby/moby-containerd/1.4.3+azure/windows/windows_amd64/moby-containerd-1.4.3+azure-1.amd64.zip"
+$global:containerdPackageUrl = "https://acs-mirror.azureedge.net/containerd/windows/v0.0.1/binaries/containerd-v0.0.1-windows-amd64.tar.gz"
```
AbelHu, line 18: I think that we should not cherry-pick chore: update containerd package (#745) for hotfix since it is used to build VHD.

AbelHu: Same for the commit `fix: enables DNS resolution of SMB shares for containerD`

PR #795, opened 2021-04-30
chore: keep one env path for containerd

mainred:


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -264,12 +267,14 @@ function Install-ContainerD {
         Expand-Archive -path $containerdTmpDest -DestinationPath $installDir -Force
     } else {
         tar -xzf $containerdTmpDest --strip=1 -C $installDir
+        mv -Force $installDir\bin\* $installDir\
```
mainred, line 270: Need to use powershell cmdlet instead

parts/windows/windowscontainerdfunc.ps1
```diff
@@ -160,17 +160,19 @@ function Install-Containerd {
...
-    del $zipfile
+    Remove-Item -Path $zipfile -Force
+
   }
   elseif ($ContainerdUrl.endswith(".tar.gz")) {
     # upstream containerd package is a tar 
     $tarfile = [Io.path]::Combine($ENV:TEMP, "containerd.tar.gz")
     DownloadFileOverHttp -Url $ContainerdUrl -DestinationPath $tarfile
     Create-Directory -FullPath $global:ContainerdInstallLocation -DirectoryUsage "storing containerd"
     tar -xzf $tarfile -C $global:ContainerdInstallLocation
```
AbelHu, line 171: I think that we can remove elseif. tar -xzf does not work in Windows node.

mainred: tar -xzf should work in Windows node. We used this command the uncompress the `https://acs-mirror.azureedge.net/containerd/windows/v0.0.1/binaries/containerd-v0.0.1-windows-amd64.tar.gz`

mainred: Also, this works on my Windows PC

AbelHu: It seems like that tar.exe is installed by default in Windows.

parts/windows/windowscontainerdfunc.ps1
```diff
@@ -160,17 +160,19 @@ function Install-Containerd {
...
-    mv -Force $global:ContainerdInstallLocation\bin\* $global:ContainerdInstallLocation\
-    del $tarfile
-    del -Recurse -Force $global:ContainerdInstallLocation\bin
+
+    Get-ChildItem -Path $global:ContainerdInstallLocation\bin -Recurse -File | Move-Item -Destination $installDir
```
AbelHu: Please be careful. They are not equal. mv will move all files under bin\* to dest with keeping the folder structure in bin.
You can create a folder test under bin and put some files in the folder test to run these two commands.
mv will move the folder test to the dest but Get-ChildItem|Move-Item will move files under the folder test to the dest but do not create the folder test in the dest. So if there are some files in different folders with the same name, new command will have the issue.

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -264,12 +268,14 @@ function Install-ContainerD {
         Expand-Archive -path $containerdTmpDest -DestinationPath $installDir -Force
     } else {
         tar -xzf $containerdTmpDest --strip=1 -C $installDir
+        Get-ChildItem -Path $installDir\bin -Recurse -File | Move-Item -Destination $installDir
+        Remove-Item -Path $installDir\bin -Force
     }
     Remove-Item -Path $containerdTmpDest | Out-Null
 
-    $newPaths = [Environment]::GetEnvironmentVariable("Path", [EnvironmentVariableTarget]::Machine) + ";$installDir;$installDir/bin"
+    $newPaths = [Environment]::GetEnvironmentVariable("Path", [EnvironmentVariableTarget]::Machine) + ";$installDir"
```
AbelHu, line 276: It seems like that the changes in this PR is not for fixing any issue but an improvement. I strongly suggest to hold on it to avoid some unexpected issues to block containerd plan.

mainred: sure, we can keep this in review, and it's not a  fix as you said, as we use only "tar.gz" as the only source of containerd.
I tested with a "zip" file, we will have two set of containerd.exe, but we won't have this kind of package.

parts/windows/windowscontainerdfunc.ps1
AbelHu: cc @mainred

mainred: The binaries so far have no conflicted files, and shouldn't have otherwise they should be included in the env path.
But I don't mind use mv instead to avoid any possible accident.

parts/windows/windowscontainerdfunc.ps1
```diff
@@ -160,17 +160,18 @@ function Install-Containerd {
...
     $tarfile = [Io.path]::Combine($ENV:TEMP, "containerd.tar.gz")
     DownloadFileOverHttp -Url $ContainerdUrl -DestinationPath $tarfile
     Create-Directory -FullPath $global:ContainerdInstallLocation -DirectoryUsage "storing containerd"
     tar -xzf $tarfile -C $global:ContainerdInstallLocation
+
     mv -Force $global:ContainerdInstallLocation\bin\* $global:ContainerdInstallLocation\
-    del $tarfile
-    del -Recurse -Force $global:ContainerdInstallLocation\bin
+    Remove-Item -Path $tarfile -Force
+    Remove-Item -Path $global:ContainerdInstallLocation\bin -Force
```
AbelHu, line 174: Do we need `-Recurse` for the folder?

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -154,12 +158,14 @@ function Install-ContainerD {
         Expand-Archive -path $containerdTmpDest -DestinationPath $installDir -Force
     } else {
         tar -xzf $containerdTmpDest --strip=1 -C $installDir
+        mv -Force $installDir\bin\* $installDir
```
AbelHu, line 161: mv => Move-Item? To keep the same format.

AbelHu (approved):

parts/windows/windowscontainerdfunc.ps1
mainred: Sure, why not.

vhdbuilder/packer/configure-windows-vhd.ps1
mainred: I reverted the code the use mv instead.

PR #797, opened 2021-05-04
update kubeproxy/kubelet/kubectl image 1.18.18/1.19.10/1.20.6/1.21.0

xuto2:


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -318,11 +320,16 @@ KUBE_PROXY_IMAGE_VERSIONS="
...
+1.19.9-hotfix.20210505
+1.19.10-hotfix.20210504
 1.20.2
 1.20.2-hotfix.20210310
 1.20.2-hotfix.20210310.2
 1.20.5-hotfix.20210322
 1.20.5-hotfix.20210322.1
+1.20.5-hotfix.20210505
+1.20.6-hotfix.20210504
+1.21.0
```
AbelHu, line 332: Will we support 1.21.0? If so, I will add this in AKS Windows VHD

AbelHu (approved):

vhdbuilder/packer/install-dependencies.sh
xuto2: yes we will

PR #798, opened 2021-05-04
fix:  HYPERV_GENERATION missing for sig mode and unbound variable for windows cache test

mainred:


vhdbuilder/packer/test/run-test.sh
```diff
@@ -6,6 +6,11 @@ TEST_RESOURCE_PREFIX="vhd-test"
 TEST_VM_ADMIN_USERNAME="azureuser"
 TEST_VM_ADMIN_PASSWORD="TestVM@$(date +%s)"
 
+if [ "$OS_TYPE" == "Linux" ] && ([ "$OS_SKU" == "CBLMariner" ] || [ "$OS_VERSION" == "16.04" ] || [ "$MODE" == "gen2Mode" ]); then
```
mainred, line 9: - Windows has no definition of OS_SKU.
- move the check before creating resource group

vhdbuilder/packer/test/run-test.sh
```diff
@@ -6,6 +6,11 @@ TEST_RESOURCE_PREFIX="vhd-test"
 TEST_VM_ADMIN_USERNAME="azureuser"
 TEST_VM_ADMIN_PASSWORD="TestVM@$(date +%s)"
 
+if [ "$OS_TYPE" == "Linux" ] && {[ "$OS_SKU" == "CBLMariner" ] || [ "$OS_VERSION" == "16.04" ] || [ "$MODE" == "gen2Mode" ]}; then
```
mainred, line 9: use {...; } per suggestion from
https://github.com/koalaman/shellcheck/wiki/SC2235

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -40,6 +40,7 @@ steps:
       -e SIG_GALLERY_NAME=${SIG_GALLERY_NAME} \
       -e SIG_IMAGE_NAME=${SIG_IMAGE_NAME} \
       -e SIG_IMAGE_VERSION=${SIG_IMAGE_VERSION} \
+      -e HYPERV_GENERATION="V1" \
```
xuto2, line 43: But I don't see where do we use this HYPERV_GENERATION in the run-test.sh?

xuto2 (approved):

.pipelines/templates/.builder-release-template-windows.yaml
mainred: This is a fix of https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/init-variables.sh#L97

vhdbuilder/packer/test/run-test.sh
```diff
@@ -6,6 +6,13 @@ TEST_RESOURCE_PREFIX="vhd-test"
 TEST_VM_ADMIN_USERNAME="azureuser"
 TEST_VM_ADMIN_PASSWORD="TestVM@$(date +%s)"
 
+if [ "$OS_TYPE" == "Linux" ]; then
```
mainred, line 9: This is a workaround, going to extract the code used for Windows and Linux so that we don't too much dependency

PR #800, opened 2021-05-06
feat: Collect kubeclusterconfig.json

AbelHu:
kubeclusterconfig.json does not contain any credentials but it contains some parameters for trouble shooting, for example, `WinDSR=true, WinOverlay=false` is included in FeatureGates when enabling WinDSR.


mainred (approved):
Please make sure there's no user credential in it. I think we don't but have no cluster around to check.

PR #802, opened 2021-05-06
feat: install blobfuse 1.3.7 by default

andyzhangx:
@xuto2


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -37,7 +37,7 @@ installDeps
 cat << EOF >> ${VHD_LOGS_FILEPATH}
   - apache2-utils
   - apt-transport-https
-  - blobfuse=1.3.5
+  - blobfuse=1.3.7
```
xuto2, line 40: This is just log, are you missing the real change?

andyzhangx: fixed

xuto2 (approved):

PR #804, opened 2021-05-07
update Windows containerd version to v0.0.41

mainred:


vhdbuilder/cgmanifest.json
```diff
@@ -16,8 +16,8 @@
                 "Type": "other",
                 "Other": {
                     "Name": "ContainerdWindows",
-                    "Version": "0.0.1",
-                    "DownloadUrl": "https://acs-mirror.azureedge.net/containerd/windows/v0.0.1/binaries/containerd-v0.0.1-windows-amd64.tar.gz"
+                    "Version": "0.0.41",
+                    "DownloadUrl": "https://k8sreleases.blob.core.windows.net/containerd/windows/v0.0.41/binaries/containerd-v0.0.41-windows-amd64.tar.gz"
```
mainred, line 20: 0.0.1 is for private preview and is not a formal one, we want to use 0.0.41 for public preview.
The big version gap is caused by building tools.

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -15,7 +15,7 @@ $ErrorActionPreference = "Stop"
 
 filter Timestamp { "$(Get-Date -Format o): $_" }
 
-$global:containerdPackageUrl = "https://acs-mirror.azureedge.net/containerd/windows/v0.0.1/binaries/containerd-v0.0.1-windows-amd64.tar.gz"
+$global:containerdPackageUrl = "https://k8sreleases.blob.core.windows.net/containerd/windows/v0.0.41/binaries/containerd-v0.0.41-windows-amd64.tar.gz"
```
AbelHu, line 18: Can it be downloaded by ff/mc? Or do we only public preview in global azure? If cx has firewall, it may not be downloaded

mainred: Updated to use acs-mirrored one. Thanks

AbelHu (approved):

PR #807, opened 2021-05-07
Update post-VHD tests for Gen2 builds

anujmaheshwari1:
Earlier post VHD build tests only tested for gen1. This change incorporates testing for gen2 builds as well.


vhdbuilder/packer/test/run-test.sh
```diff
@@ -29,24 +29,38 @@ trap cleanup EXIT
...
-    exit 1
-  fi
+    id=$(az sig image-definition show \
+      --resource-group ${AZURE_RESOURCE_GROUP_NAME} \
+      --gallery-name ${SIG_GALLERY_NAME} \
+      --gallery-image-definition ${SIG_IMAGE_NAME}) || id=""
+    if [ -z "$id" ]; then
+      echo "Image definition ${SIG_IMAGE_NAME} does not exist in gallery ${SIG_GALLERY_NAME} resource group ${AZURE_RESOURCE_GROUP_NAME}"
+      exit 1
+    fi
```
xuto2, line 57: remvoe the extra space

vhdbuilder/packer/test/run-test.sh
```diff
@@ -29,24 +29,38 @@ trap cleanup EXIT
...
 
-  IMG_DEF="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/${SIG_GALLERY_NAME}/images/${SIG_IMAGE_NAME}/versions/${SIG_IMAGE_VERSION}"
+    IMG_DEF="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/${SIG_GALLERY_NAME}/images/${SIG_IMAGE_NAME}/versions/${SIG_IMAGE_VERSION}"
+  else
```
mainred, line 60: Could we use the check `if [ "$MODE" == "gen2Mode" ];` otherwise this mode will be regarded as the default matched mode and Windows won't get here.

vhdbuilder/packer/test/run-test.sh
anujmaheshwari1: that is not extra space, there is an indentation increase at that level because of how I rewrote the code. I will check again though.

vhdbuilder/packer/test/run-test.sh
anujmaheshwari1: there are 3 build modes, 1) default 2) sigMode 3) gen2Mode

I dont see how Windows misses any required mode? Am I missing something?

mainred: For now, Windows use only default and sigmode. But this is not a blocker.

mainred (approved):

mainred:
Please make sure all the checks pass before we merge the pr, thanks.

xuto2 (approved):

PR #808, opened 2021-05-08
nmi: vendor nmi 1.7.5 image

bcho:


vhdbuilder/packer/components.json
```diff
@@ -325,7 +325,8 @@
       "downloadURL": "mcr.microsoft.com/oss/azure/aad-pod-identity/nmi:*",
       "versions": [
         "v1.7.0",
```
andyliuliming, line 327: all cluster switched to use 1.7.4 now?
if yes, we can remove the 1.7.0

andyliuliming (approved):

vhdbuilder/packer/components.json
bcho: updated, PTAL

PR #813, opened 2021-05-11
chore: Cache omsagent ciprod in Windows VHD

AbelHu:
Cache omsagent ciprod in Windows VHD


mainred:
It will be great if we have more context for this change.

AbelHu:
@mainred ciprod is used by omsagent. It is used to create omsagent pods in Windows nodes when monitoring addon is enabled.

AbelHu:
ciprod is used to create omsagent pods in Windows nodes when monitoring addon is enabled. It is introduced by omsagent team.

andyliuliming:
this will take another 1.7 g space, for others lgtm

PR #814, opened 2021-05-11
chore: Add Calico Windows v3.19.0

AbelHu:
- Add Calico Windows v3.19.0
- Remove Calico Windows v3.17.2

### DIFF
--- a/vhdbuilder/cgmanifest.json
+++ b/vhdbuilder/cgmanifest.json
@@ -5,8 +5,8 @@
                 "Type": "other",
                 "Other": {
                     "Name": "CalicoWindows",
-                    "Version": "3.17.1",
-                    "DownloadUrl": "https://acs-mirror.azureedge.net/calico-node/v3.17.1/binaries/calico-windows-v3.17.1.zip"
+                    "Version": "3.19.0",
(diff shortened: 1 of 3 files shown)


vhdbuilder/cgmanifest.json
```diff
@@ -5,8 +5,8 @@
                 "Type": "other",
                 "Other": {
                     "Name": "CalicoWindows",
-                    "Version": "3.17.1",
-                    "DownloadUrl": "https://acs-mirror.azureedge.net/calico-node/v3.17.1/binaries/calico-windows-v3.17.1.zip"
+                    "Version": "3.19.0",
+                    "DownloadUrl": "https://acs-mirror.azureedge.net/calico-node/v3.19.0/binaries/calico-windows-v3.19.0.zip"
```
AbelHu, line 9: @mainred, I think that we do not need to change it every time since the license should never be changed. I change it since you change containerd in last pr.

mainred: Hi @AbelHu 
This is from CELA team. We may not able to register all oss components but we can try to make sure what has been done are well registered with different versions we are using in the vhd. I think it's not too much effort.
_Correct – all OSS is registered and that includes use of different versions of a component._

mainred (approved):
Left a comment regarding the oss registration. Not a blocker, but I think we can try.

PR #820, opened 2021-05-12
Update calico images for CVE

mattstam:
CVE's in found in pod2daemon-flexvol so we need to go from v3.8.9 -> v3.8.9.1

Tigera-operator also needs updating so adding that in with the relevant Calico images. I will list what images tigera-operator pulls below:
```
Calico v3.17.2 via tigera-operator v1.13.5
Calico v3.18.1 via tigera-operator v1.15.1
Calico v3.19.0 via tigera-operator v1.17.1
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -168,7 +168,7 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/calico/pod2daemon-flexvol:*",
       "versions": [
-        "v3.8.9"
+        "v3.8.9.1"
       ]
     },
     {
@@ -180,7 +180,7 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/tigera/operator:*",
       "versions": [
-        "v1.13.5"
+        "v1.17.1"
       ]
     },
     {
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -154,6 +154,7 @@
       "downloadURL": "mcr.microsoft.com/oss/calico/node:*",
       "versions": [
         "v3.17.2",
+        "v3.18.1",
         "v3.19.0",
         "v3.8.9.2",
         "v3.8.9.5"
@@ -163,6 +164,7 @@
       "downloadURL": "mcr.microsoft.com/oss/calico/typha:*",
       "versions": [
         "v3.17.2",
+        "v3.18.1",
         "v3.19.0",
         "v3.8.9"
       ]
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/components.json
```diff
@@ -180,7 +180,7 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/tigera/operator:*",
       "versions": [
-        "v1.13.5"
+        "v1.17.1"
```
paulgmiller, line 183: technically theres a bunch of otehr images this is brining in

robbiezhang, line 183: please append new versions, and keep the old versions until the weekly OverlayMgr release for the new image is rolled out to all regions.

mattstam: Should we add in all the Calico 3.19 images that this indirectly brings?

mattstam: We're currently going from v1.15.1 -> v1.17.1, so I think we forgot to update that previously and that's why we have v1.13.5 here. I could add in v1.15.1 as well if it would make sense to do so.

paulgmiller: thanks

paulgmiller (approved):

vhdbuilder/packer/components.json
```diff
@@ -154,6 +154,7 @@
       "downloadURL": "mcr.microsoft.com/oss/calico/node:*",
       "versions": [
         "v3.17.2",
+        "v3.19.0",
```
paulgmiller, line 157: Thought there were UT's around this file maybe I'm out of date.

mattstam: It looks like [what would be the appropriate UT](https://github.com/Azure/AgentBaker/blob/master/pkg/agent/k8s_versions_for_baker_test.go) is out of date, and currently would not even hit k8s > 1.20 scenarios which is what we care about for these tigera-operator images. 

Looking at recent examples, these image update changes for linux don't usually have a corresponding UT change such as [#821](https://github.com/Azure/AgentBaker/pull/821/files)

vhdbuilder/packer/components.json
```diff
@@ -162,25 +164,31 @@
...
+        "v3.18.1",
+        "v3.19.0"
       ]
     },
     {
       "downloadURL": "mcr.microsoft.com/oss/tigera/operator:*",
       "versions": [
-        "v1.13.5"
+        "v1.13.5",
+        "v1.15.1",
```
mattstam, line 190: Adding in operator v1.15.1 (and corresponding calico v3.18.1) images as requested by @andyliuliming.

andyliuliming (approved):

andyliuliming (approved):

PR #821, opened 2021-05-12
updating omsagent version

daweim0:
Updating to omsagent version 05122021. I'm not removing ciprod03262021 because ciprod04222021 isn't done rolling out yet.

We aren't updating the windows image with this release, so no new windows image.

@docker-provider-devs

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -1,4 +1,4 @@
-{
+W{
   "ContainerImages": [
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/kubernetes-dashboard:*",
@@ -139,8 +139,8 @@
     {
       "downloadURL": "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:*",
       "versions": [
-        "ciprod03262021",
         "ciprod04222021",
+        "ciprod03262021",
         "ciprod05122021"
       ]
     },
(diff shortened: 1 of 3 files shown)


daweim0:
It looks like the windows VHD builds are failing because of a build pipeline issue, `az login` prints an error message `ArgumentParseError: argument --password/-p: expected one argument` https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=42451242&view=logs&j=4cae91f8-9300-5d2a-325c-cec7e691ae6d&t=ba6e6658-3216-5a7b-1df5-e991efde6751&l=41
I don't think that's a problem with this PR.

vhdbuilder/packer/components.json
```diff
@@ -140,7 +140,8 @@
       "downloadURL": "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:*",
       "versions": [
         "ciprod03262021",
-        "ciprod04222021"
+        "ciprod04222021",
+        "ciprod05122021"
```
AbelHu, line 144: Can you also help to add ciprod05122021 for Windows? Example: https://github.com/Azure/AgentBaker/pull/813/files

daweim0: We aren't publishing a windows agent with this release (no changes were made to it. We could republish the old windows image under a new tag, but we would prefer to keep using the old windows tag)

AbelHu: Do you mean that ciprod04222021 is still used in AKS Windows nodes?

AbelHu (approved):

vhdbuilder/packer/components.json
daweim0: Correct, ciprod04222021 is still the latest and greatest for Windows nodes.

PR #822, opened 2021-05-13
chore: remove old Windows images

AbelHu:
1. Separate images for containerd from docker's.
2. Remove PauseImage 1.4.1 since we have updated it to 3.4.1 with last Windows VHD in AKS
3. Remove old versions v0.5.1, v0.6.0, for azure-cloud-node-manager
3. Only keep one latest version for each image for containerd as a workaround for the issue that disk space is not enough in the VHD since containerd is in public preview

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -102,29 +102,41 @@ function Get-ContainerImages {
     $imagesToPull = @()
 
     switch ($windowsSKU) {
-        { '2019', '2019-containerd'} {
+        '2019' {
             $imagesToPull = @(
                 "mcr.microsoft.com/windows/servercore:ltsc2019",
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -102,29 +102,41 @@ function Get-ContainerImages {
     $imagesToPull = @()
 
     switch ($windowsSKU) {
-        { '2019', '2019-containerd'} {
+        '2019' {
```
mainred, line 105: Is it better to use common images for docker and containerd and images for docker only? 
The images for docker only is the old ones. And when people should add new images to the common image list.

AbelHu: This is a workaround. We will combine them later after finding a better solution

mainred (approved):

PR #832, opened 2021-05-17
fix: Log disk info of C drive

AbelHu:
Old code does not always work so fix it.

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -433,6 +433,16 @@ function Update-Registry {
     }
 }
 
+function Get-SystemDriveDiskInfo {
+    Write-Log "Get Disk info"
+    $disksInfo=Get-CimInstance -ClassName Win32_LogicalDisk
+    foreach($disk in $disksInfo) {
+        if ($disk.DeviceID -eq "C:") {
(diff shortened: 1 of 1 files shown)


AbelHu:
@mainred It works. You can see the disk info in the build logs.

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -472,8 +482,7 @@ switch ($env:ProvisioningPhase) {
         Update-Registry -containerRuntime $containerRuntime
         Get-ContainerImages -containerRuntime $containerRuntime -windowsSKU $windowsSKU
         Get-FilesToCacheOnVHD -containerRuntime $containerRuntime
-        # Show disk space
-        Get-CimInstance -ClassName Win32_LogicalDisk
```
mainred, line 476: In which case this does not work?

AbelHu: No idea. I just found that it does not log the disk info as expected sometimes.

AbelHu: @mainred do you have other concern for this PR and need me to update it?

mainred (approved):
Thanks for the fix.

PR #833, opened 2021-05-17
Add Konnectivity v0.0.17

jveski:


vhdbuilder/packer/components.json
```diff
@@ -110,6 +110,7 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/apiserver-network-proxy/agent:*",
       "versions": [
+        "v0.0.17",
         "v0.0.16",
         "v0.0.15",
```
robbiezhang, line 115: shall we delete the old ones? Only v0.0.16 is in use, and we can remove the rest.

xuto2 (approved):

vhdbuilder/packer/components.json
jveski: good call. Will confirm and address in a separate PR.

PR #835, opened 2021-05-18
chore: share configuration in Windows VHD build and test

### LINKED ISSUE #824: Refine windows-vhd-content-test.ps1 to use the same policy to filter k8s versions in configure-windows-vhd.ps1
In configure-windows-vhd.ps1
```

            # Windows containerD supports Windows containerD, starting from Kubernetes 1.20
            if ($containerRuntime -eq 'containerd' -And $dir -eq "c:\akse-cache\win-k8s\") {
                $k8sMajorVersion = $fileName.split(".",3)[0]
                $k8sMinorVersion = $fileName.split(".",3)[1]
                if ($k8sMinorVersion -lt "20" -And $k8sMajorVersion -eq "v1") {
                    Write-Log "Skip to download $url for containerD is supported from Kubernets 1.20"
                    continue
                }
            }

```
But in windows-vhd-content-test.ps1, it uses two list to validate cached k8s versions. This is not convenient. I think that we can use the same policy as that in configure-windows-vhd.ps1

AbelHu:
1. Share configuration in Windows VHD build and test
2. Remove 2004
3. Fix #824
I compared the generated release notes and LGTM.


vhdbuilder/packer/test/run-test.sh
```diff
@@ -109,13 +110,22 @@ if [ "$OS_TYPE" == "Linux" ]; then
...
+  SCRIPT_PATH="$CDIR/../$WIN_CONFIGURATION_SCRIPT_PATH"
+  echo "Run $SCRIPT_PATH"
+  az vm run-command invoke --command-id RunPowerShellScript \
+    --name $VM_NAME \
+    --resource-group $RESOURCE_GROUP_NAME \
+    --scripts @$SCRIPT_PATH \
+    --output json
+
   SCRIPT_PATH="$CDIR/$WIN_SCRIPT_PATH"
+  echo "Run $SCRIPT_PATH"
```
mainred, line 122: combine the test to take less time?

mainred: otherwise, how do we handle the first run-command error?

vhdbuilder/packer/test/windows-vhd-content-test.ps1
```diff
@@ -121,8 +50,13 @@ function Test-FilesToCacheOnVHD
             $fileName = [IO.Path]::GetFileName($URL)
             $dest = [IO.Path]::Combine($dir, $fileName)
 
-            if ($containerRuntime -eq "containerd" -And $fakeDir -eq "c:\akse-cache\win-k8s-docker\") {
-                continue
+            if ($containerRuntime -eq "containerd" -And $fakeDir -eq "c:\akse-cache\win-k8s\") {
```
mainred, line 53: just a reminder, we can benefit from using the same code for biz and test for simplicity but if we have an error in the share code, we won't be able to catch it in the test.

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -6,16 +6,11 @@
...
 
-param (
-    $containerRuntime,
-    $windowsSKU
-)
-
 $ErrorActionPreference = "Stop"
 
-filter Timestamp { "$(Get-Date -Format o): $_" }
+. c:\windows-vhd-configuration.ps1
```
mainred, line 11: I find it's used in multiple places,  is it possible to create a variable for it?

mainred (approved):
Looks good and left a few comments, I think you compared the release notes of both the windows containerd and docker images. Thanks

vhdbuilder/packer/test/run-test.sh
AbelHu: I did not find a way to run two scripts at the same time in `az vm run-command`.
The first step normally is not thrown errors. If it has some issue, we can capture the errors when running the second script.

vhdbuilder/packer/test/windows-vhd-content-test.ps1
AbelHu: The shared part is only for configuration. We should never share functions for this reason.

vhdbuilder/packer/configure-windows-vhd.ps1
AbelHu: Only twice. One for import and another is for deletion. We can improve it if we think that it is worth later.

PR #836, opened 2021-05-18
preload more kube proxy images into the vhd.

andyliuliming:


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -317,16 +315,14 @@ KUBE_PROXY_IMAGE_VERSIONS="
...
-1.19.9-hotfix.20210322
+1.19.7-hotfix.20210511
 1.19.9-hotfix.20210322.1
 1.19.9-hotfix.20210505
 1.19.10-hotfix.20210504
 1.20.2
-1.20.2-hotfix.20210310
 1.20.2-hotfix.20210310.2
-1.20.5-hotfix.20210322
+1.20.2-hotfix.20210511
```
AbelHu, line 325: Do you need to add 1.20.2-hotfix.20210518, 1.20.5-hotfix.20210518 and 1.20.6-hotfix.20210518 built by Qi Ni today?

AbelHu (approved):

vhdbuilder/packer/install-dependencies.sh
andyliuliming: I only care the cve. maybe qi want to switch to use it later.

PR #837, opened 2021-05-18
Update VHD build workflow

xuto2:


vhdbuilder/packer/init-variables.sh
```diff
@@ -59,6 +59,9 @@ echo "storage name: ${STORAGE_ACCOUNT_NAME}"
 
 # If SIG_IMAGE_NAME hasnt been provided in Gen2 mode, set it to the default value
 if [[ "$MODE" == "gen2Mode" ]]; then
+	if 	[[ -z "$SIG_GALLERY_NAME" ]]; then
+		SIG_GALLERY_NAME="PackerSigGalleryEastUS"
+	fi
```
hbeberman, line 64: cleanup.sh also needs this block since it'll try to cleanup the imported SIG image after a build.

xuto2: will do

PR #838, opened 2021-05-19
fix: Fix write-log is not used before defined

AbelHu:
Fix write-log is not used before defined.
Use `throw` instead of `log & exit 1` in `windows-vhd-configuration.ps1`


vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -1,16 +1,15 @@
+# MUST define global variable with "global"
```
mainred, line 1: Maybe we can add a comment to describe the intention of this script, like
This script is to define **global** variables and please do not add extra biz logic here.

AbelHu: Added

mainred (approved):

PR #848, opened 2021-05-21
added linux container insights agent image 05202021

daweim0:
Adding the Container Insights 5-20-2021 release. This release is also Linux only, so there is no updated Windows image to add.

I left the 03262021 image and removed the other two newer images because CI rolled back to 03262021 in production. We discovered an issue in the following two images and abandon those two releases. They will never be deployed again, so no need to have them in the agent baker.

@microsoft/docker-provider-devs


alexeldeib:
lgtm

PR #852, opened 2021-05-26
Add Windows k8s hotfixes

AbelHu:
Add k8s hotfixes for Windows: 
1.21: https://github.com/kubernetes/kubernetes/pull/102001
1.20: https://github.com/kubernetes/kubernetes/pull/102000
1.19: https://github.com/kubernetes/kubernetes/pull/101999


feiskyer (approved):

mainred (approved):

AbelHu:
cc @xuto2: These hotfixes is only for Windows so I think that Linux does not need to pick up them.

PR #853, opened 2021-05-26
Set FipsAlgorithmPolicy to enable fips mode on windows nodes based on the property of windows agentpool profile

chengliangli0918:
Set FipsAlgorithmPolicy to enable fips mode on windows nodes if this windows agentpool profile is fips enabled.


pkg/agent/baker.go
```diff
@@ -787,5 +787,8 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return ""
 		},
+		"FIPSEnabled": func() bool {
+			return config.FIPSEnabled
```
mainred, line 791: Linux and windows share the same configuration? and disabling or enabling FIPs will have an impact on both Linux and Windows.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -529,6 +529,17 @@ try
             Remove-Item $kubeConfigFile
         }
 
+        # Enable FIPS-mode
+        $fipsEnabled = [System.Convert]::ToBoolean("{{ FIPSEnabled }}")
+        if ( $fipsEnabled ) {
```
AbelHu, line 534: Could we add a function in `windowsconfigfunc.ps1` and call the function here? We will introduce new Windows version soon and it may need to check os version and then set different configurations.
1. Put `$fipsEnabled = [System.Convert]::ToBoolean("{{ FIPSEnabled }}")` before `try`
2. Call `Set-FIPS -Enable $fipsEnabled` in line 532

pkg/agent/baker.go
chengliangli0918: As we chat on Teams just now, 'FIPSEnabled' is an AgentPoolProfile property, not an AKS cluster level property at AKS-RP side, so FIPSEnabled (disabling or enabling) has impact only to the nodes of this AgentPool.

mainred (approved):

parts/windows/kuberneteswindowssetup.ps1
chengliangli0918: Move code to set the registry to a separate function, and move '$fipsEnabled = [System.Convert]::ToBoolean("{{ FIPSEnabled }}") ' before try according to this comment.

AbelHu (approved):

PR #856, opened 2021-05-28
preload some image with cve fixes.

andyliuliming:


andyliuliming:
@xuto2 @adilliadil looks like the test should not be working, because the ones in install-dependency.sh and linux-vhd-content-test.sh already out of sync.

xuto2 (approved):

xuto2:
> @xuto2 @adilliadil looks like the test should not be working, because the ones in install-dependency.sh and linux-vhd-content-test.sh already out of sync.

Your change is missing the comma and breaking the test

https://github.com/Azure/AgentBaker/pull/856/files



"1.20.2-hotfix.20210525",        "1.20.5-hotfix.20210505"        "1.20.5-hotfix.20210526",

PR #857, opened 2021-06-01
e2e tests for AgentBaker, basic scenario

anujmaheshwari1:
AgentBaker E2E test suite for basic scenario.


e2e/e2e_test.go
```diff
@@ -0,0 +1,213 @@
...
+				CaCertificate:        string(caCertDecoded),
+				ClientCertificate:    string(clientCertDecoded),
+				APIServerCertificate: string(apiServerCertDecoded),
+				ClientPrivateKey:     string(clientKeyDecoded),
+			},
+		},
+	}
+
+	//Adding a dummy key because we are not actually ssh'ing into the node. 
+	cs.Properties.LinuxProfile.SSH.PublicKeys = []datamodel.PublicKey{{
```
eriksywu, line 164: the SSH profile can be left empty if not needed?

e2e/e2e-script.sh
```diff
@@ -0,0 +1,164 @@
...
+            )
+
+if [[ -z ${tlsbootstrap} ]]; then
+    echo "TLS Bootstrap disabled"
+else
+    addJsonToFile "tlsbootstraptoken" $tlsbootstrap
+fi
+
+# Call AgentBaker to generate CustomData and cseCmd
+go test -v
```
eriksywu, line 119: nit: call out the actual _test.go file for clarity? this is a critical step so the more clear the better

e2e/e2e_test.go
anujmaheshwari1: we need the AdminUsername for the cseCmd which is populated through the SSh profile. And if the AdminUsername is define, the logic requires SSH key to be present as well. So I had to insert a dummy value. However, I can try applying the cseCmd without the AdminUsername.

e2e/e2e-script.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
+
+echo "Starting e2e tests"
+
+# TODO 1: Probably move the helper functions to some other file to imrpove readability
```
xuto2, line 5: yes  let's do that

e2e/e2e-script.sh
```diff
@@ -0,0 +1,181 @@
...
+
+addJsonToFile() {
+    k=$1; v=$2
+    jq -r --arg key $k --arg value $v '. + { ($key) : $value}' < fields.json > dummy.json && mv dummy.json fields.json
```
xuto2, line 9: make fields.json an argument?

e2e/e2e-script.sh
```diff
@@ -0,0 +1,181 @@
...
+    declare -a properties=("mode" "name" "nodeImageVersion")
+
+    for property in "${properties[@]}"; do
+        value=$(az aks show -n $CLUSTER_NAME -g $RESOURCE_GROUP_NAME | jq -r .agentPoolProfiles[].${property})
+        addJsonToFile $property $value
+    done
+}
+
+getFQDN() {
+    fqdn=$(az aks show -n $CLUSTER_NAME -g $RESOURCE_GROUP_NAME | jq -r '.fqdn')
```
xuto2, line 22: calling az aks show multiple times may have performance issue. We can save the output in a file and keep using the same file for getting properties.

e2e/e2e-script.sh
```diff
@@ -0,0 +1,181 @@
...
+# truncate -s 0 ~/.kube/config
+
+# Create a resource group for the cluster
+if [ $(az group exists -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID) == "false" ]; then
+    echo "Creating resource group"
+    az group create -l $LOCATION -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID
+fi
+
+# Create the aks cluster and get the credentials(kube/config populated) to kubectl 
+if [ -z $(az aks list -g $RESOURCE_GROUP_NAME | jq '.[].name') ]; then
```
xuto2, line 51: should do az aks show -g <> -n <> to check the existence of the cluster only

e2e/e2e-script.sh
```diff
@@ -0,0 +1,181 @@
...
+    echo "Cluster doesnt exist, creating"
+    az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys
+    az aks get-credentials -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME
+fi
+
+MC_RESOURCE_GROUP_NAME=$(az aks show -n $CLUSTER_NAME -g $RESOURCE_GROUP_NAME | jq -r '.nodeResourceGroup')
```
xuto2, line 57: save the az aks show content to a file and reuse it

e2e/e2e-script.sh
xuto2: why do we need this call at all?

e2e/e2e_test.go
```diff
@@ -0,0 +1,213 @@
...
+					StorageProfile: "ManagedDisks",
+					OSType:         datamodel.Linux,
+					AvailabilityProfile: datamodel.VirtualMachineScaleSets,
+					CustomNodeLabels: map[string]string{
+						"kubernetes.azure.com/node-image-version": values.NodeImageVersion,
+						"kubernetes.azure.com/mode":               "system", //values.Mode,
+					},
+					KubernetesConfig: &datamodel.KubernetesConfig{
+						ContainerRuntime: datamodel.Containerd,
+						KubeletConfig: map[string]string{
```
xuto2, line 103: this can be moved to NodeBootstrappingConfiguration

e2e/e2e-script.sh
anujmaheshwari1: that makes a lot of sense since I am calling az aks show ~10 times. should have thought of that, ill update the logic.

e2e/e2e-script.sh
```diff
@@ -0,0 +1,164 @@
...
+# Apply extension to the VM
+az vmss extension set --resource-group $MC_RESOURCE_GROUP_NAME \
+    --name CustomScript \
+    --vmss-name agentbaker-test-vmss \
+    --publisher Microsoft.Azure.Extensions \
+    --protected-settings settings.json \
+    --version 2.0
+
+# Sleep to let the automatic upgrade of the VM finish
+sleep 60s
```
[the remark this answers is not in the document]
anujmaheshwari1: I believe so, it has worked with every test run I tried till now but to be safe, I will increase it to 120.

eriksywu (approved):

PR #862, opened 2021-06-02
have vhd cache multiple versions of runc

eriksywu:


vhdbuilder/packer/packer_source.sh
```diff
@@ -103,3 +103,15 @@ cpAndMode() {
   src=$1; dest=$2; mode=$3
   DIR=$(dirname "$dest") && mkdir -p ${DIR} && cp $src $dest && chmod $mode $dest || exit $ERR_PACKER_COPY_FILE
 }
+
+downloadDebPkgToFile() {
```
xuto2, line 107: shouldn't be here, it should be in cse_helper?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -117,8 +117,21 @@ else
...
+if [[ $OS == $UBUNTU_OS_NAME ]]; then
+  RUNC_VERSIONS="
+  1.0.0-rc92
+  1.0.0-rc95
+  "
+  for RUNC_VERSION in $RUNC_VERSIONS; do
+    downloadDebPkgToFile "moby-runc" ${RUNC_VERSION} ${RUNC_DOWNLOADS_DIR}
+    echo "  - [cached] runc ${RUNC_VERSION}" >> ${VHD_LOGS_FILEPATH}
+  done
+fi
```
xuto2: should this be in ensureRunc() so all the runc install/caching logic stay together?

xuto2: also rc95 is used for docker nodes as well, did Brian confirm it's ok?

vhdbuilder/packer/packer_source.sh
eriksywu: I want to separate VHD building code (i.e here) with cse code since downloading to file is vhd-only

eriksywu: moving to cse_helper

vhdbuilder/packer/install-dependencies.sh
eriksywu: ensureRunc is for installing a specific runc version though - this is to just cache runc binaries during vhd build. i think they should be separate. 

checking w/ brian

eriksywu: docker w/ r95 is good and recommended

PR #870, opened 2021-06-04
Increase the timeout in checking API Server connectivity

AbelHu:
Use the same timeout 10 seconds as that for Linux.
https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_main.sh#L213

    retrycmd_if_failure ${API_SERVER_CONN_RETRIES} 1 10 nc -vz ${API_SERVER_NAME} 443 || time nc -vz ${API_SERVER_NAME} 443 || VALIDATION_ERR=$ERR_K8S_API_SERVER_CONN_FAIL<br class="Apple-interchange-newline">


karataliu (approved):

andyliuliming:
is it possible to do back-off?

andyliuliming (approved):

feiskyer (approved):

AbelHu:
> is it possible to do back-off?

I do not have the data. I do not think that we should add more retries since currently we retries 100. It should be an unknown network issue in canary.

PR #671, opened 2021-03-17
Turn off update node labels

charliedmcb:
We no longer need this as it will be supported within aks instead. Dropping this to remove the kubectl calls on startup.

PR #884, opened 2021-06-11
Turn off update node labels for windows node

mainred:
reference from Linux side work: https://github.com/Azure/AgentBaker/pull/671
Tested resetting sp, and the feature works without the label.


parts/windows/kuberneteswindowsfunctions.ps1
```diff
@@ -215,17 +215,6 @@ function Register-NodeResetScriptTask {
     Register-ScheduledTask -TaskName "k8s-restart-job" -InputObject $definition
 }
 
-function Register-NodeLabelSyncScriptTask {
```
mainred, line 218: The script is included in the sign-script, considering we needs a new release of the script and there's no harm to keep it there, as the first step, we remove the task creation

AbelHu (approved):

PR #891, opened 2021-06-16
Update Azure CNI to v1.4.1

mattstam:
Caching this release: https://github.com/Azure/azure-container-networking/releases/tag/v1.4.1 for linux and windows.

The pipelines: [Networking-Aquarius-Linux-Official-master](https://msazure.visualstudio.com/One/_build?definitionId=144451&_a=summary) and [Azure CNI (signed+unsigned)](https://dev.azure.com/AzureContainerUpstream/Kubernetes/_build?definitionId=74&_a=summary) need to be ran with this release  as well before this goes in.

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -178,6 +178,7 @@ for imageToBePulled in ${ContainerImages[*]}; do
 done
 
 VNET_CNI_VERSIONS="
+1.4.1
 1.4.0
 1.2.7
 "
@@ -189,6 +190,7 @@ done
 
 # merge with above after two more version releases
 SWIFT_CNI_VERSIONS="
+1.4.1
 1.4.0
 1.2.7
 "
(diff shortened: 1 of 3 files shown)


eriksywu:
@mattstam can you also update the components.json file with the new versions? our vhd test will fail otherwise

mattstam:
Thanks @eriksywu, updated that as well.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -178,6 +178,7 @@ for imageToBePulled in ${ContainerImages[*]}; do
 done
 
 VNET_CNI_VERSIONS="
+1.4.1
 1.4.0
```
mattstam, line 182: Going to keep 1.4.0 cached just to avoid any possible complications since that is what is currently in some regions. Will remove once it is completely unused.

mattstam:
cc @paulgmiller for review

eriksywu (approved):

paulgmiller:
already on 1.4.7

PR #894, opened 2021-06-19
Add workaround for networkd config in CBLMariner

hbeberman:
The default systemd-networkd configuration in CBLMariner is matching on the Mellanox virtual function and is waiting for a 2 minute timeout for DHCP to assign it an IP address. This virtual device isn't supposed to get an IP address so it's an unnecessary 2 minute delay during cluster provisioning.

This change updates the default 99-dhcp-en.network config to only match on eth0, which matches the behavior on Ubuntu.

### DIFF
--- a/vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
+++ b/vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
@@ -25,3 +25,7 @@ forceEnableIpForward() {
     net.bridge.bridge-nf-call-iptables = 1
 EOF
 }
+
+networkdWorkaround() {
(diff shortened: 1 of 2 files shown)


vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -25,3 +25,7 @@ forceEnableIpForward() {
     net.bridge.bridge-nf-call-iptables = 1
 EOF
 }
+
```
adilliadil, line 28: Can you also put a comment above the function what the workaround is about, so that the future readers of the code are aware

adilliadil (approved):
minor comment, otherwise LGTM

PR #897, opened 2021-06-21
Stop caching azure npm its only in .4% of linux deployments

paulgmiller:
Alternatively we could keep caching and bumpt to 1.4.2.


paulgmiller:
@andyliuliming  if we go down this route can we remove npm from the egress test?

PR #901, opened 2021-06-24
Cache v1.19.12, v1.20.8 and v1.21.2 in AKS Windows VHD

AbelHu:
James fixed https://github.com/kubernetes/kubernetes/issues/98509 in this PR [Ignore transient errors when gathering stats for Windows Containers in Dockershim by jsturtevant · Pull Request #98510 · kubernetes/kubernetes (github.com)](https://github.com/kubernetes/kubernetes/pull/98510)

They are included in below k8s patch versions.
- v1.19.12 Kubernetes v1.19.12
- v1.20.8 Kubernetes v1.20.8
- v1.21.2 Kubernetes v1.21.2

Cache this in next AKS Windows VHD so we do not need to release a new AKS Windows VHD when AKS supports above versions. We can replace/remove these binaries when AKS supports above versions with hotfix or skip them.

### DIFF
--- a/vhdbuilder/packer/generate-windows-vhd-configuration.ps1
+++ b/vhdbuilder/packer/generate-windows-vhd-configuration.ps1
@@ -130,14 +130,17 @@ $global:map = @{
         "https://acs-mirror.azureedge.net/kubernetes/v1.19.9-hotfix.20210526/windowszip/v1.19.9-hotfix.20210526-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.19.11/windowszip/v1.19.11-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.19.11-hotfix.20210526/windowszip/v1.19.11-hotfix.20210526-1int.zip",
+        "https://acs-mirror.azureedge.net/kubernetes/v1.19.12/windowszip/v1.19.12-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.20.2-hotfix.20210428/windowszip/v1.20.2-hotfix.20210428-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.20.2-hotfix.20210511/windowszip/v1.20.2-hotfix.20210511-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.20.5-hotfix.20210519/windowszip/v1.20.5-hotfix.20210519-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.20.5-hotfix.20210526/windowszip/v1.20.5-hotfix.20210526-1int.zip",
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -130,14 +130,17 @@ $global:map = @{
         "https://acs-mirror.azureedge.net/kubernetes/v1.19.9-hotfix.20210526/windowszip/v1.19.9-hotfix.20210526-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.19.11/windowszip/v1.19.11-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.19.11-hotfix.20210526/windowszip/v1.19.11-hotfix.20210526-1int.zip",
+        "https://acs-mirror.azureedge.net/kubernetes/v1.19.12/windowszip/v1.19.12-1int.zip",
```
mainred, line 133: Do we plan to support these new versions or use hotfixes instead? have you synced with @xuto2 and @palma21 ?

AbelHu: @mainred  Reference the PR's description
```
Cache this in next AKS Windows VHD so we do not need to release a new AKS Windows VHD when AKS supports above versions. We can replace/remove these binaries when AKS supports above versions with hotfix or skip them.
```

mainred: but my question is if these versions will be supported? if not, we may need to remove them. You know, we cache **limited** supported versions.

AbelHu: We will remove them if they are skipped. This is an exception since we have some ICMs which are impacted by the fixed issue.
```
Cache this in next AKS Windows VHD so we do not need to release a new AKS Windows VHD when AKS supports above versions.
```

mainred: If they are for icms, I guess hotfixes for other existing k8s versions need also to be included? because usually, we support more than three versions.

mainred (approved):
Thanks for your change, approve this PR, my concern is not a blocker but a suggestion, let you do the decision.

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
AbelHu: It is not so urgent for us to prepare the hotfix for the old versions. Since these new k8s patch versions include the fix and normally we should support them soon, I cached these k8s versions for nice to have. I want to avoid that we release the next AKS Windows VHD without these k8s versions and then we need to release a new one. k8s package's size is small so it is not a big concern. If it is an image, we should be more serious.

PR #903, opened 2021-06-29
discard unpacked layers to reduce disk space when building containerd VHD

mainred:
Windows containerd VHD building recently hit disk space insufficient issue and containerd is consuming more disk space compared to docker VHD even when docker caches more resources than containerd.

The reason is we are pulling images with default contained configuration with have discard_unpacked_layers  disabled, which _let Containerd stores the content of an image as is (the raw tars/tarballs and image manifests_, and eat 5GB or so disk space. 

We in this PR enable discard_unpacked_layers  when pre-pull images and use crictl to pull images instead which is required as discard_unpacked_layers is a cri option.

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -91,12 +91,10 @@ function Disable-WindowsUpdates {
 function Get-ContainerImages {
     if ($containerRuntime -eq 'containerd') {
         Write-Log "Pulling images for windows server 2019 with containerd"
-        # start containerd to pre-pull the images to disk on VHD
-        # CSE will configure and register containerd as a service at deployment time
-        Start-Job -Name containerd -ScriptBlock { containerd.exe }
         foreach ($image in $imagesToPull) {
+            Write-Log "Pulling image $image"
             Retry-Command -ScriptBlock {
-                & ctr.exe -n k8s.io images pull $image
+                & crictl.exe pull $image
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -166,6 +165,17 @@ function Install-ContainerD {
     $newPaths = [Environment]::GetEnvironmentVariable("Path", [EnvironmentVariableTarget]::Machine) + ";$installDir"
     [Environment]::SetEnvironmentVariable("Path", $newPaths, [EnvironmentVariableTarget]::Machine)
     $env:Path += ";$installDir"
+
+    $containerdConfigPath = [Io.Path]::Combine($installDir, "config.toml")
+    # enabling discard_unpacked_layers allows GC to remove layers from the content store after
+    # successfully unpacking these layers to the snapshotter to reduce the disk space caching Windows containerd images
+    (containerd config default)  | %{$_ -replace "discard_unpacked_layers = false", "discard_unpacked_layers = true"}  | Out-File  -FilePath $containerdConfigPath -Encoding ascii
```
mainred, line 172: If no configuration file is provided, the default configurations will be used. `discard_unpacked_layers` is unique across the setting and is the only difference between the default configuration and the currently used one.
The configuration file will be overwritten during provisioning the windows node and installing containerd

dcantah (approved):

AbelHu (approved):

PR #905, opened 2021-06-30
Adding Krustlet option for Kubelet type

julia-yin:


parts/linux/cloud-init/artifacts/krustlet.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Krustlet
+
+[Service]
+Restart=on-failure
+RestartSec=5s
+Environment=KUBECONFIG=/var/lib/kubelet/kubeconfig
+Environment=KRUSTLET_CERT_FILE=/etc/krustlet/config/krustlet.crt
```
alexeldeib, line 8: We probably want to use the kubelet server TLS certs here: https://github.com/Azure/AgentBaker/blob/f91e734bdad2b692dd03d5dc3b90a3ad68d8d590/parts/linux/cloud-init/artifacts/cse_config.sh#L82-L83

The certs should be created with CSE, so we can just use those paths as-is

alexeldeib:
I think you need to run a `make generate` to update the generated files

pkg/agent/baker.go
```diff
@@ -584,6 +584,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return ""
 		},
+		"IsKrustlet": func() bool {
+			return profile.KubeletType == "Krustlet"
```
eriksywu, line 588: nit: use strings.EqualsFold()

pkg/agent/baker.go
```diff
@@ -585,7 +585,7 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return ""
 		},
 		"IsKrustlet": func() bool {
-			return profile.KubeletType == "Krustlet"
+			return strings.EqualFold(profile.KubeletType, "Krustlet")
```
alexeldeib, line 588: one more nit: please make this a const similar to `datamodel.TempDiskContainerDataDir` above

pkg/agent/baker.go
alexeldeib: was fixed/reverted?

alexeldeib:
we need to add an `installKrustlet/downloadKrustlet` function which downloads and installs the binaries, checkout Azure CNI as an example:

https://github.com/Azure/AgentBaker/blob/8a3a63b8021912266009a7dfc5b318a9fcf4e4a8/parts/linux/cloud-init/artifacts/cse_install.sh#L43
https://github.com/Azure/AgentBaker/blob/8a3a63b8021912266009a7dfc5b318a9fcf4e4a8/parts/linux/cloud-init/artifacts/cse_install.sh#L120

alexeldeib:
and a rebase on master :)

parts/linux/cloud-init/artifacts/krustlet.service
```diff
@@ -0,0 +1,17 @@
...
+Environment=KRUSTLET_CERT_FILE=/etc/kubernetes/certs/kubeletserver.crt
+Environment=KRUSTLET_PRIVATE_KEY_FILE=/etc/kubernetes/certs/kubeletserver.key
+Environment=KRUSTLET_DATA_DIR=/etc/krustlet
+Environment=RUST_LOG=wasi_provider=info,main=info
+Environment=KRUSTLET_BOOTSTRAP_FILE=/etc/krustlet/config/bootstrap.conf
+ExecStart=/usr/local/bin/krustlet-wasi
+
+[Install]
+WantedBy=multi-user.target
+EOF
```
alexeldeib, line 17: erroneous EOF from krustlet copy-pasted over, remove dis

pkg/templates/templates_generated.go
```diff
@@ -1352,6 +1360,18 @@ downloadCNI() {
     retrycmd_get_tarball 120 5 "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" ${CNI_PLUGINS_URL} || exit $ERR_CNI_DOWNLOAD_TIMEOUT
 }
 
+{{- if IsKrustlet }}
+downloadKrustlet() {
+    mkdir -p $KRUSTLET_DOWNLOAD_DIR
```
alexeldeib, line 1365:
```suggestion
    if [ -f "$KRUSTLET_BIN_DIR/krustlet-wasi" ]; then
         echo "Krustlet already installed, no op"
         return
    fi
    mkdir -p $KRUSTLET_DOWNLOAD_DIR
```

pkg/templates/templates_generated.go
```diff
@@ -1352,6 +1360,18 @@ downloadCNI() {
     retrycmd_get_tarball 120 5 "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" ${CNI_PLUGINS_URL} || exit $ERR_CNI_DOWNLOAD_TIMEOUT
 }
 
+{{- if IsKrustlet }}
```
alexeldeib, line 1363: remove this for reuse in VHD build

pkg/templates/templates_generated.go
```diff
@@ -1325,6 +1328,11 @@ K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
 UBUNTU_RELEASE=$(lsb_release -r -s)
 TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
+{{- if IsKrustlet }}
```
alexeldeib, line 1331: remove for reuse in VHD vuild, or redefine vars in packer source

alexeldeib:
likely superceded by #949

pkg/templates/templates_generated.go
alexeldeib: this allows for vanilla ubuntu to download from storage while vhd will use prebaked binaries (best practice is keeping behavior comparable)

PR #912, opened 2021-07-07
MIG partition systemd service

qinchen352:
add a systemd service that does multi-instance GPU partitioning on MIG nodes


pkg/agent/baker.go
```diff
@@ -308,6 +308,12 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return profile.GetKubernetesLabels(normalizeResourceGroupNameForLabel(config.ResourceGroupName),
 				true, config.EnableNvidia, config.FIPSEnabled, config.OSSKU)
 		},
+		"GetGPUInstanceProfile": func() string {
+			return config.GPUInstanceProfile
+		},
+		"IsMIGEnabledNode": func() bool {
+			return config.GPUInstanceProfile != ""
```
adilliadil, line 315: let's have tests in baker_test for MIGNODE = True case

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -64,6 +64,7 @@ API_SERVER_NAME={{GetKubernetesEndpoint}}
 IS_VHD={{GetVariable "isVHD"}}
```
adilliadil, line 64: have we tested these changes in standalone?

adilliadil (commented):
Would be great if we could test these changes in standalone to see with the RP changes you have the solution is working e2e

parts/linux/cloud-init/artifacts/mig-partition.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Apply MIG configuration on Nvidia A100 GPU
+
+[Service]
+Restart=on-failure
+ExecStartPre=/usr/bin/nvidia-smi -mig 1
```
xuto2, line 6: is this intentional or from testing? I mean the "-mig 1" part.

parts/linux/cloud-init/artifacts/mig-partition.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+#TODO: use mig-parted library to do the partition after the issue is fixed
```
xuto2, line 3: add a link  or description to the issue

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -40,6 +40,7 @@ source {{GetCSEInstallScriptDistroFilepath}}
 wait_for_file 3600 1 {{GetCSEConfigScriptFilepath}} || exit $ERR_FILE_WATCH_TIMEOUT
 source {{GetCSEConfigScriptFilepath}}
 
+
```
eriksywu, line 43: nit: remove

eriksywu (approved):

pkg/agent/variables.go
```diff
@@ -120,6 +122,8 @@ func getCSECommandVariables(config *datamodel.NodeBootstrappingConfiguration) pa
 		"sgxNode":                         strconv.FormatBool(datamodel.IsSgxEnabledSKU(profile.VMSize)),
 		"configGPUDriverIfNeeded":         config.ConfigGPUDriverIfNeeded,
 		"enableGPUDevicePluginIfNeeded":   config.EnableGPUDevicePluginIfNeeded,
+		"migNode":                         strconv.FormatBool(datamodel.IsMIGNode(config.GPUInstanceProfile)),
+		"gpuInstanceProfile":              config.GPUInstanceProfile,
```
eriksywu, line 126: is this var used?

eriksywu (approved):

adilliadil (approved):
Please comment or resolve on the earlier comments so that in the future we can go back to it and understand the progress. Otherwise LGTM

pkg/agent/variables.go
qinchen352: migNode is used in cse_cmd.sh

parts/linux/cloud-init/artifacts/mig-partition.service
```diff
@@ -0,0 +1,11 @@
...
+ExecStart=/bin/bash /opt/azure/containers/mig-partition.sh {{GetGPUInstanceProfile}}
+
+[Install]
+WantedBy=multi-user.target
+#EOF
```
alexeldeib, line 11: I don't believe systemd artifact files should contain EOF here, have you tested this? 

I just did this by accident recently and broke systemd, but you might get away with it since you have a leading # (so it could be a comment)

either way, please remove this if you don't know why it's here, I don't think it's necessary.

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -110,6 +110,22 @@ write_files:
...
+{{- if IsMIGEnabledNode}}
+- path: /etc/systemd/system/mig-partition.service
+  permissions: "0644"
+  encoding: gzip
+  owner: root
+  content: !!binary |
+    {{GetVariableProperty "cloudInitData" "migPartitionSystemdService"}}
+
+- path: /opt/azure/containers/mig-partition.sh
+  permissions: "0744"
```
alexeldeib, line 122: nit: 544

write XOR execute is a nice file ownership pattern.

pkg/agent/baker_test.go
```diff
@@ -416,6 +416,16 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 			config.EnableNvidia = true
 			config.EnableRuncShimV2 = true
 			config.KubeletConfig = map[string]string{}
+		}),
+
+		Entry("AKSUbuntu1804 containerd with multi-instance GPU", "AKSUbuntu1804+Containerd+MIG", "1.19.13", func(config *datamodel.NodeBootstrappingConfiguration) {
```
alexeldeib, line 421: nice

pkg/agent/datamodel/types.go
```diff
@@ -1298,6 +1298,7 @@ type NodeBootstrappingConfiguration struct {
 	HTTPProxyConfig                *HTTPProxyConfig
 	KubeletConfig                  map[string]string
 	EnableRuncShimV2               bool
+	GPUInstanceProfile             string
```
alexeldeib, line 1301: for future reference, if you need to plumb a type into agentbaker, it's way easier to do the updates in agentbaker first, because they don't depend on anything from RP -- just make sure the logic is a no-op when the new field hasn't been passed from RP

that way, you can do:
- agentbaker api changes
- agentbaker logic changes (if it's a big feature, nice to separate)
- RP agentbaker vendor with no other changes (will be done for you in weekly vhd/vendor)
- RP api changes + type conversion to agentbaker
- additional RP logic, validation, new field values, etc etc

pkg/agent/variables.go
alexeldeib: hmm...don't we pass all the scripts through go templating anyway? is there a reason for adding the var here vs as a go template of some sort inside the script?

genuinely curious, I don't know the pros/cons and usually add to the templates except for artifacts

alexeldeib: I'd assume both are fine but wanted to check

alexeldeib:
lgtm, if you don't mind fixing the nits + re-run `make` I think this is fine to merge.

qinchen352:
add mig partition systemd

PR #914, opened 2021-07-08
Remove unused fields from type "AgentPoolProfile".

yizhang4321:


pkg/agent/datamodel/types.go
```diff
@@ -616,55 +616,30 @@ type SysctlConfig struct {
...
-	LoadBalancerBackendAddressPoolIDs   []string             `json:"loadBalancerBackendAddressPoolIDs,omitempty"`
-	CustomVMTags                        map[string]string    `json:"customVMTags,omitempty"`
-	DiskEncryptionSetID                 string               `json:"diskEncryptionSetID,omitempty"`
-	UltraSSDEnabled                     *bool                `json:"ultraSSDEnabled,omitempty"`
-	EncryptionAtHost                    *bool                `json:"encryptionAtHost,omitempty"`
-	ProximityPlacementGroupID           string               `json:"proximityPlacementGroupID,omitempty"`
-	CustomKubeletConfig                 *CustomKubeletConfig `json:"customKubeletConfig,omitempty"`
-	CustomLinuxOSConfig                 *CustomLinuxOSConfig `json:"customLinuxOSConfig,omitempty"`
+	Name                   string               `json:"name"`
+	Count                  int                  `json:"count"`
```
marwanad, line 620: question: for this PR, we are only removing parameters that are not populated in `paramsMap` - correct?

I could be wrong but I think `Count` is one of those that get written to the map but aren't actually used anywhere.

yizhang4321: Thanks for the info. Yeah, you are right, this PR only removes the easy, obvious ones. Once this goes in, I'll look into "Count". If as you mentioned it is not used, I'll remove it.

marwanad (approved):

pkg/agent/datamodel/types.go
```diff
@@ -616,55 +616,30 @@ type SysctlConfig struct {
...
+	ScaleSetEvictionPolicy string               `json:"scaleSetEvictionPolicy,omitempty"`
+	StorageProfile         string               `json:"storageProfile,omitempty"`
+	DiskSizesGB            []int                `json:"diskSizesGB,omitempty"`
+	VnetSubnetID           string               `json:"vnetSubnetID,omitempty"`
+	Subnet                 string               `json:"subnet"`
+	Distro                 Distro               `json:"distro,omitempty"`
+	CustomNodeLabels       map[string]string    `json:"customNodeLabels,omitempty"`
+	PreprovisionExtension  *Extension           `json:"preProvisionExtension"`
+	KubernetesConfig       *KubernetesConfig    `json:"kubernetesConfig,omitempty"`
+	AvailabilityZones      []string             `json:"availabilityZones,omitempty"`
```
adilliadil, line 638: I don't think we're using this anywhere are we?

cmd/generate_test.go
```diff
@@ -31,18 +31,11 @@ const (
 				{
 					"availabilityZones": null,
 					"count": 1,
-					"enableAutoScaling": null,
-					"maxCount": null,
-					"minCount": null,
 					"name": "nodepool1",
-					"orchestratorVersion": "1.19.3",
```
adilliadil, line 38: Just to clarify, are we removing agentPoolProfile.orchestratorVersion because we have the same field as properties. orchestratorProfile.orchestratorVersion already present?

adilliadil (commented):
LGTM, let's just make sure the e2es on RP change passes with these changes in Agentbaker

cmd/generate_test.go
yizhang4321: The orchestratorVersion version in AgentPoolProfile is never used in AgentBaker, thus I removed it.

pkg/agent/datamodel/types.go
yizhang4321: Thanks for the info. I will go through them and remove unused ones in future PRs.

PR #920, opened 2021-07-09
Remove unused fields from AgentPoolProfile.

yizhang4321:
Removed the following fields:
    AvailabilityZones, Subnet, DiskSizesGB, ScaleSetEvictionPolicy, ScaleSetPriority, and OSDiskSizeGB.
Some of them were put into the parameters or variables map, but those parameters or variables were never used.

### DIFF
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -616,29 +616,23 @@ type SysctlConfig struct {
 
 // AgentPoolProfile represents an agent pool definition
 type AgentPoolProfile struct {
-	Name                   string               `json:"name"`
-	VMSize                 string               `json:"vmSize"`
-	OSDiskSizeGB           int                  `json:"osDiskSizeGB,omitempty"`
-	KubeletDiskType        KubeletDiskType      `json:"kubeletDiskType,omitempty"`
-	DNSPrefix              string               `json:"dnsPrefix,omitempty"`
-	OSType                 OSType               `json:"osType,omitempty"`
-	Ports                  []int                `json:"ports,omitempty"`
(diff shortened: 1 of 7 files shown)


pkg/agent/datamodel/types.go
```diff
@@ -616,29 +616,23 @@ type SysctlConfig struct {
...
+	AvailabilityProfile   string               `json:"availabilityProfile"`
+	StorageProfile        string               `json:"storageProfile,omitempty"`
+	VnetSubnetID          string               `json:"vnetSubnetID,omitempty"`
+	Distro                Distro               `json:"distro,omitempty"`
+	CustomNodeLabels      map[string]string    `json:"customNodeLabels,omitempty"`
+	PreprovisionExtension *Extension           `json:"preProvisionExtension"`
+	KubernetesConfig      *KubernetesConfig    `json:"kubernetesConfig,omitempty"`
+	VnetCidrs             []string             `json:"vnetCidrs,omitempty"`
+	WindowsNameVersion    string               `json:"windowsNameVersion,omitempty"`
+	CustomKubeletConfig   *CustomKubeletConfig `json:"customKubeletConfig,omitempty"`
```
adilliadil, line 634: I know we need these two fields, but in RP PR the agentbaker datamodel did not contain these two fields, I may have looked at an older PR though.

adilliadil (approved):
Left a comment, otherwise LGTM/

pkg/agent/datamodel/types.go
yizhang4321: Thanks for your review, Adil. Are you referring to fields "WindowsNameVersion" and "CustomKubeletConfig"? I found both of them in our RP code.

PR #926, opened 2021-07-13
Add an AgentBaker interface for RP to use.

yizhang4321:
This is the preparation step to separate AgentBaker into a standalone service. RP will call the API provided by AgentBaker to get CustomData, CSE, and VHD image to provision nodes.
We'll use protocol buffer and gRPC for the API definition. For now, use straight Go interface as API to help testing the end to end flow.

### DIFF
--- a/pkg/agent/bakerapi.go
+++ b/pkg/agent/bakerapi.go
@@ -5,6 +5,7 @@ package agent
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/Azure/agentbaker/pkg/agent/datamodel"
 )
@@ -27,10 +28,13 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context,
 		CSE:        templateGenerator.GetNodeBootstrappingCmd(config),
 	}
 
-	if osImageConfigMap, hasCloud := datamodel.AzureCloudToOSImageMap[config.CloudSpecConfig.CloudName]; hasCloud {
-		if osImageConfig, hasImage := osImageConfigMap[config.AgentPoolProfile.Distro]; hasImage {
-			nodeBootstrapping.OSImageConfig = osImageConfig
-		}
+	osImageConfigMap, hasCloud := datamodel.AzureCloudToOSImageMap[config.CloudSpecConfig.CloudName]
+	if !hasCloud {
+		return nil, fmt.Errorf("don't have settings for cloud %s", config.CloudSpecConfig.CloudName)
+	}
+
+	if osImageConfig, hasImage := osImageConfigMap[config.AgentPoolProfile.Distro]; hasImage {
+		nodeBootstrapping.OSImageConfig = osImageConfig
 	}
 
 	return nodeBootstrapping, nil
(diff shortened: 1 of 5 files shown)


pkg/agent/bakerapi.go
```diff
@@ -0,0 +1,37 @@
...
+
+package agent
+
+import (
+	"context"
+
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+)
+
+type AgentBaker interface {
```
eriksywu, line 12: does the agentbaker pkg itself need to define the interface contract that it implements? In go it's more common to define the interface at the consuming pkg (i.e our backend rp code)

pkg/agent/bakerapi.go
```diff
@@ -0,0 +1,37 @@
...
+		CSE:        templateGenerator.GetNodeBootstrappingCmd(config),
+	}
+
+	if osImageConfigMap, hasCloud := datamodel.AzureCloudToOSImageMap[config.CloudSpecConfig.CloudName]; hasCloud {
+		if osImageConfig, hasImage := osImageConfigMap[config.AgentPoolProfile.Distro]; hasImage {
+			nodeBootstrapping.OSImageConfig = osImageConfig
+		}
+	}
+
+	return nodeBootstrapping, nil
```
eriksywu, line 36: do we need to return an error if the input distro doesn't have a corresponding AzureCloudOSImage?

pkg/agent/datamodel/osimageconfig.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	RHELOSImageConfig = AzureOSImageConfig{
+		ImageOffer:     "RHEL",
+		ImageSku:       "7.3",
+		ImagePublisher: "RedHat",
+		ImageVersion:   "latest",
+	}
+
+	CoreOSImageConfig = AzureOSImageConfig{
```
eriksywu, line 35: I doubt this os config is used at all

pkg/agent/bakerapi.go
eriksywu: actually do we need to move SIG configs here as well? Will that be done in a separate PR?

pkg/agent/bakerapi.go
yizhang4321: In our code base, do we have an example with interface being defined at the consuming side? I saw more examples with interfaces defined at the library code. For example, our core/clients defined all the interfaces. Another example is the Azure SDK for Go library we use: it defines the interfaces: https://github.com/Azure/azure-sdk-for-go/blob/main/services/compute/mgmt/2021-03-01/compute/computeapi/interfaces.go

pkg/agent/bakerapi.go
yizhang4321: I changed the code to return an error when cloud is not found.

Not returning an error if image is not found yet because for some cases the image is only found in SIG.

I'll prepare a new PR to move over the SIG settings. When that is done, if the image is not found in both places, I'll make the function return an error.

pkg/agent/datamodel/osimageconfig.go
yizhang4321: Yeah, you are right. I just copied the whole thing. CoreOS has been removed now.

eriksywu (approved):

PR #938, opened 2021-07-14
Copy over sig_config code to AgentBaker.

yizhang4321:


pkg/agent/datamodel/sig_config.go
```diff
@@ -0,0 +1,417 @@
...
+	GalleryName   string `json:"galleryName"`
+	ResourceGroup string `json:"resourceGroup"`
+}
+
+type SigImageConfigOpt func(*SigImageConfig)
+
+// GetCloudTargetEnv determines and returns whether the region is a sovereign cloud which
+// have their own data compliance regulations (China/Germany/USGov) or standard
+// Azure public cloud or air gapped clouds
+// NOTE(tonyxu)
```
eriksywu, line 42: should we remove some of these comments here

eriksywu (approved):

pkg/agent/datamodel/sig_config.go
yizhang4321: Yeah, sure. I removed those comments.

PR #944, opened 2021-07-15
Update Windows VHD to 17763.2061.210714

yizhang4321:


pkg/agent/datamodel/sig_config.go
```diff
@@ -201,7 +201,7 @@ const (
 
 const (
 	LinuxSIGImageVersion   string = "2021.07.10"
-	WindowsSIGImageVersion string = "17763.1999.210609"
+	WindowsSIGImageVersion string = "17763.2061.210714"
```
AbelHu, line 204: We have two images, one is docker and another is containerd. The versions may be different

yizhang4321: Do I need to make some change to this PR? This change is based on your RP pull request that updates Windows VHD version. I don't see two versions being specified there.

AbelHu (approved):

PR #949, opened 2021-07-19
Julia/krustlet vhd

julia-yin:


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -13,6 +13,11 @@ K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
 UBUNTU_RELEASE=$(lsb_release -r -s)
 TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
+{{- if IsKrustlet }}
```
alexeldeib, line 16: remove this for vhd build

parts/linux/cloud-init/artifacts/krustlet.service
```diff
@@ -0,0 +1,17 @@
...
+Environment=KRUSTLET_CERT_FILE=/etc/kubernetes/certs/kubeletserver.crt
+Environment=KRUSTLET_PRIVATE_KEY_FILE=/etc/kubernetes/certs/kubeletserver.key
+Environment=KRUSTLET_DATA_DIR=/etc/krustlet
+Environment=RUST_LOG=wasi_provider=info,main=info
+Environment=KRUSTLET_BOOTSTRAP_FILE=/etc/krustlet/config/bootstrap.conf
+ExecStart=/usr/local/bin/krustlet-wasi
+
+[Install]
+WantedBy=multi-user.target
+EOF
```
alexeldeib, line 17: still needs to be removed

pkg/agent/datamodel/const.go
```diff
@@ -112,3 +112,8 @@ const (
 	// when KubeletDiskType == TempDisk.
 	TempDiskContainerDataDir = "/mnt/aks/containers"
 )
+
+const (
+	// Krustlet is the kubelet type specified as WasmWasi in the AgentBaker API that can run WASI workloads
+	Krustlet = "WasmWasi"
```
alexeldeib, line 118: dupe of below?

alexeldeib:
few things to disable for docker/kubelet/containerd when using krustlet (or vice versa)

this whole block should only happen if is not krustlet: https://github.com/Azure/AgentBaker/blob/45e0c3bfd552eaaa0935e5f1b92168a46563dc9e/parts/linux/cloud-init/nodecustomdata.yml#L176-L203

same for basically all of this -- docker/containerd should only run when using kubelet, but need to maintain the existing logic when doing so: https://github.com/Azure/AgentBaker/blob/45e0c3bfd552eaaa0935e5f1b92168a46563dc9e/parts/linux/cloud-init/nodecustomdata.yml#L286-L510

pkg/agent/datamodel/const.go
alexeldeib, line 118: replaced by wasmwasi

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -13,6 +13,10 @@ K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
 UBUNTU_RELEASE=$(lsb_release -r -s)
 TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
+KRUSTLET_DOWNLOAD_DIR="/opt/KRUSTLET/downloads"
+KRUSTLET_BIN_DIR="/usr/local/bin"
+KRUSTLET_VERSION="v0.7.0"
+KRUSTLET_URL="https://krustlet.blob.core.windows.net/releases/krustlet-${KRUSTLET_VERSION}-linux-amd64.tar.gz"
```
alexeldeib, line 19: with our new binary build, this will be: https://acs-mirror.azureedge.net/krustlet/v0.7.0/linux/amd64/krustlet-wasi, and it won't be a tarball any more, just the binary. 

We need to download the binary, chmod +x it, and then move it into /usr/local/bin/krustlet-wasi

alexeldeib: see https://github.com/Azure/AgentBaker/blob/45e0c3bfd552eaaa0935e5f1b92168a46563dc9e/vhdbuilder/scripts/linux/tool_installs.sh#L12-L38 for a good example (only the bpftrace binary, there is other stuff in /tools you can ignore)

alexeldeib: I went ahead and took a stab at this so we don't have to wait for a fresh vhd build in the morning.

alexeldeib: we'll still need the prod binaries, this is the staging upstream storage account which is for testing only.

sercan will need to approve the prod build in the morning and then we can use the acs-mirror.azureedge.net CDN

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -13,6 +13,10 @@ K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
 UBUNTU_RELEASE=$(lsb_release -r -s)
 TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
+KRUSTLET_DOWNLOAD_DIR="/opt/KRUSTLET/downloads"
+KRUSTLET_BIN_DIR="/usr/local/bin"
+KRUSTLET_VERSION="v0.7.0"
+KRUSTLET_URL="https://acs-mirror.azureedge.net/krustlet/${KRUSTLET_VERSION}/linux/amd64/krustlet-wasi"
```
alexeldeib, line 19: oops -- I duplicated this below. either way is fine.

alexeldeib: ironically, this is the correct prod url, and the one below is the correct staging url.

pkg/agent/testdata/AKSUbuntu1604+Disable1804SystemdResolved=false/CustomData
```diff
@@ -6,7 +6,7 @@ write_files:
...
...FKEFtu5YTGMvvPXm85yl2Avpl5hK4QlTq3CuSMebClWlp4L3OiO8+gJ1BewWfV4IcvznNIHw1V5DXSHaN9KXFavdgPEqnIM8jewhSMljkOfyfYzzwHTqzzeKkLx5ZWtyx6UXzxTpWLB/UmK/GOsv6S+Cy/zf3IA8BSlbDT5+zAuwn/tiqO8kPx+ZfSXIMvqmGiSAaRCW6f8C8xu2Wwl06T2yiVIPlnHd+3MVsvp4Hqwmsr/wY5ljo1qCSLB3IVuC3PudVyWLs66ClCA4m4KRxtfISh704A02eweFCn25PP8NVUmQbAKyD9U6T68+qW4VLOnSTc7v3gk6KutK9DmK2cM45l0UxV4Yf51oQ/Q4XngPLFOjb5XpPA2YeOwhWPD7VOBNXuMg8gZJpPLjqP2G2ygOlv9zxPKkf2hWJ37k3b5OUX4oYkXePya3qRTwrXwdr99LJidGllOSfhxCHlkY+cGC3sVZm8UsiqEqHSZfbFM4+Fv0YXEAVem3KjyD+Ci8hGeYMW8C8gLQURUUBaoSqlbWlU334gXvAXUxsUuqu6PHgBoNOIFWA6R0PwSnPwvZjuq4tiL9o+wH/Oq9ikBCIMuLQF56dywEeQ7voP7o8S+2O3GISkhOsbV5cFfZEFxeAOQFg8YOzeLig2AcrOYTWAQx3DJIMmaTamVvJQS1/hR+gj3fDYCafKfNk8Jm2CKdFPSq5VP0BG4/b0K8nEuxGlo6kkk5ClLtTwu5zP0ollcLP45AlqerOe9XfkV+FtcDfuVAtYiFj/6YVbcSLp8FkO1UQrzFXyIzWC6LZO4o36vCCKyIk+77TUZGHK4Y36UH7xS4hUrEHh5Z2A0ell7h21l8i1NVkQ7TuwSCZxivYpAnUP17lV9a0VEJ2smhzZegNzf8jpbHXos5yd901lVxAS5d2fiMsUkevhAAir7wDHEIB/w/H9BsNi8hXXzm381YFF+m861Ih3rPVvLrfLrGupqnLiVB5GnzC6lLT+Wg633Z5024rkyCPxZcZzV2a93fOUGveDu2Bv305xwx9fzv7CcRqZm/0nSCu45JrjMFfrif+CHIS5BK5sT2BqIZm8/HMza+z9pCsbtN1E6PwuUqmk1Aeip5rvMjYGdcefO1Gw3wlrF8x2LINiRC8PTXSvosUl//8rpEgiW/m765wEMLExubvcq/AwAA//+qKFzppRwAAA==
```
alexeldeib, line 9: I manually checked the diff of this file for sanity: 
```bash
$ diff --color=always -u ./old ./new
--- ./old	2021-07-20 18:47:09.983536998 -0700
+++ ./new	2021-07-20 18:47:00.471231454 -0700
@@ -81,6 +81,8 @@
 ERR_DISBALE_IPTABLES=170 
 
 ERR_MIG_PARTITION_FAILURE=180 
+ERR_KRUSTLET_DOWNLOAD_TIMEOUT=171 
+
 
 OS=$(sort -r /etc/*-release | gawk 'match($0, /^(ID_LIKE=(coreos)|ID=(.*))$/, a) { print toupper(a[2] a[3]); exit }')
 UBUNTU_OS_NAME="UBUNTU"
ace@ace-vm:~/go/src/sigs.k8s.io/cluster-api-provider-azure$ 
```

parts/linux/cloud-init/artifacts/krustlet.service
```diff
@@ -0,0 +1,16 @@
...
+
+[Service]
+Restart=on-failure
+RestartSec=5s
+Environment=KUBECONFIG=/var/lib/kubelet/kubeconfig
+Environment=KRUSTLET_CERT_FILE=/etc/kubernetes/certs/kubeletserver.crt
+Environment=KRUSTLET_PRIVATE_KEY_FILE=/etc/kubernetes/certs/kubeletserver.key
+Environment=KRUSTLET_DATA_DIR=/etc/krustlet
+Environment=RUST_LOG=wasi_provider=info,main=info
+Environment=KRUSTLET_BOOTSTRAP_FILE=/etc/krustlet/config/bootstrap.conf
```
alexeldeib, line 12: hmmm. this seems to have lost several changes from the debugging session Christine and I had? https://github.com/alexeldeib/AgentBaker/blob/334016268770cf08f810414da551e1aa5c5f05eb/parts/linux/cloud-init/artifacts/krustlet.service#L15-L17

alexeldeib: krustlet-fix-ca.sh: https://github.com/alexeldeib/AgentBaker/blob/334016268770cf08f810414da551e1aa5c5f05eb/parts/linux/cloud-init/artifacts/krustlet-fix-ca.sh

we can drop the After= and Require=krustlet-fix-ca.service, and keep the ExecStartPre block instead

parts/linux/cloud-init/artifacts/krustlet.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Krustlet
+ConditionPathExists=/usr/local/bin/krustlet-wasi
+
```
alexeldeib, line 4: can you add one here for the krustlet-fix-ca.sh script too?

alexeldeib: a ConditionPathExists I mean. we don't want the service to try to start without these files.

julia-yin: like this? ConditionPathExists=/usr/local/bin/krustlet-fix-ca.sh

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -173,7 +182,8 @@ write_files:
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "kubeletMonitorSystemdService"}}
 
-{{- if NeedsContainerd}}
+{{- if not IsKrustlet}}
```
alexeldeib, line 185: I did some testing on this and unfortunately it doesn't work super cleanly. Can we reenable containerd and docker for now?

We have a ton of container images prebaked into the VHD. At the end of bootstrapping, we use docker/containerd and a corresponding CLI to delete all of them (we're talking gigabytes of data here, very impactful, but after bootstrap so it doesn't affect provisioning latency).

With krustlet, if we disable containerd/docker, we can't do that cleanup. This would be one reason for example having a separate VHD might have been nice. I think for now, we can just let containerd be installed so we can do cleanup properly, and down the road we can iterate.

alexeldeib: https://github.com/Azure/AgentBaker/blob/a59bc79933e073baa740a55a7f3b006438bcaf5d/parts/linux/cloud-init/artifacts/cse_main.sh#L75 is the cleanup i'm talking about as reference.

pkg/agent/baker_test.go
```diff
@@ -426,6 +426,10 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 			config.AgentPoolProfile.VMSize = "Standard_ND96asr_v4"
 			config.EnableNvidia = true
 			config.GPUInstanceProfile = "mig-3g"
+		}),
+		
+		Entry("AKSUbuntu1804 with krustlet", "AKSUbuntu1804+krustlet", "1.20.7", func(config *datamodel.NodeBootstrappingConfiguration) {
+			config.ContainerService.Properties.AgentPoolProfiles[0].WorkloadRuntime = datamodel.WasmWasi
```
alexeldeib, line 432: can you add: 

```go
config.ContainerService.Properties.AgentPoolProfiles[0].KubernetesConfig = &datamodel.KubernetesConfig{
	ContainerRuntime: datamodel.Containerd,
}
```

we won't use this, but RP will still set it internally and pass it down, and we will need it to generate cloud-init testdata which contains containerd (which we want so cleanup works)

pkg/agent/const.go
```diff
@@ -105,6 +105,7 @@ const (
 	updateNodeLabelsSystemdService            = "linux/cloud-init/artifacts/update-node-labels.service"
 	kubernetesCustomSearchDomainsScript       = "linux/cloud-init/artifacts/setup-custom-search-domains.sh"
 	kubeletSystemdService                     = "linux/cloud-init/artifacts/kubelet.service"
+	krustletSystemdService                    = "linux/cloud-init/artifacts/krustlet.service"
```
alexeldeib, line 108: krustletFixCaScript needs to be added here

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -103,12 +103,21 @@ write_files:
     {{GetVariableProperty "cloudInitData" "reconcilePrivateHostsService"}}
 {{- end}}
 
+ {{- if IsKrustlet}}
+- path: /etc/systemd/system/krustlet.service
+  permissions: "0644"
+  encoding: gzip
+  owner: root
+  content: !!binary |
+    {{GetVariableProperty "cloudInitData" "krustletSystemdService"}}
```
alexeldeib, line 112: krustlet-fix-ca.sh needs to be written to disk as part of cloud init here, you can use 0544 for the permissions (owner read + execute, others read)

pkg/agent/const.go
alexeldeib: https://github.com/Azure/AgentBaker/blob/304ad9410059f42721dcb0f2f68651dc18fb1dd0/pkg/agent/variables.go#L29

krustletFixCa and krustletSystemdService need to be added here too

alexeldeib:
https://github.com/Azure/AgentBaker/blob/304ad9410059f42721dcb0f2f68651dc18fb1dd0/parts/linux/cloud-init/artifacts/cse_main.sh#L165

can you move ensureKubelet below both ensureJournal and ensureSysctl, and then wrap it with {{ if IsKrustlet }}:

something like this

```sh
ensureSysctl
ensureJournal
{{- if IsKrustlet}}
systemctlEnableAndStart krustlet
{{- else}}
ensureKubelet
{{- if NeedsContainerd}} {{- if and IsKubenet (not HasCalicoNetworkPolicy)}}
ensureNoDupOnPromiscuBridge
{{- end}} {{- end}}
{{- end}}
```

alexeldeib:
after making the above changes, can you push a tag, checkout a branch from Anchita's RP branch with the API changes, and try to do a standalone build + create a cluster and ensure nodes come up? let me know if you need any help with any part of that.

I think this should be good, but trying it is always the best :)

PR #950, opened 2021-07-20
Capture installed package versions in VHD release notes

hbeberman:
Print out the list of all installed packages and versions at the end of VHD build into the release notes so we can easily lookup which images have which package versions.


eriksywu (approved):

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -57,4 +57,8 @@ cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST
 }
 
+listInstalledPackages() {
```
alexeldeib, line 60: this is never used in CSE right? can we possibly move it to tools-install for VHD build instead?

hbeberman: Good point, moved these functions into the tool_installs_distro scripts.

alexeldeib (approved):
lgtm

anujmaheshwari1 (approved):

PR #953, opened 2021-07-22
extract kube proxy images information into json file

yangzuo0621:
In order to better manage the images version information, and support automatically patch image versions, extract kube proxy images information into separate files.


mainred (approved):
Look good, let but Linux folks know this change for it's not urgent.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -314,46 +315,8 @@ done
 # v1.21.1
```
adilliadil, line 315: we can also remove these comments, even better let's add them as comment field to the json file

andyliuliming:
> Overall I like this change, but why are we creating separate file instead of adding these to the https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/components.json file?

the reason why Zuo put it into a separate file is because if we put it to the components.json.
then I think all the images will be pulled?

vhdbuilder/packer/install-dependencies.sh
adilliadil: Talked to @andyliuliming, we don't want to download all versions since we have some custom logic per version thus having those configs in separate file is more suitable for this scenario

adilliadil (approved):

andyliuliming (approved):

PR #955, opened 2021-07-22
open containerd metric endpoint with 0.0.0.0:10257

mainred:
open metrics endpoint for Windows containerd,

containerd is listening on 10257
```
PS C:\Users\azureuser> Get-NetTCPConnection -Localport 10257

LocalAddress                        LocalPort RemoteAddress                       RemotePort State       AppliedSetting OwningProcess
------------                        --------- -------------                       ---------- -----       -------------- -------------
::                                  10257     ::                                  0          Listen                     3164
127.0.0.1                           10257     127.0.0.1                           19565      Established Datacenter     3164
```

```
PS C:\Users\azureuser> Get-Process -id 3164

Handles  NPM(K)    PM(K)      WS(K)     CPU(s)     Id  SI ProcessName
-------  ------    -----      -----     ------     --  -- -----------
    138      14    23676      27496      12.38   3164   0 containerd
```

request the endpoint
```
PS C:\Users\azureuser> curl http://127.0.0.1:10257/v1/metrics -UseBasicParsing  |  Select-Object -Expand Content |more
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 0
go_gc_duration_seconds{quantile="0.25"} 0
go_gc_duration_seconds{quantile="0.5"} 0
go_gc_duration_seconds{quantile="0.75"} 0
go_gc_duration_seconds{quantile="1"} 0.0020003
go_gc_duration_seconds_sum 0.0058501
go_gc_duration_seconds_count 223
# HELP go_goroutines Number of goroutines that currently exist.
.. ...
```


parts/windows/containerdtemplate.toml
```diff
@@ -14,7 +14,7 @@ state = "C:\\ProgramData\\containerd\\state"
   level = "info"
 
 [metrics]
-  address = ""
+  address = "0.0.0.0:10257"
   grpc_histogram = false
```
mainred, line 18: Need to consider whether or not we should turn on or off gRPC histogram metrics. The current setting meets the requirement of AKS need, per my test.

AbelHu: gRPC histogram metrics may be very expensive for Prometheus to retain and query so I think we should disable it.

mainred: Agree and thanks, let me close this and keep grpc_histogram  as false.

parts/windows/containerdtemplate.toml
```diff
@@ -14,7 +14,7 @@ state = "C:\\ProgramData\\containerd\\state"
   level = "info"
 
 [metrics]
-  address = ""
+  address = "0.0.0.0:10257"
```
AbelHu, line 17: Q: Who will consume this, metric server and omsagent?

mainred: So far, we use prometheus service discovery to find Kubernetes nodes and replace the metric URL with a proxied one.

mainred: @keikhara, Do we have any concerns from the point of security, as we expose 0.0.0.0 and 10257?  Linux is using this endpoint for containerd since June.

AbelHu (approved):

PR #971, opened 2021-07-27
add kubelet start time

haitch:
- also contain it in a map
- execution duration will be slower, but it match the real world case, current calculation is cse finish time - kubelet start time, while actual execution duration should be cse finish time - cse start time


pkg/agent/datamodel/types.go
```diff
@@ -1749,6 +1748,8 @@ type CSEStatus struct {
 	CSEStartTime string `json:"cseStartTime,omitempty"`
 	// GuestAgentStartTime indicate starttime of GuestAgent, output from systemctl show walinuxagent.service -p ExecMainStartTimestamp
 	GuestAgentStartTime string `json:"guestAgentStartTime,omitempty"`
+	// Data contains key-value pairs passed out from CSE.
+	Data map[string]string `json:"data,omitempty"`
```
jveski, line 1752: Any reason to use a map instead of struct fields?

haitch: I feel too much property is added, I am planning to move the other timestamp into this map as well.

jveski: What do you think about using a nested struct? Something like `CSEStatusStats` or `CSEStatusMetadata`? I'd prefer to use a structured type to document which values can be expected.

haitch: updated to bootDatapoints

jveski (approved):

PR #972, opened 2021-07-27
 feat: Support gMSAv2 in Windows nodes

AbelHu:
AKS RP will set windowsGmsaPackageUrl if gMSAv2 is enabled. Otherwise it will be set to empty.


parts/windows/windowsconfigfunc.ps1
```diff
@@ -196,3 +196,133 @@ function Enable-FIPSMode
...
+
+    # Enable the logging manifest.
+    Write-Log "Importing the CCGEvents manifest file"
+    wevtutil.exe im "$tempInstallPackageFoler\CCGEvents.man"
+    if($LASTEXITCODE) {
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GMSA_IMPORT_CCGEVENTS -ErrorMessage "Failed to import the CCGEvents.man manifest file."
+    }
+
+    # Enable the PowerShell privilege to set the registry permissions.
+    Write-Log "Enabling the PowerShell privilege"
```
mainred, line 281: Curious, will setting PowerShell privilege change the behaviors of the following commands outside of `Install-GmsaPlugin`

parts/windows/windowsconfigfunc.ps1
```diff
@@ -196,3 +196,133 @@ function Enable-FIPSMode
...
+        $acl = $key.GetAccessControl()
+        $acl.SetOwner($owner)
+        $key.SetAccessControl($acl)
+        
+        $key = [Microsoft.Win32.Registry]::LocalMachine.OpenSubKey(
+            $ccgKeyPath,
+            [Microsoft.Win32.RegistryKeyPermissionCheck]::ReadWriteSubTree,
+            [System.Security.AccessControl.RegistryRights]::ChangePermissions)
+        $acl = $key.GetAccessControl()
+        $rule = New-Object System.Security.AccessControl.RegistryAccessRule(
```
mainred, line 306: curious, PropagationFlags and AccessControlType are now used?

parts/windows/windowsconfigfunc.ps1
```diff
@@ -196,3 +196,133 @@ function Enable-FIPSMode
...
+            $owner,
+            [System.Security.AccessControl.RegistryRights]::FullControl,
+            [System.Security.AccessControl.AccessControlType]::Allow)
+        $acl.SetAccessRule($rule)
+        $key.SetAccessControl($acl)
```
mainred, line 311: $key.Close()? 
_Closes the key and flushes it to disk if its contents have been modified._

parts/windows/windowsconfigfunc.ps1
```diff
@@ -196,3 +196,133 @@ function Enable-FIPSMode
...
+        [String] $GmsaPackageUrl
+    )
+
+    if ( $GmsaPackageUrl -eq "" ) {
+        Write-Log "GmsaPackageUrl is not set so skip installing GMSA plugin."
+        return
+    }
+
+    $tempInstallPackageFoler = $env:TEMP
+    $tempPluginZipFile = [Io.path]::Combine($ENV:TEMP, "gmsa.zip")
```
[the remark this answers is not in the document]
AbelHu: This is the name of local temp file. We still use the same naming and format in remote package url

parts/windows/windowsconfigfunc.ps1
```diff
@@ -196,3 +196,133 @@ function Enable-FIPSMode
...
+
+    Write-Log "Getting the GMSA plugin package"
+    DownloadFileOverHttp -Url $GmsaPackageUrl -DestinationPath $tempPluginZipFile
+    Expand-Archive -Path $tempPluginZipFile -DestinationPath $tempInstallPackageFoler -Force
```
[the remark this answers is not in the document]
AbelHu: We do not see any failure in Expand-Archive. We can do it later if we see such failure.

parts/windows/windowsconfigfunc.ps1
AbelHu: It works for the current process to set values under the registry key "System\CurrentControlSet\Control\CCG\COMClasses"

parts/windows/windowsconfigfunc.ps1
AbelHu: Cannot get your point. Could you explain your question more?

parts/windows/windowsconfigfunc.ps1
AbelHu: It works from my testing. This is only used to run `reg.exe import ` in this process so I think that we do not need to save the data.

mainred:
I compared the powershell script used here with the one provided by Harika Vedati , could you please also invite him to take a look?

parts/windows/windowsconfigfunc.ps1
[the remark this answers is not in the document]
mainred: Sorry you did not answer my question, I mean does the gmsa package has a version? if we have a version we need to differentiate the local cache as well.

parts/windows/windowsconfigfunc.ps1
mainred: my comment is based on the comparision from Harika Vedati's installplugin.ps1.

parts/windows/windowsconfigfunc.ps1
mainred: I guess it's like a file.close(), the decision is given to you.

mainred (approved):
Not an expert of PowerShell and windows stack. If you are confident with your PR, please go ahead.

parts/windows/windowsconfigfunc.ps1
AbelHu: If we need to persistent this change for usage after reboot, we definitely need to call it. But we only use it after enabling privilege and do not want to use it after reboot, I think that it is better not to close it.

parts/windows/windowsconfigfunc.ps1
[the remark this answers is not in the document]
AbelHu: After offline sync, we can close this.

parts/windows/windowsconfigfunc.ps1
AbelHu: We use different script for the reference.

PR #976, opened 2021-07-29
Fix bool convert issue

AbelHu:
The old code does not convert it to bool so it will always enable Windows secure TLS by default.
$windowsSecureTlsEnabled = "false"

### DIFF
--- a/parts/windows/kuberneteswindowssetup.ps1
+++ b/parts/windows/kuberneteswindowssetup.ps1
@@ -202,7 +202,7 @@ Expand-Archive scripts.zip -DestinationPath "C:\\AzureData\\"
 $useContainerD = ($global:ContainerRuntime -eq "containerd")
 $global:KubeClusterConfigPath = "c:\k\kubeclusterconfig.json"
 $fipsEnabled = [System.Convert]::ToBoolean("{{ FIPSEnabled }}")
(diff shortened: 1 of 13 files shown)


karataliu (approved):

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -202,7 +202,7 @@ Expand-Archive scripts.zip -DestinationPath "C:\\AzureData\\"
 $useContainerD = ($global:ContainerRuntime -eq "containerd")
 $global:KubeClusterConfigPath = "c:\k\kubeclusterconfig.json"
 $fipsEnabled = [System.Convert]::ToBoolean("{{ FIPSEnabled }}")
-$windowsSecureTlsEnabled = "{{GetVariable "windowsSecureTlsEnabled" }}"
+$windowsSecureTlsEnabled = [System.Convert]::ToBoolean("{{GetVariable "windowsSecureTlsEnabled" }}");
```
mainred, line 205: nit: why end with `;` when no `;` around this line

mainred (approved):
Left a nit, thanks.

parts/windows/kuberneteswindowssetup.ps1
AbelHu: Both work. I copied it from line 171. You can see above lines ending with ";", for example, line 157 - 177. We can normalize them later.

PR #977, opened 2021-07-30
add krustlet to vhd + cse_install

alexeldeib:


alexeldeib:
note that this is not invoked in cse_main

alexeldeib:
passed ci except windows which should be no-op, merging.

PR #978, opened 2021-07-30
Add support for k8s patch versions: 1.19.12.1 and 1.20.8.1

ganeshkumar5699:
This change adds support for kubernetes patch versions: 1.19.12.1 and 1.20.8.1.


adilliadil (approved):
lgtm

anujmaheshwari1 (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -370,11 +370,11 @@ KUBE_BINARY_VERSIONS="
 1.19.7-hotfix.20210310
 1.19.9-hotfix.20210505
 1.19.11
-1.19.12
+1.19.12.1
```
AbelHu, line 373: Is this the k8s official patch release, or a AKS private patch release? I am curious why the version pattern is different, neither 1.19.12 nor 1.19.12-hotfix.2021mmdd?
Should we also apply it to Windows nodes?

AbelHu (approved):

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: Asked @sozercan and learned that it is a patch release and not a hotfix and that's why it's in this different format. I'm not sure if it's applicable for Windows nodes. @andyliuliming or @sozercan, could you answer this? Thanks!

andyliuliming:
where's the test side change?

PR #997, opened 2021-08-09
Change proxy cert format from DER to PEM

anujmaheshwari1:
The cert format for HTTP Proxy is currently stored in the DER format. update-ca-certificates require that the format be PEM instead of DER. Sources:

1) https://serverfault.com/questions/1051880/ca-certificates-crt-doesnt-include-my-custom-ca-certificate-after-sudo-update
2) https://unix.stackexchange.com/questions/514136/what-certificate-format-does-usr-local-share-ca-certificates-accept
3) https://superuser.com/questions/437330/how-do-you-add-a-certificate-authority-ca-to-ubuntu


ganeshkumar5699 (approved):
lgtm

PR #1007, opened 2021-08-11
chore: Update gmsa

AbelHu:
Store `registerplugin.reg` in this repo. It is used for registering `CCGAKVPlugin.dll`. This file will be downloaded by AKS build pipeline to build `windows-gmsa-ccgakvplugin-[VERSION].zip` with `CCGAKVPlugin.dll`.
We do not copy its content to the Windows PowerShell script because it is related to `CCGAKVPlugin.dll` and may be changed with a new version. And CustomData may exceed the maximum length.


vhdbuilder/scripts/windows/gmsa/registerplugin.reg
```diff
@@ -0,0 +1,32 @@
+Windows Registry Editor Version 5.00
```
mainred, line 1: This file is removed from gmsa package? will the reg file change?

parts/windows/windowsconfigfunc.ps1
```diff
@@ -237,19 +237,10 @@ function Install-GmsaPlugin {
...
-
     # Copy the plugin DLL file.
     Write-Log "Installing the GMSA plugin"
     Copy-Item -Force -Path "$tempInstallPackageFoler\CCGAKVPlugin.dll" -Destination "${env:SystemRoot}\System32\"
 
-    # Enable the logging manifest.
-    Write-Log "Importing the CCGEvents manifest file"
-    wevtutil.exe im "$tempInstallPackageFoler\CCGEvents.man"
-    if ($LASTEXITCODE) {
-        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GMSA_IMPORT_CCGEVENTS -ErrorMessage "Failed to import the CCGEvents.man manifest file."
```
mainred, line 250: Should we keep the exist code WINDOWS_CSE_ERROR_GMSA_IMPORT_CCGEVENTS?

vhdbuilder/scripts/windows/gmsa/registerplugin.reg
AbelHu: For #1, copy the PR description to here
```
Store registerplugin.reg in this repo. It is used for registering CCGAKVPlugin.dll. This file will be downloaded by AKS build pipeline to build windows-gmsa-ccgakvplugin-[VERSION].zip with CCGAKVPlugin.dll.
We do not copy its content to the Windows PowerShell script because it is related to CCGAKVPlugin.dll and may be changed with a new version. And CustomData may exceed the maximum length.
```
For #2, 
Reg file shouldn’t change (From Manoj in the email thread). But in future, it may need to be updated with the change in CCGAKVPlugin.dll. At that time, we will need a different reg file in this folder.

parts/windows/windowsconfigfunc.ps1
AbelHu: Good catch. Since we have not release it, I think that we can remove it and update the error code. Later if we have release a new feature and remove some code, I think that we can remove the error code but should not change any code values.

vhdbuilder/scripts/windows/gmsa/registerplugin.reg
mainred: For #1, because your change,  you use `reg.exe import "$tempInstallPackageFoler\registerplugin.reg" 2>$null 1>$null`, which means the registerplugin.reg is from gmsa package, right?

AbelHu: For #1, we still import the reg file from the gmsa package.
We store this file in AgentBaker so our pipeline can download it in build gmsa package. The gmsa package published by us will only contain two files: CCGAKVPlugin.dll and registerplugin.reg.

mainred: For #2,
Reg file shouldn’t change (From Manoj in the email thread). But in future, it may need to be updated with the change in CCGAKVPlugin.dll. At that time, we will need a different reg file in this folder.

Together with #1, that's why I left a comment here, if previously dll pairs with reg under gmsa package, we should rely on the package-provided dll and reg.

It seems from this change that the reg is not included in the gmsa package, but no idea why the gmsa feature work before this change.


I think I have explained well in this comment, will approve the pr and leave the decision to you,

mainred (approved):

mainred:
tag this PR with `vhd` so that it can show up easily next time we list changes related to the Windows VHD building.

PR #1008, opened 2021-08-11
Remove unused fields from KubernetesConfig/LinuxProfile

anujmaheshwari1:
This PR removes all the fields that have never been referenced outside types.go for the struct KubernetesConfig and LinuxProfile


ganeshkumar5699:
I may not have the full context about this task so I have a few questions about this:

1. How/what sets the default values for these removed fields from KubernetesConfig (before this PR)? I see that the fields are not referenced in the agentbaker code but are they used/set as defaults in rp?
2. Are we confident that these removed fields will not be used in the future? Or should we add them back in later when needed?
3. For KubernetesConfig, did you manually check each field to see if it was referenced elsewhere or needed to be removed? Or was there a faster way to do it?

Thanks in advance for answering the questions!

ganeshkumar5699 (approved):

PR #1011, opened 2021-08-11
chore: add win release notes 17763.2114.210811

mainred:
This PR is created by https://msazure.visualstudio.com/DefaultCollection/CloudNativeCompute/_release?definitionId=716


AbelHu:
I think that we should remove `Allowed security protocols: Tls, Tls11, Tls12` in the release notes. I created a PR for it: https://github.com/Azure/AgentBaker/pull/1013

AbelHu (approved):

PR #506, opened 2020-12-25
chore: Remove secure protocol check which is set within session scope

mainred:
Remove secure protocol check for Windows VHD building test. Secure protocol setting is session scoped.
This test passed somehow for Windows Server 2019, which is unexpected and fails for Windows Server 2004.

PR #1013, opened 2021-08-11
remove unuseful code Set-AllowedSecurityProtocols

AbelHu:
Actually the tls configuration in building Windows VHD does not work as expected. We have disabled insecure TLS protocols in node provisioning so I think that we can remove these code now.


mainred (approved):
agree, had a PR https://github.com/Azure/AgentBaker/pull/506 but did not move forward.

_Remove secure protocol check for Windows VHD building test. Secure protocol setting is session scoped._

PR #1015, opened 2021-08-12
remove client-go dep; cleanup aks-engine legacy code

alexeldeib:


vhdbuilder/packer/test/run-pretest.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
alexeldeib, line 1: I don't think this was actually necessary, I just needed to force git to change this file to LF (which was failing CI with CRLF). I should just fix my git config, but usr/bin/env is slightly more unix-y

versioning.mk
```diff
@@ -3,14 +3,8 @@ GIT_SHA    = $(shell git rev-parse --short HEAD)
 GIT_TAG    = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null || echo "canary")
 GIT_DIRTY  = $(shell test -n "`git status --porcelain`" && echo "dirty" || echo "clean")
 
-LDFLAGS += -X github.com/Azure/agentbaker/cmd.BuildSHA=${GIT_SHA}
```
alexeldeib, line 6: files no longer exist, so variables are pointless

marwanad (approved):

PR #1016, opened 2021-08-12
Add lsof and pstree tools to CBLMariner VHD

hbeberman:
lsof and pstree are missing on the current CBLMariner VHDs causing some log-collection scripts to fail.
This commit adds those two tools into the CBLMariner VHD.


ganeshkumar5699 (approved):
Lgtm

alexeldeib (approved):
lgtm

might be worth adding a unit test for the generated Mariner data similar to what exists for other combinations in testdata folder? not a big deal

PR #1024, opened 2021-08-17
Add hotfixes for 1.20.7 and 1.21.2

ganeshkumar5699:
Adding latest CVE fixes.

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -205,8 +205,10 @@ testKubeBinariesPresent() {
   1.19.13
   1.20.2-hotfix.20210310
   1.20.5-hotfix.20210322
+  1.20.7-hotfix.20210816
   1.20.8
   1.20.9
+  1.21.2-hotfix.20210816
   "
   for patchedK8sVersion in ${k8sVersions}; do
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -205,8 +205,10 @@ testKubeBinariesPresent() {
   1.19.13
   1.20.2-hotfix.20210310
   1.20.5-hotfix.20210322
+  1.20.7-hotfix.20210816
```
alexeldeib, line 208: so we weren't previously testing either of these versions in vhd content test? oops

alexeldeib:
lgtm pending the make issue? the error output doesn't have much, might need to check the make target and manually debug it a tiny bit

vhdbuilder/packer/test/linux-vhd-content-test.sh
ganeshkumar5699: Yes, we were not and I was surprised by that too.

ganeshkumar5699:
> lgtm pending the make issue? the error output doesn't have much, might need to check the make target and manually debug it a tiny bit

That was a transient issue. Checked with @anujmaheshwari1 and re-ran.

alexeldeib (approved):
Lgtm

PR #1025, opened 2021-08-18
Add previous versions: 1.20.7 and 1.21.2 for caching

ganeshkumar5699:
Adding previous k8s versions present before PR (https://github.com/Azure/AgentBaker/pull/1024) for caching.


alexeldeib (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -378,10 +378,12 @@ KUBE_BINARY_VERSIONS="
 1.19.13
 1.20.2-hotfix.20210310
 1.20.5-hotfix.20210505
+1.20.7
```
alexeldeib, line 381: @AbelHu fyi, the content check fails for these versions because we normalize to MAJOR.MINOR.PATCH, so the hotfix versions overwrite the patch versions. I believe removing this was correct.

AbelHu: @alexeldeib Thanks for involving me. When a hotfix version `1.20.7-hotfix.20210816` is introduced and 1.20.7 is never used before, I think that we should only add `1.20.7-hotfix.20210816`.
I am a little confused for `I believe removing this was correct.`. Should we remove it? Why does it look like that this pr is adding it?

alexeldeib (changes requested):
we can close this based on the normalization

vhdbuilder/packer/install-dependencies.sh
alexeldeib: ah sorry I thought you commented on #1024 which originally had the non-hotfix versions. we should only keep the hotfix version, yes. the hotfix would overwrite the older version on disk anyway due to normalization.

PR #1032, opened 2021-08-19
krustlet polish + functionality 

alexeldeib:


Makefile
```diff
@@ -55,7 +55,7 @@ export GO111MODULE=on
 # Add the tools bin to the front of the path
 export PATH := $(TOOLSBIN):$(PATH)
 
-all: build
```
alexeldeib, line 58: removing the cmd/ directory in previous PR broke the go-build target. It's otherwise unused, remove it.

PR #1033, opened 2021-08-19
krustlet polish and functionality, base64 fix

alexeldeib:


alexeldeib:
some notes:
- this leaves containerd running so we can do container image cleanup. we can follow up with removing containerd/images completely from these nodes.
- capacity on krustlet nodes is hardcoded. we should fix that for the sake of scheduling + autoscaling.
- hostnetwork only afaik.
- krustlet doesn't seem to detect external IPs for the node. should fix that.

parts/linux/cloud-init/artifacts/krustlet.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Krustlet
+
+[Service]
+Restart=on-failure
+RestartSec=5s
+EnvironmentFile=/etc/default/kubelet
```
marwanad, line 7: do we need to read any env vars from /etc/default/kubelet?

marwanad: seems like KUBELET_NODE_LABELS comes from there?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -173,7 +182,7 @@ write_files:
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "kubeletMonitorSystemdService"}}
 
-{{- if NeedsContainerd}}
```
marwanad, line 176: curious: is this diff needed?

marwanad: most ignorance on the difference between {{-if}} vs {{if -}. I assume the else clause but just checking for my sanity.

marwanad (approved):
just questions to clarify but looks good to me

parts/linux/cloud-init/artifacts/krustlet.service
alexeldeib: Yep just node labels for now

parts/linux/cloud-init/nodecustomdata.yml
alexeldeib: It mostly helps keep the generated diff cleaner

PR #1048, opened 2021-08-23
Remove unsupported kube binary and kube proxy versions to increase space

ganeshkumar5699:
VHDs for certain OS SKUs are about 97-99% full, which leads to build errors. Removing these versions from `kubectl` and `kube-proxy` to increase available space.

Checked for versions supported in AKS (`eastus`) using: `az aks get-versions -l eastus -o table`.


ganeshkumar5699:
Comment from `install-dependencies.sh`:

```
# kubelet and kubectl
# need to cover previously supported version for **VMAS scale up scenario**
# So keeping as many versions as we can - those unsupported version can be removed when we don't have enough space
```
What is the VMAS scale up scenario? And why would unsupported versions be useful for that?

alexeldeib:
versions lgtm

my understanding of the VMAS scale case is for unsupported versions, removing the images induces pull latency. However that’s the same for VMSS. Removing the binaries would be similar (might just force the download, not sure of the top of my head).

Another possibility is since each VM is an individual resource, each node ends up with the latest VHD. That would mean scale operations could fail/induce latency if the binaries arent there.

I’m in favor of merging this since we need the space and these versions are well out of support

mainred:
> my understanding of the VMAS scale case is for unsupported versions, removing the images induces pull latency. However that’s the same for VMSS. Removing the binaries would be similar (might just force the download, not sure of the top of my head).

Agree. I think our earlier try is to do make cx happy even their k8s version is out of support.

As regards to the versions supported, I think we can keep the caches related to k8s versions 30 days after the version goes to the end per [AKS release and deprecation process](https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#release-and-deprecation-process) as follows:
_Users have 30 days from version removal to upgrade to a supported minor version release to continue receiving support._


reference: [aks-kubernetes-release-calendar](https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#aks-kubernetes-release-calendar)

AbelHu:
> versions lgtm
> 
> my understanding of the VMAS scale case is for unsupported versions, removing the images induces pull latency. However that’s the same for VMSS. Removing the binaries would be similar (might just force the download, not sure of the top of my head).
> 
> Another possibility is since each VM is an individual resource, each node ends up with the latest VHD. That would mean scale operations could fail/induce latency if the binaries arent there.
> 
> I’m in favor of merging this since we need the space and these versions are well out of support

I think that we should balance them. If a version is not supported, I think that we can remove the cache. Cx's nodes still can download the image/binaries from the internet although there is some latency. If cx does not want to hit this issue, they should upgrade their cluster to AKS supported versions.

vhdbuilder/packer/kube-proxy-images.json
```diff
@@ -52,19 +33,12 @@
       {
         "downloadURL": "mcr.microsoft.com/oss/kubernetes/kube-proxy:v*",
         "versions": [
-          "1.19.1-hotfix.20200923",
-          "1.19.1-hotfix.20200923.1",
+          "1.19.9-hotfix.20210526.2",
```
AbelHu, line 36: Should the latest hotfix version `1.19.9-hotfix.20210526.3` be always used if the cx has updated the node image to the latest one? If so, we can only keep `1.19.9-hotfix.20210526.3`

ganeshkumar5699:
Thanks for the answers and reviews @alexeldeib , @mainred and @AbelHu.

vhdbuilder/packer/kube-proxy-images.json
ganeshkumar5699: I believe so but I want to confirm before removing. Perhaps @palma21  or @qpetraroia can help answer the requirement based on AKS release policies/users' expectations?

alexeldeib: Should be fine to follow up on this

alexeldeib (approved):

ganeshkumar5699:
Merging despite Windows Build failures in the Pipeline as the PR changes should not affect it.

vhdbuilder/packer/kube-proxy-images.json
palma21: Yes, that is the policy/expectation.

PR #1050, opened 2021-08-23
move all cse variables to files baked in cloud init

alexeldeib:
this makes a runtime node a clean copy we can snapshot and run the root cse script against with no component version changes. it's a nice feature for node snapshot/custom vhd/whatever


anujmaheshwari1 (approved):

alexeldeib:
unsure if I want to merge this as-is, or maybe work on https://github.com/Azure/AgentBaker/pull/1060 as a more aggressive approach

alexeldeib:
probably close for now...

PR #1054, opened 2021-08-24
cleanup: remove unused/unpopulated variables

alexeldeib:
these values are never populated and totally unused as far as I can tell

taints are passed from RP into KUBELET_FLAGS, this is legacy from aks-engine
register-node is legacy from aks-engine


anujmaheshwari1 (approved):

ganeshkumar5699:
Just to make sure: was KUBELET_FLAGS intended to be used for the kubeletconfig feature (that is supposed to be enabled in the future)? I remember that the kubelet config use feature was disabled after issues earlier but it does not seem KUBELET_FLAGS variable is related to that.

alexeldeib:
> Just to make sure: was KUBELET_FLAGS intended to be used for the kubeletconfig feature (that is supposed to be enabled in the future)? I remember that the kubelet config use feature was disabled after issues earlier but it does not seem like the KUBELET_FLAGS variable is related to that.

KUBELET_FLAGS is how we pass a ton of flags today. The ones I'm eliminating are already part of KUBELET_FLAGS from RP.

Inside agentbaker, we split out the flags compatible with kubelet file configuration and remove them from KUBELET_FLAGS. we still need both sets because kubelet's config file doesn't have all flags yet.

PR #1055, opened 2021-08-24
chore: move containerd flags from go tmpl to drop in

alexeldeib:
removing go template from these files lets us do more things at runtime on the node, vs from RP side. using drop ins is a nice alternative for composability, even though we're still relying on go tempates to inject them via cloud init rn.


marwanad (approved):

pkg/agent/testdata/AKSUbuntu1604+Disable1804SystemdResolved=false/line66.sh
```diff
@@ -19,10 +19,10 @@ ExecStartPre=-/sbin/iptables -t nat --numeric --list
 ExecStart=/usr/local/bin/kubelet \
         --enable-server \
         --node-labels="${KUBELET_NODE_LABELS}" \
-        --v=2  \
+        --v=2 \
         --volume-plugin-dir=/etc/kubernetes/volumeplugins \
-        $KUBELET_FLAGS \
-        $KUBELET_REGISTER_NODE $KUBELET_REGISTER_WITH_TAINTS
```
alexeldeib, line 25: 🤔 should've been removed by previous PR, awkward

PR #1101, opened 2021-09-02
adding omsagnet image ciprod08052021-1

daweim0:
Title says it all. The only difference between ciprod08052021 and ciprod08052021-1 is the image tag

@ganga1980 @vishiy @rashmichandrashekar @gracewehner

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -144,7 +144,8 @@
         "ciprod06112021",
         "ciprod05202021-hotfix",
         "ciprod06112021-1",
-        "ciprod08052021"
+        "ciprod08052021",
+        "ciprod08052021-1"
       ]
     },
     {


ganga1980 (approved):

vhdbuilder/packer/components.json
```diff
@@ -144,7 +144,8 @@
         "ciprod06112021",
         "ciprod05202021-hotfix",
         "ciprod06112021-1",
-        "ciprod08052021"
+        "ciprod08052021",
```
ganeshkumar5699, line 147: Is it ok to have this older version? And can that be removed since it will not be used?

anujmaheshwari1 (approved):
lgtm pending all VHDs build successfully!

chengliangli0918 (approved):

vhdbuilder/packer/components.json
ganga1980: yes, ciprod08052021 ok to have and we will remove old ones next agent release.

daweim0: The older version will still be used. We only plan to use the newer version if there is an issue toggling a few customers to the new version

ganeshkumar5699 (approved):
Lgtm. And thanks for the info about the previous version and the plan to remove the older versions.

ganga1980 (approved):

juan-lee (approved):

PR #1109, opened 2021-09-03
retag the images with mooncake mcr for the vhd based on containerd too.

andyliuliming:


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -427,7 +427,20 @@ sed -i 's/After=network-online.target/After=multi-user.target/g' /lib/systemd/sy
 
 # retag all the mcr for mooncake
 # shellcheck disable=SC2207
-allMCRImages=($(docker images | grep '^mcr.microsoft.com/' | awk '{str = sprintf("%s:%s", $1, $2)} {print str}'))
+if [[ ${cliTool} == "ctr" ]]; then
+  # shellcheck disable=SC2016
+  allMCRImages=($(ctr --namespace k8s.io images list | grep '^mcr.microsoft.com/' awk '{print $1}'))
+elif [[ ${cliTool} == "crictl" ]]; then
```
AbelHu, line 433: when is cliTool set to crictl? I cannot find it in this file

andyliuliming: there's one place like this:
{{- if NeedsContainerd}}
# If crictl gets installed then use it as the cri cli instead of ctr
# crictl is not a critical component so continue with boostrapping if the install fails
# CLI_TOOL is by default set to "ctr"
installCrictl && CLI_TOOL="crictl"
{{- end}}

andyliuliming: oh, but there's only ctrl in install dependency, will remove this if.

AbelHu (approved):

PR #1120, opened 2021-09-03
Use containerd 1.5.5

ganeshkumar5699:
Updating for k8s version 1.22.


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -98,7 +98,7 @@ installStandaloneContainerd() {
     
     #if there is no containerd_version input from RP, use hardcoded version
     if [[ -z ${CONTAINERD_VERSION} ]]; then
-        CONTAINERD_VERSION="1.4.8"
+        CONTAINERD_VERSION="1.5.5"
```
alexeldeib, line 101: we need to gate this on the actual kubernetes version, I think we can use bash + KUBERNETES_VERSION or else use Go templates.

ganeshkumar5699: Ok. So if version < 1.22, we use containerd 1.4.8 and if it's >= to it then use containerd 1.5.5?
Is the latest containerd not backwards compatible with the earlier k8s versions?

alexeldeib:
please also bump here: https://github.com/Azure/AgentBaker/blob/2b4fc07296196f7ddc26ad9cbd1bdc04f23cc18c/vhdbuilder/packer/install-dependencies.sh#L96

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
alexeldeib: it's not a compatibility issue, I think that's just how PM want to ship the change (containerd 1.5 + k8s 1.22 as a coupled pair). clusters with older k8s versions should keep what they have (barring hotfixes)

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -95,10 +95,17 @@ installStandaloneContainerd() {
     # azure-built runtimes have a "+azure" suffix in their version strings (i.e 1.4.1+azure). remove that here.
     CURRENT_VERSION=$(containerd -version | cut -d " " -f 3 | sed 's|v||' | cut -d "+" -f 1)
     # v1.4.1 is our lowest supported version of containerd
-    
+    CURRENT_KUBERNETES_VERSION=$(kubectl -version)
+    echo "Current version of kubernetes used is: $CURRENT_KUBERNETES_VERSION"
     #if there is no containerd_version input from RP, use hardcoded version
+    echo "(for all cases): KUBERNETES_VERSION IS ${KUBERNETES_VERSION}"
```
anujmaheshwari1, line 101: did you mean CURRENT_KUBERNETES_VERSION instead of KUBERNETES_VERSION?
the latter variable isnt defined anywhere.

ganeshkumar5699: I did want KUBERNETES_VERSION there to test the value as it is defined in cse_cmd.sh. The build in the PR ([link](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=46517365&view=logs&j=6e436a35-c994-55aa-033b-9a96e1fbff66&t=fa694a85-6f82-5b17-9708-d28cddf5ec77&l=2844)) says that it is an unbound variable, so the variables in cse_cmd.sh do not seem to be accessible here.

ganeshkumar5699: The CURRENT_KUBERNETES_VERSION variable that I created is currently empty in the build PR. I am trying to find out other ways to access the k8s version.

ganeshkumar5699: Thanks @anujmaheshwari1 for the suggestion to check in cse_install.sh.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -102,6 +102,9 @@ if [[ ${CONTAINER_RUNTIME:-""} == "containerd" ]]; then
     containerd_version="1.4.4"
     downloadContainerd ${containerd_version}
     echo "  - [cached] containerd v${containerd_version}" >> ${VHD_LOGS_FILEPATH}
+    updated_containerd_version="1.5.5"
```
alexeldeib, line 105: nit: this doesn't really need to be a separate variable, or even a variable. but let's leave it as-is if you've already tested this.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -102,6 +102,9 @@ if [[ ${CONTAINER_RUNTIME:-""} == "containerd" ]]; then
     containerd_version="1.4.4"
```
alexeldeib, line 102: leave for now, but let's remove 1.4.4 soon.

alexeldeib (approved):
lgtm if you've tested this with standalone :)

ganeshkumar5699:
Thanks @alexeldeib and @anujmaheshwari1.

The standalone build for default version 1.19.11 has containerd 1.4.8 in var/log/azure/cluster-provision.log (as desired).
RP changes for k8s 1.22 need to be merged to test containerd 1.5.5 change in the standalone environment (as we cannot create a cluster with 1.22 at this time). That will be tested after the 1.22 PR in RP is merged. 

The Linux and Windows VHD checks in this PR passed earlier (before latest run with the base branch update).

PR #1127, opened 2021-09-06
remove bunch of images not in use anymore from vhd.

andyliuliming:


feiskyer (approved):
LGTM

ganeshkumar5699:
Thanks for removing this @andyliuliming. I am curious: how did you determine that all of these were not being used in agentbaker and rp (including recent releases). I see that you kept the latest versions and when there were recent versions in July/August you usually kept them. Beyond the Linux VHD build test, did you check something in rp for this too?

Thanks again for answering my questions.

PR #1133, opened 2021-09-07
Chore: Update NPM to v1.4.9

vakalapa:
Updating Azure NPM image to 1.4.9


vhdbuilder/packer/components.json
```diff
@@ -77,8 +77,7 @@
     {
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-npm:*",
       "versions": [
-        "v1.1.8",
```
paulgmiller, line 80: think these are used by 5 clusters please remove that in aks code as well?

paulgmiller (approved):
clean up 1.1.8 in the toggle too?

andyliuliming (approved):

PR #1134, opened 2021-09-08

andyliuliming:


vhdbuilder/packer/kube-proxy-images.json
```diff
@@ -4,20 +4,20 @@
       {
         "downloadURL": "mcr.microsoft.com/oss/kubernetes/kube-proxy:v*",
         "versions": [
-          "1.19.11-hotfix.20210526.3",
           "1.19.11-hotfix.20210823",
+          "1.19.11-hotfix.20210823.1",
```
mainred, line 8: What's the principle of keeping and removing an old version?

AbelHu (approved):

vhdbuilder/packer/kube-proxy-images.json
```diff
@@ -4,20 +4,20 @@
...
           "1.19.12.1",
-          "1.19.13-hotfix.20210830",
+          "1.19.13-hotfix.20210830.1",
```
mainred, line 10: keeping 1.19.11-hotfix.20210823 while adding 1.19.11-hotfix.20210823.1 to make sure we have one copy of the last image? so that even though new vhd is published but aks-rp code does not change, the cache is still there.
But why we remove 1.19.13-hotfix.20210830 when adding 1.19.13-hotfix.20210830.1?

andyliuliming: this is because we have no vhd release for this week.

vhdbuilder/packer/kube-proxy-images.json
andyliuliming: this PR is special.
all due to we have no vhd release this week.
so we jump the image tag.

PR #1142, opened 2021-09-09
Enable CBLMariner dnf-automatic installs

hbeberman:


vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -42,3 +42,18 @@ listInstalledPackages() {
...
+    # Stop the notify only dnf timer since we've enabled the auto install one.
+    # systemctlDisableAndStop adds .service to the end which doesn't work on timers.
+    systemctl disable dnf-automatic-notifyonly.timer
+    systemctl stop dnf-automatic-notifyonly.timer
+    # At 6:00:00 UTC (1 hour random fuzz) download and install package updates.
+    systemctlEnableAndStart dnf-automatic-install.timer || exit $ERR_SYSTEMCTL_START_FAIL
+    # At 8:000:00 UTC check if a reboot-required package was installed
+    # Touch /var/run/reboot-required if a reboot required pacakge was installed.
+    # This helps avoid a Mariner specific reboot check command in kured.
+    systemctlEnableAndStart check-restart.timer || exit $ERR_SYSTEMCTL_START_FAIL
```
alexeldeib, line 58: this is the newly added package above, but it doesn't enable itself by default? or where is this coming from otherwise?

alexeldeib (approved):
minor q, lgtm. this makes me minorly uncomfortable, but only because our ubuntu strategy also makes me uncomfortable :)

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
hbeberman: Once we switch over to a Marketplace base image I believe that check-restart's going to be installed by default but disabled. Running the systemctlEnableAndStart here to be doubly sure that the timer is running in the AKS images.

PR #1147, opened 2021-09-10
Byo CNI network plugin option

mattstam:
This is the first step for `az aks create … --network-plugin=none` option, which will not install a CNI at cluster creation time. We will also use this to transition AzureCNI / Kubenet to daemonset installers.

For full design doc, please see: https://dev.azure.com/msazure/CloudNativeCompute/_git/CloudNativeCompute.wiki/pullrequest/5104561


pkg/agent/datamodel/types.go
```diff
@@ -1000,6 +1000,14 @@ func (o *OrchestratorProfile) IsAzureCNI() bool {
 	return false
 }
 
+// IsByoCNI returns true if Byo CNI network plugin is enabled
+func (o *OrchestratorProfile) IsByoCNI() bool {
+	if o.KubernetesConfig != nil {
+		return strings.EqualFold(o.KubernetesConfig.NetworkPlugin, NetworkPluginByo)
```
alexeldeib, line 1006: tagentially, can this ever be empty? e.g. for default azure cni?

alexeldeib:
this mostly looks good, lmk if you're ready for a full review? 

have you tested this with standalone?

mattstam:
Yep @alexeldeib a full review would be great!

Tested on a standalone, works as expected in the doc (nodes come up NotReady with no CNI installed), then I applied a Calico installation on the cluster and set it up with VXLAN + Calico IPAM, which seemed to work great.

pkg/agent/datamodel/types.go
mattstam: Nope we're pretty strict about only allowing in either `--network-plugin kubenet` or `--network-plugin azure`. After this is available in AgentBaker, I will make the adjustments in RP to allow `--network-plugin byo`.

mattstam: And if the flag isn't given we'll always default to `kubenet` and set it up like that.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -93,7 +93,9 @@ installContainerRuntime
 installTeleportdPlugin
 {{- end}}
 
+{{- if not IsByoCNI }}
```
alexeldeib, line 96: do you still want the vanilla cni plugins (installCNI inside installAzureCNI) even with byo? or does calico do everything fine itself? 

I assume the latter but just checking

pkg/agent/testdata/AKSUbuntu1604+K8S117/line33.sh
```diff
@@ -75,20 +75,20 @@ else
 fi
 
 installContainerRuntime
-
 installNetworkPlugin
 
+
```
alexeldeib, line 80: nit: any chance you can tweak the go templates to make this diff go away

alexeldeib (approved):
minor comments

parts/linux/cloud-init/artifacts/cse_main.sh
mattstam: Yep the latter, Calico is all good without any CNI installed since their daemonset installs it. There is the interesting case of when we use it for just policies (ontop of azure CNI), but that will be configured in addon-manager.

pkg/agent/testdata/AKSUbuntu1604+K8S117/line33.sh
mattstam: Which templates would that that be? This was just spit out from `make generate` so I assume it should be kept this way otherwise next person to run that will also get weird diffs.

PR #1149, opened 2021-09-13
fix: Use Invoke-RestMethod instead of curl to download packages

AbelHu:
Use Invoke-WebRequest instead of curl.exe to download files
curl.exe does not retry when failed in handshake.

2021-09-08T01:12:44.6532522+00:00: Download kubelet binaries and unzip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:21 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:21 --:--:--     0
curl: (35) schannel: failed to receive handshake, SSL/TLS connection failed

https://github.com/Azure/aks-engine/pull/4646


parts/windows/kuberneteswindowsfunctions.ps1
```diff
@@ -139,11 +141,15 @@ function Retry-Command {
         $RetryDelaySeconds
     )
 
-    for ($i = 0; $i -lt $Retries; $i++) {
+    for ($i = 0; ; ) {
```
andyliuliming, line 144: where's the exit condition?

AbelHu: It will return if success or throw an exception if it fails with max retries.

parts/windows/kuberneteswindowsfunctions.ps1
```diff
@@ -139,11 +141,15 @@ function Retry-Command {
         $RetryDelaySeconds
     )
 
-    for ($i = 0; $i -lt $Retries; $i++) {
+    for ($i = 0; ; ) {
         try {
             return & $Command @Args
         }
         catch {
+            $i++
```
andyliuliming, line 149: curious how to write the unit test for powershell.

andyliuliming (approved):

parts/windows/kuberneteswindowsfunctions.ps1
AbelHu: Good point. We have not written unit tests for cse. I have validated these code in the node directly. Later we will vendor this change to RP and then validate it again.

PR #1178, opened 2021-09-22
chore: removing networkmonitor and azure-vnet-telemetry docker images

pjohnst5:
networkmonitor and azure-vnet-telemtry docker images are no longer used with versions of azure-cni 1.2.0+

Removing them from agent baker

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -67,23 +67,9 @@
       ]
     },
     {
-      "downloadURL": "mcr.microsoft.com/containernetworking/networkmonitor:*",
-      "versions": [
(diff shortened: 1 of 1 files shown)


paulgmiller:
Not sure we want to remove these till we have a PR  in to disable this in new clusters. 
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/3947551

alexeldeib:
> Sigh I think this has to stay till 1.20 gets deprecated....

hello from the future @paulgmiller where 1.20 deprecation is currently rolling to customers!

pjohnst5:
Closing in lieu of #1745  which does the same and is merged

PR #1181, opened 2021-09-23

andyliuliming:


mainred (approved):

vhdbuilder/packer/components.json
```diff
@@ -230,7 +232,8 @@
       "versions": [
         "v1.2.0.5",
         "v1.5.0",
-        "v1.7.0"
+        "v1.7.0",
+        "v1.7.0.1"
```
mainred, line 236: Thought it was "v1.7.1", but after checking we have used "v1.2.0.5", so I think "v1.7.0.1" is the correct one.

PR #1186, opened 2021-09-24
Agent baker as a web service

MirzaSikander:
Exposed agentbaker api as a web service.

Goal:
Eventually we want to be able to run multiple versions of agent baker at the same time each as a separate pod.


apiserver/nodebootstrapconfig.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func handleError(err error, w http.ResponseWriter) {
+	log.Println(err.Error())
+	http.Error(w, err.Error(), http.StatusBadRequest)
+}
+
+// GetNodeBootstrapConfig endpoint for getting node bootstrapping data.
+func (api *APIServer) GetNodeBootstrappingConfig(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
```
alexeldeib, line 25: do you know if there is a default context timeout here? I kind of doubt it, can you add one? can be longish, even 30s would be fine.

alexeldeib (approved):
minor comment

PR #1215, opened 2021-10-04
Fixed error handling.

MirzaSikander:


alexeldeib (approved):
lgtm. I think we can boil this down even further, but let’s get it working first :)

PR #1218, opened 2021-10-05
update npm version to 1.4.12

huntergregory:


vhdbuilder/packer/components.json
```diff
@@ -78,7 +78,7 @@
     {
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-npm:*",
       "versions": [
-        "v1.4.9"
+        "v1.4.12"
```
vakalapa, line 81: Can you keep the old v1.4.9 too ? if we have any problem with 1.4.12 we should be able to roll back to 1.4.9

vakalapa (approved):

vhdbuilder/packer/components.json
paulgmiller: ditto I can merge once you add that back. For a period of time both will be used.

huntergregory:
@paulgmiller I added 1.4.9 back in the above commit. Can you approve?

paulgmiller (approved):

PR #1224, opened 2021-10-07
Shorten outbound connectivity check wait

marwanad:
This PR makes two changes:
- Shorten the retries for outbound connectivity check from 100 to 50 retries.
- Avoid blocking on `nc` command and timeout after 30s. I'm not sure what the intention of the second `if` clause was because on success, we won't print the time. I've repurposed to be a final retry that would tell us if our timeout for the first clause is too aggressive but pretty sure we can just drop it and alternatively print the retry co

Basically, the second or conditional would allow us to uncover "slowness" in case our 10s timeout for the first clause is too long. This gives it a grace period of 30s for one last retry. In the worst case, we'll end up waiting ~10min. 

Planning on following-up with a further shorter number of retries.


marwanad:
😭

PR #1229, opened 2021-10-12
fix: cleanup images for different clouds

alexeldeib:


alexeldeib:
thanks! I also wanted to test in standalone one more time after a final tweak

PR #1269, opened 2021-10-23
 fix: fabric manager + device plugin with mig instances

alexeldeib:
without fabric manager, A100 instances with nvlink will have non-functioning CUDA installations. this means everything looks like it works (including GPU partitioning) right up until a user actually tries to deploy a workload.


parts/linux/cloud-init/artifacts/mig-partition.sh
```diff
@@ -21,7 +21,7 @@ case ${MIG_PROFILE} in
         ;;  
     *)
         echo "not a valid GPU instance profile"
-        exit ${ERR_MIG_PARTITION_FAILURE}
```
alexeldeib, line 24: this variable is invalid in this context, it's sourced from CSE and this is a systemd unit. switch to exit 1.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -489,7 +489,15 @@ configGPUDrivers() {
     retrycmd_if_failure_no_stats 120 5 25 update-initramfs -u || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
     wait_for_apt_locks
     {{/* if the unattened upgrade is turned on, and it may takes 10 min to finish the installation, and we use the 1 second just to try to get the lock more aggressively */}}
-    retrycmd_if_failure 600 1 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-container-runtime="${NVIDIA_CONTAINER_RUNTIME_VERSION}+${NVIDIA_DOCKER_SUFFIX}" || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+    retrycmd_if_failure 600 1 3600 apt-get -o Dpkg::Options::="--force-confold" install -y "cuda-drivers-fabricmanager-$GPU_FABRICMGR_VERSION" nvidia-container-runtime="${NVIDIA_CONTAINER_RUNTIME_VERSION}+${NVIDIA_DOCKER_SUFFIX}" || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+    if [[ "${MIG_NODE}" == "true" ]]; then
+        retrycmd_if_failure 600 1 3600 apt-get -o Dpkg::Options::="--force-confold" install -y "cuda-drivers-fabricmanager-$GPU_FABRICMGR_VERSION" || exit $ERR_FABRIC_MANAGER_INSTALL_TIMEOUT
+    else
+        if dpkg -s "cuda-drivers-fabricmanager-$GPU_FABRICMGR_VERSION"; then
```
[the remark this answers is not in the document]
ganeshkumar5699: Line 492 and 494 are almost the same except for the `nvidia-container-runtime`. So are we attempting to re-install `fabric-manager` when we are on a MIG node? That’s probably not the case though.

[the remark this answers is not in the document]
ganeshkumar5699: Instead of removing fabric manager for non-MIG nodes, is it ok to only install it for MIG nodes? In that case, perhaps the installation time for non-MIG nodes would go down a bit too?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -109,11 +109,24 @@ if [[ "${GPU_NODE}" = true ]]; then
...
+            wait_for_file 3600 1 /etc/systemd/system/nvidia-device-plugin.service.d/10-mig_strategy.conf || exit $ERR_FILE_WATCH_TIMEOUT
+        fi
         systemctlEnableAndStart nvidia-device-plugin || exit $ERR_GPU_DEVICE_PLUGIN_START_FAIL
     else
         systemctlDisableAndStop nvidia-device-plugin
     fi
 fi
+# If it is a MIG Node, enable mig-partition systemd service to create MIG instances
+if [[ "${MIG_NODE}" == "true" ]]; then
+    REBOOTREQUIRED=true
```
ganeshkumar5699, line 122: how did you determine that reboot was required to create mig instances? was it from the Nvidia MIG guide?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -109,11 +109,24 @@ if [[ "${GPU_NODE}" = true ]]; then
...
+    ensureMigPartition
+else
+    if dpkg -s "cuda-drivers-fabricmanager-$GPU_FABRICMGR_VERSION"; then
```
ganeshkumar5699, line 125: Are there restrictions on which SKUs support MIG nodes (I assume it's only the GPU skus)? If so, which section of the code would those constraints be present in?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -132,7 +132,17 @@ write_files:
...
-    {{GetVariableProperty "cloudInitData" "migPartitionScript"}}  
+    {{GetVariableProperty "cloudInitData" "migPartitionScript"}}
+
+- path: /etc/systemd/system/nvidia-device-plugin.service.d/10-mig_strategy.conf
+  permissions: "0644"
+  owner: root
+  content: |
+    [Service]
+    Environment="MIG_STRATEGY=--mig-strategy single"
+    ExecStart=
```
ganeshkumar5699, line 143: Is this empty ExecStart needed?

parts/linux/cloud-init/artifacts/cse_config.sh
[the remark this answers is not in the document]
alexeldeib: oops-nice catch. L492 shouldn't have the fabric manager, only MIG nodes will have it.

since we don't have a mig specific vhd (just generic gpu), and there's minimal effect, I figured just tidying up the bits at runtime was ok. I think there needs to be a more general cleanup after this.

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: testing and discussion with nvidia. also this logic was already here :) just moved earlier.

actually would be nice to talk with hyperv/hpc folks if we can remove this requirement, it's not strictly necessary on other platforms.

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: yes, RP handles that, but the GPU optimized vhd will have these components for non-MIG which isn't so great. I'd rather remove than add at runtime though, it'll be a bit dirty either way sadly :/

parts/linux/cloud-init/nodecustomdata.yml
alexeldeib: yes, you can't override execstart directly otherwise as I just learned :)

parts/linux/cloud-init/artifacts/cse_config.sh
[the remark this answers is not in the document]
ganeshkumar5699: got it, thanks!

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -109,11 +109,32 @@ if [[ "${GPU_NODE}" = true ]]; then
...
     fi
 fi
+# If it is a MIG Node, enable mig-partition systemd service to create MIG instances
+if [[ "${MIG_NODE}" == "true" ]]; then
+    REBOOTREQUIRED=true
+    STATUS=`systemctl is-active nvidia-fabricmanager`
+    if [ ${STATUS} = 'active' ]; then
+        echo "Fabric Manager service is running, no need to install."
+    else
+        retrycmd_if_failure 30 5 3600 wget "https://developer.download.nvidia.com/compute/cuda/redist/fabricmanager/linux-x86_64/fabricmanager-linux-x86_64-${GPU_DV}.tar.gz"
```
alexeldeib, line 127: remove before merge -- not allowed for egress. should update docs.

alexeldeib: we'll bake it into vanilla vhd, just not install it until runtime

ganeshkumar5699 (approved):
lgtm. nice!

PR #1299, opened 2021-11-02
demo code for a new provisioning path

alexeldeib:


parts/linux/cloud-init/config.ign
```diff
@@ -0,0 +1,8 @@
+{
+    "TODO": "POPULATE_ME",
```
alexeldeib, line 2: if we template as cloud-config and then convert to ignition, that works fine too. I didn't know how to easily generate ignition from cloud-config yaml without shelling to a binary

alexeldeib: his is really where the magic sauce should be

PR #1302, opened 2021-11-04
Trigger multiple ev2 artifact builds

anujmaheshwari1:


.pipelines/.vsts-vhd-automation.yaml
```diff
@@ -2,17 +2,19 @@ pool:
   vmImage: ubuntu-latest
 
 steps:
+- bash: |
+        echo $ADO_PAT | az devops login --organization=https://dev.azure.com/msazure
```
anujmaheshwari1, line 6: This does not actually log the PAT, pipeline run for reference: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=48614630&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=76467409-274e-589c-5c88-4286504c504b

vhdbuilder/scripts/automate_ev2pipeline_trigger.sh
```diff
@@ -5,17 +5,29 @@ source vhdbuilder/scripts/automate_helpers.sh
...
+            exit 1
+        else
+            echo "Build ID successfully updated"
+        fi
 
-configure_az_devops $system_access_token
-set -x
+        # Run the pipeline and fetch the run ID to poll for success later
+        run_id=$(az pipelines run --id $pipeline_id | jq -r '.id')
+        while ! az pipelines runs show --id $run_id | grep -q '"result": "succeeded"'; do
```
anujmaheshwari1, line 25: The pipeline "result" is null until it succeeds. Should a "failed" state be considered? This logic presumes that the ev2 artifact pipeline will eventually succeed.

alexeldeib: I think as long as this has a reasonable timeout it's okay to just search for the success case

vhdbuilder/scripts/automate_ev2pipeline_trigger.sh
```diff
@@ -5,17 +5,29 @@ source vhdbuilder/scripts/automate_helpers.sh
...
-system_access_token=$2
+build_ids=$1
 
 trigger_pipeline() {
-    az pipelines variable update --name VHD_PIPELINE_RUN_ID --pipeline-id 165939 --value $build_id  # Update the VHD_PIPELINE_RUN_ID with the build ID
-    az pipelines run --id 165939  # 165939 is the pipeline ID for ev2 artifacts pipeline
-}
+    pipeline_id="165939"   # 165939 is the pipeline ID for ev2 artifacts pipeline
+    for build_id in $build_ids; do
+        echo "Build ID for the release is $build_id"
```
alexeldeib, line 13: consider parallelizing these, logging to file, and checking the pids for success like we did for vhd reduction (checkout the script there for how to do it)

alexeldeib (approved):
minor comments, lgtm

vhdbuilder/scripts/automate_ev2pipeline_trigger.sh
ganeshkumar5699: So around a 2 hr timeout for the polling to detect success or exit the loop would be good? I've seen some builds take up to 1.5 hrs.

ganeshkumar5699 (approved):

vhdbuilder/scripts/automate_ev2pipeline_trigger.sh
anujmaheshwari1: This is ev2 artifact build, it doesn't take more than 3/4 minutes.

vhdbuilder/scripts/automate_ev2pipeline_trigger.sh
anujmaheshwari1: I didn't parallelize this because we have to update the VHD_PIPELINE_RUN_ID for each build_id, what about a race condition where both processes try to update the variable at once? Since the pipeline takes only 3/4 minutes to run, I left it sequential. Thoughts?

vhdbuilder/scripts/automate_ev2pipeline_trigger.sh
ganeshkumar5699: Ok, got it - thought it was the build for some reason :). The ev2 build artifact runs have usually succeeded so a timeout a bit more generous than the usual time to complete would be good.

PR #1305, opened 2021-11-08
bake in default kubernetes CNI plugins always

mattstam:
I originally made the assumption that we wouldn't need to bake in https://github.com/containernetworking/plugins, but it seems we will still want do so. 

FYI this function is:

```
installNetworkPlugin() {
    if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
        installAzureCNI
    fi
    installCNI
    rm -rf $CNI_DOWNLOADS_DIR &
}
```

```
installCNI() {
    CNI_TGZ_TMP=${CNI_PLUGINS_URL##*/} # Use bash builtin ## to remove all chars ("*") up to the final "/"
    if [[ ! -f "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" ]]; then
        downloadCNI
    fi
    mkdir -p $CNI_BIN_DIR
    tar -xzf "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" -C $CNI_BIN_DIR
    chmod -R 755 $CNI_BIN_DIR
}
```


paulgmiller:
To be clear the reason we think we do is that containerd has a dependency on at least loopback.

paulgmiller (approved):
Next up actually install recent ones!

PR #1312, opened 2021-11-10
Remove unused cahced omsagent windows image

AbelHu:
Remove unused cahced omsagent windows image


AbelHu:
@andyzhangx @ZeroMagic @nilo19 @feiskyer could you help to check the cached Windows images for azure csi/disk/file and azure-cloud-node-manager.
I think that we need to keep all versions support in AKS even for those old k8s versions. For example, we need to cache azuredisk-csi:v1.8.0.1 for >=1.21.0 and azuredisk-csi:v1.2.0.5 for < 1.21.0.

ganga1980 (approved):

PR #1323, opened 2021-11-12
another attempt at rebooting vm during build

alexeldeib:


vhdbuilder/packer/vhd-image-builder.json
```diff
@@ -14,11 +14,13 @@
     "hyperv_generation": "{{env `HYPERV_GENERATION`}}",
     "container_runtime": "{{env `CONTAINER_RUNTIME`}}",
     "teleportd_plugin_download_url": "{{env `TELEPORTD_PLUGIN_DOWNLOAD_URL`}}",
-    "enable_fips": "{{env `ENABLE_FIPS`}}"
+    "enable_fips": "{{env `ENABLE_FIPS`}}",
+    "allowed_ip": "{{env `PUBLIC_IP`}}"
```
alexeldeib, line 18: @anujmaheshwari1 allowed IP plumbed here

vhdbuilder/packer/init-variables.sh
```diff
@@ -213,7 +214,8 @@ cat <<EOF > vhdbuilder/packer/settings.json
   "arm64_sig_resource_group_name": "${ARM64_SIG_RESOURCE_GROUP_NAME}",
   "arm64_sig_gallery_name": "${ARM64_SIG_GALLERY_NAME}",
   "arm64_sig_image_name": "${ARM64_SIG_IMAGE_NAME}",
-  "arm64_sig_image_version": "${ARM64_SIG_IMAGE_VERSION}"
+  "arm64_sig_image_version": "${ARM64_SIG_IMAGE_VERSION}",
+  "allowed_ip": "${PUBLIC_IP}"
```
alexeldeib, line 218: @anujmaheshwari1 allowed IP plumbed here, grabbed from ifconfig.me which is a public service. 

it would be better to know the expected IP in advance, but I think the VM has no IP and this is some Azure NAT'd pub IP so may not be possible. In the future if we adjusted our packer build to run on an agent in one subnet plus build a vm in the same subnet, they could both run without allowed_ip

alexeldeib:
we need this without all the cruft. need to restrict nsg to msft datacenter ips.

PR #1332, opened 2021-11-18
feat: add motd support

alexeldeib:


anujmaheshwari1 (approved):

ganeshkumar5699:
Perhaps a basic question:
Why are changes for message of the day in agentbaker/through CSE scripts? The string to be displayed is provided through RP, so can it not show the CLI changes without going through agentbaker? Is it because we want to create `etc/motd`?

ganeshkumar5699:
> Perhaps a basic question: Why are changes for message of the day in agentbaker/through CSE scripts? The string to be displayed is provided through RP, so can it not show the CLI changes without going through agentbaker? Is it because we want to create `etc/motd`?

It's displayed in each node but it's at the agentpool level (based on my understanding of the RP PR), isn't it?

alexeldeib:
> Is it because we want to create etc/motd

yes, this is the entire point of this feature. for e.g. FedRAMP compliance, users need to have a customize message on the nodes. 

I just felt it made sense at the nodepool level naturally

> so can it not show the CLI changes without going through agentbaker

I don't follow this. CLI calls RP, RP calls AB with the new property, nodes still need to write the value to disk. We always need to plumb new features like this through cse/cloud init to "do the thing" when we are adding new node features.

ganeshkumar5699 (approved):
Got it, thanks for explaining. So since we are making node level writes, it needs to be here. I'd mixed up thinking about agentpool level features vs node-level features that need to modify the node.

Lgtm.

ganeshkumar5699:
Had thought of an alternate (but poorly designed) approach where we obtain the message from the DB, each time, with the request containing the motd string. It may need fewer CSE changes but that would be time consuming and quite unnecessary when we can persist that message in the node through `etc/motd` :).

alexeldeib:
> approach where we obtain the message from the DB, each time, with the request containing the motd string

how would you even do that though? keep in mind, RP is storing the user requested model in a database the node can't access. the node needs _some_ way to access the data. 

We could, in the future, do something like have RP write the desired message to e.g. a storage account and download it at runtime, but it doesn't save much work. A better example might be detecting N series (gpu) vms. That can be done either by: checking sku api, checking IMDS, or using nvidia-smi / on-host device detection to figure out if there are GPUs available, and then removing that helper and associated GPU config from AB.

PR #1326, opened 2021-11-16
Freezing 1604

MirzaSikander:
- Freeze 1604 to the latest image
- Not build 1604 anymore

PR #1337, opened 2021-11-19
Cut official branch and validate build commits

anujmaheshwari1:


vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -47,10 +49,47 @@ update_image_version() {
...
+    git push -u origin $official_branch_name
+
+    git tag $official_tag
+    git push origin tag $official_tag
+    git checkout master
+
+}
+
+set_git_config
+find_current_image_version "pkg/agent/datamodel/osimageconfig.go"
```
alexeldeib, line 93: should we consider using sig_config here rather than osimageconfig? I see the update func hits all files, but since AB is only source of truth for sigconfig, maybe better to rely on that?

alexeldeib (approved):
lgtm pending testing, one minor comment

vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -47,10 +49,47 @@ update_image_version() {
...
+    create_pull_request $new_image_version $github_access_token $branch_name $pr_title
+    set -x
+}
+
+# This function cuts the official branch based off the commit ID that the builds were triggered from and tags it
+cut_official_branch() {
+    official_branch_name="official/v${new_image_version//.}"
+    official_tag="v0.${new_image_version//.}.0"
+    final_commit_hash=""
+    for build_id in $build_ids; do
```
ganeshkumar5699, line 66: would it be cleaner if we move this section (from l66:76) to a separate function like `check_builds_same_commit`? That would make that validation logic separate from the script to cut the branch.

vhdbuilder/scripts/automate_version_bump.sh
ganeshkumar5699: would the 1604 freeze also affect the automated PR bump?  This PR https://github.com/Azure/AgentBaker/pull/1326/files sets  `sig_config.go` to have a particular image version for 1604 and `update_image_version` replaces the versions in the same file. It may not affect this PR but I just wanted to flag it to be sure.

anujmaheshwari1: @alexeldeib is it ok if I iterate on this is a separate PR? it would require me to change the logic of extracting the current image version which depends on the file composition.

anujmaheshwari1: @ganeshkumar5699 it will be a one time thing, as soon as the version for 1604 and the latest version for other SKUs are out of sync, shouldnt matter.

vhdbuilder/scripts/automate_version_bump.sh
anujmaheshwari1: I did not move it to a separate function because the check isn't too long/complicated. Can do that in a later PR though, will add a TODO if that is ok. @ganeshkumar5699

ganeshkumar5699: yes, that sounds good :)

vhdbuilder/scripts/automate_version_bump.sh
ganeshkumar5699: Sounds good. I see how manually fixing the image version in the next VHD release would make the 1604 image correct for that and subsequent releases (as the `current_image_version` would change afterwards).

anujmaheshwari1:
> lgtm pending testing, one minor comment

Successful branch cutting test run : https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=49235539&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=12f1170f-54f2-53f3-20dd-22fc7dff55f9

Dummy official branch https://github.com/Azure/AgentBaker/tree/official/v30304050

ganeshkumar5699 (approved):

PR #1338, opened 2021-11-19
Don't action on /var/runreboot-required. 

paulgmiller:
Even though mig also doesn't its not ideal to reboot right after CSE runs as we might immediately drop pods that were just moved over and there is no drain.

This would only be hit a small subset of time. There's basicalyl an hour windows 6AM utc where it runs so even if it races your cse you would only see this 1/(24*7) of runs...

There for best to remove and let kured or someother outside service worry about reboots.


paulgmiller:
No rush on this happy for more debate.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -213,6 +207,8 @@ else
     retrycmd_if_failure ${API_SERVER_CONN_RETRIES} 1 10 nc -vz ${API_SERVER_NAME} 443 || time nc -vz ${API_SERVER_NAME} 443 || VALIDATION_ERR=$ERR_K8S_API_SERVER_CONN_FAIL
 fi
 
+#Ace: Basically the hypervisor blocks gpu reset which is required after enabling mig mode for the gpus to be usable
```
alexeldeib, line 210: nit
```suggestion
# Ace: Basically the hypervisor blocks gpu reset which is required after enabling mig mode for the gpus to be usable
```

PR #1344, opened 2021-11-29
Improve e2e script

johananl:
This PR improves the e2e script.

Summary of changes:

- Avoid altering the user's global kubeconfig file.
- Explicitly use JSON output in all `az` commands to mitigate environment variations.
- Allow overriding script params using env vars.
- Wrap fragile API calls in retry logic.
- Fix masked exit codes in calls to the Azure API.
- Misc cleanups.

### DIFF
--- a/e2e/e2e-script.sh
+++ b/e2e/e2e-script.sh
@@ -9,13 +9,12 @@ echo "Starting e2e tests"
 : "${CLUSTER_NAME:=agentbaker-e2e-test-cluster}"
 
 # Create a resource group for the cluster
-if [ $(az group exists -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID -ojson) == "false" ]; then
-    echo "Creating resource group"
-    az group create -l $LOCATION -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID -ojson
-fi
+echo "Creating resource group"
+az group create -l $LOCATION -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID -ojson
 
 # Create the AKS cluster and get the kubeconfig
-if [ -z $(az aks list -g $RESOURCE_GROUP_NAME -ojson | jq '.[].name') ]; then
+out=$(az aks list -g $RESOURCE_GROUP_NAME -ojson | jq '.[].name')
+if [ -z "$out" ]; then
     echo "Cluster doesnt exist, creating"
     az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys -ojson
 fi
--- a/e2e/e2e-script.sh
+++ b/e2e/e2e-script.sh
@@ -48,7 +48,9 @@ az vmss run-command invoke \
 
 declare -a files=("apiserver.crt" "ca.crt" "client.key")
 for file in "${files[@]}"; do
-    content=$(az vmss run-command invoke \
+    for i in $(seq 1 10); do
+        set +e
+        content=$(az vmss run-command invoke \
                 -n $VMSS_NAME \
                 -g $MC_RESOURCE_GROUP_NAME \
                 --command-id RunShellScript \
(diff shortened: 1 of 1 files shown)


e2e/e2e-script.sh
```diff
@@ -3,33 +3,32 @@ set -euxo pipefail
...
-# Clear the kube/config file for any conflicts
-truncate -s 0 ~/.kube/config
+: "${SUBSCRIPTION_ID:=8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8}" #Azure Container Service - Test Subscription
+: "${RESOURCE_GROUP_NAME:=agentbaker-e2e-tests}"
+: "${LOCATION:=eastus}"
+: "${CLUSTER_NAME:=agentbaker-e2e-test-cluster}"
 
 # Create a resource group for the cluster
-if [ $(az group exists -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID) == "false" ]; then
+out=$(az group exists -n $RESOURCE_GROUP_NAME --subscription $SUBSCRIPTION_ID -ojson)
```
alexeldeib, line 12: I think you can unconditionally create the group since it should return 200 either way, but this works for me

e2e/e2e-script.sh
```diff
@@ -49,17 +50,30 @@ az vmss run-command invoke \
...
+        retval=$?
+        set -e
+        if [ "$retval" -ne 0 ]; then
+            echo "retrying attempt $i"
+            sleep 10s
+            continue
+        fi
+        break;
+    done
+    [ "$retval" -eq 0 ]
```
alexeldeib, line 75: what's the intent here? validate the final retval and exit if not zero? not super accustomed to this syntax without `if; then; fi`

e2e/e2e-script.sh
```diff
@@ -111,12 +135,14 @@ az vmss create -n ${VMSS_NAME} \
     --instance-count 1 \
     --assign-identity $msiResourceID \
     --image "microsoft-aks:aks:aks-ubuntu-1804-gen2-2021-q2:2021.05.19" \
-    --upgrade-policy-mode Automatic
+    --upgrade-policy-mode Automatic \
+    -ojson
```
alexeldeib, line 139: isn't -o json the default? just being explicit? not seeing the output used anywhere anyway?

e2e/e2e-script.sh
```diff
@@ -130,16 +156,20 @@ az vmss extension set --resource-group $MC_RESOURCE_GROUP_NAME \
     --vmss-name ${VMSS_NAME} \
     --publisher Microsoft.Azure.Extensions \
     --protected-settings settings.json \
-    --version 2.0
+    --version 2.0 \
+    -ojson
```
alexeldeib, line 160: same as above

alexeldeib (approved):
minor comments, thanks for fixing things up :) assuming we'll iterate to bring in the flatcar bits?

e2e/e2e-script.sh
johananl: Ah, I think you're right - creating the group does seem idempotent. I'll remove the conditional :+1:

e2e/e2e-script.sh
johananl: It's basically a shorthand for the following (since we have `set -e`):

```bash
if [ "$retval" -eq 0 ]; then
    exit 1
fi
```

I can refactor to echo something and exit explicitly if that makes more sense.

e2e/e2e-script.sh
johananl: >isn't -o json the default? just being explicit?

It is, however we probably want the script to want on machines where a different output format is configured. On my laptop, for example, I had `table` configured, which made the script break.

>not seeing the output used anywhere anyway?

In this case the output isn't used, you're right. I set `-ojson` here, too, because the command's output is *printed* and I wanted to be consistent with other cases where we print stuff in JSON format.

e2e/e2e-script.sh
johananl: See https://github.com/Azure/AgentBaker/pull/1344/files#r759108725.

e2e/e2e-script.sh
alexeldeib: fine as is, just wanted to double check 👍 learned something new today

PR #1364, opened 2021-12-08
fix: Catch failure when failing in download Windows CSE scripts package

AbelHu:
1. Move downloading AKS Windows CSE Scripts Package to try block so we can return the correct exitcode
2. Set all variables definition before `# Extract cse helper script from ZIP`


parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -188,6 +188,11 @@ $global:TLSBootstrapToken = "{{GetTLSBootstrapTokenForKubeConfig}}"
 # Base64 representation of ZIP archive
 $zippedFiles = "{{ GetKubernetesWindowsAgentFunctions }}"
 
+$useContainerD = ($global:ContainerRuntime -eq "containerd")
```
andyliuliming, line 191: so that's the reason the retry can not work in this case?

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -188,6 +188,11 @@ $global:TLSBootstrapToken = "{{GetTLSBootstrapTokenForKubeConfig}}"
 # Base64 representation of ZIP archive
 $zippedFiles = "{{ GetKubernetesWindowsAgentFunctions }}"
 
+$useContainerD = ($global:ContainerRuntime -eq "containerd")
+$global:KubeClusterConfigPath = "c:\k\kubeclusterconfig.json"
```
andyliuliming, line 192: just curious can we add tests for the ps1 file?

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -196,36 +201,30 @@ Expand-Archive scripts.zip -DestinationPath "C:\\AzureData\\"
...
-$windowsSecureTlsEnabled = [System.Convert]::ToBoolean("{{GetVariable "windowsSecureTlsEnabled" }}");
-
 try
 {
+    Write-Log ".\CustomDataSetupScript.ps1 -MasterIP $MasterIP -KubeDnsServiceIp $KubeDnsServiceIp -MasterFQDNPrefix $MasterFQDNPrefix -Location $Location -AADClientId $AADClientId -NetworkAPIVersion $NetworkAPIVersion -TargetEnvironment $TargetEnvironment"
+
+    # Download CSE function scripts
+    Write-Log "Getting CSE scripts"
+    $tempfile = 'c:\csescripts.zip'
+    DownloadFileOverHttp -Url $global:CSEScriptsPackageUrl -DestinationPath $tempfile
```
andyliuliming, line 211: I think we should add logs for the retry in the
 try {
            $args = @{Uri=$Url; Method="Get"; OutFile=$DestinationPath}
            Retry-Command -Command "Invoke-RestMethod" -Args $args -Retries 5 -RetryDelaySeconds 10
        } catch {
            Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_DOWNLOAD_FILE_WITH_RETRY -ErrorMessage "Failed in downloading $Url. Error: $_"
        }

so we know whether the retry works as expected.

parts/windows/kuberneteswindowssetup.ps1
AbelHu: We should be able to do it but it needs much work. At first we need to generate values and compile this script since it actually is a go template.

parts/windows/kuberneteswindowssetup.ps1
AbelHu: Could you explain more? For "#2" in the description, I just refined the code to make it more readable.

parts/windows/kuberneteswindowssetup.ps1
AbelHu: Good suggestion. I can add logs for the retry in `function Retry-Command `

parts/windows/kuberneteswindowssetup.ps1
AbelHu: I tested these change in AKS RP side

andyliuliming (approved):

ShiqianTao (approved):

PR #1369, opened 2021-12-09
Add registry configuration breakout to containerd/config.toml

phealy:
Adds a config entry to reference /etc/containerd/certs.d, which can contain per registry configuration files. This will allow customers to drop off config files with a DaemonSet that will modify how containerd acts when it connects to certain registries.

Enabled scenarios:
- Adding a trusted root CA for a given registry, enabling private registry CAs
- Adding a client certificate for mutual TLS against a registry
- Redirecting a registry URL to a pull-through registry proxy seamlessly (so anything going to docker.io actually goes to mydockerproxy.mydomain.com)

Fixes [AKS/1940](https://github.com/Azure/AKS/issues/1940) and helps with some of the people in [AKS/2259](https://github.com/azure/aks/issues/2259) for Linux nodepools.

See:
- [Containerd registry configuration](https://github.com/containerd/containerd/blob/main/docs/cri/config.md#:~:text=%23%20The%20cri%20plugin%20will%20look%20for%20and%20use%20config_path/host%2Dnamespace/hosts.toml%0A%20%20%20%20%23%20%20%20configs%20if%20present%20OR%20load%20certificate%20files%20as%20laid%20out%20in%20the%20Docker/Moby%0A%20%20%20%20%23%20%20%20specific%20layout%20https%3A//docs.docker.com/engine/security/certificates/)
- [Docker TLS configuration (which is respected by containerd if placed in /etc/containerd/certs.d)](https://docs.docker.com/engine/security/certificates/)


phealy:
I tested this in a standalone and it seems to work great - the config is now limited to 1.22.x only where we have containerd 1.5, which is where the option support came in.

Only issue I see - none of our actual test cases in AgentBaker seem to build 1.22 clusters right now, so none of the test data includes the config statement. I'm not quite sure what the best way to add that test case in would be.

phealy:
#add-reviewer @alexeldeib

alexeldeib:
k8s version is in the describle table tests here: https://github.com/Azure/AgentBaker/blob/7a18f182ed101ebdef746f78d0463c3dfe1db343/pkg/agent/baker_test.go#L23-L30

a case like this with 1.22+containerd would be good to cover this scenario I think: https://github.com/Azure/AgentBaker/blob/7a18f182ed101ebdef746f78d0463c3dfe1db343/pkg/agent/baker_test.go#L530-L535

otherwise PR lgtm, appreciate it :)

phealy:
@alexeldeib I don't have write access, so would please you merge or help me fix that?

PR #1372, opened 2021-12-11
New Ubuntu VHD build skeleton for ARM arch

chengliangli0918:
1. Add a new vhd build skeletion for new arch
2. New vhd build VM for new arch is created and then install-dependencies.sh starts
3. new installation logic will be added in following PRs to keep each PR small.

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -65,6 +65,10 @@ parameters:
   displayName: Build 1804 FIPS Gen2 GPU+containerd
   type: boolean
   default: true
+- name: build1804arm64gen2containerd
+  displayName: Build 1804 ARM64 Gen2 containerd
+  type: boolean
+  default: false
 - name: dryrun
   displayName: Dry run
   type: boolean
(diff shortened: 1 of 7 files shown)


vhdbuilder/packer/init-variables.sh
```diff
@@ -182,6 +182,18 @@ if [ ! -z "${WINDOWS_SKU}" ]; then
 	esac
 fi
 
+# no source image on marketplace, use MSFT/AME SIG as source image
+if [[ ${ARCHITECTURE,,} == "arm64" ]]; then
+  ARM64_SIG_SUBSCRIPTION_ID="a84c0852-ea2d-4359-88e4-11a80a4fb6b9"
```
alexeldeib, line 187: how will we access this, is the image public?

chengliangli0918: I asked the ARM64 team to grant read access to aksvhdbuilderSP for the AME AMR64 SIG,  aksvhdbuildSP and aks_agentbaker_vhd_test for the MSFT ARM64 SIG. The packer build pipeline can create the ARM64 VMs, start the script install-dependencies.sh. I will make changes in install-dependencies.sh and file related in following PRs to keep each PR not too big.

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -365,3 +369,24 @@ stages:
...
+    jobs:
+    - job: build
+      timeoutInMinutes: 180
+      steps:
+        - bash: |
+            echo '##vso[task.setvariable variable=DRY_RUN]${{parameters.dryrun}}'
+            echo '##vso[task.setvariable variable=OS_SKU]Ubuntu'
+            echo '##vso[task.setvariable variable=OS_VERSION]18.04'
+            echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
+            echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D8darm_V3'
```
ganeshkumar5699, line 384: Does using `Standard_D8darm_V3` need any additional testing/changes elsewhere? Asking since other builds use `Standard_DS2_v`2 and I wasn’t able to see `Standard_D8darm_v3` vm size used in Agentbaker.

ganeshkumar5699 (approved):

.pipelines/.vsts-vhd-builder-release.yaml
chengliangli0918: no, the packer pipeline will use the vm size that we ask it to use and it creates a VM and do the installations. For now the ARM SKU is only visible on the subscriptions (1 MSFT subscription and 1 AME subscription) that the VHD build pipeline is using.

PR #1385, opened 2021-12-28
Add arm distro sig info in order that AKS RP can consume it

chengliangli0918:
Add arm distro sig info in order that AKS RP can consume it

### DIFF
--- a/pkg/agent/datamodel/sig_config.go
+++ b/pkg/agent/datamodel/sig_config.go
@@ -66,7 +66,9 @@ var AvailableUbuntu1804Distros []Distro = []Distro{
 	AKSUbuntuFipsContainerd1804,
 	AKSUbuntuFipsContainerd1804Gen2,
 	AKSUbuntuFipsGPUContainerd1804,
-	AKSUbuntuFipsGPUContainerd1804Gen2}
+	AKSUbuntuFipsGPUContainerd1804Gen2,
+	AKSUbuntuArm64Containerd1804Gen2,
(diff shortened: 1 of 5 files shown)


pkg/agent/datamodel/sig_config.go
```diff
@@ -202,6 +206,8 @@ const (
 const (
 	LinuxSIGImageVersion   string = "2021.12.07"
 	WindowsSIGImageVersion string = "17763.2366.211215"
+	// will not do weekly vhd release as amd64 when ARM64 Compute/AKS is still under development
+	Arm64LinuxSIGImageVersion string = "2021.12.28"
```
anujmaheshwari1, line 210: can this be merged with the LinuxSIGImageVersion once CCOA is done and there is an RP release next week? otherwise, this would require changes in the VHD Automation pipeline as well to incorporate a separate version for the Arm64 VHD version.

anujmaheshwari1 (approved):

pkg/agent/datamodel/sig_config.go
chengliangli0918: @anujmaheshwari1 thank for mentioning it. Before arm64 compute is in public preview or GA, I will not enable aks arm64 vhd build leg in AKS weekly VHD build/release pipeline to cause trouble to Linux VHD weekly release manager. In next 2-3 months, I will take care of arm vhd build/release(will do it weekly or bi-weekly)

PR #1393, opened 2022-01-04
upgrade csi driver version

andyzhangx:
`azurefile-csi:v1.8.0`, `azuredisk-csi:v1.9.0` is not used in AKS rp yet, so it's safe to upgrade in AgentBaker directly.


andyzhangx:
cc @AbelHu

AbelHu (approved):

PR #1395, opened 2022-01-04

HDYA:

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -71,7 +71,8 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/coredns:*",
       "amd64OnlyVersions": [
-        "1.6.6"
+        "1.6.6",
+        "1.8.6"
       ],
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/components.json
```diff
@@ -238,8 +244,11 @@
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/azure-cloud-node-manager:*",
       "amd64OnlyVersions": [
         "v0.6.0",
+        "v0.7.9",
         "v1.0.2",
-        "v1.1.1"
+        "v1.0.6",
+        "v1.1.1",
+        "v1.1.2"
```
andyliuliming, line 251: why a new image tag is added. (v1.1.2)

andyliuliming (approved):

vhdbuilder/packer/components.json
HDYA: latest version 1.1.2 is in section multiArchVersions, pipeline added it to multiArchVersions as we will not touch multiArchVersions for now

PR #1396, opened 2022-01-05
Remove versions from amd64OnlyVersions because they are in multiArchVersions already

chengliangli0918:
Remove some versions from amd64OnlyVersions because they are in multiArchVersions already

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -71,8 +71,7 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/coredns:*",
       "amd64OnlyVersions": [
-        "1.6.6",
-        "1.8.6"
+        "1.6.6"
       ],
       "multiArchVersions": [
         "1.8.6"
@@ -118,9 +117,7 @@
     },
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/apiserver-network-proxy/agent:*",
-      "amd64OnlyVersions": [
-        "v0.0.24-hotfix.20211027"
-      ],
+      "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v0.0.24-hotfix.20211027",
         "v0.0.27"
@@ -136,8 +133,7 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes-csi/secrets-store/driver:*",
       "amd64OnlyVersions": [
-        "v0.0.21.4",
-        "v1.0.0"
+        "v0.0.21.4"
       ],
       "multiArchVersions": [
         "v1.0.0"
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/components.json
```diff
@@ -224,11 +218,9 @@
     },
     {
       "downloadURL": "mcr.microsoft.com/oss/tigera/operator:*",
-      "amd64OnlyVersions": [
-        "v1.20.0"
-      ],
+      "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.22.3",
```
andyliuliming, line 231: the 1.22.3 is not in use?

vhdbuilder/packer/components.json
```diff
@@ -306,7 +292,7 @@
         "v1.8.0.2"
       ],
       "multiArchVersions": [
-        "v1.10.0"
```
andyliuliming, line 309: "v1.10.0" is not in use anymore?

vhdbuilder/packer/components.json
```diff
@@ -316,7 +302,7 @@
         "v1.7.0"
       ],
       "multiArchVersions": [
-        "v1.9.0"
```
andyliuliming, line 319: also here.

vhdbuilder/packer/components.json
HDYA, line 309: could you verify this change? seems https://github.com/Azure/AgentBaker/pull/1395/files did not touch it

vhdbuilder/packer/components.json
HDYA, line 319: could you verify this change? seems https://github.com/Azure/AgentBaker/pull/1395/files did not touch it

HDYA (commented):
original PR did not touch multiArchVersions, please help verify

vhdbuilder/packer/components.json
chengliangli0918: 1.22.3 is the first multi-arch I plan to use, but we backfilled the 1.20.0 currently in use to 'multi-arch'.

vhdbuilder/packer/components.json
chengliangli0918: yes, you're right. azuredisk-csi:v1.9.0 is the minimum version with 'multi-arch'. I will use 1.9.0 instead of 1.10.0.

vhdbuilder/packer/components.json
chengliangli0918: yes, you're right. azurefile-csi:v1.8.0 is the first version with 'multi-arch'. I will use 1.8.0 instead of 1.9.0.

vhdbuilder/packer/components.json
chengliangli0918: azuredisk-csi:v1.9.0 is the minimum version with 'multi-arch'. I will use 1.9.0 instead of 1.10.0.  Current RP doesn't use 1.10.0 at all.

vhdbuilder/packer/components.json
chengliangli0918: azurefile-csi:v1.8.0 is the first version with 'multi-arch'. I will use 1.8.0 instead of 1.9.0. Current RP doesn't use 1.9.0 at all.

andyliuliming (approved):

HDYA (approved):

PR #1400, opened 2022-01-07
update image version in v20220101 branch

chengliangli0918:
update image version in v20220101 branch

### DIFF
--- a/pkg/agent/datamodel/osimageconfig.go
+++ b/pkg/agent/datamodel/osimageconfig.go
@@ -20,9 +20,9 @@ var (
 
 	Ubuntu1804Gen2OSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks",
-		ImageSku:       "aks-ubuntu-1804-gen2-2021-q3",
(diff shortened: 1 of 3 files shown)


pkg/agent/datamodel/osimageconfig.go
```diff
@@ -20,9 +20,9 @@ var (
 
 	Ubuntu1804Gen2OSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks",
-		ImageSku:       "aks-ubuntu-1804-gen2-2021-q3",
+		ImageSku:       "aks-ubuntu-1804-gen2-2022-q1",
```
anujmaheshwari1, line 23: oh, this was a good catch @chengliangli0918. I think @ganeshkumar5699 made the version PR before creating a new plan for the 2 PIR images. so there would have been a mismatch, but good catch.

andyliuliming (approved):

pkg/agent/datamodel/osimageconfig.go
ganeshkumar5699: Thanks for this PR @chengliangli0918! Great catch!

PR #1404, opened 2022-01-07
move mcr retag to runtime

alexeldeib:


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -184,6 +184,7 @@ ensureMonitorService
 # https://github.com/kubernetes/kubernetes/issues/51017
 # can remove when fixed
 cleanupRetaggedImages
```
alexeldeib, line 186: remove before merge, just for testing runtime retag

ganeshkumar5699: just curious, how did you test this? did you check the tags in a node in moon cake, in a standalone environment?

pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line33.sh
```diff
@@ -97,7 +97,9 @@ ensureMonitorService
 # must run before kubelet starts to avoid race in container status using wrong image
 # https://github.com/kubernetes/kubernetes/issues/51017
 # can remove when fixed
-cleanupRetaggedImages
+if [[ "AzurePublicCloud" == "AzureChinaCloud" ]]; then
```
ganeshkumar5699, line 100: basic question: isn't this check only doing a string comparison that always evaluates to false? I went through the link in the comment but I'm probably missing something about this check.

anujmaheshwari1: you are looking at an autogenerated file, line 33.sh
these test files essentially act as what the cse will look like for certain NodeBootstrappingConfigurations(in this case, it is AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig) (look at the file path)
The actual check is in cse_main.sh 
if [[ "{{GetTargetEnvironment}}" == "AzureChinaCloud" ]]; then
    retagMCRImagesForChina
fi

so what the check essentially does is if the target environment is AzureChinaCloud, then the retag function is called. For the file you commented on, the target env is AzurePublicCloud, as expected from the test configuration,

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -257,6 +257,32 @@ retagContainerImage() {
     fi
 }
 
+retagMCRImagesForChina() {
+    # retag all the mcr for mooncake
+    if [[ "${CONTAINER_RUNTIME}" == "containerd" ]]; then
+        # shellcheck disable=SC2016
+        allMCRImages=($(ctr --namespace k8s.io images list | grep '^mcr.microsoft.com/' | awk '{print $1}'))
+    else
+        # shellcheck disable=SC2016
```
anujmaheshwari1, line 266: Does this disable the shellcheck? I remember it failing for a format like ($(...))

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: i changed the logic a bit, I testing by forcing cleanup and then running retag in public cloud, then changing the if condition from != back to ==

parts/linux/cloud-init/artifacts/cse_install.sh
alexeldeib: I just copied this from install-deps.sh, but I think it's the $vars in single quotes https://github-wiki-see.page/m/koalaman/shellcheck/wiki/SC2016

I think it's fine as is

anujmaheshwari1 (approved):

pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line33.sh
ganeshkumar5699: Thanks for explaining this @anujmaheshwari1! So `GetTargetEnvironment` would be replaced depending on the environment in which it is run in.

PR #1418, opened 2022-01-12
chore: Update Windows SIG image version to 2022-1B

ShiqianTao:


AbelHu (approved):

pkg/agent/datamodel/sig_config.go
```diff
@@ -205,7 +205,7 @@ const (
 
 const (
 	LinuxSIGImageVersion   string = "2022.01.08"
-	WindowsSIGImageVersion string = "17763.2366.211215"
+	WindowsSIGImageVersion string = "17763.2452.220107"
```
AbelHu, line 208: Why is the version 220107?

AbelHu: It should be today's date 220112.

ShiqianTao: Updated, thank you.

AbelHu (approved):

PR #1430, opened 2022-01-17
Revert "chore: Update Windows SIG image version to 2022-1B (#1418)"

AbelHu:
This reverts commit bd16ffcf645590df7aa94045ac927d107928bd4c.
gMSAv2 is blocked by 2022-1B. We need to revert this change to avoid others to bring this change to AKS RP before we figure out the solution.


AbelHu:
I have reverted the commit in official/v20220108

ShiqianTao (approved):

PR #1436, opened 2022-01-18
pin moby-containerd to 1.4.11

anujmaheshwari1:
moby-containerd 1.4.11 is the last stable version that runs with runc 1.0.0-rc95. installing the above runc version with a version of moby containerd newer than 1.4.11 will cause it to be uninstalled therefore also causing moby-engine to be uninstalled.


cpuguy83 (approved):
LGTM

alexeldeib (approved):

PR #1449, opened 2022-01-21
Add 1.23.2, hotfixes to prev 1.23 versions, and update docker version in test data

ganeshkumar5699:


pkg/agent/testdata/AKSWindows2019+CustomCloud/CustomData
```diff
@@ -89,7 +89,7 @@ $global:ContainerdUrl = "https://k8swin.blob.core.windows.net/k8s-windows/contai
 $global:ContainerdSdnPluginUrl = ""
 
 ## Docker Version
-$global:DockerVersion = "20.10.6"
+$global:DockerVersion = "20.10.9"
```
alexeldeib, line 92: where'd this come from? missed make generate somewhere?

alexeldeib (approved):
lgtm but curious how that slipped in

pkg/agent/testdata/AKSWindows2019+CustomCloud/CustomData
ganeshkumar5699: yes, I think `make generate` was not run with this commit: https://github.com/Azure/AgentBaker/commit/0d64f6cd36ae463ef383ca42b007ec6ad772babd

ganeshkumar5699: Perhaps, the CI doesn't check that `make generate` updates files in `testdata`. Should check and fix that in the future, in that case.

PR #1453, opened 2022-01-22
support configurable containerd and runc url

mainred:
Currently, only the versions of containerd and runc are configurable by picking the pre-cached deb package file, this makes tests with containerd and runc from other sources impossible, this change introduces new fields in NodeBootstrappingConfiguration, `ContainerdPackageURL` and `RuncPackageURL` compared to existing `ContainerdVersion` and `RuncVersion` . These changes also changed the CSE file to always install containerd or runc from the package URLs first if specified, and exit without trying other installation methods.


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -129,42 +129,66 @@ updateAptWithMicrosoftPkg() {
...
+    CONTAINERD_PACKAGE_URL="${CONTAINERD_PACKAGE_URL:=}"
+    # the user-defined package url is always picked first, and the others options won't tried when this one fails
+    if [[ ! -z ${CONTAINERD_PACKAGE_URL} ]]; then
+        echo "Installing containerd from user input: ${CONTAINERD_PACKAGE_URL}"
+        # NOTE(mainred): the containerd arch depends on the input from the user
+        mkdir -p $CONTAINERD_DOWNLOADS_DIR
+        CONTAINERD_DEB_TMP=${CONTAINERD_PACKAGE_URL##*/}
+        CONTAINERD_DEB_FILE="$CONTAINERD_DOWNLOADS_DIR/${CONTAINERD_DEB_TMP}"
+        retrycmd_curl_file 120 5 60 ${CONTAINERD_DEB_FILE} ${CONTAINERD_PACKAGE_URL} || exit $ERR_CONTAINERD_DOWNLOAD_TIMEOUT
+        if [[ ! -f "${CONTAINERD_DEB_FILE}" ]]; then
```
alexeldeib, line 146: checking the file right after downloading? just let the download fail?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -129,42 +129,66 @@ updateAptWithMicrosoftPkg() {
...
+            echo "Failed to install containerd from user input: ${CONTAINERD_PACKAGE_URL}, downloaded file is not found"
+            exit $ERR_CONTAINERD_INSTALL_FILE_NOT_FOUND
+        fi
         removeMoby
```
alexeldeib, line 150: don't need these the way you've refactored this, was only necessary because different version was already installed

mainred: It's a bit different, though we have the same containerd installed on the node, still, we want to use a user-defined one considering a hotfix with the same version or that from the user-owned pipeline, etc.
In one word, we want to use the containerd anyway when specified. I can add a comment here.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
mainred: exit with `$ERR_CONTAINERD_DOWNLOAD_TIMEOUT`, though not accurate in case of other failures, won't co-exist with $CONTAINERD_DEB_FILE is not null, right? Let me simplify the code to remove this `if` then.

mainred:
validated the change in standalone environment for URLs are specified and not.

alexeldeib:
sorry, caused some conflicts because we had to tweak containerd/runc install a bit for 1.5.9/1.4.12 upgrade. but I think it should make your life easier -- we always use the URL for containerd.

mostly lgtm if you can rebase the containerd install

mainred:
@alexeldeib, I have tested my change in AKS, please take another look, Thanks.

feiskyer (approved):

PR #1454, opened 2022-01-23
Add node fqdn to Kubelet certificate

phealy:
Adds the node FQDN to the Kubelet certificate.


phealy:
Doesn't work yet - do not merge.

phealy:
So this works in that the commands are run, but `hostname -f` isn't returning the FQDN this early in the build cycle. Any thoughts on a good way to fix this, @UtheMan @alexeldeib @anujmaheshwari1?

alexeldeib:
@phealy just doing some housekeeping -- what do you want to do with this one? don't think we found a good way in the short term, might be worth just waiting for us to have all the stuff for proper CSR bootstrapping?

cameronmeissner:
no longer needed

PR #1456, opened 2022-01-23
Fix quoting on systemd Environment= drop-ins

phealy:
Fixes quoting on systemd environment file drop-ins. This actually works fine on newer versions, but breaks node registration on older systemd versions (e.g. you can't build a 1.15.12 test cluster without this change).


alexeldeib (approved):
nice catch!

PR #1467, opened 2022-01-27
use manifest install for containerd, add some manifest schema validation

alexeldeib:


e2e/e2e-script.sh
```diff
@@ -167,7 +164,9 @@ done
 waitForNodeEndTime=$(date +%s)
 log "Waited $((waitForNodeEndTime-waitForNodeStartTime)) seconds for node to join"
 
-# trap 'az vmss delete -g $MC_RESOURCE_GROUP_NAME -n $VMSS_NAME --no-wait' EXIT
+kubectl get nodes
+
+trap 'az vmss delete -g $MC_RESOURCE_GROUP_NAME -n $VMSS_NAME --no-wait' EXIT
```
anujmaheshwari1, line 169: should this step be moved to after the check in L172? i.e should we delete the vmss even if the node failed to joined the cluster before fetching the logs?

anujmaheshwari1 (approved):

e2e/e2e-script.sh
alexeldeib: it's a trap so it only executes when the entire script exits, we fetch logs before deleting the vmss

vhdbuilder/packer/manifest.json
```diff
@@ -0,0 +1,43 @@
...
+        "installedVersion": "1.0.3"
+    },
+    "nvidia-container-runtime": {
+        "fileName":         "",
+        "downloadLocation": "",
+        "downloadURL":      "",
+        "versions": [],
+        "installedVersion": ""
+    },
+    "nvidia-drivers": {
```
alexeldeib, line 29: TODO

PR #1488, opened 2022-01-31
e2e script improvement #1

anujmaheshwari1:
What this PR addresses:

1) Bug: ACS-Test does Garbage Collection every 3 days. This deleted the MC RG corresponding to agentbaker_e2e cluster, however, the cluster itself isn't deleted, causing an error while retrieving cluster-info referring to a deleted MC RG. 
    Fix: Added logic to check if cluster exists and MC RG doesn't and carry out proper cleanup

2) Bug: Logic to check if Pod successfully runs on a newly joined Node checks as -> kubectl get pods -o wide | grep 'Running', which can result in false-positive w.r.t any Running pod.
    Fix: grep pod first, then test for 'Running'

3) Improvement: Randomize the Pod container name to avoid errors in concurrent runs


alexeldeib (approved):
lgtm

PR #1491, opened 2022-01-31
chore: install containerd from manifest in cse

alexeldeib:


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -14,26 +14,48 @@ UBUNTU_RELEASE=$(lsb_release -r -s)
...
+        stable_containerd="$(jq -r .containerd.installed.stable "$MANIFEST_FILEPATH")"
+        latest_containerd="$(jq -r .containerd.installed.latest "$MANIFEST_FILEPATH")"
+    else
+        echo "WARNING: containerd version not found in manifest, defaulting to hardcoded."
+    fi
+
+    # todo(ace): read 1.22 from a manifest and track it against supported versions
+    if semverCompare ${KUBERNETES_VERSION} "1.22.0"; then
+        containerd_version="$(echo "$latest_containerd" | cut -d- -f1)"
+        containerd_patch_version="$(echo "$latest_containerd" | cut -d- -f2)"
```
ganeshkumar5699, line 41: Getting the `stable_containerd` and `latest_containerd` versions and using it in such places is nice!

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -14,26 +14,48 @@ UBUNTU_RELEASE=$(lsb_release -r -s)
...
+        if [ -z "$container_verson" ]; then
+            containerd_version="1.5.9"
```
ganeshkumar5699, line 43: in the future, perhaps, we can even remove this hardcoded version with the assumption that the manifest file is present (and error after that file check instead of returning a warning)?

parts/linux/cloud-init/artifacts/cse_install.sh
alexeldeib: yeah, I'd love to find a way to make it more "static" so that we can catch failures earlier. e.g., this still requires indexing into the manifest. maybe we could have `make generate` parse the manifest and feed raw vars into the bash scripts or something. 

as-is, if the jq parse has typos, you'll have a bad time

alexeldeib: eh, actually jq parse failures return null, not empty str, which does error out the containerd install...I guess that's good enough since it'll fail e2e as well as vhd pr build

parts/linux/cloud-init/artifacts/cse_install.sh
alexeldeib: I went ahead and removed this to force the errors to the surface.

ganeshkumar5699 (approved):

PR #1492, opened 2022-02-01
Support new patch versions: 1.20.15, 1.21.9, 1.22.6, with latest hotfix

ganeshkumar5699:


anujmaheshwari1 (approved):

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -221,13 +221,13 @@ testKubeBinariesPresent() {
   k8sVersions="
   1.19.11
   1.19.13
-  1.20.9
   1.20.13
-  1.21.2-hotfix.20220130
-  1.21.7-hotfix.20220130
```
alexeldeib, line 227: err, we downgraded 1.21.7, 1.22.4, etc? that doesn't seem right

alexeldeib: ah content test only checks major.minor.patch?

alexeldeib: hmm no seems like it checks both down there...

ganeshkumar5699: I think the checks still pass if both the correct major.minor.patch versions are present (the VHD didn't have the same hotfix versions in this test file for an earlier PR, for instance)

ganeshkumar5699: With frequent changes in versions, I'd thought it would be easier (in terms of making consistent changes) to just check major.minor.patch instead of the hotfix version as well and it wouldn't change much of the value provided by this test.

ganeshkumar5699: I'll add the full hotfix versions back for now though - eventually, we can move these versions to a common config file.

ganeshkumar5699: fixed it, thanks!

PR #1497, opened 2022-02-02
Install FIPS-updates instead of FIPS to get more updates

chengliangli0918:
Install FIPS-updates instead of FIPS to get more updates

In 2021, the FIPS modules installed by ‘**ua enable fips-updates**’ is not NIST certified (Figure 1), so we use '**ua enable fips**' to install fips modules (kernel version is always linux-image-4.15.0-**1002**-azure-fips)

Recently Cannonical got NIST certification for more patches to FIPS modules, now packages installed by ‘**ua enable fips-updates**’ are also NIST-Certified (Figure 2). So we can use '**ua enable fips-updates**’ to install more recent NSIT-certified FIPS modules.

**Figure 1**
image

**Figure 2**
image


anujmaheshwari1 (approved):
nice, this should avoid issues like out of sync symlink in the future. lgtm, approved.

PR #1507, opened 2022-02-04
fix: windows dynamic-config-dir

alexeldeib:


pkg/agent/baker.go
```diff
@@ -263,7 +263,9 @@ func validateAndSetLinuxNodeBootstrappingConfiguration(config *datamodel.NodeBoo
 	if config.KubeletConfig != nil {
 		kubeletFlags := config.KubeletConfig
 		delete(kubeletFlags, "--dynamic-config-dir")
-		if IsKubernetesVersionGe(config.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion, "1.11.0") {
+		if IsKubernetesVersionGe(config.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion, "1.24.0") {
+			kubeletFlags["--feature-gates"] = remoteFeatureGateString(kubeletFlags["--feature-gates"], "DynamicKubeletConfig")
```
ganeshkumar5699, line 267: is this k8s >= 1.24 check added since dynamic kubelet config flag is removed (and not just set to false) from k8s 1.24?

pkg/agent/utils.go
```diff
@@ -576,6 +576,21 @@ func strKeyValToMapBool(str string, strDelim string, pairDelim string) map[strin
 	return m
 }
 
+func remoteFeatureGateString(featureGates string, key string) string {
```
ganeshkumar5699, line 579: is this supposed to be "remove" instead of "remote"?

ganeshkumar5699 (approved):
lgtm. And asked those two questions to confirm.

pkg/agent/baker.go
alexeldeib: correct https://github.com/kubernetes/kubernetes/pull/106932

pkg/agent/utils.go
alexeldeib: oops :)

pkg/agent/baker.go
alexeldeib: I think it's actually locked to false anyway, but might as well tidy it up on our side

PR #1508, opened 2022-02-04
e2e script improvement #2

anujmaheshwari1:
This PR addresses the following:
1) Flaky permission error for cluster-provision.log, add sudo to always be able to fetch the log
2) Delete VMSS and subsequently also delete node from the cluster(since it turns to NotReady post VMSS deletion)

### DIFF
--- a/e2e/e2e-script.sh
+++ b/e2e/e2e-script.sh
@@ -181,12 +181,7 @@ done
 waitForNodeEndTime=$(date +%s)
 log "Waited $((waitForNodeEndTime-waitForNodeStartTime)) seconds for node to join"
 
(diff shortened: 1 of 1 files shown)


e2e/e2e-script.sh
```diff
@@ -247,5 +242,13 @@ else
     exit 1
 fi
 
+waitForDeleteStartTime=$(date +%s)
+
+az vmss delete -g $MC_RESOURCE_GROUP_NAME -n $VMSS_NAME
```
anujmaheshwari1, line 247: the delete sequence takes ~1minute to run, doing some more test runs to see if cloud controller manager deletes the node after certain amount of time, if yes, will add --no-wait to vmss delete and remove the kubectl delete.

alexeldeib (approved):

PR #1512, opened 2022-02-05
chore: try to auto-update containerd

alexeldeib:


.github/workflows/update-containerd.yml
```diff
@@ -0,0 +1,112 @@
...
+          which rg
+
+          work=$(mktemp -d)
+          pushd $work
+          trap 'popd; rm -rf $work' EXIT TERM
+
+          git clone https://github.com/containerd/containerd
+          cd containerd/
+          
+          latest_two_major_minor="$(git tag -l | sort -V | rg 'v[0-9]+.[0-9]+.[0-9]+$' | cut -d'.' -f1,2 | uniq | tail -n 2)"
```
ganeshkumar5699, line 56: Clever! It’s nice that you are using git tags in the repo to determine the latest versions

.github/workflows/update-containerd.yml
```diff
@@ -0,0 +1,112 @@
...
+          should_update_stable="false"
+
+          exists=false
+          i=1
+          retval=0
+          curl -fsSL -I "https://moby.blob.core.windows.net/moby/moby-containerd/${stable_patch_version}+azure/bionic/linux_amd64/moby-containerd_${stable_patch_version}+azure-${i}_amd64.deb" || retval=$?
+          while [ "$retval" == "0" ]; do
+            exists="true"
+            i=$((i+1))
+            curl -fsSL -I "https://moby.blob.core.windows.net/moby/moby-containerd/${stable_patch_version}+azure/bionic/linux_amd64/moby-containerd_${stable_patch_version}+azure-${i}_amd64.deb" || retval=$?
```
ganeshkumar5699, line 77: Could there be a case where azure-${i} is not available but azure ${i + n}, n > 0, is available? Seems like this would be a rare edge case (if it happens at all), as my understanding is that older versions would still available.

alexeldeib: quite possibly. tbh we should reshuffle the whole thing to just take latest from apt if we can list it that way (even though we're in storage). 

or maybe we can figure out a way to list storage blobs (don't love that)

alexeldeib: or if we can search for latest revision somehow by major.minor.patch...

ganeshkumar5699: yeah, that would be nice. Last time I checked, `apt-cache madison`  for `moby-containerd` only had containerd versions < 1.5 and  `apt-cache madison` for `containerd` had versions > 1.5 but it didn't have the latest versions. So it seems like the results were not comprehensive.

ganeshkumar5699 (approved):
lgtm pending a branch update.

ganeshkumar5699:
This containerd version automation will be really helpful in improving the time it takes to test and update containerd and it should also improve AB security due to faster patch updates.

Devinwong:
abandon this PR as it's stale

PR #1522, opened 2022-02-09
Handling for default container log max size

UtheMan:
Set default value for Container Log Max Size to 50MB


pkg/templates/templates_generated.go
```diff
@@ -4568,6 +4623,43 @@ func linuxCloudInitArtifactsUbuntuCse_install_ubuntuSh() (*asset, error) {
 	return a, nil
 }
 
+var _linuxCloudInitConfigHcpmumanskiebld51438502EastusSvc0 = []byte(`apiVersion: v1
```
anujmaheshwari1, line 4626: some make target probably messed up depending on which you ran? these shouldnt be here

pkg/templates/templates_generated.go
```diff
@@ -58,6 +61,7 @@
 // linux/cloud-init/artifacts/sysctl-d-60-CIS.conf
 // linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
 // linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
+// linux/cloud-init/config.hcpmumanskiebld51438502-eastus-svc-0
```
anujmaheshwari1, line 64: aah this is the culprit I suppose. and L3-5.

pkg/agent/utils_test.go
```diff
@@ -258,3 +417,11 @@ var _ = Describe("Assert ParseCSE", func() {
 		Expect(err.Error()).To(ContainSubstring(testMessage))
 	})
 })
+
+func TestKubeletConfigDefaultsShouldBeSetIfCustomKubeletConfigNotProvided(t *testing.T) {
```
anujmaheshwari1, line 421: why do we have empty tests?

pkg/templates/templates_generated.go
ganeshkumar5699: re-running `make generate` again with a clean branch with only necessary changes, may make it work or at least help check that this was added by the make target.

UtheMan: Seems like removing accidentally added files helped with make generate

alexeldeib (approved):

ganeshkumar5699 (approved):

PR #1523, opened 2022-02-09
Switch to using templated JSON for e2e

anujmaheshwari1:
This PR switches to using a templated JSON (percluster and NodeBootstrappingConfig) to ease adding more test cases.
Adds logic to populate a percluster JSON with relevant cluster info, then does a patch merge to the actual NodeBootstrappingConfig.


e2e/e2e-script.sh
```diff
@@ -123,9 +120,9 @@ clusterInfoEndTime=$(date +%s)
 log "Retrieved cluster info in $((clusterInfoEndTime-clusterInfoStartTime)) seconds"
 
 set +x
-addJsonToFile "apiserver.crt" "$(cat apiserver.crt)"
-addJsonToFile "ca.crt" "$(cat ca.crt)"
-addJsonToFile "client.key" "$(cat client.key)"
+addJsonToFile "apiserverCrt" "$(cat apiserver.crt)"
```
anujmaheshwari1, line 123: changed from file format to camel case because bash doesnt allow exporting variables with special chars.

alexeldeib (approved):

e2e/e2e-script.sh
```diff
@@ -141,6 +138,13 @@ addJsonToFile "mcRGName" $MC_RESOURCE_GROUP_NAME
 addJsonToFile "clusterID" $CLUSTER_ID
 addJsonToFile "subID" $SUBSCRIPTION_ID
 
+set +x
+jq -r 'keys[] as $k | "export \($k)=\(.[$k])"' fields.json
```
anujmaheshwari1, line 142: @alexeldeib this line is troublesome. need to fine another way to export the var to be able to set +x around it.

PR #1530, opened 2022-02-11
Update hotfixes and remove 1.19 versions for linux

ganeshkumar5699:


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -450,15 +450,13 @@ done
 # Please do not use the .1 suffix, because that's only for the base image patches
 
 AMD64_ONLY_KUBE_BINARY_VERSIONS="
```
ganeshkumar5699, line 452: Leaving this var for now in case we end up adding previous versions again. Should be able to remove it in the near future.

anujmaheshwari1 (approved):

PR #1547, opened 2022-02-18
Use disk snapshot instead of managed image to capture the arm64 OS disk

chengliangli0918:
For AMD64 Gen2, after k8s related installation, the OS is generalized, and captured as **managed image**, the managed images then is converted into a SIG, at last the SIG is converted into a VHD file with SAS to be published.

However **Managed Image** is in maintenance mode, Azure Compute is not planning to support ARM64 architecture scenarios for managed image resource. So capturing generalized OS disk of ARM64 VM to a managed image is forbidden by Azure Compute stack. To workaround this restriction, the generalized OS disk of ARM64 VM is **snapshotted**, then the os disk snapshot is converted into a SIG, the remaining operations are the same as AMD64 Gen2.

This change is on a private branch for some time, and was used to build weekly AKS ARM64 VHD for a few weeks in Jan and Feb, and worked fine. Now merge this change to master using this PR.

All amd64 image build still works fine: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=52128238


anujmaheshwari1 (approved):
lgtm, quick question, I suppose you tested building the VHD already?

chengliangli0918:
Yes, tests all vhd building in https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=52128238.
> lgtm, quick question, I suppose you tested building the VHD already?

PR #1573, opened 2022-02-25
Always cache cni versions currently in use by ARM64

chengliangli0918:
In January, the CNI 0.8.7 was cached on VHDs, so ARM64 uses 0.8.7 CNI. Later 0.8.7 was removed and 0.9.1 was added to VHD. In most case, this is not an issue because 0.8.7 will be downloaded during cse provisioning on ARM64 nodes. But for 'egress lockdown' case, the cse will fail with CSE error 41.

This change will always cache cni version currently in use by ARM64 node, no matter how the common versions list is changed. This change doesn't affect contents of amd64 VHD.


mattstam:
cc @rbtr

mattstam:
Change sounds good to me, though want to get Evan's input on this as well.

mattstam (approved):

rbtr:
this change looks fine to me as a one-off, but my question is how do we maintain this? what is source-of-truth for the version of the CNIs that's used by ARM64 vs AMD64?

chengliangli0918:
> this change looks fine to me as a one-off, but my question is how do we maintain this? what is the source-of-truth for the version of the CNIs used by ARM64 vs AMD64?

For amd64, the AKS RP still uses 0.7.6 cni binaries (aks/rp/common/datamodel/const.go), and 1.4.12 azure cni and swift cni binaries (aks/rp/resourceprovider/server/microsoft.com/containerservice/flags/flags.go). Those version doesn't have corresponding ARM64 binaries. So ARM64 have to use 0.8.7 cni and 1.4.14 azure cni/swift cni binaries. If amd64 bump cni/azure&swift cni up to 0.8.7/1.4.13+ in RP side, then amd64/arm64 can use the same version, the installation logic here could be simple too.

Does AKS Networking SIG have plan to upgrade cni/azure&swift cni binaries version?

PR #1575, opened 2022-02-28
build: upgrade azure-cloud-node-manager versions

lzhecheng:
Signed-off-by: Zhecheng Li <zhechengli@microsoft.com>


vhdbuilder/packer/components.json
```diff
@@ -240,14 +240,14 @@
...
       ],
       "multiArchVersions": [
-        "v0.7.11",
-        "v0.7.12",
-        "v1.0.8",
-        "v1.0.9",
-        "v1.1.4",
-        "v1.1.5",
-        "v1.23.1",
-        "v1.23.2"
```
lodrem, line 250: We should keep the last updated versions.
```
v0.7.12  // last updated 
v0.7.14  // new one
```

lzhecheng: Is it correct? Even this time I bump up 2 versions, I still need to keep the last updated version?

lodrem: I think the reason we keep the last updated version is that RP would use it according to its config. We need to update RP and bring it to the latest version. (but still, it could be a gap that RP uses the newest VHD and requires the last updated version)

For this cause, should we bump the latest versions which are `v0.7.14` / `v1.0.11` / `v1.1.7` / `v1.23.4` and the last updated versions? And we don't need other versions such as `v0.7.13`.

@feiskyer would you plz confirm the logic that we keep "two" versions here?

zhiweiv:
Can you update the version to 1.23.5? It contains an important fix for AAD Pod Identity.
https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1205
https://github.com/kubernetes-sigs/cloud-provider-azure/issues/1155
https://github.com/Azure/aad-pod-identity/issues/1246

vhdbuilder/packer/components.json
feiskyer: yes, we need to keep the last version updated in RP and the latest version.

@lzhecheng please update the versions to today's new releases.

lzhecheng: Got it.

lzhecheng:
@feiskyer @zhiweiv I have updated the code. PTAL

Keep for RP:
```
0.6.0
0.7.12
1.0.9
1.1.5
1.23.2
```

Latest:
```
0.7.15
1.0.12
1.1.8
1.23.5
```

lzhecheng:
Check `validate-image-version` fails because the site is unstable right now.
http://mcr.microsoft.com/v2/oss/kubernetes/azure-cloud-node-manager/tags/list

feiskyer (approved):

PR #1592, opened 2022-03-03
Enable/disable unattended upgrades using smartCSE

MirzaSikander:


clayvan:
Hey [MirzaSikander](https://github.com/MirzaSikander), I'm super interested in this and curious if you ever plan to pick this back up? Maybe share some original intentions behind it?

pkg/agent/baker.go
```diff
@@ -787,5 +787,8 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"HasMessageOfTheDay": func() bool {
 			return profile.MessageOfTheDay != ""
 		},
+		"IsUnattendedUpgradeEnabled": func() bool {
+			return config.IsUnattendedUpgradeEnabled
```
paulgmiller, line 791: can't start with this defaulted to false or have to change rp when you vendor it in which is risky

PR #1610, opened 2022-03-10
Containerd 1.6.4 for k8s >= 1.24

ganeshkumar5699:
Containerd 1.6.4 works with k8s 1.24, in the standalone environment!


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -30,23 +30,31 @@ installContainerRuntime() {
...
         stable_containerd="$(jq -r .containerd.stable "$MANIFEST_FILEPATH")"
         latest_containerd="$(jq -r .containerd.latest "$MANIFEST_FILEPATH")"
+        edge_containerd="$(jq -r .containerd.edge "$MANIFEST_FILEPATH")"
     else
         echo "WARNING: containerd version not found in manifest, defaulting to hardcoded."
     fi
 
     # todo(ace): read 1.22 from a manifest and track it against supported versions
-    if semverCompare ${KUBERNETES_VERSION} "1.22.0"; then
+    if semverCompare ${KUBERNETES_VERSION} "1.24.0"; then
```
alexeldeib, line 39: would be nice if we can figure out a way to structure this so we can just enumerate (k8s version, containerd version) tuples and compare against desired version instead of branching this case more. 

in practice we probably only have a few

alexeldeib (approved):

PR #139, opened 2020-06-26
Support dynamic kubelet

bowang-666:
Translate current flags we set to config file before running kubelet.
Flags can be found in https://github.com/Azure/aks-engine/blob/aks-release-v0.47.0-1/pkg/api/defaults-kubelet.go
and some RP files.
Remove translated flags from flag string, but some flags will still remain (for example azure specific flags)
Configurations in config file can be changed at run time following https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/

PR #1611, opened 2022-03-10
feat: support customized kubelet configuration

mainred:
Besides customizing node configuration to change the default kubelet config, we introduce in this PR the ability to configure both the args of kubelet and download URL, as part of our custom configuration feature.


pkg/agent/utils.go
```diff
@@ -358,18 +358,44 @@ func GetOrderedKubeletConfigFlagString(k map[string]string, cs *datamodel.Contai
 	kubeletConfigFileEnabled := IsKubeletConfigFileEnabled(cs, profile, kubeletConfigFileToggleEnabled)
 	keys := []string{}
 	for key := range k {
-		if !kubeletConfigFileEnabled || !TranslatedKubeletConfigFlags[key] {
-			keys = append(keys, key)
+		// TODO(qinhao): need to consider the case when enable-kubelet-config-file is enabled
+		if kubeletConfigFileEnabled && TranslatedKubeletConfigFlags[key] {
```
mainred, line 362: `&&`  is more straightforward.

mainred, line 362: Custom configuration includes not only the configuration of kubelet but also kube URL, as well as other components' images and configurations for internal use. These will be a conflict with [custom node configuration](https://docs.microsoft.com/en-us/azure/aks/custom-node-configuration), but the two features are not expected to be used for the same cluster.
TranslatedKubeletConfigFlags is introduced for dyanmic kubelet by https://github.com/Azure/AgentBaker/pull/139/files, and has been disabled by @anujmaheshwari1 in AKS

mainred:
@alexeldeib @ganeshkumar5699 @feiskyer  please take a look.

pkg/agent/params.go
```diff
@@ -62,6 +62,15 @@ func getParameters(config *datamodel.NodeBootstrappingConfiguration, generatorCo
 		}
 	}
 
+	if properties.CustomConfiguration != nil && properties.CustomConfiguration.KubernetesConfigurations != nil {
+		for componentName, configuration := range properties.CustomConfiguration.KubernetesConfigurations {
```
alexeldeib, line 66: nit: why bother looping over a map when you are going to switch on component name anyway, and hardcode both componentName and customKubeBinaryUrl (i.e., don't use the keys for templating)

pkg/agent/utils.go
```diff
@@ -370,6 +378,50 @@ func GetOrderedKubeletConfigFlagString(k map[string]string, cs *datamodel.Contai
 	return buf.String()
 }
 
+func getOrderedKubeletConfigFlagWithCustomConfigurationString(customConfig, defaultConfig map[string]string) string {
+	config := customConfig
+
+	if defaultConfig != nil {
+		for k, v := range defaultConfig {
+			// override configurations from default values to customized ones
+			if _, ok := config[k]; ok {
```
alexeldeib, line 387: nit: you don't need the if statement if you overwrite default config with custom config

on L395 you can range a nil map safely so it doesn't matter if defaultConfig is nil on L384 (assuming this is why you had it flipped to begin with)

alexeldeib (approved):
mostly nits, would appreciate an answer to https://github.com/Azure/AgentBaker/pull/1611#discussion_r838767904 though -- seems at a glance that extractKubeBinaries should work fine, not sure what the diff is here. but seems to work either way :)

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -227,21 +227,37 @@ extractHyperkube() {
...
+        # remove the kubelet binaries to make sure the only binary left is from the CUSTOM_KUBE_BINARY_DOWNLOAD_URL
+        rm -rf /usr/local/bin/kubelet-* /usr/local/bin/kubectl-*
+
+        # NOTE(mainred): we expect kubelet binary to be under `kubernetes/node/bin`. This suits the current setting of
+        # kube binaries used by AKS and Kubernetes upstream.
+        # TODO(mainred): let's see if necessary to auto-detect the path of kubelet
+        mkdir -p ${K8S_DOWNLOADS_DIR}
+        K8S_TGZ_TMP=${CUSTOM_KUBE_BINARY_DOWNLOAD_URL##*/}
+        retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" ${CUSTOM_KUBE_BINARY_DOWNLOAD_URL} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
+        tar -xzvf "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" --strip-components=3 -C /usr/local/bin kubernetes/node/bin/kubelet kubernetes/node/bin/kubectl
```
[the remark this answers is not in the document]
mainred: Thanks for pointing this out, will reuse `extractKubeBinaries`. My original thought is that we want to accept a tarball with a more flexible dir structure. But it seems at least for now it's not necessary as the tarball used by AKS and upstream are with the same structure.

pkg/agent/params.go
mainred: will use logic like `if val, ok := m["kubelet"]; ok` instead.

mainred:
Thanks @alexeldeib. The change now looks more simple and straightfoward.

PR #1613, opened 2022-03-10
add logrotate for rsyslog warn

alexeldeib:
we split part of /var/log/messages out to /var/log/warn, but there is no default logrotate configuration for /var/log/warn like there is for /var/log/messages. this means /var/log/warn can grow infinitely. this PR adds some configuration for log rotation on that file.


phealy (approved):
lgtm

PR #1632, opened 2022-03-16
update cni versions

rbtr:
Signed-off-by: Evan Baker <rbtr@users.noreply.github.com>
/cc @AbelHu @chengliangli0918

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -453,7 +453,7 @@
       "downloadURL": "https://acs-mirror.azureedge.net/azure-cni/v*/binaries",
       "versions": [
         "1.2.7",
-        "1.4.16",
+        "1.4.14",
         "1.4.21"
       ]
     },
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/components.json
```diff
@@ -453,9 +453,8 @@
       "downloadURL": "https://acs-mirror.azureedge.net/azure-cni/v*/binaries",
       "versions": [
         "1.2.7",
-        "1.4.13",
-        "1.4.14",
-        "1.4.16"
+        "1.4.16",
```
paulgmiller, line 456: have to leave 1.4.14 since apparently thats what we're on?
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/toggles/global/azure-cni-linux-url.yaml

rbtr: and swift is on 1.4.12, so i have left 1.4.14 and 1.4.12 cached respectively.

paulgmiller (approved):

PR #1638, opened 2022-03-19
update omsagent images

pfrcks:
- update linux image to multiarch ciprod03172022
- update windows image to win-ciprod03172022


ganga1980 (approved):

AbelHu:
@pfrcks Will win-ciprod03172022 be used as next version? Could we remove win-ciprod03112022 since the VHD disk is not enough to cache three omsagent-win images? cc @ShiqianTao

ganga1980:
> @pfrcks Will win-ciprod03172022 be used as next version? Could we remove win-ciprod03112022 since the VHD disk is not enough to cache three omsagent-win images? cc @ShiqianTao

> @AbelHu ,  win-ciprod03112022 can be removed if there is space constraint.

PR #1646, opened 2022-03-21
Add release notes for 2022.03.21

anujmaheshwari1:


ganeshkumar5699 (commented):
Just curious: why were the mariner builds skipped? I checked the weekly VHD builds and it seemed like mariner builds were not selected in the build process. The azsecpack fix does add the "SkipLinuxAzSecPack": "true" to mariner builds though, so I was just wondering why mariner was also not part of the build.

anujmaheshwari1 (commented):
defer to @alexeldeib since he triggered the builds. I am guessing it has something to do with the mariner hotfix we did recently?

ganeshkumar5699 (approved):

PR #1653, opened 2022-03-22
generate image version from build_id for automation pipeline

anujmaheshwari1:
This PR aims at removing the IMAGE_VERSION as input from the automation pipeline. This reduces scope for human introduced error while inputting the image version for a given release.
Changing the logic to rely on the artifacts published from the build ID to fetch the image version and plumb it down to the other stages of the pipeline.

Test Run: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=53332454&view=results using the following build https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=53226610&view=results

Result: Correctly generated release notes PR: https://github.com/Azure/AgentBaker/pull/1652
(skipped image bump step to avoid cutting official branches that have no significance)


vhdbuilder/scripts/automate_generate_version.sh
```diff
@@ -0,0 +1,23 @@
...
+            az pipelines runs artifact download --artifact-name $artifact --path $(pwd) --run-id $build_id
+            current_image_version=$(jq -r .image_version < vhd-publishing-info.json)
+            if [[ $global_image_version != $current_image_version ]]; then
+                if [[ -z $global_image_version ]]; then
+                    global_image_version=$current_image_version
+                else 
+                    exit 1
+                fi
+            fi
+        fi
```
MirzaSikander, line 18: Will there be multiple artifacts matching the condition ever? From the code I feel there wouldn't be. 

It might be better to use to jq's filter to get the artifact you want. Also if the filter results in multiple items with the same name (maybe in the future there is a change that causes this) then just error out instead of using the wrong value.

Just put a TODO for now so that we can get it checked in.

MirzaSikander (approved):

vhdbuilder/scripts/automate_generate_version.sh
MirzaSikander: as discussed offline. Lets just put the context as comment.

MirzaSikander (approved):

PR #1654, opened 2022-03-22
Output cluster-provision.log instead of cluster-provision-cse-output.log

ganeshkumar5699:
- Changing this to provide additional logs for debugging CSE failures without entering a node


parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -2,7 +2,7 @@ CSE_STARTTIME=$(date)
 /bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1
 EXIT_CODE=$?
 systemctl --no-pager -l status kubelet >> /var/log/azure/cluster-provision-cse-output.log 2>&1
-OUTPUT=$(head -c 3000 "/var/log/azure/cluster-provision-cse-output.log")
+OUTPUT=$(perl -ne '/\n/ && print' "/var/log/azure/cluster-provision.log" "/var/log/azure/cluster-provision-cse-output.log")
```
alexeldeib, line 5: I think we should `tail -n $SOME_NUMBER_LESS_THAN_3000 cluster-provision.log`

I am not sure I understanding why perl is needed here, since the other file is multi-line and didn't have this?

ganeshkumar5699: Thank you. As per the discussion, I've switched to `tail -c 3000 cluster-provision.log`.

alexeldeib (approved):
lgtm, thanks!

ganeshkumar5699:
> lgtm, thanks!

Thanks for the find and suggestions for this @alexeldeib!

PR #1655, opened 2022-03-22
CSE logging test

ganeshkumar5699:


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -22,6 +22,8 @@ cleanupContainerdDlFiles() {
 
 installContainerRuntime() {
 {{if NeedsContainerd}}
+    echo "just testing CSE exit: ERR_RUNC_DOWNLOAD_TIMEOUT"
```
ganeshkumar5699, line 25: example of a deliberate CSE error, @Miranda

PR #1657, opened 2022-03-23
Skip install Azure Security Pack on AKS Windows nodes

ShiqianTao:


AbelHu (approved):

vhdbuilder/packer/windows-vhd-builder-sig.json
```diff
@@ -56,7 +56,8 @@
                 "os": "Windows",
                 "now": "{{user `create_time`}}",
                 "createdBy": "aks-vhd-pipeline",
-                "SkipNRMSNSG": "295551581"
+                "SkipNRMSNSG": "295551581",
+                "SkipASMAzSecPack": "true"
```
AbelHu, line 60: True?

AbelHu (changes requested):
I think that we should strictly follow the guidance to use `True` instead of `true`

vhdbuilder/packer/windows-vhd-builder-sig.json
ShiqianTao: Yes. Updated. Thank you.

AbelHu (approved):

PR #1658, opened 2022-03-23
chore: Cache new Azure CNI v1.4.22

AbelHu:
Cache new Azure CNI v1.4.22


ShiqianTao (approved):

ShiqianTao (approved):

vhdbuilder/packer/components.json
```diff
@@ -456,7 +456,8 @@
       "versions": [
         "1.2.7",
         "1.4.14",
```
paulgmiller, line 458: ar we still runnign tight on space? 1.4.14 ready for removal?

AbelHu: @paulgmiller I have no idea. Normally I always add a new Azure CNI but do not remove any old Azure CNI version for Linux since I do not know whether I can do it. If it needs to be removed, I think that it should be done in a separate PR. How do you think?

paulgmiller: defaultValue: "https://acs-mirror.azureedge.net/azure-cni/v1.4.14/binaries/azure-vnet-cni-linux-amd64-v1.4.14.tgz"

Turns out we never used 1.4.21 so you could remove that but .14 has to stay. I'd encourage you to remove .21 but won't block on it.

vhdbuilder/packer/components.json
```diff
@@ -466,7 +467,8 @@
       "versions": [
         "1.2.7",
         "1.4.12",
```
paulgmiller, line 469: her keep 12 but remove defaultValue: "https://acs-mirror.azureedge.net/azure-cni/v1.4.12/binaries/azure-vnet-cni-swift-linux-amd64-v1.4.12.tgz"

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -252,6 +252,7 @@ AMD64_ONLY_CNI_VERSIONS="
 MULTI_ARCH_VNET_CNI_VERSIONS="
 1.4.14
 1.4.21
```
paulgmiller, line 254: Again prety sure 1.4.21 can be removed though unless arm using it somehow. Have to check with Charlie li.

paulgmiller (approved):
Not sure why th extra exclusion of single tenancy-windows is neede in ps1  is needed but doesn't seem harmful so apprive. Charlie Li would know if arm is using 1.4.21 but didn't see it when I seached our code base.

AbelHu:
@paulgmiller Thanks for the checking. I will merge this PR to save the test cost. I can create a new PR to remove 1.4.12 later after Charlie Li confirm it. Thanks.

vhdbuilder/packer/install-dependencies.sh
AbelHu: @paulgmiller Charlie shared "1.4.21 is cached in VHD recently by networking SIG. not used in RP side so far." so I think that v1.4.21 should not be removed.

vhdbuilder/packer/components.json
AbelHu: @paulgmiller Charlie shared "v1.4.14 can not be removed." since it is the default toggle value.

vhdbuilder/packer/components.json
AbelHu: @paulgmiller Charlie shared "arm64 and amd64 is using swift cni 1.4.12" so it seems like that it should be kept.

PR #1666, opened 2022-03-24
Add an image architecture property in vhd-publishing-info

chengliangli0918:
Add an image architecture property in vhd-publishing-info in order that the SIG image definition can be created with right architecture properties in VHD release pipeline

SIG image definition for AMD64/ARM64 has subtle difference, otherwise a SIG version cannot be used to create VM/VMSS of corresponding sku. 'az sig image-definition create' will have a new property for this. 

In Public cloud, I manually create '1804gen2arm64containerd' image definition using private cli a few months ago. When this new property is available in new official CLI and architecture feature is available in FF/MC. The release script will use this new property in vhd-publishing-info to create right SIG image definition.


alexeldeib:
@chengliangli0918 looking to merge this? lgtm.

alexeldeib (approved):

chengliangli0918:
> @chengliangli0918 looking to merge this? lgtm.
@alexeldeib thanks for approving it. I am looking to merge it but not in a very hurry(that's the reason I didn't ping you guys to review it). Because for now in FF/MC, the Azure Compute stack doesn't support creating ARM64 image definition yet (az sig image-definition create --architecture Arm64 ...). This is also the reason that currently I always use separated SIG pipeline to publish arm64 sig to public cloud only, but not to FF/MC.

PR #1667, opened 2022-03-25
test: Check Azure Extensions in AKS Windows VHD build pipeline

ShiqianTao:
Test PackagesPlugins in Windows VHD build pipeline

### DIFF
--- a/vhdbuilder/packer/test/windows-vhd-content-test.ps1
+++ b/vhdbuilder/packer/test/windows-vhd-content-test.ps1
@@ -216,17 +216,17 @@ function Test-DefenderSignature {
     }
 }
 
-function Test-AzurePlugins {
-    $expectedPlugins = @(
+function Test-AzureExtensions {
+    $expectedExtensions = @(
         "Microsoft.CPlat.Core.RunCommandWindows"
     )
-    $actualPlugins = (Get-ChildItem "C:\Packages\Plugins").Name
-    $compareResult = (Compare-Object $expectedPlugins $actualPlugins)
+    $actualExtensions = (Get-ChildItem "C:\Packages\Plugins").Name
+    $compareResult = (Compare-Object $expectedExtensions $actualExtensions)
     if ($compareResult) {
-        Write-Error "Packages Plugins are not expected. Details: $compareResult"
+        Write-Error "Azure extensions are not expected. Details: $($compareResult | Out-String)"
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/test/windows-vhd-content-test.ps1
```diff
@@ -165,8 +165,25 @@ function Test-RegistryAdded {
     }
 }
 
+function Test-PackagesPlugins {
+    $expectedPlugins = @(
```
AbelHu, line 169: Are these package installed in current VHD? We may need to consider to avoid installing these packages.

ShiqianTao: Yes, the pipeline installs them.
1. AntimalwareSignature. If we need to remove it, I will investigate that how to avoid installing it.
2. RunCommandWindows. I think we can ignore it (just keep it in code), it exists in this pipeline process, and it is automatically removed from Windows VM & AKS Cluster Node.

AbelHu: You need to update both PR title and commit message to start with "test:". `extensions` may be better than `Plugins`.

vhdbuilder/packer/test/windows-vhd-content-test.ps1
```diff
@@ -213,9 +213,24 @@ function Test-DefenderSignature {
     }
 }
 
+function Test-AzureExtensions {
+    $expectedExtensions = @(
+        "Microsoft.CPlat.Core.RunCommandWindows"
```
AbelHu, line 218: I think whether we can remove this extension in the last step.

ShiqianTao: According to my investigation, it will be removed automatically after the pipeline.

ShiqianTao: Additional investigation: this Windows test is called by ```az vm run-command invoke --command-id RunPowerShellScript```, and this extension is installed by it. So it's expected in this context, and it is not existing in windows node.

ShiqianTao: Added notes for it.

AbelHu (approved):

PR #1668, opened 2022-03-25
Support using IPAddress for apiserver for linux node

gossion:
Currently, agentbaker requires apiserver endpoint to be FQDN. When we enable SWIFT stack, it is possible that we can remove the dns dependency and use IP address of apiserver directly. This PR is to add support for using IP address of apiserver for linux nodes.


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -226,28 +226,33 @@ VALIDATION_ERR=0
...
-STS=0
-{{- end}}
-if [[ $STS != 0 ]]; then
-    time nslookup ${API_SERVER_NAME}
-    if [[ $RES == *"168.63.129.16"*  ]]; then
-        VALIDATION_ERR=$ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL
+if ! [[ ${API_SERVER_NAME} =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+    API_SERVER_DNS_RETRIES=100
+    if [[ $API_SERVER_NAME == *.privatelink.* ]]; then
+    API_SERVER_DNS_RETRIES=200
```
levimm, line 237: more space here

gossion: added.

levimm (approved):

feiskyer (approved):
LGTM

This is only for Linux, please file a new PR for Windows

PR #1670, opened 2022-03-25
doc: add issue and PR templates

andyzhangx:
We need templates!


mainred (approved):
Let's go with this first and see how to suit our repo better. Thanks.

PR #1675, opened 2022-03-28
fix: chown azure cni to root

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -186,6 +186,7 @@ installAzureCNI() {
         downloadAzureCNI
     fi
     tar -xzf "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" -C $CNI_BIN_DIR
+    chown -R root:root $CNI_BIN_DIR
```
paulgmiller, line 189: Thanks for the quick suggestion Ace
The "Which issue(s) this PR fix" seems relevant.
Shoudd everything isntalled by CSE be root or do we really believe just cni is special?  Surely we're doing other tar -xzf's here....

alexeldeib: > Surely we're doing other tar -xzf's here

we are

> Shoudd everything isntalled by CSE be root
I think generally, yes (I think if anything needs to be accessible by non-root can use file permissions?)

> do we really believe just cni is special?

well, it's the only one which broke users ;) 

also, given this was due to a change in build infra and this is one of the few deps upstream doesn't build, I think it _is_ special in some sense (acknowledging the weakness of that argument in practice since upstream could have the same issues).

alexeldeib: oh, I noticed we do the chown in setupCNIDirs, it just happens before we untar.

alexeldeib:
passed manual re-run after screwing up e2e scenarios: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=53846536&view=results

e2e/e2e-scenario.sh
```diff
@@ -168,6 +169,7 @@ kubectl apply -f pod-nginx.yaml
 waitForPodStartTime=$(date +%s)
 for i in $(seq 1 10); do
     set +e
+    kubectl get pods -o wide | grep $podName
```
anujmaheshwari1, line 172: workaround to make e2e pass because of pods not coming up?

anujmaheshwari1 (approved):

e2e/e2e-scenario.sh
alexeldeib: to show the pod status if it's not already in running

PR #1689, opened 2022-04-04
update kube-proxy and kube-binary images with NAT hotfix

ganeshkumar5699:
**What type of PR is this?**
/kind bug

Hotfix for a kube-proxy NAT issue: https://msazure.visualstudio.com/CloudNativeCompute/_git/CloudNativeCompute.wiki/pullrequest/5604191 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/kube-proxy-images.json
```diff
@@ -18,8 +18,10 @@
           "1.22.6-hotfix.20220130",
           "1.22.6-hotfix.20220310.1",
           "1.23.3-hotfix.20220130.2",
-          "1.23.3-hotfix.20220310.2",
-          "1.23.4.1"
+          "1.23.3-hotfix.20220401.1",
+          "1.23.4.1",
```
ganeshkumar5699, line 22: Did not remove older version, in case the hotfix version has an issue detected in RP and we need to use the previous version (following the VHD release for this week).

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -464,8 +464,9 @@ MULTI_ARCH_KUBE_BINARY_VERSIONS="
 1.21.9-hotfix.20220204
 1.22.4-hotfix.20220201
 1.22.6-hotfix.20220130
-1.23.3-hotfix.20220130
-1.23.4
+1.23.3-hotfix.20220401.1
```
anujmaheshwari1, line 467: will this require a change for the post VHD tests as well?

ganeshkumar5699: yes, it will.

anujmaheshwari1: i meant this file right here https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/test/linux-vhd-content-test.sh

ganeshkumar5699: nice catch, thanks! I've updated the test file.

PR #1704, opened 2022-04-08
Add k8s 1.24.0-beta.0 to VHD

ganeshkumar5699:
**What type of PR is this?**

To test k8s 1.24 on AKS. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -468,6 +468,7 @@ MULTI_ARCH_KUBE_BINARY_VERSIONS="
 1.23.3-hotfix.20220401
 1.23.4-hotfix.20220331
 1.23.5-hotfix.20220331
+1.24.0-beta.0
```
ganeshkumar5699, line 471: Seems like this needs a beta**.0**, which should not be needed for other k8s binaries. Tested with URL.

ganeshkumar5699:
- Windows binary for this 1.24 beta version does not seem to be available but that's ok for now since we are doing this to test.
- All builds pass. Will update the branch after a review - kept it so build successes are easily visible.

anujmaheshwari1 (approved):

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: cc: @AbelHu

PR #1718, opened 2022-04-14
Ensure a reasonable open file descriptor limit

clayvan:
**What type of PR is this?**
Over the past few weeks we've noticed some of our workloads running on AKS throwing "Too many open files" in the logs, causing some performance degradation. We identified that our linux OS sysctl setting for "fs.file-max" was set to a seemingly low number of `65,536`. Note that this was in our "user" node pool. We found however that our "system" node pool, the limit was set to `1,428,951` (more in line with what we would expect).

From what I can tell, the only difference in the pools is the nodeImageVersion property:
- System node pool: `AKSUbuntu-1804containerd-2022.01.24`
- User node pool: `AKSUbuntu-1804gen2containerd-2022.01.24`

I believe that there's some underlying difference in the ubuntu OS on the 2 images, and to be safe we should set a reasonable limit like `fs.file-max=1428951`.

Note: this will still allow users to override this value via the [custom node configuration API](https://docs.microsoft.com/en-us/azure/aks/custom-node-configuration#file-handle-limits)

/kind bug

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib:
thanks! could you run `make` to regenerate some of the embedded artifacts?

alexeldeib:
based on the discussion in slack I think we're leaning away from merging this, as it could _lower_ the defaults for larger VM sizes. 

still unclear why some VM images from january booted up with wrong defaults. I suspect syslogs/dmesg from a repro would tell us something interesting, but none of us can seem to get a fresh repro. unfortunately probably going to chalk that up as a flake for now

clayvan:
Agreed. [Slack thread for future travelers](https://kubernetes.slack.com/archives/C5HJXTT9Q/p1649966267538419)

PR #1720, opened 2022-04-15
Add a check for customized image

MirzaSikander:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/bakerapi.go
+++ b/pkg/agent/bakerapi.go
@@ -29,12 +29,16 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context,
 		CSE:        templateGenerator.GetNodeBootstrappingCmd(config),
 	}
 
+	distro := config.AgentPoolProfile.Distro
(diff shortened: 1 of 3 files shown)


pkg/agent/bakerapi.go
```diff
@@ -29,12 +29,16 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context,
 		CSE:        templateGenerator.GetNodeBootstrappingCmd(config),
 	}
 
+	distro := config.AgentPoolProfile.Distro
+	if distro == datamodel.CustomizedWindowsOSImage || distro == datamodel.CustomizedImage {
```
alexeldeib, line 33: where is the image coming from in this case? RP sets it from the headers correctly?

MirzaSikander: yes. Before the changes made to use `bakerapi.go`, the behavior was just to use agentbaker to generate CSE and cmd. The logic in vmss.go figured out the right image to use.

alexeldeib (approved):

PR #1736, opened 2022-04-21

HDYA:


andyliuliming (approved):

vhdbuilder/packer/components.json
```diff
@@ -11,6 +11,9 @@
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/dashboard:*",
       "amd64OnlyVersions": [
         "v2.0.0",
+        "v2.0.0",
```
andyliuliming, line 14: why multiple entries

andyliuliming (changes requested):

vhdbuilder/packer/components.json
HDYA: it appears to have issue handling versions starting with v
will open work item to track its fix

andyliuliming (approved):

PR #1742, opened 2022-04-22
fix: prepull images for mariner, fix watcher image

alexeldeib:
**What type of PR is this?**
/kind bug

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -213,9 +213,6 @@ cat << EOF >> ${VHD_LOGS_FILEPATH}
   - libbcc-examples
 EOF
 
-installImg
```
alexeldeib, line 216: this func is undefined 😞

PR #1755, opened 2022-04-26
 azure npm - remove unused versions

ganeshkumar5699:
**What type of PR is this?**
/kind cleanup 

Removing azure npm versions that are not used in RP, for space reduction in VHD (we are hitting max space thresholds)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -90,8 +90,6 @@
     {
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-npm:*",
       "amd64OnlyVersions": [
-        "v1.4.9",
-        "v1.4.18"
       ],
       "multiArchVersions": []
     },


vhdbuilder/packer/components.json
```diff
@@ -90,8 +90,6 @@
     {
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-npm:*",
       "amd64OnlyVersions": [
-        "v1.4.9",
```
huntergregory, line 93: we should delete these and replace with 1.4.21 and 1.4.22, the current images for default and default in sov cloud respectively

huntergregory (changes requested):

vhdbuilder/packer/components.json
ganeshkumar5699: Thanks for sharing this. Does it need RP changes for the VHD image preload check as well? Who will be making those? I don't have much context about azure npm - making these changes mainly to reduce space for the VHD (so we can create a successful build for the weekly release)

huntergregory: Discussed offline that we can go ahead with this PR and include default NPM images in another PR

huntergregory (approved):

PR #1771, opened 2022-04-29
chore: remove list permission and redact token in logs

AbelHu:
**What type of PR is this?**
/kind: build

1. remove list permission
2. redact token in logs

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/generate-vhd-publishing-info.sh
+++ b/vhdbuilder/packer/generate-vhd-publishing-info.sh
@@ -60,4 +60,4 @@ cat <<EOF > vhd-publishing-info.json
 EOF
 
 # Do not log sas token
-cat vhd-publishing-info.json | sed 's/?.*\",/?***\",/g'
+sed 's/?.*\",/?***\",/g' < vhd-publishing-info.json
--- a/vhdbuilder/packer/generate-vhd-publishing-info.sh
+++ b/vhdbuilder/packer/generate-vhd-publishing-info.sh
@@ -35,14 +35,15 @@ fi
 
 start_date=$(date +"%Y-%m-%dT00:00Z" -d "-1 day")
 expiry_date=$(date +"%Y-%m-%dT00:00Z" -d "+1 year")
-sas_token=$(az storage container generate-sas --name vhds --permissions lr --connection-string ${CLASSIC_SA_CONNECTION_STRING} --start ${start_date} --expiry ${expiry_date} | tr -d '"')
+sas_token=$(az storage container generate-sas --name vhds --permissions r --connection-string ${CLASSIC_SA_CONNECTION_STRING} --start ${start_date} --expiry ${expiry_date} | tr -d '"')
 if [ "$sas_token" == "" ]; then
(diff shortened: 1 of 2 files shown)


AbelHu:
cc @alexeldeib @ganeshkumar5699 `vhdbuilder/packer/generate-vhd-publishing-info.sh` is used by both Windows and Linux so please help to take a look at it.

ShiqianTao (approved):

ganeshkumar5699:
Just to confirm, was this tested with a build to make sure the change works as expected?

ganeshkumar5699:
We used to copy this URL for PIR publishing and after we stopped publishing to PIR, I don't think this URL has been useful. Perhaps it may be useful if we are asked to publish to PIR again (doubtful) or for debugging (image replication issues can usually be debugged through the portal so it should be ok), but it seems unlikely. 

Thanks for noticing this. Lgtm.

ganeshkumar5699:
just icymi, found these logs with suggestions in `check-generated`from shell-shell (ex: https://github.com/koalaman/shellcheck/wiki/SC2001): 

> redact_vhd_url=$(echo $vhd_url |sed -e 's/?.*/?***/')
>                  ^-- SC2001: See if you can use ${variable//search/replace} instead.
> 
> 
> make: *** [Makefile:93: generate] Error 2
> In ./vhdbuilder/packer/generate-vhd-publishing-info.sh line 63:
> cat vhd-publishing-info.json | sed 's/?.*\",/?***\",/g'
>     ^----------------------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
> 
>

ganeshkumar5699 (approved):
Lgtm besides the couple of small comments above about testing + check-generated suggestion. Thank you!

AbelHu:
> ments above about testing + check-generated suggestion.

I have manually tested redact commands. I do not validate the SAS token with removing the SAS token since it needs to publish a new SIG image. We can revert this PR if we hit any issue. That is why I need to involve you :), @ganeshkumar5699.

AbelHu:
Thanks for the reminder, @ganeshkumar5699 . I have updated it in https://github.com/Azure/AgentBaker/pull/1771/commits/1ba18fd51e7b9e6872387f03587974eecc1fc71c


> just icymi, found these logs with suggestions in `check-generated`from shell-shell (ex: https://github.com/koalaman/shellcheck/wiki/SC2001):
> 
> > redact_vhd_url=$(echo $vhd_url |sed -e 's/?._/?_**/')
> > ^-- SC2001: See if you can use ${variable//search/replace} instead.
> > make: *** [Makefile:93: generate] Error 2
> > In ./vhdbuilder/packer/generate-vhd-publishing-info.sh line 63:
> > cat vhd-publishing-info.json | sed 's/?._",/?_**",/g'
> > ^----------------------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.

PR #1781, opened 2022-05-04
Fix DNS for 20.04, disable systemd resolv

anujmaheshwari1:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -239,18 +239,14 @@ configureCNIIPTables() {
 disable1804SystemdResolved() {
     ls -ltr /etc/resolv.conf
     cat /etc/resolv.conf
-    {{- if Disable1804SystemdResolved}}
     UBUNTU_RELEASE=$(lsb_release -r -s)
-    if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
+    if [[ ${UBUNTU_RELEASE} == "18.04" || ${UBUNTU_RELEASE} == "20.04" ]]; then
```
alexeldeib, line 243: != 16.04 would probably be better, as we support nothin < 16

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -54,7 +54,7 @@ fi
 configureHTTPProxyCA
 {{- end}}
 
-disable1804SystemdResolved
+disable1804and2004SystemdResolved
```
alexeldeib, line 57: just -- disableSystemdResolved?

alexeldeib (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
anujmaheshwari1: right, however, I had a problem with Mariner for != 16.04 when I was updating chrony, I will merge this now(to start the weekly release) and revisit later in case this grows. hope that is ok.

PR #1783, opened 2022-05-05

HDYA:


vhdbuilder/packer/components.json
```diff
@@ -11,6 +11,9 @@
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/dashboard:*",
       "amd64OnlyVersions": [
         "v2.0.0",
+        "v2.0.0",
```
andyliuliming, line 14: why dup entries.

vhdbuilder/packer/components.json
```diff
@@ -122,15 +125,18 @@
       ],
       "multiArchVersions": [
         "v1.0.0.2",
-        "v1.0.0.3"
+        "v1.0.0.3",
+        "v1.1.1.3",
```
andyliuliming, line 129: curious why v1.1.1.3 and v1.1.1.4 are both in use.

andyliuliming (changes requested):

vhdbuilder/packer/components.json
HDYA: there seem to be some issue in dedup process when there is a leading `v` in image tag
fixed this PR with a new commit for now

vhdbuilder/packer/components.json
HDYA: I think v1.1.1.3 is in use but not preloaded,
while v1.1.1.4 is latest patch version

andyliuliming (approved):

PR #1798, opened 2022-05-10
build: cleanup storage account

wanqingfu:
**What type of PR is this?**
/kind build

Automatically clean temporary storage accounts created 24 * 7 hours ago.
Set default value of VHD_DEBUG and DRY_RUN to False. Fixed typos.

[Work Item] https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/14246382/
The default value of VHD_DEBUG and DRY_RUN are false now. For consistency, they should be changed to False for Boolean comparisons. Currently ${VHD_DEBUG} is used in .builder-release-template-windows.yaml. Since it comes from an environment variable set in .vsts-builder-windows.yaml, this PR replaces parameter quote ${} with variable quote $().

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/cleanup.sh
+++ b/vhdbuilder/packer/cleanup.sh
@@ -98,13 +98,13 @@ echo "Looking for storage accounts in ${AZURE_RESOURCE_GROUP_NAME} created over 
 echo "That is, those created before $(date -d@$deadline) As shown below"
 az storage account list -g ${AZURE_RESOURCE_GROUP_NAME} | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | tr -d '\"' || ""
 for storage_account in $(az storage account list -g ${AZURE_RESOURCE_GROUP_NAME} | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | tr -d '\"' || ""); do
-    if [[ "${DRY_RUN}" = false ]]; then
+    if [[ "${DRY_RUN}" == "False" ]]; then
        if [[ $storage_account = aksimages* ]]; then
           echo "Will delete storage account ${storage_account}# from resource group ${AZURE_RESOURCE_GROUP_NAME}..."
           az storage account delete --name ${storage_account} -g ${AZURE_RESOURCE_GROUP_NAME} --yes  || echo "unable to delete storage account ${storage_account}, will continue..."
           echo "Deletion completed"
         fi
     else
-        echo "skipping because DRY_RUN is set to true"
+        echo "skipping because DRY_RUN is set to True"
     fi
 done
--- a/vhdbuilder/packer/cleanup.sh
+++ b/vhdbuilder/packer/cleanup.sh
@@ -38,34 +38,6 @@ if [ -n "${VNET_RESOURCE_GROUP_NAME}" ]; then
   fi
 fi
 
-#clean up the temporary storage account
-id=$(az storage account show -n ${SA_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id)
-if [ -n "$id" ]; then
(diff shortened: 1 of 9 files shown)


AbelHu:
I suggest to use `ci` instead of `experimenting` in your commit message.

vhdbuilder/packer/cleanup.sh
```diff
@@ -44,6 +44,28 @@ if [ -n "$id" ]; then
   az storage account delete -n ${SA_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} --yes
 fi
 
+#clean up storage account created over X hours ago (testing; X=19)
+# set subscription to AKS VHD Image Release
+az account set -s a15c116e-99e3-4c59-aebc-8f864929b4a0
```
AbelHu, line 49: I think that we do not need to set account. If we must set it, it should also impact all old commands after this line.

vhdbuilder/packer/cleanup.sh
```diff
@@ -44,6 +44,28 @@ if [ -n "$id" ]; then
...
+AZURE_RESOURCE_GROUP_NAME="akswinvhdbuilderrg"
```
AbelHu, line 50: Please use `AZURE_RESOURCE_GROUP_NAME` instead of hardcoded value.

vhdbuilder/packer/cleanup.sh
```diff
@@ -44,6 +44,28 @@ if [ -n "$id" ]; then
...
+EXPIRATION_IN_HOURS=20
```
AbelHu, line 51: The setting does not match the value in PR's title and description (18)

vhdbuilder/packer/cleanup.sh
```diff
@@ -44,6 +44,28 @@ if [ -n "$id" ]; then
...
+# deadline = the "date +%s" representation of the oldest age we're willing to keep
+(( deadline=$(date +%s)-${expirationInSecs%.*} ))
+echo "Current time is $(date)"
+echo "Looking for storage accounts in ${AZURE_RESOURCE_GROUP_NAME} created over ${EXPIRATION_IN_HOURS} hours ago..."
+echo "That is, those created before $(date -d@$deadline) As shown below"
+az storage account list -g ${AZURE_RESOURCE_GROUP_NAME} | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | tr -d '\"' || ""
+for storage_account in $(az storage account list -g ${AZURE_RESOURCE_GROUP_NAME} | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | tr -d '\"' || ""); do
+    if [[ $storage_account = aksimages* ]]; then
+       echo "Will delete storage account ${storage_account}
+       # from resource group ${AZURE_RESOURCE_GROUP_NAME}..."
```
AbelHu, line 63: You can move the comment into above logs

vhdbuilder/packer/cleanup.sh
wanqingfu: > I think that we do not need to set account. If we must set it, it should also impact all old commands after this line.

Yes. It was set in Line 7, init-variables.sh. Checked the log of "Build Windows VHD" step. It was already set to this subscription.

vhdbuilder/packer/cleanup.sh
wanqingfu: Sorry about the confusion. It was meticulously set to delete accounts manually created X hours ago as a trial run. The time had to be precise, and so the inconsistency comes from the time delay. Will change to 24*7. Or better yet, if there is no gap in completely cleaning the storage account upon cancellation, the manual deletion is not necessary.

AbelHu (approved):

ShiqianTao (approved):

PR #1801, opened 2022-05-10
Add new versions of CoreDNS

AYM1607:
**What type of PR is this?**

We need to add new patch versions to be used in RP as well as add the newest version to be released with k8s 1.24 preview.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/components.json
```diff
@@ -75,8 +75,7 @@
         "1.6.6"
       ],
       "multiArchVersions": [
-        "1.8.6",
-        "v1.8.6"
```
robbiezhang, line 79: don't remove v1.8.6 until we fully rollout v1.8.7

AYM1607: done

PR #1808, opened 2022-05-11
image list output for builds

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


chengliangli0918 (approved):

alexeldeib:
repoDigests seems empty for containerd even though locally testing the same images it works. not sure it matters since we still get image ID (although there are interesting quirks - https://github.com/distribution/distribution/issues/1662#issuecomment-213079540)

also this defaults to installing crictl latest version, should probably tweak that a bit more or use ctr instead

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -119,7 +119,6 @@ installCrictl() {
         tar zxvf "$CRICTL_DOWNLOAD_DIR/${CRICTL_TGZ_TEMP}" -C ${CRICTL_BIN_DIR}
         chmod 755 $CRICTL_BIN_DIR/crictl
     fi
-    rm -rf ${CRICTL_DOWNLOAD_DIR}
```
alexeldeib, line 122: why i do this?

ganeshkumar5699:
> repoDigests seems empty for containerd even though locally testing the same images it works. not sure it matters since we still get image ID (although there are interesting quirks - [distribution/distribution#1662 (comment)](https://github.com/distribution/distribution/issues/1662#issuecomment-213079540))
> 
> also this defaults to installing crictl latest version, should probably tweak that a bit more or use ctr instead

Should ctr also be used for generating this list (and ctr's list seems to be a superset of the crictl list?) ? @sozercan had shared earlier about finding images through `ctr -n k8s.io images list` and also identified some SHAs in the image list (on the node) whose component was not clear.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -153,6 +153,9 @@ if [[ ${CONTAINER_RUNTIME:-""} == "containerd" ]]; then
     downloadCrictl ${CRICTL_VERSION}
     echo "  - crictl version ${CRICTL_VERSION}" >> ${VHD_LOGS_FILEPATH}
   done
+  
+  KUBERNETES_VERSION=$(echo -e "$CRICTL_VERSIONS" | head -n 1) installCrictl || exit $ERR_CRICTL_DOWNLOAD_TIMEOUT
```
ganeshkumar5699, line 157: Is the output stored in KUBERNETES_VERSION, being used elsewhere?

ganeshkumar5699, line 157: Do we need to handle separate installation of `crictl` for docker vs containerd builds here?

alexeldeib:
ah, need to recreate this pr from not-my-fork for security on PR checks

vhdbuilder/packer/install-dependencies.sh
alexeldeib: in installCrictl :)

alexeldeib: shouldn't really matter since docker will use...docker I suppose

alexeldeib:
moved to https://github.com/Azure/AgentBaker/pull/1985

PR #1836, opened 2022-05-20
feat: custom configuration support for Windows kubelet and kubeproxy

mainred:
**What type of PR is this?**
/kind feature

Support customized configuration for Windows kubelet and kubeproxy. 
WindowsKubernetesConfigurations provides kubelet and kubeproxy args so that customers can providers their own configuration for these two components. The args provided by customers will override existing default ones.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/types_test.go
```diff
@@ -2039,15 +2039,18 @@ func TestKubernetesConfigGetOrderedKubeletConfigString(t *testing.T) {
 		expectedForPowershell string
 	}{
 		{
-			name:                  "zero value kubernetesConfig",
-			config:                &NodeBootstrappingConfiguration{},
+			name: "zero value kubernetesConfig",
```
AbelHu, line 2042: Should we keep the old unit test even if it is a negative test case that ContainerService is nil? We may also need another negative test case that Properties is nil?

mainred: I checked these are several places assuming that `NodeBootstrappingConfiguration.ContainerService.Properties` is not nil, so I did not check the nil pointer in the newly introduced code.

AbelHu: The old test case covers that ContainerService is nil but your PR removed this test case since you updated the existing unit test. You can keep it and add your new unit test

mainred: Reverted the change on the old tests.

AbelHu (approved):
Please select the first checkbox `uses [conventional commit messages](https://www.conventionalcommits.org/)` in the PR

PR #1840, opened 2022-05-22
Cache crictl 1.24.0 in VHD

ganeshkumar5699:
**What type of PR is this?**
/kind chore

To use with k8s 1.24.0 preview

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -142,11 +142,10 @@ if [[ ${CONTAINER_RUNTIME:-""} == "containerd" ]]; then
   echo "  - [installed] containerd v${containerd_version}-${containerd_patch_version}" >> ${VHD_LOGS_FILEPATH}
 
   CRICTL_VERSIONS="
```
alexeldeib, line 144: guess it would be nice to eventually switch to use the json too heh. only in the test rn I think?

ganeshkumar5699: Yes, it would be good to change here too

PR #1841, opened 2022-05-23
Remove dockershim flags for containerd and --non-masquerade-cidr for all k8s versions

ganeshkumar5699:
**What type of PR is this?**
/kind cleanup

Removing dockershim related flags for kubelet (and also the --non-masquerade-cidr) flag. 
These flags have been removed with k8s 1.24. While RP passes them in, HCP which stores these flags may pass them in. Deleting for containerd images and this would affect both Linux and WIndows nodes.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):

pkg/agent/variables.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/blang/semver"
 )
 
+var dockerShimFlags = []string{"--image-pull-progress-deadline", "--network-plugin", "--cni-bin-dir", "--cni-conf-dir", "--docker-endpoint", "--network-plugin", "--network-plugin-mtu"}
```
AbelHu, line 14: Should we also need to add `--cni-cache-dir` in this list?

e2e/nodebootstrapping_template.json
```diff
@@ -166,7 +166,6 @@
         "--max-pods": "110",
         "--network-plugin": "kubenet",
         "--node-status-update-frequency": "10s",
-        "--non-masquerade-cidr": "0.0.0.0/0",
```
AbelHu, line 169: Thanks for updating this. I think that Windows should not use this flag since Windows uses ExceptionList. Please run all Windows test pipelines with this change to validate it.
<html>
<body>


--non-masquerade-cidr string     Default: 10.0.0.0/8
--
  | Traffic to IPs outside this range will use IP masquerade. Set to '0.0.0.0/0' to never masquerade. (DEPRECATED: will be removed in a future version)


</body>
</html>

pkg/agent/baker.go
```diff
@@ -260,10 +260,16 @@ func normalizeResourceGroupNameForLabel(resourceGroupName string) string {
 func validateAndSetLinuxNodeBootstrappingConfiguration(config *datamodel.NodeBootstrappingConfiguration) {
 	// If using kubelet config file, disable DynamicKubeletConfig feature gate and remove dynamic-config-dir
 	// we should only allow users to configure from API (20201101 and later)
+	profile := config.AgentPoolProfile
 	if config.KubeletConfig != nil {
 		kubeletFlags := config.KubeletConfig
 		delete(kubeletFlags, "--dynamic-config-dir")
-		delete(kubeletFlags, "--image-pull-progress-deadline")
+		delete(kubeletFlags, "--non-masquerade-cidr")
```
AbelHu, line 267: Does 1.24 not support this flag --non-masquerade-cidr? Should we also remove this for Windows?

ganeshkumar5699: Yes, k8s 1.24 does not support --non-masquerade-cidr flag

ganeshkumar5699: It was removed in this PR: https://github.com/kubernetes/kubernetes/pull/107096

ganeshkumar5699: It does seem like it can be removed for Windows too

pkg/agent/variables.go
ganeshkumar5699: Thanks! I've added it to the list. `--cni-cache-dir` is not present in RP/AB at the moment, so it should be good.

e2e/nodebootstrapping_template.json
ganeshkumar5699: Should I also test it with the Windows VHD pipeline in this PR? Or just the Windows RP pipelines? Asking since the Windows pipelines are skipped for this PR.

AbelHu (approved):

e2e/nodebootstrapping_template.json
ganeshkumar5699: Vendored + tested in the AKS Windows E2E RP pipelines

PR #1862, opened 2022-06-04
Add DisableAcceleratorUsageMetrics feature gate with default value: false for accelerator usage metrics in Kubelet cAdvisor endpoint

ganga1980:
**What type of PR is this?**
feature gate 

ContainerInsights depends on GPU accelerator Usage metrics from Kubelet cAdvisor endpoint but deprecation of this feature moved to beta which breaks the ContainerInsights customers with K8s version 1.20 or higher
Until Container Insights move to new API adding this feature gate to get the GPU metrics continue to work
Reference - https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1867-disable-accelerator-usage-metrics

GPU Accelerator Usage metrics continue to available in Kubelet cAdvisor endpoint

Fixes #

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


alexeldeib:
do you know if/when this flag will be removed? I'm browsing but i'm not finding it. I'd just ask that you bound it on both min/max or else it's a pain for clusters later. we need to fix this in other places as well, but prefer doing it now if we know this will be deprecated?

ganga1980:
> do you know if/when this flag will be removed? I'm browsing but i'm not finding it. I'd just ask that you bound it on both min/max or else it's a pain for clusters later. we need to fix this in other places as well, but prefer doing it now if we know this will be deprecated?

This flag will be removed when this feature "DisableAcceleratorMetrics" goes to GA. Currently, its in Beta. The GA date will be 1 year  from beta. I cant find the exact date when this feature went to BETA. 

https://kubernetes.io/blog/2020/12/16/third-party-device-metrics-reaches-ga/#nvidia-gpu-metrics-deprecated
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features.

alexeldeib (approved):
lgtm pending one more rebase

PR #1871, opened 2022-06-08
feat: enable 2.0 Mariner images

Camelron:
**What type of PR is this?**
/kind feature

Enables Mariner 2.0 AKS image builds from Mariner 2.0 marketplace images. 

Contains a cloud-init custom data script to allow sha1 ssh auth during packer build. Cleans it up after. 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests


packer.mk
```diff
@@ -38,6 +39,18 @@ else
 	@echo "${MODE}: Building with Hyper-v generation 1 VM and save to Classic Storage Account"
 	@packer build -var-file=vhdbuilder/packer/settings.json vhdbuilder/packer/vhd-image-builder-mariner.json
 endif
+else ifeq (${OS_VERSION},V2)
```
alexeldeib, line 42: @anujmaheshwari1 I feel like we had discussed some better way to do this?

vhdbuilder/packer/sha1rsa_custom_data
```diff
@@ -0,0 +1,6 @@
+#cloud-config to reenable sha1 temporarily for packer
```
alexeldeib, line 1: this is due to issues with packer ssh failing? 

can we just use some more secure config?

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -34,6 +34,25 @@ networkdWorkaround() {
     sed -i "s/Name=e\*/Name=eth0/g" /etc/systemd/network/99-dhcp-en.network
 }
 
+# On Mariner 2.0 Marketplace images, the default systemd network config is removed.
+# Because Azure data sources is disabled in AKS, we need to readd a default configuration for eth0
+setMarinerNetworkdConfig() {
```
alexeldeib, line 39: what's the actual behavior difference between this and networkdWorkaround?

one seds the config in place, one overwrites the whole file? can we collapse?

alexeldeib: or is there some real difference in the base config?

vhdbuilder/packer/sha1rsa_custom_data
hbeberman: This workaround was added because Packer was only capable of using SHA-1 RSA due to some [longstanding limitations ](https://github.com/golang/go/issues/39885)in the underlying golang crypto lib. Because Mariner 2.0 is using a newer version of OpenSSH that has SHA-1 disabled by default, we use this customdata to temporarily reenable it for the lifetime of the Packer run.

For other clouds they have [the option ](https://github.com/hashicorp/packer-plugin-amazon/pull/179) of using ed25519 as their temporary keypair, but that's [not supported](https://docs.microsoft.com/en-us/troubleshoot/azure/virtual-machines/ed25519-ssh-keys) on Azure.

The good news is this looks like it may have been addressed in the [packer-plugin-azure v1.0.7](https://github.com/hashicorp/packer-plugin-azure/releases/tag/v1.0.7) release at the start of May. 

@Camelron can we try removing this customdata and testing a build on the latest version of the azure packer plugin?

Camelron: Testing now

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
Camelron: There is a slight difference between 1.0 and 2.0 configs in this case, but I combined them into setMarinerNetworkdConfig and branch inside of the function rather than inside the caller.

vhdbuilder/packer/sha1rsa_custom_data
Camelron: The version of Packer we are using doesn't pull in the latest packer-plugin-azure. The earliest version of packer that does so is v1.8.1 (https://github.com/hashicorp/packer/commit/9bf17ec95292f5729510cdce02828e58287f63fe). Added a todo comment to this workaround so we can remove it when we are able to.

alexeldeib (approved):
lgtm, one comment for @anujmaheshwari1 but think you're fine to merge 👍

hbeberman (approved):

PR #1874, opened 2022-06-09

HDYA:


vhdbuilder/packer/components.json
```diff
@@ -11,6 +11,9 @@
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/dashboard:*",
       "amd64OnlyVersions": [
         "v2.0.0",
+        "v2.0.0",
```
andyliuliming, line 14: why always have dup entries.

vhdbuilder/packer/components.json
```diff
@@ -293,7 +299,8 @@
       ],
       "multiArchVersions": [
         "v1.17.0",
-        "v1.18.0"
+        "v1.18.0",
```
andyliuliming, line 302: why keep 2 version.

vhdbuilder/packer/components.json
```diff
@@ -124,14 +127,17 @@
       ],
       "multiArchVersions": [
         "v1.0.0.2",
-        "v1.0.0.3"
+        "v1.0.0.3",
```
andyliuliming, line 130: is it expected to have 4 versions?

andyliuliming (changes requested):

vhdbuilder/packer/components.json
HDYA: fixed

vhdbuilder/packer/components.json
HDYA: 1.0 and 1.1 are considered different minor versions, should be by design

vhdbuilder/packer/components.json
HDYA: I don't think auto patch will remove a minor version, shall we delete it manually?

HDYA: I think it should be owner's call to remove another minor version from VHD, when they did a bump manually

andyliuliming (approved):

PR #1876, opened 2022-06-09
update blobfuse driver to v1.4.4

andyzhangx:
**What type of PR is this?**
/kind feature

blobfuse v1.4.4 is a stable version released 6/1/2022

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
update blobfuse driver to v1.4.4
```


pkg/templates/templates_generated.go
```diff
@@ -4318,7 +4318,7 @@ installDeps() {
     aptmarkWALinuxAgent hold
     apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
     apt_get_dist_upgrade || exit $ERR_APT_DIST_UPGRADE_TIMEOUT
-    BLOBFUSE_VERSION="1.4.3"
+    BLOBFUSE_VERSION="1.4.4"
```
mainred, line 4321: 👍 
`installDeps` is shared by both vhd and cse, so modification here will work for both vhd cache and cse.

mainred: I think we need to change https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/install-dependencies.sh#L51 by removing the specific version, this is used for vhd release notes, and the real installed version can be found later in the release notes, for example, https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/release-notes/AKSUbuntu/gen2/1804/2022.06.08.txt#L246

PR #1878, opened 2022-06-10
fix: curl mcr through proxy in cse

alexeldeib:
**What type of PR is this?**
/kind bug

In egress restricted environments with a proxy, the curl check still fails since it doesn't actually go through the proxy without other networking config. 

Not sure how this ever worked for users -- maybe some routing configuration was different?

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


phealy (approved):
lgtm

PR #1902, opened 2022-06-17
feat: upgrade nvidia driver to 510 for A10 support

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib:
oh good. we've finally hit a point where Azure supports GPUs which the newer nvidia drivers don't support. 

the VM size used in E2E uses an older GPU which is no longer supported in the 510 driver series. it fails to install the drivers. 

A10 GPUs _require_ the 510 series. 

this may complicate things. we may need to start differentiating them

PR #1912, opened 2022-06-21
adding mandb suppression logic.

cameronmeissner:
**What type of PR is this?**

/kind bug

This PR will bring in changes to CSE installation steps to suppress real-time man-db updates when installing dependencies. After CSE installation completes, a background process is spawned that will trigger man-db to update its indices with newly installed dependencies without delaying termination of the CSE installation process.

Fixes #
Currently, when dependencies are installed during CSE execution time, automatic updates are triggered on man-db. This massively increases the time it takes to install each dependency during CSE execution, as it can take man-db a lot of time to update its database for each downloaded dependency.

This PR will prevent man-db updates from slowing down CSE execution by suppressing man-db's auto-update feature and having it update its indices AFTER all dependencies have been installed. This update process will now take place within a background process that does not block CSE execution.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -556,3 +559,6 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
   # multi-user.target usually start at the end of the boot sequence
   sed -i 's/After=network-online.target/After=multi-user.target/g' /lib/systemd/system/motd-news.service
 fi
+
+echo "Recreating man-db auto-update flag file and kicking off man-db update process.."
+/usr/bin/mandb &
```
alexeldeib, line 564: this is in vhd, doesn't need to be backgrounded

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -15,6 +15,19 @@ TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
...
+MAN_DB_AUTO_UPDATE_FLAG_FILEPATH="/var/lib/man-db/auto-update"
+
+removeManDbAutoUpdateFlagFile() {
+    if [ -f "$MAN_DB_AUTO_UPDATE_FLAG_FILEPATH" ]; then
+        rm $MAN_DB_AUTO_UPDATE_FLAG_FILEPATH
+    fi
+}
+
+createManDbAutoUpdateFlagFile() {
+    if [! -f "$MAN_DB_AUTO_UPDATE_FLAG_FILEPATH" ]; then
```
alexeldeib, line 27: need a space after [ and before !

ganeshkumar5699 (commented):
It seems like you tested it with the timestamps and I wonder: were you able to measure the time it took to execute the script (until where you added the timestamp), and see any noticeable reduction in execution time?

cameronmeissner:
@ganeshkumar5699 I didn't end up testing with timestamps across the whole CSE execution, but I did add some adhoc timestamping specifically for the man-db update process after all dependencies have been installed. At the moment it looks like this process takes about ~2 seconds on vanilla and ~5 seconds on gpu.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -15,6 +15,19 @@ TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
...
+
+removeManDbAutoUpdateFlagFile() {
+    if [ -f "$MAN_DB_AUTO_UPDATE_FLAG_FILEPATH" ]; then
+        rm $MAN_DB_AUTO_UPDATE_FLAG_FILEPATH
+    fi
+}
+
+createManDbAutoUpdateFlagFile() {
+    if [ ! -f "$MAN_DB_AUTO_UPDATE_FLAG_FILEPATH" ]; then
+        touch $MAN_DB_AUTO_UPDATE_FLAG_FILEPATH
```
phealy, line 28:
```suggestion
        touch $MAN_DB_AUTO_UPDATE_FLAG_FILEPATH
```

touch won't error if the file already exists, so it's up to you, but we don't really need to have that inside an if check at all.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -15,6 +15,19 @@ TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
 KRUSTLET_VERSION="v0.0.1"
 MANIFEST_FILEPATH="/opt/azure/manifest.json"
+MAN_DB_AUTO_UPDATE_FLAG_FILEPATH="/var/lib/man-db/auto-update"
+
+removeManDbAutoUpdateFlagFile() {
+    if [ -f "$MAN_DB_AUTO_UPDATE_FLAG_FILEPATH" ]; then
+        rm $MAN_DB_AUTO_UPDATE_FLAG_FILEPATH
```
phealy, line 22: if you make this `rm -f` it won't error if the file doesn't exist, so you can skip the check here too.

parts/linux/cloud-init/artifacts/cse_install.sh
phealy, line 28: touch will just update the mtime if the file already exists, so you don't really need an if check.

cameronmeissner: It seems you've uncovered the fact that my bash/linux is a bit rusty ;).

ganeshkumar5699 (approved):
Lgtm! Congrats on the first PR!

anujmaheshwari1 (approved):

PR #1915, opened 2022-06-22
Add CSE logging upload functionality

phealy:
**What type of PR is this?**
/kind feature

This PR symlinks some log files into a subdirectory of /var/log/azure (/var/log/azure/aks), where the WALinuxAgent will pick them up and upload them to the host via WireServer. This will include these log files on an hourly upload. Additionally, a timeout has been added to provision.sh (set to 14m, one minute shorter than the CSE execution timeout), after which it will be killed and the logs will be uploaded to the host before the script terminates with the regular error conditions. This means that the provisioning logs for any failing Linux node can now be downloaded from the host.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


phealy:
Opened against the wrong branch.

PR #1917, opened 2022-06-22
Custom CA - handle dirs not existing in update script

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


anujmaheshwari1 (approved):

parts/linux/cloud-init/artifacts/update_certs.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -uo pipefail
```
ganeshkumar5699, line 2: Just curious, what prompted this change to not exit on errors? Asking since this seems to affect all error handling and not just directory/file related errors

ganeshkumar5699: You'd mentioned about no errors when feature is not enabled -> so this checks whether feature was enabled?

UtheMan: The problem here was this line: 

`ls "$certDestination" | grep -E '^[0-9]{14}' | while read -r line; do`

with -e enabled we would get a non-zero exit code when $certDestination was empty (which is a case when feature is not enabled). 
To get around that we could either add some additional logic checking for directories or just remove -e since it's the only real error that was appearing for us here.

ganeshkumar5699: Got it, thanks for explaining!

PR #1926, opened 2022-06-23
fix: move proxy CA configuration before egress check

alexeldeib:
**What type of PR is this?**

egress with a custom ca with https proxy fails ssl validation to MCR if the cert isn't configured with TLS first. it goes through the proxy now, but fails -_-

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


phealy (approved):
lgtm!

PR #1927, opened 2022-06-24
Upload AKS logs to host on CSE failure/hourly basis

phealy:
**What type of PR is this?**
/kind feature

This PR symlinks some log files into a subdirectory of /var/log/azure (/var/log/azure/aks), where the WALinuxAgent will pick them up and upload them to the host via WireServer. This will include these log files [in the Azure Linux Agent automatic hourly log collection](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#linux-guest-agent-automatic-logs-collection). Additionally, a timeout has been added to provision.sh (set to 14m, one minute shorter than the CSE execution timeout), after which it will be killed and the logs will be uploaded to the host before the script terminates with the regular error conditions. This means that the provisioning logs for any failing Linux node can now be downloaded from the host.

These logs can be downloaded by a support engineer during a support case where the customer gives log access permissions; see the GuestVMLogs action in ASI for more details.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
AKS node provisioning logs are now included in the [Azure Linux VM Agent automatic hourly log collection](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#linux-guest-agent-automatic-logs-collection) for faster log access during support cases.


phealy:
@cameronmeissner PR for you!

cameronmeissner:
NOTE: the failing LGTM Python analysis is occurring because there are currently no Python files within the master branch of AB. Once this PR is merged into master, that check should no longer fail.

ganeshkumar5699:
This should be a pretty useful feature!

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -105,6 +105,9 @@ stages:
             echo '##vso[task.setvariable variable=DRY_RUN]${{parameters.dryrun}}'
             echo '##vso[task.setvariable variable=OS_SKU]Ubuntu'
             echo '##vso[task.setvariable variable=OS_VERSION]16.04'
+            echo '##vso[task.setvariable variable=IMG_PUBLISHER]Canonical'
+            echo '##vso[task.setvariable variable=IMG_OFFER]UbuntuServer'
+            echo '##vso[task.setvariable variable=IMG_SKU]16.04-LTS'
```
ganeshkumar5699, line 110: Just curious, why is 16.04 being added here, even though we don't support it now?

ganeshkumar5699 (approved):
PR lgtm! While I skimmed through most sections, I haven't read every section in detail. Since you both have tested it extensively, including complete runs to test the feature, it seems good. 

cc: @alexeldeib may have a more comprehensive review

parts/windows/sendlogs.ps1
```diff
@@ -0,0 +1,38 @@
+<#
```
alexeldeib, line 1: is this used anywhere? maybe break windows to a separate pr?

parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
```diff
@@ -0,0 +1,52 @@
+#! /bin/bash
+
+SRC=/var/log/containers
+DST=/var/log/azure/aks/pods
+
+# Install inotify-tools if they're missing from the image
+command -v inotifywait >/dev/null 2>&1 || apt-get -o DPkg::Lock::Timeout=300 -y install inotify-tools
```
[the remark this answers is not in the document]
alexeldeib: cc @hbeberman, might be interesting for you

.pipelines/.vsts-vhd-builder-release.yaml
phealy: We can revert that - I added it so that I could test more easily on a 1604 VHD, since even though it's not supported, we do still have clusters out there using it. It turns out the image build doesn't work for other reasons anyway, and I was able to test using the last built image.

parts/windows/sendlogs.ps1
phealy: It's not used right now, but including it in the build would mean that the windows team could use it manually. Instead of having to have support run collect-aks-logs.ps1 and then pull it from the host to send, they could stop the waagent and run send-logs.py to force the upload to the host where support can get it. If you want it in a separate PR we can pull it out, but I think it's useful to include even before they integrate it in their provisioning script.

PR #1936, opened 2022-06-29
build: Exclude the UDP source port 65330 in Windows VHD

AbelHu:
**What type of PR is this?**
/bug

Exclude the UDP source port 65330 in Windows VHD since it may fail if we run the same command in Windows CSE script.

Fixes #

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

I have tested it in https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=57999733&view=results. The settings is still reserved after running sysprep.

**Release note**:
```
none
```


ShiqianTao (approved):

AbelHu:
**NOTE**: Retry with 30 times may still not work. cc @ShiqianTao

ShiqianTao (approved):

PR #1956, opened 2022-07-07
Use official packer to build arm64 vhd

chengliangli0918:
**What type of PR is this?**

Packer always fails to capture generalized os disk of ARM64 VM to a managed image because arm64 compute stack doesn't support managed image anymore because managed image is in maintenance mode (https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/209282/Packer-always-fails-to-capture-generalized-os-disk-of-ARM64-VM). official packer 1.7.3 that we previously used in Agentbaker doesn't have a workaround for this issue so I have to use a private packer 1.7.x build to build arm64. Now the official 1.8.1 Packer provides a 'skip_create_image' to workaround this arm64 compute restriction, so switch to official packer to build arm64 vhd.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


packer.mk
```diff
@@ -5,10 +5,7 @@ ifeq (${OS_SKU},Ubuntu)
 ifeq (${ARCHITECTURE},ARM64)
 ifeq (${MODE},gen2Mode)
 	@echo "${MODE}: Building with Hyper-v generation 2 ARM64 VM"
-	@rm -f ./vhdbuilder/packer/packer
-	@curl -fsSL https://packerbinary.blob.core.windows.net/bin/packer -o ./vhdbuilder/packer/packer
```
chengliangli0918, line 9: @HDYA  JFYI. In an AKS pipeline, you also use my private packer. Now you can use packer 1.8.1+ with 'skip_create_image' setting to workaround an ARM64 stack restriction.

anujmaheshwari1 (approved):

PR #1900, opened 2022-06-17
fix: install containerd from packages.microsoft.com

alexeldeib:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #1965, opened 2022-07-13
cache containerd 1.6, add testing channel for apt, and switch to packages.microsoft.com again 

ganeshkumar5699:
**What type of PR is this?**
/kind bug

This PR is to fix 2 bugs: 
1) Containerd 1.6 not being cached in the VHD.
2) Containerd 1.6 needing additional apt config to use the "testing" channel. 

The switch to using packages.microsoft.com is from this PR: 
https://github.com/Azure/AgentBaker/pull/1900 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
This PR switches containerd downloads to use `packages.microsoft.com` with apt (instead of `moby.blob.core.windows.net`). After this VHD is released and in the corresponding RP vendor PR, it should make containerd downloads consistent with the AKS egress docs.
```


parts/linux/cloud-init/artifacts/manifest.json
```diff
@@ -4,7 +4,8 @@
         "downloadLocation": "/opt/containerd/downloads",
         "downloadURL": "https://moby.blob.core.windows.net/moby/moby-containerd/${CONTAINERD_VERSION}+azure/bionic/linux_${CPU_ARCH}/moby-containerd_${CONTAINERD_VERSION}+azure-ubuntu18.04u${CONTAINERD_PATCH_VERSION}_${CPU_ARCH}.deb",
         "versions": [
-            "1.4.13-3"
+            "1.4.13-3",
+            "1.6.4-4"
```
ganeshkumar5699, line 8: Need to add all versions besides "latest" to this list of versions, as this is the list used for downloading (the "latest" version is downloaded and installed in the VHD by default).

pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line48.sh
```diff
@@ -102,6 +102,7 @@ updateAptWithMicrosoftPkg() {
     fi
 
     retrycmd_if_failure 10 5 10 cp /tmp/microsoft-prod.list /etc/apt/sources.list.d/ || exit $ERR_MOBY_APT_LIST_TIMEOUT
+    echo "deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/18.04/multiarch/prod testing main" > /etc/apt/sources.list.d/microsoft-prod-testing.list
```
ganeshkumar5699, line 105: Added the testing channel here. Thanks @cpuguy83 for the help with this.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -162,26 +163,28 @@ installStandaloneContainerd() {
...
 
 downloadContainerdFromVersion() {
-    #containerd has arm64 binaries from 1.4.4 or later on moby.blob.core.windows.net
-    CPU_ARCH=$(getCPUArch)  #amd64 or arm64
     CONTAINERD_VERSION=$1
-    # currently upstream maintains the package on a storage endpoint rather than an actual apt repo
-    CONTAINERD_PATCH_VERSION="${2:-1}"
-    CONTAINERD_DOWNLOAD_URL="https://moby.blob.core.windows.net/moby/moby-containerd/${CONTAINERD_VERSION}+azure/bionic/linux_${CPU_ARCH}/moby-containerd_${CONTAINERD_VERSION}+azure-ubuntu18.04u${CONTAINERD_PATCH_VERSION}_${CPU_ARCH}.deb"
-    downloadContainerdFromURL $CONTAINERD_DOWNLOAD_URL
+    updateAptWithMicrosoftPkg
```
alexeldeib, line 184: is this not already done elsewhere?

alexeldeib (approved):
mostly lgtm, one minor improvement

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumar5699: It is called in two places: 
1) In`downloadContainerdFromVersion` , which is called in during VHD build time and in CSE, in `installStandaloneContainerd`. 

2)  In`installMoby`, which should be called for docker builds. 

And I just reverted the earlier revert PR for this chang.

ganeshkumar5699: So yeah, this function would be called twice, if the necessary containerd version was not cached.

ganeshkumar5699: Perhaps a better way is to move `updateAptWithMicrosoftPkg` from the `downloadContainerdFromVersion` helper function, to being called once during the VHD build. Will do that.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -102,6 +102,7 @@ updateAptWithMicrosoftPkg() {
     fi
 
     retrycmd_if_failure 10 5 10 cp /tmp/microsoft-prod.list /etc/apt/sources.list.d/ || exit $ERR_MOBY_APT_LIST_TIMEOUT
+    echo "deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/18.04/multiarch/prod testing main" > /etc/apt/sources.list.d/microsoft-prod-testing.list
```
alexeldeib, line 105: please use UBUNTU_RELEASE here for 20.04, 22.04

alexeldeib: nvm, I see it doesn't exist for 20.04 and 22.04...something to perhaps follow up on

cpuguy83: Those repos don't have "multiarch" in the url.

ganeshkumar5699: @cpuguy83, thanks for the help with this! Probably a basic question: do the containerd builds for 20.04 and 22.04 differ, from the 18.04 builds? Wondering if it's ok to use the 18.04 URL for all (or change it depending on the version).

cpuguy83: @ganeshkumar5699 Yes, the builds are different as the binaries are dynamically linked against the toolchain available for each distro.

ganeshkumar5699: Thank you. Will likely be making this change in a subsequent PR.

PR #1972, opened 2022-07-14
Upload AKS logs to host on CSE failure/hourly basis

cameronmeissner:
**What type of PR is this?**
/kind feature

This PR symlinks some log files into a subdirectory of /var/log/azure (/var/log/azure/aks), where the WALinuxAgent will pick them up and upload them to the host via WireServer. This will include these log files [in the Azure Linux Agent automatic hourly log collection](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#linux-guest-agent-automatic-logs-collection). Additionally, a timeout has been added to provision.sh (set to 14m, one minute shorter than the CSE execution timeout), after which it will be killed and the logs will be uploaded to the host before the script terminates with the regular error conditions. This means that the provisioning logs for any failing Linux node can now be downloaded from the host.

These logs can be downloaded by a support engineer during a support case where the customer gives log access permissions; see the GuestVMLogs action in ASI for more details.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
AKS node provisioning logs are now included in the [Azure Linux VM Agent automatic hourly log collection](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#linux-guest-agent-automatic-logs-collection) for faster log access during support cases.


parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
```diff
@@ -0,0 +1,56 @@
+#! /bin/bash
+
+SRC=/var/log/containers
+DST=/var/log/azure/aks/pods
+
+# Install inotify-tools if they're missing from the image
+{{- if IsMariner}}
+command -v inotifywait >/dev/null 2>&1 || dnf install -y inotify-tools
```
alexeldeib, line 8: think you still need this in artifacts/mariner/cse_install_mariner.sh to get it in the VHD? 

if possible, would be nice to use /etc/os-release and vanilla bash vars instead of templates here, but not a big deal either way (trying to reduce reliance on template vars in scripts where possible)

cameronmeissner: This check should be useful in the case where something happens to cause the CSE data to be repaved without the image being updated. @phealy 

I'll look into avoiding templating here.

alexeldeib: I think just keep the templates for now, not a big deal and simple

cameronmeissner: Just pushed changes for using /etc/os-release instead lol, can you revert if you think it's simpler to use templates - the only concern I have for sourcing from /etc/os-release is if there are already variables with the same name, they will be overwritten from what I understand. Was go and start and testing against that.

alexeldeib: haha I apologize, let's run with it, I just felt bad for nit picking :P

alexeldeib (approved):

AbelHu:
@cameronmeissner @phealy Thanks for adding parts/windows/sendlogs.ps1. May I know how you test uploading CSE logs when failing in provisioning Linux nodes? I have implemented the code for Windows but I do not know how to validate whether the cse logs is uploaded successfully. https://github.com/Azure/AgentBaker/commit/6085ee3b18fadc422be9137aab2799296dca0b7e

phealy:
> @cameronmeissner @phealy Thanks for adding parts/windows/sendlogs.ps1. May I know how you test uploading CSE logs when failing in provisioning Linux nodes? I have implemented the code for Windows but I do not know how to validate whether the cse logs is uploaded successfully. https://github.com/Azure/AgentBaker/commit/6085ee3b18fadc422be9137aab2799296dca0b7e

I'll send you a response on teams since this is a public repo.

PR #1985, opened 2022-07-20
image list output for builds

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib:
@ganeshkumar5699 

> Should ctr also be used for generating this list (and ctr's list seems to be a superset of the crictl list?)

I thought crictl is pretty much equivalent to ctr -n k8s.io? did you identify some images which are listed on way and not the other?

otherwise think this is almost there, fixing some minor issues with crictl install..

alexeldeib:
I also couldn't find a way to list images with ctr to output json :/

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
```diff
@@ -311,6 +311,23 @@
...
     },
+    {
+      "type": "file",
+      "source": "vhdbuilder/packer/list-images.sh",
+      "destination": "/home/packer/list-images.sh"
+    },
+    {
+      "type": "shell",
+      "inline": [
+        "sudo BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} HYPERV_GENERATION={{user `hyperv_generation`}} CONTAINER_RUNTIME={{user `container_runtime`}} ENABLE_FIPS={{user `enable_fips`}} /bin/bash -ux /home/packer/list-images.sh"
```
ganeshkumar5699, line 322: Just a general question (since it's the case for other parts of packer too): is there a good way to avoid repeating commands across various packer build files?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -153,6 +153,9 @@ if [[ ${CONTAINER_RUNTIME:-""} == "containerd" ]]; then
     downloadCrictl ${CRICTL_VERSION}
     echo "  - crictl version ${CRICTL_VERSION}" >> ${VHD_LOGS_FILEPATH}
   done
+  
+  KUBERNETES_VERSION=$(echo -e "$CRICTL_VERSIONS" | tail -n 2 | head -n 1 | tr -d ' ') installCrictl || exit $ERR_CRICTL_DOWNLOAD_TIMEOUT
```
ganeshkumar5699, line 157: As this uses the latest version of crictl, I was curious about the crictl version skew/compat with containerd and k8s versions and found this: https://github.com/kubernetes-sigs/cri-tools/issues/297. Seems like it should be fine.

ganeshkumar5699:
> @ganeshkumar5699
> 
> > Should ctr also be used for generating this list (and ctr's list seems to be a superset of the crictl list?)
> 
> I thought crictl is pretty much equivalent to ctr -n k8s.io? did you identify some images which are listed on way and not the other?
> 
> otherwise think this is almost there, fixing some minor issues with crictl install..

We discussed offline and the main difference is `ctr -n k8s.io images list` lists additional SHAs. @alexeldeib found that the `ctr -n k8s.io images list` has a manifest corresponding to the image tag, and that manifest has SHAs corresponding to the AMD 64 and ARM 64 image versions. And one of the SHAs (depending on the image specific to ISA), is listed by `ctr`. 

cc: @sozercan

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -123,6 +123,10 @@ stages:
           inputs:
             artifactName: 'vhd-release-notes'
             targetPath: 'release-notes.txt'
+        - task: PublishPipelineArtifact@0
```
ganeshkumar5699, line 126: nit: there are two stages with the same "PublishPipelineArtifact" name (the other for release notes). Perhaps the name can be changed?

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -123,6 +123,10 @@ stages:
...
+          inputs:
+            artifactName: 'vhd-image-bom'
+            targetPath: 'image-bom.json'
```
ganeshkumar5699, line 129: The build logs show that `image-bom.json` has been published and uploaded successfully. Is there way to view that file separately, for this build, at this stage?

ganeshkumar5699 (approved):
Lgtm. Just added some minor comments/questions.

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
alexeldeib: ideally yeah -- you can see cluster-api is a bit better at this -- you can have multiple json files for variables and merge those. the steps still need to be the same but good way to parameterize: https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/packer/azure/azure-sig.json

think @anujmaheshwari1 had some thoughts on this for the future

.pipelines/.vsts-vhd-builder.yaml
alexeldeib: you can download it from the pipeline artifacts from PR builds :)

.pipelines/.vsts-vhd-builder.yaml
alexeldeib: good point, will try a fix 👍

vhdbuilder/release-notes/autonotes/main.go
```diff
@@ -125,28 +125,44 @@ func getReleaseNotes(sku, path string, fl *flags, errc chan<- error, done chan<-
 	}
 	defer os.RemoveAll(tmpdir)
 
-	artifactName := fmt.Sprintf("vhd-release-notes-%s", sku)
-	artifactFileIn := filepath.Join(tmpdir, "release-notes.txt")
-	artifactDirOut := filepath.Join(fl.path, path)
-	artifactFileOut := filepath.Join(artifactDirOut, fmt.Sprintf("%s.txt", fl.date))
+	releaseNotesName := fmt.Sprintf("vhd-release-notes-%s", sku)
```
alexeldeib: trouble testing due to ssl errors...will revert

PR #2020, opened 2022-07-26
remove updateAptWithMicrosoftPkg call from cse main

ganeshkumar5699:
**What type of PR is this?**
/kind bug

Egress tests in RP, restrict calls to packages.microsoft.com, during CSE execution. Removing `updateAptWithMicrosoftPkg` as that makes calls to packages.microsoft.com, in `cse_main.sh`.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/e2e-scenario.sh
```diff
@@ -75,7 +75,7 @@ az vmss create -n ${VMSS_NAME} \
     --vm-sku $VM_SKU \
     --instance-count 1 \
     --assign-identity $msiResourceID \
-    --image "microsoft-aks:aks:aks-ubuntu-1804-2022-q1:2022.02.01" \
```
ganeshkumar5699, line 78: Updating this image, since it needs the newer VHD (with the updated apt packages).

ganeshkumar5699: Otherwise Agentbaker e2e will fail, even though the behavior is correct

ganeshkumar5699: Will likely change it to a different image before merging (GPU tests seem to have restrictions due to it too)

e2e/e2e-scenario.sh
```diff
@@ -75,7 +75,7 @@ az vmss create -n ${VMSS_NAME} \
...
+    --image /subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1658785995.106 \
```
alexeldeib, line 78: please revert due to auth issues

alexeldeib: err, gc issues? both? lol

alexeldeib (approved):

anujmaheshwari1 (approved):

PR #2031, opened 2022-07-28
fix: nvidia 470/510 drivers for K80 deprecation

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -544,7 +544,8 @@ ensureGPUDrivers() {
         # needs to happen even on gpu vhd because newer containerd/runc broke old 
         # nvidia-container-runtime. containerd [1.5.9, 1.4.12] + runc 1.0.2 don't work with 
         # old nvidia-container-runtime like 2.0.0, only new like 3.6.0
-        installNvidiaContainerRuntime "${NVIDIA_CONTAINER_RUNTIME_VERSION}"
```
alexeldeib, line 547: unused

pkg/agent/baker.go
```diff
@@ -807,5 +807,16 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetOutboundCommand": func() string {
 			return getOutBoundCmd(config, config.CloudSpecConfig)
 		},
+		"GPUDriverVersion": func() string {
+			if isStandardNCv1(profile.VMSize) {
```
alexeldeib, line 811: I do not like this, but I would prefer the fix to be in AB and as simple as possible for now...

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -96,6 +96,9 @@ configGPUDrivers() {
 setupGpuRunfileInstall() {
     mkdir -p $GPU_DEST/lib64 $GPU_DEST/overlay-workdir
     retrycmd_if_failure 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+    export -f isARM64
```
alexeldeib, line 99: next PR: ruthlessly cut out this file I keep forgetting exists

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -199,23 +199,31 @@ installBpftrace
...
-  echo "  - nvidia-container-runtime=${NVIDIA_CONTAINER_RUNTIME_VERSION}";
-  echo "  - nvidia-gpu-driver-version=${GPU_DV}";
-  echo "  - nvidia-fabricmanager=${GPU_DV}";
-} >> ${VHD_LOGS_FILEPATH}
-if grep -q "fullgpu" <<< "$FEATURE_FLAGS"; then
+  # for dedicated GPU VHD, install older driver version compatible with all VM sizes
+  # consider this informal "deprecation" since we can't support one version right now.
+  if grep -q "fullgpu" <<< "$FEATURE_FLAGS"; then
     echo "  - ensureGPUDrivers" >> ${VHD_LOGS_FILEPATH}
+    export GPU_DV=470.57.02
```
alexeldeib, line 206: this forces the same version we have been using on the gpu vhd to avoid breaking those users. they won't get v5 support for now. a subtle nudge to get them on the normal vhd since everything is precached with no egress there anyway..

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -199,23 +199,31 @@ installBpftrace
...
     ensureGPUDrivers
-fi
+  else
+    # for non-GPU VHD, download all the newer driver bits, but don't install them yet.
+    export GPU_DV=510.47.03
```
alexeldeib, line 210: this bumps the default version to 510.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -89,7 +89,7 @@ UBUNTU_OS_NAME="UBUNTU"
 MARINER_OS_NAME="MARINER"
 KUBECTL=/usr/local/bin/kubectl
 DOCKER=/usr/bin/docker
-export GPU_DV=470.57.02
+export GPU_DV="{{GPUDriverVersion}}"
```
alexeldeib, line 92: this is no longer user for vhd build, but will dynamically select the correct version at runtime. for the NCv1 sizes which no longer support 510 series, this will mean egress calls.

alexeldeib: this also means the vanilla aks vhd will work for all GPU sizes however, which is the point.

pkg/agent/baker.go
alexeldeib: this miiiight break the RP egress tests for dedicated vhd depending on vm size...
maybe can cache both -_-

PR #2036, opened 2022-08-01
feat: Build Windows 2022 Containerd Gen 2 VHD

wanqingfu:
**What type of PR is this?**
Build Gen2 VHD for 2022 containerd

/kind feature

To add support of building Windows 2022 containerd Gen 2 VHD in AgentBaker.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -17,6 +17,10 @@ parameters:
   displayName: Build 2022 containerd
   type: boolean
   default: True
+- name: build2022containerdgen2
+  displayName: Build 2022 containerd Gen 2
+  type: boolean
+  default: True
 - name: dryrun
   displayName: Dry run
   type: boolean
@@ -91,3 +95,25 @@ stages:
         - template: ./templates/.builder-release-template-windows.yaml
           parameters:
             artifactName: 2022-containerd
+
+  - stage: build_vhd_2022_containerd_gen2
+    dependsOn: []
+    condition: eq('${{ parameters.build2022containerdgen2 }}', True)
+    jobs:
+    - job: build
+      timeoutInMinutes: 180
+      steps:
+        - bash: |
+            echo '##vso[task.setvariable variable=DRY_RUN]${{parameters.dryrun}}'
+            echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
+            echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D4s_v3'
+            echo '##vso[task.setvariable variable=CONTAINER_RUNTIME]containerd'
+            echo '##vso[task.setvariable variable=WINDOWS_SKU]2022-containerd'
+            echo '##vso[task.setvariable variable=VNET_RESOURCE_GROUP_NAME]vnet-2022-containerd-gen2-$(Build.BuildNumber)'
+            echo '##vso[task.setvariable variable=WINDOWS_BASE_IMAGE_URL]$(WINDOWS_2022_GEN2_BASE_IMAGE_URL)'
+            echo '##vso[task.setvariable variable=WINDOWS_NANO_IMAGE_URL]$(WINDOWS_2022_NANO_IMAGE_URL)'
+            echo '##vso[task.setvariable variable=WINDOWS_CORE_IMAGE_URL]$(WINDOWS_2022_CORE_IMAGE_URL)'
+          displayName: Setup Build Variables
+        - template: ./templates/.builder-release-template-windows.yaml
+          parameters:
+            artifactName: 2022-containerd-gen2
--- a/.pipelines/.vsts-vhd-builder-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-windows.yaml
@@ -17,6 +17,10 @@ parameters:
   displayName: Build 2022 containerd
   type: boolean
   default: True
+- name: build2022containerdgen2
+  displayName: Build 2022 containerd Gen 2
+  type: boolean
+  default: True  
 - name: dryrun
   displayName: Dry run
   type: boolean
@@ -98,3 +102,26 @@ stages:
         - template: ./templates/.builder-release-template-windows.yaml
           parameters:
             artifactName: 2022-containerd
+
+  - stage: build_vhd_2022_containerd_gen2
+    dependsOn: []
+    condition: eq('${{ parameters.build2022containerdgen2 }}', True)
+    jobs:
+    - job: build
+      timeoutInMinutes: 180
+      steps:
+        - bash: |
+            echo '##vso[task.setvariable variable=DRY_RUN]${{parameters.dryrun}}'
+            echo '##vso[task.setvariable variable=VHD_DEBUG]${{parameters.vhddebug}}'            
+            echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
+            echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D4s_v3'
+            echo '##vso[task.setvariable variable=CONTAINER_RUNTIME]containerd'
+            echo '##vso[task.setvariable variable=WINDOWS_SKU]2022-containerd'
+            echo '##vso[task.setvariable variable=VNET_RESOURCE_GROUP_NAME]vnet-2022-containerd-gen2-$(Build.BuildNumber)'
+            echo '##vso[task.setvariable variable=WINDOWS_BASE_IMAGE_URL]$(WINDOWS_2022_GEN2_BASE_IMAGE_URL)'
+            echo '##vso[task.setvariable variable=WINDOWS_NANO_IMAGE_URL]$(WINDOWS_2022_NANO_IMAGE_URL)'
+            echo '##vso[task.setvariable variable=WINDOWS_CORE_IMAGE_URL]$(WINDOWS_2022_CORE_IMAGE_URL)'
+          displayName: Setup Build Variables
+        - template: ./templates/.builder-release-template-windows.yaml
+          parameters:
+            artifactName: 2022-containerd-gen2
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -4,11 +4,39 @@ parameters:
       default: 2019
 
 steps:
+# For Gen 2, if we want to use sig for test purpose only, we set the pipeline variables
+#               SIG_GALLERY_NAME, SIG_IMAGE_NAME_PREFIX, and SIG_IMAGE_VERSION.
+#               SIG_IMAGE_NAME_PREFIX will be used to define the task variable SIG_IMAGE_NAME for
+#               the packer destination SIG.
+#               SIG_IMAGE_NAME, SIG_GALLERY_NAME, and SIG_IMAGE_VERSION will be passed below.
+#               Task variable GEN2_SIG_FOR_PRODUCTION is set to False and passed to the following steps.
+#               Built sig will not be deleted because it is for future testing.
+#            if we want to publish the VHD, instead of using pipeline variables,
+#               we will hard-code the task variables SIG_GALLERY_NAME, SIG_IMAGE_NAME, and SIG_IMAGE_VERSION.
+#               Task variable GEN2_SIG_FOR_PRODUCTION is set to True and passed to the following steps.
+#               Built sig will be deleted because it has been converted to VHD, and thus not needed.
   - bash: |
-      if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then m="sigMode"; sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; else m="default"; fi && \
-      echo "Set SIG_IMAGE_NAME to $sigImageName" && \
-      echo "##vso[task.setvariable variable=SIG_IMAGE_NAME]$sigImageName" && \
+      if [[ $(HYPERV_GENERATION) == "V2" ]]; then \
+        m="sigMode"; \
+        if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then \   
+          sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
+          echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
+        else
+          sigImageName="windows-${WINDOWS_SKU}"; \
+          sigGalleryName=WS2022Gen2Gallery$(Build.BuildNumber)${RANDOM}"; \
+          sigGalleryName=$(echo $sigGalleryName | tr '-' '_'); \
+          echo "##vso[task.setvariable variable=SIG_GALLERY_NAME]$sigGalleryName"; \
+          echo "##vso[task.setvariable variable=SIG_IMAGE_VERSION]1.0.0"; \
+          echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]True"; \
+        fi
+      elif [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then 
+        m="sigMode"; sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; \
+        echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
+      else 
+        m="default"; \
+      fi && \
       echo "Set build mode to $m" && \
+      echo "##vso[task.setvariable variable=SIG_IMAGE_NAME]$sigImageName" && \
       echo "##vso[task.setvariable variable=MODE]$m" && \
       if [[ "${USE_RELEASE_DATE}" = "False" ]]; then \
         echo "use current date as build date";  BUILD_DATE=$(date +"%y%m%d"); \
@@ -18,6 +46,7 @@ steps:
       echo "Set build date to $BUILD_DATE" && \
       echo "##vso[task.setvariable variable=BUILD_DATE]$BUILD_DATE"
     displayName: Get Build Mode
+
   - script: |
       echo MODE=$(MODE) && \
       docker run --rm \
@@ -42,7 +71,8 @@ steps:
       -e SIG_GALLERY_NAME=${SIG_GALLERY_NAME} \
       -e SIG_IMAGE_NAME=${SIG_IMAGE_NAME} \
       -e SIG_IMAGE_VERSION=${SIG_IMAGE_VERSION} \
-      -e HYPERV_GENERATION="V1" \
+      -e GEN2_SIG_FOR_PRODUCTION=${GEN2_SIG_FOR_PRODUCTION} \
+      -e HYPERV_GENERATION=${HYPERV_GENERATION} \
       -e VNET_RESOURCE_GROUP_NAME=${VNET_RESOURCE_GROUP_NAME} \
       -e WINDOWS_PATCH_URL=${WINDOWS_PATCH_URL} \
       -e WINDOWS_BASE_IMAGE_URL=${WINDOWS_BASE_IMAGE_URL} \
@@ -50,8 +80,12 @@ steps:
       -e WINDOWS_CORE_IMAGE_URL=${WINDOWS_CORE_IMAGE_URL} \
       ${CONTAINER_IMAGE} make -f packer.mk run-packer-windows
     displayName: Building windows VHD
+
+# Note: use -a to grep MANAGED_SIG_ID (packer-output should be read as a binary file in Linux)
   - script: |
+      echo MODE=$(MODE) && \
       OS_DISK_URI="$(cat packer-output | grep -a "OSDiskUri:" | cut -d " " -f 2)" && \
+      MANAGED_SIG_ID="$(cat packer-output | grep -a "ManagedImageSharedImageGalleryId:" | cut -d " " -f 2)" && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
@@ -61,6 +95,7 @@ steps:
       -e SUBSCRIPTION_ID="${BUILD_SUBSCRIPTION_ID}" \
       -e AZURE_RESOURCE_GROUP_NAME=${BUILD_AZURE_RESOURCE_GROUP_NAME} \
       -e OS_DISK_URI=${OS_DISK_URI} \
+      -e MANAGED_SIG_ID=${MANAGED_SIG_ID} \
       -e AZURE_LOCATION=${BUILD_AZURE_LOCATION} \
       -e CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) \
       -e WINDOWS_SKU=$(WINDOWS_SKU) \
@@ -73,10 +108,14 @@ steps:
       -e ARCHITECTURE=${ARCHITECTURE} \
       ${CONTAINER_IMAGE} make -f packer.mk test-building-vhd
     displayName: Run VHD cache test
+
+# We can generate release notes for check-in pr and sig mode to validate whether it is expected.
   - task: PublishPipelineArtifact@0
     inputs:
         artifactName: 'vhd-release-notes-${{ parameters.artifactName }}'
         targetPath: 'release-notes.txt'
+
+# Default Gen 1: packer VHD=>VHD in classic SA
   - script: |
       echo MODE=$(MODE) && \
       OS_DISK_SAS="$(cat packer-output | grep -a "OSDiskUriReadOnlySas:" | cut -d " " -f 2)" && \
@@ -91,17 +130,52 @@ steps:
       -e CLASSIC_BLOB=${STORAGE_ACCT_BLOB_URL} \
       -e CLASSIC_SAS_TOKEN="$(STORAGE_ACCT_SAS_TOKEN)" \
       -e OS_DISK_SAS=${OS_DISK_SAS} \
-      -e VHD_NAME=${VHD_NAME} \
       ${CONTAINER_IMAGE} make -f packer.mk az-copy
     displayName: Copying resource to Classic Storage Account
     condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), eq(variables['MODE'], 'default'))
+
+# Moved conversion to VHD before cleanup.
+# Gen 2 packer outputs a sig in destination. This step: dest sig=>disk=>VHD in classic SA for publishing.
+  - bash: |
+        echo MODE=$(MODE) && \
+        create_time="$(cat vhdbuilder/packer/settings.json | grep "create_time" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        SIG_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "sig_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        docker run --rm \
+        -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
+        -w /go/src/github.com/Azure/AgentBaker \
+        -e CLIENT_ID=${CLIENT_ID} \
+        -e CLIENT_SECRET="$(CLIENT_SECRET)" \
+        -e TENANT_ID=${TENANT_ID} \
+        -e SUBSCRIPTION_ID="${SUBSCRIPTION_ID}" \
+        -e RESOURCE_GROUP_NAME="${AZURE_RESOURCE_GROUP_NAME}" \
+        -e CREATE_TIME="${create_time}" \
+        -e LOCATION="${AZURE_LOCATION}" \
+        -e OS_TYPE="Windows" \
+        -e CLASSIC_BLOB=${STORAGE_ACCT_BLOB_URL} \
+        -e CLASSIC_SAS_TOKEN="$(STORAGE_ACCT_SAS_TOKEN)" \
+        -e SIG_IMAGE_NAME=${SIG_IMAGE_NAME} \
+        -e GEN2_CAPTURED_SIG_VERSION=${gen2_captured_sig_version} \
+        ${CONTAINER_IMAGE} make -f packer.mk convert-sig-to-classic-storage-account-blob
+    displayName: Convert Shared Image Gallery To VHD Blob In Classic Storage Account
+    condition: and(eq(variables.DRY_RUN, 'False'), eq(variables.GEN2_SIG_FOR_PRODUCTION, 'True'))
+
+# SA_NAME:             Temporary storage account name
+# IMPORTED_IMAGE_NAME: Build output for sigMode is SIG. Packer does not support building a SIG from raw
+#                      VHD blob directly. Will use this as an intermediate sig to import from raw VHD url.
+#                      Can be deleted after building.
+# SIG_IMAGE_NAME:      Packer-generated SIG image for sigMode. (Gen 2 included) If GEN2_SIG_FOR_PRODUCTION = True
+#                      This image should be deleted. Otherwise, we keep it.
+#                      Used to define the IMAGE_NAME below.
+# IMAGE_NAME:          Managed image created by packer: ManagedImageName in log. No longer used.
+#                      Can be cleaned up in this step.
   - bash: |
       echo MODE=$(MODE) && \
       PKR_RG_NAME="$(cat packer-output | grep -a "ResourceGroupName" | cut -d "'" -f 2 | head -1)" && \
       SA_NAME="$(cat packer-output | grep -a "storage name:" | cut -d " " -f 3)" && \
-      gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
       IMPORTED_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "imported_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
-      if [[ $(MODE) == "gen2Mode" ]]; then IMAGE_NAME="${gen2_captured_sig_version}"; else IMAGE_NAME=$UPDATED_SIG_IMAGE_NAME; fi && \
+      SIG_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "sig_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+      IMAGE_NAME="$(cat packer-output | grep -a "ManagedImageName:" | cut -d " " -f 2)" && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
@@ -116,18 +190,29 @@ steps:
       -e DRY_RUN=${DRY_RUN} \
       -e IMAGE_NAME=${IMAGE_NAME} \
       -e VNET_RESOURCE_GROUP_NAME=${VNET_RESOURCE_GROUP_NAME} \
+      -e WINDOWS_SKU=$(WINDOWS_SKU) \
       -e IMPORTED_IMAGE_NAME=${IMPORTED_IMAGE_NAME} \
       -e SIG_GALLERY_NAME=${SIG_GALLERY_NAME} \
+      -e SIG_IMAGE_NAME=${SIG_IMAGE_NAME} \
+      -e GEN2_SIG_FOR_PRODUCTION=${GEN2_SIG_FOR_PRODUCTION} \
       ${CONTAINER_IMAGE} make -f packer.mk cleanup
     displayName: Clean Up Packer Generated Resources
     condition: always()
+
+# Set VHD_NAME and SKU_NAME for Gen 2, which will be published.
+# Note: use -a to grep OS_DISK_SAS (packer-output should be read as a binary file in Linux)
+# Perform this step only if we want to publish the VHD: Default Gen 1 or Gen 2 (sigMode) and the built sig is for production.
   - bash: |
       echo MODE=$(MODE) && \
-      OS_DISK_SAS="$(cat packer-output | grep -a "OSDiskUriReadOnlySas:" | cut -d " " -f 2)" && \
-      VHD_NAME="$(echo $OS_DISK_SAS | cut -d "/" -f 8 | cut -d "?" -f 1)" && \
+      if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
+      [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
+      SKU_NAME="windows-$WINDOWS_SKU"; \
+      else \
+      OS_DISK_SAS="$(cat packer-output | grep -a "OSDiskUriReadOnlySas:" | cut -d " " -f 2)"; \
+      VHD_NAME="$(echo $OS_DISK_SAS | cut -d "/" -f 8 | cut -d "?" -f 1)"; \
+      SKU_NAME="windows-$WINDOWS_SKU"; fi && \
       WINDOWS_VERSION="$(grep -a "OS Version" < release-notes.txt | cut -d ":" -f 2 | tr -d "[:space:]")" && \
       IMAGE_VERSION="$WINDOWS_VERSION.${BUILD_DATE}" && \
-      SKU_NAME="windows-$WINDOWS_SKU" && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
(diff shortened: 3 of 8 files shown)


.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -90,4 +94,26 @@ stages:
...
+  - stage: build_vhd_2022_containerd_gen2
+    dependsOn: []
+    condition: eq('${{ parameters.build2022containerdgen2 }}', True)
+    jobs:
+    - job: build
+      timeoutInMinutes: 180
+      steps:
+        - bash: |
+            echo '##vso[task.setvariable variable=DRY_RUN]${{parameters.dryrun}}'
+            echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
```
AbelHu, line 108: Why do you remove `echo '##vso[task.setvariable variable=VHD_DEBUG]${{parameters.vhddebug}}'`?

.pipelines/.vsts-vhd-builder-windows.yaml
```diff
@@ -97,4 +101,26 @@ stages:
...
+
+  - stage: build_vhd_2022_containerd_gen2
+    dependsOn: []
+    condition: eq('${{ parameters.build2022containerdgen2 }}', True)
+    jobs:
+    - job: build
+      timeoutInMinutes: 180
+      steps:
+        - bash: |
+            echo '##vso[task.setvariable variable=DRY_RUN]${{parameters.dryrun}}'
```
AbelHu, line 114: Same question as that for [.pipelines/.vsts-vhd-builder-release-windows.yaml](https://github.com/Azure/AgentBaker/pull/2036/files#diff-7ff44dd52bb7628aab9652bdb37642f273e4ec43cd48b5e5f4c752bdba31478c)

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -5,10 +5,11 @@ parameters:
 
 steps:
   - bash: |
-      if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then m="sigMode"; sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; else m="default"; fi && \
+      if [[ $(HYPERV_GENERATION) == "V2" ]]; then m="gen2Mode"; elif [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then m="sigMode"; sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; else m="default"; fi && \
```
AbelHu, line 8: I think that we also need to support `sigMode` in building WS2022 gen2 VHD for test purpose. I think that there are below options:
1. set mode to `sigMode` with `HYPERV_GENERATION=V2`
2. set mode to `gen2Mode` with setting `SIG_GALLERY_NAME`, `sigImageName` and `SIG_IMAGE_VERSION`
3. define a new mode `gen2SigMode`

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -42,16 +44,25 @@ steps:
...
       ${CONTAINER_IMAGE} make -f packer.mk run-packer-windows
     displayName: Building windows VHD
+
+# Note: use -a to grep MANAGED_SIG_ID (packer-output should be read as a binary file in Linux)
   - script: |
+      echo MODE=$(MODE) && \
       OS_DISK_URI="$(cat packer-output | grep -a "OSDiskUri:" | cut -d " " -f 2)" && \
+      MANAGED_SIG_ID="$(cat packer-output | grep -a "ManagedImageSharedImageGalleryId:" | cut -d " " -f 2)" && \
+      gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+      echo "OS_DISK_URI ${OS_DISK_URI}" && \
```
AbelHu, line 62: Remove these lines with `echo`?

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -73,35 +86,73 @@ steps:
...
       VHD_NAME="$(echo $OS_DISK_SAS | cut -d "/" -f 8 | cut -d "?" -f 1)" && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
-      -e CLIENT_ID=${CLIENT_ID} \
-      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-      -e TENANT_ID=${TENANT_ID} \
-      -e SUBSCRIPTION_ID="${SUBSCRIPTION_ID}" \
+      -e CLIENT_ID=${BUILD_CLIENT_ID} \
+      -e CLIENT_SECRET="$(BUILD_CLIENT_SECRET)" \
```
AbelHu, line 103: why do you change these? Currently Windows uses `CLIENT_ID`, `CLIENT_SECRET` and `TENANT_ID` to copy the built VHD to the specified storage account and it works.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -114,40 +165,50 @@ steps:
...
       IMAGE_VERSION="$WINDOWS_VERSION.${BUILD_DATE}" && \
-      SKU_NAME="windows-$WINDOWS_SKU" && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
-      -e CLIENT_ID=${CLIENT_ID} \
-      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-      -e TENANT_ID=${TENANT_ID} \
-      -e SUBSCRIPTION_ID="${SUBSCRIPTION_ID}" \
+      -e CLIENT_ID=${BUILD_CLIENT_ID} \
```
AbelHu, line 194: Same issue as above

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: It didn't work before. Will check on this.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -73,35 +86,73 @@ steps:
...
+      -e AZURE_LOCATION=${BUILD_AZURE_LOCATION} \
       -e CLASSIC_BLOB=${STORAGE_ACCT_BLOB_URL} \
       -e CLASSIC_SAS_TOKEN="$(STORAGE_ACCT_SAS_TOKEN)" \
       -e OS_DISK_SAS=${OS_DISK_SAS} \
       -e VHD_NAME=${VHD_NAME} \
       ${CONTAINER_IMAGE} make -f packer.mk az-copy
     displayName: Copying resource to Classic Storage Account
     condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), eq(variables['MODE'], 'default'))
+
+# Moved this conversion before cleanup because packer-generated SIG is already in VHD, and so can be cleaned.
```
AbelHu, line 116: `conversion` => `step`? How to understand `packer-generated SIG is already in VHD`?

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -73,35 +86,73 @@ steps:
...
       -e VHD_NAME=${VHD_NAME} \
       ${CONTAINER_IMAGE} make -f packer.mk az-copy
     displayName: Copying resource to Classic Storage Account
     condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), eq(variables['MODE'], 'default'))
+
+# Moved this conversion before cleanup because packer-generated SIG is already in VHD, and so can be cleaned.  
+  - bash: |
+        echo MODE=$(MODE) && \
+        create_time="$(cat vhdbuilder/packer/settings.json | grep "create_time" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        SIG_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "sig_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
```
AbelHu, line 120: Gen2 also needs to support sigMode so we should not reuse `SIG_IMAGE_NAME`

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -73,35 +86,73 @@ steps:
...
+        gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        docker run --rm \
+        -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
+        -w /go/src/github.com/Azure/AgentBaker \
+        -e CLIENT_ID=${BUILD_CLIENT_ID} \
+        -e CLIENT_SECRET="$(BUILD_CLIENT_SECRET)" \
+        -e TENANT_ID=${BUILD_TENANT_ID} \
+        -e SUBSCRIPTION_ID="${BUILD_SUBSCRIPTION_ID}" \
```
AbelHu, line 128: Same issue as above

packer.mk
```diff
@@ -88,7 +91,7 @@ run-packer-windows: az-login
 	@packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer-windows | tee -a packer-output)
 
 az-copy: az-login
-	azcopy-preview copy "${OS_DISK_SAS}" "${CLASSIC_BLOB}${CLASSIC_SAS_TOKEN}" --recursive=true
+#	azcopy-preview copy "${OS_DISK_SAS}" "${CLASSIC_BLOB}${CLASSIC_SAS_TOKEN}" --recursive=true
```
AbelHu, line 94: Need to uncomment this line before submitting the PR

packer.mk
```diff
@@ -66,7 +66,10 @@ else
 endif
 
 build-packer-windows:
-ifeq (${MODE},sigMode)
+ifeq (${MODE},gen2Mode)
+	@echo "${MODE}: Building with Hyper-v generation 2 VM and save to Classic Storage Account"
+	@packer build -var-file=vhdbuilder/packer/settings.json vhdbuilder/packer/windows-vhd-builder-gen2.json
+else ifeq (${MODE},sigMode)
```
AbelHu, line 72: gen2 also needs to support sigMode. sigMode: publish the built VHD as the sig image in the resource group and do not delete it in cleanup step. We can use this sig image for test later.

vhdbuilder/packer/cleanup.sh
```diff
@@ -28,6 +28,7 @@ if [[ -n "$AZURE_RESOURCE_GROUP_NAME" && -n "$IMAGE_NAME" ]]; then
     id=$(az image show -n ${IMAGE_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id)
     if [ -n "$id" ]; then
       az image delete -n ${IMAGE_NAME} -g ${AZURE_RESOURCE_GROUP_NAME}
+      echo "deleting managed image ${IMAGE_NAME} under resource group ${AZURE_RESOURCE_GROUP_NAME}"
```
AbelHu, line 31: Move the logs before the action since it is `deleting`

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -34,8 +34,9 @@ az resource create --id $disk_resource_id  --is-full-object --location $LOCATION
 }"
 # shellcheck disable=SC2102
 sas=$(az disk grant-access --ids $disk_resource_id --duration-in-seconds 3600 --query [accessSas] -o tsv)
+echo "Converting $sig_resource_id to ${CLASSIC_BLOB}/${GEN2_CAPTURED_SIG_VERSION}.vhd"
 
-azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${GEN2_CAPTURED_SIG_VERSION}.vhd${CLASSIC_SAS_TOKEN}" --recursive=true
+# azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${GEN2_CAPTURED_SIG_VERSION}.vhd${CLASSIC_SAS_TOKEN}" --recursive=true
```
chengliangli0918, line 39: Why this line is commented out? It is required for Linux Gen2 VHD build.

For Linux Gen1 VHD build, packer can capture os disk directly to a vhd blob for publish pipeline to use (e.g. https://vhdbuilderaceastustest.blob.core.windows.net/vhd/aks-1659383205-osDisk.bd14574a-4a0d-4c2b-9e73-ae1d01bbe095.vhd ...)

For Linux Gen2 VHD build, packer cannot capture os disk directly to a vhd blob for publish pipeline to use later. The packer captures the os disk to a managed image, convert the managed image to a SIG (this is for amd64, arm64 is more complext). then this .sh script is called to convert the SIG to a vhd blob for publish pipeline to use later (e.g. https://vhdbuilderaceastustest.blob.core.windows.net/vhd/1.1659383200.4650.vhd ...). if this line is commented out, the contract between the VHD build pipeline and VHD publish pipeline for Gen2 VHDs are broken.

vhdbuilder/packer/cleanup.sh
```diff
@@ -59,6 +60,45 @@ if [[ -n "${IMPORTED_IMAGE_NAME}" ]]; then
   fi
 fi
 
+#cleanup managed image sig image version
+if [[ -n "${SIG_IMAGE_NAME}" ]]; then
+   echo "SIG_IMAGE_NAME is ${SIG_IMAGE_NAME}"
+   MANAGED_IMAGE_SIG_NAME=${SIG_IMAGE_NAME}
```
AbelHu, line 66: In sigMode, SIG_IMAGE_NAME is the final created sig image name for test and should not be deleted.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
AbelHu, line 39: Should uncomment before submitting the PR

vhdbuilder/packer/init-variables.sh
```diff
@@ -105,6 +105,14 @@ if [[ "$MODE" == "gen2Mode" ]]; then
 		if [[ "$OS_SKU" == "CBLMariner" ]]; then
 			SIG_IMAGE_NAME=${OS_SKU}${OS_VERSION//./}Gen2
 		fi
+
+		if [[ "$OS_TYPE" == "Windows" ]]; then
```
AbelHu, line 109: In sigMode, SIG_IMAGE_NAME is the final created sig image name for test so I think that we should use a different name for gen2 as the sig image we create based on the raw vhd in sigMode

vhdbuilder/packer/test/run-test.sh
```diff
@@ -6,7 +6,7 @@ WIN_SCRIPT_PATH="windows-vhd-content-test.ps1"
 TEST_RESOURCE_PREFIX="vhd-test"
 TEST_VM_ADMIN_USERNAME="azureuser"
 
-set +x
```
AbelHu, line 9: what is the purpose to remove this line?

vhdbuilder/packer/test/run-test.sh
```diff
@@ -63,14 +63,26 @@ else
 
     IMG_DEF="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/${SIG_GALLERY_NAME}/images/${SIG_IMAGE_NAME}/versions/${SIG_IMAGE_VERSION}"
   else 
-    #gen2Mode check, set the IMG_DEF to the MANAGED_SIG_ID retrieved from packer-output after VHD Build
-    IMG_DEF=${MANAGED_SIG_ID}
+    # gen2Mode check, set the IMG_DEF to the MANAGED_SIG_ID retrieved from packer-output after VHD Build
+    if [ "$OS_TYPE" == "Linux" ]; then
+      IMG_DEF=${MANAGED_SIG_ID}
+    else
+      # (to-do) for Windows: MANAGED_SIG_ID reads as "file"
```
AbelHu, line 70: do we still need this comment?

vhdbuilder/packer/test/run-test.sh
```diff
@@ -80,6 +92,9 @@ else
       --size Standard_D2pds_V5 \
       --public-ip-address ""
   else
+    # echo "architecture is ${ARCHITETURE,,}"
```
AbelHu, line 95: remove these 3 lines?

vhdbuilder/packer/windows-image.env
```diff
@@ -10,7 +10,19 @@ WINDOWS_2019_BASE_IMAGE_VERSION=17763.3165.220706
 WINDOWS_2022_BASE_IMAGE_SKU=2022-Datacenter-Core-smalldisk
 WINDOWS_2022_BASE_IMAGE_VERSION=20348.825.220704
 
+# Please refer to https://vincentlauzon.com/2018/01/10/finding-a-vm-image-reference-publisher-sku/
+# , https://docs.microsoft.com/en-us/azure/virtual-machines/windows/ps-template for ARM templates.
+# Please use this PowershellScript to check for available skus that azure-arm can use:
+#   Get-AzureRmVMImageSku -Location eastus -PublisherName MicrosoftWindowsServer  -Offer WindowsServer
```
AbelHu, line 16: Can we use `azure cli` as the example as we did for gen1?

vhdbuilder/packer/windows-vhd-builder-gen2.json
```diff
@@ -0,0 +1,149 @@
+{
```
AbelHu, line 1: It seems like that we can reuse vhdbuilder/packer/windows-vhd-builder-sig.json

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Moved conversion to VHD before cleanup as suggested. Because packer-generated SIG has already been converted to classical storage account, it can be deleted in cleanup for gen2Mode. For gen2SigMode (the one you suggested), this conversion will not be stepped in.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: For `HYPERV_GENERATION=V2`, we can define a new mode `gen2SigMode` if `SIG_GALLERY_NAME`, `sigImageName` and `SIG_IMAGE_VERSION` are set. If they are not set, mode is just `gen2Mode`? This will only add one condition branch in `packer.mk`: `ifeq (${MODE},gen2SigMode)`?

packer.mk
wanqingfu: Maybe add one branch `ifeq (${MODE},gen2SigMode)`? See the options above?

vhdbuilder/packer/init-variables.sh
wanqingfu: Maybe we can use a GEN2_SIG_IMAGE_NAME and pass it to the pipeline steps in .builder-release-template-windows.yaml? But then lots of changes need to be made in this file at least. Will that clutter the code?

vhdbuilder/packer/windows-vhd-builder-gen2.json
wanqingfu: Not exactly. We have a `"gen2_captured_sig_version"` that will be used when converting from packer SIG to the VHD for publish pipeline. Also, since we don't want to reuse `sig_image_name`, the image name for the packer output in `shared_image_gallery_destination` should be something like `gen2_sig_image_name`.

.pipelines/.vsts-vhd-builder-release-windows.yaml
wanqingfu: `vhddebug` is only used in `.vsts-vhd-builder-windows.yaml` but not this? The repo does not have it in this file. It was not even here in this commit https://github.com/Azure/AgentBaker/commit/b387ed14b8686ee93ea1b19ef23dcb242bf9fdd9#diff-7ff44dd52bb7628aab9652bdb37642f273e4ec43cd48b5e5f4c752bdba31478c.

AbelHu: "vhddebug is only used in .vsts-vhd-builder-windows.yaml" yes. I did not see it in .vsts-vhd-builder-windows.yaml. I tagged a wrong file.

wanqingfu: Added back https://github.com/Azure/AgentBaker/pull/2036/commits/7dfa348264f6453af6a2be809c1b517b66b94de8.

.pipelines/.vsts-vhd-builder-windows.yaml
wanqingfu: Added back https://github.com/Azure/AgentBaker/pull/2036/commits/7dfa348264f6453af6a2be809c1b517b66b94de8.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Removed.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: I misunderstood. I think `CLIENT_ID`, `CLIENT_SECRET`, and `TENANT_ID` are set in containerd&docker pipeline, but not in the PR check-in pipeline. When I was incorrectly using DRY_RUN=False and toying around with az-copy and convert-sig-to-classic-storage-account, I set them to `BUILD_...`. Corrected in the commit https://github.com/Azure/AgentBaker/pull/2036/commits/7dfa348264f6453af6a2be809c1b517b66b94de8.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Corrected.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Does the built SIG need to be deleted for gen2Mode (either Linux or Windows). I think if DRY_RUN=False, and the conversion to VHD is done, we don't need to keep the SIG for test? (This conversion step is after testing the built VHD)

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Removed the `"BUILD"` prefix.

packer.mk
wanqingfu: Uncommented.

vhdbuilder/packer/cleanup.sh
wanqingfu: Moved the log before.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
wanqingfu: Yes. Uncommented.

vhdbuilder/packer/test/run-test.sh
wanqingfu: No. `IMG_DEF=${MANAGED_SIG_ID}` for both Linux and Windows.

vhdbuilder/packer/test/run-test.sh
wanqingfu: Removed.

vhdbuilder/packer/windows-image.env
wanqingfu: Changed to the `azure cli` example. Would you take a look https://github.com/Azure/AgentBaker/pull/2036/commits/7dfa348264f6453af6a2be809c1b517b66b94de8?

vhdbuilder/packer/test/run-test.sh
wanqingfu: Added back.

vhdbuilder/packer/cleanup.sh
wanqingfu: What about the gen2Mode? (the one that matches with Linux) GEN2_SIG_IMAGE_NAME has already been converted to VHD. The steps: `run-test` => `convert` => `cleanup` in the current pipeline. Should we delete it?

For gen2SigMode, I think the final created sig should be kept there?

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
wanqingfu: Uncommented.

vhdbuilder/packer/cleanup.sh
AbelHu: for gen2Mode, the built sig image should be deleted since we only need to upload the vhd to the storage account after converting the sig to vhd.
for gen2SigMode, we need to keep it. But we need to use the provide gallery name, image name and image version.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -5,10 +5,11 @@ parameters:
 
 steps:
   - bash: |
-      if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then m="sigMode"; sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; else m="default"; fi && \
+      if [[ $(HYPERV_GENERATION) == "V2" ]]; then m="gen2Mode"; elif [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then m="sigMode"; sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; else m="default"; fi && \
       echo "Set SIG_IMAGE_NAME to $sigImageName" && \
       echo "##vso[task.setvariable variable=SIG_IMAGE_NAME]$sigImageName" && \
       echo "Set build mode to $m" && \
+      echo "##vso[task.setvariable variable=SIG_IMAGE_NAME]$sigImageName" && \
```
AbelHu, line 12: Why adding line 12 since we have set it in line 10?

.pipelines/templates/.builder-release-template-windows.yaml
AbelHu: let us sync in teams

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -42,16 +44,21 @@ steps:
...
       -e WINDOWS_PATCH_URL=${WINDOWS_PATCH_URL} \
       -e WINDOWS_BASE_IMAGE_URL=${WINDOWS_BASE_IMAGE_URL} \
       -e WINDOWS_NANO_IMAGE_URL=${WINDOWS_NANO_IMAGE_URL} \
       -e WINDOWS_CORE_IMAGE_URL=${WINDOWS_CORE_IMAGE_URL} \
       ${CONTAINER_IMAGE} make -f packer.mk run-packer-windows
     displayName: Building windows VHD
+
+# Note: use -a to grep MANAGED_SIG_ID (packer-output should be read as a binary file in Linux)
   - script: |
+      echo MODE=$(MODE) && \
```
AbelHu, line 58: remove line 58

vhdbuilder/packer/windows-image.env
```diff
@@ -10,7 +10,19 @@ WINDOWS_2019_BASE_IMAGE_VERSION=17763.3165.220706
...
+WINDOWS_2022_GEN2_BASE_IMAGE_SKU=2022-datacenter-core-smalldisk-g2
+WINDOWS_2022_GEN2_BASE_IMAGE_VERSION=20348.825.220704
+
 # Please uncomment the following lines and set a larger os disk size that is at least 30GB when your PR check-in fails
 # WINDOWS_2019_OS_DISK_SIZE_GB=30
 # WINDOWS_2019_CONTAINERD_OS_DISK_SIZE_GB=30
-# WINDOWS_2022_CONTAINERD_OS_DISK_SIZE_GB=30
\ No newline at end of file
+# WINDOWS_2022_CONTAINERD_OS_DISK_SIZE_GB=30
+# WINDOWS_2019_CONTAINERD_GEN2_OS_DISK_SIZE_GB=30
```
AbelHu, line 27: Remove line 27

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Yeah. Just noticed that. But we can define a $gen2sigImageName?

wanqingfu: Fixed.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: ?

vhdbuilder/packer/windows-image.env
wanqingfu: Removed.

vhdbuilder/packer/init-variables.sh
wanqingfu: Will reuse ``SIG_IMAGE_NAME`` and set the values differently for ``GEN2_SIG_FOR_PRODUCTION`` and ``GEN2_SIG_FOR_TEST``.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Decided to reuse it but it needs to set values differently depending on whether the gen2 sig is for test purpose only or for production.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Synced in teams. The built SIG will be deleted if it's for production ``GEN2_SIG_FOR_PRODUCTION = True`` because ``SIG=>VHD``. Will not delete if the sig is only for test.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Decision: we will reuse `sigMode` for Gen 2 and use the variable "`GEN2_SIG_FOR_PRODUCTION`" to indicate if it's for test or for publishing. If pipeline variables `SIG_GALLERY_NAME`, `SIG_IMAGE_NAME_PREFIX`, and `SIG_IMAGE_VERSION` are set, we are using sig for test. O.w. we use sig for production.

packer.mk
wanqingfu: Only two branches `sigMode` and `Default`. This file will check `GEN2_SIG_FOR_PRODUCTION` and would go into the real "`sigMode`" if this variable is `False`. (sig is for test purpose only)

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -4,11 +4,37 @@ parameters:
...
-      echo "Set SIG_IMAGE_NAME to $sigImageName" && \
-      echo "##vso[task.setvariable variable=SIG_IMAGE_NAME]$sigImageName" && \
+      if [[ $(HYPERV_GENERATION) == "V2" ]]; then \
+        m="sigMode"; \
+        if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then \   
+          sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
+          echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
+        else
+          sigImageName="windows-${WINDOWS_SKU}"; \
+          echo "##vso[task.setvariable variable=SIG_GALLERY_NAME]WS2022Gen2Gallery_${RANDOM}"; \
```
AbelHu, line 26: Use `WS2022Gen2Gallery-${RANDOM}` instead of `WS2022Gen2Gallery_${RANDOM}`

.pipelines/templates/.builder-release-template-windows.yaml
AbelHu: You have changed the file several times so the comment is not correct now

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -91,17 +131,52 @@ steps:
       -e CLASSIC_BLOB=${STORAGE_ACCT_BLOB_URL} \
       -e CLASSIC_SAS_TOKEN="$(STORAGE_ACCT_SAS_TOKEN)" \
       -e OS_DISK_SAS=${OS_DISK_SAS} \
-      -e VHD_NAME=${VHD_NAME} \
```
AbelHu, line 94: do we not need line 94?

packer.mk
```diff
@@ -67,7 +67,15 @@ endif
 
 build-packer-windows:
 ifeq (${MODE},sigMode)
+ifeq (${HYPERV_GENRATION},V1)
 	@echo "${MODE}: Building with Hyper-v generation 1 VM and save to Shared Image Gallery"
+else
+ifeq (${GEN2_SIG_FOR_PRODUCTION},True)
```
AbelHu, line 73: indent?

vhdbuilder/packer/init-variables.sh
```diff
@@ -301,21 +314,17 @@ if [ "$OS_TYPE" == "Windows" ]; then
...
 		# WINDOWS_IMAGE_URL to a custom VHD to use for your base image. If this value is set, image_publisher, image_offer, image_sku, or image_version should not be set.
 		WINDOWS_IMAGE_PUBLISHER=""
 		WINDOWS_IMAGE_OFFER=""
 		WINDOWS_IMAGE_SKU=""
 		WINDOWS_IMAGE_VERSION=""
-	fi
 
-	# Need to use a sig image to create the build VM
-	if [[ "$MODE" == "sigMode" || "$MODE" == "gen2Mode" ]]; then
-		if [ -n "${WINDOWS_BASE_IMAGE_URL}" ]; then
```
AbelHu, line 318: Does removing "WINDOWS_BASE_IMAGE_URL" mean that we do not support using raw vhd?

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Yeah. I rebased a couple times. Did you remember what was the issue?

AbelHu: I cannot remember it now. let us close this for now.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: I tried to use dash before. The log says **"ERROR: (BadRequest) Gallery name is invalid. Allowed characters are English alphanumeric characters, with underscores and periods allowed in the middle, up to 80 characters total. All other special characters, including dashes, are disallowed.".** 

See the most recent pipeline [https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=59422027&view=logs&j=05b0285b-7424-5da1-4fe4-29b48f7b1d3a&t=4f5de716-8c9d-57d8-0c59-15d673379c58](url)

AbelHu: I see. Maybe we can use "WS2022Gen2Gallery{RANDOM}". I am not sure whether it is better to use "WS2022Gen2Gallery$(Build.BuildNumber)".

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: I don't think we need? az-copy uses `OS_DISK_SAS`. generate-sas uses `VHD_NAME`?

vhdbuilder/packer/init-variables.sh
wanqingfu: We still support. Just moving the conditions `if [ -n "${WINDOWS_BASE_IMAGE_URL}" ]; then` outside and `if [[ "$MODE" == "sigMode" ]]; then` inside raw vhd check. So we will check if raw VHD is used first (**line 309**), and then check if we are using sigMode (**line 327**) (Removed `"$MODE" == "gen2Mode"` since we are reusing sigMode for Gen 2. ) The logic works for both Gen 1 & Gen 2. 

**Context in the master branch:** Line 296 and 318 of [init-variables from the official repo](https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/init-variables.sh) both check if we are using raw VHD. If under `sigMode` or `gen2Mode` and no raw VHD is given, line 360 to 365 will be stepped in to initialize the variables. (_empty strings_)

**In this commit**, line 301 through 307 initializes the sigmode source variables. If under `sigMode` and raw VHD, line 386 to 390 will change the sigmode sources accordingly.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Maybe $(Build.BuildNumber) is more informative since the BuildNumber like `20220823.5_merge_****` tells the date? When we check the gallery on portal to see if the gallery has been deleted - sometimes undeleted resources will break the pipeline, one can trace back to the code commit on that day and see what went wrong???

Further, I think we can use something similar to `STORAGE_ACCOUNT_NAME="aksimages${CREATE_TIME}$RANDOM"`, i.e. `WS2022Gen2Gallery$(Build.BuildNumber)${RANDOM}`?

AbelHu: storageaccount has length limitation "must be between 3 and 24 characters in length and may contain numbers and lowercase letters only."

wanqingfu: I meant we can emulate `STORAGE_ACCOUNT_NAME` to set `SIG_GALLERY_NAME` <= `WS2022Gen2Gallery$(Build.BuildNumber)${RANDOM}`. Gallery name is "up to 80 characters total".

**Proposal**: `WS2022Gen2Gallery$(Build.BuildNumber)${RANDOM}`?

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -116,38 +191,55 @@ steps:
...
       IMAGE_VERSION="$WINDOWS_VERSION.${BUILD_DATE}" && \
-      SKU_NAME="windows-$WINDOWS_SKU" && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
-      -e CLIENT_ID=${CLIENT_ID} \
-      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-      -e TENANT_ID=${TENANT_ID} \
-      -e SUBSCRIPTION_ID="${SUBSCRIPTION_ID}" \
+      -e CLIENT_ID=${BUILD_CLIENT_ID} \
```
AbelHu, line 220: why do we change CLIENT_ID to BUILD_CLIENT_ID?

AbelHu: We need to generate sas token for the published vhd but not for the built vhd

pkg/agent/datamodel/sig_config.go
```diff
@@ -208,6 +208,7 @@ var AvailableWindowsSIGDistros []Distro = []Distro{
 	AKSWindows2019,
 	AKSWindows2019Containerd,
 	AKSWindows2022Containerd,
+	AKSWindows2022ContainerdGen2,
```
AbelHu, line 211: It is better to separate these code in another PR from the build code.

vhdbuilder/packer/init-variables.sh
```diff
@@ -326,19 +335,35 @@ if [ "$OS_TYPE" == "Windows" ]; then
 				--location $AZURE_LOCATION \
 				--hyper-v-generation $HYPERV_GENERATION \
 				--os-type ${OS_TYPE}
+			
+			# Showing the created managed image from imported image
+			id=$(az image show --name ${IMPORTED_IMAGE_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id)
```
AbelHu, line 340: Do you use these code to show the created image, definition and version for debug? Suggest to remove it for better perf

.pipelines/templates/.builder-release-template-windows.yaml
AbelHu: I see. WS2022Gen2Gallery$(Build.BuildNumber)${RANDOM} LGTM

packer.mk
wanqingfu: `ifeq` is makefile syntax and cannot be preceded by TAB according to the explanation [here](https://stackoverflow.com/questions/28712585/when-to-use-space-or-tab-in-makefile)? 

Adding indentation fails [here](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=59467327&view=logs&j=05b0285b-7424-5da1-4fe4-29b48f7b1d3a&t=4f5de716-8c9d-57d8-0c59-15d673379c58). 
`-c: line 0: syntax error near unexpected token `,V1'
`/bin/bash: -c: line 0: `ifeq (,V1)'`

See the `"build-packer"` block in this file?

pkg/agent/datamodel/sig_config.go
wanqingfu: Moved https://github.com/Azure/AgentBaker/pull/2113.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Fixed.

vhdbuilder/packer/init-variables.sh
wanqingfu: Removed.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -73,10 +106,17 @@ steps:
       -e ARCHITECTURE=${ARCHITECTURE} \
       ${CONTAINER_IMAGE} make -f packer.mk test-building-vhd
     displayName: Run VHD cache test
+
+# Write release notes only if the mode is Default Gen 1 or it's sigMode Gen 2 (built sig is for production)
   - task: PublishPipelineArtifact@0
     inputs:
         artifactName: 'vhd-release-notes-${{ parameters.artifactName }}'
         targetPath: 'release-notes.txt'
+    condition: and(succeeded(), eq(variables.DRY_RUN, 'False'),
```
AbelHu, line 115: What is the purpose to add the condition for publishing release notes? We can generate release notes for check-in pr and sig mode to validate whether it is expected.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -34,6 +34,7 @@ az resource create --id $disk_resource_id  --is-full-object --location $LOCATION
 }"
 # shellcheck disable=SC2102
 sas=$(az disk grant-access --ids $disk_resource_id --duration-in-seconds 3600 --query [accessSas] -o tsv)
+echo "Converting $sig_resource_id to ${CLASSIC_BLOB}/${GEN2_CAPTURED_SIG_VERSION}.vhd"
```
AbelHu, line 37: Is it duplicated since we have the same log after line 39?

wanqingfu: Removed.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Removed.

AbelHu:
Please remove the empty file "cd"

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Just to clarify, for the step `build`, `test`, and `cleanup` we are using `BUILD_CLIENT_ID`. For the steps related to publishing `az-copy`, `convert`, and `generate-sas` we are using `CLIENT_ID`?

**Context**: For windows build pipeline `.vsts-vhd-builder-windows.yaml` it has variable `BUILD_CLIENT_ID`. For windows publish pipeline `.vsts-vhd-builder-release-windows.yaml` it has both `BUILD_CLIENT_ID` and `CLIENT_ID`. Both files will reuse the step template which is this file `.builder-release-template-windows.yaml`

For Linux, both build and publish pipeline have only the variable `CLIENT_ID`.

**Question**: I wonder what is the difference.

AbelHu: You can check the code which is used to generate sas token. Different credentials are used for different tenant/sub/storage accounts.

AbelHu (approved):

PR #2050, opened 2022-08-02
Add custom ca trust data model changes

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/templates/templates_generated.go
```diff
@@ -5399,6 +5399,16 @@ write_files:
     #EOF
 {{- end}}
 
+{{- if ShouldConfigureCustomCATrust}}
+{{range $i, $cert := GetCustomCATrustConfigCerts}}
+- path: /usr/local/share/ca-certificates/00000000000000cert{{$i}}.crt
```
UtheMan, line 5404: To match the logic of binary used by pod that removes/updates certs, a prefix of length 14 is required. We want these initial certs to be removed when update comes, so for now i just added all zeros. 
Is there some more elegant way of doing that maybe?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -300,6 +300,16 @@ write_files:
     #EOF
 {{- end}}
 
+{{- if ShouldConfigureCustomCATrust}}
+{{range $i, $cert := GetCustomCATrustConfigCerts}}
+- path: /usr/local/share/ca-certificates/00000000000000cert{{$i}}.crt
```
alexeldeib, line 305: lol

I guess this is to match the regex in our update logic? can you leave a comment ?

{{/ /}} I think in go tmpl

UtheMan: Yeah added a comment in PR in generated file explaining that. This is so that these certs get removed once the pod/ds is running and update flow begins.

Will add comment in code too.

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -300,6 +300,16 @@ write_files:
...
+  permissions: "0644"
+  owner: root
+  content: |
+  {{$cert}}
```
alexeldeib, line 309:
```suggestion
  {{$cert}}
  #EOF
```

annoyingly this is how wait_for_file checks if cloud init finished writing it to disk

alexeldeib: also something seems a tiny bit of with the indentation but I can't tell what or if it will even matter - https://github.com/Azure/AgentBaker/blob/8166500194be7a250b2943a58c03a447309c4e98/pkg/agent/testdata/AKSUbuntu1804%2BCustomCATrust/CustomData#L92-L113

look at the previous file 99periodic vs yours

pkg/agent/baker.go
```diff
@@ -807,5 +807,18 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
 			return getOutBoundCmd(config, config.CloudSpecConfig)
 		},
+		"ShouldConfigureCustomCATrust": func() bool {
+			return config.CustomCATrustConfig != nil && len(config.CustomCATrustConfig.CustomCATrustCerts) > 0
+		},
+		"GetCustomCATrustConfigCerts": func() []string {
+			if config.CustomCATrustConfig != nil && config.CustomCATrustConfig.CustomCATrustCerts != nil {
+				indentedCerts := make([]string, 0, len(config.CustomCATrustConfig.CustomCATrustCerts))
+				for _, cert := range config.CustomCATrustConfig.CustomCATrustCerts {
+					indentedCerts = append(indentedCerts, datamodel.IndentString(cert, 4))
```
alexeldeib, line 817: oh that's why the indentation is off

pkg/agent/baker.go
```diff
@@ -807,5 +807,18 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetOutboundCommand": func() string {
 			return getOutBoundCmd(config, config.CloudSpecConfig)
 		},
+		"ShouldConfigureCustomCATrust": func() bool {
+			return config.CustomCATrustConfig != nil && len(config.CustomCATrustConfig.CustomCATrustCerts) > 0
+		},
+		"GetCustomCATrustConfigCerts": func() []string {
+			if config.CustomCATrustConfig != nil && config.CustomCATrustConfig.CustomCATrustCerts != nil {
```
alexeldeib, line 814: nit: if we didn't have a struct and only an array, could range directly. but maybe worth it for future changes...

parts/linux/cloud-init/nodecustomdata.yml
alexeldeib: ah my bad, read the pr sequentially and started commenting :D

pkg/agent/baker.go
UtheMan: Yeah wasn't sure what the indentation should be, I looked at how other GetHTTPProxyCA was implemented in baker.go and tried to do the same.

pkg/agent/baker.go
UtheMan: Im open to changing it to an array, at first I wanted to add additional stuff like timestamp in the struct, but decided to go with a prefix of zeros instead. What would you prefer?

pkg/templates/templates_generated.go
```diff
@@ -5399,6 +5399,18 @@ write_files:
...
 
+{{- if ShouldConfigureCustomCATrust}}
+{{range $i, $cert := GetCustomCATrustConfigCerts}}
+{{/* adding a prefix made of zeros to match removal logic used by custom ca trust pod, which handles old cert removal */}}
+- path: /usr/local/share/ca-certificates/00000000000000cert{{$i}}.crt
+  permissions: "0644"
+  owner: root
+  content: |
+    {{$cert}}
+  #EOF
```
alexeldeib, line 5410: I think you do actually want the EOF indented, otherwise it's going to count as a yaml key not part of multiline string

good thing to test e2e :)

PR #2051, opened 2022-08-03
containerized driver install for 470/510 branches

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -444,67 +444,28 @@ configAzurePolicyAddon() {
...
 configGPUDrivers() {
-    blacklistNouveau
-    addNvidiaAptRepo
-    installNvidiaContainerRuntime "${NVIDIA_CONTAINER_RUNTIME_VERSION}"
-    installNvidiaDocker "${NVIDIA_DOCKER_VERSION}"
 
-    # tidy
-    rm -rf $GPU_DEST/tmp
+    # install gpu driver
+    mkdir -p /opt/{actions,gpu}
```
alexeldeib, line 450: TODO: delete all of this, there should be exactly one pivot in cse_main for dedicated vs non dedicated vhd, this was just a dirty hack to try and test it quickly

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -444,67 +444,27 @@ configAzurePolicyAddon() {
...
-    blacklistNouveau
-    addNvidiaAptRepo
-    installNvidiaContainerRuntime "${NVIDIA_CONTAINER_RUNTIME_VERSION}"
-    installNvidiaDocker "${NVIDIA_DOCKER_VERSION}"
-
-    # tidy
-    rm -rf $GPU_DEST/tmp
+    # install gpu driver
+    mkdir -p /opt/{actions,gpu}
+    ctr image pull docker.io/alexeldeib/aks-gpu:latest
```
anujmaheshwari1, line 450: just for testing?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -195,28 +195,44 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
...
+      ctr run --privileged --net-host --with-ns pid:/proc/1/ns/pid --mount type=bind,src=/opt/gpu,dst=/mnt/gpu,options=rbind --mount type=bind,src=/opt/actions,dst=/mnt/actions,options=rbind docker.io/alexeldeib/aks-gpu:latest gpuinstall /entrypoint.sh install.sh
+      ret=$?
+      if [[ "$ret" != "0" ]]; then
+        echo "Failed to install GPU driver, exiting..."
+        exit $ret
+      fi
+    else
+      mkdir -p /opt/{actions,gpu}
+      ctr image pull docker.io/alexeldeib/aks-gpu:latest
+      ctr run --privileged --net-host --with-ns pid:/proc/1/ns/pid --mount type=bind,src=/opt/gpu,dst=/mnt/gpu,options=rbind --mount type=bind,src=/opt/actions,dst=/mnt/actions,options=rbind docker.io/alexeldeib/aks-gpu:latest gpuinstall /entrypoint.sh copy
```
chengliangli0918, line 212: just wondering whether this new approach works for FIPS+GPU?

currently for FIPS+GPU, at VHD build time, the build VM runs on non-fips kernel, but fips kernel is installed, so GPU driver built&installed are not for target fips kernel at this moment.

At AKS node provisioning time, the VM starts with FIPS kernel, the GPU driver is rebuilt and installed for the targeting fips kernel. In your new approach, will the gpu driver is built&installed again in CSE?

in cse_config.sh, the container image docker.io/alexeldeib/aks-gpu:latest is pulled and run, it looks FIPS+GPU should work as before, just want to confirm.

parts/linux/cloud-init/artifacts/cse_config.sh
alexeldeib: yeah was trying to get this into mcr, almost there

vhdbuilder/packer/install-dependencies.sh
alexeldeib: this is also something we could improve for all our builds -- rebooting before build completion to get the latest kernel updates.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -89,13 +89,15 @@ UBUNTU_OS_NAME="UBUNTU"
...
-export GPU_DV=470.57.02
+export GPU_DV="{{GPUDriverVersion}}"
 export GPU_DEST=/usr/local/nvidia
 NVIDIA_DOCKER_VERSION=2.8.0-1
 DOCKER_VERSION=1.13.1-1
-NVIDIA_CONTAINER_RUNTIME_VERSION=3.6.0
-NVIDIA_CONTAINER_TOOLKIT_VER=1.6.0
-NVIDIA_PACKAGES="libnvidia-container1 libnvidia-container-tools nvidia-container-toolkit"
+NVIDIA_CONTAINER_RUNTIME_VERSION="3.6.0"
+export NVIDIA_DRIVER_IMAGE_TAG="${GPU_DV}-v0.0.2-rev16-sha-05eb02"
```
alexeldeib, line 97: gross...but unique

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -89,13 +89,15 @@ UBUNTU_OS_NAME="UBUNTU"
...
+export NVIDIA_DRIVER_IMAGE="mcr.microsoft.com/aks/aks-gpu"
+export CTR_GPU_INSTALL_CMD="ctr run --privileged --rm --net-host --with-ns pid:/proc/1/ns/pid --mount type=bind,src=/opt/gpu,dst=/mnt/gpu,options=rbind --mount type=bind,src=/opt/actions,dst=/mnt/actions,options=rbind"
```
alexeldeib, line 99: this also got gross...

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -444,67 +444,37 @@ configAzurePolicyAddon() {
...
-    {{else}}
-    retrycmd_if_failure 120 5 25 pkill -SIGHUP dockerd || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
-    {{end}}
-
     # install gpu driver
-    setupGpuRunfileInstall
+    mkdir -p /opt/{actions,gpu}
+    if [[ "${CONTAINER_RUNTIME}" == "containerd" ]]; then
+        ctr image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG
+        bash -c "$CTR_GPU_INSTALL_CMD $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG gpuinstall /entrypoint.sh install"
```
alexeldeib, line 452: annoying but doing this to dynamically choose the right tag -_-

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -142,19 +139,6 @@ fi
 # If it is a MIG Node, enable mig-partition systemd service to create MIG instances
 if [[ "${MIG_NODE}" == "true" ]]; then
     REBOOTREQUIRED=true
-    STATUS=`systemctl is-active nvidia-fabricmanager`
```
alexeldeib, line 145: might need this back, at least `systemctlEnableAndStart nvidia-fabricmanager`

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -541,14 +511,9 @@ ensureGPUDrivers() {
     if [[ "${CONFIG_GPU_DRIVER_IF_NEEDED}" = true ]]; then
         configGPUDrivers
     else
-        # needs to happen even on gpu vhd because newer containerd/runc broke old 
-        # nvidia-container-runtime. containerd [1.5.9, 1.4.12] + runc 1.0.2 don't work with 
-        # old nvidia-container-runtime like 2.0.0, only new like 3.6.0
-        installNvidiaContainerRuntime "${NVIDIA_CONTAINER_RUNTIME_VERSION}"
-        installNvidiaDocker "${NVIDIA_DOCKER_VERSION}"
```
alexeldeib, line 548: need to bring back nvidia docker

alexeldeib: added

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -89,13 +89,15 @@ UBUNTU_OS_NAME="UBUNTU"
...
 DOCKER=/usr/bin/docker
-export GPU_DV=470.57.02
+export GPU_DV="{{GPUDriverVersion}}"
 export GPU_DEST=/usr/local/nvidia
 NVIDIA_DOCKER_VERSION=2.8.0-1
 DOCKER_VERSION=1.13.1-1
-NVIDIA_CONTAINER_RUNTIME_VERSION=3.6.0
-NVIDIA_CONTAINER_TOOLKIT_VER=1.6.0
-NVIDIA_PACKAGES="libnvidia-container1 libnvidia-container-tools nvidia-container-toolkit"
+NVIDIA_CONTAINER_RUNTIME_VERSION="3.6.0"
```
alexeldeib, line 96: todo: remove this in follow ups

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -194,28 +194,49 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
   done
 fi
 
+if [[ $OS == $UBUNTU_OS_NAME && $(isARM64) != 1 ]]; then  # no ARM64 SKU with GPU now
+  gpu_action="copy"
+  if grep -q "fullgpu" <<< "$FEATURE_FLAGS"; then
+    # force older driver on dedicated vhd to avoid build failures.
+    export NVIDIA_DRIVER_IMAGE_TAG="470.57.02-v0.0.2-rev17-sha-ff708e"
```
alexeldeib, line 201: gross but effective to keep gpu vhd pinned to old version

alexeldeib: I guess with this change we might (?) be able to bump the gpu vhd anyway, since it will be able to egress to MCR? not sure what RP e2e will do...

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -194,28 +194,49 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
...
-} >> ${VHD_LOGS_FILEPATH}
-if grep -q "fullgpu" <<< "$FEATURE_FLAGS"; then
-    echo "  - ensureGPUDrivers" >> ${VHD_LOGS_FILEPATH}
-    ensureGPUDrivers
-fi
-fi
+cat << EOF >> ${VHD_LOGS_FILEPATH}
+  - nvidia-docker2=${NVIDIA_DOCKER_VERSION}
+  - nvidia-container-runtime=${NVIDIA_CONTAINER_RUNTIME_VERSION}
+  - nvidia-gpu-driver-version=${GPU_DV}
```
ganeshkumar5699, line 237: nit: should this have the major.minor.patch from NVIDIA_DRIVER_IMAGE_TAG, instead, because of the switch between the two versions? Or maybe I'm missing something.

alexeldeib: yeah, I wasn't sure if it was better here to list the container image version or the underlying components. I think keeping it like this is okay, but it means we have to sync these vars in cse_helpers.sh with the GPU image tags. a bit annoying. 

might have the installer export a file with this info we could pick up at build time instead

parts/linux/cloud-init/artifacts/cse_helpers.sh
ganeshkumar5699: just to confirm, we would need to update this tag, each time we have a new release of the `aks-gpu` image right? That's probably not going to be that often.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -194,28 +194,49 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
   done
 fi
 
+if [[ $OS == $UBUNTU_OS_NAME && $(isARM64) != 1 ]]; then  # no ARM64 SKU with GPU now
+  gpu_action="copy"
+  if grep -q "fullgpu" <<< "$FEATURE_FLAGS"; then
+    # force older driver on dedicated vhd to avoid build failures.
+    export NVIDIA_DRIVER_IMAGE_TAG="470.57.02-v0.0.2-rev20-sha-a0bbd5"
```
ganeshkumar5699, line 201: So the dedicated VHD would now need the user to reinstall the drivers? Or it may not work for Nvidia V5 SKUs? I noticed a related comment in your email about this, and it seems like CSE is not dynamically updating the driver version for dedicated vhd.

alexeldeib: so right now this leaves the vhd at the 470 sku, but if compatible, it will used the 510 driver instead, which is most..

I should probably flip this since only ncv1 is compatible with 470 but not 510, but that happens to be what we run e2e with so I was worried about tests failing :D

maybe can just do that tho, and fix the tests

ganeshkumar5699: Thanks for explaining, that makes sense. 

Flipping to 510 as default for dedicated VHD, would also probably reduce the average install latency for GPU dedicated VHD across SKUs.

ganeshkumar5699 (approved):
Lgtm. This is a pretty elaborate change, and an elegant way to support both driver versions, depending on the SKU!

parts/linux/cloud-init/artifacts/cse_helpers.sh
alexeldeib: yes, hopefully not too often, also I can make the tag less obscenely long

vhdbuilder/packer/install-dependencies.sh
alexeldeib: aaand I can't because we don't have quota to build on other vm sizes -_-

could install but not validate them..

alexeldeib: trying to get more gpu quota to switch off ncv1

vhdbuilder/packer/install-dependencies.sh
alexeldeib: ran a fips test build for gpu and seemed to work okay https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=58954783&view=logs&j=2e57e679-818b-5c8d-ac48-4ec72e0831ac

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -286,90 +279,4 @@ ensureRunc() {
...
-    if [ -f "/etc/apt/sources.list.d/nvidia-docker.list" ]; then
-        echo "nvidia-docker.list already exists, no need to update"
-        return
-    fi
-    local release=$(lsb_release -r -s)
-    retrycmd_if_failure_no_stats 120 5 25 curl -fsSL https://nvidia.github.io/nvidia-docker/gpgkey | gpg --dearmor > /tmp/aptnvidia.gpg || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
-    wait_for_apt_locks
-    retrycmd_if_failure 10 5 10 cp /tmp/aptnvidia.gpg /etc/apt/trusted.gpg.d || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
-    wait_for_apt_locks
-    retrycmd_if_failure_no_stats 120 5 25 curl -fsSL https://nvidia.github.io/nvidia-docker/ubuntu${release}/nvidia-docker.list > /tmp/nvidia-docker.list || exit  $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
```
alexeldeib, line 312: ah...might still need to account for this -_-

alexeldeib: ${release} -- the base docker image for the installer should be distro specific.

alexeldeib: will follow up on this next week since we don't have any non-18.04 gpu skus yet

PR #2081, opened 2022-08-11
feat: Supports Windows agent pools disable OutBoundNAT

ShiqianTao:
**What type of PR is this?**
/kind feature

Support Windows agent pools disable OutBoundNAT

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/types.go
```diff
@@ -662,7 +662,8 @@ type AgentPoolProfile struct {
 	MessageOfTheDay       string               `json:"messageOfTheDay,omitempty"`
 	// This is a new property and all old agent pools do no have this field. We need to keep the default
 	// behavior to reboot Windows node when it is nil
-	NotRebootWindowsNode *bool `json:"notRebootWindowsNode,omitempty"`
+	NotRebootWindowsNode *bool                    `json:"notRebootWindowsNode,omitempty"`
+	WindowsProfile       *AgentPoolWindowsProfile `json:"agentPoolWindowsProfile,omitempty"`
```
AbelHu, line 666: use `agentPoolWindowsProfile` to differentiate from the property in cluster level?

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -93,6 +93,30 @@ function Set-AzureCNIConfig
...
+            Name = 'EndpointPolicy'
+            Value = $valueObj
+        }
+
+        # AdditionalArgs[0] is OutBoundNAT with ExceptionList; AdditionalArgs[1] is Route.
+        # Disable OutBoundNAT&Route if WinDSR is enabled. Otherwise, only disable OutBoundNAT and reserve Route.
+        if ($global:KubeproxyFeatureGates.Contains("WinDSR=true")) {
+            $configJson.plugins[0].AdditionalArgs = @($jsonContent)
+        } else {
+            $configJson.plugins[0].AdditionalArgs[0] = $jsonContent
```
AbelHu, line 116: It will remove other settings, for example, settings when IsDualStackEnabled is true

AbelHu: From the line 63-74, AdditionalArgs is always an array.
```
    $osBuildNumber = (get-wmiobject win32_operatingsystem).BuildNumber
    if ($osBuildNumber -le 17763){
        # In WS2019 and below rules in the exception list are generated by dropping the prefix lenght and removing duplicate rules.
        # If multiple execptions are specified with different ranges we should only include the broadest range for each address.
        # This issue has been addressed in 19h1+ builds

        $processedExceptions = GetBroadestRangesForEachAddress $exceptionAddresses
        Write-Log "Filtering CNI config exception list values to work around WS2019 issue processing rules. Original exception list: $exceptionAddresses, processed exception list: $processedExceptions"
        $configJson.plugins.AdditionalArgs[0].Value.ExceptionList = $processedExceptions
    }
    else {
        $configJson.plugins.AdditionalArgs[0].Value.ExceptionList = $exceptionAddresses
    }
```

pkg/agent/datamodel/types.go
AbelHu: I think that we can also use agentPoolWindowsProfile in API. If not, we can use `agentPoolWindowsProfile` in code but expose windowsProfile in API

ShiqianTao: Updated to AgentPoolWindowsProfile.

staging/cse/windows/azurecnifunc.ps1
ShiqianTao: Split this logic from this PR.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -184,6 +184,9 @@ $global:TLSBootstrapToken = "{{GetTLSBootstrapTokenForKubeConfig}}"
 
 $global:IsNotRebootWindowsNode = [System.Convert]::ToBoolean("{{GetVariable "isNotRebootWindowsNode" }}");
 
+# Disable OutBoundNAT in Azure CNI configuration
+$global:IsDisableWindowsOutboundNat = [System.Convert]::ToBoolean("{{GetVariable "isDisableWindowsOutboundNat" }}");
```
AbelHu, line 188: Should we set `$WindowsCSEScriptsPackage` to "aks-windows-cse-scripts-v0.0.14.zip" in this PR?

ShiqianTao: Updated. Thank you.

AbelHu (approved):

PR #2083, opened 2022-08-11
explicitly set further kubelet (node-status-report-frequency) config coming from RP to avoid inconsistent defaulting behavior

cameronmeissner:
**What type of PR is this?**
/kind bug

AgentBaker needs to be updated to accept the new configuration flags that will now be coming from RP as a result of the larger change to explicitly set as much kubelet config as possible. This should prevent unexpected kubelet behavior when clusters are eventually configured with a kubelet config file as opposed to only kubelet command line flags.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

This change hasn't been completely finished yet, as further work still needs to be done around adding specific logic for flags that are either deprecated from the command line to the config file or deprecated completely after certain k8s versions. I'm currently doing this research on all minor k8s versions between 1.13 and 1.24


pkg/agent/utils.go
```diff
@@ -276,7 +283,11 @@ func GetOrderedKubeletConfigFlagString(k map[string]string, cs *datamodel.Contai
 	keys := []string{}
 	for key := range k {
 		if !kubeletConfigFileEnabled || !TranslatedKubeletConfigFlags[key] {
-			keys = append(keys, key)
+			// --node-status-report-frequency is not a valid command line flag
+			// and should only be explicitly set in the config file
+			if key != nodeStatusReportFrequencyFlag {
```
alexeldeib, line 288: how about creating a new map of "this is not in config file"?

would be nice to avoid hardcoding single key names here

alexeldeib (approved):

PR #2095, opened 2022-08-15
Pablotrivino/apport fix

PabloTriv:
**What type of PR is this?**
/kind bug

Uninstalls apport on AKS host OS to prevent PII leaks to external entities.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -29,6 +29,10 @@ cleanupContainerdDlFiles() {
     rm -rf $CONTAINERD_DOWNLOADS_DIR
 }
 
+removeApport() {
+    sudo apt-get purge apport
```
alexeldeib, line 33: I would move this to https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/release-notes/AKSUbuntu/gen2/1804/2022.08.02.txt#L212-L213

similar to https://github.com/Azure/AgentBaker/blob/4da8de64fe542bd849a62cc7625e0b5fa487f00d/vhdbuilder/packer/install-dependencies.sh#L556-L562 and conditioned on ubuntu os

preferably before this line so the package output is right -__- https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/install-dependencies.sh#L518

PabloTriv:
Release notes no longer include apport.

https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=59247529&view=artifacts&pathAsName=false&type=publishedArtifacts

alexeldeib (approved):
lgtm!

PR #2104, opened 2022-08-20
setting up additional logging for steps within CSE

chmill-zz:
**What type of PR is this?**
/kind feature

This will allow logging to be scraped internally by azure to give us more insight into delays in the cse 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -23,6 +26,21 @@ JSON_STRING=$( jq -n \
 mkdir -p /var/log/azure/aks
 echo $JSON_STRING | tee /var/log/azure/aks/provision.json
 
+# arg names are defined by GA and all these are required to be correctly read by GA
+# EventPid, EventTid are required to be int. No use case for them at this point.
+EVENT_JSON=$( jq -n \
+    --arg Timestamp   "${CSE_STARTTIME}" \
+    --arg OperationId "${CSE_ENDTIME}" \
+    --arg Version     "1.23" \
```
chmill-zz, line 34: is there a smart way to get the version somewhere? I didn't spot it right away. cse_main is in the same boat where right now I just manually set, but not one wants that forever

PR #2111, opened 2022-08-22
Deeper logging cse

chmill-zz:
**What type of PR is this?**

/kind feature

This is going to allow internal tracing for various steps in the CSE to identity points of improvement. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -23,6 +26,21 @@ JSON_STRING=$( jq -n \
 mkdir -p /var/log/azure/aks
 echo $JSON_STRING | tee /var/log/azure/aks/provision.json
 
+# arg names are defined by GA and all these are required to be correctly read by GA
+# EventPid, EventTid are required to be int. No use case for them at this point.
+EVENT_JSON=$( jq -n \
+    --arg Timestamp   "${CSE_STARTTIME}" \
+    --arg OperationId "${CSE_ENDTIME}" \
+    --arg Version     "1.23" \
```
chmill-zz, line 34: Didn't spot version anywhere, but if it's stored somewhere, it'd be great to replace this here and in cse_main. Or just directly in the helper func

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -289,4 +290,28 @@ isARM64() {
         echo 0
     fi
 }
+
+logs_to_events() {
+    task=$1; shift
```
chmill-zz, line 295: do we see value in allowing the task name to always be defined. Right now all the cse_main references are AKS.CSE._func describe_

If we see value in being able to change AKS.CSE to whatever we want I can leave this as is. Or only accept a _func describe_ and have this logs_to_events append the _func describe_ to hard coded AKS.CSE in the func

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -1,12 +1,19 @@
 CSE_STARTTIME=$(date)
+CSE_STARTTIME_FORMATTED=$(date +"%F %T.%3N")
 timeout -k5s 15m /bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1
 EXIT_CODE=$?
 systemctl --no-pager -l status kubelet >> /var/log/azure/cluster-provision-cse-output.log 2>&1
 OUTPUT=$(tail -c 3000 "/var/log/azure/cluster-provision.log")
 KERNEL_STARTTIME=$(systemctl show -p KernelTimestamp | sed -e  "s/KernelTimestamp=//g" || true)
+KERNEL_STARTTIME_FORMATTED=$(echo ${KERNEL_STARTTIME} | xargs -i date -d {} +"%F %T.%3N" )
```
chmill-zz, line 8: this _FORMATTED stuff is to make the returned time into something the tools are happy to parse as a date. 

Before: Fri 2022-05-06 21:26:31 UTC
After: 2022-05-06 21:26:31.000

If we just want to keep the formatted variable and remake the existing to the formatted output I'll make that change. Not sure if there's any reason behind the existing date convention so left it alone for now.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -105,89 +105,89 @@ else
...
 {{- if HasDCSeriesSKU}}
 if [[ ${SGX_NODE} == true && ! -e "/dev/sgx" ]]; then
-    installSGXDrivers
+    logs_to_events "AKS.CSE.installSGXDrivers" installSGXDrivers
 fi
 {{end}}
 
 {{- if HasCustomSearchDomain}}
-wait_for_file 3600 1 {{GetCustomSearchDomainsCSEScriptFilepath}} || exit $ERR_FILE_WATCH_TIMEOUT
+logs_to_events "AKS.CSE.GetCustomSearchDomainsCSEScriptFilepath" "wait_for_file 3600 1 {{GetCustomSearchDomainsCSEScriptFilepath}} || exit $ERR_FILE_WATCH_TIMEOUT"
```
alexeldeib, line 168: does this still exit correctly? I think exit will work inside the func, but sanity check

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -196,25 +196,26 @@ if [[ "{{GetTargetEnvironment}}" == "AzureChinaCloud" ]]; then
...
+logs_to_events "AKS.CSE.configureSwapFile" configureSwapFile
 {{- end}}
 
-ensureSysctl
-ensureJournal
+logs_to_events "AKS.CSE.ensureSysctl" ensureSysctl
+logs_to_events "AKS.CSE.ensureJournal" ensureJournal
 {{- if IsKrustlet}}
-systemctlEnableAndStart krustlet
+logs_to_events "AKS.CSE.krustlet" systemctlEnableAndStart krustlet
```
alexeldeib, line 213:
```suggestion
logs_to_events "AKS.CSE.krustlet" "systemctlEnableAndStart krustlet"
```
i think?

parts/linux/cloud-init/artifacts/cse_main.sh
chmill-zz: I had the same question... I think based on how wait_for_file was used I should do this instead?

( exit outside the "" )
`logs_to_events "AKS.CSE.GetCustomSearchDomainsCSEScriptFilepath" "wait_for_file 3600 1 {{GetCustomSearchDomainsCSEScriptFilepath}}" || exit $ERR_FILE_WATCH_TIMEOUT`

parts/linux/cloud-init/artifacts/cse_start.sh
alexeldeib: xargs doesn't feel necessary here vs pipe alone but i'll allow it _shrugs_

alexeldeib, line 8:
```suggestion
KERNEL_STARTTIME_FORMATTED=$(date -d "${KERNEL_STARTTIME}" +"%F %T.%3N" )
```
I think

pkg/agent/datamodel/sig_config.go
```diff
@@ -1,503 +1,503 @@
...
+	return false
+}
+func (d Distro) IsGen2Distro() bool {
+	for _, distro := range AvailableGen2Distros {
+		if d == distro {
+			return true
+		}
+	}
+	return false
+}
```
alexeldeib, line 155: errrr did you format this whole file by accident?

pkg/agent/datamodel/types.go
```diff
@@ -1,1899 +1,1899 @@
...
-
-func NewError(code CSEStatusParsingErrorCode, message string) *CSEStatusParsingError {
-	return &CSEStatusParsingError{Code: code, Message: message}
-}
-
-func (err *CSEStatusParsingError) Error() string {
-	return fmt.Sprintf("CSE has invalid message=%q, InstanceErrorCode=%s", err.Message, err.Code)
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
```
alexeldeib, line 2: pls `git checkout master -- pkg/agent/datamodel`

guessing it's line endings or something but hard to eyeball this

parts/linux/cloud-init/artifacts/cse_start.sh
chmill-zz: ah i was doing `| date -d - xyz` before and failed. Turns out it failed because typo. I felt fancy with an excuse to xargs. But will drop

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -264,10 +264,10 @@ if ! [[ ${API_SERVER_NAME} =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
             VALIDATION_ERR=$ERR_K8S_API_SERVER_DNS_LOOKUP_FAIL
         fi
     else
-        logs_to_events "AKS.CSE.apiserverNC" "retrycmd_if_failure ${API_SERVER_CONN_RETRIES} 1 10 nc -vz ${API_SERVER_NAME} 443 || time nc -vz ${API_SERVER_NAME} 443 || VALIDATION_ERR=$ERR_K8S_API_SERVER_CONN_FAIL"
+        logs_to_events "AKS.CSE.apiserverNC" "retrycmd_if_failure ${API_SERVER_CONN_RETRIES} 1 10 nc -vz ${API_SERVER_NAME} 443 || time nc -vz ${API_SERVER_NAME} 443" || VALIDATION_ERR=$ERR_K8S_API_SERVER_CONN_FAIL
```
alexeldeib, line 267: I think this actually was right the first time, I think now if the cmd fails, the logs_to_events func will return success, so it won't exit for sure

if you can just do a sanity check in local bash like
```bash
function logs_to_events() { ... }
logs_to_events "false || exit 1"
echo "made it"
````
and make sure it actually kills the shell (i.e., never reaches "made it"), that's fine for me.

File: pkg/agent/datamodel/types.go (hunk 7 above)
Comment (alexeldeib) [reply pkg/agent/datamodel/types.go]: not resolved?

File: parts/linux/cloud-init/artifacts/cse_main.sh (hunk 4 above)
Comment (alexeldeib) [reply parts/linux/cloud-init/artifacts/cse_main.sh]: oops commented elsewhere -- I think you need to execute the whole command, but I wasn't sure if double interpolated vars would work correctly. we can just run with it.

File: pkg/agent/datamodel/types.go (hunk 7 above)
Comment (chmill-zz) [reply pkg/agent/datamodel/types.go]: well... That is confusing. When I look at https://github.com/Azure/AgentBaker/pull/2111/files _pkg/agent/datamodel_ doesn't show up. But it's in the commit...

And git checkout master -- pkg/agent/datamodel/types.go after making sure master is current doesn't pick up a change

This doesn't spot it either... git diff --name-status master..deeperLoggingCSE

Comment (chmill-zz) [reply pkg/agent/datamodel/types.go]: not confident if we want to continue to argue with git on this, or start a fresh pr and avoid this :\

File: parts/linux/cloud-init/artifacts/cse_helpers.sh [hunk 9]
```diff
@@ -96,12 +96,13 @@ export GPU_DEST=/usr/local/nvidia
 NVIDIA_DOCKER_VERSION=2.8.0-1
 DOCKER_VERSION=1.13.1-1
 NVIDIA_CONTAINER_RUNTIME_VERSION="3.6.0"
-export NVIDIA_DRIVER_IMAGE_TAG="${GPU_DV}-v0.0.2-rev20-sha-a0bbd5"
```
alexeldeib, line 99: why is this diff here? still missing a rebase? I can try and fix this if you want

alexeldeib: master has what's on the righ which is why it seems like some weird history issue

PR #2121, opened 2022-08-25

HDYA:


AbelHu:
@HDYA does cve impact Windows?

HDYA:
this PR is generated based on azcu API for latest patch release
does Windows share the same component list? @AbelHu

AbelHu:
@HDYA if cve only impacts images, it should not impact windows since windows uses kubelet and kubeproxy in k8s binaries. We cache k8s binaries for Windows in [generate-windows-vhd-configuration.ps1](https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/generate-windows-vhd-configuration.ps1#L171-L194)

andyliuliming (approved):

PR #2122, opened 2022-08-25
add moby-cli and moby-engine to containerd vhd, no longer build docker vhds

cameronmeissner:
**What type of PR is this?**

This PR brings in changes to the VHD build process. We're now installing moby-cli and moby-engine when building containerd VHDs so we no longer need to build independent docker-based VHDs, which are no longer being built also as a part of these changes.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -59,7 +59,18 @@ installStandaloneContainerd() {
...
+}
 
+installMissingMobyComponentsForContainerd() {
+    # install moby components common between containerd and docker
+    local MOBY_VERSION="19.03.14"
+    MOBY_CLI=${MOBY_VERSION}
+    if [[ "${MOBY_CLI}" == "3.0.4" ]]; then
+        MOBY_CLI="3.0.3"
+    fi
+    echo "Installing moby-engine version ${MOBY_VERSION}, moby-cli version ${MOBY_CLI}"
```
alexeldeib, line 72: idk if this will work...I don't think we ever supported docker + mariner

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -59,7 +59,18 @@ installStandaloneContainerd() {
...
     fi
+    installMissingMobyComponentsForContainerd
+}
 
+installMissingMobyComponentsForContainerd() {
+    # install moby components common between containerd and docker
+    local MOBY_VERSION="19.03.14"
+    MOBY_CLI=${MOBY_VERSION}
+    if [[ "${MOBY_CLI}" == "3.0.4" ]]; then
+        MOBY_CLI="3.0.3"
```
alexeldeib, line 70: err what?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -267,16 +299,82 @@ ensureRunc() {
...
-    if [ -f $VHD_LOGS_FILEPATH ]; then
-        RUNC_DEB_PATTERN="moby-runc_${TARGET_VERSION/-/\~}+azure-*_${CPU_ARCH}.deb"
+
+    RUNC_DEB_PATTERN="moby-runc_${TARGET_VERSION/-/\~}+azure-*_${CPU_ARCH}.deb"
+    RUNC_DEB_FILE=$(find ${RUNC_DOWNLOADS_DIR} -type f -iname "${RUNC_DEB_PATTERN}" | sort -V | tail -n1)
+    if [[ -f "${RUNC_DEB_FILE}" ]]; then
+        installDebPackageFromFile ${RUNC_DEB_FILE} || exit $ERR_RUNC_INSTALL_TIMEOUT
+    else
+        downloadRuncFromVersionAndCPUArch ${TARGET_VERSION/-/\~} $CPU_ARCH
         RUNC_DEB_FILE=$(find ${RUNC_DOWNLOADS_DIR} -type f -iname "${RUNC_DEB_PATTERN}" | sort -V | tail -n1)
```
alexeldeib, line 309: dupe of line 304 right? can we invert this? check for the file, if not present download or fail?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
alexeldeib: also don't think we really need this for mariner, it is mostly to avoid breaking older ubuntu users, you can no op it or fail as unimplemented

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -94,6 +95,8 @@ KUBECTL=/usr/local/bin/kubectl
 DOCKER=/usr/bin/docker
 export GPU_DV="{{GPUDriverVersion}}"
 export GPU_DEST=/usr/local/nvidia
+MOBY_VERSION=20.10.15
```
cameronmeissner, line 98: I believe this is where we should be doing the docker bump.

cameronmeissner (commented):
I'm also wondering: can we get away with removing the cleanupContainerd call within cse_main on L76 at this point?

parts/linux/cloud-init/artifacts/cse_helpers.sh
alexeldeib: yeah maybe let's just bump to 20.10.17 if that works for all distros?

alexeldeib:
should update this bit to force both? https://github.com/Azure/AgentBaker/blob/0b9d929a71b540a72040e69534bbf6503cf99f96/parts/linux/cloud-init/artifacts/cse_main.sh#L181-L185

cameronmeissner:
> should update this bit to force both?
> 
> https://github.com/Azure/AgentBaker/blob/0b9d929a71b540a72040e69534bbf6503cf99f96/parts/linux/cloud-init/artifacts/cse_main.sh#L181-L185

I figured these two would clash if they both ran, I guess that's not the case?

alexeldeib:
right now they disable each other, I think we can just merge the logic. 

as-is, when we boot up the containerd vhd, dockerd will get disabled, so we can't share the vhd :D

alexeldeib:
> as-is, when we boot up the containerd vhd, dockerd will get disabled, so we can't share the vhd :D

I guess that's not true since it will come from RP, so maybe this is fine...let me marinate on that

alexeldeib:
so...installStandaloneContainerd will install it, but then ensureContainerd would remove it for docker vhd...
ensure docker would remove containerd tho...actually not even sure how that works right now? I suppose docker might be running its own containerd daemon, not a separate unit. 

I think if we actually want to use the containerd unit we install, we should always manually be installing both containerd and docker like in installStandaloneContainerd, and ensureDocker/ensureContainerd should not remove their counterparts?

but then it might make sense to merge a bit more code...

PR #2153, opened 2022-09-05
fix: Build Windows 2022 Containerd Gen 2 VHD

wanqingfu:
**What type of PR is this?**
Fix the issues of Gen 2 VHD building.

1. Gen 1 and Gen 2 VHD sig destinations are not differentiated.

2. Gen 2 VHD build succeeds but test failed.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -15,6 +15,7 @@ steps:
 #               we will hard-code the task variables SIG_GALLERY_NAME, SIG_IMAGE_NAME, and SIG_IMAGE_VERSION.
 #               Task variable GEN2_SIG_FOR_PRODUCTION is set to True and passed to the following steps.
 #               Built sig will be deleted because it has been converted to VHD, and thus not needed.
+#               Use the suffix $RANDOM to avoid duplication while running concurrent builds.
   - bash: |
       if [[ $(HYPERV_GENERATION) == "V2" ]]; then \
         m="sigMode"; \
@@ -22,8 +23,8 @@ steps:
           sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
           echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
         else
-          sigImageName="windows-${WINDOWS_SKU}"; \
-          sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d")"; \          
+          sigImageName="windows-${WINDOWS_SKU}-gen2-$RANDOM"; \
+          sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d%H%M%S")$RANDOM"; \          
           echo "##vso[task.setvariable variable=SIG_GALLERY_NAME]$sigGalleryName"; \
           echo "##vso[task.setvariable variable=SIG_IMAGE_VERSION]1.0.0"; \
           echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]True"; \
@@ -135,25 +136,26 @@ steps:
 
 # Moved conversion to VHD before cleanup.
 # Gen 2 packer outputs a sig in destination. This step: dest sig=>disk=>VHD in classic SA for publishing.
+# Credentials and resource group name come from the BUILD_**** pipeline variables because source sig is in the build subscription.
   - bash: |
         echo MODE=$(MODE) && \
-        create_time="$(cat vhdbuilder/packer/settings.json | grep "create_time" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
         SIG_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "sig_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
         gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
         docker run --rm \
         -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
         -w /go/src/github.com/Azure/AgentBaker \
-        -e CLIENT_ID=${CLIENT_ID} \
-        -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-        -e TENANT_ID=${TENANT_ID} \
-        -e SUBSCRIPTION_ID="${SUBSCRIPTION_ID}" \
-        -e RESOURCE_GROUP_NAME="${AZURE_RESOURCE_GROUP_NAME}" \
-        -e CREATE_TIME="${create_time}" \
-        -e LOCATION="${AZURE_LOCATION}" \
+        -e CLIENT_ID=${BUILD_CLIENT_ID} \
+        -e CLIENT_SECRET="$(BUILD_CLIENT_SECRET)" \
+        -e TENANT_ID=${BUILD_TENANT_ID} \
+        -e SUBSCRIPTION_ID="${BUILD_SUBSCRIPTION_ID}" \
+        -e RESOURCE_GROUP_NAME="${BUILD_AZURE_RESOURCE_GROUP_NAME}" \
+        -e LOCATION="${BUILD_AZURE_LOCATION}" \
         -e OS_TYPE="Windows" \
-        -e CLASSIC_BLOB=${STORAGE_ACCT_BLOB_URL} \
         -e CLASSIC_SAS_TOKEN="$(STORAGE_ACCT_SAS_TOKEN)" \
+        -e CLASSIC_BLOB="$(STORAGE_ACCT_BLOB_URL)" \
         -e SIG_IMAGE_NAME=${SIG_IMAGE_NAME} \
+        -e SIG_GALLERY_NAME=${SIG_GALLERY_NAME} \
+        -e SIG_IMAGE_VERSION=${SIG_IMAGE_VERSION} \
         -e GEN2_CAPTURED_SIG_VERSION=${gen2_captured_sig_version} \
         ${CONTAINER_IMAGE} make -f packer.mk convert-sig-to-classic-storage-account-blob
     displayName: Convert Shared Image Gallery To VHD Blob In Classic Storage Account
--- a/vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
+++ b/vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
@@ -18,7 +18,20 @@ do
     fi
 done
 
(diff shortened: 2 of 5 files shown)


vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -7,6 +7,7 @@ required_env_vars=(
     "LOCATION"
     "OS_TYPE"
     "SIG_IMAGE_NAME"
+    "SIG_GALLERY_NAME"
```
AbelHu, line 10: Will it block Linux?

wanqingfu: Maybe?

AbelHu: cc @ganeshkumar5699 for the change in the shared code.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -18,7 +18,11 @@ do
     fi
 done
 
-sig_resource_id="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/${SIG_IMAGE_NAME}/versions/${GEN2_CAPTURED_SIG_VERSION}"
+if [ -z ${SIG_GALLERY_NAME} ]; then
```
AbelHu, line 21: Please use the same validated code in cleanup.sh
```
if [[ -z "$SIG_GALLERY_NAME" ]]; then
  SIG_GALLERY_NAME="PackerSigGalleryEastUS"
fi
```

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
wanqingfu: [https://github.com/Azure/AgentBaker/pull/2153/commits/c4f63ffd1705520c63e38a96b4546306bc8b93de](url)

**Context**: Changed gallery name in `sig_resource_id` from `PackerSigGalleryEastUS` to `SIG_GALLERY_NAME` because windows might need a different gallery destination for Gen 2 VHD.

**Sanity check**: Linux does not pass `SIG_GALLERY_NAME` to this file. So if `SIG_GALLERY_NAME` is empty, this file will set its value to `PackerSigGalleryEastUS` as in `cleanup.sh`.

AbelHu (approved):

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -22,8 +22,8 @@ steps:
           sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
           echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
         else
-          sigImageName="windows-${WINDOWS_SKU}"; \
-          sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d")"; \          
+          sigImageName="windows-${WINDOWS_SKU}-gen2"; \
+          sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d%m%H%M")"; \
```
AbelHu, line 26: Please add %s to avoid the confliction

AbelHu: And please publish the pr for code review

wanqingfu: Adding %S so that gallery name looks like "WS2022Gen2Gallery220909022841" instead of "WS2022Gen2Gallery22090902291662690573".

AbelHu: Will it not fail when two PRs are submitted at the same time in seconds?

wanqingfu: Sorry are you suggesting adding more % formatting suffix to the gallery name? I think it's highly unlikely that two PRs are simultaneously submitted either manually or automatically. Besides, there's some time from the time of submission untill the code `sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d%H%M%S")"; ` gets executed?

AbelHu: %S may be enough, but I really do not want to waste time to investigate the failures for the same sigGalleryName

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
AbelHu: It is not forcedly applied the same format in all files in AgentBaker for now.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -15,15 +15,16 @@ steps:
...
       if [[ $(HYPERV_GENERATION) == "V2" ]]; then \
         m="sigMode"; \
         if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then \   
           sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
           echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
         else
-          sigImageName="windows-${WINDOWS_SKU}"; \
-          sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d")"; \          
+          sigImageName="windows-${WINDOWS_SKU}-gen2-$RANDOM"; \
+          sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d%H%M%S")$RANDOM"; \
```
AbelHu, line 27: Why do you add $RANDOM back? Will it not fail when using $RANDOM?

.pipelines/templates/.builder-release-template-windows.yaml
AbelHu: > %S may be enough, but I really do not want to waste time to investigate the failures for the same sigGalleryName

I prefere to use %s but not %S for this reason.

wanqingfu: True. It seems like the packer error message was not concrete or sensible enough. "Do not rat hole a problem. If something fundamental issues found in the"

According to [this bug](https://github.com/hashicorp/packer/issues/8653)
"Our issue ended up being that our concurrent builds were sharing a managed_image_name and shared_image_gallery_destination.image_version".

So changing the sig gallery name and sig image name to unique values in [this](https://github.com/Azure/AgentBaker/pull/2153/commits/708693f780a58e1337aed3f82ed27fdcb913dacc).

https://github.com/hashicorp/packer-plugin-azure/issues/41

AbelHu: I think that https://github.com/hashicorp/packer/issues/8653 may not be the root cause since you have use $random to generate different galleryname in your failed pipelines.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Based on the packer github issue, the same managed image name `windows-$WINDOWS_SKU-gen2` will also fail. I was just trying to see if under absolutely different `sigImageName` and `sigGalleryName`, will concurrent builds fail again.

Even if the gallery name contains $RANDOM as before, I suspect the same `sigImageName`: `windows-2022-containerd-gen2` might also trigger the error in Packer.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -122,10 +123,10 @@ steps:
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
-      -e CLIENT_ID=${CLIENT_ID} \
-      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-      -e TENANT_ID=${TENANT_ID} \
-      -e SUBSCRIPTION_ID="${SUBSCRIPTION_ID}" \
+      -e CLIENT_ID=${BUILD_CLIENT_ID} \
```
AbelHu, line 126: Please use CLIENT_ID but not BUILD_CLIENT_ID since it is used to logon the VHD destination

AbelHu (changes requested):
Please update the PR to use the correct sp

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Tested three times in the PR check-in pipeline. I think that post is trying to say it might be caused by the same managed_image_name and image version. Managed image name, in our case, is `sigImageName="windows-${WINDOWS_SKU}-gen2"`. So to avoid confusion, $RANDOM is added at the back of both image name and gallery name.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Replied above.

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Replied you in a separate email. Just for the record (for the future audience), this is a part where we've been changing back and forth for some time. Needed some concrete and well-organized clarifications.

wanqingfu:
> Please update the PR to use the correct sp

Needed some more fundamental understanding on the correct sp for az-copy, conversion, and generate-sas. Since the production pipeline is not working under CLIENT_ID etc., I suggest we hold onto this before rushing into changes that won't work either. Being brash isn't helpful for both the author and the reviewer regarding this issue, for Linux only has one set of credentials where Windows has separate ones for PR check-in and the production counterpart. I do think it's confusing.

wanqingfu:
> Please update the PR to use the correct sp

I am using the "_correct?_" SP and testing four cases in PR check-in. For the prod. pipeline, it seemed to work twice - see the latest two runs. Would you be so kind to take a look at the code and pipeline logs?

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
wanqingfu: Could we close this?

AbelHu: Yes since you have fixed it

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -115,6 +116,7 @@ steps:
         targetPath: 'release-notes.txt'
 
 # Default Gen 1: packer VHD=>VHD in classic SA
+# Credentials are without the BUILD prefix because we need AME account to copy to the classic SA
```
AbelHu, line 119: Actually, we do not need any credential for copy since both src and dest have the sas token. I think that the credential is passed since it is needed for az login and it may be used. But for now, we do not use it.

wanqingfu: Will remove it.

AbelHu (approved):

PR #2156, opened 2022-09-06
Update log symlink logic to check for file existence

phealy:
**What type of PR is this?**
/kind bug

Fixes the logic on symlinking log files to check that they exist first, so that we won't bomb out if one of the files isn't there.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner (approved):
lgtm, I guess there's multiple edge cases this protects against.

alexeldeib:
I think we determined this wasn't necessary?

PR #2165, opened 2022-09-07
freezing docker images at current versions

cameronmeissner:
**What type of PR is this?**

Freezes all docker images at their current version, as we'll no longer be building new docker VHDs. Moving forward, customers will need to be using containerd instead.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/agent/datamodel/sig_config.go
+++ b/pkg/agent/datamodel/sig_config.go
@@ -245,31 +245,33 @@ var (
 		Definition:    "1604",
 		Version:       "2021.11.06",
 	}
+
 	SIGUbuntu1804ImageConfigTemplate = SigImageConfigTemplate{
 		ResourceGroup: AKSUbuntuResourceGroup,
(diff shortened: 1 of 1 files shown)


pkg/agent/datamodel/sig_config.go
```diff
@@ -245,31 +245,33 @@ var (
 		Definition:    "1604",
 		Version:       "2021.11.06",
 	}
+
 	SIGUbuntu1804ImageConfigTemplate = SigImageConfigTemplate{
 		ResourceGroup: AKSUbuntuResourceGroup,
 		Gallery:       AKSUbuntuGalleryName,
 		Definition:    "1804",
-		Version:       LinuxSIGImageVersion,
+		Version:       "2022.08.29",
```
alexeldeib, line 253: nit: shared const since they are all the same?

alexeldeib (approved):

PR #2170, opened 2022-09-08
set kernel.pid_max to 4194304 (PID_MAX_LIMIT on 64-bit) to avoid node instability

cameronmeissner:
**What type of PR is this?**

To avoid customer-reported node instability, we should be setting kernel.pid_max to the maximum value on 64 bit Linux distros (2^22). Increasing this value is also recommended by the k8s documentation: https://kubernetes.io/docs/concepts/policy/pid-limiting/

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests


alexeldeib:
why 2^22?

it's a nice number, but curious :)

alexeldeib:
I guess it's max (I thought it was different) but it'd be nice for posterity to include it either in the PR message or commit message ;)

cameronmeissner:
> 

Found this: https://unix.stackexchange.com/questions/162104/how-to-change-the-kernel-max-pid-number.

Looks like 2^22 is the max value for 64-bit platforms.

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -765,6 +765,8 @@ write_files:
     net.ipv4.tcp_retries2=8
     net.core.message_burst=80
     net.core.message_cost=40
+    # To avoid node instability
+    kernel.pid_max=4194304
```
alexeldeib, line 769: can we move this to a drop in and add at build time? similar to https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/sysctl-d-60-CIS.conf

cameronmeissner: for sure, are these drop ins more "hard-codey" type settings that we use for safety?

cameronmeissner: nvm, did some research and makes much more sense now :)

alexeldeib: yeah mostly it's nice to put as much static stuff as possible, we could certainly improve that tbh

alexeldeib (approved):
would be nice if we could sanity test this by booting a VM from test VHD and ensuring 1) we can fork bomb to above the default max, and 2) see what happens (I assume failed fork()) when we hit the new limit?

PR #2171, opened 2022-09-08
fix AgentBaker E2E test failures due to clusters in failed state

cameronmeissner:
**What type of PR is this?**

To fix transient failures within the AgentBaker E2E test pipeline due to the GC not properly deleting existing cluster resources. This PR adds a check to the pipeline step that performs cluster creation - we now check to see if the existing cluster also has a provisioning state of "Succeeded" before attempting to reuse it for testing.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests


e2e/e2e-starter.sh
```diff
@@ -21,24 +31,31 @@ if [ -n "$out" ]; then
...
-        az aks delete -n $CLUSTER_NAME -g $RESOURCE_GROUP_NAME --yes
-        clusterDeleteEndTime=$(date +%s)
-        log "Deleted cluster in $((clusterDeleteEndTime-clusterDeleteStartTime)) seconds"
+        # The cluster exists, but has no associated MC resource group
+        log "Cluster $CLUSTER_NAME exists but has no associated MC resource group"
+        deleteCluster $CLUSTER_NAME $RESOURCE_GROUP_NAME
         create_cluster="true"
+    else
+        # The cluster exists and has an associated MC resource group
+        provisioning_state=$(az aks show -n $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -ojson | jq '.provisioningState' | tr -d "\"")
```
anujmaheshwari1, line 40: nit for refactor: we can probably move this check for provisioning state outside, as soon as we check if the cluster exists or not. 

following that, 
if exists == false or provisioning state == failed
    delete cluster
fi

saves the redundant calls from L35-37 but can follow up later.

anujmaheshwari1 (approved):

e2e/e2e-starter.sh
cameronmeissner: yep, I think I originally planned on going down this path, but figured we could avoid an unnecessary call to "aks show" - although I think the readability benefits/code reduction do trump that :)

PR #2175, opened 2022-09-11
allow us to disable unattended upgrades for ubuntu

paulgmiller:
**What type of PR is this?**

/kind feature
Unattended upgrade can be harmful and redundant if you have node image uprgade. 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -316,6 +316,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"Disable1804SystemdResolved": func() bool {
 			return config.Disable1804SystemdResolved
 		},
+		"EnableUnattendedUpgrade": func() bool {
+			return !config.DisableUnattendedUpgrades
```
yewmsft, line 320: nit: shall we call it enable to avoid the double negative?

paulgmiller: default for a bool is false though and for now the default needs to be unattended upgrades enabled.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -227,7 +227,9 @@ if $FULL_INSTALL_REQUIRED; then
 fi
 
 {{- /* re-enable unattended upgrades */}}
+{{- if EnableUnattendedUpgrade }} 
 rm -f /etc/apt/apt.conf.d/99periodic
```
alexeldeib, line 231: how about we remove this, and then in cloud init, we set the correct value (enabled/disabled) based on the bool?

alexeldeib (approved):
lgtm. long term i'd prefer disabling in the base image and consume user data as an enable/disable mechanism but that's probably not happening soon. 

actually would be nice to disable in the base image and enable when required just to avoid races...but still fine either way

pkg/agent/baker.go
paulgmiller: with new refactor at least go rid of the not

PR #2177, opened 2022-09-12
Add k8s 1.25.0 binaries to VHD.

yizhang4321:
**What type of PR is this?**

/kind feature

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -495,6 +495,7 @@ MULTI_ARCH_KUBE_BINARY_VERSIONS="
 1.23.8-hotfix.20220620
 1.24.0-hotfix.20220615
 1.24.3
+1.25.0
 "
```
ganeshkumar5699, line 500: `crictl` binary for k8s 1.25.0 needs to be cached in the VHD as well. 
The `1.25.0` version needs to be added in line 150 of `install-dependencies.sh` and it needs to be added to crictl version in components.json as well (line 469).

ganeshkumar5699: Other parts of the PR look good

ganeshkumar5699: And we also want to confirm that the VHD builds succeed (might need a retry here).

yizhang4321: crictl 1.25 added

ganeshkumar5699:
The VHD build tests fail during the download step of the "Download Asc baseline". Creating the PR from a branch in Azure/Agentbaker directly, instead of a fork, should make the tests run.

alexeldeib:
@ganeshkumar5699 assuming we can close since we shipped 1.25.2?

PR #2185, opened 2022-09-13
Add custom ca trust v2 changes (datamodel + cse)

UtheMan:
**What type of PR is this?**

Adding datamodel/CSE changes for custom ca v2

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -826,6 +826,19 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return 0
 		},
+		"ShouldConfigureCustomCATrust": func() bool {
+			return config.CustomCATrustConfig != nil && len(config.CustomCATrustConfig.CustomCATrustCerts) > 0
```
cameronmeissner, line 830: nit: consistency on the nil/empty check on CustomCATrustCerts. apparently len(<nil slice>) will return 0,  so len(<slice>) == 0 and <slice> == nil actually are the same.

cameronmeissner: could also refactor that check out to its own function and reuse within "GetCustomCATrustConfigCerts"

pkg/agent/baker.go
```diff
@@ -826,6 +826,19 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+		},
+		"GetCustomCATrustConfigCerts": func() []string {
+			if config.CustomCATrustConfig != nil && config.CustomCATrustConfig.CustomCATrustCerts != nil {
+				indentedCerts := make([]string, 0, len(config.CustomCATrustConfig.CustomCATrustCerts))
+				for _, cert := range config.CustomCATrustConfig.CustomCATrustCerts {
+					indentedCerts = append(indentedCerts, datamodel.IndentString(cert, 4))
+				}
+				return config.CustomCATrustConfig.CustomCATrustCerts
```
cameronmeissner, line 838: I might be missing something, but do you mean to be returning indentedCerts here?

alexeldeib: ironically the output looks good in the generated yaml so I think the indentation may not be necessary?

UtheMan: Yeah just tested, i think it really is not needed. Will remove indentation

pkg/agent/baker.go
```diff
@@ -826,10 +826,23 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
 	}
 }
 
 func isStandardNCv1(size string) bool {
 	tmp := strings.ToLower(size)
 	return strings.HasPrefix(tmp, "standard_nc") && !strings.Contains(tmp, "_v")
 }
+
+func areCustomCATrustCertsPopulated(config datamodel.NodeBootstrappingConfiguration) bool {
+	return config.CustomCATrustConfig != nil && config.CustomCATrustConfig.CustomCATrustCerts != nil
```
alexeldeib, line 847: nit
```suggestion
	return config.CustomCATrustConfig != nil && len(config.CustomCATrustConfig.CustomCATrustCerts) > 0
```

https://go.dev/play/p/y_7HlM92Phr

alexeldeib: don't know if we validate the empty array on rp side or not

alexeldeib (approved):
lgtm with one comment

pkg/agent/baker.go
alexeldeib: lol I see cameron left the opposite review, but I think keeping len > 0 and skipping the nil check is correct?

UtheMan: On rp side empty array will be treated as a valid input, but on AB side we don't want to do anything with that. I think both of these will work. Discussed on the side with Cameron that switching from len to nil check there might be less readable so I'm okay with letting it stay like it was originally.

cameronmeissner: yep, either works fine, I was mainly just thinking we should be consistent in what we choose there :)

PR #2201, opened 2022-09-16
systemd service for ci-syslog integration

pfrcks:
**What type of PR is this?**

/kind feature

Add systemd service to process ci syslog changes, adding and removing ci-syslog config and restarting rsyslog

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/ci-syslog-watcher.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Restarts ci-syslog service
+
+[Service]
+Type=oneshot
+ExecStart=/bin/systemctl restart ci-syslog.service
```
alexeldeib, line 6: is there a reason to do this?

make ci-syslog a oneshot and move the path trigger to the same service?

parts/linux/cloud-init/artifacts/ci-syslog.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
alexeldeib, line 1: you need to add all of the files in VHD to https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/vhd-image-builder-gen2.json for each sku you care about (likely all of them, except windows)

PR #2210, opened 2022-09-19
bump nvidia version to prep for grid drivers

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -197,7 +197,7 @@ fi
 
 if [[ $OS == $UBUNTU_OS_NAME && $(isARM64) != 1 ]]; then  # no ARM64 SKU with GPU now
   gpu_action="copy"
-  export NVIDIA_DRIVER_IMAGE_TAG="510.47.03-sha-8e58eb"
```
ganeshkumar5699, line 200: nit (perhaps for future): can we use `NVIDIA_DRIVER_IMAGE_TAG` to avoid copy-pasting the SHA twice?

ganeshkumar5699 (approved):

vhdbuilder/packer/install-dependencies.sh
alexeldeib: open to a suggestion, the annoying part is we have to change `NVIDIA_DRIVER_IMAGE_TAG` in CSE depending on VM size, so it can't be static :/

I suppose we at least split out the sha...

alexeldeib: better?

anujmaheshwari1 (approved):
LGTM!

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: yep, I was thinking of something similar! Looks good, thanks!

PR #2214, opened 2022-09-19
add: support for nvidia grid drivers

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/const.go
```diff
@@ -114,3 +114,18 @@ const (
...
 )
+
+const (
+	Nvidia470CudaDriverVersion = "cuda-470.82.01"
+	Nvidia510CudaDriverVersion = "cuda-510.47.03"
+	Nvidia510GridDriverVersion = "grid-510.73.08"
+)
+
+var GridGPUSizes = map[string]bool{
+	"standard_nv6ads_a10_v5":   true,
```
anujmaheshwari1, line 125: are these only 6? i.e do they need to be hardcoded? as and when grid GPU sizes are added, they need to be added here?

anujmaheshwari1 (approved):

pkg/agent/datamodel/const.go
alexeldeib: discussed offline. yes we will need to update it but there's no mechanism for discovery currently, so this is all we can do.

PR #2221, opened 2022-09-20
Add CI-Syslog systemd service

pfrcks:
**What type of PR is this?**
/kind feature

Adds a systemd service which tracks CI-Syslog enable/disable and correspondingly updates rsyslog config and restarts rsyslog

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
```diff
@@ -0,0 +1,29 @@
...
+        cat >/etc/rsyslog.d/60-rsyslog-forward-mdsd.conf << 'EOL'
+$ModLoad omuxsock 
+$OMUxSockSocket /var/run/mdsd/default_syslog.socket 
+template(name="MDSD_RSYSLOG_TraditionalForwardFormat" type="string" string="<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag%%msg:::sp-if-no-1st-sp%%msg%")
+$OMUxSockDefaultTemplate MDSD_RSYSLOG_TraditionalForwardFormat
+*.* :omuxsock:
+EOL
+elif [[ "$status" == "remove" ]]; then
+        echo "Status changed to $status."
+        [ -f "/etc/rsyslog.d/60-rsyslog-forward-mdsd.conf" ] && rm /etc/rsyslog.d/60-rsyslog-forward-mdsd.conf
```
alexeldeib, line 20: dumb question, have you tried this with mariner // does Azure Monitor support mariner?

you've added it to all the mariner vhds as well, please either confirm it works there, or remove it from those files. As we open up mariner to 3P we should probably doc the limitation if it exists.

alexeldeib:
mostly lgtm, would appreciate clarification on the mariner support, also needs a rebase/merge from master

parts/linux/cloud-init/artifacts/ci-syslog-watcher.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Update CI syslog config based on status change
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/ci-syslog-watcher.sh
```
alexeldeib, line 6: nit: do you care about Restart or RestartSec? default I think is no restart so would trigger on next path change, if you do Restart=on-failure you probably want to increase RestartSec to avoid spamming it? minor stuff tho

parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
ganga1980, line 20: will this path "/etc/rsyslog.d/60-rsyslog-forward-mdsd.conf" conflicts with other MDSD running on the AKS node?

parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o pipefail
+
+[ ! -f "/var/run/mdsd/update.status" ] && exit 0
```
ganga1980, line 6: I think, plan was to use container  insights specific path  to avoid any conflicts with MDSD, right? are you planning to update this path to container insights specific path for example, /var/run/containerinsights/update.status?

parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
```diff
@@ -0,0 +1,29 @@
...
+status=$(cat /var/run/mdsd/update.status)
+
+if [[ "$status" == "add" ]]; then
+        echo "Status changed to $status."
+        cat >/etc/rsyslog.d/60-rsyslog-forward-mdsd.conf << 'EOL'
+$ModLoad omuxsock 
+$OMUxSockSocket /var/run/mdsd/default_syslog.socket
```
ganga1980, line 13: Cant we configure the MDSD to create and listen on specific socket instead of using default socket which could avoid any potentially conflict with MDSD running on the host?

parts/linux/cloud-init/artifacts/ci-syslog-watcher.path
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Monitor the CI syslog status file for changes
+
+[Path]
+PathModified=/var/run/mdsd/update.status
```
ganga1980, line 5: is that Path Modified trigger the watcher? if yes, what and all cases this being triggered, meaning file create and update etc. ?

pfrcks: yes https://www.freedesktop.org/software/systemd/man/systemd.path.html

parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
pfrcks: yes that is the plan. this is a draft PR to test things out. I'll have the actual paths in next iteration

parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
pfrcks: yes we will use something like /var/run/mdsd-ci/default_syslog.socket

parts/linux/cloud-init/artifacts/ci-syslog-watcher.sh
pfrcks: currently there is not other scenario where mdsd is running on the node. however even if it was, upon renaming the config file and the socket there will not be any conflicts

alexeldeib: won't 1p often have mdsd (I guess that would require them having container monitoring enabled too)

pfrcks: We will support mariner. I tested it with mariner v1 and v2

pfrcks: we will be starting mdsd at a custom location to prevent any conflicts

parts/linux/cloud-init/artifacts/ci-syslog-watcher.service
pfrcks: if it fails it fails and we wait for the next path change to trigger it

pfrcks:
@alexeldeib @ganga1980 sorry for the PR ping pong but since this PR is from my fork and forks don't have permissions to run some pipelines I have opened another PR directly against this repo https://github.com/Azure/AgentBaker/pull/2267

cameronmeissner:
feel free to reopen without using a fork if needed

PR #2250, opened 2022-09-27
feat: Enable Mariner kata image build

hbeberman:
**What type of PR is this?**
/kind feature

Enables image builds for Mariner 2.0 Kata enabled container host.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -112,10 +112,15 @@ if [[ $OS == $MARINER_OS_NAME ]]; then
...
     setMarinerNetworkdConfig
-    enableDNFAutomatic
     fixCBLMarinerPermissions
     addMarinerNvidiaRepo
     overrideNetworkConfig || exit 1
+    if grep -q "kata" <<< "$FEATURE_FLAGS"; then
+      enableMarinerKata
+    else
+      # Leave automatic package update disabled for the kata image
+      enableDNFAutomatic
```
alexeldeib, line 122: cc @paulgmiller

this fancy kata containers image won't have dnf automatic updates enabled for compat reasons. 

(no action for mariner folks)

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -114,3 +114,11 @@ EOF
 # Future base images will already have rsyslog installed with 755 /etc/rsyslog.d
     chmod 755 /etc/rsyslog.d
 }
+
+enableMarinerKata() {
+    # Enable the mshv boot path
+    sudo sed -i -e 's@menuentry "CBL-Mariner"@menuentry "Dom0" {\n    search --no-floppy --set=root --file /EFI/Microsoft/Boot/bootmgfw.efi\n        chainloader /EFI/Microsoft/Boot/bootmgfw.efi\n}\n\nmenuentry "CBL-Mariner"@'  /boot/grub2/grub.cfg
```
alexeldeib, line 120: this seems...resilient...

is there really no better way than sed'ing it? you can't write the whole file out?

whitespace/newlines feels like a recipe for breakage

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -114,3 +114,11 @@ EOF
...
+
+    # Start kata VM initrd generation service
+    systemctlEnableAndStart kata-osbuilder-generate
```
alexeldeib, line 123: Can you leave a brief (well, slightly longer than current :D) comment of what this does, I think I saw it elsewhere but a bit unfamiliar with why it's needed.

alexeldeib:
https://github.com/Azure/AgentBaker/blob/master/pkg/agent/datamodel/sig_config.go#L391

need to update this to actually use it, since it's already there but unexposed to customers we can just update it now?

alexeldeib (approved):
minor comments

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
hbeberman: I'm not so thrilled about such an unwieldy sed either, but since it's what we've been building with for a few months I'd rather not change it at the last minute before we get a few images out.

This line gets removed when we've updated the boot path to fall back to defaults on a non-nested VM SKU.

Alternatively, @neha170 could we look into having both regular and DOM0 menu entries in our base image then just setting something like "GRUB_DEFAULT=1" as we're customizing the image in AgentBaker?

Camelron (approved):
Things look fine on the Mariner AKS side:

https://dev.azure.com/mariner-org/mariner/_build/results?buildId=243031&view=results
https://dev.azure.com/mariner-org/mariner/_build/results?buildId=243040&view=results

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
alexeldeib: np, can leave as is

PR #2254, opened 2022-09-28
fix: add gen2 as suffix in generate-sas for sku_name

wanqingfu:
**What type of PR is this?**
Add gen2 as suffix in `sku_name` so that the resulted VM image def for Gen 1 and Gen 2 will not have name conflicts. 
**Q**: How will it look like after this change?
**A**: Gen 1: windows-2022-containerd; Gen 2: windows-2022-containerd-gen2.

/kind bug

To fix the bug. Currently the prod. sub has the same "windows-2022-containerd" VM image def for both Gen 1 and Gen 2.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -20,10 +20,10 @@ steps:
       if [[ $(HYPERV_GENERATION) == "V2" ]]; then \
         m="sigMode"; \
         if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then \   
-          sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
+          sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}"; \
           echo "##vso[task.setvariable variable=GEN2_SIG_FOR_PRODUCTION]False"; \
         else
-          sigImageName="windows-${WINDOWS_SKU}-gen2-$RANDOM"; \
+          sigImageName="windows-${WINDOWS_SKU}-$RANDOM"; \
           sigGalleryName="WS2022Gen2Gallery$(date +"%y%m%d%H%M%S")$RANDOM"; \          
           echo "##vso[task.setvariable variable=SIG_GALLERY_NAME]$sigGalleryName"; \
           echo "##vso[task.setvariable variable=SIG_IMAGE_VERSION]1.0.0"; \
--- a/vhdbuilder/packer/generate-windows-vhd-configuration.ps1
+++ b/vhdbuilder/packer/generate-windows-vhd-configuration.ps1
@@ -9,7 +9,7 @@ if (-not ($validContainerRuntimes -contains $containerRuntime)) {
 }
 
 $global:windowsSKU = $env:WindowsSKU
-$validSKU = @("2019", "2019-containerd", "2022-containerd")
+$validSKU = @("2019", "2019-containerd", "2022-containerd", "2022-containerd-gen2")
 if (-not ($validSKU -contains $windowsSKU)) {
     throw "Unsupported windows image SKU: $windowsSKU"
 }
@@ -37,9 +37,8 @@ $global:defaultContainerdPackageUrl = "https://acs-mirror.azureedge.net/containe
 
 $global:defaultDockerVersion = "20.10.9"
 
-switch ($windowsSKU) {
-    "2019" {
-        $global:imagesToPull = @(
+if ($windowsSku -eq "2019") {
+    $global:imagesToPull = @(
             "mcr.microsoft.com/windows/servercore:ltsc2019",
             "mcr.microsoft.com/windows/nanoserver:1809",
             "mcr.microsoft.com/oss/kubernetes/pause:3.6-hotfix.20220114",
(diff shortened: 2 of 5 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -20,10 +20,10 @@ steps:
       if [[ $(HYPERV_GENERATION) == "V2" ]]; then \
         m="sigMode"; \
         if [[ -n ${SIG_GALLERY_NAME} && -n ${SIG_IMAGE_NAME_PREFIX} && -n ${SIG_IMAGE_VERSION} ]]; then \   
-          sigImageName="${SIG_IMAGE_NAME_PREFIX}-${WINDOWS_SKU}-gen2"; \
```
wanqingfu, line 23: Removing the "gen2" suffix because there's a new WINDOWS_SKU?

vhdbuilder/packer/init-variables.sh
```diff
@@ -276,7 +276,7 @@ if [ "$OS_TYPE" == "Windows" ]; then
 			os_disk_size_gb=${WINDOWS_2019_CONTAINERD_OS_DISK_SIZE_GB}
 		fi
 		;;
-	"2022-containerd")
+	"2022-containerd" | "2022-containerd-gen2")
```
wanqingfu, line 279: For Gen 1 and Gen 2, they both go into the same code block here. Initially, the WINDOWS_SKUs are the same "2022-containerd". Adding "gen2" suffix to differentiate??

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -89,7 +89,7 @@ switch ($windowsSKU) {
             # OMS-Agent (Azure monitor). Owner: ganga1980 (Ganga Mahesh Siddem)
             "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:win-ciprod08102022")
     }
-    "2022-containerd" {
+    "2022-containerd*" {
```
wanqingfu, line 92: Using regular expression matching to avoid redundant duplication?

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -207,7 +207,7 @@ steps:
       echo MODE=$(MODE) && \
       if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
       [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
-      SKU_NAME="windows-$WINDOWS_SKU-gen2"; \
+      SKU_NAME="windows-$WINDOWS_SKU"; \
```
wanqingfu, line 210: Rolling back to the master code version w/o the "gen2" suffix because the added WINDOWS_SKU already ends with "gen2"??

ShiqianTao (approved):

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -207,7 +207,7 @@ steps:
       echo MODE=$(MODE) && \
       if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
       [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
-      SKU_NAME="windows-$WINDOWS_SKU"; \
+      SKU_NAME="windows-$WINDOWS_SKU-gen2"; \
```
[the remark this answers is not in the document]
wanqingfu: I think the easiest way to fix is the one you are commenting here (the first commit). It requires only the `SKU_NAME` to have "gen2" suffix and then be outputted to the published artifact `vhd-publishing-info.json`. However, this is inconsistent with the one defined in line 119 above.

The better way may be defining a new `WINDOWS_SKU` in line 119 since logically there should be a new SKU for V2? This way, the names are more consistent. "gen2" will be included in `WINDOWS_SKU`, `VNET_RESOURCE_GROUP_NAME`, `artifactName` etc., basically everywhere involving Gen 2?

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -37,7 +37,7 @@ $global:defaultContainerdPackageUrl = "https://acs-mirror.azureedge.net/containe
 
 $global:defaultDockerVersion = "20.10.9"
 
-switch ($windowsSKU) {
+switch -Regex ($windowsSKU) {
     "2019" {
```
AbelHu, line 41: Did you test your changes? 2019-containerd will also fit "2019" with regex switch

AbelHu (changes requested):
It seems like that there are some issues by mixing 2019 and 2019-containerd in generate-windows-vhd-configuration.ps1

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
wanqingfu: Need to inspect the logs. Yes, I tested four cases.

wanqingfu:
> It seems like that there are some issues by mixing 2019 and 2019-containerd in generate-windows-vhd-configuration.ps1

Yes, you are right. Need to figure out the neatest way to write this, I guess?

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
wanqingfu: Outdated. Replaced by "if else" statement clauses.

wanqingfu: Checked the pulled images in pipeline logs. `pulling images for ****` I think it's now pulling the correct images per `windowsSku`. Are there any other issues?

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -88,11 +85,9 @@ switch ($windowsSKU) {
             "mcr.microsoft.com/oss/kubernetes/azure-cloud-node-manager:v1.23.11", # for k8s 1.23.x
             "mcr.microsoft.com/oss/kubernetes/azure-cloud-node-manager:v1.24.3", # for k8s 1.24.x
             # OMS-Agent (Azure monitor). Owner: ganga1980 (Ganga Mahesh Siddem)
-            "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:win-ciprod10042022-3c05dd1b"
-        )
-    }
-    "2022-containerd" {
-        $global:imagesToPull = @(
+            "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:win-ciprod10042022-3c05dd1b")
+} elseif ($windowsSku.Contains("2022-containerd")) {
```
AbelHu, line 89: Can we compare the sku strictly to avoid any human mistakes?

wanqingfu: ok

AbelHu (approved):

PR #2259, opened 2022-09-30
Updating AgentBaker for new image version for Azure Monitor Metrics

rashmichandrashekar:
**What type of PR is this?**
Updating AgentBaker for new image version for Azure Monitor Metrics


vhdbuilder/packer/components.json
```diff
@@ -141,7 +141,8 @@
     {
       "downloadURL": "mcr.microsoft.com/azuremonitor/containerinsights/ciprod/prometheus-collector/images:*",
       "amd64OnlyVersions": [
-        "5.1.0-main-09-23-2022-df3e2703"
+        "5.1.0-main-09-23-2022-df3e2703",
```
chengliangli0918, line 144: AKS ARM64 Ubuntu 22.04 are GA now. Any container images of new partner add-on feature should be 'multi-arch'.  On AKS RP repo side, we have dedicated pipeline to catch this violation (using new amd64 only add-on container images)

rashmichandrashekar: @chengliangli0918 - We are not supporting multi arch for now.. We are going public preview at ignite. Could you bypass this check to get this merged?

chengliangli0918 (approved):

PR #2230, opened 2022-09-22
Update k8s patch versions: 1.22.15, 1.23.12, 1.24.6, 1.25.2- emergency release

ganeshkumar5699:
**What type of PR is this?**

**What this PR does / why we need it**:
Update k8s patch versions for all k8s minor versions, with CVE fix and related bug fix in the emergency patch version release. 

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #2275, opened 2022-10-06
chore: Add new Windows Kubernetes hotfixes for v1.24

AbelHu:
**What type of PR is this?**

Add new Windows Kubernetes hotfixes for v1.24
Reference: https://github.com/Azure/AKS/issues/3246

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -192,10 +192,8 @@ $global:map = @{
         "https://acs-mirror.azureedge.net/kubernetes/v1.23.7/windowszip/v1.23.7-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.23.8-hotfix.20220728/windowszip/v1.23.8-hotfix.20220728-1int.zip",
         "https://acs-mirror.azureedge.net/kubernetes/v1.23.12-hotfix.20220922/windowszip/v1.23.12-hotfix.20220922-1int.zip", 
-        "https://acs-mirror.azureedge.net/kubernetes/v1.24.0/windowszip/v1.24.0-1int.zip",
-        "https://acs-mirror.azureedge.net/kubernetes/v1.24.3/windowszip/v1.24.3-1int.zip",
-        "https://acs-mirror.azureedge.net/kubernetes/v1.24.6/windowszip/v1.24.6-1int.zip",
-        "https://acs-mirror.azureedge.net/kubernetes/v1.25.2/windowszip/v1.25.2-1int.zip"
```
ganeshkumar5699, line 198: k8s 1.25.2 will be previewed this week, so we would need it on the VHD for Windows

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -192,10 +192,8 @@ $global:map = @{
...
+        "https://acs-mirror.azureedge.net/kubernetes/v1.24.3-hotfix.20221006/windowszip/v1.24.3-hotfix.20221006-1int.zip", 
+        "https://acs-mirror.azureedge.net/kubernetes/v1.24.6-hotfix.20221006/windowszip/v1.24.6-hotfix.20221006-1int.zip"
     );
```
ganeshkumar5699, line 197: Would it also be possible to make the corresponding version updates for Linux, like here: https://github.com/Azure/AgentBaker/pull/2230/files? While the contents would be the same for k8s binaries on Linux, it would be good to have parity for hotfix versions (before the auto patching pipeline is run), as we may also need to make corresponding changes on the RP side. Thank you.

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
AbelHu: @ganeshkumar5699, v1.25.2 is not available so I only add v1.24 in this PR

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
AbelHu: @ganeshkumar5699 This hotfix only has the hotfix for Windows so I think that Linux does not need to update the k8s packages. If we cache these packages for Linux, it means that we need a change in RP code. Will you do it?

AbelHu: @ganeshkumar5699 Added the hotfix for 1.25.2 on Linux

ShiqianTao (approved):

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
ganeshkumar5699: Thank you!

PR #2276, opened 2022-10-06
Explicitly purge and hold non-fips kernel in FIPS VHD

chengliangli0918:
**What type of PR is this?**
/kind bug

Explicitly purge and hold non-fips kernel in FIPS VHD

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
+++ b/vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
@@ -140,6 +140,12 @@ installFIPS() {
     echo "Installing FIPS..."
     wait_for_apt_locks
 
+    # installing fips kernel doesn't remove non-fips kernel now, purge current linux-image-azure
+    echo "purging linux-image-azure..."
+    apt_get_purge 5 10 120 linux-image-azure
+    retrycmd_if_failure 120 5 25 apt-mark hold linux-image-azure
+    retrycmd_if_failure 120 5 25 apt-mark hold linux-image-$(uname -r)
+
     echo "adding ua repository..."
     retrycmd_if_failure 5 10 120 add-apt-repository -y ppa:ua-client/stable || exit $ERR_ADD_UA_APT_REPO
     apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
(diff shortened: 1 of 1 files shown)


vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -140,6 +140,12 @@ installFIPS() {
     echo "Installing FIPS..."
     wait_for_apt_locks
 
+    # installing fips kernel doesn't remove non-fips kernel now, purge current linux-image-azure
+    echo "purging linux-image-azure..."
+    apt_get_purge 5 10 120 linux-image-azure
+    retrycmd_if_failure 120 5 25 apt-mark hold linux-image-azure
+    retrycmd_if_failure 120 5 25 apt-mark hold linux-image-$(uname -r)
```
alexeldeib, line 147: if `apt upgrade` pulls in a newer kernel, this will only purge the currently running kernel, but the other apt package will still exist

you mentioned this scenario should be caught by run test anyway, but jfyi

alexeldeib: can otherwise grep/wildcard somehow and remove all packages

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -140,6 +140,12 @@ installFIPS() {
     echo "Installing FIPS..."
     wait_for_apt_locks
 
+    # installing fips kernel doesn't remove non-fips kernel now, purge current linux-image-azure
+    echo "purging linux-image-azure..."
+    apt_get_purge 5 10 120 linux-image-azure
+    retrycmd_if_failure 120 5 25 apt-mark hold linux-image-azure
```
alexeldeib, line 146: none of these functions exit on failure and we don't `set -e` by default, consider adding `|| exit 1` or similar to force these cases

alexeldeib (approved):
few comments but proactively approving

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
chengliangli0918: use grep/wildcard to purge & hold all non-fips kernel packages.

PR #2113, opened 2022-08-24
Add new Windows sig distro for gen 2

wanqingfu:
**What type of PR is this?**
Add new sig image in AgentBaker for RP to use later.

**What this PR does / why we need it**:
To support Windows Gen 2 VM sizes. Currently AgentBaker does not have Windows Gen 2 sig.

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #2289, opened 2022-10-08
fix: add new gen 2 sig version and sig config template in AgentBaker

wanqingfu:
**What type of PR is this?**
To support the new Windows Gen 2 sig image in AgentBaker.

/kind feature

1. Add a new Gen 2 sig version.
2. Add a new Gen 2 sig config template for RP to use.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

The sig image version comes from the build artifact `vhd-publishing-info.json` of the latest Gen 2 VHD production pipeline run.

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -207,7 +207,7 @@ steps:
       echo MODE=$(MODE) && \
       if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
       [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
-      SKU_NAME="windows-$WINDOWS_SKU-gen2"; \
+      SKU_NAME="windows-$WINDOWS_SKU"; \
       else \
       OS_DISK_SAS="$(cat packer-output | grep -a "OSDiskUriReadOnlySas:" | cut -d " " -f 2)"; \
       VHD_NAME="$(echo $OS_DISK_SAS | cut -d "/" -f 8 | cut -d "?" -f 1)"; \
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -207,7 +207,7 @@ steps:
       echo MODE=$(MODE) && \
       if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
       [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
-      SKU_NAME="windows-$WINDOWS_SKU"; \
+      SKU_NAME="windows-$WINDOWS_SKU-gen2"; \
       else \
       OS_DISK_SAS="$(cat packer-output | grep -a "OSDiskUriReadOnlySas:" | cut -d " " -f 2)"; \
       VHD_NAME="$(echo $OS_DISK_SAS | cut -d "/" -f 8 | cut -d "?" -f 1)"; \
--- a/pkg/agent/datamodel/sig_config.go
+++ b/pkg/agent/datamodel/sig_config.go
@@ -236,6 +236,7 @@ const (
 
 	Windows2019SIGImageVersion string = "17763.3406.220913"
 	Windows2022SIGImageVersion string = "20348.1006.220913"
+	Windows2022Gen2SIGImageVersion string = "20348.1006.220930"
 
 	Arm64LinuxSIGImageVersion    string = "2022.09.27"
 	Ubuntu2204SIGImageVersion    string = "2022.09.27"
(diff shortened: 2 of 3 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -207,7 +207,7 @@ steps:
       echo MODE=$(MODE) && \
       if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
       [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
-      SKU_NAME="windows-$WINDOWS_SKU"; \
+      SKU_NAME="windows-$WINDOWS_SKU-gen2"; \
```
wanqingfu, line 210: This comes from the unmerged PR https://github.com/Azure/AgentBaker/pull/2254, I think.

pkg/agent/datamodel/sig_config.go
```diff
@@ -236,6 +236,7 @@ const (
 
 	Windows2019SIGImageVersion string = "17763.3406.220913"
 	Windows2022SIGImageVersion string = "20348.1006.220913"
+	Windows2022Gen2SIGImageVersion string = "20348.1006.220930"
```
wanqingfu, line 239: You may check the version value from the production pipeline run of the Gen 2 VHD.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -207,7 +207,7 @@ steps:
       echo MODE=$(MODE) && \
       if [[ "${GEN2_SIG_FOR_PRODUCTION}" == "True" ]]; then gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')"; \
       [ -n "${gen2_captured_sig_version}" ] && VHD_NAME="${gen2_captured_sig_version}.vhd"; \
-      SKU_NAME="windows-$WINDOWS_SKU-gen2"; \
+      SKU_NAME="windows-$WINDOWS_SKU"; \
```
wanqingfu, line 210: Rolling back to the master code. (without gen2 suffix)

.pipelines/templates/.builder-release-template-windows.yaml
wanqingfu: Deleted in the later commit to only include sig image version changes?

pkg/agent/datamodel/sig_config.go
AbelHu, line 239: Since we always build and publish new AKS Windows images at the same time. We share the image version for ws2019docker and ws2019containerd. It seems like that we can also share the image version for ws2022 gen1 and ws2022 gen2.

For the test purpose since you can separate them in your private branch.

wanqingfu: Ok. So I should follow "AKS-RP tests the forked AgentBaker" and do standalone testing in RP using this private branch?

wanqingfu: Removed. Understood you only wanna keep one version eventually. In the latest change, the gen2 sig version was hard-coded https://github.com/Azure/AgentBaker/pull/2289/commits/5756472689443fa97449f4f7155c35767cd213b3#L427. Was wondering if we can merge this into AB master temporarily for test purpose only. The change is pretty similar to my previous one https://github.com/Azure/AgentBaker/pull/2113/files except changing line `410`.

I believe it's the quickest way to test if the published Gen 2 sig can be used by RP for POC. For argument's sake, reasons are as follows:
1. Suspected `"Generating Template Failed (Details: can't find image for distro aks-windows-2022-containerd-gen2"` trying to create Gen 2 nodepool was because - the version in `SIGWindows2022ContainerdGen2ImageConfigTemplate` used to be `Windows2022SIGImageVersion`. This was set to `20348.1006.220913`, that is different from Gen 2: `20348.1006.220930`. Of course the creation would fail.

    My understanding is if this version is correct, standalone should be able to create Gen 2 nodepool if the published sig actually works. Personally, I am more comfortable with this. Logically this isn't done right for it's just a premature PR change. I totally get it. But if we can get this workaround working, I believe it's very easy to roll back this change any time.

    AFAIK, all standalone experiments excluding Gen 2 seemed to be fine. At least the standalone env can be created since it is using the master AB version?

2. The issues relating to AKS-RP testing my forked AB are around `make update_agent_baker`. I think it may not be the most efficient way if I try to phrase my questions because
- From the standalone creation logs, the problem relates to this `enableTelemetry`. I guess it was deprecated.  

  - https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp?version=GBmaster&path=/protos/hcp/types/managedcluster/v1/featureflags.proto - `  // deprecated: enableTelemtry
  reserved 4;`  
  - I don't seem to find this feature flag in `typeconversion.go/toAgentBakerFeatureFlags` either. 
  - Concerning this `telemetry`, I suppose AB & RP are not supporting it very well - might be wrong anyway.
  
- Sorting out the issues and explaining take time both from the technical side and the language side. I did rebased my private forked AB repo on master, ran this command on master RP, rebased my RP on master, toyed with Telemetry profile in AB and RP etc and etc - I can start a whole new thread explaining the error logs and changes. 

   I can do this simultaneously, but I do think it's ratholing a new bigger issue. After observing logs of multiple runs of AKS Dev Deploy pipelines, I sense it is not an easy fix - let alone say I was just blindly exploring without understanding the fundamentals underneath.
   
Anyway, with two options at hand I think we may 1) merge this 2) standalone testing RP using the master AB, and at the same time find a way to vendor my forked AB into RP. 

The above only represents my personal opinion. I am open to vendoring the forked AB into RP if we decide to go down this right way.

@AbelHu @ShiqianTao @junjiezhang1997 what do you think? 

Thanks for reading.

PR #2297, opened 2022-10-11
chore: Remove unused CNI version: 1.4.27

ganeshkumar5699:
**What type of PR is this?**
/kind cleanup

Linux VHD is full. Removing unused CNI version to clear space.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -345,7 +345,6 @@ for VNET_CNI_VERSION in $SWIFT_CNI_VERSIONS; do
 done
 
 OVERLAY_CNI_VERSIONS="
-1.4.27
```
vipul-21, line 348: Couple of things: 
1. did we verify with all the consumers of this version (if there are any)  that this can be removed ? Last I remember overlay-cni 1.4.27 was used in aks-rp side.
2. Should we remove the same from component.json file as well: https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/components.json#L445

ganeshkumar5699: 1. I did a brief check in AKS-RP, and did not see this version referenced. Do you know who the owner of the component is? We can ask them to confirm, as component owners are supposed to maintain this. 
2. Good point, thanks. Yes, we can remove this from component.json, if it's not needed.

vipul-21: We can ask Will daily about it just to be sure.

rbtr: we keep n and n-1 for continuity since the aks-rp toggle rollouts take some time to hit all regions, so even if it's not currently referenced in the codebase it might be in the current prod toggles for certain regions still

alexeldeib: that's fine, this is merely a cache. unfortunately, our cache is full 🙂 we need to evict some free space 😄 

we can add to allow list in rp if we hit issues on the other side.

ganeshkumar5699: Thanks, and in this case the versions were not added recently. So I'll go ahead and merge the PR.

mattstam (changes requested):
Confirmed no reference in RP, should be good to remove this.

Should also remove `1.4.27` in [components.json](https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/components.json#L445) in this PR.

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: cc: @wedaly

mattstam (approved):

PR #2318, opened 2022-10-13
add TL skeleton to build pipeline

anujmaheshwari1:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -34,17 +34,34 @@ fi
...
 
-az resource create --id $disk_resource_id  --is-full-object --location $LOCATION --properties "{\"location\": \"$LOCATION\", \
-  \"properties\": { \
-    \"osType\": \"$OS_TYPE\", \
-    \"creationData\": { \
-      \"createOption\": \"FromImage\", \
-      \"galleryImageReference\": { \
-        \"id\": \"${sig_resource_id}\" \
+if [[ ${ENABLE_TRUSTED_LAUNCH} == "True" ]]; then
+  az resource create --id $disk_resource_id  --is-full-object --location $LOCATION --properties "{\"location\": \"$LOCATION\", \
```
alexeldeib, line 38: hmm, we should figure out if we can extract this more cleanly, e.g. mutate a base template

alexeldeib: not for now

alexeldeib (approved):
some opportunities for later refactoring but lgtm

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
anujmaheshwari1: will merge and iterate on this in follow up PR

anujmaheshwari1:
successful test run: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=61916011&view=results

PR #2324, opened 2022-10-13
one linux VHD build mode, eliminate gen1, gen2, and sig modes for linux VHD builds

cameronmeissner:
- Merged gen1, gen2, and sig modes into one linuxVhdMode for Linux VHD builds
- Windows VHD builds will maintain their current state and retain all three original modes
- We now build MarinerV1 Gen1 VHD images from a marketplace base image instead of a pre-supplied VHD blob
- All Linux VHD builds through packer will now output a managed image which then gets published to a SIG version
- SIG image versions created by Packer will be deleted after build completion when dry-running

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -130,30 +132,15 @@ stages:
...
+          condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), eq(variables['MODE'], 'linuxVhdMode'))
         - bash: |
             echo MODE=$(MODE) && \
             PKR_RG_NAME="$(cat packer-output | grep "ResourceGroupName" | cut -d "'" -f 2 | head -1)" && \
             SA_NAME="$(cat packer-output | grep "storage name:" | cut -d " " -f 3)" && \
             gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
             IMPORTED_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "imported_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
             ARM64_OS_DISK_SNAPSHOT_NAME="$(cat vhdbuilder/packer/settings.json | grep "arm64_os_disk_snapshot_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
-            if [[ $(MODE) == "gen2Mode" ]]; then IMAGE_NAME="${gen2_captured_sig_version}"; else IMAGE_NAME=$SIG_IMAGE_NAME; fi && \
+            if [[ "${HYPERV_GENERATION,,}" == "v2" ]]; then IMAGE_NAME="${gen2_captured_sig_version}"; else IMAGE_NAME=$SIG_IMAGE_NAME; fi && \
```
alexeldeib, line 143: typo?

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -193,10 +180,10 @@ stages:
             -e GEN2_CAPTURED_SIG_VERSION=${gen2_captured_sig_version} \
             ${CONTAINER_IMAGE} make -f packer.mk convert-sig-to-classic-storage-account-blob
           displayName: Convert Shared Image Gallery To VHD Blob In Classic Storage Account
-          condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), eq(variables['MODE'], 'gen2Mode'))
+          condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), eq(variables['MODE'], 'linuxVhdMode'))
```
alexeldeib, line 183: do we need this is it's our only mode?

cameronmeissner: not necessary but figured I'd keep it for verbosity's sake, though fine with removing if you think that's best

.pipelines/.vsts-vhd-builder.yaml
cameronmeissner: instead of checking MODE we now check HYPERV_GENERATION, don't think I see a typo here? unless you're saying that we shouldn't be using HYPERV_GENERATION here

cameronmeissner: granted I'm thinking about how we can change this logic altogether since we'll always need a unique captured sig version at this point

cameronmeissner: yeah, so I'm gonna change this such that we always get the image name from the captured_sig_version since we'll no longer have a case where packer outputs a disk URI

wanqingfu: sorry to interrupt here. are you suggesting packer will always output a managed sig image, and so the image will have to be snapshot into an OS disk blob before publishing?

cameronmeissner: these changes will have it such that packer always outputs a managed image and publishes it to a SIG version for Linux-based VHDs - this means that we will always be converting a SIG image version to classic storage account blob, which will then be forwarded on to the appropriate release pipelines

I've made sure to keep the windows build definitions/modes the same with these changes however

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -140,7 +140,7 @@ steps:
   - bash: |
         echo MODE=$(MODE) && \
         SIG_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "sig_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
-        gen2_captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "gen2_captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
```
cameronmeissner, line 143: @wanqingfu @alexeldeib ended up renaming "gen2_captured_sig_version" to "captured_sig_version" since we're now using it in all cases on the linux side, if you don't think this makes sense then I can revert - functionality wise everything should remain the same on the windows side.

wanqingfu: i'm okay with it. yeah the conversion should function properly.

AbelHu:
@wanqingfu could you help to review this PR to make sure that Windows SIG build and GEN2 build still work with this change?

wanqingfu:
I've gone through all the file changes. This PR is compatible with WINDOWS SIG build and GEN2 build @AbelHu. For the windows part changes, they are basically only renaming the `GEN2_CAPTURED_SIG_VERSION` to `CAPTURED_SIG_VERSION` - functionality remains the same. 

Confirmed with @cameronmeissner and will continue to follow up on this. I've been monitoring the pipeline also, and I don't think I'm seeing any *fail* in most recent runs relating to this PR.

anujmaheshwari1 (approved):
lgtm!

hbeberman (approved):
:shipit:

PR #2334, opened 2022-10-18
feat: support ExternalTrafficPolicy Cluster with IPv6 LoadBalancer services

tyler-lloyd:
**What type of PR is this?**

/kind feature

In AKS, nodes act as a router and forward traffic to the container's network namespace. Per RFC 4193, "Routers must not forward any packets with Link-Local source or destination addresses to other links", which means when the link-local source packet arrives on the node to probe for the health of the service, the node cannot forward that packet to a pod network namespace running on that node. This prevents k8s LoadBalancer services from using `externalTrafficPolicy: Cluster` because Kubernetes is attempting to determine the health of pods backing a service directly (versus ETP: Local, where the health probe is replied to by kube-proxy itself on the node which doesn't need to forward health probe packets).

By changing the src IP of the probe packet to the globally unique address mentioned above, nodes are able to forward the probe packet to the correct destination network namespace thus enabling the use of `externalTrafficPolicy: Cluster` for IPv6 services in a dual-stack AKS cluster.

This PR ~~includes a new step in CSE, `configureCNINFTables`, which runs as part of the CNI setup function and~~ will create nftables rules for re-writing the source IP of IPv6 packets from the Azure Load Balancer probe's address link-local address, `fe80::1234:5678:9abc`, to a globally unique address `2603:1062:0:1:fe80:1234:5678:9abc`.

update: Instead of putting a new step in CSE, we're adding a systemd unit to run at boot every time. This ensures the rules are written between restarts and bakes the service into the VHD. The service will run a script to query IMDS for to see if the instance has an IPv6 private address and if so, writes the nftables rules. Otherwise it is a no-op.

The new rules try to be as non-disruptive as possible and are setup as stateless NAT rules, meaning no connection tracking occurs.

**These new rules will only be written for clusters that have IPv6 enabled** (currently only dual-stack).

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

- Adds `nftables` to image dependencies. Happy to discuss if we should have these rules just be written with iptables but there might be advantages to trying to move to nftables in the future as it offers better performance and priority.
- Updates the KubernetesConfig model to include recent updates for IPv6 clusters (ipFamilies, clusterSubnets, and serviceCIDRs)


alexeldeib:
I'm okay with these changes conceptually but have a few clarifying questions (especially since I am not familiar with the RFC)

> By changing the src IP of the probe packet to the globally unique address

Just reiterating...we can't forward the packet because source IP is link local, but we can SNAT it then forward it?

> a globally unique address 2603:1062:0:1:fe80:1234:5678:9abc

what is this? looks like it's publicly routable, this is the ipv6 equivalent of 168.63.129.16 for Azure LB? 

How is this different from ipv4? Seems like with ipv4 source IP is already 168.63.129.16, is there something semantically wrong with using a non-link local address with ipv6 from the LB side as well (really a question more for SLB team I guess, but unfortunately picking on you)

and yeah, why nftables 🙂 not really opposed, but would love a thorough sanity check that this doesn't mess with e.g. kube-proxy

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -241,6 +242,27 @@ configureCNIIPTables() {
     fi
 }
 
+configureCNINFTables() {
+    if [[ "${IS_IPV6}" = "true" ]]; then
+        # Delete the table in a subshell so that we can eat the failed return code
+        (nft -na -- list table ip6 azureSLBProbe >/dev/null 2>&1 && nft -- delete table ip6 azureSLBProbe; exit 0)
+
+        # Create a custom table
+        nft -- add table ip6 azureSLBProbe
```
alexeldeib, line 251: would it make sense to have the full table/chain defined in a file and just `nft -f $file` ? I would fine that a little more readable personally

phealy (changes requested):
A few items to address:

1. I never tested the nftables commands on 16.04, I have no idea if they work. We should probably block this to be 18.04 only.
1. Isn't this going to fail on nodes running new CSE but old node image because they won't have nftables installed? Do we need a command -v nftables >/dev/null || apt-get install -y nftables for Ubuntu?
1. What does this do to mariner, since we probably don't have the package their either?

phealy:
> Just reiterating...we can't forward the packet because source IP is link local, but we can SNAT it then forward it?

Yes, if a packet matching fe80::/10 gets to ROUTING with that source address, it gets dropped as "beyond scope"

> a globally unique address 2603:1062:0:1:fe80🔢5678:9abc

An IP address assigned permanently by Dawn Bedard from Microsoft owned space specifically for AKS probe SNAT. It'll never be used for anything else, so we don't have to worry about these rules breaking something else down the line.

> How is this different from ipv4? Seems like with ipv4 source IP is already 168.63.129.16, is there something semantically wrong with using a non-link local address with ipv6 from the LB side as well (really a question more for SLB team I guess, but unfortunately picking on you)

I talked to SLB extensively about this - they agree with the problem of link-local addresses and that it shouldn't have been done this way, but changing it carries a huge amount of risk and technical debt and would take more than a semester. So, workaround.

> and yeah, why nftables 🙂 not really opposed, but would love a thorough sanity check that this doesn't mess with e.g. kube-proxy

You literally can't make this change in iptables - iptables only allows the SNAT action in the NAT table POSTROUTING action and won't let you mangle the source IP before ROUTING, and the packet's already been dropped by then. nftables is more flexible and lets you mangle src early (at priority -300) before the system really sees it.

alexeldeib:
> Isn't this going to fail on nodes running new CSE but old node image because they won't have nftables installed? Do we need a command -v nftables >/dev/null || apt-get install -y nftables for Ubuntu

alternatively, we can bake this into the vhd as a systemd unit/oneshot or similar. downside is it keeps it pinned to the node image. benefit is less CSE which we do want. depends on UX

> Mariner

let's see if we can get this working on mariner, cc @hbeberman

parts/linux/cloud-init/artifacts/ipv6_nftables
```diff
@@ -0,0 +1,15 @@
+#!/sbin/nft -f
```
alexeldeib, line 1: does this actually work? you are directly executing with bash in the system oneshot, I would expect bash to parse this as a shell script and fail?

alexeldeib: oo nvm I see how it works, yeah would drop the shebang if this isn't a real script 

I see it seems to be a convention, albeit a dated (?) one: https://bugzilla.redhat.com/show_bug.cgi?id=1756347

up to you if you want to leave it

tyler-lloyd:
> alternatively, we can bake this into the vhd as a systemd unit/oneshot or similar.

latest updates move it to a systemd unit with a 10s delay between restarts if it fails. The script will check if the instance has IPv6 configured with a call to IMDS, looking at `network.interface` for non-empty IPv6 private IP addresses. The script will also install nftables if it isn't detected.

parts/linux/cloud-init/artifacts/ipv6_nftables
tyler-lloyd: not sure if this comment was on the latest commit but right now:

- service has `ExecStart=/bin/bash /opt/scripts/ipv6_nftables.sh`
- then that script has `nft -f $NFTABLES_RULESET_FILE`
- and the nftables rule file still has `#!/sbin/nft -f`

this all seems to work. However, I noticed that when installing nft on a node it gets installed to `/usr/sbin` so should I just remove the `#!/sbin/nft -f` from the rule file altogether?

parts/linux/cloud-init/artifacts/ipv6_nftables.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Configure nftables rules for handling Azure SLB IPv6 health probe packets
+
+[Service]
```
alexeldeib, line 4: this allows service invocation and restart on failure, but is of Type=simple and so will default to assuming execution succeeds as soon as `fork()` returns, meaning we won't actually track success of the script

suggest the workaround in https://unix.stackexchange.com/questions/272313/how-to-assure-a-oneshot-service-gets-retried-if-it-fails-the-first-time, making the service `Type=notify`, with `systemd-notify --ready` after the script succeeds? I don't see a better way for the same restart behavior with simple/oneshot. Exec gets you the error handling, but systemd won't wait for it to exit like oneshot.

alexeldeib: this is also relatively minor, I mean if the script fails the traffic will just break regardless of service type, this is mostly to provide a good signal (systemd unit failing) that something is wrong.

parts/linux/cloud-init/artifacts/ipv6_nftables
alexeldeib: yep sorry, I misread the ExecStart as directly executing the table, missed the misdirection with the IMDS bits initially .

tyler-lloyd: 👍  ... disregard my comment then, I wrote it before the page updated with your second comment. I'll just remove the shebang from ipv6_nftables file

parts/linux/cloud-init/artifacts/ipv6_nftables.sh
```diff
@@ -0,0 +1,53 @@
...
+IPV6_ADDR_COUNT=$(curl -sSL -H "Metadata: true" "http://169.254.169.254/metadata/instance/network/interface?api-version=2021-02-01" | \
+    jq '[.[].ipv6.ipAddress[] | select(.privateIpAddress != "")] | length')
+
+if [[ $IPV6_ADDR_COUNT -eq 0 ]]; then
+    echo "instance is not configured with IPv6, skipping nftables rules"
+    exit 0
+fi
+
+# Install nftables if it's not already on the node
+command -v nft >/dev/null || {
```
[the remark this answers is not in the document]
alexeldeib: doesn't distinguish mariner/ubuntu regardless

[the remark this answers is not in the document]
alexeldeib: actually this is entirely a no-op since the service is in the vhd :D please remove :)

alexeldeib:
couple remaining minor comments, LGTM after that 👍

parts/linux/cloud-init/artifacts/ipv6_nftables.sh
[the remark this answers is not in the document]
tyler-lloyd: updated

parts/linux/cloud-init/artifacts/ipv6_nftables.service
tyler-lloyd: latest update is `Type=notify` with the `systemd-notify --ready` at the end of the script. Tested on node and seems to work fine.

So is the tl;dr that this makes it easier to troubleshoot if this service is failing because it will stop other services from starting before it is finished?

alexeldeib: Sorry, brain shortcircuited. I don't think you actually need Type=notify, although it won't hurt. 

> is the tl;dr that this makes it easier to troubleshoot if this service is failing because it will stop other services from starting before it is finished

right. but, there are no dependent services :D

My thinking for context:
- Type=simple considered successfully executed after fork()
- Type=exec ... after exec()
- Type=oneshot ... after exit with non-zero code
- Type=notify...after notification sent for readiness

The think my brain lost track of, is of course systemd will consider the unit failed if it exits non-zero regardless. Since there are no deps, there's not much point blocking initial startup, you can just let it fail execution and enter failed state regardless of service type (but probably still want something other than oneshot for Restart=on-failure on e.g. imds failures or something).

alexeldeib:
I think PR build is bust because we had to force PRs from this repo not forks (one of our PR checks has secrets, TODO to refactor that), would you mind reopening this PR from a branch pushed to Azure/AgentBaker directly? lmk if you need write access

lgtm after that depending on what you want to keep in the unit

parts/linux/cloud-init/artifacts/ipv6_nftables.service
tyler-lloyd: makes sense, I'll move it back to `simple` with the restarts.

tyler-lloyd:
> lmk if you need write access

pretty sure I will. Getting a 403 trying to push to Azure/AgentBaker.

alexeldeib:
@tyler-lloyd you're in agentbakerwrite: 
image

maybe need to SSO approve your ssh key/PAT?

tyler-lloyd:
completed in #2393

PR #1451, opened 2022-01-22
IP SANs in Kubelet serving cert

UtheMan:
Adds IP SANs to Kubelet serving cert to allow components to use secure TLS.

PR #2352, opened 2022-10-21
fix: Add subjectAltName for kubeletserver cert

yaqisun7:
**What type of PR is this?**
/kind bug

Current kubeletserver certificate does not have subjectAltName field, which is not compliant to the description in https://www.rfc-editor.org/rfc/rfc2818. 
RFC 2818: "If a subjectAltName extension of type dNSName is present, that MUST be used as the identity. Otherwise, the (most specific) Common Name field in the Subject field of the certificate MUST be used. Although the use of the Common Name is existing practice, it is deprecated and Certification Authorities are encouraged to use the dNSName instead."

Some packages may remove support for use of common names when a certificate doesn't provide a subjectAltName, such as https://github.com/urllib3/urllib3/issues/497. So, the kubeletserver certificate without subjectAltName will cause problems for users.

This pr adds a subjectAltName with node name as DNSName.

Fixes # https://github.com/Azure/AKS/issues/2770

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


karataliu (approved):

alexeldeib (changes requested):
please do not merge this, it breaks FIPS

https://github.com/Azure/AgentBaker/pull/1451
https://github.com/Azure/AgentBaker/pull/1471

I am working on redesigning the entire flow for kubelet server TLS bootstrap, please see https://github.com/Azure/aks-tls-bootstrap and https://github.com/alexeldeib/supreme-goggles

alexeldeib:
cc @chengliangli0918

alexeldeib:
alternatively, please test this with standalone on FIPS to validate it works.

alexeldeib (approved):
lgtm, tested in standalone and seems fine on fips.

alexeldeib:
for posterity: hostname SAN works, IP SAN appeared to break FIPS.

PR #2353, opened 2022-10-21
disable UU in image and enable just-in-time

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -285,7 +285,15 @@ if $REBOOTREQUIRED; then
...
-        /usr/lib/apt/apt.systemd.daily &
+        {{- if EnableUnattendedUpgrade }}
+        systemctl unmask apt-daily.service apt-daily-upgrade.service
+        systemctl enable apt-daily.service apt-daily-upgrade.service
+        systemctl enable apt-daily.timer apt-daily-upgrade.timer
+        {{- end }}
+        # this is the DOWNLOAD service
+        # meaning we are wasting IO without even triggering an upgrade 
+        # -________________-
+        systemctl restart apt-daily.service
```
alexeldeib, line 296: I changed this to invoke the oneshot service directly since backgrounded task is pretty useless and would exit on cse handler termination without nohup or daemonize. 

but the timer itself will execute it nearly immediately, not sure we should bother with this.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -562,4 +562,17 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
   # update message-of-the-day to start after multi-user.target
   # multi-user.target usually start at the end of the boot sequence
   sed -i 's/After=network-online.target/After=multi-user.target/g' /lib/systemd/system/motd-news.service
-fi
\ No newline at end of file
+
+  # disable and mask all UU timers/services
```
alexeldeib, line 566: _with fire_

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -235,15 +235,15 @@ write_files:
...
 
 {{if not IsMariner}}
-{{if DisableUnattendedUpgrade }}
+{{if EnableUnattendedUpgrade }}
 - path: /etc/apt/apt.conf.d/99periodic
   permissions: "0644"
   owner: root
   content: |
-    APT::Periodic::Update-Package-Lists "0";
+    APT::Periodic::Update-Package-Lists "1";
```
alexeldeib, line 243: this just matches the defaults in the image before I disabled them

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: @paulgmiller for thoughts?

ganeshkumar5699 (approved):

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -235,15 +235,15 @@ write_files:
...
     APT::Periodic::Download-Upgradeable-Packages "0";
```
paulgmiller, line 244: If we're not going to change autoclean or Download-Upgradeable-Packages why have them her at all.

alexeldeib: touche

vhdbuilder/packer/install-dependencies.sh
paulgmiller: more comments please about why you're not just removing the pacakge or relying on the configfile. I am not sure this is necessary and makes it a bit harder a customer to re-nable on their own with a daemonset (which you coul argue is a plus or a minus).

parts/linux/cloud-init/artifacts/cse_main.sh
paulgmiller, line 296: id didn't understand why this is restart rather than start but I'm just a systemd ignoramus.

paulgmiller (approved):
I think this is probably fine though per comments a bit concerned that touching actualy services is overkill and more complicated than removing the package.

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: pretty much the same, afaik there's no cases where start is actually preferable to restart

PR #2368, opened 2022-10-25
Remove and reuse crictl versions to download in VHD build, for k8s versioning

ganeshkumar5699:
**What type of PR is this?**
/kind cleanup

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -136,12 +136,20 @@ if [[ ${CONTAINER_RUNTIME:-""} == "containerd" ]]; then
...
+      if [[ ${CRICTL_VERSIONS_STR} != null ]]; then
+        CRICTL_VERSIONS=$(echo "${CRICTL_VERSIONS_STR}" | jq -r ".[]")
+      fi
+      break
+    fi
+  done
+  echo $CRICTL_VERSIONS
+
   for CRICTL_VERSION in ${CRICTL_VERSIONS}; do
     downloadCrictl ${CRICTL_VERSION}
```
ganeshkumar5699, line 154: dowwnloadcrictl can also be refactored to use the download URL present in components.json. Since the URL is not expected to change soon (and we need to include parts like CPU arch), that refactoring can be done in a future PR.

alexeldeib (approved):

PR #2372, opened 2022-10-25
fix slow log rotation for /var/log/

cameronmeissner:
**What type of PR is this?**
/bug

Fixes an existing issue with slow log rotation for logs dumped in /var/log/. With these changes, we now invoke the logrotate service once every hour on the middle of the hour. Also added a size constraint on said logs:
- 100 MB for /var/log/syslog
- 50 MB for all others

We now also perform rotation on logs within /var/log/ on a daily basis regardless of size. In all cases, we maintain a max of 4 rotated logs on agent nodes at a time. These changes should prevent customers from experiencing excess disk pressure on their agent nodes.

Tested this locally in a standalone environment to verify that logs are in fact being rotated based on the modified logrotate config.

Fixes issue #[2802](https://github.com/Azure/AKS/issues/2802)


parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -313,14 +313,41 @@ write_files:
     {{GetMessageOfTheDay}}
 {{- end}}
 
-- path: /etc/logrotate.d/rsyslog-warn
+- path: /etc/logrotate.d/rsyslog
   permissions: "0644"
   owner: root
   content: |
+    /var/log/syslog
```
alexeldeib, line 320: can we do this in VHD? for the existing file as well?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -331,6 +358,14 @@ write_files:
             endscript
     }
 
+- path: /etc/cron.d/logrotate
```
alexeldeib, line 361: nit: I personally prefer systemd timers.

also isn't there an existing cronjob? is this overriding that?

cameronmeissner: Yep, this is one commonly used method of overriding the daily cron. I'll look into systemd timers.

PR #2381, opened 2022-10-27
improve slow log rotation in /var/log/

cameronmeissner:
**What type of PR is this?**
/bug

Fixes an existing issue with slow log rotation for logs dumped in /var/log/. With these changes, we now invoke the logrotate service once every hour. Also added a size constraint on said logs:
- 300 MB for /var/log/{syslog, messages, kern.log}
- 50 MB for all 11 others

This will have it such that, in the worst case, we'd have (3 * 2 * 300) + (11 * 2 * 75) + (grace period accumulation) = ~3450 MB of uncompressed logs at any given time before a rotation occurs.

We now also perform rotation on logs within /var/log/ on a daily basis regardless of size. We maintain 5 copies (4 compressed) of /var/log/{syslog, messages, kern.log}, and 7 copies of the others at a time. These changes should prevent customers from experiencing disk pressure on their agent nodes as a result of log pollution.

Tested this locally in a standalone environment to verify that logs are in fact being rotated based on the modified logrotate config.

Fixes #[2802](https://github.com/Azure/AKS/issues/2802)


pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line33.sh
```diff
@@ -177,6 +177,21 @@ if [[ ${ID} != "mariner" ]]; then
...
 
+# handle starting the logrotate-aks service responsible for log rotation
+if [ -f /etc/systemd/system/logrotate-aks.timer ] && [ -f /etc/systemd/system/logrotate-aks.service ]; then
+    # we're on a VHD that has definitions for the logrotate-aks service
+    logs_to_events "AKS.CSE.logrotate-aks" "systemctlEnableAndStart logrotate-aks.timer"
+    if [ -f /etc/cron.daily/logrotate ]; then
+        rm -f /etc/cron.daily/logrotate
+        echo "Removed /etc/cron.daily/logrotate"
+    fi
+    echo "Enabled and started logrotate-aks.timer, removed"
```
cameronmeissner, line 188: need to fix this typo...

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -273,6 +273,21 @@ if [[ ${ID} != "mariner" ]]; then
     /usr/bin/mandb && echo "man-db finished updates at $(date)" &
 fi
 
+# handle starting the logrotate-aks service responsible for log rotation
+if [ -f /etc/systemd/system/logrotate-aks.timer ] && [ -f /etc/systemd/system/logrotate-aks.service ]; then
```
alexeldeib, line 277: remove this, please do it in VHD only

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -496,6 +496,66 @@ ls -ltr /usr/local/bin/* >> ${VHD_LOGS_FILEPATH}
...
+/var/log/user.log
+/var/log/lpr.log
+/var/log/cron.log
+/var/log/debug
+/var/log/messages
+/var/log/warn
+{
+  rotate 4
+  daily
+  size 25M
```
alexeldeib, line 531: did we determine this plays nicely with daily vs the (poorly documented) `maxsize`?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -496,6 +496,66 @@ ls -ltr /usr/local/bin/* >> ${VHD_LOGS_FILEPATH}
...
+  delaycompress
+  sharedscripts
+  postrotate
+      /usr/lib/rsyslog/rsyslog-rotate
+  endscript
+}
+EOF
+
+# populate the logrotate-aks unit and timer definitions
+cat << EOF >> /etc/systemd/system/logrotate-aks.service
```
alexeldeib, line 544: okay, it would be nice if these were in the packer json admittedly, a little unwieldy in bash

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -496,6 +496,66 @@ ls -ltr /usr/local/bin/* >> ${VHD_LOGS_FILEPATH}
...
+[Unit]
+Description=runs the logrotate utility for log rotation
+[Service]
+ExecStart=/usr/sbin/logrotate /etc/logrotate.conf
+EOF
+cat << EOF >> /etc/systemd/system/logrotate-aks.timer
+[Unit]
+Description=a timer that runs the logrotate-aks on the top of every hour of every day
```
alexeldeib, line 552: may want to randomize this a bit

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: Yep, from the manual testing I've done the size constraint is mutually exclusive from the daily - so if it does grow too large before the daily time frame is hit it will be rotated regardless.

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: you're saying we can and should start the timer in the VHD itself?

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: trying with RandomizedDelaySec=60s

alexeldeib (approved):

parts/linux/cloud-init/artifacts/aks-rsyslog
```diff
@@ -0,0 +1,40 @@
+/var/log/syslog
+/var/log/messages
+/var/log/kern.log
+{
+  rotate 5
+  daily
+  size 300M
```
alexeldeib, line 7: was this default? or just to keep a bit more? fine either way, just checking

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -555,3 +559,7 @@ if [[ ${UBUNTU_RELEASE} == "18.04" || ${UBUNTU_RELEASE} == "22.04" ]]; then
 fi
 
 
+<<<<<<< HEAD
```
alexeldeib, line 562: git conflict?

vhdbuilder/packer/packer_source.sh
```diff
@@ -65,6 +65,14 @@ copyPackerFiles() {
...
   CI_SYSLOG_WATCHER_SCRIPT_SRC=/home/packer/ci-syslog-watcher.sh
   CI_SYSLOG_WATCHER_SCRIPT_DEST=/usr/local/bin/ci-syslog-watcher.sh
+  AKS_LOGROTATE_SCRIPT_SRC=/home/packer/aks-logrotate.sh
+  AKS_LOGROTATE_SCRIPT_DEST=/usr/local/bin/aks-logrotate.sh
+  AKS_LOGROTATE_SERVICE_SRC=/home/packer/aks-logrotate.service
+  AKS_LOGROTATE_SERVICE_DEST=/etc/systemd/system/aks-logrotate.service
+  AKS_LOGROTATE_TIMER_SRC=/home/packer/aks-logrotate.timer
+  AKS_LOGROTATE_TIMER_DEST=/etc/systemd/system/aks-logrotate.timer
+  AKS_LOGROTATE_CONF_SRC=/home/packer/aks-rsyslog
+  AKS_LOGROTATE_CONF_DEST=/etc/logrotate.d/aks-rsyslog
```
alexeldeib, line 75: I suppose this automatically gets picked up by logrotate when the timer invokes it?

parts/linux/cloud-init/artifacts/aks-rsyslog
cameronmeissner: weekly -> daily, rotate 4 -> rotate 5, added size 300M

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: should be taken care of

vhdbuilder/packer/packer_source.sh
cameronmeissner: yep yep, /etc/logrotate.conf has an include directive that brings in the dropins from /etc/logrotate.d/, which in this case will incluse aks-rsyslog which is overriding rsyslog for our purposes.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -41,11 +41,15 @@ systemctlEnableAndStart disk_queue || exit 1
 mkdir /opt/certs
 chmod 666 /opt/certs
 systemctlEnableAndStart update_certs.path || exit 1
-systemctlEnableAndStart update_certs.timer || exit 1
+systemctlEnableAndStart update_certs.timer || exit
```
alexeldeib, line 44: where did the 1 go :o

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -41,11 +41,15 @@ systemctlEnableAndStart disk_queue || exit 1
 mkdir /opt/certs
 chmod 666 /opt/certs
 systemctlEnableAndStart update_certs.path || exit 1
```
alexeldeib, line 43: dumb question: does enabling the timer automatically enable the service? I think it is directly enable-able, right?

honestly not sure

alexeldeib (approved):

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: haven't been able to find definitive answers online, I'm not sure either, but from what I've observed the target service does indeed get invoked by the timer after the timer has been enabled and started

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: ended up switching to maxsize :)

PR #2393, opened 2022-10-31
feat: support `externalTrafficPolicy: Cluster` with IPv6 LoadBalancer service

tyler-lloyd:
**What type of PR is this?**

/kind feature

moved from #2334 

In AKS, nodes act as a router and forward traffic to the container's network namespace. Per RFC 4193, "Routers must not forward any packets with Link-Local source or destination addresses to other links", which means when the link-local source packet arrives on the node to probe for the health of the service, the node cannot forward that packet to a pod network namespace running on that node. This prevents k8s LoadBalancer services from using `externalTrafficPolicy: Cluster` because Kubernetes is attempting to determine the health of pods backing a service directly (versus ETP: Local, where the health probe is replied to by kube-proxy itself on the node which doesn't need to forward health probe packets).

By changing the src IP of the probe packet to the globally unique address mentioned above, nodes are able to forward the probe packet to the correct destination network namespace thus enabling the use of `externalTrafficPolicy: Cluster` for IPv6 services in a dual-stack AKS cluster.

This PR ~~includes a new step in CSE, `configureCNINFTables`, which runs as part of the CNI setup function and~~ will create nftables rules for re-writing the source IP of IPv6 packets from the Azure Load Balancer probe's address link-local address, `fe80::1234:5678:9abc`, to a globally unique address `2603:1062:0:1:fe80:1234:5678:9abc`.

update: Instead of putting a new step in CSE, we're adding a systemd unit to run at boot every time. This ensures the rules are written between restarts and bakes the service into the VHD. The service will run a script to query IMDS for to see if the instance has an IPv6 private address and if so, writes the nftables rules. Otherwise it is a no-op.

The new rules try to be as non-disruptive as possible and are setup as stateless NAT rules, meaning no connection tracking occurs.

**These new rules will only be written for clusters that have IPv6 enabled** (currently only dual-stack).

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

- Adds `nftables` to image dependencies. Happy to discuss if we should have these rules just be written with iptables but there might be advantages to trying to move to nftables in the future as it offers better performance and priority.
- Updates the KubernetesConfig model to include recent updates for IPv6 clusters (ipFamilies, clusterSubnets, and serviceCIDRs)


parts/linux/cloud-init/artifacts/ipv6_nftables.sh
```diff
@@ -0,0 +1,56 @@
...
+#         {
+#           "privateIpAddress": "fd85:534e:4cd6:ab02::5"
+#         }
+#       ]
+#     },
+#     "macAddress": "000D3A98DA20"
+#   }
+# ]
+
+# every 60 min, query IMDS and update nftables rules
```
tyler-lloyd, line 37: not a fan of this right now. need to understand why our rules aren't working after 2 hours.

alexeldeib (approved):

parts/linux/cloud-init/artifacts/ipv6_nftables
```diff
@@ -0,0 +1,12 @@
+define slb_lla = fe80::1234:5678:9abc
+define slb_gua = 2603:1062:0:1:fe80:1234:5678:9abc
+
+add table ip6 azureSLBProbe
+
+flush table ip6 azureSLBProbe
```
tyler-lloyd, line 6: this wipes the rules in the table before writing them. Only here because we're reconciling this table every hour and for an unknown reason need to reset the rules for the probe packets to flow correctly.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -352,6 +353,11 @@ for CNI_PLUGIN_VERSION in $CNI_PLUGIN_VERSIONS; do
     echo "  - CNI plugin version ${CNI_PLUGIN_VERSION}" >> ${VHD_LOGS_FILEPATH}
 done
 
+# IPv6 nftables, only Ubuntu for now
+if [[ $OS == $UBUNTU_OS_NAME ]]; then
```
tyler-lloyd, line 357: @hbeberman - FYI when I tried testing this on a mariner nodepool I wasn't able to install nftables. Any chance we could get it added in the future?

```sh
> root@aks-mariner-38336011-vmss000000 [ / ]# yum install nftables
Loaded plugin: tdnfrepogpgcheck
No package nftables available
Error(1011) : No matching packages
```

alexeldeib: discussed offline, fine to scope to Ubuntu for now and follow up.

parts/linux/cloud-init/artifacts/ipv6_nftables.sh
tyler-lloyd: fixed

parts/linux/cloud-init/artifacts/ipv6_nftables.sh
```diff
@@ -0,0 +1,47 @@
...
+# check the number of IPv6 addresses this instance has from IMDS
+IPV6_ADDR_COUNT=$(curl -sSL -H "Metadata: true" "http://169.254.169.254/metadata/instance/network/interface?api-version=2021-02-01" | \
+    jq '[.[].ipv6.ipAddress[] | select(.privateIpAddress != "")] | length')
+
+if [[ $IPV6_ADDR_COUNT -eq 0 ]];
+then
+    echo "instance is not configured with IPv6, skipping nftables rules"
+else
+    echo "writing nftables from $NFTABLES_RULESET_FILE"
+    nft -f $NFTABLES_RULESET_FILE
```
alexeldeib, line 46: no `-set -e` so this never actually exits non-zero on fails? (bash will exit 0?)

tyler-lloyd: updated with set -e

parts/linux/cloud-init/artifacts/ipv6_nftables
```diff
@@ -0,0 +1,18 @@
+define slb_lla = fe80::1234:5678:9abc
```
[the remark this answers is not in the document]
tyler-lloyd: nope, that's better. updated.

PR #2422, opened 2022-11-10

HDYA:


andyliuliming (approved):

vhdbuilder/packer/components.json
```diff
@@ -278,8 +278,8 @@
       ],
       "multiArchVersions": [
         "v1.23.0",
-        "v1.23.0.1",
```
andyliuliming, line 281: "v1.23.0.1", not in use?

HDYA: yes, it was skipped when (another) Andy bump to 1.24

PR #2436, opened 2022-11-14
chore: Set Windows SIG image verions to fake values

AbelHu:
**What type of PR is this?**
/kind feature

Set Windows SIG image verions to fake values

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


ShiqianTao (approved):

junjiezhang1997 (approved):

alexeldeib (approved):
lgtm assuming rp e2e failures are unrelated

would be nice to make it even more obviously fake -- to a new dev, it's very confusing that these images wouldn't work right. maybe "SHOULD_NEVER_BE_USED" or something similarly obvious

but I don't know if we rely on the images being reasonable on RP side

AbelHu:
> lgtm assuming rp e2e failures are unrelated
> 
> would be nice to make it even more obviously fake -- to a new dev, it's very confusing that these images wouldn't work right. maybe "SHOULD_NEVER_BE_USED" or something similarly obvious
> 
> but I don't know if we rely on the images being reasonable on RP side

I think that the e2e failure is not related to this change. We have a wiki about how to update the sig image versions in AgentBaker and will update it again after merging this PR

PR #2440, opened 2022-11-16
fix: Mariner-specific update_certs.sh

dgarrett-microsoft:
**What type of PR is this?**
/kind bug

Mariner and RPM-based systems have a different method of ingesting trusted CA certs. The cert needs to be added to a different path and then `update-ca-trust` needs to be run instead of `update-ca-certificates`

See [Fedora documentation](https://docs.fedoraproject.org/en-US/quick-docs/using-shared-system-certificates/) or the [`update-ca-trust` man page](https://www.mankier.com/8/update-ca-trust)

Fixes ADO OS/41969197

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

I was able to test my changes adding and removing a CA cert successfully.
I followed this to create a valid CA cert: https://www.digitalocean.com/community/tutorials/how-to-set-up-and-configure-a-certificate-authority-ca-on-ubuntu-20-04
And I followed this to create an AKS cluster using this custom CA feature: https://learn.microsoft.com/en-us/azure/aks/custom-certificate-authority

I then checked:
ls /opt/certs/
ls /usr/share/pki/ca-trust-source/anchors/
journalctl -u update_certs
trust list | grep Easy-RSA

to see the cert added to the node and successfully ingested.
I used the same commands to watch it get deleted if I removed it from custom-ca-trust-secret

One difference from the original update_certs.sh is that `update-ca-trust` doesn't support certs in subdirectories, so I wasn't able to copy the automated certs into a specific subdirectory and avoid messing with any existing certs.

**Release note**:
```
none
```


dgarrett-microsoft:
> One difference from the original update_certs.sh is that `update-ca-trust` doesn't support certs in subdirectories, so I wasn't able to copy the automated certs into a specific subdirectory and avoid messing with any existing certs.

I'm considering getting a similar result by prepending cert filenames copied to $certDestination with a string like `aks-custom-` and then only deleting those files. Thoughts?

alexeldeib:
> I'm considering getting a similar result by prepending cert filenames copied to $certDestination with a string like aks-custom- and then only deleting those files. Thoughts?

this sounds like a good idea. 

it looks like besides the subdirectory, the only changes are 1) the command, and 2) the source dir the distro uses for certs? I wonder if we can just align on the `aks-custom` naming somehow for all cases?

@utheman previously used timestamps for versioning, maybe we can combined that + unique naming, and parameterize the command + dir and share one script?

UtheMan:
Generally LGTM, I really like the idea of combining both Ubuntu and Mariner scripts. 
One thing I'm worried about is that updated custom CA takes CAs passed during PUT MC request and creates files with zeroed "timestamp" to indicate that its the CAs passed during boot time (see [here](https://github.com/Azure/AgentBaker/blob/de4b5f10f99a811e020a46d48a6876a233a046c1/parts/linux/cloud-init/artifacts/cse_config.sh#L101))

I'd need to verify that this will still be working as expected after those changes (files created properly and removed properly after update)

parts/linux/cloud-init/artifacts/update_certs.sh
```diff
@@ -2,13 +2,18 @@
 set -uo pipefail
 
 certSource=/opt/certs
-certDestination=/usr/local/share/ca-certificates/certs
+certDestination="${1:-/usr/local/share/ca-certificates/certs}"
+updateCmd="${2:-update-ca-certificates -f}"
+destPrefix="aks-custom-"
```
UtheMan, line 8: We should create certDestination dir here, as the new way of copying files from source will not create it automatically

UtheMan: [ ! -d "$certDestination" ] && mkdir "$certDestination"

UtheMan:
```suggestion
[ ! -d "$certDestination" ] && mkdir "$certDestination"
```

UtheMan (approved):

PR #2447, opened 2022-11-21
lowering the timeout for initial mcr check

chmill-zz:
**What type of PR is this?**

Believe this long time out is making us land in the error 124, but really it's a mcr timeout.

If I can do math we're still going to wait/retry for ~5min. Which is excessive still, but gets us under the 15min window we've set ourselves for a timeout.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):
I might bump it a bit higher, to keep it close to the timeout limit while still keeping the whole script length below the all-up timeout. 

but fine with this and iterating

PR #2452, opened 2022-11-23
reboot packer VM multiple times during VHD builds

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/vhd-image-builder-mariner-gen2.json
```diff
@@ -324,9 +338,26 @@
     {
       "type": "shell",
       "inline": [
+        "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} HYPERV_GENERATION={{user `hyperv_generation`}} CONTAINER_RUNTIME={{user `container_runtime`}} TELEPORTD_PLUGIN_DOWNLOAD_URL={{user `teleportd_plugin_download_url`}} ENABLE_FIPS={{user `enable_fips`}} SGX_INSTALL={{user `sgx_install`}} /bin/bash -ux /home/packer/pre-install-dependencies.sh",
```
alexeldeib, line 341: shouldn't there be a 2nd reboot between these for fips? or am I missing something?

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -0,0 +1,63 @@
...
+echo "kubelet/kubectl downloaded:" >> ${VHD_LOGS_FILEPATH}
+ls -ltr /usr/local/bin/* >> ${VHD_LOGS_FILEPATH}
+
+# shellcheck disable=SC2010
+ls -ltr /dev/* | grep sgx >>  ${VHD_LOGS_FILEPATH} 
+
+echo -e "=== Installed Packages Begin\n$(listInstalledPackages)\n=== Installed Packages End" >> ${VHD_LOGS_FILEPATH}
+
+echo "Disk usage:" >> ${VHD_LOGS_FILEPATH}
+df -h >> ${VHD_LOGS_FILEPATH}
```
alexeldeib, line 30: suggest to re-run some of our cleanup after a reboot - should strip old kernels etc.

```
  apt-get -y autoclean || exit 1
  apt-get -y autoremove --purge || exit 1
  apt-get -y clean || exit 1
```

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -0,0 +1,63 @@
...
+} >> ${VHD_LOGS_FILEPATH}
+
+if [[ $(isARM64) != 1 ]]; then
+  # no asc-baseline-1.1.0-268.arm64.deb
+  installAscBaseline
+fi
+
+if [[ ${UBUNTU_RELEASE} == "18.04" || ${UBUNTU_RELEASE} == "22.04" ]]; then
+  if [[ ${ENABLE_FIPS,,} == "true" || ${CPU_ARCH} == "arm64" ]]; then
+    relinkResolvConf
```
alexeldeib, line 59: this...is odd. not related to you. but why is this conditioned on arm or fips?

should be done for all of 18, 20, and 22.04. probably already done elsewhere...

vhdbuilder/packer/init-variables.sh
```diff
@@ -419,7 +419,9 @@ cat <<EOF > vhdbuilder/packer/settings.json
   "windows_sigmode_source_resource_group_name": "${windows_sigmode_source_resource_group_name}",
   "windows_sigmode_source_gallery_name": "${windows_sigmode_source_gallery_name}",
   "windows_sigmode_source_image_name": "${windows_sigmode_source_image_name}",
-  "windows_sigmode_source_image_version": "${windows_sigmode_source_image_version}"
+  "windows_sigmode_source_image_version": "${windows_sigmode_source_image_version}",
+  "vnet_name": "nodesig-pool-vnet",
```
alexeldeib, line 423: did you test windows? if so nice, otherwise would leave this alone until @AbelHu or someone else can take a look. I hope it "just works" but not 100% sure how the network topology looks there.

cameronmeissner: haven't tested on windows as of yet

AbelHu: @cameronmeissner could you help to introduce what this PR is for?
When building Windows VHDs,
- we have used a pre-created vnet: https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/init-variables.sh#L39-L66
- we have rebooted the build VM: https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/windows-vhd-builder.json#L76

alexeldeib: this is for 1es build pools. we pre-create vnet with **no inbound connectivity**. this fixes our security exception!

the 1es build agents are in the vnet Cameron specified here. it doesn't look like windows nsg has anything special, just allow WinRM SSL? that should work here, but if you're already pre-creating a vnet, likely we could refactor a bit more.

AbelHu: Yes. we pre-create the vnet for Windows VHD build VM since we need to add the tag `SkipASMAzSecPack=True` and allow WinRM with SSL 5986. It seems like that Windows also can reuse this PR to avoid the security exception if we use the vnet Cameron specified here and 1es build pools. Please free to correct me if my understanding is wrong

AbelHu: can we create nodesig-pool-vnet dynamically? Is it ok to use the same vent?

cameronmeissner: do you mean dynamically creating nodesig-pool-vnet on demand or something? pretty sure as long as there isn't already a vnet with the same name in the target resource group it should be fine. although, I think a simpler approach would be to just modify the bicep templates we're using for deploying the 1es pools to add another subnet within nodesig-pool-vnet that we can then point to from the packer templates in a similar fashion as we're doing now for linux.

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -0,0 +1,80 @@
...
+source /home/packer/provision_source_distro.sh
+source /home/packer/tool_installs.sh
+source /home/packer/tool_installs_distro.sh
+
+CPU_ARCH=$(getCPUArch)  #amd64 or arm64
+VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
+
+if [[ $OS == $MARINER_OS_NAME ]]; then
+  if grep -q "kata" <<< "$FEATURE_FLAGS"; then
+    enableMarinerKata
```
alexeldeib, line 17: shouldn't we do this before rebooting, since it generates a new grub config iirc?

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -0,0 +1,80 @@
...
+  # shellcheck disable=SC2021
+  current_kernel="$(uname -r | cut -d- -f-2)"
+  dpkg --get-selections | grep -e "linux-\(headers\|modules\|image\)" | grep -v "$current_kernel" | tr -s '[[:space:]]' | tr '\t' ' ' | cut -d' ' -f1 | xargs -I{} apt-get remove -yq {}
+
+  # strip old kernels/packages
+  apt-get -y autoclean || exit 1
+  apt-get -y autoremove --purge || exit 1
+  apt-get -y clean || exit 1
+  # remove apport
+  apt-get purge --auto-remove apport open-vm-tools -y
```
alexeldeib, line 34: do this before autoclean/autoremove/clean so those commands catch the transitive deps which will no longer be required

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -0,0 +1,80 @@
...
+ls -ltr /dev/* | grep sgx >>  ${VHD_LOGS_FILEPATH} 
+
+echo -e "=== Installed Packages Begin\n$(listInstalledPackages)\n=== Installed Packages End" >> ${VHD_LOGS_FILEPATH}
+
+echo "Disk usage:" >> ${VHD_LOGS_FILEPATH}
+df -h >> ${VHD_LOGS_FILEPATH}
+
+
+# warn at 75% space taken
+[ -s $(df -P | grep '/dev/sda1' | awk '0+$5 >= 75 {print}') ] || echo "WARNING: 75% of /dev/sda1 is used" >> ${VHD_LOGS_FILEPATH}
```
alexeldeib, line 51: unrelated: can you create a backlog task to do two things here:
- use disk detection by label for OS disk (/dev/sda not reliable!)
- check capacity of OS disk by label
- hardcode max capacity as 30 GB. I want to change os disk size to 512GB to alleviate IO latency during VHD builds. So need to avoid relying on the disk size itself, only current util.

alexeldeib: not blocking for merge

vhdbuilder/packer/vhd-image-builder-mariner-gen2.json
alexeldeib: ah, touche

vhdbuilder/packer/vhd-image-builder-mariner-gen2.json
```diff
@@ -328,9 +338,26 @@
...
         "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} HYPERV_GENERATION={{user `hyperv_generation`}} CONTAINER_RUNTIME={{user `container_runtime`}} TELEPORTD_PLUGIN_DOWNLOAD_URL={{user `teleportd_plugin_download_url`}} ENABLE_FIPS={{user `enable_fips`}} /bin/bash -ux /home/packer/install-dependencies.sh"
       ]
     },
+    {
+      "type": "shell",
+      "inline": [
+        "set -x",
```
alexeldeib, line 348: not really necessary for a simple one line reboot..

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -0,0 +1,80 @@
...
+  else
+    # Leave automatic package update disabled for the kata image
+    enableDNFAutomatic
+  fi
+fi
+
+if [[ $OS == $UBUNTU_OS_NAME ]]; then
+  # shellcheck disable=SC2021
+  current_kernel="$(uname -r | cut -d- -f-2)"
+  dpkg --get-selections | grep -e "linux-\(headers\|modules\|image\)" | grep -v "$current_kernel" | tr -s '[[:space:]]' | tr '\t' ' ' | cut -d' ' -f1 | xargs -I{} apt-get remove -yq {}
```
alexeldeib, line 27: please sanity check the running kernel listed in the final release notes vs the apt packages in the release notes // the logs from these steps

I want to be super careful to make sure we 1) reboot before this and 2) never remove the latest kernel // make sure apt/dpkg/ubuntu fail _hard_ if we do so by accident.

vhdbuilder/packer/vhd-image-builder-mariner-gen2.json
cameronmeissner: will remove, was mainly for debugging anyways

vhdbuilder/packer/post-install-dependencies.sh
cameronmeissner: new items created

vhdbuilder/packer/post-install-dependencies.sh
alexeldeib: pretty sure we can at least swap this one sku over to standard_d4ads_v5 without issue, and we could probably swap everything (I think default quota is 100 cores, need to double check tho)

alexeldeib: should be able to test easily

cameronmeissner: looks like this works, will use this approach for now and can hopefully upgrade the vm sizes for other vhd builds in the future as well

PR #2456, opened 2022-11-25
upgrade Azure disk csi driver to v1.25.0

andyzhangx:
**What type of PR is this?**
/kind feature

upgrade Azure disk csi driver to v1.25.0

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
upgrade Azure disk csi driver to v1.25.0
```


vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -48,8 +48,8 @@ if ($windowsSku -eq "2019") {
             "mcr.microsoft.com/oss/kubernetes-csi/livenessprobe:v2.6.0",
             "mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar:v2.4.0",
             "mcr.microsoft.com/oss/kubernetes-csi/csi-node-driver-registrar:v2.5.0",
-            "mcr.microsoft.com/oss/kubernetes-csi/azuredisk-csi:v1.23.0",
```
AbelHu, line 51: Has we upgraded it from 1.23 to 1.24 in production? If no, we should still keep the old version until v1.23.0 is not used in production.

andyzhangx: yes, 1.24 is current version in production now which is released one month ago.

AbelHu: I think that we should cache the new image version for better perf next time and then update the new version in production.

AbelHu (approved):
Has we upgraded it from 1.23 to 1.24 in production? If no, we should still keep the old version until v1.23.0 is not used in production.

PR #2475, opened 2022-12-01
remove dangling shas leftover from image deletion using ctr

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -338,7 +338,14 @@ removeContainerImage() {
     CLI_TOOL=$1
     CONTAINER_IMAGE_URL=$2
     if [[ ${CLI_TOOL} == "ctr" ]]; then
-        ctr --namespace k8s.io image rm $CONTAINER_IMAGE_URL
+        IFS=':'
+        read -ra urlParts <<< "${CONTAINER_IMAGE_URL}"
+        url="${urlParts[0]}"
+        tag="${urlParts[1]}"
+
+        sha=$(crictl images --digests | grep "${url}" | grep "${tag}" | tr -s ' ' | cut -d' ' -f 3)
```
alexeldeib, line 346: if we're going to depend on crictl, just `crictl rmi` instead of this foo?

alexeldeib: cli_tool used to be how we differentiated, because it was only set to crictl if the install succeeded at runtime lol. but it should be fine in vhd now.

cameronmeissner: okay that makes sense, I guess we can just simplify this to only pivot based off docker?

cameronmeissner: also, do we need to be worried about this?

https://github.com/Azure/AgentBaker/blob/91a34ad5b28d510fad33689c80b0176927437a54/parts/linux/cloud-init/artifacts/cse_install.sh#:~:text=%23%20always%20use,removes%20too%20much.

alexeldeib:
That only applies to retagged images where we have 2 tags to 1 sha, and
want to keep 1 tag + 1 sha but delete 1 tag (this is for China MCR vs
normal MCR)

alexeldeib (approved):

alexeldeib:
hmm vhd build issues?

cameronmeissner:
> hmm vhd build issues?

looked transient I believe, rerunning now

PR #2482, opened 2022-12-06
feat: one Windows VHD build mode

wanqingfu:
**What type of PR is this?**
Merge gen1, gen2, and sig modes into one mode for Windows VHD builds - use sig only.

/kind feature

Let packer output a managed image for all Windows VHD builds.

Collapse all existing modes to one sig mode, even for default.

Pros:
- to simplify build mode so that it's easier to maintain in the long run
- we can have cleaner packer templates

Also added some logic to do some sort of backfill deletion on older managed image/SIG versions/packer etc.
Get rid of those empty galleries / galleries with image-definitions all empty in the backfill deletion step.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

- Cleaning outdated provisioned resources is only performed while not in dry-run mode (production pipeline) or when `SKIP_IMAGE_DELETION_BACKFILL` is overridden to `False` e.g. in testing/PR-check in pipeline.
- Outdated resources mean galleries and associated SIG image defs/versions basically.
- backfill deletion was set as a parallel stage(_job/task_) to reduce execution time in monthly VHD building.

**Release note**:
```
none
```


vhdbuilder/packer/windows-image.env
```diff
@@ -19,5 +19,5 @@ WINDOWS_2022_GEN2_BASE_IMAGE_VERSION=20348.1366.221207
 
 # Please uncomment the following lines and set a larger os disk size that is at least 30GB when your PR check-in fails
 # WINDOWS_2019_OS_DISK_SIZE_GB=30
-# WINDOWS_2019_CONTAINERD_OS_DISK_SIZE_GB=30
-# WINDOWS_2022_CONTAINERD_OS_DISK_SIZE_GB=30
+WINDOWS_2019_CONTAINERD_OS_DISK_SIZE_GB=40
```
AbelHu, line 22: Should we increase the vhd size for one build mode?

vhdbuilder/packer/init-variables.sh
```diff
@@ -334,18 +338,18 @@ if [ "$OS_TYPE" == "Windows" ]; then
 		WINDOWS_IMAGE_VERSION=""
 
 		# Need to use a sig image to create the build VM
-		if [[ "$MODE" == "sigMode" ]]; then
-			# Reuse IMPORTED_IMAGE_NAME so the shared code in cleanup.sh can delete the temporary resource
-			IMPORTED_IMAGE_NAME=$imported_windows_image_name
-			echo "Creating new image for imported vhd ${WINDOWS_IMAGE_URL}"
+		if [[ "$MODE" == "windowsVhdMode" ]]; then
+			# a. create a new managed image $IMPORTED_IMAGE_NAME from os disk source $IMPORTED_IMAGE_URL
```
AbelHu, line 342: Suggest not using a. b. c. d. in multiple comments and it need additional work to update them if another step is added in future.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -235,14 +205,12 @@ steps:
       ${CONTAINER_IMAGE} make -f packer.mk generate-sas
     displayName: Getting Shared Access Signature URI
     condition: and(succeeded(), eq(variables.DRY_RUN, 'False'), 
-                   or(and(eq(variables['MODE'], 'sigMode'), eq(variables.GEN2_SIG_FOR_PRODUCTION, 'True')), 
-                      eq(variables['MODE'], 'default')))
+                   and(eq(variables['MODE'], 'windowsVhdMode'), eq(variables.SIG_FOR_PRODUCTION, 'True')))
```
AbelHu, line 208: do we have only one mode after this PR is merged? If so, maybe we can remove this check `eq(variables['MODE'], 'windowsVhdMode')`

packer.mk
```diff
@@ -60,20 +60,21 @@ else
...
+endif
+else
+ifeq (${HYPERV_GENRATION},V1)
+	@echo "${MODE}: Building with Hyper-v generation 1 VM and save to Shared Image Gallery"
 else
-	@echo "${MODE}: Building and save to Shared Image Gallery"
+	@echo "${MODE}: Building with Hyper-v generation 2 VM and save to Shared Image Gallery"
 endif
 endif
 	@packer build -var-file=vhdbuilder/packer/settings.json vhdbuilder/packer/windows-vhd-builder-sig.json
```
AbelHu, line 77: Should we remove this line `@packer build -var-file=vhdbuilder/packer/settings.json vhdbuilder/packer/windows-vhd-builder-sig.json` and the file `vhdbuilder/packer/settings.json vhdbuilder/packer/windows-vhd-builder-sig.json`?

vhdbuilder/packer/cleanup.sh
```diff
@@ -181,7 +181,58 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN,,}" == "false" ]]; then
...
+# attempt to clean up Windows managed images and SIG image versions created over a week ago in SIG_GALLERY_NAME (except the production gallery)
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "False" && "${SIG_GALLERY_NAME}" != "AKSWindows" ]]; then
+  echo "Looking for Windows managed images in ${AZURE_RESOURCE_GROUP_NAME} created over ${EXPIRATION_IN_HOURS} hours ago..."
+
+  managed_image_ids=""
+  sig_version_ids=""
+
+  # delete outdated Windows managed images and associated SIG versions (.tags.os must be "Windows")
+  images=$(az image list -g ${AZURE_RESOURCE_GROUP_NAME} | jq --arg dl $deadline '.[] | select(.tags.os == "Windows") | select(.tags.now < $dl).name' | tr -d '\"' || "")
+  for image in $images; do
```
AbelHu, line 193: Do we only delete the managed images and sig versions for the last image since line 199-211 is not in the for loop?

vhdbuilder/packer/cleanup.sh
```diff
@@ -181,7 +181,58 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN,,}" == "false" ]]; then
...
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "False" ]]; then
+  gallery_lists=$(az sig list -g ${AZURE_RESOURCE_GROUP_NAME} | jq '.[].name' | tr -d '\"' || "")
+  for gallery in $gallery_lists; do
+    if [[ $gallery == "AKSWindows" ]]; then
+        continue
+    fi
+
+    # delete old Windows SIG image versions in gallery (image definitions must have .osType == "Windows")
+    image_defs=$(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq '.[] .name' | tr -d '\"' || "")
+    for image_definition in $(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq '.[] | select(.osType == "Windows").name' | tr -d '\"' || ""); do
```
AbelHu, line 224: I think that we should avoid list all image-definition in every for loop.
1. It will increase the finish time
2. it may not exit if some PRs are being submitted to create new image definitions continously.

vhdbuilder/packer/cleanup.sh
```diff
@@ -181,7 +181,58 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN,,}" == "false" ]]; then
   set -x
 fi
 
-#clean up storage account created over a week ago
+# attempt to clean up Windows managed images and SIG image versions created over a week ago in SIG_GALLERY_NAME (except the production gallery)
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "False" && "${SIG_GALLERY_NAME}" != "AKSWindows" ]]; then
```
AbelHu, line 185: we may not run so many cleanup steps in `"${DRY_RUN}" == "False"` which means that it will be run only when we try to publish new Windows images. It will increase additional cost for us.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -4,37 +4,26 @@ parameters:
...
-#               SIG_IMAGE_NAME, SIG_GALLERY_NAME, and SIG_IMAGE_VERSION will be passed below.
-#               Task variable GEN2_SIG_FOR_PRODUCTION is set to False and passed to the following steps.
-#               Built sig will not be deleted because it is for future testing.
-#            if we want to publish the VHD, instead of using pipeline variables,
-#               we will hard-code the task variables SIG_GALLERY_NAME, SIG_IMAGE_NAME, and SIG_IMAGE_VERSION.
-#               Task variable GEN2_SIG_FOR_PRODUCTION is set to True and passed to the following steps.
-#               Built sig will be deleted because it has been converted to VHD, and thus not needed.
-#               Use the suffix $RANDOM to avoid duplication while running concurrent builds.
+# Merge gen1, gen2, and sig modes into one mode for Windows VHD builds - use sig only.
+# 1. If sig is for test purpiose only, SIG_GALLERY_NAME, SIG_IMAGE_NAME_PREFIX, and SIG_IMAGE_VERSION are set.
```
cameronmeissner, line 8: typo: purpiose -> purpose

vhdbuilder/packer/cleanup.sh
cameronmeissner: 199-211 use the string of managed image and SIG version IDs to actually run the delete operation, the loop just builds up those strings

vhdbuilder/packer/cleanup.sh
```diff
@@ -181,7 +181,59 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN,,}" == "false" ]]; then
...
+  done
+
+  if [[ -n "${managed_image_ids}" ]]; then
+    echo "Attempting to delete $(echo ${managed_image_ids} | wc -w) Windows managed images..."
+    az resource delete --ids ${managed_image_ids} > /dev/null || echo "Windows managed image deletion was not successful, continuing..."
+  else
+    echo "Did not find any Windows managed images eligible for deletion"
+  fi
+
+  if [[ -n "${sig_version_ids}" ]]; then
```
cameronmeissner, line 206: if you're planning on keeping the same strategy of generating a new unique gallery for each build, i'd add cleanup logic for old galleries themselves - I've seen a lot of instances of empty galleries / galleries with image-definitions that are all empty, etc.

vhdbuilder/packer/cleanup.sh
```diff
@@ -181,7 +181,59 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN,,}" == "false" ]]; then
...
+    echo "Attempting to delete $(echo ${sig_version_ids} | wc -w) Windows SIG image versions associated with old managed images..."
+    az resource delete --ids ${sig_version_ids} > /dev/null || echo "Windows SIG image version deletion was not successful, continuing..."
+  else
+    echo "Did not find any SIG versions associated with old Windows managed images eligible for deletion"
+  fi
+fi
+
+# attempt to clean up Windows SIG image versions in all galleries except "AKSWindows" created over a week ago
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "True" ]]; then
+  gallery_lists=$(az sig list -g ${AZURE_RESOURCE_GROUP_NAME} | jq '.[].name' | tr -d '\"' || "")
```
cameronmeissner, line 216: nit: gallery_lists -> gallery_list?

vhdbuilder/packer/cleanup.sh
```diff
@@ -193,4 +245,22 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "False" ]]; then
...
+# clean up old packer and vnet resource groups created over a week ago that didn't get cleaned in the building vhd step in case the pipeline fails
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "True" ]]; then
+  pkr_groups=$(az group list | jq --arg dl $deadline '.[] | select(.name | test("pkr-Resource-Group*")) | select(.tags.now < $dl).name' | tr -d '\"' || "")
+  for pkr_group in $pkr_groups; do
+      echo "Deleting packer resource group $pkr_group"
+      az group delete --name ${pkr_group} --yes 
+      echo "Deleted packer resource group $pkr_group"
+  done
+
+  # the old vnet groups do not have "now" in tags
```
cameronmeissner, line 259: maybe only do this in the windows case since we don't have these RGs for linux builds?

vhdbuilder/packer/cleanup.sh
```diff
@@ -193,4 +245,22 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "False" ]]; then
           echo "Deletion completed"
       fi
   done
-fi
\ No newline at end of file
+fi
+
+# clean up old packer and vnet resource groups created over a week ago that didn't get cleaned in the building vhd step in case the pipeline fails
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN}" == "True" ]]; then
```
cameronmeissner, line 251: I'd consider cleaning up old pkr resource groups irregardless of dry-run

cameronmeissner: hmm, I'm actually wondering if it would make more sense to have a separate pipeline variable that we could use to skip the image/SIG/image-definition deletion backfill steps, as opposed to overloading the functionality/meaning of dry-run mode. I feel like dry-run mode should just determine the nature of the output of the build itself, rather than the other miscellaneous cleanup it does during this step.

Basically, the new pipeline variable would default to false on the production pipelines, and default to true on the testing pipelines (which would also be able to be overridden). If the variable is set to true, we don't do the backfill deletion, otherwise we always attempt the backfill deletion. The variable could be named something like SKIP_RESOURCE_DELETION_BACKFILL or something.

this would be something we can maybe swing back around to after this PR is merged in

vhdbuilder/packer/init-variables.sh
```diff
@@ -141,8 +141,8 @@ fi
 
 echo "Using finalized SIG_IMAGE_NAME: ${SIG_IMAGE_NAME}, SIG_GALLERY_NAME: ${SIG_GALLERY_NAME}"
 
-# If we're building a Linux VHD or we're building a windows VHD in sigMode, ensure SIG resources
-if [[ "$MODE" == "linuxVhdMode" || "$MODE" == "sigMode" ]]; then
+# If we're building a Linux VHD or we're building a windows VHD in windowsVhdMode, ensure SIG resources
+if [[ "$MODE" == "linuxVhdMode" || "$MODE" == "windowsVhdMode" ]]; then
```
cameronmeissner, line 145: we can probably get rid of this conditional since we'll always be in either linuxVhdMode or windowsVhdMode now right?

though might be good to keep for sanity's sake...?

vhdbuilder/packer/init-variables.sh
```diff
@@ -334,18 +340,18 @@ if [ "$OS_TYPE" == "Windows" ]; then
 		WINDOWS_IMAGE_VERSION=""
 
 		# Need to use a sig image to create the build VM
-		if [[ "$MODE" == "sigMode" ]]; then
-			# Reuse IMPORTED_IMAGE_NAME so the shared code in cleanup.sh can delete the temporary resource
-			IMPORTED_IMAGE_NAME=$imported_windows_image_name
-			echo "Creating new image for imported vhd ${WINDOWS_IMAGE_URL}"
+		if [[ "$MODE" == "windowsVhdMode" ]]; then
```
cameronmeissner, line 343: can we also potentially get rid of this conditional?

vhdbuilder/packer/cleanup.sh
cameronmeissner: yeah I commented about this below, thinking we can potentially create a separate pipeline variable to dictate this behavior and not complicate the behavior or dry-run

cameronmeissner (approved):

vhdbuilder/packer/cleanup.sh
```diff
@@ -6,6 +6,14 @@ EXPIRATION_IN_HOURS=168
 # deadline = the "date +%s" representation of the oldest age we're willing to keep
 (( deadline=$(date +%s)-${expirationInSecs%.*} ))
 
+# we will clean old sig image-definitions/versions/galleries, RGs only if 
+# we have overriden the pipeline variable SKIP_IMAGE_DELETION_BACKFILL to False
```
AbelHu, line 10: Could we run the cleanup as a separate parallel task to avoid running them in all tasks and increase the duration when building VHDs for production?

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -0,0 +1,100 @@
...
+# this can be used in PR check-in pipelines together with a set SIG_GALLERY_NAME from which we'd like to free up resources
+if [[ "${MODE}" == "windowsVhdMode" && -n "${AZURE_RESOURCE_GROUP_NAME}" && "${CLEAN_OLD_WINDOWS_RESOURCES}" == "True" && "${SIG_GALLERY_NAME}" != "AKSWindows" ]]; then
+  echo "Looking for Windows managed images in ${AZURE_RESOURCE_GROUP_NAME} created over ${EXPIRATION_IN_HOURS} hours ago..."
+
+  managed_image_ids=""
+  sig_version_ids=""
+
+  # delete outdated Windows managed images and associated SIG versions (.tags.os must be "Windows")
+  images=$(az image list -g ${AZURE_RESOURCE_GROUP_NAME} | jq --arg dl $deadline '.[] | select(.tags.os == "Windows") | select(.tags.now < $dl).name' | tr -d '\"' || "")
+  for image in $images; do
```
AbelHu, line 27: is it expected to always delete the last image after a for loop?

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -0,0 +1,100 @@
...
+(( expirationInSecs = ${EXPIRATION_IN_HOURS} * 60 * 60 ))
+# deadline = the "date +%s" representation of the oldest age we're willing to keep
+(( deadline=$(date +%s)-${expirationInSecs%.*} ))
+
+# we will clean old sig image-definitions/versions/galleries, RGs only if 
+# we have overriden the pipeline variable SKIP_IMAGE_DELETION_BACKFILL to False
+# or we are running on production pipelines
+CLEAN_OLD_WINDOWS_RESOURCES="False"
+if [[ "${SKIP_IMAGE_DELETION_BACKFILL}" == "False" || "${DRY_RUN}" == "False" ]]; then
+  CLEAN_OLD_WINDOWS_RESOURCES="True"
```
AbelHu, line 14: could we exit with 0 if we do not need to cleanup the old resources?

AbelHu: Since this is only used by Windows, it may be better to exit with 0 when MODE is not windowsVhdMode

vhdbuilder/packer/backfill-cleanup.sh
cameronmeissner: the for-loop builds up the string of resource IDs that then get deleted by a single `az resource delete` afterwards

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -0,0 +1,104 @@
...
+(( expirationInSecs = ${EXPIRATION_IN_HOURS} * 60 * 60 ))
+# deadline = the "date +%s" representation of the oldest age we're willing to keep
+(( deadline=$(date +%s)-${expirationInSecs%.*} ))
+
+if [[ "${MODE}" != "windowsVhdMode" ]]; then
+  exit 0
+fi
+
+# we will clean old sig image-definitions/versions/galleries, RGs only if 
+# we have overriden the pipeline variable SKIP_IMAGE_DELETION_BACKFILL to False
```
cameronmeissner, line 14: not necessary, but I personally think this entire flow would be a lot more readable if we renamed SKIP_IMAGE_DELETION_BACKFILL to something like PERFORM_RESOURCE_DELETION_BACKFILL and swap around the boolean values - generally better to not have negatives/negations in the names of booleans

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -0,0 +1,104 @@
...
+# or we are running on production pipelines
+CLEAN_OLD_WINDOWS_RESOURCES="False"
+if [[ "${SKIP_IMAGE_DELETION_BACKFILL}" == "False" || "${DRY_RUN}" == "False" ]]; then
```
cameronmeissner, line 17: to clarify, we backfill resource deletion when this variable is overridden or during any build not in dry-run mode?

cameronmeissner: should separate out chained conditionals with brackets:

[[ "${SKIP_IMAGE_DELETION}" == "False" ]] || [[ "${DRY_RUN}" == "False" ]]

vhdbuilder/packer/backfill-cleanup.sh
cameronmeissner: yeah think this would be cleaner if we just say:

`
[[ "${SKIP_IMAGE_DELETION_BACKFILL"} == "True" ]] && [[ "${DRY_RUN}" == "True" ]] && exit 0
`

then we don't need `CLEAN_OLD_WINDOWS_RESOURCES` at all, unless I'm missing something.

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -0,0 +1,104 @@
...
+
+# attempt to clean up Windows SIG image versions in all galleries except "AKSWindows" created over a week ago
+if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${CLEAN_OLD_WINDOWS_RESOURCES}" == "True" ]]; then
+  gallery_list=$(az sig list -g ${AZURE_RESOURCE_GROUP_NAME} | jq '.[] | select(.name != "AKSWindows") | .name' | tr -d '\"' || "")
+  for gallery in $gallery_list; do
+    # delete old Windows SIG image versions in gallery (image definitions must have .osType == "Windows")
+    image_defs=$(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq '.[] | select(.osType == "Windows").name' | tr -d '\"' || "")
+    for image_definition in $image_defs; do
+        echo "Finding sig image versions associated with ${image_definition} in gallery ${gallery}"
+        image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | tr -d '\"' || "")
```
cameronmeissner, line 60: nit: everywhere we do a selection from an array with jq, you can tack `-r` onto it so we don't need the `tr -d '\"'`

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -0,0 +1,101 @@
...
+(( expirationInSecs = ${EXPIRATION_IN_HOURS} * 60 * 60 ))
+# deadline = the "date +%s" representation of the oldest age we're willing to keep
+(( deadline=$(date +%s)-${expirationInSecs%.*} ))
+
+if [[ "${MODE}" != "windowsVhdMode" ]]; then
+  exit 0
+fi
+
+# any build not in dry-run mode for Windows must be a check-in build
+if [[ "${BACKFILL_RESOURCE_DELETION}" == "False" ]] && [[ "${DRY_RUN}" == "True" ]]; then
```
AbelHu, line 14: This can be more simple to remove the parameter DRY_RUN.
 In check-in pipeline, the default value of  `BACKFILL_RESOURCE_DELETION` is `false` and we can set it to `true` manually. It works without `DRY_RUN`.
We can always set `BACKFILL_RESOURCE_DELETION` to `true` in prod pipeline instead of using `DRY_RUN`.

vhdbuilder/packer/backfill-cleanup.sh
AbelHu: Can we use `managed_image_ids += " /subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/images/${image}"` 

It is difficult to identify the blank between `"${managed_image_ids}` and `/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/images/${image}"`

Same for sig_version_ids

AbelHu: I think that `+` should make it to clearer that this is an append operation.

vhdbuilder/packer/cleanup.sh
```diff
@@ -180,17 +171,3 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY_RUN,,}" == "false" ]]; then
   
   set -x
 fi
-
-#clean up storage account created over a week ago
```
AbelHu, line 184: Will removing these lines impact the cleanup on linuxVhdMode?

AbelHu:
@wanqingfu PR LGTM now. Please check and fix all failures in the test pipelines.

wanqingfu:
> @wanqingfu PR LGTM now (some issues have been detected by the failed test pipelines). Please check and fix all failures in the test pipelines.

Will do. Do I need to set `BACKFILL_RESOURCE_DELETION` to `True` in prod pipeline before 2B? This can be tested on check-in test pipelines as well since we removed 'DRY_RUN` condition checking.

AbelHu:
> > @wanqingfu PR LGTM now (some issues have been detected by the failed test pipelines). Please check and fix all failures in the test pipelines.
> 
> Will do. Do I need to set `BACKFILL_RESOURCE_DELETION` to `True` in prod pipeline before 2B? This can be tested on check-in test pipelines as well since we removed 'DRY_RUN` condition checking.

I think that we should set `BACKFILL_RESOURCE_DELETION` to `True` in prod pipeline after this PR is merged. 

I do not understand `This can be tested on check-in test pipelines as well since we removed 'DRY_RUN' condition checking.`

wanqingfu:
> > > @wanqingfu PR LGTM now (some issues have been detected by the failed test pipelines). Please check and fix all failures in the test pipelines.
> > 
> > 
> > Will do. Do I need to set `BACKFILL_RESOURCE_DELETION` to `True` in prod pipeline before 2B? This can be tested on check-in test pipelines as well since we removed 'DRY_RUN` condition checking.
> 
> I think that we should set `BACKFILL_RESOURCE_DELETION` to `True` in prod pipeline after this PR is merged.
> 
> I do not understand `This can be tested on check-in test pipelines as well since we removed 'DRY_RUN' condition checking.`

should I simply set its default value to `True` in `.vsts-vhd-builder-release-windows.yaml`? (current value is `False`)

I meant do I need to queue a pipeline on containerd&docker using this branch to test the backfill deletion? I can test it on PR check-in pipeline by setting `BACKFILL_RESOURCE_DELETION` to `True`.

AbelHu:
> ant do I need to queue a pipeline on containerd&docker using this branch to test the backfill deletion? I can test it on PR check-in pipeline by setting `BACKFILL_RESOURCE_DELETION` to `True`.

I think that we should set BACKFILL_RESOURCE_DELETION to True in prod pipeline. If you do not add a pipeline variable so you should change the default value in the code.

It works in check-in pipeline so it should also work in prod pipeline since I think that there is no difference for your change in this PR.

wanqingfu:
> > ant do I need to queue a pipeline on containerd&docker using this branch to test the backfill deletion? I can test it on PR check-in pipeline by setting `BACKFILL_RESOURCE_DELETION` to `True`.
> 
> I think that we should set BACKFILL_RESOURCE_DELETION to True in prod pipeline. If you do not add a pipeline variable so you should change the default value in the code.
> 
> It works in check-in pipeline so it should also work in prod pipeline since I think that there is no difference for your change in this PR.

 
changed default to `True` in prod pipeline. 

ok.

wanqingfu:
> > > > ant do I need to queue a pipeline on containerd&docker using this branch to test the backfill deletion? I can test it on PR check-in pipeline by setting `BACKFILL_RESOURCE_DELETION` to `True`.
> > > 
> > > 
> > > I think that we should set BACKFILL_RESOURCE_DELETION to True in prod pipeline. If you do not add a pipeline variable so you should change the default value in the code.
> > > It works in check-in pipeline so it should also work in prod pipeline since I think that there is no difference for your change in this PR.
> > 
> > 
> > changed default to `True` in prod pipeline.
> > ok.
> 
> FYI: With this change, you can run the test in prod pipeline with setting `DRY_RUN` to `True` to test this PR without uploading VHDs to prod.

I see. Now I understand why you said we can remove `DRY_RUN=True` check in `backfill-cleanup.sh` since we can set it to `True` in prod pipeline, which is convenient. Just queued on in prod with `DRY_RUN` and `BACKFILL_RESOURCE_DELETION` both checked.

PR #2485, opened 2022-12-06
lockdown windows VHD build networking with 1es pools

cameronmeissner:
**What type of PR is this?**

This PR will have windows VHD builds take place on our newly deployed (or soon to be deployed) self-created 1ES hosted pools . These pools host agents which live in a vnet that has been peered with the vnet(s) where the packer VM which performs the windows VHD build gets provisioned and run, thus eliminating the need for a network exception to allow the currently used ADO/1ES build agent IP to WinRM into the packer VM from an external network.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests


vhdbuilder/packer/init-variables.sh
```diff
@@ -36,35 +43,6 @@ if [ -z "$rg_id" ]; then
...
 fi
 
-if [ -n "${VNET_RESOURCE_GROUP_NAME}" ]; then
-	VIRTUAL_NETWORK_NAME="vnet"
-	VIRTUAL_NETWORK_SUBNET_NAME="subnet"
-	NETWORK_SECURITY_GROUP_NAME="nsg"
-
-	echo "creating resource group ${VNET_RESOURCE_GROUP_NAME}, location ${AZURE_LOCATION} for VNET"
-	az group create --name ${VNET_RESOURCE_GROUP_NAME} --location ${AZURE_LOCATION} \
-		--tags 'os=Windows' 'createdBy=aks-vhd-pipeline' 'SkipASMAzSecPack=True'
```
cameronmeissner, line 46: Do we need these tags on every resource in the network stack? Packer is already configured to tag the build VM with these tags as is.

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -43,7 +43,6 @@ stages:
             echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D4s_v3'
             echo '##vso[task.setvariable variable=CONTAINER_RUNTIME]docker'
             echo '##vso[task.setvariable variable=WINDOWS_SKU]2019'
-            echo '##vso[task.setvariable variable=VNET_RESOURCE_GROUP_NAME]vnet-2019-$(Build.BuildNumber)'
```
cameronmeissner, line 46: Can we get away with not creating these networking resources on the fly like it's currently implemented?

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -55,7 +55,7 @@ steps:
       -e CLIENT_ID=${BUILD_CLIENT_ID} \
       -e CLIENT_SECRET="$(BUILD_CLIENT_SECRET)" \
       -e TENANT_ID=${BUILD_TENANT_ID} \
-      -e SUBSCRIPTION_ID="${BUILD_SUBSCRIPTION_ID}" \
+      -e SUBSCRIPTION_ID="8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8" \
```
wanqingfu, line 58: Is it for debugging or it's hard-coded? I think the subscription id is not the same as the one that appears in the PR-checkin pipeline?

cameronmeissner: yeah this is just a temporary thing since I wanted to make sure the build could actually work (and it does :D), that's why I still have this marked as a draft - after speaking with @AbelHu it sounds like the best approach will be to use separate dedicated sub for vhd builds in which we'll deploy the 1ES pool and associated networking resources.

PR #2496, opened 2022-12-09
fix: add timeout and panic recovery handler for all absvc routes

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/apiserver/getdistrosigimageconfig.go
+++ b/apiserver/getdistrosigimageconfig.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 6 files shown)


apiserver/routers.go
```diff
@@ -41,5 +42,7 @@ func (api *APIServer) NewRouter(ctx context.Context) *mux.Router {
 		Name("GetDistroSigImageConfig").
 		HandlerFunc(api.GetDistroSigImageConfig)
 
+	router.Use(handlers.RecoveryHandler(handlers.PrintRecoveryStack(true)))
```
paulgmiller, line 45: if you use a different logger you can set that here too.

paulgmiller, line 45: You could also put in the log handler to get a better idea of which request caused the failure (thought it won't get operation id ) 
https://pkg.go.dev/github.com/gorilla/handlers#LoggingHandler

But sicne this is a hotfix I'll let you decide if you want to save it for later.

paulgmiller: by default it just uses the log package

paulgmiller (approved):

apiserver/routers.go
alexeldeib: yeeees. I don't want to log the whole thing indiscriminately since I just caused another issue that way but will at least try to get some of the panic struct bits explicitly

alexeldeib:
we use goroutines for timeouts which messes up panic handlers. need to adjust.

apiserver/routers.go
alexeldeib: okay ended up refactoring more, might stick with this for first pass.

apiserver/routers.go
```diff
@@ -41,5 +42,34 @@ func (api *APIServer) NewRouter(ctx context.Context) *mux.Router {
...
+func handleOK(w http.ResponseWriter, r *http.Request) {
+	setHeaders(w)
+	w.WriteHeader(http.StatusOK)
+}
+
+func recoveryHandler() mux.MiddlewareFunc {
+	return handlers.RecoveryHandler(handlers.PrintRecoveryStack(true))
+}
+
+func timeoutHandler() mux.MiddlewareFunc {
```
alexeldeib, line 71: this should behavior very similarly to our timeout goroutine, but properly tracks recovery by passing panic stacks across goroutines, which is very tricky. our implementation was not close to cutting it. https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/net/http/server.go;l=3394-3437

apiserver/routers.go
```diff
@@ -41,5 +42,34 @@ func (api *APIServer) NewRouter(ctx context.Context) *mux.Router {
...
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+}
+
+func handleOK(w http.ResponseWriter, r *http.Request) {
+	setHeaders(w)
+	w.WriteHeader(http.StatusOK)
+}
+
+func recoveryHandler() mux.MiddlewareFunc {
+	return handlers.RecoveryHandler(handlers.PrintRecoveryStack(true))
```
alexeldeib, line 68: this captures panics and logs them, which ironically should be standard go http behavior, but because we used goroutines we escaped the panic handler in the location of the panic.

alexeldeib: correction: the default is recovery but does not include the panic stack from the handler apparently.

apiserver/getnodebootstrapdata.go
```diff
@@ -17,60 +17,39 @@ const (
...
-}
-
 // GetNodeBootstrapConfig endpoint for getting node bootstrapping data.
 func (api *APIServer) GetNodeBootstrapData(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
-	processResult := make(chan Result)
-	go func() {
```
alexeldeib, line 32: doing this in a goroutine avoided our panic handler which swallows a lot of useful info.

apiserver/routers.go
```diff
@@ -41,5 +42,34 @@ func (api *APIServer) NewRouter(ctx context.Context) *mux.Router {
...
 	return router
 }
+
+func healthz(w http.ResponseWriter, r *http.Request) {
+	handleOK(w, r)
+}
+
+func setHeaders(w http.ResponseWriter) {
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
```
paulgmiller, line 59: no idea why this is needed.

apiserver/getdistrosigimageconfig.go
```diff
@@ -22,46 +23,36 @@ func (api *APIServer) GetDistroSigImageConfig(w http.ResponseWriter, r *http.Req
...
 
+	result, err := json.Marshal(allDistros)
+	if err != nil {
+		log.Println(err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
 	}
+
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprint(w, string(result))
```
paulgmiller, line 57: not important for this pr at all (in fact please don't make this change but you know you can deseriize directly to w with json.Encoder.

apiserver/routers.go
alexeldeib: it's not needed, I like to be a good citizen and set content type headers correctly. it's copy paste from myself years ago :P

will remove for the sake of keeping this small.

alexeldeib: or will shove in handleOK and drop the nosniff since that's really for browsers.

PR #2498, opened 2022-12-09
refactor: add configurable timeout handler, apiserver recorded tests

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/apiserver/routers.go
+++ b/apiserver/routers.go
@@ -45,7 +45,7 @@ func (api *APIServer) NewRouter(ctx context.Context) *mux.Router {
 	router.Methods("GET").Path("/healthz").Name("healthz").HandlerFunc(healthz)
 
 	// global timeout and panic handlers.
-	router.Use(timeoutHandler(), recoveryHandler())
+	router.Use(requestContextInjector(api.Options.Timeout), timeoutHandler(api.Options.Timeout), recoveryHandler())
 
 	return router
 }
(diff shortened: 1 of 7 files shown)


apiserver/routers.go
```diff
@@ -63,8 +63,18 @@ func recoveryHandler() mux.MiddlewareFunc {
...
 
-func timeoutHandler() mux.MiddlewareFunc {
+func timeoutHandler(timeout time.Duration) mux.MiddlewareFunc {
 	return func(h http.Handler) http.Handler {
-		return http.TimeoutHandler(h, time.Second*30, "")
+		return http.TimeoutHandler(h, timeout, "")
+	}
+}
+
+func requestContextInjector(timeout time.Duration) mux.MiddlewareFunc {
```
paulgmiller, line 72: didn't understand how this is being used.

alexeldeib: ah will remove. timeout handler accepts a context from the request itself, which I was using this to override with a per-request timeout. I was going to use this in UT to force timeouts without mocking a clock (since timeouthandler doesn't accept one).

instead I moved the timeout to the server level, and configured it to zero for UT.

cameronmeissner (approved):
lgtm I think

PR #2500, opened 2022-12-13
fix kubelet startup

cameronmeissner:
**What type of PR is this?**

Fixes #

This PR fixes a bug where, after start-stop scenarios, the kubelet systemd unit attempts to start before the containerd unit has a chance to come alive - due to systemd's default restarting behavior, this state would prevent kubelet from ever properly starting, rendering the node dead.

Fixes include:
- add explicit dependency on containerd from kubelet unit definition
- add a 1-second delay between restarts of the kubelet service

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/kubelet.service
```diff
@@ -2,10 +2,11 @@
...
 ConditionPathExists=/usr/local/bin/kubelet
 Wants=network-online.target
-After=network-online.target
+After=network-online.target containerd.service
 
 [Service]
 Restart=always
+RestartSec=2
 EnvironmentFile=/etc/default/kubelet
 SuccessExitStatus=143
```
alexeldeib, line 11: unrelated but can you figure out what 143 is and leave a comment

not blocking

parts/linux/cloud-init/artifacts/kubelet.service
```diff
@@ -2,10 +2,11 @@
 Description=Kubelet
 ConditionPathExists=/usr/local/bin/kubelet
 Wants=network-online.target
-After=network-online.target
+After=network-online.target containerd.service
```
alexeldeib, line 5: can we sanity check this in standalone? 

I'd also argue to add containerd to Wants to avoid races.

alexeldeib: possibly even Requires vs Wants...up to you

alexeldeib (approved):
lgtm with some iterative improvements.

PR #2508, opened 2022-12-14
fix: disable fabricmanager for nc_a100_v4, enable for nd_v4 when mig disabled, and reboot when required for mig

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib:
hmmm im wondering if https://github.com/Azure/AgentBaker/pull/1338/files#diff-59521d60320265e88e46f4702029a64c03d857baee2923ff450f4997f6184fe8R253 is also breaking this

weirdly gpu profiles seem to partition fine without it, but mig mode change should require a reboot?

debug with a sample pod using MIG7g on A100 tomorrow.

alexeldeib:
confirmed earlier comments, driver install and everything works, mig-partition fails but curiously does not fail CSE (!) even though it looks like it should. guessing we are mistaking an exit code somewhere on systemctl restart (not a oneshot so guessing some service behavior comes into play for simple vs forking vs oneshot)

it fails only when trying to deploy a CUDA workload, with CUDA_ERR_NOT_INITIALIZED or similar, but then continues using CPU only (???)

so definitely need that reboot after gpu reset.

but should check current mig mode first, using `nvidia-smi` with format csv or similar.

alexeldeib:
I've tested this with:

"normal" GPU
Standard_NC6s_v3 - plain VHD - MIG N/A
Standard_NC6s_v3 - GPU VHD - MIG N/A

MIG-capable, single card, (incompatible with fabric manager)
Standard_NC24ads_A100_v4 - plain VHD - MIG disabled
Standard_NC24ads_A100_v4 - plain VHD - MIG7g
Standard_NC24ads_A100_v4 - GPU VHD - MIG disabled
Standard_NC24ads_A100_v4 - GPU VHD - MIG7g

MIG-capable, 8 card (requires fabric manager)
Standard_ND96asr_v4 - plain VHD - MIG disabled
Standard_ND96asr_v4 - plain VHD - MIG7g
Standard_ND96asr_v4 - GPU VHD - MIG disabled
Standard_ND96asr_v4 - GPU VHD - MIG7g

all cases seem to provision successfully. hit infoROM host platform bug 2x, but that's unrelated. all configurations seem able to run the sample mnist job from https://learn.microsoft.com/en-us/azure/aks/gpu-cluster#run-a-gpu-enabled-workload

this is really tough to test. it seems like the default job behavior is just to proceed with CPU. we need to grep the logs or have cuda fail on init deliberately if we don't find gpus.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -120,6 +120,9 @@ logs_to_events "AKS.CSE.installNetworkPlugin" installNetworkPlugin
     logs_to_events "AKS.CSE.downloadKrustlet" downloadContainerdWasmShims
 {{- end }}
 
+# By default, never reboot new nodes.
```
alexeldeib, line 123: @paulgmiller pretty sure this has been broken for a while :/ only requested review as fyi, but MIG still needs this.

pkg/agent/baker.go
```diff
@@ -231,11 +231,17 @@ func (t *TemplateGenerator) getBakerFuncMap(config *datamodel.NodeBootstrappingC
 // similar to what the ARM template used to do.
 //
 // When ARM template was used, the following is used:
-//   variables('labelResourceGroup')
+//
```
alexeldeib, line 234: errr gofmt?

alexeldeib:
> nvidia-smi --query-gpu=mig.mode.current,mig.mode.pending --format=csv,noheader | tr -d ' '
 
useful to check current/pending mig mode, if mismatched on VM boot, likely host platform issue not us.

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: at least people are trying to use MIG now :P

alexeldeib: also this won't be required for next gen gpu as mentioned below yay

pkg/agent/baker_test.go
```diff
@@ -532,8 +532,19 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 			config.KubeletConfig = map[string]string{}
 			config.AgentPoolProfile.VMSize = "Standard_ND96asr_v4"
 			config.EnableNvidia = true
-			config.GPUInstanceProfile = "mig-3g"
+			config.GPUInstanceProfile = "MIG7g"
```
ganeshkumar5699, line 535: Why is this being changed from "mig-3g" to "MIG7g", in this test (for the same VM size)? Just to test the 7 partitions?

ganeshkumar5699 (approved):

pkg/agent/baker_test.go
alexeldeib: mig-3g is invalid syntax, figured it's better to use appropriate profiles.

MIG3g is a valid profile. 1,2,3,4,7 are all valid currently valid. see https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#a100-profiles

PR #2533, opened 2022-12-19
fix: run renamed unit test

junjiezhang1997:
**What type of PR is this?**
/kind bug

run renamed unit test for windows cse

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/templates/templates_generated.go
+++ b/pkg/templates/templates_generated.go
@@ -84,7 +84,6 @@
 // windows/kuberneteswindowssetup.ps1
 // windows/sendlogs.ps1
 // windows/windowscsehelper.ps1
-// windows/windowscsehelper.tests.ps1
 package templates
 
 import (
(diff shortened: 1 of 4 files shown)


pkg/templates/templates_generated.go
```diff
@@ -7447,7 +7447,8 @@ func windowsWindowscsehelperPs1() (*asset, error) {
 }
 
 var _windowsWindowscsehelperTestsPs1 = []byte(`BeforeAll {
```
AbelHu, line 7449: interesting. Could we exclude _windowsWindowscsehelperTestsPs1 in templates_generated.go since it is only for ut?

junjiezhang1997: If we exclude _windowsWindowscsehelperTestsPs1, we would also need to exclude windowscsehelper.tests.ps1 in templates_generated.go. Is it ok?

AbelHu: yes. I think that we should not include the test file in the template. It was a mistake.

junjiezhang1997: Thanks for the comment. Correct the mistake now.

AbelHu (approved):

PR #2543, opened 2022-12-20
move a couple small things from CSE to VHD

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line40.sh
```diff
@@ -70,16 +70,6 @@ downloadAzureCNI() {
     retrycmd_get_tarball 120 5 "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" ${VNET_CNI_PLUGINS_URL} || exit $ERR_CNI_DOWNLOAD_TIMEOUT
 }
 
-setupCNIDirs() {
```
alexeldeib, line 73: be careful with this kind of change. I think CNI is safe because we already extract much of it in install-deps.

but there are edge cases like SP reset where newer CSE runs against older vhd. non-existent directories are really tricky until everything is self-contained.

for some of the stuff like ensureJournal, the same issue exists, but it doesn't cause _functional_ failures, mostly minor issues until next node image upgrade.

alexeldeib: ab e2e against old vhd would help validate this (do make sure it hits this code path)

cameronmeissner: yeah looks like this doesn't work with older images unfortunately

PR #2544, opened 2022-12-20
add parameterization for custom SIG version ID to AB E2E

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/e2e-local.sh
```diff
@@ -7,15 +7,21 @@ set -euxo pipefail
 : "${LOCATION:=eastus}"
 : "${CLUSTER_NAME:=agentbaker-e2e-test-cluster}"
 
+# Error out if SCENARIO_NAME, VM_SKU, or SCENARIO_NAME are unbound
```
anujmaheshwari1, line 10: SIG_VERSION instead of SCENARIO_NAME?

Error out if SCENARIO_NAME, VM_SKU, or SIG_VERSION are unbound

e2e/e2e-scenario.sh
```diff
@@ -67,9 +67,21 @@ EOF
 cat $SCENARIO_NAME-vmss.json
 
 # Create a test VMSS with 1 instance 
-# TODO 3: Discuss about the --image version, probably go with aks-ubuntu-1804-gen2-2021-q2:latest
-#       However, how to incorporate chaning quarters?
 log "Creating VMSS"
+
+# If a custom SIG_VERSION_ID was not provided via the scenario matrix or via the command line, default to a locked 1804Gen2 image for now
+if [ -z "$SIG_VERSION_ID" ]; then
```
anujmaheshwari1, line 73: nit: can ignore if too much effort

another way of doing it could be to instantiate SIG_VERSION_ID with the id we use if input isnt provided.
this will help reduce the code following here a bit, essentially, only the else will follow.

can totally leave it as is, its fine honestly. just an observation.

anujmaheshwari1 (approved):

e2e/e2e-local.sh
cameronmeissner: woops good catch

e2e/e2e-scenario.sh
cameronmeissner: to clarify: you mean instantiate it here or in e2e-local.sh if it's not provided?

PR #2555, opened 2023-01-09
Add Edge Zone SIG definition

yunrugemicrosoft:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/sig_config.go
```diff
@@ -245,6 +253,9 @@ const (
 	// We do not use AKS Windows image versions in AgentBaker. These fake values are only used for unit tests
 	Windows2019SIGImageVersion string = "17763.2019.221114"
 	Windows2022SIGImageVersion string = "20348.2022.221114"
+
+	// Used for edge zone scenario
+	EdgeZoneSIGImageVersion string = "2023.01.09"
```
cameronmeissner, line 258: how often would this be bumped?

pkg/agent/datamodel/sig_config.go
```diff
@@ -543,6 +585,14 @@ func withACSSIGConfig(acsSigConfig SIGConfig, osSKU string) (SigImageConfigOpt,
 	}, nil
 }
 
+func withEdgeZoneConfig(acsSigConfig SIGConfig) (SigImageConfigOpt, error) {
```
UtheMan, line 588: We're using this to avoid dependency on ACS configs for edge zones. ACS configs have no edge zones in them and cluster deployment would fail if we tried to read from them.

pkg/agent/datamodel/sig_config.go
UtheMan: @yunrugemicrosoft do you know?

cameronmeissner (approved):

pkg/agent/datamodel/sig_config.go
yunrugemicrosoft: Currently, we are manually kicking off the build for edge zones on the AKS SIG replication pipeline and need to update this version manually here for a short time. Right now the replication to edge zones is not quite stable, so most builds are for testing. But we should fix it soon and during this time we would publish aligning with the AKS normal Linux SIG build version and schedule, just manually and in a separate build.

Later we are also looking to set the edge zone replication auto-triggered in the AKS SIG replication pipeline after our build is stable. Then the EdgeZoneSIGImageVersion here could be the same as the existing LinuxSIGImageVersion.

UtheMan: Okay if that's the case then we can leave it as is for now I think? Thoughts @cameronmeissner ?

cameronmeissner: sgtm

PR #2570, opened 2023-01-14
Move linux sig image version to config and change image bump pipeline

ganeshkumar5699:
**What type of PR is this?**
/kind feature

Agentbaker/VHD releases need to be tracked by the AKS release tracker. Currently, the release tracker only tracks RP release. 

To get the latest version that is released from Agentbaker, the release tracker needs to obtain the SIG version from a config file. This PR moves it from Go code to a config file and refactors references, and the image bump pipeline step to make it work.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/sig_config.go
```diff
@@ -260,191 +266,236 @@ const (
...
+
+	SIGCBLMarinerV1ImageConfigTemplate = SigImageConfigTemplate{
+		ResourceGroup: AKSCBLMarinerResourceGroup,
+		Gallery:       AKSCBLMarinerGalleryName,
+		Definition:    "V1",
+		Version:       linuxSIGImageVersion,
+	}
+}
+
+func init() {
```
alexeldeib, line 445: grrr..can we avoid this init function

pkg/agent/datamodel/sig_config.go
```diff
@@ -260,191 +266,236 @@ const (
...
-	}
+	var sigImageStruct SIGVersion
+	json.Unmarshal([]byte(contents), &sigImageStruct)
 
-	SIGUbuntuGPU1804Gen2ImageConfigTemplate = SigImageConfigTemplate{
-		ResourceGroup: AKSUbuntuResourceGroup,
-		Gallery:       AKSUbuntuGalleryName,
-		Definition:    "1804gen2gpu",
-		Version:       FrozenLinuxSIGImageVersionForDocker,
+	if len(contents) == 0 {
```
cameronmeissner, line 300: should we check contents length before the unmarshal call?

pkg/agent/datamodel/sig_config.go
```diff
@@ -260,191 +266,236 @@ const (
...
+var LinuxSIGImageVersion string
+
 var (
-	SIGUbuntu1604ImageConfigTemplate = SigImageConfigTemplate{
-		ResourceGroup: AKSUbuntuResourceGroup,
-		Gallery:       AKSUbuntuGalleryName,
-		Definition:    "1604",
-		Version:       "2021.11.06",
-	}
+	SIGUbuntuContainerd1804Gen2ImageConfigTemplate,
```
alexeldeib, line 272: suggestion: create a function that returns the populated versions, not an init func that populates the vars

vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -17,24 +17,18 @@ build_ids=$3
...
 find_current_image_version() {
     filepath=$1
-    while read -r p; do
-        if [[ $p == *"LinuxSIGImageVersion"* ]]; then
-            current_image_version=$(echo $p | awk -F'\"' '{print $2}')
-            echo "Image version is $current_image_version, cut from line $p"
-            break
-        fi
-    done < $filepath
+    current_image_version=$(jq .version $filepath)
```
cameronmeissner, line 23: not sure if this is used elsewhere, but might want to tack `-r` on here so we get rid of the surrounding quotes

pkg/agent/datamodel/sig_config.go
```diff
@@ -260,191 +266,236 @@ const (
...
+		ResourceGroup: AKSCBLMarinerResourceGroup,
+		Gallery:       AKSCBLMarinerGalleryName,
+		Definition:    "V1",
+		Version:       linuxSIGImageVersion,
+	}
+}
+
+func init() {
+	var err error
+
```
cameronmeissner, line 447: nit: unnecessary new lines?

pkg/agent/datamodel/sig_config.go
```diff
@@ -260,191 +266,236 @@ const (
...
 
 	SIGCBLMarinerV2Arm64ImageConfigTemplate = SigImageConfigTemplate{
 		ResourceGroup: AKSCBLMarinerResourceGroup,
 		Gallery:       AKSCBLMarinerGalleryName,
 		Definition:    "V2gen2arm64",
-		Version:       LinuxSIGImageVersion,
+		Version:       linuxSIGImageVersion,
+	}
+
+	SIGUbuntuContainerd1804ImageConfigTemplate = SigImageConfigTemplate{
```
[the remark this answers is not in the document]
ganeshkumar5699: Yes, this just moved without changing

pkg/agent/datamodel/linux_sig_version.json
```diff
@@ -0,0 +1,4 @@
+{
+  "ostype": "LinuxSIGImageVersion",
```
alexeldeib, line 2: just doing this one as a first example?

cameronmeissner, line 2: doesn't look like "ostype" is ever used? I might be missing something tho

ganeshkumar5699: For the other versions I will be splitting them into different PRs, and since the LinuxSIGImageVersion is the one most commonly used and updated, I started with that one.

pkg/agent/datamodel/sig_config.go
ganeshkumar5699: I had used the init function mainly for error checking. Switched to storing the version date returned from the parsing function directly, and using panics instead.

pkg/agent/datamodel/sig_config.go
```diff
@@ -244,9 +247,12 @@ const (
 	AKSUbuntuEdgeZoneResourceGroup string = "AKS-Ubuntu-EdgeZone"
 )
 
-const (
-	LinuxSIGImageVersion string = "2023.01.19"
+type SIGVersion struct {
```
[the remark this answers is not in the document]
ganeshkumar5699: Avoided exporting this

pkg/agent/datamodel/linux_sig_version.json
ganeshkumar5699: Currently not used but it will be used later to differentiate between different SKUs (while using same helper function)

vhdbuilder/scripts/automate_version_bump.sh
ganeshkumar5699: Thanks, updated.

pkg/agent/datamodel/sig_config.go
ganeshkumar5699: I'd created the previous initialization helper function, because it was needed when using an init function (since the original package variables are evaluated before the init function).
With the removal of the init function, the helper function is also not needed.

pkg/agent/datamodel/sig_config.go
```diff
@@ -260,6 +260,33 @@ const (
 	EdgeZoneSIGImageVersion string = "2023.01.09"
 )
 
+type sigVersion struct {
+	OSType  string `json:"ostype"`
+	Version string `json:"version"`
+}
+
+//go:embed linux_sig_version.json
+var LinuxVersionJSONContentsEmbedded string
```
alexeldeib, line 269: nit: make private var

alexeldeib (approved):

alexeldeib:
would be nice to avoid panics but this is all internal and it's a const so much bigger refactoring

lgtm otherwise

vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -82,6 +76,6 @@ cut_official_branch() {
 }
 
 set_git_config
-find_current_image_version "pkg/agent/datamodel/sig_config.go"
+find_current_image_version "linux_sig_version.json"
```
anujmaheshwari1, line 79: @ganeshkumar5699 this should have been the absolute file path, its failing in vhd automation. it never got caught because automation never worked since this PR merged. I will fix it.

vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -17,24 +17,18 @@ build_ids=$3
...
-    done < $filepath
+    current_image_version=$(jq -r .version $filepath)
     echo "Current image version is: ${current_image_version}"
 }
 
 # This function replaces the old image version with the new input image version for all relevant files
 update_image_version() {
     sed -i "s/${current_image_version}/${new_image_version}/g" pkg/agent/bakerapi_test.go
     sed -i "s/${current_image_version}/${new_image_version}/g" pkg/agent/datamodel/sig_config.go
-    sed -i "s/${current_image_version}/${new_image_version}/g" pkg/agent/datamodel/sig_config_test.go
```
anujmaheshwari1, line 37: this was also incorrect. sig_config_test.go has some hardcoded values that still rely on this. this has to be updated or the test would have failed https://github.com/Azure/AgentBaker/blob/e6086d14d13e3acc1c1e87eae03706cb6a56c4bd/pkg/agent/datamodel/sig_config_test.go#L60

again, not caught because automation didnt run. please run automation pipeline when making such changes to validate.

PR #2580, opened 2023-01-18
adding func for cni vhd steps and readding some that got dropped

chmill-zz:
**What type of PR is this?**

noticed in an investigation that the multi cni wasn't getting the untar in vhdbuild. So adding that in as we as others.

We do this for speed during CSE

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -289,6 +289,15 @@ else
     retagContainerImage "docker" ${watcherFullImg} ${watcherStaticImg}
 fi
 
+# doing this at vhd allows CSE to be faster with just mv
+cniUnTarNClean() {
```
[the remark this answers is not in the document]
alexeldeib: ahh nvm. changed the loop var in 2579 but VNET_CNI_PLUGINS_URL is still the same

alexeldeib, line 293:
```suggestion
unpackAzureCNI() {
```

or similar (not picky but no stop-n-shop style abbreviation please)

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -289,6 +289,15 @@ else
...
+  CNI_TGZ_TMP=${VNET_CNI_PLUGINS_URL##*/}
+  CNI_DIR_TMP=${CNI_TGZ_TMP%.tgz}
+  mkdir "$CNI_DOWNLOADS_DIR/${CNI_DIR_TMP}" 
+  tar -xzf "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" -C $CNI_DOWNLOADS_DIR/$CNI_DIR_TMP
```
alexeldeib, line 297: can we add a check to run-test that the binaries exist where cse thinks they do?

if no, can you clarify why? I would like to fix content checks to be more thorough later, looking for obstacles.

vhdbuilder/packer/install-dependencies.sh
chmill-zz: in an interview when they ask about my weaknesses. The answer is naming things.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -319,6 +323,7 @@ SWIFT_CNI_VERSIONS="
 for VNET_CNI_VERSION in $SWIFT_CNI_VERSIONS; do
     VNET_CNI_PLUGINS_URL="https://acs-mirror.azureedge.net/azure-cni/v${VNET_CNI_VERSION}/binaries/azure-vnet-cni-swift-linux-${CPU_ARCH}-v${VNET_CNI_VERSION}.tgz"
     downloadAzureCNI
+    unpackAzureCNI
```
alexeldeib, line 326: won't this overwrite the files every time? or do I need some sleep

alexeldeib: only thing changing is source, and dest is the same every time?

alexeldeib: hmm, no I think it's okay...will check logs tomorrow...

VNET_CNI_PLUGINS_URL -> last part becomes CNI_TGZ_TMP
CNI_TGZ_TMP without .tgz becomes CNI_DIR_TMP

$CNI_DOWNLOADS_DIR/$CNI_DIR_TMP is final unpack dir...

so as long as the file name/versions are unique for each version, it works?

vhdbuilder/packer/install-dependencies.sh
chmill-zz: linux-vhd-content-test.sh runs through components.json today. But it's doing a fileSize check https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/test/linux-vhd-content-test.sh#L43

Rather than a 1:1 of the binaries. 

This change I've just made actually correctly failed this test because the swift and overlay cni's where added with this unpack check, but I didn't adjust the components.json to look for a non *.tgz

vhdbuilder/packer/install-dependencies.sh
alexeldeib: aaaand for my own sanity: from the runtime side --

https://github.com/Azure/AgentBaker/blob/39611fbce3331dae7e9c3598134c5e737c1d52ee/parts/linux/cloud-init/artifacts/cse_install.sh#L194-L195

https://github.com/Azure/AgentBaker/blob/39611fbce3331dae7e9c3598134c5e737c1d52ee/parts/linux/cloud-init/artifacts/cse_install.sh#L200-L202

still looking in $CNI_DOWNLOADS_DIR/$CNI_DIR_TMP, then mv to final location, and fall back to tarball, then fall back to download.

bravo for managing this

alexeldeib (approved):

PR #2588, opened 2023-01-19
chore: use self-created 1es hosted pools to build windows vhd

junjiezhang1997:
**What type of PR is this?**
/kind improvement

Use self-created 1es hosted pools to build windows vhd so that we do not need to create security exception to expose WinRM port 5986 in the sub “AKS VHD Image Release”

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/windows-vhd-builder.json
+++ b/vhdbuilder/packer/windows-vhd-builder.json
@@ -17,7 +17,8 @@
         "windows_image_version": "{{env `WINDOWS_IMAGE_VERSION`}}",
         "windows_patch_url": "{{env `WINDOWS_PATCH_URL`}}",
         "vnet_name": "{{env `VNET_NAME`}}",
-        "subnet_name": "{{env `SUBNET_NAME`}}"
+        "subnet_name": "{{env `SUBNET_NAME`}}",
+        "vnet_resource_group_name": "{{env `VNET_RESOURCE_GROUP_NAME`}}"
     },
     "builders": [
         {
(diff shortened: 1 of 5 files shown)


cameronmeissner (approved):
I just double-checked and looks like you did this already, but I'll just stress here to make sure the names of your pools are unique and don't overlap with existing Linux pools (I think the bicep templates would probably yell at you for that anyways now that I think about it).

cameronmeissner:
I'll also mention that we've been experiencing network issues with the pool and connecting to VMs in the peered vnet during our ARM64 builds specifically - I doubt this will be something that affects you but we're planning on getting around to fixing that on our side soon.

vhdbuilder/packer/windows-vhd-builder.json
```diff
@@ -49,10 +50,9 @@
                 "createdBy": "aks-vhd-pipeline",
                 "SkipASMAzSecPack": "True"
             },
-            "virtual_network_resource_group_name": "{{user `vnet_resource_group_name`}}",
-            "virtual_network_name": "vnet",
-            "virtual_network_subnet_name": "subnet",
-            "private_virtual_network_with_public_ip": "true"
+            "virtual_network_resource_group_name": "nodesigtest-agent-pool",
```
ShiqianTao, line 53: Variable may be more flexible. BTW: this work is interesting, could we have a wiki to note it?

junjiezhang1997: Fine. Modify the use of variable and add the wiki.

AbelHu (approved):

PR #2594, opened 2023-01-20
Add support for Mariner 2.0 Gen 1

SeanDougherty:
**What type of PR is this?**

/kind feature

Ease customer adoption of Mariner AKS by removing any potential friction for moving from Ubuntu to Mariner. Several customers have raised concerns for the need to keep using a Gen1 only VM SKU. Supporting Gen1 Mariner 2.0 AKS would also improve our OSSKU migration story by allowing Mariner 1.0 customers to migrate existing clusters to Mariner 2.0. 

In support of [Deliverable #43020951](https://microsoft.visualstudio.com/OS/_workitems/edit/43020951)

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

Requesting feedback from @alexeldeib on the naming of the SIGCBLMarinerV2Gen1ImageConfigTemplate and the SIGCBLMarinerV2ImageConfigTemplate variables. The latter is already part of the codebase, but breaks the convention of gen2 possessing the '...gen2' naming convention.

I wish to update these variables to follow the common practice of the codebase (See below), but want to verify that such a change won't have knock-on effects in other areas such as breaking contracts with AKSRP or corrupting the production SIG gallery's image definitions. With your approval, I will update this PR with the following changes:

Current variable names:
SIGCBLMarinerV2Gen1ImageConfigTemplate // for gen1
SIGCBLMarinerV2ImageConfigTemplate // for gen2

Proposed variable names:
SIGCBLMarinerV2ImageConfigTemplate // for gen1
SIGCBLMarinerV2Gen2ImageConfigTemplate // for gen2

**Release note**:
```
none
```


pkg/agent/datamodel/sig_config.go
```diff
@@ -417,6 +419,13 @@ var (
 		Version:       LinuxSIGImageVersion,
 	}
 
+	SIGCBLMarinerV2Gen1ImageConfigTemplate = SigImageConfigTemplate{
```
hbeberman, line 422: As you mentioned @SeanDougherty, this would ideally be SIGCBLMarinerV2ImageConfigTemplate, and the one below would be SIGCBLMarinerV2Gen2ImageConfigTemplate to pattern match with the rest.

@alexeldeib is there any risk with renaming the existing Mariner V2 Gen2 config template and fixing it up in the SigImageConfig map?

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -46,6 +46,10 @@ parameters:
   displayName: Build MarinerV1 Gen2
   type: boolean
   default: true
+- name: buildMarinerV2gen1
```
hbeberman, line 49:
```suggestion
- name: buildMarinerV2
```

Should probably drop the gen1 suffix from the changes in this file for consistency with the existing build targets.

vhdbuilder/packer/vhd-image-builder-mariner2.json
```diff
@@ -0,0 +1,391 @@
+{
```
hbeberman, line 1: I don't think we need to add a new image builder json for gen1, HYPERV_GENERATION is passed in by the caller and there's no delta between mariner2 and mariner2-gen2 jsons. I'd just rename the existing one to vhd-image-builder-mariner2.json

hbeberman (changes requested):

hbeberman (approved):

hbeberman (approved):

alexeldeib (approved):

pkg/agent/datamodel/sig_config.go
alexeldeib: this isn't used externally afaik, i'm fine either way

worst case we can revert it

alexeldeib: the key piece is the mapping of the distro name "aks-cblmariner-v2" to the image config (the string "aks-cblmariner-v2" must be the same in aks-rp)

TODO future: extract those to common data model somewhere else

PR #2598, opened 2023-01-20
Managed Prometheus addon image latest image: 1.2.3-test

gracewehner:


gracewehner:
This is an automation test

PR #2612, opened 2023-01-23
refactor: remove sgx code

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_main.sh
+++ b/parts/linux/cloud-init/artifacts/cse_main.sh
@@ -181,12 +181,6 @@ logs_to_events "AKS.CSE.installKubeletKubectlAndKubeProxy" installKubeletKubectl
 
 createKubeManifestDir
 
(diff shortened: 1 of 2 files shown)


alexeldeib:
checked with ACC folks, this VM size is EOL in Azure and they agree we should be able to delete this code

cameronmeissner (approved):
is this the only place we use SGX code? I remember factoring out installSGX to a new flag that gets passed around the yaml files back when we merged the different build modes - wondering if we can delete that stuff as well? (I can follow-up with a new PR if so)

alexeldeib:
we should be able to remove installSGXDrivers entirely now, I just didn't do so because it doesn't have go templates :D

sgx functionality from later SKUs comes from ACC addon on RP side

PR #2635, opened 2023-01-24
feat: add IPFamilies to config for better IsDualStackEnabled

tyler-lloyd:
**What type of PR is this?**

/kind feature

Adds the property `IPFamilies` to the KubernetesConfig which can be used to determine if the cluster
is dual-stack or not. Currently, AKS does not use the FeatureFlag `IsIPv6DualStackFeatureEnabled` and
instead uses the IPFamilies property.

A change is also made to the function, `IsIPv6DualStackFeatureEnabled`, which will return true
for IPFamilies = [ipv4, ipv6]. This is needed to setup the correct ext network in Windows inside
the `New-ExternalHnsNetwork` function.

It will also ensure DHCPv6 on Linux nodes but this is inconsequential since IPv6 is done via
IPv6 Stateless Address Autoconfiguration.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/types.go
```diff
@@ -537,6 +537,7 @@ type KubernetesConfig struct {
 	NetworkMode                       string            `json:"networkMode,omitempty"`
 	ContainerRuntime                  string            `json:"containerRuntime,omitempty"`
 	MaxPods                           int               `json:"maxPods,omitempty"`
+	IPFamilies                        []string          `json:"ipFamilies,omitempty"`
```
ashvindeodhar, line 540: when / where do you set KubernetesConfig.IPFamilies to v4 and v6?

tyler-lloyd: all of these values get set in aks-rp. there is a agent baker bootstrap config that gets built by rp and then passed to agent baker during vmss create/update

staging/cse/windows/provisioningscripts/windowsnodereset.ps1
```diff
@@ -15,7 +15,8 @@ $global:NetworkMode = "L2Bridge"
 $global:NetworkPlugin = $Global:ClusterConfiguration.Cni.Name
 $global:ContainerRuntime = $Global:ClusterConfiguration.Cri.Name
 $UseContainerD = ($global:ContainerRuntime -eq "containerd")
-$IsDualStackEnabled = $Global:ClusterConfiguration.Kubernetes.Kubeproxy.FeatureGates -contains "IPv6DualStack=true"
+$IsDualStackEnabled = ($Global:ClusterConfiguration.Kubernetes.Kubeproxy.FeatureGates -contains "IPv6DualStack=true") `
+                        -Or ($Global:ClusterConfiguration.Kubernetes.Kubelet.NodeLabels -like "*kubernetes.azure.com/podv6network-type=overlay*")
```
tyler-lloyd, line 19: open to something else besides checking if this label exists. the FeatureGate won't be set so we'll need something else and this was the best option when I looked at the ClusterConfig json object

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -157,7 +157,7 @@ $global:AzureCNIConfDir = [Io.path]::Combine("$global:AzureCNIDir", "netconf")
 # $global:NetworkPolicy = "{{GetParameter "networkPolicy"}}" # BUG: unused
 $global:NetworkPlugin = "{{GetParameter "networkPlugin"}}"
 $global:VNetCNIPluginsURL = "{{GetParameter "vnetCniWindowsPluginsURL"}}"
-$global:IsDualStackEnabled = {{if IsIPv6DualStackFeatureEnabled}}$true{{else}}$false{{end}}
+$global:IsDualStackEnabled = {{if or IsIPv6DualStackFeatureEnabled IsAzureCNIOverlayDualStackFeatureEnabled}}$true{{else}}$false{{end}}
```
AbelHu, line 160: Do we need any change for DualStack with WinDSR enabled?
Current `Set-AzureCNIConfig` in staging/cse/windows/azurecnifunc.ps1 only handles DualStack when WinDSR is disabled.

pkg/agent/baker.go
```diff
@@ -564,6 +564,10 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"IsAzureCNIOverlayFeatureEnabled": func() bool {
 			return cs.Properties.OrchestratorProfile.KubernetesConfig.IsUsingNetworkPluginMode("overlay")
 		},
+		"IsAzureCNIOverlayDualStackFeatureEnabled": func() bool {
```
AbelHu, line 567: could we use IsDualStack instead of adding IsAzureCNIOverlayDualStackFeatureEnabled?

pkg/agent/datamodel/types.go
```diff
@@ -1261,6 +1262,23 @@ func (k *KubernetesConfig) IsIPMasqAgentEnabled() bool {
 	return k.IsAddonEnabled(IPMASQAgentAddonName)
 }
 
+func (k *KubernetesConfig) IsDualStack() bool {
```
AbelHu, line 1265: is it possible to set `EnableIPv6DualStack` in `FeatureFlags` when DualStack is enabled so we do not change these code in AgentBaker?

parts/windows/kuberneteswindowssetup.ps1
tyler-lloyd: Do you know if there's any reason they need to be mutually exclusive? Can the dualstack section in `Set-AzureCNIConfig` run regardless of if DSR is enabled?

pkg/agent/baker.go
tyler-lloyd: that's probably fine, I can make the change

pkg/agent/datamodel/types.go
tyler-lloyd: I avoided using the FeatureFlags because we do not currently use it for Kubenet dual stack so if we start using it here then it could have unintended consequences on the Linux side.

AbelHu: How does Linux support IPv6? should we also need to update other components if we enable IPv6 on Windows agent pools, for example, LoadBalancerSku?

tyler-lloyd: everything on the Linux side is already taken care of, we just don't happen to use the "FeatureFlag" property for it. It's controlled via toggle and then a cluster can be checked if it is dualstack by looking at the ipFamilies property on the managed cluster to see if it is `ipv4` or `ipv4,ipv6`.

AbelHu: @tyler-lloyd you can search `EnableIPv6DualStack` in AKS RP code and check whether it is needed when using ipFamilies property

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -157,7 +157,7 @@ $global:AzureCNIConfDir = [Io.path]::Combine("$global:AzureCNIDir", "netconf")
 # $global:NetworkPolicy = "{{GetParameter "networkPolicy"}}" # BUG: unused
 $global:NetworkPlugin = "{{GetParameter "networkPlugin"}}"
 $global:VNetCNIPluginsURL = "{{GetParameter "vnetCniWindowsPluginsURL"}}"
-$global:IsDualStackEnabled = {{if IsIPv6DualStackFeatureEnabled}}$true{{else}}$false{{end}}
+$global:IsDualStackEnabled = {{if or IsIPv6DualStackFeatureEnabled IsDualStack}}$true{{else}}$false{{end}}
```
AbelHu, line 160: If `IsIPv6DualStackFeatureEnabled` is not used, I suggest you removing this function and code in AgentBaker. We keep this for dualstack but now a new property will be used so we should remove it.

parts/windows/kuberneteswindowssetup.ps1
AbelHu: We only have one customer with WinDSR disabled. @allyford is helping on this to enable WinDSR in their clusters.
I think that we can remove all code for WinDSR disabled.

PR #2657, opened 2023-01-25
refactor: remove containerd exec_start from cloud init

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/containerd_exec_start.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+ExecStartPost=/sbin/iptables -P FORWARD ACCEPT
```
UtheMan, line 2: Not too familiar with this systemd command, any reason why ExecStartPost is used when there's no ExecStart command to run?

UtheMan (approved):

parts/linux/cloud-init/artifacts/containerd_exec_start.conf
alexeldeib: it's a drop in, if you look at the file path it ends up in `/etc/systemd/system/containerd.service.d/exec_start.conf`

systemd services often live in files like `/etc/systemd/system/containerd.service`, and by convention any files in a directory like `/etc/systemd/system/containerd.service.d/` will be applied as drop in overrides 

see https://www.freedesktop.org/software/systemd/man/systemd.unit.html informative but not visually pleasing 

https://flatcar-linux.org/docs/latest/setup/systemd/drop-in-units/#override-only-specific-option
also for example or perhaps https://steveazz.xyz/micro/systemd-drop-in/ to keep it short and sweet

PR #2672, opened 2023-01-25
refactor: remove many unnecessary go templates

alexeldeib:
- systemd units can be on disk without being enabled
- same for scripts
- some of these things can be done unconditionally with no change like docker exec_start

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/CustomData
```diff
@@ -149,8 +149,19 @@ write_files:
   content: |
     U2VjcmV0CiNFT0Y=
 
+- path: /etc/systemd/system/dhcpv6.service
```
alexeldeib, line 152: looks normal

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -261,6 +258,7 @@ write_files:
     ExecStartPost=/sbin/iptables -P FORWARD ACCEPT
     #EOF
 
+{{if RequiresDocker}}
```
ganeshkumar5699, line 261: I think we need an `{{end}}`` corresponding to this?

cameronmeissner (approved):

parts/linux/cloud-init/nodecustomdata.yml
alexeldeib: L284 I think -- diff cuts it off

PR #2674, opened 2023-01-25
refactor: http proxy config entirely out of cloud init

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -61,11 +61,14 @@ if [[ "${DISABLE_SSH}" == "true" ]]; then
     disableSSH || exit $ERR_DISABLE_SSH
 fi
 
-if [[ "${SHOULD_CONFIGURE_HTTP_PROXY_CA}" == "true" ]]; then
-    configureHTTPProxyCA || exit $ERR_UPDATE_CA_CERTS
+if [[ "${SHOULD_CONFIGURE_HTTP_PROXY}" == "true" ]]; then
+    if [[ "${SHOULD_CONFIGURE_HTTP_PROXY_CA}" == "true" ]]; then
+        configureHTTPProxyCA || exit $ERR_UPDATE_CA_CERTS
+    fi
     configureEtcEnvironment
```
alexeldeib, line 68: only calling configureEtcEnvironment when provided with a trusted CA was a bug, guess no one noticed

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -69,18 +69,43 @@ configureSwapFile() {
 }
 
 configureEtcEnvironment() {
+    mkdir -p /etc/systemd/system.conf.d/
+    touch /etc/systemd/system.conf.d/proxy.conf
+    chmod 0644 /etc/systemd/system.conf.d/proxy.conf
```
cameronmeissner, line 74: unrelated, but wondering if we can get away with properly hardening file perms all around eventually, if not already (0644 -> 0600)

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
alexeldeib: hah, yes excellent point. I just noticed this in one of my later commits

probably good follow up

PR #2678, opened 2023-01-26
Move Mariner, 2204 and Edge Zone SIG versions to config

ganeshkumar5699:
**What type of PR is this?**
/kind feature

Moving SIG image versions for Mariner v2, Ubuntu 2204 and Edge Zones to a config file, so that it can eventually be tracked by the AKS release tracker (which customers have asked for). 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):
lgtm but would be nice to not have a bunch of files for this, vs one file parsed with a default/overrides

e.g. we always have one linux version, except for docker, and we occasionally pin mariner

ganeshkumar5699:
> 

Okay. I thought the goal was to have one json file per distro. Can also switch it.

ganeshkumar5699:
discussed offline and it seems like it's fine to have it this way for now, and decide whether to change depending on how this is consumed

PR #2704, opened 2023-01-30
chore: test changes in windows cse package automatically

junjiezhang1997:
**What type of PR is this?**
/kind improvement

test changes in windows cse package automatically

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/e2e/e2e-windows-scenario.sh
+++ b/e2e/e2e-windows-scenario.sh
@@ -15,27 +15,27 @@ collect-logs() {
     VMSS_INSTANCE_ID="$(az vmss list-instances --name $DEPLOYMENT_VMSS_NAME -g $MC_RESOURCE_GROUP_NAME | jq -r '.[0].instanceId')"
     set +x
     expiryTime=$(date --date="2 day" +%Y-%m-%d)
-    token=$(az storage container generate-sas --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --permissions 'rwacdl' --expiry $expiryTime --name $STORAGE_CONTAINER_NAME --https-only)
+    token=$(az storage container generate-sas --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --permissions 'rwacdl' --expiry $expiryTime --name $STORAGE_LOG_CONTAINER --https-only)
     # Use .ps1 file to run scripts since single quotes of parameters for --scripts would fail in check-shell
     az vmss run-command invoke --command-id RunPowerShellScript \
         --resource-group $MC_RESOURCE_GROUP_NAME \
         --name $DEPLOYMENT_VMSS_NAME \
         --instance-id $VMSS_INSTANCE_ID \
         --scripts @upload-cse-logs.ps1 \
-        --parameters arg1=$STORAGE_ACCOUNT_NAME arg2=$STORAGE_CONTAINER_NAME arg3=$DEPLOYMENT_VMSS_NAME arg4=$token
+        --parameters arg1=$STORAGE_ACCOUNT_NAME arg2=$STORAGE_LOG_CONTAINER arg3=$DEPLOYMENT_VMSS_NAME arg4=$token
     if [ "$retval" != "0" ]; then
         echo "failed in uploading cse logs"
     fi
-    wget https://aka.ms/downloadazcopy-v10-linux
-    tar -xvf downloadazcopy-v10-linux
+    # wget https://aka.ms/downloadazcopy-v10-linux
+    # tar -xvf downloadazcopy-v10-linux
     tokenWithoutQuote=${token//\"}
     # use array to pass shellcheck
     array=(azcopy_*)
-    ${array[0]}/azcopy copy "https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/${STORAGE_CONTAINER_NAME}/${DEPLOYMENT_VMSS_NAME}-cse.log?${tokenWithoutQuote}" $SCENARIO_NAME-logs/CustomDataSetupScript.log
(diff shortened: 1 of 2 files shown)


e2e/e2e-windows-scenario.sh
```diff
@@ -45,6 +44,36 @@ collect-logs() {
...
+csePackageURL="https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/${STORAGE_PACKAGE_CONTAINER}/${DEPLOYMENT_VMSS_NAME}-aks-windows-cse-scripts.zip?${tokenWithoutQuote}"
+export csePackageURL
+
+cd ../../../e2e
+
+array=(azcopy_*)
+${array[0]}/azcopy copy ${DEPLOYMENT_VMSS_NAME}-aks-windows-cse-scripts.zip "https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/${STORAGE_PACKAGE_CONTAINER}/${DEPLOYMENT_VMSS_NAME}-aks-windows-cse-scripts.zip?${tokenWithoutQuote}"
+set -x
+
+echo "upload cse packages done"
```
AbelHu, line 73: could we exit immediately if it fails in uploading the test cse package?

e2e/e2e-windows-scenario.sh
```diff
@@ -109,6 +136,10 @@ az deployment group create --resource-group $MC_RESOURCE_GROUP_NAME \
 retval=$?
 set -e
 
+# delete cse package in storage account
+${array[0]}/azcopy rm "https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/${STORAGE_PACKAGE_CONTAINER}/${DEPLOYMENT_VMSS_NAME}-aks-windows-cse-scripts.zip?${tokenWithoutQuote}"
```
AbelHu, line 140: this step may not be executed for some failures. I think that we have two options:
1. use `--overwrite` in uploading the test cse package
2. check whether a file with the same name exists before uploading the test cse package.

e2e/e2e-windows-scenario.sh
```diff
@@ -45,6 +44,36 @@ collect-logs() {
...
+DEPLOYMENT_VMSS_NAME="$(mktemp -u winXXXXX | tr '[:upper:]' '[:lower:]')"
+export DEPLOYMENT_VMSS_NAME
+
+# download azcopy
+wget https://aka.ms/downloadazcopy-v10-linux
+tar -xvf downloadazcopy-v10-linux
+
+# zip and upload cse package
+cd ../staging/cse/windows
+zip -r ../../../e2e/${DEPLOYMENT_VMSS_NAME}-aks-windows-cse-scripts.zip ./* -x ./*.tests.ps1 -x "*azurecnifunc.tests.suites*" -x README -x provisioningscripts/*.md -x debug/update-scripts.ps1
```
AbelHu, line 56: It is better to add timestamp in the package name so we can cleanup all old packages when it fails in deleting the uploaded package

e2e/e2e-windows-scenario.sh
junjiezhang1997: Sure. I will add it

e2e/e2e-windows-scenario.sh
junjiezhang1997: I use the random-generated vmss name to specify a cse package. cse packages with the same name may bring the problem when one pipeline wants to use the package while another pipeline wants to delete it.

e2e/e2e-windows-scenario.sh
junjiezhang1997: Maybe use timestamp together with the random-generated vmss name to name cse package is better?

e2e/e2e-windows-scenario.sh
AbelHu: for option 2, we can use a different name when it exists. In most scenarios, it should have no problem if we add timestamp in the name. But it is better to check it before uploading it or check the upload result if it fails because the file exists.

junjiezhang1997: Thanks. I will take option 2 and add it.

e2e/e2e-windows-scenario.sh
```diff
@@ -45,6 +44,57 @@ collect-logs() {
...
+array=(azcopy_*)
+noExistStr="File count: 0"
+listResult=$(${array[0]}/azcopy list $csePackageURL --running-tally)
+
+for i in $(seq 1 10); do
+    if [[ "$listResult" != *"$noExistStr"* ]]; then
+        echo "Cse package with the same exists, retry $i to use new name..."
+        timeStamp=$(date +%s)
+        listResult=$(${array[0]}/azcopy list $csePackageURL --running-tally)
+        sleep 10
```
AbelHu, line 79: why do we need to sleep 10 seconds?

e2e/e2e-windows-scenario.sh
```diff
@@ -45,6 +44,57 @@ collect-logs() {
...
+
+listResult=$(${array[0]}/azcopy list $csePackageURL --running-tally)
+if [[ "$listResult" == *"$noExistStr"* ]]; then
+    err "failed to upload cse package"
+    exit 1
+fi
+
+echo "upload cse packages done"
+
+envsubst < scenarios/$SCENARIO_NAME/property-$SCENARIO_NAME-template.json > scenarios/$SCENARIO_NAME/property-$SCENARIO_NAME.json
```
AbelHu, line 96: What is the purpose to add this new line?

e2e/e2e-windows-scenario.sh
junjiezhang1997: I want to use "sleep 10" to guarantee the timestamp is different from before. I think I need to move this line before generate timeStamp. Maybe "sleep 5" is just enough?

e2e/e2e-windows-scenario.sh
junjiezhang1997: I need to use new cse package url in property-windows-json. This new line is used to set the new cse package url.

e2e/e2e-windows-scenario.sh
AbelHu: I think that we may not need to sleep since it should take several seconds in the command "azcopy list". How do you think?

AbelHu (approved):

PR #2706, opened 2023-01-30
add old pkr rg deletion backfill logic

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/cleanup.sh
+++ b/vhdbuilder/packer/cleanup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -x
 
-# deadline for packer resource group deletion is 24 hours
+# deadline for old packer resource group deletion is 24 hours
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/cleanup.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash -x
 
+# deadline for old packer resource group deletion is 24 hours
+(( packerRgExpirationInSecs = 24 * 60 * 60 ))
```
alexeldeib, line 4: this is fun new syntax

I always thought this only worked like var=$(( arithmetic * foo ))

Bryce-Soghigian: yes very fancy syntax

cameronmeissner:
will revisit after escalation to ADO

PR #2727, opened 2023-02-06
Restart containerd once during provisioning if custom ca certs were passed to cluster

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/update_certs_provisioning.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Adds certificates passed in managed cluster body during node provisioning in AKS
+
+[Service]
+Type=oneshot
+ExecStart=/opt/scripts/update_certs.sh
+ExecStart=/opt/scripts/restart_containerd.sh
```
alexeldeib, line 7: do you really want this? it will trigger the restart every time the certs on disk update, including from DS.

probably the right thing for users but a bit of a behavior change for people who didn't need this specific functionality

parts/linux/cloud-init/artifacts/restart_containerd.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo "Restarting containerd"
+systemctl restart containerd
```
alexeldeib, line 4: you can inline this into the systemd unit, fyi

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -122,7 +122,7 @@ configureCustomCaCertificate() {
         declare varname=CUSTOM_CA_CERT_${i} 
         echo "${!varname}" > /opt/certs/00000000000000cert${i}.crt
     done
-    systemctl restart update_certs.service || exit $ERR_UPDATE_CA_CERTS
+    systemctl restart update_certs_provisioning.service || exit $ERR_UPDATE_CA_CERTS
```
alexeldeib, line 125: ah this is annoying. it's not guaranteed to be on disk for older vhds. can you avoid changing the name

alexeldeib: wait, this is an addition to the previous service? you don't need this at all...

alexeldeib, line 125:
```suggestion
    # this block until the service is considered active!
    systemctl restart update_certs || exit $ERR_UPDATE_CA_CERTS
    systemctl restart containerd
```

now make update_certs.service `type=Notify`, and add `systemd-notify --ready` to the end of update_certs.sh

now `systemctl restart update_certs` blocks until `systemd-notify --ready` was called, and you don't need the extra unit

PR #2730, opened 2023-02-06
cli upgrade + optimizations to generating image versions

anujmaheshwari1:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/scripts/automate_generate_version.sh
+++ b/vhdbuilder/scripts/automate_generate_version.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-#set -euxo pipefail
+set -euxo pipefail
 
 build_ids=$1
 global_image_version="${IMAGE_VERSION:=}"
--- a/vhdbuilder/scripts/automate_generate_version.sh
+++ b/vhdbuilder/scripts/automate_generate_version.sh
@@ -13,16 +13,14 @@ for build_id in $build_ids; do
         # So in the past, it has happened that you trigger a build at say 3/4pm PST, 
         # some SKUs will have todays date some will have tomorrows based on when they are triggered because of UTC conversion
         # TODO(amaheshwari): Change VHD script to use a common var for image version that is plumbed down to all SKUs
-        if [[ $artifact == *"publishing-info"* ]]; then
-            az pipelines runs artifact download --artifact-name $artifact --path $(pwd) --run-id $build_id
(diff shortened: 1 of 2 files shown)


vhdbuilder/scripts/automate_generate_version.sh
```diff
@@ -4,22 +4,23 @@ set -euxo pipefail
 build_ids=$1
 global_image_version="${IMAGE_VERSION:=}"
 for build_id in $build_ids; do
-    for artifact in $(az pipelines runs artifact list --run-id $build_id | jq -r '.[].name'); do    # Retrieve what artifacts were published
+    artifacts=($(az pipelines runs artifact list --run-id $build_id | jq -r '.[].name'))
+    artifacts=(${artifacts[@]//vhd*})
```
anujmaheshwari1, line 8: this filters out all the files starting with vhd, so all that is left in the artifacts list is publishing info, which is what we really ever needed.

vhdbuilder/scripts/automate_generate_version.sh
```diff
@@ -4,22 +4,23 @@ set -euxo pipefail
...
+    printf '%s\n' "${artifacts[@]}"
+    for artifact in ${artifacts[@]}; do    # Retrieve what artifacts were published
         # This loop is because of how the Image Version is set for builds. 
         # It uses the UTC time of when the build for a particular SKU ends. 
         # So in the past, it has happened that you trigger a build at say 3/4pm PST, 
         # some SKUs will have todays date some will have tomorrows based on when they are triggered because of UTC conversion
         # TODO(amaheshwari): Change VHD script to use a common var for image version that is plumbed down to all SKUs
-        if [[ $artifact == *"publishing-info"* ]]; then
```
cameronmeissner, line 13: might be worth keeping this check for sanity's sake? 

I doubt the artifact names are gonna change anytime soon but would still be nice to have I think, up to you though

anujmaheshwari1: fair enough, since we already got some gains on speed, I will keep this check as is.

anujmaheshwari1: resolved

cameronmeissner (approved):

PR #2731, opened 2023-02-06
restart containerd during provisioning if custom ca certs are passed

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -122,7 +122,9 @@ configureCustomCaCertificate() {
         declare varname=CUSTOM_CA_CERT_${i} 
         echo "${!varname}" > /opt/certs/00000000000000cert${i}.crt
     done
+    # This will block until the service is considered active. Update_certs.service is a oneshot type of unit, it is considered active when the ExecStart= command terminates with a zero status code.
     systemctl restart update_certs.service || exit $ERR_UPDATE_CA_CERTS
+    systemctl restart containerd
```
cameronmeissner, line 127: seems straightforward, though why do we need to do this exactly? (mainly just curious)

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
UtheMan: after new certs are added to trust store, containerd will not pick them up properly before restart. If restart happens for some reason then all good and it will work, but the aim here is to have this working straight away for a freshly provisioned node. So we will force a restart ourselves after certs are added.

alexeldeib: can you add this comment to the code above restart containerd

alexeldeib: and potentially a comment about how the DS works, and that in the future we in theory could do this in the unit for maxxx reactiveness

PR #2742, opened 2023-02-08
restore repoDigests within image-bom.json for VHD builds

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/list-images.sh
```diff
@@ -3,9 +3,31 @@ set -euxo pipefail
...
+        digest=$(echo $image | awk '{print $2}')
+        id=$(echo "$crictl_list" | grep -e "$tag" | awk '{print $2}')
+
+        jq --arg repoTag "$tag" --arg repoDigest "$digest" --arg id "$id" -n '{id:$id, repoTags:[$repoTag], repoDigests:[$repoDigest]}' >> $temp_image_bom
+    done
+
+    IFS=$IFS_backup
+    jq --slurpfile images $temp_image_bom -n '$images | group_by(.id) | map({id:.[0].id, repoTags:map(.repoTags | add) | unique, repoDigests:map(.repoDigests | add) | unique})' > /opt/azure/containers/image-bom.json
+    rm -f $temp_image_bom
+    jq < /opt/azure/containers/image-bom.json
```
alexeldeib, line 26: why input?

vhdbuilder/packer/list-images.sh
```diff
@@ -3,9 +3,31 @@ set -euxo pipefail
...
+function generate_image_bom_for_containerd() {
+    temp_image_bom=/opt/azure/containers/temp-image-bom.json
+    IFS_backup=$IFS
+    IFS=$'\n'
+    ctr_list=$(ctr --namespace k8s.io image list | sed 1d | awk '{print $1, $3}')
+    crictl_list=$(crictl images --no-trunc | sed 1d | awk '{printf "%s:%s %s\n", $1,$2,$3;}')
+
+    for image in $ctr_list; do
+        tag=$(echo $image | awk '{print $1}')
+        # skip image tags that consist of a sha256 hash
```
alexeldeib, line 15: i vaguely recall this. you get one line per tag/sha pair per image, or something? why is this necessary? we should only have tagged images so IIRC this was something funny with ctr behavior

vhdbuilder/packer/list-images.sh
```diff
@@ -3,9 +3,31 @@ set -euxo pipefail
...
+        [[ $tag == sha256* ]] && continue
+        digest=$(echo $image | awk '{print $2}')
+        id=$(echo "$crictl_list" | grep -e "$tag" | awk '{print $2}')
```
alexeldeib, line 18: similar q above, if ctr has the tag, confused why we need both

vhdbuilder/packer/list-images.sh
alexeldeib: oh it's written above, this is just validation I suppose?

vhdbuilder/packer/list-images.sh
cameronmeissner: yep, each line is "<tag> <sha>" from the ctr output, then each line is "<tag> <id>" from the crictl output

cameronmeissner: and yeah not exactly sure why, but ctr will output a bunch of tags that are just shas in and of themselves, I cross-referenced previous image-bom's and saw that these images aren't included, so I added this to simply skip them for now

vhdbuilder/packer/list-images.sh
cameronmeissner: ctr has both the tag and the digest sha, but not the id sha - that comes from crictl

vhdbuilder/packer/list-images.sh
cameronmeissner: yeah was mainly for validation/so we can see the output during the build, through I'm seeing now this line might be producing errors during test builds...

cameronmeissner: ended up removing altogether

vhdbuilder/packer/list-images.sh
cameronmeissner: I was wrong! all can come from ctr, now simplified

alexeldeib (approved):

PR #2753, opened 2023-02-10
Revert "upgrade azure file csi driver to v1.26.0"

andyzhangx:
Reverts Azure/AgentBaker#2750
v1.25.1 ~ v1.26.0 has unmount issue when there is unmount stuck failure, force unmount is not a good solution


ZeroMagic (approved):
/lgtm

PR #2756, opened 2023-02-13
persist cleanup-vhd.sh on disk for invocation in subsequent builds/releases

cameronmeissner:
**What type of PR is this?**

New flow is:

if RUN_VHD_CLEANUP is true:
perform existing functionality (call cleanup-vhd.sh during VHD build, don't copy it to the VHD)
else:
don't call clean-up-vhd.sh during the VHD build, but copy it to the VHD so it can be later called by AIB during SIG release
fi

RUN_VHD_CLEANUP has been introduced as a new pipeline variable to all 3 linux build pipelines

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


anujmaheshwari1 (approved):

vhdbuilder/packer/cleanup-vhd.sh
```diff
@@ -1,9 +1,13 @@
...
-rm -f /root/.ssh/authorized_keys
-rm -f /home/packer/.ssh/authorized_keys
-rm -f /var/log/cloud-init.log /var/log/cloud-init-output.log 
-rm -f /etc/machine-id
-touch /etc/machine-id
-chmod 644 /etc/machine-id
\ No newline at end of file
+[ -z "${RUN_VHD_CLEANUP}" ] && echo "RUN_VHD_CLEANUP must be set for cleanup-vhd.sh to run" && exit 1
+
+if [ "${RUN_VHD_CLEANUP,,}" == "true" ]; then
```
alexeldeib, line 5: I would leave this as it was

you may want to shuffle the ssh key deletion for AIB steps tho

alexeldeib (approved):

PR #2762, opened 2023-02-14
add retry for gpu install

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -537,7 +537,7 @@ configGPUDrivers() {
         mkdir -p /opt/{actions,gpu}
         if [[ "${CONTAINER_RUNTIME}" == "containerd" ]]; then
             ctr image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG
-            bash -c "$CTR_GPU_INSTALL_CMD $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG gpuinstall /entrypoint.sh install" 
+            retrycmd_if_failure 5 10 600 bash -c "$CTR_GPU_INSTALL_CMD $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG gpuinstall /entrypoint.sh install"
```
cameronmeissner, line 540: so I guess worst case this could take an hour? :( 

hopefully that isn't often the case lol

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
UtheMan: tbh those numbers are kind of randomly chosen... I think we should tweak them. I'm open to suggestions Unfortunately the issue that I'm trying to fix with this has been pretty... elusive. One retry should be enough (most of the times...)

PR #2767, opened 2023-02-15
fix undesired sig versions/galleries cleanup in backfill deletion

wanqingfu:
**What type of PR is this?**
Fix the following issue

- backfill deletion step tries to free a sig image-def/gallery that is being used.

- This might happen when the current pipeline build creates a sig gallery & image-def but has not yet created a sig version (sig version would be empty). Or better yet, say there's another concurrent pipeline build just created a sig gallery & image def w/o attaching any sig versions.

Need to add some sort of check to see if the image def has older sig versions and if they can be reduced. If that's the case, the sig image-def can then be safely removed, hence the same about the sig gallery. o.w. we shouldn't delete the sig version => delete the sig image-def => delete the sig gallery if that makes sense.

/kind fix

fix the backfill deletion step

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

check build results of parallel-running pipelines (_pipeline that stay roughly close in start time_)

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/backfill-cleanup.sh
+++ b/vhdbuilder/packer/backfill-cleanup.sh
@@ -55,7 +55,6 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" ]]; then
         echo "Finding sig image versions associated with ${image_definition} in gallery ${gallery}"
         old_image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq --arg dl $deadline -r '.[] | select(.tags.now < $dl).name')
         for old_image_version in $old_image_versions; do
-            az sig image-version show -e $old_image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id
             echo "Deleting sig image-version ${old_image_version} ${image_definition} from gallery ${gallery} rg ${AZURE_RESOURCE_GROUP_NAME}"
             az sig image-version delete -e $old_image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME}
         done
--- a/vhdbuilder/packer/backfill-cleanup.sh
+++ b/vhdbuilder/packer/backfill-cleanup.sh
@@ -53,15 +53,15 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" ]]; then
     image_defs=$(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq -r '.[] | select(.osType == "Windows").name')
     for image_definition in $image_defs; do
         echo "Finding sig image versions associated with ${image_definition} in gallery ${gallery}"
-        image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq --arg dl $deadline -r '.[] | select(.tags.now < $dl).name')
-        for image_version in $image_versions; do
-            az sig image-version show -e $image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id
-            echo "Deleting sig image-version ${image_version} ${image_definition} from gallery ${gallery} rg ${AZURE_RESOURCE_GROUP_NAME}"
-            az sig image-version delete -e $image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME}
+        old_image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq --arg dl $deadline -r '.[] | select(.tags.now < $dl).name')
+        for old_image_version in $old_image_versions; do
+            az sig image-version show -e $old_image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id
+            echo "Deleting sig image-version ${old_image_version} ${image_definition} from gallery ${gallery} rg ${AZURE_RESOURCE_GROUP_NAME}"
+            az sig image-version delete -e $old_image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME}
         done
-        image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition})
-        # clean the image-definition if ALL associated sig versions have been deleted
-        if [[ "$image_versions" == "[]" ]]; then
+        cur_image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition})
+        # clean the image-definition if the current image versions are empty after cleaning older ones provided they exist
+        if [[ -n "${old_image_versions}" ]] && [[ "${cur_image_versions}" == "[]" ]]; then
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -53,15 +53,15 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" ]]; then
     image_defs=$(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq -r '.[] | select(.osType == "Windows").name')
     for image_definition in $image_defs; do
         echo "Finding sig image versions associated with ${image_definition} in gallery ${gallery}"
-        image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq --arg dl $deadline -r '.[] | select(.tags.now < $dl).name')
-        for image_version in $image_versions; do
-            az sig image-version show -e $image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id
-            echo "Deleting sig image-version ${image_version} ${image_definition} from gallery ${gallery} rg ${AZURE_RESOURCE_GROUP_NAME}"
-            az sig image-version delete -e $image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME}
+        old_image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq --arg dl $deadline -r '.[] | select(.tags.now < $dl).name')
```
AbelHu, line 56:
```
sigGalleryName="WS${WS_SKU}Gallery$(date +"%y%m%d%H%M%S")$RANDOM";
```
Could we parse the datetime from the generated SIG image name and decide whether to skip it if it is not expired?

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -53,15 +53,15 @@ if [[ -n "${AZURE_RESOURCE_GROUP_NAME}" ]]; then
...
+        for old_image_version in $old_image_versions; do
+            az sig image-version show -e $old_image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id
```
cameronmeissner, line 58: unrelated, but do we really need this call to `az sig image-version show`? seems like this would slow things down a fair amount

cameronmeissner (approved):

vhdbuilder/packer/backfill-cleanup.sh
AbelHu: @wanqingfu I think that we want to use this script to cleanup all those created SIG galleries including all resources in them and we can use the creation datetime in the gallery name. If the gallery needs to be deleted, we will delete all resources in it and then delete this gallery. Is it clear now?

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -49,28 +49,29 @@ fi
...
+
+    due_date=$(date +%y%m%d -d "7 days ago")
+    echo "create_date is ${create_date}"
+    echo "due_date is ${due_date}"
+    # clean the entire SIG resources if it's one week ago
+    if [[ $create_date < $due_date ]]; then
       echo "Deleting gallery ${gallery}"
       az sig delete --gallery-name ${gallery} --resource-group ${AZURE_RESOURCE_GROUP_NAME}
+      image_defs=$(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq -r '.[] | select(.osType == "Windows").name')
+      if [[ -n "${image_defs}" ]]; then
```
AbelHu, line 72: can a gallery be deleted before deleting all nested resources? Should we keep the old logic to delete all image definitions and versions?

wanqingfu: I think I need to do some manual experiments: create gallery => SIG image => SIG version and delete gallery completely before deleting all nested resources versus bottom-up deletion (_old logic_)

wanqingfu: short answer is we can't by running this single command alone. error message is "can not delete resource before nested resources are deleted".

image

we can do this on portal which I often did before by typing "yes" when asked if you wanna clean all nested resources, as shown below.
image

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -49,28 +49,46 @@ fi
...
+    else
+      continue
+    fi
+
+    due_date=$(date +%y%m%d -d "7 days ago")
+    echo "create_date is ${create_date}"
+    echo "due_date is ${due_date}"
+    # clean the entire SIG resources if it's one week ago
+    if [[ $create_date < $due_date ]]; then
+      echo "Finding sig image definitions from gallery ${gallery}"
```
wanqingfu, line 69: adding back old logic because running commands cannot delete nested resources unlike the portal.

The experiments - I created a bunch of sig image defs, versions.

image

dry-run of line `71` to line `91` locally showing that the bottom-up deletion succeeded.
image

vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -49,28 +49,46 @@ fi
...
+              az sig image-version delete -e $image_version -i ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME}
+          done
+          image_versions=$(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} -i ${image_definition} | jq -r '.[].name')
+          echo "image versions are $image_versions"
+          if [[ -z "${image_versions}" ]]; then
+            echo "Deleting sig image-definition ${image_definition} from gallery ${gallery} rg ${AZURE_RESOURCE_GROUP_NAME}"
+            az sig image-definition delete --gallery-image-definition ${image_definition} -r ${gallery} -g ${AZURE_RESOURCE_GROUP_NAME}
+          fi
+      done
+      image_defs=$(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${gallery} | jq -r '.[] | select(.osType == "Windows").name')
```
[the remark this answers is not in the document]
wanqingfu: I think it's safe to run another check since if it's empty it won't take any extra time. But I DO observe in the past that there were some resources (sig image-def in this case) that were not deleted. And I think that's part of the reason why we need a backfill deletion because if all resources are GC as we desire we won't need this _convoluted_ logic at all at the first place.

I think run another check will make sure the definitions will not be deleted if they are being used somehow by another pipeline. And we won't need to debug the RNF (resource not found) error message.

But if running the list introduces overhead I'm happy to remove it. I just feel it's safer to double check the emptiness/existence before cleaning all.

vhdbuilder/packer/backfill-cleanup.sh
wanqingfu: The proof of the dry-run test will be posted on the channel. (showing that this logic works)

vhdbuilder/packer/backfill-cleanup.sh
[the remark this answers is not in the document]
AbelHu: if for the debug purpose, I think that it may be better to log images defs if image_defs is not empty but always delete the gallery without checking it. It will fail in deleting the gallery but it has no harm, right?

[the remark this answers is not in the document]
wanqingfu: logged and deleted gallery anyway w/o checking emptiness.

AbelHu (approved):

AbelHu:
It is better to use the pipeline to test it after you have created some test resources manually.

junjiezhang1997 (approved):

PR #2768, opened 2023-02-15
feat: Validate generated sas token before publishing artifacts

AbelHu:
**What type of PR is this?**
/kind feature

Validate generated sas token before publishing artifacts

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Validated it in https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=68716276&view=logs&j=26d3b11b-de37-5bdb-7f8b-686d226a8fb0&t=e57474fd-ef2a-530e-8fb7-25622dad9b67 with reverting @junjiezhang1997 's fix.

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/generate-vhd-publishing-info.sh
+++ b/vhdbuilder/packer/generate-vhd-publishing-info.sh
@@ -57,6 +57,14 @@ if [ "$sas_token" == "" ]; then
 fi
 vhd_url="${STORAGE_ACCT_BLOB_URL}/${VHD_NAME}?$sas_token"
 
+echo "Testing whether the generated sas token works"
+vhd_size=$(curl -sI $vhd_url | grep -i Content-Length | awk '{print $2}')
+if [ "$vhd_size" == "" ]; then
+    echo "The genrated sas token does not work"
+    exit 1
+fi
+echo "The generated sas token works"
+
 # Do not log sas token
 echo "COPY ME ---> ${STORAGE_ACCT_BLOB_URL}/${VHD_NAME}?***"
 sku_name=$(echo $SKU_NAME | tr -d '.')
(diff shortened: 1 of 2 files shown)


ShiqianTao (approved):

junjiezhang1997 (approved):

wanqingfu (approved):

vhdbuilder/packer/generate-vhd-publishing-info.sh
```diff
@@ -57,6 +57,14 @@ if [ "$sas_token" == "" ]; then
 fi
 vhd_url="${STORAGE_ACCT_BLOB_URL}/${VHD_NAME}?$sas_token"
 
+echo "Testing whether the generated sas token works"
+vhd_size=$(curl -sI $vhd_url | grep -i Content-Length | awk '{print $2}')
```
alexeldeib, line 61: should we just check the response code instead? I think even for a 401/403/404, the response size would be non-zero XML payload?

alexeldeib:
```bash
res=$(curl -o /dev/null -w "%{http_code}" -fsSL -I "$vhd_url")
if test "$res" != "200"; then
  exit 1
fi
```

or similar?

AbelHu: I think that it is same. This command gets the content length. I have tested that the invalid token will show empty result but the valid token will get a non-empty `integer`.

AbelHu: ` curl -sI $vhd_url` with a valid token but a non-exist blob returns below response
```
HTTP/1.1 404 The specified blob does not exist.
Transfer-Encoding: chunked
Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
x-ms-request-id: 0da001d6-901e-000f-56a3-414af3000000
x-ms-version: 2018-11-09
x-ms-error-code: BlobNotFound
Date: Thu, 16 Feb 2023 01:09:47 GMT
```

PR #2769, opened 2023-02-15
feat: Add bcc-examples for Mariner

liunan-ms:
**What type of PR is this?**
/kind feature

Added bcc-examples package for all AKS Mariner images to allow AKS Mariner customers to use the example eBPF programs.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Build result: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=68674644&view=results

**Release note**:
```
none
```


vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -10,6 +10,8 @@ installBcc() {
     echo "Installing BCC tools..."
     dnf_makecache || exit $ERR_APT_UPDATE_TIMEOUT
     dnf_install 120 5 25 bcc-tools || exit $ERR_BCC_INSTALL_TIMEOUT
+    echo "Installing BCC examples..."
```
SeanDougherty, line 13: If we expect more bcc packages to be added sometime in the future, it could be good to lay the groundwork for that now. 

https://github.com/Azure/AgentBaker/blob/ff3ff212947242de5ee8ca7d8d87a3d1d8d84dbc/parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh#L20

Uses a For loop to make adding additional packages to the "install list" very easy. We could use that same logic here.

However, if we don't foresee additional packages ever being added to this function call, I think it looks great as is and you can resolve this comment :)

SeanDougherty (approved):

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
liunan-ms: Great point Sean! It is unlikely that there are other bcc packages. In the future new tools and more examples may be added into the existing packages, so I will keep the change as is here.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -218,6 +218,7 @@ EOF
 installBcc
 cat << EOF >> ${VHD_LOGS_FILEPATH}
   - bcc-tools
+  - bcc-examples
```
hbeberman, line 221: No need to add `- bcc-examples` to this list, especially since this logging statement is for both Ubuntu and Mariner. 
We already print out the package list into the log at the end of the AgentBaker run.
```suggestion
```

hbeberman (approved):

alexeldeib (approved):

juan-lee (approved):

PR #2783, opened 2023-02-21
Download and install mirror-proxy components for artifact streaming (not enabled)

ganeshkumar5699:
**What type of PR is this?**
/kind feature 

This is to download and install ACR's mirror-proxy components, for 22.04 and 20.04 (AMD) but not enable any of the systemd units corresponding to it. This is for Artifact Streaming on AKS. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -257,6 +257,32 @@ EOF
...
+  sudo /opt/acr/tools/overlaybd/enable-http-auth.sh || exit $ERR_MIRROR_PROXY_INSTALL_ERR
+  modprobe target_core_user || exit $ERR_MIRROR_PROXY_INSTALL_ERR
+  
+}
+
+enableArtifactStreamingServices() {
+  sudo /opt/acr/tools/overlaybd/enable.sh || exit $ERR_MIRROR_PROXY_INSTALL_ERR
+}
+
+if [[ "${CONTAINER_RUNTIME:=}" == "containerd"  && "${UBUNTU_RELEASE}" == "18.04" ]]; then
```
alexeldeib, line 282: I wouldn't bother with container runtime at this point, we should finish stripping that out of vhd build at least if not runtime (for old users)

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -195,6 +195,32 @@ EOF
 
 echo "${CONTAINER_RUNTIME} images pre-pulled:" >> ${VHD_LOGS_FILEPATH}
 
+installAndConfigureArtifactStreaming() {
```
juliusl, line 198: do you need a seperate function for mariner?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -195,6 +195,32 @@ EOF
...
+  echo "  - [installed] acr mirror-proxy v${MIRROR_PROXY_VERSION}" >> ${VHD_LOGS_FILEPATH}
+  rm "./acr-mirror-${UBUNTU_VERSION_CLEANED}.deb"
+  popd || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
+
+  sudo apt install libnl-3-dev libnl-genl-3-dev libc6 libssl3 -y
+
+  echo "  - [installed] libnl-3-dev, libnl-genl-3-dev, libc6 and libssl3" >> ${VHD_LOGS_FILEPATH}
+
+  sudo /opt/acr/tools/overlaybd/install.sh || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
+  sudo /opt/acr/tools/overlaybd/enable-http-auth.sh || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
```
juliusl, line 217: I think this is missing a step to curl the mirror to configure it's settings,

For example,

```sh
# Namespace means the registry host, _default will configure the mirror to be the default host,
# When an incoming request is received that doesn't match an accepted suffix, 
# a config will be created so that subsequent requests will not avoid calling the mirror 
# ex. docker.io request will first go to the mirror and a config will be created so that
#      subsequent requests go directly to docker.io
NAMESPACE=_default
# This controls which suffix's to enable the proxy w/
ENABLE_SUFFIX=azurecr.io
# Use 'overlaybd_vnext' to use the new artifact type
STREAM_FORMAT=overlaybd
# If true, this will edit containerd/config.toml to use overlaybd snapshotter and set cri defaults for the `--hosts-dir` setting
ENABLE_CONTAINERD=true

curl \ 
  -X PUT \
  "http://localhost:8578/config?ns=$NAMESPACE&enable_suffix=$ENABLE_SUFFIX&stream_format=$STREAM_FORMAT&enable_containerd=$ENABLE_CONTAINERD"
```

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: Depends on how the mirror will handle it. Currently it's installing for 22.04 only: https://github.com/juliusl/lifec_registry/blob/7ae610b456d48d17788917a23fa340ee54d60f57/lib/sh/overlaybd/install.sh#L6, so we will need to change that first. 

Mariner VHD builds are also done in this script, so we can use the same function for both Ubuntu and Mariner.

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: This will go in the CSE script (separate PR), as that will actually enable it for node pools that have the streaming flag.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -205,6 +205,28 @@ EOF
...
+
+  echo "  - [installed] acr mirror-proxy v${MIRROR_PROXY_VERSION}" >> ${VHD_LOGS_FILEPATH}
+  rm "./acr-mirror-${UBUNTU_VERSION_CLEANED}.deb"
+
+  sudo apt install libnl-3-dev libnl-genl-3-dev libc6 libssl3 -y
+
+  echo "  - [installed] libnl-3-dev, libnl-genl-3-dev, libc6 and libssl3" >> ${VHD_LOGS_FILEPATH}
+}
+
+if [[ ( "${UBUNTU_RELEASE}" == "20.04" || "${UBUNTU_RELEASE}" == "22.04" ) && ( "${CPU_ARCH}" == "amd64" ) ]]; then
```
alexeldeib, line 226: might be worth saving yourself future pain and splitting the major/minor, and doing >= 20 on the version.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -205,6 +205,28 @@ EOF
...
+  # download acr-mirror proxy
+  MIRROR_PROXY_VERSION='19'
+  UBUNTU_VERSION_CLEANED="${UBUNTU_RELEASE//.}"
+  MIRROR_PROXY_URL="https://acrmirrordev.blob.core.windows.net/bin/Release-${MIRROR_PROXY_VERSION}/acr-mirror-${UBUNTU_VERSION_CLEANED}.deb"
+  
+  wget $MIRROR_PROXY_URL || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
+  apt_get_install 30 1 600 "./acr-mirror-${UBUNTU_VERSION_CLEANED}.deb" || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
+  systemctl stop acr-mirror.service
+
+  echo "  - [installed] acr mirror-proxy v${MIRROR_PROXY_VERSION}" >> ${VHD_LOGS_FILEPATH}
```
alexeldeib, line 218: I think we won't need this, we list all installed debs anyway

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -205,6 +205,28 @@ EOF
 
 echo "${CONTAINER_RUNTIME} images pre-pulled:" >> ${VHD_LOGS_FILEPATH}
 
+installAndConfigureArtifactStreaming() {
+  # download acr-mirror proxy
+  MIRROR_PROXY_VERSION='19'
+  UBUNTU_VERSION_CLEANED="${UBUNTU_RELEASE//.}"
+  MIRROR_PROXY_URL="https://acrmirrordev.blob.core.windows.net/bin/Release-${MIRROR_PROXY_VERSION}/acr-mirror-${UBUNTU_VERSION_CLEANED}.deb"
+  
+  wget $MIRROR_PROXY_URL || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
```
alexeldeib, line 214: suggest a retry for network issues

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: Updated. May add 18.04 support for FIPS depending on usage and will add Mariner support later.

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: Added retry, thanks.

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: Removed

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -205,6 +205,28 @@ EOF
...
+  apt_get_install 30 1 600 "./acr-mirror-${UBUNTU_VERSION_CLEANED}.deb" || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
+  systemctl stop acr-mirror.service
+
+  echo "  - [installed] acr mirror-proxy v${MIRROR_PROXY_VERSION}" >> ${VHD_LOGS_FILEPATH}
+  rm "./acr-mirror-${UBUNTU_VERSION_CLEANED}.deb"
+
+  sudo apt install libnl-3-dev libnl-genl-3-dev libc6 libssl3 -y
```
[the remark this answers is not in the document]
ganeshkumar5699: It's for runtime dependency during installation of overlaybd components. I'd removed the overlaybd installation steps recently and should've removed as well. Can add it back, when adding overlaybd installation

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -158,6 +158,27 @@ for CRICTL_VERSION in ${CRICTL_VERSIONS}; do
...
+  retrycmd_curl_file 10 5 60 $MIRROR_DOWNLOAD_PATH $MIRROR_PROXY_URL || exit ${ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL}
+  
+  apt_get_install 30 1 600 $MIRROR_DOWNLOAD_PATH || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
+  systemctl stop acr-mirror.service
+
+  rm "./acr-mirror-${UBUNTU_VERSION_CLEANED}.deb"
+}
+
+UBUNTU_MAJOR_VERSION=$(echo $UBUNTU_RELEASE | cut -d. -f1)
+if [ $UBUNTU_MAJOR_VERSION -ge 20 ]; then
```
alexeldeib, line 177: still do need the mariner filter here. you could arguably filter the whole func on mariner

alexeldeib (approved):
lgtm with the mariner fix

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: Added check for Ubuntu/to filter Mariner

PR #2786, opened 2023-02-23

HDYA:


HDYA:
this should be merged IF we already publish image with previous patch versions, otherwise merge https://github.com/Azure/AgentBaker/pull/2827

andyliuliming (approved):

PR #2788, opened 2023-02-23
fix: add bootstrap kubeconfig and /etc/default/kubelet

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -277,6 +277,49 @@ write_files:
     [Install]
     WantedBy=multi-user.target
 
+- path: /etc/default/kubelet
+  permissions: "0644"
```
cameronmeissner, line 281: should probably harden these but I don't _think_ these files were part of the CIS benchmark

cameronmeissner (approved):

parts/linux/cloud-init/nodecustomdata.yml
alexeldeib: touche, I don't think so but maybe a good follow up

PR #2805, opened 2023-02-24
Update pod log collection to include all kube-system pods

phealy:
**What type of PR is this?**
/kind feature

Updates the sync tunnel logs process, which is used to provide logs for the walinuxagent hourly logs collection, to include logs for all kube-system pods

N/A

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

none

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
```diff
@@ -39,7 +39,7 @@ while true; do
 done &
 
 # Manually sync all matching logs once
-for TUNNEL_LOG_FILE in $(compgen -G "$SRC/@(aks-link|azure-cns|cilium|konnectivity|tunnelfront)-*_kube-system_*.log"); do
+for TUNNEL_LOG_FILE in $(compgen -G "$SRC/*_kube-system_*.log"); do
```
alexeldeib, line 42: TIL. if I understand, this is a bash built-in, so fine on mariner?

phealy: yeah, it's part of bash, and is a good way to expand the glob.

parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
```diff
@@ -48,7 +48,7 @@ echo "Starting inotifywait..."
 # Monitor for changes
 inotifywait -q -m -r -e delete,create $SRC | while read DIRECTORY EVENT FILE; do
     case $FILE in
-        aks-link-*_kube-system_*.log | azure-cns-*_kube-system_*.log | cilium-*_kube-system_*.log | konnectivity-*_kube-system_*.log | tunnelfront-*_kube-system_*.log)
+        *_kube-system_*.log)
```
alexeldeib, line 51: do we actually need this case statement now since we want all files?

alexeldeib: ah nvm still catches other namespaces otherwise

parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
alexeldeib, line 42: I suppose TUNNEL_LOG_FILE is no longer accurate ;)

alexeldeib (approved):
lgtm, will let @cameronmeissner review as well

parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
cameronmeissner: TUNNEL_LOG_FILE -> KUBE_SYSTEM_LOG_FILE, or maybe just LOG_FILE would work

cameronmeissner:
lgtm :)

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
phealy: renamed to CONTAINER_LOG_FILE.

parts/linux/cloud-init/artifacts/sync-tunnel-logs.sh
phealy: Right. It could be an if instead, but case has the nice wildcard matching built in, and it's easy to expand back out if we decide to add something like calico-system or tigera-operator in the future.

cameronmeissner:
moved everything to VHD, will proceed with testing

cameronmeissner:
tested with standalone, looks good

cameronmeissner (approved):

PR #2813, opened 2023-02-27
feat: add --node-ip to kubelet args for Windows dual-stack

tyler-lloyd:
**What type of PR is this?**

/kind feature

In overlay dual-stack there is a circular dependency with CNS and cloud-node-manager. CNS
is waiting for dual-stack InternalIPs to be set on the node, which CNM would normally update
a node with, but CNM cannot start on Windows because it does not run in a hostNetwork or
hostProcess mode and so it waits to get an IP. But it cannot get an IP from CNS because it's
waiting for the IPv6 IP.

This PR adds the IPv6 and IPv4 IPs from the host NIC via the Kubelet arg `--node-ip`.

Tested on a dual-stack windows node

```
PS C:\> $KubeletArgList = @()
PS C:\>     $defaultRouteIfIndex = (Get-NetRoute -DestinationPrefix 0.0.0.0/0).ifIndex
PS C:\> 
PS C:\>     # use addresses from dhcp and sort by family so IPv4 is always first
PS C:\>     $ipAddrs = (Get-NetIPAddress -InterfaceIndex $defaultRouteIfIndex | ? { $_.SuffixOrigin -eq "Dhcp" } | Sort-Object AddressFamily).IPAddress
PS C:\>
PS C:\>     if ($ipAddrs.Length -eq 2) {
>>         $KubeletArgList += @("--node-ip=$($ipAddrs -join ",")")
>>     } else {
>>         # error instead?
>>         Write-Host "Expected exactly 2 IPs for dual-stack, got $($ipAddrs.Length). IP addresses from interface ${defaultRouteIfIndex}: $($ipAddrs -join " ")"
>>     }
PS C:\>     $KubeletArgList
--node-ip=10.224.216.0,fd87:3fab:8538:e78c:2600::
```

Fixes https://github.com/kubernetes-sigs/cloud-provider-azure/issues/3213

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -31,6 +31,16 @@ if ($global:ContainerRuntime -eq "containerd") {
 
 ipmo $global:HNSModule
 
+function
+Test-NodeIPArgExists($argList) {
+    foreach ($arg in $argList) {
+        if ($arg.ToLower() -like "--node-ip*") {
```
paulgmiller, line 37: why the wild card?

paulgmiller: oh no this is just to --node-ip= 
Can we fully define to the = so we don't catch other random arguments?

tyler-lloyd: it's testing to see if `--node-ip={already provided my ip}` is in the kubelet args list already. So if for whatever reason something has already set it in the kubelet config we don't want to overwrite it (or maybe we do but I am taking the very passive approach here). Guess it could be clearer if I added the `=` in the search string

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -66,6 +76,22 @@ if ($global:ContainerRuntime -eq "containerd") {
     $KubeletArgList += @("--container-runtime=remote", "--container-runtime-endpoint=npipe://./pipe/containerd-containerd")
 }
 
+# Update args to get both IPs from NIC using the default route 0.0.0.0/0 if IsDualStackEnabled
+# and the --node-ip argument is not present in the existing kubelet arg list.
+if ($global:IsDualStackEnabled -and !(Test-NodeIPArgExists $KubeletArgList)) {
+    $defaultRouteIfIndex = (Get-NetRoute -DestinationPrefix 0.0.0.0/0).ifIndex
+
+    # use addresses from dhcp and sort by family so IPv4 is always first
```
paulgmiller, line 84: Somewhat nervous that we do this in windows but not linux. like to be consistent. another possiblity is using an annottion for dnc-rc and leaving this argument alone. Fine with this as short term mitigation but we might want to say this is only realyl needed by dnc-rc.

paulgmiller (approved):

staging/cse/windows/provisioningscripts/kubeletstart.ps1
tyler-lloyd: yeah I think adding an annotation would work, and +1 to putting it in Linux. imo I like using this argument for kubelet because then the node is created with both IPs and there's no period of waiting for the IPv6 address. currently, even in linux, the node gets created with just an IPv4 address and then after an arbitrary amount of time the node is patched with the IPv6 address. the inconsistency concern would be if we use --node-ip for dual-stack only but single-stack still just relies on kubelet defaulting

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -66,6 +76,22 @@ if ($global:ContainerRuntime -eq "containerd") {
     $KubeletArgList += @("--container-runtime=remote", "--container-runtime-endpoint=npipe://./pipe/containerd-containerd")
 }
 
+# Update args to get both IPs from NIC using the default route 0.0.0.0/0 if IsDualStackEnabled
+# and the --node-ip argument is not present in the existing kubelet arg list.
+if ($global:IsDualStackEnabled -and !(Test-NodeIPArgExists $KubeletArgList)) {
```
AbelHu, line 81: When will `Test-NodeIPArgExists $KubeletArgList` return `true`?

AbelHu, line 81: This script will be run by nssm.exe as a service. How will `IsDualStackEnabled` be set?

tyler-lloyd: If --node-ip was already set in the kubeletconfig. We don't do it now, but it's just taking a more cautious approach instead of overwriting the value if it does exist.

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -66,6 +76,22 @@ if ($global:ContainerRuntime -eq "containerd") {
...
+    $defaultRouteIfIndex = (Get-NetRoute -DestinationPrefix 0.0.0.0/0).ifIndex
+
+    # use addresses from dhcp and sort by family so IPv4 is always first
+    $ipAddrs = (Get-NetIPAddress -InterfaceIndex $defaultRouteIfIndex | ? { $_.SuffixOrigin -eq "Dhcp" } | Sort-Object AddressFamily).IPAddress
+
+    if ($ipAddrs.Length -eq 2) {
+        $KubeletArgList += @("--node-ip=$($ipAddrs -join ",")")
+    } else {
+        # error instead?
+        Write-Host "Expected exactly 2 IPs for dual-stack, got $($ipAddrs.Length). IP addresses from interface ${defaultRouteIfIndex}: $($ipAddrs -join " ")"
```
AbelHu, line 91: I think that we should set this parameter in `c:\k\kubeclusterconfig.json` in node provisioning so we can return a new error code for this in cse exitcode. I think that you can reference the function `Write-KubeClusterConfig` in staging/cse/windows/kubernetesfunc.ps1

staging/cse/windows/provisioningscripts/kubeletstart.ps1
tyler-lloyd: ah that's good to know. I updated it to reference the kubeproxy feature gates which will have "IPv6DualStack=true" instead

staging/cse/windows/provisioningscripts/kubeletstart.ps1
tyler-lloyd: will `Write-KubeClusterConfig` run even on a node restart? if a node gets a new IP between reboots will it write a new config? otherwise it could keep around stale IP data

staging/cse/windows/provisioningscripts/kubeletstart.ps1
tyler-lloyd: updated with `=`

staging/cse/windows/provisioningscripts/kubeletstart.ps1
AbelHu: cse will only be run once in provisioning the node. Could you share when the IP will be changed after a reboot? I did not know this before.

AbelHu: You can see that currently we also setup the node with the IP in cse

tyler-lloyd: also, I am not sure I think moving it to `Write-KubeConifg` is better. We already append additional arguments to the kubelet args here in kubeletstart.ps1 and we aren't doing that in `Write-KubeConfig`. What about just adding validation inside CSE to check `if ($global:IsDualStackEnabled) { expect 2 IPs on the NIC else return CSE error }` but keep the `--node-ip` setup here in kubeletstart?

AbelHu: @tyler-lloyd I think that the cse will be run again if the vm is started from deallocation. For IPv4, we always need to get the current host ip in provisioning Windows nodes in https://github.com/Azure/AgentBaker/blob/master/staging/cse/windows/azurecnifunc.ps1#L363

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -66,6 +77,22 @@ if ($global:ContainerRuntime -eq "containerd") {
     $KubeletArgList += @("--container-runtime=remote", "--container-runtime-endpoint=npipe://./pipe/containerd-containerd")
 }
 
+# Update args to get both IPs from NIC using the default route 0.0.0.0/0 if IsDualStackEnabled
+# and the --node-ip argument is not present in the existing kubelet arg list.
+if (($global:KubeproxyFeatureGates -contains "IPv6DualStack=true") -and !(Test-NodeIPArgExists $KubeletArgList)) {
```
AbelHu, line 82: could we move this to cse scripts to check whether it exists in `$Global:ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs` before writing it into `c:\k\kubeclusterconfig.json` in Write-KubeClusterConfig?
With this change,
1. we can return WINDOWS_CSE_ERROR_DUAL_STACK_NO_EXACT_TWO_IPS if we cannot get two IPs
2. we can append ipAddrs to KubeletArgList if KubeletArgList does not contain them
3. we do not need to change kubeletstart.ps1 and cannot do anything if it fails to get two IPs.

tyler-lloyd:
@AbelHu - I moved the logic to a new function `Update-KubeletConfigArgsForDualStack` in kubernetesfunc.ps1. This will get called right before the ClusterConfig json is written and does the validation/update there. Let me know if this is what you were thinking.

staging/cse/windows/provisioningscripts/kubeletstart.ps1
tyler-lloyd: sounds good, updated in https://github.com/Azure/AgentBaker/pull/2813/commits/e90990d18d93a6ff92c66b6569e548335dc3e972

staging/cse/windows/provisioningscripts/kubeletstart.ps1
tyler-lloyd: updated

staging/cse/windows/kubernetesfunc.ps1
```diff
@@ -129,9 +129,68 @@ function Write-KubeClusterConfig {
...
+        return
+    }
+
+    # nothing to do if not a dualstack clsuter or the --node-ip is already in the arg list
+    if ((!$global:IsDualStackEnabled) -or `
+        (Test-NodeIPArgExists -Args $ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs)) {
+        return
+    }
+
+    $defaultRouteIfIndex = (Get-NetRoute -DestinationPrefix 0.0.0.0/0).ifIndex
```
AbelHu, line 159: it may be better to exit with an error code when `Get-NetRoute -DestinationPrefix 0.0.0.0/0` returns empty and `.ifIndex` will fail if there is some bug in vmss instance's network

staging/cse/windows/kubernetesfunc.ps1
```diff
@@ -129,9 +129,68 @@ function Write-KubeClusterConfig {
...
+
+    # use addresses from dhcp and sort by family so IPv4 is always first
+    $ipAddrs = (Get-NetIPAddress -InterfaceIndex $defaultRouteIfIndex | `
```
AbelHu, line 162: Same here.

staging/cse/windows/kubernetesfunc.ps1
```diff
@@ -129,9 +129,68 @@ function Write-KubeClusterConfig {
...
+    Update-KubeletConfigArgsForDualStack -ClusterConfiguration $Global:ClusterConfiguration
+
     $Global:ClusterConfiguration | ConvertTo-Json -Depth 10 | Out-File -FilePath $global:KubeClusterConfigPath
 }
 
+# Update-KubeletConfigArgsForDualStack updates the $Global:ClusterConfiguration object
+# to append the --node-ip argument for kubelet if the cluster is dualstack and the
+# host is configured with an IPv4 and IPv6 on the interface using the default route.
+# If the host does not have both IPs it will raise erro WINDOWS_CSE_ERROR_DUAL_STACK_NO_EXACT_TWO_IPS
+function Update-KubeletConfigArgsForDualStack {
```
AbelHu, line 141: Could you help to test this change with a private cse package and a agentbaker tag in RP to validate whether this works with both dualstack and non-dualstack?

1. Create a folder aks-window-cse and copy all files including folder in staging/cse/windows to this folder
2. Use `zip -j aks-windows-cse-scripts-v1.0.22.zip aks-windows-cse/*` (Linux command) to create a test cse script pacakge. NOTE: Please not reuse the current cse script package version since the cache in AKS Windows image will be used instead of downloading it
3. Upload this cse package to a storage container and assure that it can be downloaded with anonymous permission
4. Update `parts/windows/kuberneteswindowssetup.ps1` in your test branch to set `CSEScriptsPackageUrl` to the full url to your test cse package in the storage account
5. Run `make generate` and commit it in your test branch
6. Create a test tag **in your forked repo**, for exmaple, ` v1.20230216.6` (The latest absvc tag in RP is `v0.20230216.5`)
7. Update absvc tag in RP and test both

tyler-lloyd:
likely going to close this PR as --node-ip with dual-stack IPs is actually not supported if using cloud-provider: https://github.com/kubernetes/kubernetes/blob/efe20f6c9b54fbe36b4ff4c47d7d0bc857699b5e/cmd/kubelet/app/server.go#L1137

tyler-lloyd:
closing due to incompatibility with dual-stack `--node-ip` arg and using a cloud provider

PR #2814, opened 2023-02-28
fix: curl proxy cse vars; e2e for noProxy

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/nodebootstrapping_template.json
```diff
@@ -141,7 +141,9 @@
     "ContainerdPackageURL": "",
     "RuncPackageURL": "",
     "FIPSEnabled": false,
-    "HTTPProxyConfig": null,
+    "HTTPProxyConfig": {
```
alexeldeib, line 144: doesn't do anything but tests var syntax in bash

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -76,6 +76,9 @@ if [[ "${SHOULD_CONFIGURE_CUSTOM_CA_TRUST}" == "true" ]]; then
 fi
 
 if [[ -n "${OUTBOUND_COMMAND}" ]]; then
+    if [[ -n "${PROXY_VARS}" ]]; then
+        eval $PROXY_VARS
```
alexeldeib, line 80: kinda annoying to form this export string properly without eval...

don't love eval, but pretty sure we sanitize this reasonably well in RP

and if we don't...congratulations, you're paying to hack your own VMs via shell script injection

anujmaheshwari1 (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -63,6 +63,9 @@ if [[ "${DISABLE_SSH}" == "true" ]]; then
     disableSSH || exit $ERR_DISABLE_SSH
 fi
 
+apt update && apt install -yq tinyproxy
```
alexeldeib: this works but doesn't handle https serving endpoint for HTTPS_PROXY

you can set HTTPS_PROXY to http://url, but it's not testing full scenario

PR #2823, opened 2023-03-01
fix: clean dynamic-config-dir in CSE as well as custom data

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker_test.go
```diff
@@ -142,6 +142,17 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 			configUpdator(config)
 		}
 
+		// CSE
+		cseCommand := baker.GetNodeBootstrappingCmd(config)
```
alexeldeib, line 146: GetNodeBootstrappingPayload (customData) cleaned the input, then used by GetNodeBootStrappingCmd (CSE)

flipping the order and showing the testdata doesn't change to add --dyamic-config-dir is the validation of the change

previous to this, the flag was still present in CSE.

PR #2827, opened 2023-03-02

HDYA:


HDYA:
this should be merged IF no existing build will be used to release, otherwise merge https://github.com/Azure/AgentBaker/pull/2786

PR #2828, opened 2023-03-02
chore: resolve conflict of creating in AgentBaker Windows E2E

junjiezhang1997:
**What type of PR is this?**
/kind fix

1. Conflict may happen in AgentBaker Windows E2E pipeline when multiple pipelines try to create new clusters or nodepools at the same time. Add code to resolve conflict.
2. Conflict may happen in AgentBaker Windows E2E pipeline when multiple pipelines try to execute "kubectl" command after creating the cluster. Use "az vmss run-command" to upload linux cert files.
3. Add missing "retval=$?" in previous code.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -2,6 +2,15 @@ parameters:
   - name: windowsImage
     type: string
     default: 2019-containerd
+  - name: windowsOSSKU
+    type: string
+    default: Windows2019
+  - name: windowsVMSize
+    type: string
+    default: Standard_D2s_v3
+  - name: storageAccount
+    type: string
+    default: storewin19
 
 jobs:
 - job: Setup_Test_Cluster
@@ -13,14 +22,18 @@ jobs:
     - bash: bash .pipelines/scripts/setup_go.sh
       displayName: setup go
     - bash: |
+        export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
+        export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
+        export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
+        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}
+        mkdir -p $WINDOWS_E2E_IMAGE
+        cp -r e2e/* $WINDOWS_E2E_IMAGE
         go version
-        cd e2e
-        WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
-        export WINDOWS_E2E_IMAGE
+        cd $WINDOWS_E2E_IMAGE
         bash ./e2e-starter.sh
         bash e2e-create-windows-nodepool.sh
       displayName: Setup Test Cluster
-    - publish: $(System.DefaultWorkingDirectory)/e2e
+    - publish: $(System.DefaultWorkingDirectory)/${{ parameters.windowsImage }}
       artifact: ${{ parameters.windowsImage }}-clusterConfig
       condition: always()
 
@@ -47,13 +60,13 @@ jobs:
       displayName: az login
     - bash: |
         export SCENARIO_NAME
-        WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
-        export WINDOWS_E2E_IMAGE
-        cp -a $(Pipeline.Workspace)/${{ parameters.windowsImage }}-clusterConfig/* e2e
-        cd e2e
+        export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
+        mkdir -p $WINDOWS_E2E_IMAGE
+        cp -a $(Pipeline.Workspace)/${{ parameters.windowsImage }}-clusterConfig/* $WINDOWS_E2E_IMAGE
+        cd $WINDOWS_E2E_IMAGE
         bash e2e-windows-scenario.sh
       env:
         MAPPED_ACCOUNT_KEY: $(STORAGE_ACCOUNT_KEY)
-    - publish: $(System.DefaultWorkingDirectory)/e2e/$(SCENARIO_NAME)-logs
+    - publish: $(System.DefaultWorkingDirectory)/${{ parameters.windowsImage }}/$(SCENARIO_NAME)-logs
       artifact: ${{ parameters.windowsImage }}-$(SCENARIO_NAME)-logs
       condition: always()
--- a/e2e/e2e-create-windows-nodepool.sh
+++ b/e2e/e2e-create-windows-nodepool.sh
@@ -7,7 +7,7 @@ source e2e-helper.sh
 log "Starting to create windows nodepool"
 
 if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
-    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"
+    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
 fi
 
 out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME | jq '.[].name')
@@ -17,16 +17,10 @@ if [[ "$out" != *"winnp"* ]]; then
     retval=0
     az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku $WINDOWS_E2E_OSSKU --node-vm-size $WINDOWS_E2E_VMSIZE --node-count 1 || retval=$?
     if [ "$retval" -ne 0 ]; then
-        provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
-        if [ "$provisioning_state" == "Creating" ]; then
-            log "Other pipeline may be creating the same nodepool, waiting for ready"
-            az aks nodepool wait --nodepool-name "winnp" --cluster-name $CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --created --interval 60 --timeout 1800
-        elif [ "$provisioning_state" != "Succeeded" ]; then
-            err "Failed to create windows nodepool"
-            exit 1
-        fi
+        log "Other pipeline may be creating the same nodepool, waiting for ready"
+    else
+        log "Created windows nodepool"
     fi
-    log "Created windows nodepool"
 else
     log "Already create windows nodepool"
 fi
--- a/e2e/e2e-create-windows-nodepool.sh
+++ b/e2e/e2e-create-windows-nodepool.sh
@@ -35,6 +35,6 @@ provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOU
 if [ "$provisioning_state" == "Succeeded" ]; then
     log "Windows nodepool is in succeed state"
 else
-    err "Windows nodepool is not in succeeded state"
+    err "Windows nodepool is not in succeeded state. Current state is $provisioning_state."
     exit 1
 fi
--- a/e2e/e2e-create-windows-nodepool.sh
+++ b/e2e/e2e-create-windows-nodepool.sh
@@ -14,17 +14,33 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME 
 
 if [[ "$out" != *"winnp"* ]]; then
     log "Creating windows nodepool"
-    if [ "$WINDOWS_E2E_IMAGE" == "2019-containerd" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --node-count 1
-    elif [ "$WINDOWS_E2E_IMAGE" == "2022-containerd" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_D2_v2 --node-count 1
-    elif [ "$WINDOWS_E2E_IMAGE" == "2022-containerd-gen2" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_DS2_v2 --node-count 1
-    else
-        err "Invalid windows e2e image version"
-        exit 1
+    retval=0
+    az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku $WINDOWS_E2E_OSSKU --node-vm-size $WINDOWS_E2E_VMSIZE --node-count 1 || retval=$?
+    if [ "$retval" -ne 0 ]; then
+        provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
+        if [ "$provisioning_state" == "Creating" ]; then
+            log "Other pipeline may be creating the same nodepool, waiting for ready"
+            az aks nodepool wait --nodepool-name "winnp" --cluster-name $CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --created --interval 60 --timeout 1800
+        elif [ "$provisioning_state" != "Succeeded" ]; then
+            err "Failed to create windows nodepool"
+            exit 1
+        fi
     fi
     log "Created windows nodepool"
 else
     log "Already create windows nodepool"
+fi
+
+provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
+if [ "$provisioning_state" == "Creating" ]; then
+    log "Other pipeline may be creating the same nodepool, waiting for ready"
+    az aks nodepool wait --nodepool-name "winnp" --cluster-name $CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --created --interval 60 --timeout 1800
+fi
+
+provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
+if [ "$provisioning_state" == "Succeeded" ]; then
+    log "Windows nodepool is in succeed state"
+else
+    err "Windows nodepool is not in succeeded state"
+    exit 1
 fi
--- a/e2e/e2e-helper.sh
+++ b/e2e/e2e-helper.sh
@@ -33,7 +33,7 @@ upload_linux_file_to_storage_account() {
     set +x
     expiryTime=$(date --date="2 day" +%Y-%m-%d)
     account_key=$(az storage account keys list --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --query "[0].value")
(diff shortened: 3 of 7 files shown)


e2e/e2e-create-windows-nodepool.sh
```diff
@@ -15,16 +15,36 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME
...
+        errorMessage=$(az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --node-count 1 --only-show-errors 2>&1)
     elif [ "$WINDOWS_E2E_IMAGE" == "2022-containerd" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_D2_v2 --node-count 1
+        errorMessage=$(az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_D2_v2 --node-count 1 --only-show-errors 2>&1)
     elif [ "$WINDOWS_E2E_IMAGE" == "2022-containerd-gen2" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_DS2_v2 --node-count 1
+        errorMessage=$(az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_DS2_v2 --node-count 1 --only-show-errors 2>&1)
```
AbelHu, line 22: it seems like that we can create two variables ossku and vmsize per $WINDOWS_E2E_IMAGE and only use one command to create the windows agent pool. It should be more simple

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -15,16 +15,36 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME
...
     else
         err "Invalid windows e2e image version"
         exit 1
     fi
+    if [ "$errorMessage" != "" ]; then
```
AbelHu, line 27: can we check whether it fails because other pipeline is also creating it? I think that we should not wait if it fails with other errors, for example, capacity issue

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -15,16 +15,36 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME
...
+        log "Other pipeline may be creating the same nodepool, waiting for ready"
+        az aks nodepool wait --nodepool-name "winnp" --cluster-name $CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --created --interval 60 --timeout 1800
+        sleep 5
```
AbelHu, line 30: why do we sleep 5 seconds?

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -15,16 +15,36 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME
...
     elif [ "$WINDOWS_E2E_IMAGE" == "2022-containerd-gen2" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_DS2_v2 --node-count 1
+        errorMessage=$(az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_DS2_v2 --node-count 1 --only-show-errors 2>&1)
     else
         err "Invalid windows e2e image version"
         exit 1
     fi
+    if [ "$errorMessage" != "" ]; then
+        log "Other pipeline may be creating the same nodepool, waiting for ready"
+        az aks nodepool wait --nodepool-name "winnp" --cluster-name $CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --created --interval 60 --timeout 1800
```
AbelHu, line 29: should we fail if it exits after the timeout 1800 but the new agent pool is still not available?

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -15,16 +15,36 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME
...
+        sleep 5
+    fi
     log "Created windows nodepool"
 else
     log "Already create windows nodepool"
 fi
+
+provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
+if [ "$provisioning_state" == "Creating" ]; then
```
AbelHu, line 38: same here and below to wait it to be Succeeded.

e2e/e2e-windows-scenario.sh
```diff
@@ -101,6 +101,7 @@ WINDOWS_PASSWORD=$({
     choose '0123456789'
     choose 'abcdefghijklmnopqrstuvwxyz'
     choose 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+    choose '#*'
```
AbelHu, line 104: this may still do not work as expected to generate a password with a special character with # or *

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: Just to avoid the inconsistency between the result of "az aks nodepool wait" and the actual status of windows nodepool.

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: Resolved.

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: We can check this at the end of this file.

e2e/e2e-windows-scenario.sh
junjiezhang1997: ‘*’ causes a problem in generating password. I changed it to '@' and tested it.

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: check at the end of this file.

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: check at the end of the file.

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -16,7 +22,11 @@ jobs:
         go version
         cd e2e
         WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
```
AbelHu, line 24: export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}?

e2e/e2e-create-windows-nodepool.sh
AbelHu: Cannot get your point

junjiezhang1997: Just checked in [here](https://learn.microsoft.com/en-us/cli/azure/aks/nodepool?view=azure-cli-latest#az-aks-nodepool-wait-optional-parameters). '--created' parameter guarantees that this command will wait until provisioningState becomes 'Succeeded'. Will delete the sleep time here.

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -14,17 +14,33 @@ out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME
...
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_D2_v2 --node-count 1
-    elif [ "$WINDOWS_E2E_IMAGE" == "2022-containerd-gen2" ]; then
-        az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku Windows2022 --node-vm-size Standard_DS2_v2 --node-count 1
-    else
-        err "Invalid windows e2e image version"
-        exit 1
+    retval=0
+    az aks nodepool add --resource-group $RESOURCE_GROUP_NAME --cluster-name $CLUSTER_NAME --name "winnp" --os-type Windows --os-sku $WINDOWS_E2E_OSSKU --node-vm-size $WINDOWS_E2E_VMSIZE --node-count 1 || retval=$?
+    if [ "$retval" -ne 0 ]; then
+        provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
```
AbelHu, line 20: Do we need line 20-26 since line 34-46 will wait it to be ready?

e2e/e2e-helper.sh
```diff
@@ -16,6 +16,97 @@ exec_on_host() {
...
+upload_linux_file_to_storage_account() {
+    local retval=0
+    MC_RESOURCE_GROUP_NAME="MC_${RESOURCE_GROUP_NAME}_${CLUSTER_NAME}_$LOCATION"
+    MC_VMSS_NAME=$(az vmss list -g $MC_RESOURCE_GROUP_NAME --query "[?contains(name, 'nodepool')]" -ojson | jq -r '.[0].name')
+    VMSS_INSTANCE_ID="$(az vmss list-instances --name $MC_VMSS_NAME -g $MC_RESOURCE_GROUP_NAME | jq -r '.[0].instanceId')"
+
+    set +x
+    expiryTime=$(date --date="2 day" +%Y-%m-%d)
+    account_key=$(az storage account keys list --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --query "[0].value")
+    token=$(az storage container generate-sas --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --account-key $account_key --permissions 'rwacdl' --expiry $expiryTime --name "tmp")
```
AbelHu, line 36: do we need 'rwacdl' to only upload the file?

e2e/e2e-helper.sh
```diff
@@ -16,6 +16,97 @@ exec_on_host() {
...
+    local retval=0
+    for i in $(seq 1 20); do
+        account_key=$(az storage account keys list --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --query "[0].value") || retval=$?
+        if [ "$retval" -ne 0 ]; then
+            sleep 10
+            continue
+        fi
+        break;
+    done
+    token=$(az storage container generate-sas --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --account-key $account_key --permissions 'rwacdl' --expiry $expiryTime --name "tmp")
```
AbelHu, line 79: can we only grant 'r' to download the file?

e2e/e2e-starter.sh
```diff
@@ -42,10 +53,24 @@ fi
...
     if [[ "$RESOURCE_GROUP_NAME" == *"windows"* ]]; then
-        az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys --network-plugin azure -ojson
+        az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys --network-plugin azure -ojson || retval=$?
     else
-        az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys --network-plugin kubenet -ojson
+        az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys --network-plugin kubenet -ojson || retval=$?
+    fi
+
+    if [ "$retval" -ne 0  ]; then
+        log "Other pipelines may be creating cluster $CLUSTER_NAME, waiting for ready"
```
AbelHu, line 64: would this happen since it seems like that only the pipeline creating the cluster successfully will set create_cluster to true?

e2e/e2e-starter.sh
```diff
@@ -63,22 +88,30 @@ az vmss list -g $MC_RESOURCE_GROUP_NAME --query "[?contains(name, 'nodepool')]"
...
-exec_on_host "cat /etc/kubernetes/certs/apiserver.crt | base64 -w 0" apiserver.crt
-exec_on_host "cat /etc/kubernetes/certs/ca.crt | base64 -w 0" ca.crt
-exec_on_host "cat /etc/kubernetes/certs/client.key | base64 -w 0" client.key
-exec_on_host "cat /var/lib/kubelet/bootstrap-kubeconfig" bootstrap-kubeconfig
-
-clusterInfoEndTime=$(date +%s)
-log "Retrieved cluster info in $((clusterInfoEndTime-clusterInfoStartTime)) seconds"
+if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
+    if [ "$create_cluster" == "true" ]; then
+        create_storage_account
```
AbelHu, line 93: NOTE: the storage account is unique so it will fail if the storage account exists

AbelHu: One more question: When and who will delete the resource group?

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: Sure. I will modify code here.

e2e/e2e-helper.sh
junjiezhang1997: Will test permissions needed and change code.

e2e/e2e-starter.sh
junjiezhang1997: Yes. Even the pipeline didn't find any cluster at first, it could meet conflict when trying to create a cluster.

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -7,24 +7,34 @@ source e2e-helper.sh
 log "Starting to create windows nodepool"
 
 if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
-    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"
+    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
```
AbelHu, line 10: Q: Is this used to avoid conflict with current pipelines?

e2e/e2e-create-windows-nodepool.sh
```diff
@@ -7,24 +7,34 @@ source e2e-helper.sh
...
+if [ "$provisioning_state" == "Creating" ]; then
+    log "Other pipeline may be creating the same nodepool, waiting for ready"
+    az aks nodepool wait --nodepool-name "winnp" --cluster-name $CLUSTER_NAME --resource-group $RESOURCE_GROUP_NAME --created --interval 60 --timeout 1800
+fi
+
+provisioning_state=$(az aks nodepool show --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME -n "winnp" -ojson | jq '.provisioningState' | tr -d "\"")
+if [ "$provisioning_state" == "Succeeded" ]; then
+    log "Windows nodepool is in succeed state"
+else
+    err "Windows nodepool is not in succeeded state"
```
AbelHu, line 38: Could we log the unexpected state for troubleshooting?

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: Yes. Also avoid conflict of hotfix on official branch.

e2e/e2e-create-windows-nodepool.sh
junjiezhang1997: Sure. Will add it.

e2e/e2e-windows-scenario.sh
```diff
@@ -15,36 +15,35 @@ collect-logs() {
...
         --resource-group $MC_RESOURCE_GROUP_NAME \
         --name $DEPLOYMENT_VMSS_NAME \
         --instance-id $VMSS_INSTANCE_ID \
         --scripts @upload-cse-logs.ps1 \
-        --parameters arg1=$STORAGE_ACCOUNT_NAME arg2=$STORAGE_LOG_CONTAINER arg3=$DEPLOYMENT_VMSS_NAME arg4=$token
-    if [ "$retval" != "0" ]; then
-        echo "failed in uploading cse logs"
+        --parameters arg1=$STORAGE_ACCOUNT_NAME arg2=$STORAGE_LOG_CONTAINER arg3=$DEPLOYMENT_VMSS_NAME arg4=$token || retval=$?
+    if [ "$retval" -ne 0 ]; then
+        err "failed in uploading cse logs"
```
AbelHu, line 27: same to log error and error code in the log

e2e/e2e-windows-scenario.sh
```diff
@@ -15,36 +15,35 @@ collect-logs() {
...
-    if [ "$retval" != "0" ]; then
-        echo "failed in downloading cse logs"
+    ${array[0]}/azcopy copy "https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/${STORAGE_LOG_CONTAINER}/${DEPLOYMENT_VMSS_NAME}-cse.log?${tokenWithoutQuote}" $SCENARIO_NAME-logs/$WINDOWS_E2E_IMAGE-CustomDataSetupScript.log || retval=$?
+    if [ "$retval" -ne 0 ]; then
+        err "failed in downloading cse logs"
     else
-        ${array[0]}/azcopy rm "https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/${STORAGE_LOG_CONTAINER}/${DEPLOYMENT_VMSS_NAME}-cse.log?${tokenWithoutQuote}"
-        if [ "$retval" != "0" ]; then
-            echo "failed in deleting cse logs in remote storage"
+        echo "collect cse logs done"
```
AbelHu, line 37: same to log error and error code in the log

e2e/e2e-starter.sh
```diff
@@ -63,22 +88,30 @@ az vmss list -g $MC_RESOURCE_GROUP_NAME --query "[?contains(name, 'nodepool')]"
...
-
-exec_on_host "cat /etc/kubernetes/azure.json" fields.json
-exec_on_host "cat /etc/kubernetes/certs/apiserver.crt | base64 -w 0" apiserver.crt
-exec_on_host "cat /etc/kubernetes/certs/ca.crt | base64 -w 0" ca.crt
-exec_on_host "cat /etc/kubernetes/certs/client.key | base64 -w 0" client.key
-exec_on_host "cat /var/lib/kubelet/bootstrap-kubeconfig" bootstrap-kubeconfig
-
-clusterInfoEndTime=$(date +%s)
-log "Retrieved cluster info in $((clusterInfoEndTime-clusterInfoStartTime)) seconds"
+if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
```
AbelHu, line 91: @alexeldeib  FYI: @junjiezhang1997 found that Linux also hit the same issue in getting the files from the Linux VMSS instance.

AbelHu:
@junjiezhang1997 Please investigate why "AgentBaker E2E" failed.

e2e/e2e-starter.sh
junjiezhang1997: For example: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=70452078&view=logs&j=df77c11a-d70e-5fda-f10b-e543dae019aa&t=42160a71-98d9-5950-cc5c-ad5fea7cc94c

Actually, Linux has move e2e to go so I think change in e2e-starter.sh would not influence Linux AgentBaker E2E pipeline. Just use this if condition to avoid any conflict.

junjiezhang1997:
> @junjiezhang1997 Please investigate why "AgentBaker E2E" failed.

It seems like after the resource group is deleted automatically, the pipeline failed at Line 14 of e2e/cluster.go rather than try to create a new cluster. @alexeldeib Need your help to fix it, thanks very much!

AbelHu (approved):

PR #2829, opened 2023-03-02
feat: Enable Multi-Instance GPU Support for Mariner AKS

henryli001:
**What type of PR is this?**

This PR adds support for MIG mode for Azure Linux: 
- 1. nvidia-smi takes longer than the timeout value set in the code to return success on A100 SKU. Based on current investigation, the issue seems to be caused by the NVIDIA GPU driver and thus, enabling persistence mode to accelerate nvidia-smi operations later on.
- 2. install nvidia fabric manager when the condition is satisifed

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -438,7 +438,15 @@ ensureMigPartition(){
 [Service]
 Environment="GPU_INSTANCE_PROFILE=${GPU_INSTANCE_PROFILE}"
 EOF
-    systemctlEnableAndStart mig-partition || exit $ERR_SYSTEMCTL_START_FAIL
+    # Noticed on Mariner that "nvidia-smi -mig 1" does not always enable MIG mode successfully and
+    # requires a reboot to fully actiavte MIG mode. Thus, as a workaround, we'll only run the MIG enable
+    # operation here and do the rest of the GPU instances creation after the reboot
+    if [[ $OS == $MARINER_OS_NAME ]]; then
+        retrycmd_if_failure 24 5 120 nvidia-smi -mig 1 || exit $ERR_GPU_DRIVERS_START_FAIL
+        sed '6d' /etc/systemd/system/mig-partition.service
```
alexeldeib, line 446: what's this?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -438,7 +438,15 @@ ensureMigPartition(){
 [Service]
 Environment="GPU_INSTANCE_PROFILE=${GPU_INSTANCE_PROFILE}"
 EOF
-    systemctlEnableAndStart mig-partition || exit $ERR_SYSTEMCTL_START_FAIL
+    # Noticed on Mariner that "nvidia-smi -mig 1" does not always enable MIG mode successfully and
+    # requires a reboot to fully actiavte MIG mode. Thus, as a workaround, we'll only run the MIG enable
+    # operation here and do the rest of the GPU instances creation after the reboot
+    if [[ $OS == $MARINER_OS_NAME ]]; then
+        retrycmd_if_failure 24 5 120 nvidia-smi -mig 1 || exit $ERR_GPU_DRIVERS_START_FAIL
```
alexeldeib, line 445: did we find out if ubuntu has the same issue?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -333,6 +333,10 @@ if $REBOOTREQUIRED; then
         # logs_to_events should not be run on & commands
         aptmarkWALinuxAgent unhold &
     fi
+    # Start the mig-partition.service for Mariner after rebooting to complete the rest ofd the cxonfigurations
+    if [[ $OS == $MARINER_OS_NAME ]]; then
+        systemctlEnableAndStart mig-partition || exit $ERR_SYSTEMCTL_START_FAIL
```
alexeldeib, line 338: isn't this already required elsewhere for both distros?

parts/linux/cloud-init/artifacts/cse_config.sh
henryli001: Have revised the implementation for mig-partition.service. This code no longer applies.

parts/linux/cloud-init/artifacts/cse_config.sh
henryli001: On Ubuntu "nvidia-smi -mig 1" also does not immediately enable MIG unless performing a reboot, which is the same as on Mariner. Have eliminated this chunk of code since the root cause of the failure of mig-partition.service is related to timeout.

parts/linux/cloud-init/artifacts/cse_main.sh
henryli001: This code has already been eliminated.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -34,6 +34,14 @@ downloadGPUDrivers() {
     if ! dnf_install 30 1 600 cuda-${CUDA_VERSION}; then
       exit $ERR_APT_INSTALL_TIMEOUT
     fi
+
+    # Check the NVIDIA driver version installed and install nvidia-fabric-manager
+    NVIDIA_DRIVER_VERSION=$(cut -d - -f 2 <<< "$(rpm -qa cuda)")
+    for nvidia_package in nvidia-fabric-manager-${NVIDIA_DRIVER_VERSION} nvidia-fabric-manager-devel-${NVIDIA_DRIVER_VERSION}; do
+      if ! dnf_install 30 1 600 $nvidia_package; then
```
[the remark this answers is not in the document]
henryli001: Makes sense. Have already implemented installing fabric manager in a separate function and only call this function when "${GPU_NEEDS_FABRIC_MANAGER}" evaluates to True.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -560,6 +568,13 @@ configGPUDrivers() {
     if [[ $OS == $UBUNTU_OS_NAME ]]; then
         retrycmd_if_failure 120 5 25 nvidia-modprobe -u -c0 || exit $ERR_GPU_DRIVERS_START_FAIL
     fi
+
+    # Noticed that NVIDIA driver version R515 and later versions will experience long loading time of nvidia-smi on 
+    # ND A100-v4 and NDm A100-v4 series. For now, use the workaround solution to enable NVIDIA persistence mode which 
+    # will accelerate the loading time of future nvidia-smi operations
+    if [[ $OS == $MARINER_OS_NAME ]]; then
+        retrycmd_if_failure 120 5 35 nvidia-smi -pm 1 || exit $ERR_GPU_DRIVERS_START_FAIL
```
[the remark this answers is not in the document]
henryli001: I was also able to capture this on Ubuntu when testing using ab e2e. One thing I would like to open for discussion is that "nvidia-smi -pm 1" will also take quite long to complete (> 25sec on Mariner), in that case, from the perspective of total cluster creation time, do we go for enabling persistence mode and then check "nvidia-smi" or simply increasing the timeout of nvidia-smi?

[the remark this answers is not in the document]
alexeldeib: when do we get the benefits of persistence mode? on reboot when we don't have to reload nvidia-smi? or any time we invoke it?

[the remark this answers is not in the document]
henryli001: After enabling persistence mode, we'll expect nvidia-smi or any applications that attempt to trigger GPU initialization to complete much faster. But this will only occur in the current session. If you do a reboot, then persistence mode will be disabled and needs to re-activated. 

Turning on persistence mode will also bring the benefits of data persistence. Otherwise the driver state won't be preserved and it will revert back to default when the GPU is initialized again next time.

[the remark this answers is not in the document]
henryli001: Replaced "nvidia-smi -pm 1" with using nvidia persistence daemon instead because "nvidia-smi -pm 1" is a legacy solution that will be deprecated.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -560,6 +560,13 @@ configGPUDrivers() {
     if [[ $OS == $UBUNTU_OS_NAME ]]; then
         retrycmd_if_failure 120 5 25 nvidia-modprobe -u -c0 || exit $ERR_GPU_DRIVERS_START_FAIL
     fi
+
+    # Noticed that NVIDIA driver version R515 and later versions will experience long loading time of nvidia-smi on 
+    # ND A100-v4 and NDm A100-v4 series. For now, use the workaround solution to enable NVIDIA persistence mode which 
+    # will accelerate the loading time of future nvidia-smi operations
+    if [[ $OS == $MARINER_OS_NAME ]]; then
+        retrycmd_if_failure 120 5 35 nvidia-persistenced --persistence-mode || exit $ERR_GPU_DRIVERS_START_FAIL
```
alexeldeib, line 568: is this sufficient for reboots, cc @henryli001 ?

should we install a systemd service and enable + start that instead?

ganeshkumar5699: Reboots will reset/turn off persistence. Question is do we want to enable persistenced by default for all SKUs, and always

ganeshkumar5699: Can create a systemd service for this, if we want to have above experience

henryli001: +1 to what Ganeshkumar's response. I don't have customer usage data to justify whether this should be generalized back when I developed this, so I only implemented in this way so as to resolve the provisioning error.

pkg/agent/testdata/RawUbuntu/line9.sh
```diff
@@ -228,7 +228,7 @@ version_gte() {
 }
 
 systemctlEnableAndStart() {
-    systemctl_restart 100 5 30 $1
+    systemctl_restart 100 5 300 $1
```
hbeberman, line 231: I'm a bit concerned about this stretching out the time before service start failures are detected for everything else invoking systemctlEnableAndStart.

Could this timeout be parameterized so that every other invocation maintains the 30s timeout?

henryli001: Have reverted this change as it is no longer needed for the latest R525 driver version.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -50,6 +60,28 @@ installNvidiaContainerRuntime() {
     done
 }
 
+enableNvidiaPersistenceMode() {
```
alexeldeib, line 63: silly question. would it make sense to package this for mariner?

alexeldeib (approved):
think this lgtm, excited for A100 on Azure Linux :)

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
henryli001: Not at all, this is a great question! I actually looked into this and nvidia-persistenced is actually being [packaged](https://github.com/NVIDIA/yum-packaging-nvidia-persistenced/blob/main/nvidia-persistenced.spec) by the upstream. But when I checked the [source code](https://download.nvidia.com/XFree86/nvidia-persistenced/), looks like NVIDIA does not publish the source for each driver version, including the R525.85.12 that Azure Linux is supporting. 

Here's the [link](https://github.com/NVIDIA/nvidia-persistenced/issues/9#issuecomment-1555163280) to the github issue I filed to the upstream community regarding why source code for R525.85.12 is not released and the answer I got back. 

I also researched on the binaries/files shipped by the upstream nvidia-persistenced [package](https://rpmfind.net/linux/RPM/rpmfusion/nonfree/fedora/devel/x86_64/Packages/n/nvidia-persistenced-535.86.05-1.fc39.x86_64.html). The only instances that are related to actual functionality are **/usr/bin/nvidia-persistenced** and **/usr/lib/systemd/system/nvidia-persistenced.service** whereas the other files are either man pages or license files. /usr/bin/nvidia-persistenced is already shipped by the cuda rpm and thus the only thing left is the service file and since this file is not subject to changes in each release, I don't see a benefit of forcefully binding it in a package in Mariner unless we do it systematically like what the upstream does.

henryli001: @adithyaj thoughts on this?

hbeberman (approved):

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
alexeldeib: makes sense, I actually hit exactly the same issue :D I think the source from https://developer.download.nvidia.com/compute/cuda/redist/ _might_ have fabric manager source? but unsure.

not a blocker

henryli001: Yes, this link does contain the fabric manager source. This is actually where I grabbed the source to package fabric manager in Mariner. Unfortunately, I didn't see nvidia-persistenced source over there.

alexeldeib: d'oh, component confusion

PR #2843, opened 2023-03-08
test: decode customdata/cse and enable testing against contents

alexeldeib:
**What type of PR is this?**

/kind test

enable testing on generated testdata instead of using human eyeballs to guide snapshots

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -59,7 +59,7 @@ func (t *TemplateGenerator) getLinuxNodeCustomDataJSONObject(config *datamodel.N
 // GetWindowsNodeCustomDataJSONObject returns Windows customData JSON object in the form
 // { "customData": "<customData string>" }
 func (t *TemplateGenerator) getWindowsNodeCustomDataJSONObject(config *datamodel.NodeBootstrappingConfiguration) string {
-	// validtae and fix input
+	// validate and fix input
 	validateAndSetWindowsNodeBootstrappingConfiguration(config)
 
 	cs := config.ContainerService
--- a/pkg/agent/baker_test.go
+++ b/pkg/agent/baker_test.go
@@ -1,26 +1,62 @@
 package agent
 
 import (
+	"bytes"
+	"compress/gzip"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
+	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/Azure/agentbaker/pkg/agent/datamodel"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/barkimedes/go-deepcopy"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
 )
 
 func generateTestData() bool {
 	return os.Getenv("GENERATE_TEST_DATA") == "true"
 }
 
+// this regex looks for groups of the following forms, returning KEY and VALUE as submatches
+// - KEY=VALUE
+// - KEY="VALUE"
+// - KEY=
+// - KEY="VALUE WITH WHITSPACE"
+const cseRegexString = `([^=\s]+)=(\"[^\"]*\"|[^\s]*)`
+
+type nodeBootstrappingOutput struct {
+	customData string
+	cseCmd     string
+	files      map[string]*decodedValue
+	vars       map[string]string
+}
+
+type decodedValue struct {
+	encoding cseVariableEncoding
+	value    string
+}
+
+type cseVariableEncoding string
+
+const (
+	cseVariableEncodingBase64 cseVariableEncoding = "base64"
+	cseVariableEncodingGzip   cseVariableEncoding = "gzip"
+)
+
+type outputValidator func(*nodeBootstrappingOutput)
+
 var _ = Describe("Assert generated customData and cseCmd", func() {
-	DescribeTable("Generated customData and CSE", func(folder, k8sVersion string, configUpdator func(*datamodel.NodeBootstrappingConfiguration)) {
+	DescribeTable("Generated customData and CSE", func(folder, k8sVersion string, configUpdator func(*datamodel.NodeBootstrappingConfiguration), validator outputValidator) {
 		cs := &datamodel.ContainerService{
 			Location: "southcentralus",
 			Type:     "Microsoft.ContainerService/ManagedClusters",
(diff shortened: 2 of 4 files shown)


pkg/agent/baker_test.go
```diff
@@ -915,10 +919,89 @@ func backfillCustomData(folder, customData string) {
 	if strings.Contains(folder, "AKSWindows") {
 		return
 	}
-	err := exec.Command("/bin/sh", "-c", fmt.Sprintf("./testdata/convert.sh testdata/%s", folder)).Run()
+	err := decodeCustomDataFiles(fmt.Sprintf("testdata/%s", folder))
 	Expect(err).To(BeNil())
 }
 
+func decodeCustomDataFiles(dir string) error {
+	files, err := filepath.Glob(filepath.Join(dir, "*.sh"))
```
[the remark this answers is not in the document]
alexeldeib: actually, we can eliminate test data. see latest commits?

pkg/agent/baker_test.go
```diff
@@ -522,29 +562,39 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 				}),
 				TrustedCA: to.StringPtr(EncodedTestCert),
 			}
+		}, []outputValidator{
+			func(o *nodeBootstrappingOutput) {
+				cseMain := o.files["/opt/azure/containers/provision.sh"]
```
alexeldeib, line 567: testing with validation on content

pkg/agent/baker_test.go
```diff
@@ -142,36 +157,61 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
...
-	}, Entry("AKSUbuntu1604 with k8s version less than 1.18", "AKSUbuntu1604+K8S115", "1.15.7", nil),
-		Entry("AKSUbuntu1604 with k8s version 1.18", "AKSUbuntu1604+K8S118", "1.18.2", nil),
-		Entry("AKSUbuntu1604 with k8s version 1.17", "AKSUbuntu1604+K8S117", "1.17.7", nil),
+	}, Entry("AKSUbuntu1604 with k8s version less than 1.18", "AKSUbuntu1604+K8S115", "1.15.7", func(config *datamodel.NodeBootstrappingConfiguration) {
+		config.KubeletConfig["--dynamic-config-dir"] = "/var/lib/kubelet/"
+	}, []outputValidator{
+		func(o *nodeBootstrappingOutput) {
+			Expect(strings.Contains(o.cseCmd, "DynamicKubeletConfig")).To(BeTrue())
+			Expect(strings.Contains(o.vars["KUBELET_FLAGS"], "DynamicKubeletConfig")).To(BeTrue())
+			Expect(strings.Contains(o.vars["KUBELET_FLAGS"], "--dynamic-config-dir")).To(BeFalse())
```
alexeldeib, line 210: validate cse doesn't set dynamic config dir

pkg/agent/baker.go
```diff
@@ -95,6 +95,8 @@ func (t *TemplateGenerator) GetNodeBootstrappingCmd(config *datamodel.NodeBootst
 
 // getLinuxNodeCSECommand returns Linux node custom script extension execution command
 func (t *TemplateGenerator) getLinuxNodeCSECommand(config *datamodel.NodeBootstrappingConfiguration) string {
+	// validate and fix input
+	validateAndSetLinuxNodeBootstrappingConfiguration(config)
```
alexeldeib, line 99: this fixes the dynamic config dir issue, otherwise the new unit tests break because I added the deepcopy to avoid mutation in baker_test.go

pkg/agent/baker_test.go
```diff
@@ -142,8 +173,14 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 			configUpdator(config)
 		}
 
+		configCustomDataClone, err := deepcopy.Anything(config)
```
alexeldeib, line 176: this avoids mutation

pkg/agent/baker_test.go
```diff
@@ -1,26 +1,57 @@
...
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
 )
 
 func generateTestData() bool {
 	return os.Getenv("GENERATE_TEST_DATA") == "true"
 }
 
+const cseRegexString = `([^=\s]+)=(\"[^\"]*\"|[^\s]*)`
```
UtheMan, line 30: Mind adding a comment explaining what exactly this checks for?

UtheMan, line 30: Mind adding a comment explaining what exactly this checks for?

pkg/agent/baker_test.go
UtheMan: why do we need two deep copies of the same thing? 
configCseClone, err := deepcopy.Anything(config) and configCustomDataClone, err := deepcopy.Anything(config)

pkg/agent/baker_test.go
```diff
@@ -153,25 +190,71 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
...
+
+	}, Entry("AKSUbuntu1604 with k8s version less than 1.18", "AKSUbuntu1604+K8S115", "1.15.7", func(config *datamodel.NodeBootstrappingConfiguration) {
+		config.KubeletConfig["--dynamic-config-dir"] = "/var/lib/kubelet/"
+	}, func(o *nodeBootstrappingOutput) {
+		etcDefaultKubelet := o.files["/etc/default/kubelet"].value
+
+		Expect(o.vars["KUBELET_FLAGS"]).NotTo(BeEmpty())
+		Expect(strings.Contains(o.vars["KUBELET_FLAGS"], "DynamicKubeletConfig")).To(BeTrue())
+		Expect(strings.Contains(o.vars["KUBELET_FLAGS"], "DynamicKubeletConfig")).To(BeTrue())
+		Expect(strings.Contains(o.vars["KUBELET_FLAGS"], "--dynamic-config-dir")).To(BeFalse())
```
UtheMan, line 230: Is there some sort of list of all deprecated flags per version that we could check for? I assume similar error could happen with other flags in the future as well.

pkg/agent/baker_test.go
```diff
@@ -919,6 +1013,170 @@ func backfillCustomData(folder, customData string) {
...
+
+	for _, cseVar := range cseVariableList {
+		if len(cseVar) < 3 {
+			return nil, fmt.Errorf("expected 3 results (match, key, value) from regex, found %d, result %q", len(cseVar), cseVar)
+		}
+
+		key := cseVar[1]
+		maybeQuotededValue := cseVar[2]
+
+		if len(maybeQuotededValue) > 1 && maybeQuotededValue[0] == '"' && maybeQuotededValue[len(maybeQuotededValue)-1] == '"' {
```
UtheMan, line 1034: I'd move it to separate function with descriptive name

pkg/agent/baker_test.go
alexeldeib: yep

pkg/agent/baker_test.go
alexeldeib: to avoid mutation of the original config -- this is what happened with the dynamic config dir bug.

GetNodeBootstrappingPayload mutates the input so it's not the same as what gets passed to GetNodeBootstrappingCmd which causes bugs. agentbaker as a service uncovered this.

will comment.

UtheMan: Hmm gotcha, but why are we deep copying the same thing twice? In the end we have configCustomDataClone and configCseClone - different names, but wouldn't the result be the same thing? Both are created by deepcopy.Anything(config) using the same config variable.

alexeldeib: I couldn't decide which one to use the original thing for, so I renamed both :P

alexeldeib: I can strictly do one fewer copy but I don't mind being explicit for the test logic with a big WARNING

UtheMan: Two copies are prefectly fine, I was just confused by the naming i guess 😄

UtheMan (approved):

PR #2848, opened 2023-03-09
fix: 5 sysctls do not correctly respect defaults

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -373,3 +373,111 @@ write_files:
   owner: root
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+- path: /etc/sysctl.d/999-sysctl-aks.conf
```
alexeldeib, line 377: old code, was just diffing behavior

pkg/agent/baker.go
```diff
@@ -938,31 +938,46 @@ const sysctlTemplateString = `# This is a partial workaround to this upstream Ku
...
 {{- end}}
+{{- else }}
+net.core.somaxconn=16384
+net.ipv4.tcp_max_syn_backlog=16384
+net.ipv4.neigh.default.gc_thresh1=4096
+net.ipv4.neigh.default.gc_thresh2=8192
+net.ipv4.neigh.default.gc_thresh3=16384
+{{- end}}
+{{- else }}
+net.core.somaxconn=16384
```
alexeldeib, line 975: annoying to duplicate but not sure how to do this better

pkg/agent/baker.go
```diff
@@ -938,31 +938,46 @@ const sysctlTemplateString = `# This is a partial workaround to this upstream Ku
 net.ipv4.tcp_retries2=8
 net.core.message_burst=80
 net.core.message_cost=40
-{{- if .CustomLinuxOSConfig}}{{ if .CustomLinuxOSConfig.Sysctls}}{{ if .CustomLinuxOSConfig.Sysctls.NetCoreSomaxconn}}
```
alexeldeib, line 941: issue was 3 levels of ifs, where the inner layer is per sysctl, but the outer 2 are for all sysctl/linux config.

so if you set any custom sysctl/linux config but did NOT override those 5 values, you didn't get our defaults. 

this was a regression from January's refactor to avoid using reflection - https://github.com/Azure/AgentBaker/commit/92060d06a657f38e8750433f7cbd41309dae54c4

alexeldeib: the reflection avoided the nested ifs

UtheMan: Did this happen because we would get zero values for some of these fields, meaning that it would never end up setting defaults (cause of all the ifs)?

alexeldeib: expanding from the bad commit:
```go
{{- if .CustomLinuxOSConfig}}
  {{ if .CustomLinuxOSConfig.Sysctls}}
    {{ if .CustomLinuxOSConfig.Sysctls.NetCoreSomaxconn}}
    net.core.somaxconn={{.CustomLinuxOSConfig.Sysctls.NetCoreSomaxconn}}
    {{end}}
    // missing else
  {{end}}
   // missing else
{{- else}}
net.core.somaxconn=16384
{{- end}}
```

make sense now?

alexeldeib: why did I do that? not sure honestly...maybe we can simplify further

PR #2853, opened 2023-03-09
fix: clean cse as well as custom data for old flags

alexeldeib:
this removes e.g. dynamic-config-dir uniformly, where it was previously left in CSE but cleaned for 16 months from custom data

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -95,6 +95,8 @@ func (t *TemplateGenerator) GetNodeBootstrappingCmd(config *datamodel.NodeBootst
 
 // getLinuxNodeCSECommand returns Linux node custom script extension execution command
 func (t *TemplateGenerator) getLinuxNodeCSECommand(config *datamodel.NodeBootstrappingConfiguration) string {
+	// validate and fix input
+	validateAndSetLinuxNodeBootstrappingConfiguration(config)
 	// get parameters
 	parameters := getParameters(config, "", "")
 	// get variable
@@ -116,6 +118,8 @@ func (t *TemplateGenerator) getLinuxNodeCSECommand(config *datamodel.NodeBootstr
 
 // getWindowsNodeCSECommand returns Windows node custom script extension execution command
 func (t *TemplateGenerator) getWindowsNodeCSECommand(config *datamodel.NodeBootstrappingConfiguration) string {
+	// TODO(ace): linux cleans the input here for CSE now.
+	// should we do the same for windows?
 	// get parameters
 	parameters := getParameters(config, "", "")
 	// get variable


ganeshkumar5699 (approved):

pkg/agent/baker.go
```diff
@@ -116,6 +118,8 @@ func (t *TemplateGenerator) getLinuxNodeCSECommand(config *datamodel.NodeBootstr
 
 // getWindowsNodeCSECommand returns Windows node custom script extension execution command
 func (t *TemplateGenerator) getWindowsNodeCSECommand(config *datamodel.NodeBootstrappingConfiguration) string {
+	// TODO(ace): linux cleans the input here for CSE now.
```
alexeldeib, line 121: @AbelHu 

context: Linux moved kubelet flags from custom data to CSE.

custom data generation already calls validateAndSetLinuxNodeBootstrappingConfiguration and validateAndSetWindowsNodeBootstrappingConfiguration.

those functions remove deprecated flags like `--dynamic-config-dir` which cause kubelet to crash with 1.24+.

it also removes _docker_ flags for example.

I am not sure if you want all those flags removed, but I suspect if windows does the same refactor linux just did, you will cause an icm due to `--dynamic-confir-dir` in AKS database being passed to CSE of 1.24+ nodes.

I suggest you add the call to validateAndSetWindowsNodeBootstrappingConfiguration here as well.

I am not confident about the flag changes on windows so left it as-is, but FYI.

anujmaheshwari1 (approved):

pkg/agent/baker.go
AbelHu: @alexeldeib Windows removes all dockershim configs and `--non-masquerade-cidr` in RP side. Windows has the similar logic to call validateAndSetWindowsNodeBootstrappingConfiguration in getWindowsNodeCustomDataJSONObject.
It seems like that we also need to call validateAndSetWindowsNodeBootstrappingConfiguration in getWindowsNodeCSECommand but we need to know more details why it causes an ICM.

For Windows, KubeletConfig is used in CSE in below call stack.
$global:KubeletConfigArgs  => GetKubeletConfigKeyValsPsh => GetOrderedKubeletConfigStringForPowershell
$global:KubeletConfigArgs will be written to `c:\k\kubeclusterconfig.json` as `Kubernetes.Kubelet.ConfigArgs` in Write-KubeClusterConfig
staging/cse/windows/provisioningscripts/kubeletstart.ps1 will use `Kubernetes.Kubelet.ConfigArgs`

AbelHu:
```
func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context,
	config *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error) {
	templateGenerator := InitializeTemplateGenerator()
	nodeBootstrapping := &datamodel.NodeBootstrapping{
		CustomData: templateGenerator.GetNodeBootstrappingPayload(config),
		CSE:        templateGenerator.GetNodeBootstrappingCmd(config),
	}
```
Without the change in this PR, validateAndSetLinuxNodeBootstrappingConfiguration is called in getLinuxNodeCustomDataJSONObject which is called by GetNodeBootstrappingPayload for Linux. If `CustomData: templateGenerator.GetNodeBootstrappingPayload(config),` is always executed before `CSE:        templateGenerator.GetNodeBootstrappingCmd(config),`, the config in GetNodeBootstrappingCmd will always be updated by GetNodeBootstrappingPayload so it seems like that the issue will not occur.

https://kuree.gitbooks.io/the-go-programming-language-report/content/15/text.html

PR #2859, opened 2023-03-11
test: move e2e to go

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


ganeshkumar5699 (commented):
Excited to see AB e2e in Go! Should make it easier to add tests - thanks!

e2e/go.mod
```diff
@@ -0,0 +1,73 @@
+module github.com/Azure/agentbakere2e
```
alexeldeib, line 1: avoid polluting root module dependencies with test deps

e2e/kube.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cluster kubeconfig bytes")
+	}
+
+	restConfig, err := clientcmd.RESTConfigFromKubeConfig(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert kubeconfig bytes to rest config")
+	}
+	restConfig.NegotiatedSerializer = serializer.WithoutConversionCodecFactory{CodecFactory: scheme.Codecs}
+	restConfig.APIPath = "/api"
```
alexeldeib, line 53: I swear this is wrong but it works _shrugs_

should be /api for v1.Node,Pod, but /apis for DS. dynamic client works as expected for DS, didn't test typed client yet.

e2e/nsenter.go
```diff
@@ -0,0 +1,257 @@
...
...TBkaHg3YXpQd0ZmS1FRQUEKQU1BdklPZk8wMVVpRjdPS05xSHdaV0txU2VkVVRUa2hqd2RzdmZ5bzg0bDViQWZDVTVRRVNkbzdBOGlSZUhpUW5ZckQ5ZwpuR2IrNEJrZDI4S1E1UkxOb1lxZDgrK2xhQWUyNE1SVFFPc3psUS9LK3RRWWdWazBOaTRjeFhsWFc1TzVyTEFTbEw3THQ5CmorbzJiQkluSWs2TThZYStnRFJrWWtmYXhIWENoK0FxOUxXNEF4bHhJZEc2SzRQOXlYcUhGakNyMVRvNGk2NUtHUkQ1Wm4KNDFVR1paVUpWakZBL0E3SUM2KzVlSDZQSkE4MUk3VER2Tmt0WGQ0VGhQRTd5S2lCMEFBQURCQU9VWGU2b2JGWk13NjFMUgozVE84NFd4Q09TUjgvamNIRDNxSHRKdFdMakkvYk53MGthQVhOMUpsdHFiTElBL08ydFdPY2ZEeEpFYU1sQjRUcmdZV0FUCnI1VkZFV1hZS0MySGxSVGEva2c3QS9HUk9Od1hIS2V5dXlscFVNN0N6MTdMa2k5ZTF5QkVOYTRGQmg4V3c3VGtjaGtrVDcKc3dsbEh3QlI3dGlTWlBEZXl2UzZqZUo3OWxwNlFGMTc3NTZmdHg1RjAxbDZVV3hUYmlPSndiU3FxZUxQS1E5SldnRWRTQwp2MjNzcTNQV1F2Q1ZiSWtkOS9YK1piWVFCTzhOUWYvUUFBQU1FQXlsU2lvclF4dDlQY2ljTy9taXJjdlpzTC9tdldORy95ClY1L1BoRml3WTJ2c0d1RUFSYkxFWEQvSkVDck5rM3VZS0VZRnFKZXRIbTl6ajdlcWlJSHFWemtHaU13ZE9zNXZQM25sREsKK05PWjFiNG9rVzgwdjRJWWduZ2pqN1B5a2kreUxMQzlnZS9lQ3NhM1pvNDNMUTNoQXNhcFZ6N3YxUnYvMFAzcGVteW8zdgo5K0lUUlVTdnpuaTd2bGIxK0JIMzBMdTNOWm9pWmZ4K0JpMHJFU1lBN29qaXFjK21SZG5BT3Q3UWw0QlJ4N1BZRWkyWkdKCk8xZ0JLWnY2OFRzQ294QUFBQUlYSnZiM1JBVldKMWJuUjFMVEl3TURRdFptOWpZV3d0TmpRdGJXbHVhVzFoYkFFPQotLS0tLUVORCBPUEVOU1NIIFBSSVZBVEUgS0VZLS0tLS0K`
```
alexeldeib, line 36: TODO: remove this and use the generated key again

log the generated key somewhere?

e2e/nsenter.go
```diff
@@ -0,0 +1,257 @@
...
+	// 	"0",
+	// 	"networkInterfaces",
+	// 	"2018-10-01",
+	// 	nil,
+	// )
+	if err != nil {
+		return nil, err
+	}
+
+	respBytes, err := io.ReadAll(resp.Body)
```
alexeldeib, line 70: close(resp.Body)?

e2e/nsenter.go
```diff
@@ -0,0 +1,257 @@
...
+
+	if err := json.Unmarshal(respBytes, &instanceNICResult); err != nil {
+		return nil, err
+	}
+
+	privateIP := instanceNICResult.Value[0].Properties.IPConfigurations[0].Properties.PrivateIPAddress
+
+	sshCommand := fmt.Sprintf(sshCommandTemplate, notAnSshKey, privateIP)
+
+	commandList := map[string]string{
```
alexeldeib, line 85: this is how to extract arbitrary files

e2e/vmss.go
```diff
@@ -0,0 +1,146 @@
...
+		return fmt.Errorf("failed to validate private key: %q", err)
+	}
+
+	publicRsaKey, err := ssh.NewPublicKey(&privateKey.PublicKey)
+	if err != nil {
+		return fmt.Errorf("failed to convert private to public key: %q", err)
+	}
+
+	pubKeyBytes := ssh.MarshalAuthorizedKey(publicRsaKey)
+	_ = pubKeyBytes
```
alexeldeib, line 39: @anujmaheshwari1

e2e/cloud.go
```diff
@@ -0,0 +1,113 @@
...
+	coreClient     *azcore.Client
+	vmssClient     *armcompute.VirtualMachineScaleSetsClient
+	vmssVMClient   *armcompute.VirtualMachineScaleSetVMsClient
+	resourceClient *armresources.Client
+	aksClient      *armcontainerservice.ManagedClustersClient
+}
+
+func newAzureClient(subscription string) (*azureClient, error) {
+	httpClient := &http.Client{
+		// For Now using the defaults recommended by Track 2
```
ganeshkumar5699, line 31: Should we add a link to the doc that mentions the recommended defaults? In case the defaults change

e2e/cloud.go
```diff
@@ -0,0 +1,113 @@
...
+	credential, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create credential: %q", err)
+	}
+
+	plOpts := runtime.PipelineOptions{}
+	clOpts := &azcore.ClientOptions{
+		Transport: httpClient,
+		PerCallPolicies: []policy.Policy{
+			runtime.NewBearerTokenPolicy(credential, []string{"https://management.azure.com/.default"}, nil),
```
ganeshkumar5699, line 74: nit: better to move this hardcoded string to a variable?

e2e/cloud.go
```diff
@@ -0,0 +1,113 @@
...
+			logger,
+		},
+	}
+
+	coreClient, err := azcore.NewClient("agentbakere2e.e2e_test", "v0.0.0", plOpts, clOpts)
```
ganeshkumar5699, line 79: When would this version need to be changed?

e2e/suite_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"github.com/barkimedes/go-deepcopy"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var cases = map[string]scenarioConfig{
+	"base": {},
+	// "gpu": {
```
ganeshkumar5699, line 23: Do we want to remove this commented code/move it to a separate gpu case (without using it)?

e2e/template_test.go
```diff
@@ -0,0 +1,501 @@
+package e2e_test
+
+import "github.com/Azure/agentbaker/pkg/agent/datamodel"
+
+// this is huge, but accurate, so leave it here
+// TODO(ace): minimize the actual required defaults.
+// this is what we previously used for bash e2e from e2e/nodebootstrapping_template.json
+// which itself was extracted from baker_test.go logic, which was inherited from aks-engine.
+func baseTemplate() *datamodel.NodeBootstrappingConfiguration {
```
ganeshkumar5699, line 9: How do we plan on testing with different NBCs: 1) mutate the result NBC of baseTemplate() when necessary, 2) create another function with a new NBC from scratch (will be quite large then) or some other way?

ganeshkumar5699: If we go with option 1, then we need to make sure that other callers/dependencies of NBC all use the mutated NBC and not the base one by mistake.

e2e/vmss.go
```diff
@@ -0,0 +1,160 @@
...
+										KeyData: to.Ptr(string(pubKeyBytes)),
+										Path:    to.Ptr("/home/azureuser/.ssh/authorized_keys"),
+									},
+								},
+							},
+						},
+					},
+					StorageProfile: &armcompute.VirtualMachineScaleSetStorageProfile{
+						ImageReference: &armcompute.ImageReference{
+							ID: to.Ptr("/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1677169694.31375"),
```
ganeshkumar5699, line 105: nit: Should we move this (and similar long hardcoded strings) to variables? I can also move it in a different PR for testing e2e with latest VHD

e2e/suite_test.go
```diff
@@ -0,0 +1,181 @@
+package e2e_test
```
ganeshkumar5699, line 1: future readability suggestion: would moving the specific test cases (and the suite run) to a subfolder in e2e/ make it easier to read? And having one test per file? since we plan on adding more test cases

e2e/suite_test.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	baseConfig, err := getBaseBootstrappingConfig(ctx, t, cloud, suiteConfig, clusterParams)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for name, tc := range cases {
```
ganeshkumar5699, line 66: Looks like this is done serially? Can we parallelize the test case runs?

AbelHu:
cc @junjiezhang1997

e2e/suite_test.go
alexeldeib: t.Run actually already runs parallel in this setup I believe

e2e/vmss.go
```diff
@@ -0,0 +1,165 @@
...
+									KeyData: to.Ptr(sshPublicKey),
+									Path:    to.Ptr("/home/azureuser/.ssh/authorized_keys"),
+								},
+							},
+						},
+					},
+				},
+				StorageProfile: &armcompute.VirtualMachineScaleSetStorageProfile{
+					ImageReference: &armcompute.ImageReference{
+						ID: to.Ptr("/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1677169694.31375"),
```
cameronmeissner, line 124: I guess we're leaving out the parameterized SIG_VERSION_ID for now, should be fine to add in after initial merge

e2e/nsenter.go
```diff
@@ -0,0 +1,247 @@
...
+		"bash",
+		"-c",
+	}
+}
+
+const sshCommandTemplate = `echo '%s' > sshkey && chmod 0600 sshkey && ssh -i sshkey -o PasswordAuthentication=no -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ConnectTimeout=5 azureuser@%s sudo`
+
+func extractLogsFromFailedVM(ctx context.Context, t *testing.T, cloud *azureClient, kube *kubeclient, subscription, resourceGroupName, clusterName, vmssName, sshPrivateKey string) (map[string]string, error) {
+	pl := cloud.coreClient.Pipeline()
+	url := fmt.Sprintf("https://management.azure.com/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/virtualMachineScaleSets/%s/virtualMachines/%d/networkInterfaces?api-version=2018-10-01",
```
cameronmeissner, line 40: maybe refactor this string out to another template variable like sshCommandTemplate?

e2e/template_test.go
alexeldeib: this is in fact done

I think we can also reduce the base template size, it has many defaults that aren't needed

e2e/suite_test.go
alexeldeib: I think we can refactor most of the code still yes :D

e2e/suite_test.go
alexeldeib: added t.Parallel to make it work

e2e/cloud.go
alexeldeib: I don't actually know where these are documented but they're relatively sane compared to many stdlib defaults

avoid slow loris, avoid indefinite timeouts, etc

e2e/suite_test.go
alexeldeib: uncommented

e2e/vmss.go
alexeldeib: this can be done in code by mutating vmss directly now, but not sure otherwise best approach yeah

e2e/cloud.go
alexeldeib: will comment

e2e/nsenter.go
```diff
@@ -0,0 +1,285 @@
...
+		t.Log(stderr.String())
+
+		if err != nil {
+			return nil, err
+		}
+
+		result[file] = stdout.String()
+	}
+
+	litter.Dump(result)
```
alexeldeib, line 130: could probably do something more clever here but pretty good for now

e2e/nsenter.go
```diff
@@ -0,0 +1,285 @@
...
+
+	podName := podList.Items[0].ObjectMeta.Name
+
+	var result = map[string]string{}
+	for file, sourceCmd := range commandList {
+		cmd := append(nsenterCommandArray(), sourceCmd)
+
+		req := kube.typed.CoreV1().RESTClient().Post().Resource("pods").Name(podName).Namespace("default").SubResource("exec")
+
+		option := &corev1.PodExecOptions{
```
alexeldeib, line 159: TODO: extract this to a helper

e2e/vmss.go
```diff
@@ -0,0 +1,165 @@
...
+	"encoding/pem"
+	"fmt"
+	mrand "math/rand"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"golang.org/x/crypto/ssh"
+)
+
+func createVMSSWithPayload(ctx context.Context, r *mrand.Rand, cloud *azureClient, location, resourceGroupName, name, subnetID string, customData, cseCmd string, mutator func(*armcompute.VirtualMachineScaleSet)) (sshPrivateKey []byte, e error) {
```
cameronmeissner, line 17: I might be blind, but is 'r' ever used in here?

cameronmeissner (approved):

e2e/cases.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"testing"
+
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+)
+
+var cases = map[string]scenarioConfig{
```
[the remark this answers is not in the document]
alexeldeib: they're just normal Go test functions so the thinking was not to use one big table but provide good helpers to make it easy to write one off tests

open to whatever folks want/can hack up, show me ideas :)

e2e/cluster.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+)
+
+func setupCluster(ctx context.Context, cloud *azureClient, location, resourceGroupName, clusterName string) error {
+	aksCluster, err := cloud.aksClient.Get(ctx, resourceGroupName, clusterName, nil)
+	if err != nil {
+		return fmt.Errorf("failed to get aks cluster: %q", err)
+	}
+
+	rgExistence, err := cloud.resourceGroupClient.CheckExistence(ctx, resourceGroupName, nil)
```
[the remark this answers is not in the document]
alexeldeib: it's the MC resource group so it's expected to exist only if the AKS cluster also exists

e2e/cluster.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get MC RG: %q", err)
+	}
+
+	if !rgExistence.Success || aksCluster.Properties == nil || aksCluster.Properties.ProvisioningState == nil || *aksCluster.Properties.ProvisioningState == "Failed" {
```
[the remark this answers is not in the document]
alexeldeib: that's exactly what this is -- I think you're looking at the wrong RG name? or did I mess it up?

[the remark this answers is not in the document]
alexeldeib: I did mess it up

e2e/cluster.go
```diff
@@ -0,0 +1,93 @@
...
+	for pager.More() {
+		nextResult, err := pager.NextPage(ctx)
+		if err != nil {
+			return "", fmt.Errorf("failed to advance page: %q", err)
+		}
+		for _, v := range nextResult.Value {
+			if v == nil {
+				return "", fmt.Errorf("aks vnet id was empty")
+			}
+			return fmt.Sprintf("%s/subnets/%s", *v.ID, "aks-subnet"), nil
```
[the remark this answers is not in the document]
alexeldeib: this is extracting the AKS created VNET for a single cluster, the subnet name is typically aks-subnet by default

it'll work fine for multiple clusters when we pass in new RG/cluster name instead of constants

e2e/kube.go
```diff
@@ -0,0 +1,106 @@
...
+						return true, nil
+					}
+				}
+			}
+		}
+
+		return false, nil
+	})
+
+	if err == nil {
```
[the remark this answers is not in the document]
alexeldeib: hah, just caught this in e2e

e2e/nsenter.go
```diff
@@ -0,0 +1,287 @@
...
+	if err := json.Unmarshal(respBytes, &instanceNICResult); err != nil {
+		return nil, err
+	}
+
+	privateIP := instanceNICResult.Value[0].Properties.IPConfigurations[0].Properties.PrivateIPAddress
+
+	sshCommand := fmt.Sprintf(sshCommandTemplate, sshPrivateKey, privateIP)
+
+	commandList := map[string]string{
+		"/var/log/azure/cluster-provision/log": "cat /var/log/azure/cluster-provision.log",
```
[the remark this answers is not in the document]
alexeldeib: hah, yes.

e2e/vmss.go
alexeldeib: fixed -- using the global func without a seed is deprecated.

e2e/nsenter.go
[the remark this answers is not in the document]
alexeldeib: fixed

e2e/cluster.go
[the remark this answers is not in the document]
alexeldeib: fixed to use const

e2e/e2e-scenario.sh
```diff
@@ -46,7 +46,9 @@ kubectl rollout status deploy/debug
 echo "Scenario is $SCENARIO_NAME"
 jq -s '.[0] * .[1]' nodebootstrapping_config.json scenarios/$SCENARIO_NAME/property-$SCENARIO_NAME.json > scenarios/$SCENARIO_NAME/nbc-$SCENARIO_NAME.json
 
-go test -run TestE2EBasic
+go test -tags=bash_e2e -run TestE2EBasic
+
+exit
```
alexeldeib, line 51: oops

e2e/nsenter.go
```diff
@@ -0,0 +1,290 @@
...
+
+	if len(podList.Items) < 1 {
+		return nil, fmt.Errorf("failed to find debug pod, list by selector returned no results")
+	}
+
+	podName := podList.Items[0].ObjectMeta.Name
+
+	var result = map[string]string{}
+	for file, sourceCmd := range commandList {
+		mergedCmd := fmt.Sprintf("%s %s", sshCommand, sourceCmd)
```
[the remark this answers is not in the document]
alexeldeib: good point. trying now!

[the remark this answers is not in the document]
cameronmeissner: I guess I'm confused why we need the SSH command at all, if we can just exec into the pod and nsenter in order to get the files - it seems like right now we just create the debug pod, exec into it, and SSH into the target VM from there since it's within the same private network as the VM

why couldn't we just add a label selector to the debug pod to always guarantee it spawns on the target VM -> exec -> nsenter -> cat, eliminating SSH altogether

juan-lee (approved):

e2e/nsenter.go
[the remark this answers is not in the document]
anujmaheshwari1: I think this was in case the node didnt actually join the api server for some reason or cse failed to execute. I think.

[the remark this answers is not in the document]
alexeldeib: there are two cases for extraction:
- to get the parameters from an existing (healthy) node - easiest way is nsenter via pod, no ssh
- to extract logs from a failed node - requires ssh or similar, can't be done with nsenter
  - however, nodes only have private IPs, so easiest way is to jumpbox an existing node via a pod

you don't need to nsenter and then ssh for the 2nd case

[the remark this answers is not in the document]
alexeldeib: openssh isn't installed by the container by default so this doesn't work

alexeldeib:
https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=70225182&view=results passed e2e

ganeshkumar5699 (approved):

PR #2862, opened 2023-03-13
fix: Wrong echo of Hyper-v generation in the Windows VHD build pipeline

wanqingfu:
**What type of PR is this?**
fix the wrong echo output of windowsvhdmode
/kind bug

`HYPERV_GENRATION` was incorrectly used in `packer.mk`. Output was always `Building with Hyper-v generation 2 VM and save to .....`

Should use `HYPERV_GENERATION`.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

It does not affect the packer build. It's just an echo output typo.

**Release note**:
```
none
```

### DIFF
--- a/packer.mk
+++ b/packer.mk
@@ -66,13 +66,13 @@ endif
 build-packer-windows:
 ifeq (${MODE},windowsVhdMode)
 ifeq (${SIG_FOR_PRODUCTION},True)
-ifeq (${HYPERV_GENRATION},V1)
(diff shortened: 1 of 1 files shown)


ShiqianTao (approved):

ShiqianTao:
Are PR's labels correct?

AbelHu (approved):

PR #2867, opened 2023-03-13
feat: use cluster subnets for dual-stack

tyler-lloyd:
**What type of PR is this?**

/kind feature

Updates the exception list for the CNI conflist to include all CIDRs
in kubeclustercidr. It already handles "," separated string by default
but only adds the first one currently. This will add both so the IPv4
and IPv6 addresses are in the exception list.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/parts/windows/kuberneteswindowssetup.ps1
+++ b/parts/windows/kuberneteswindowssetup.ps1
@@ -117,6 +117,7 @@ $global:PrimaryAvailabilitySetName = "{{GetVariable "primaryAvailabilitySetName"
 $global:PrimaryScaleSetName = "{{GetVariable "primaryScaleSetName"}}"
 
 $global:KubeClusterCIDR = "{{GetParameter "kubeClusterCidr"}}"
+$global:KubeClusterCIDRs = "{{GetParameter "kubeClusterCidrs"}}"
 $global:KubeServiceCIDR = "{{GetParameter "kubeServiceCidr"}}"
 $global:VNetCIDR = "{{GetParameter "vnetCidr"}}"
 {{if IsKubernetesVersionGe "1.16.0"}}
@@ -407,6 +408,7 @@ try
     Set-AzureCNIConfig -AzureCNIConfDir $global:AzureCNIConfDir `
         -KubeDnsSearchPath $global:KubeDnsSearchPath `
         -KubeClusterCIDR $global:KubeClusterCIDR `
+        -KubeClusterCIDRs $global:KubeClusterCIDRs `
         -KubeServiceCIDR $global:KubeServiceCIDR `
         -VNetCIDR $global:VNetCIDR `
         -IsDualStackEnabled $global:IsDualStackEnabled `
--- a/parts/windows/kuberneteswindowssetup.ps1
+++ b/parts/windows/kuberneteswindowssetup.ps1
@@ -117,7 +117,6 @@ $global:PrimaryAvailabilitySetName = "{{GetVariable "primaryAvailabilitySetName"
 $global:PrimaryScaleSetName = "{{GetVariable "primaryScaleSetName"}}"
 
 $global:KubeClusterCIDR = "{{GetParameter "kubeClusterCidr"}}"
-$global:KubeClusterCIDRs = "{{GetParameter "kubeClusterCidrs"}}"
 $global:KubeServiceCIDR = "{{GetParameter "kubeServiceCidr"}}"
 $global:VNetCIDR = "{{GetParameter "vnetCidr"}}"
 {{if IsKubernetesVersionGe "1.16.0"}}
@@ -408,7 +407,6 @@ try
     Set-AzureCNIConfig -AzureCNIConfDir $global:AzureCNIConfDir `
         -KubeDnsSearchPath $global:KubeDnsSearchPath `
         -KubeClusterCIDR $global:KubeClusterCIDR `
-        -KubeClusterCIDRs $global:KubeClusterCIDRs `
         -KubeServiceCIDR $global:KubeServiceCIDR `
         -VNetCIDR $global:VNetCIDR `
         -IsDualStackEnabled $global:IsDualStackEnabled `
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -84,9 +84,9 @@ function Set-AzureCNIConfig
         # Fill in DNS information for kubernetes.
         $exceptionAddresses = @()
         if ($IsDualStackEnabled) {
-            $podCIDRs = $KubeClusterCIDR -split ","
-            foreach ($cidr in $podCIDRs) {
-                $exceptionAddresses += $cidr
+            $subnetsToPass = $KubeClusterCIDR -split ","
+            foreach ($subnet in $subnetsToPass) {
+                $exceptionAddresses += $subnet
             }
         } else {
             $exceptionAddresses += $KubeClusterCIDR
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -83,9 +83,11 @@ function Set-AzureCNIConfig
     } else {
         # Fill in DNS information for kubernetes.
         $exceptionAddresses = @()
-        if ($IsDualStackEnabled){
-            $subnetToPass = $KubeClusterCIDR -split ","
-            $exceptionAddresses += $subnetToPass[0]
+        if ($IsDualStackEnabled) {
+            $podCIDRs = $KubeClusterCIDR -split ","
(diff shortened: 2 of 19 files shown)


staging/cse/windows/azurecnifunc.tests.suites/AzureCNI.Default.OverlayDualStack.conflist
```diff
@@ -0,0 +1,52 @@
...
+      "mode": "bridge",
+      "bridge": "azure0",
+      "executionMode": "v4swift",
+      "capabilities": {
+        "portMappings": true,
+        "dns": true
+      },
+      "ipam": {
+        "type": "azure-cns",
+        "mode": "v4overlay"
```
ashvindeodhar, line 17: for dualstack, the ipam mode should be different - https://github.com/Azure/azure-container-networking/pull/1772/files

tyler-lloyd: fixed

tyler-lloyd: @ashvindeodhar is the conflist in the pr you linked correct? it has `"executionMode": "v4swift",` and then no IPv6 addresses for nameservers or endpoint policies. is that expected?

staging/cse/windows/azurecnifunc.tests.suites/AzureCNI.Default.OverlayDualStack.conflist
```diff
@@ -0,0 +1,52 @@
+{
+  "cniVersion": "0.3.0",
+  "name": "azure",
+  "adapterName": "",
+  "plugins": [
+    {
+      "type": "azure-vnet",
+      "mode": "bridge",
+      "bridge": "azure0",
+      "executionMode": "v4swift",
```
ashvindeodhar, line 10: checking on this

tyler-lloyd: it would still be good to know but we don't update this value in AB so for testing purposes it isn't a big concern. We're more concerned with making sure the right exception list written

staging/cse/windows/azurecnifunc.tests.suites/AzureCNI.Expect.OverlayDualStack.conflist
```diff
@@ -0,0 +1,76 @@
...
+          }
+        },
+        {
+          "Name": "EndpointPolicy",
+          "Value": {
+            "Type": "ACL",
+            "Protocols": "6",
+            "Action": "Block",
+            "Direction": "Out",
+            "RemoteAddresses": "168.63.129.16/32",
```
AbelHu, line 46: Does wireserver work with ipv6? We may need to block the ipv6 address of the wireserver if it works.

staging/cse/windows/azurecnifunc.tests.suites/AzureCNI.Default.OverlayDualStack.conflist
```diff
@@ -0,0 +1,47 @@
...
+        "portMappings": true,
+        "dns": true
+      },
+      "ipam": {
+        "type": "azure-cns",
+        "mode": "dualStackOverlay"
+      },
+      "dns": {
+        "Nameservers": [
+          "10.0.0.10",
```
AbelHu, line 21: Q: Do we need to update dns servers for ipv6?

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -117,6 +117,7 @@ $global:PrimaryAvailabilitySetName = "{{GetVariable "primaryAvailabilitySetName"
 $global:PrimaryScaleSetName = "{{GetVariable "primaryScaleSetName"}}"
 
 $global:KubeClusterCIDR = "{{GetParameter "kubeClusterCidr"}}"
+$global:KubeClusterCIDRs = "{{GetParameter "kubeClusterCidrs"}}"
```
AbelHu, line 120: Could we reuse `KubeClusterCIDR` since it supports multiple cidrs with `,` by design?
$exceptionAddresses += $subnetToPass[0] was added in https://github.com/Azure/AgentBaker/commit/fa17dfd271a7e8c060be70d5c214b35ca067c33c. So you only need to update `$exceptionAddresses += $subnetToPass[0]` to 
```
foreach ($cidr in $subnetToPass) {
                $exceptionAddresses += $cidr
            }
```

AbelHu: With this change, it will block the current users even without new cse script package.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -407,6 +408,7 @@ try
     Set-AzureCNIConfig -AzureCNIConfDir $global:AzureCNIConfDir `
         -KubeDnsSearchPath $global:KubeDnsSearchPath `
         -KubeClusterCIDR $global:KubeClusterCIDR `
+        -KubeClusterCIDRs $global:KubeClusterCIDRs `
```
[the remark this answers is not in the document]
tyler-lloyd: yeah let's simplify this and just re-use kubeClusterCidr. RP can pass the cluster cidrs as a comma separated string for dual-stack which seems like it was the original intent here.

parts/windows/kuberneteswindowssetup.ps1
tyler-lloyd: yeah, latest update just re-uses kubeClusterCidr

staging/cse/windows/azurecnifunc.tests.suites/AzureCNI.Default.OverlayDualStack.conflist
tyler-lloyd: I don't think right now because we don't support accepting a kube dns service IP from users in IPv6 format. Even in dual-stack clusters we only accept the v4 one.

staging/cse/windows/azurecnifunc.tests.suites/AzureCNI.Expect.OverlayDualStack.conflist
ashvindeodhar: there is no ipv6 IP for wireserver

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -83,9 +83,11 @@ function Set-AzureCNIConfig
     } else {
         # Fill in DNS information for kubernetes.
         $exceptionAddresses = @()
-        if ($IsDualStackEnabled){
-            $subnetToPass = $KubeClusterCIDR -split ","
-            $exceptionAddresses += $subnetToPass[0]
+        if ($IsDualStackEnabled) {
+            $subnetsToPass = $KubeClusterCIDR -split ","
```
AbelHu, line 87: will current rp pass multiple values in KubeClusterCIDR and this change will block anything?

tyler-lloyd: currently rp doesn't pass multiple values in kubeClusterCIDR. there is a separate PR for that. this shouldn't have any affect on any current clusters as IsDualStack shouldn't be enabled for any prod cluster.

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -83,9 +83,11 @@ function Set-AzureCNIConfig
     } else {
         # Fill in DNS information for kubernetes.
         $exceptionAddresses = @()
-        if ($IsDualStackEnabled){
-            $subnetToPass = $KubeClusterCIDR -split ","
-            $exceptionAddresses += $subnetToPass[0]
+        if ($IsDualStackEnabled) {
```
AbelHu, line 86: Please test this change with a private cse package and update the test result.

tyler-lloyd: ran a standalone test and verified:

1. the cluster cidr is passed in as comma sep list
  ```
  "Network": {
            "ClusterCidr": "10.240.0.0/16,fd81:6b3c:7bb7:5fbc::/64",
            "ServiceCidr": "10.0.0.0/16",
            "DnsIp": "10.0.0.10"
        }
  ```
2. the conflist contains the right exception list
```
"AdditionalArgs": [
                {
                    "Name": "EndpointPolicy",
                    "Value": {
                        "Type": "OutBoundNAT",
                        "ExceptionList": [
                            "10.240.0.0/16",
                            "fd81:6b3c:7bb7:5fbc::/64"
                        ]
                    }
                },
```

AbelHu (approved):
LGTM

PR #2893, opened 2023-03-17
upload VM logs and cluster parameter files during abe2e go, some slight refactoring

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/nsenter.go
```diff
@@ -84,47 +83,22 @@ func extractLogsFromVM(ctx context.Context, t *testing.T, cloud *azureClient, ku
...
 
-		var stdout, stderr bytes.Buffer
-
-		err = exec.StreamWithContext(ctx, remotecommand.StreamOptions{
-			Stdout: &stdout,
-			Stderr: &stderr,
-		})
-
-		t.Log("stdout")
+		t.Log("----------------------------------- begin stdout -----------------------------------")
```
alexeldeib, line 91: should we even bother doing this? it was a stop gap mostly

e2e/nsenter.go
```diff
@@ -149,32 +123,7 @@ func extractClusterParameters(ctx context.Context, t *testing.T, kube *kubeclien
...
-		}
-
-		var stdout, stderr bytes.Buffer
-
-		err = exec.StreamWithContext(ctx, remotecommand.StreamOptions{
-			Stdout: &stdout,
-			Stderr: &stderr,
-		})
-
+		stdout, _, err := execOnPod(ctx, kube, defaultNamespace, podName, cmd)
```
alexeldeib, line 126: I think you may want stderr if the executed command fails inside the pod?

e2e/nsenter.go
cameronmeissner: figured it would be nice to have the output directly in the pipeline logs, though can remove if we want to everything from the uploaded logs

e2e/suite_test.go
```diff
@@ -51,12 +51,31 @@ func Test_All(t *testing.T) {
...
+
+	dumpClusterParams := func() {
+		t.Logf("dumping cluster parameters to local directory: %s", clusterParamsDir)
+		if err := dumpFileMapToDir(clusterParamsDir, clusterParams); err != nil {
+			t.Log("error dumping cluster parameters")
+			t.Error(err)
+		}
+	}
+
+	defer dumpClusterParams()
```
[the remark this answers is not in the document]
cameronmeissner: figured we would do all log dumping + cleanup on the deferred call stack, can do it here instead tho

e2e/nsenter.go
cameronmeissner: actually yeah I think it makes more sense to only output stderr here

e2e/nsenter.go
```diff
@@ -85,46 +84,17 @@ func extractLogsFromVM(ctx context.Context, t *testing.T, cloud *azureClient, ku
...
-		t.Log("stderr")
-		t.Log(stderr.String())
+		t.Logf("executing command on pod %s/%s: %q", defaultNamespace, podName, strings.Join(cmd, " "))
 
+		stdout, stderr, err := execOnPod(ctx, kube, defaultNamespace, podName, cmd)
 		if err != nil {
 			return nil, err
 		}
 
+		checkStdErr(stderr, t)
```
alexeldeib, line 94: this still isn't quite right...on L91 you'll return before logging the error if err != nil (for example, exit 1) and stderr may have contents (script failure output)

alexeldeib: could either log by default or bubble to caller if err != nil somehow

e2e/suite_test.go
```diff
@@ -51,12 +51,27 @@ func Test_All(t *testing.T) {
...
+		t.Error(err)
+	}
+
 	baseConfig, err := getBaseBootstrappingConfig(ctx, t, cloud, suiteConfig, clusterParams)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	for name, tc := range cases {
+		caseLogsDir, err := createVMLogsDir(name)
```
alexeldeib, line 70: this is outside t.Run() so it fails the whole suite, not a single case, probably should be per case

e2e/nsenter.go
cameronmeissner: herp

e2e/suite_test.go
cameronmeissner: derp

e2e/suite_test.go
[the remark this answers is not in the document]
alexeldeib: I was mostly using deferred for things we always want to run, only if the corresponding "dirty" task succeded. 

I wasn't sure if we even need this uploaded but ok either way

alexeldeib (approved):
lgtm!

I think we can probably still refactor some of these helpers a bit from the original PR for clarity/organization, but can follow up as we add scenarios to move parameters around

juan-lee (approved):

PR #2896, opened 2023-03-20
refactor: abe2e go scenario config

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/scenario/gpu.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+)
+
+// Returns config for the 'gpu' E2E scenario
+func gpu() *Scenario {
+	return &Scenario{
+		Name:        "gpu",
+		Description: "Tests that a GPU-enabled SKU can be properly bootstrapped",
+		ScenarioConfig: ScenarioConfig{
+			BootstrapConfigMutator: func(t *testing.T, nbc *datamodel.NodeBootstrappingConfiguration) {
```
UtheMan, line 17: Thoughts on using functional options instead of BootstrapConfigMutator to set values in NodeBootstrappingConfiguration? I think it could be a bit better for readability, the drawback is we'd need to add many many functional options for different parameters (but I feel like it should be possible to generate them all automatically? I never attempted that though :D). Something along the lines of [this](https://gist.github.com/UtheMan/04039c7edcbd29f899113a9ceab7473f). I feel like it would make the tests a bit more readable since option names would explain what's happening straightaway. Drawback is obviously bloat. 

I might be missing some context on why we wanted to use a function like BootstrapConfigMutator though.

cameronmeissner: yeah, I think this is something that we previously talked about moving towards as well - I like this approach since it allows to compose a lot of potentially different things together to move towards desired state for a given scenario

cameronmeissner: I also don't think we necessarily need to pre-define every permutation of modification to nbc, that's more thing we can do on the fly when actually defining the given scenario - though it'll become apparent what common stuff can be moved into predefined mutators

cameronmeissner:
reimplemented in https://github.com/Azure/AgentBaker/pull/2982

PR #2898, opened 2023-03-21
fix: use a fixed blobfuse2 version to avoid unexpected version change

andyzhangx:
**What type of PR is this?**
/kind bug

fix: use a fixed blobfuse2 version to avoid unexpected version change
This PR does not change current blobfuse2 version in AgentBaker image since latest version is already v2.0.2, so it has no impact currently, this PR would use a fixed version to avoid unexpected version change in the future

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
fix: use a fixed blobfuse2 version to avoid unexpected version change
```


cvvz (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -38,7 +39,7 @@ installDeps() {
         # blobfuse2 is installed for all ubuntu versions, it is included in pkg_list
         # for 22.04, fuse3 is installed. for all others, fuse is installed
         # for 16.04, installed blobfuse1.3.7, for all others except 22.04, installed blobfuse1.4.5
-        pkg_list+=(blobfuse2)
+        pkg_list+=(blobfuse2=${BLOBFUSE2_VERSION})
         if [[ "${OSVERSION}" == "22.04" ]]; then
             pkg_list+=(fuse3)
```
AbelHu, line 44: should we specify the version for fuse3?

andyzhangx: it's not necessary to specify fuse2 since that's a kernel driver which is quite stable, we only want to make azure blobfuse2 driver as fixed since it would upgrade frequently. @AbelHu

AbelHu: I think that fixed version should be better since we have validated it in our test.

AbelHu: I think that that is why you plan to fix the version for fuse3

andyzhangx: these are Ubuntu packages, we usually don't specify a version for every package, e.g.
https://github.com/Azure/AgentBaker/blob/ccb15a86a1976e504d46e4eb58c2f3daf4c09d59/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh#L26

we don't want to specify a version for every package above.

AbelHu: This PR LGTM. It seems like that this is a generic issue. @alexeldeib should we create a repair item to fix the version for all these packages?

AbelHu (approved):
Suggest to fix other versions too for the same purpose "avoid unexpected version change"

PR #2904, opened 2023-03-21
drop unused images from vhd

haitch:
**What type of PR is this?**
/kind cleanup

- RunCommand image
we don't need runCommand image in VHD, it's a runtime feature, we only need the image upon user request.
missing this image would only block runCommand API request, have no impact on ControlPlane/Node health.
we can save 200MB in the VHD size.

- Kubernetes Dashboard image
Kubernetes dashboard image has no longer been used since 1.18, so drop them.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/components.json
```diff
@@ -1,15 +1,5 @@
...
-      "downloadURL": "mcr.microsoft.com/oss/kubernetes/kubernetes-dashboard:*",
-      "amd64OnlyVersions": [],
-      "multiArchVersions": []
-    },
-    {
-      "downloadURL": "mcr.microsoft.com/oss/kubernetes/dashboard:*",
-      "amd64OnlyVersions": [],
-      "multiArchVersions": []
-    },
     {
```
robbiezhang, line 3: you can remove this as well. It's part of the kube-dashboard

chmill-zz:
can we remove the 1.22.* kube-proxy images while we're here. https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/kube-proxy-images.json

haitch:
> can we remove the 1.22.* kube-proxy images while we're here. https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/kube-proxy-images.json  

I don't understand the impact in this case.

alexeldeib:
> I don't understand the impact in this case.

no longer supported, so easy size win

vhdbuilder/packer/components.json
sozercan: +1 on metrics-scaper

sozercan (approved):

alexeldeib (approved):

robbiezhang (approved):

PR #2919, opened 2023-03-23
fix cluster recreation logic

### LINKED ISSUE #2916: AgentBaker E2E pipeline fails to create new cluster and reports "suite_test.go:32: failed to get aks cluster"
**What happened**:
Since the resource group of AgentBaker E2E pipeline is created on a test sub, it will be deleted after a certain time (3 days). After the resource group is deleted, current AgentBaker E2E pipeline fails to create a new cluster and reports "suite_test.go:32: failed to get aks cluster".

**What you expected to happen**:
AgentBaker E2E pipeline recreates a new resource group and cluster, then runs test successfully.

**How to reproduce it**:
Delete the resource manually and run AgentBaker E2E pipeline.

It's a bit strange that I noticed a new cluster was created yesterday by an old branch with bash version of AgentBaker E2E pipeline https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=70508837&view=results. But the cluster was deleted in 24 hours. Not sure whether there is any new change for cleaning resource automatically.

**Environment**:
- AgentBaker version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

cameronmeissner:
**What type of PR is this?**

Fixes cluster recreation logic within the new AB E2E go implementation. Specifically, the testing code will now create a new cluster if the 'agentbaker-e2e-test-cluster' in 'agentbaker-e2e-tests' RG does not already exist when the test is run. It will check for both RG and cluster existence to determine whether or not a new cluster needs to be created.

Fixes #2916

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -6,30 +6,64 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 )
 
 func setupCluster(ctx context.Context, cloud *azureClient, location, resourceGroupName, clusterName string) error {
-	aksCluster, err := cloud.aksClient.Get(ctx, resourceGroupName, clusterName, nil)
+	var needNewCluster bool
+
+	rgExistence, err := cloud.resourceGroupClient.CheckExistence(ctx, resourceGroupName, nil)
 	if err != nil {
-		return fmt.Errorf("failed to get aks cluster: %q", err)
+		return fmt.Errorf("failed to get AB E2E RG %q: %q", resourceGroupName, err)
 	}
 
-	rgExistence, err := cloud.resourceGroupClient.CheckExistence(ctx, agentbakerTestResourceGroupName, nil)
-	if err != nil {
-		return fmt.Errorf("failed to get MC RG: %q", err)
+	if !rgExistence.Success {
+		needNewCluster = true
+		_, err := cloud.resourceGroupClient.CreateOrUpdate(
+			ctx,
+			resourceGroupName,
+			armresources.ResourceGroup{
+				Location: to.Ptr("eastus"),
+				Name:     to.Ptr(resourceGroupName),
+			},
+			nil)
+
+		if err != nil {
+			return fmt.Errorf("failed to create AB E2E RG %q: %q", resourceGroupName, err)
+		}
 	}
 
-	if !rgExistence.Success || aksCluster.Properties == nil || aksCluster.Properties.ProvisioningState == nil || *aksCluster.Properties.ProvisioningState == "Failed" {
-		poller, err := cloud.aksClient.BeginDelete(ctx, resourceGroupName, clusterName, nil)
+	if !needNewCluster {
+		aksCluster, err := cloud.aksClient.Get(ctx, resourceGroupName, clusterName, nil)
 		if err != nil {
-			return fmt.Errorf("failed to start aks cluster deletion: %q", err)
+			if !isResourceNotFoundError(err) {
+				return fmt.Errorf("failed to get aks cluster %q: %q", clusterName, err)
+			}
+			needNewCluster = true
 		}
 
-		_, err = poller.PollUntilDone(ctx, nil)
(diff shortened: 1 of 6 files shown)


e2e/cluster.go
```diff
@@ -6,30 +6,64 @@ import (
...
 
-	if !rgExistence.Success || aksCluster.Properties == nil || aksCluster.Properties.ProvisioningState == nil || *aksCluster.Properties.ProvisioningState == "Failed" {
-		poller, err := cloud.aksClient.BeginDelete(ctx, resourceGroupName, clusterName, nil)
+	if !rgExistence.Success {
+		needNewCluster = true
+		_, err := cloud.resourceGroupClient.CreateOrUpdate(
+			ctx,
+			resourceGroupName,
+			armresources.ResourceGroup{
+				Location: to.Ptr("eastus"),
```
ganeshkumar5699, line 26: nit: can this be a variable, if we want to change it to a different location later?

e2e/cluster.go
```diff
@@ -6,30 +6,64 @@ import (
...
-		return fmt.Errorf("failed to get aks cluster: %q", err)
-	}
+	var needNewCluster bool
 
-	rgExistence, err := cloud.resourceGroupClient.CheckExistence(ctx, agentbakerTestResourceGroupName, nil)
+	rgExistence, err := cloud.resourceGroupClient.CheckExistence(ctx, resourceGroupName, nil)
 	if err != nil {
-		return fmt.Errorf("failed to get MC RG: %q", err)
+		return fmt.Errorf("failed to get AB E2E RG %q: %q", resourceGroupName, err)
 	}
```
ganeshkumar5699, line 19: Just to confirm, is it possible for err==nil and rgExistence.Success to be false? Based on error message above it seems like non-existent RG will lead to an error, and exit the function before?

cameronmeissner: this function specifically checks for existence of the resource group, rather than just a plain get - my understanding is that if the RG doesn't exist then err will indeed be nil, and will be non-nil due to any other sort of error. I tested this and it seems to be working fine.

e2e/nsenter.go
```diff
@@ -134,6 +122,32 @@ func extractClusterParameters(ctx context.Context, t *testing.T, kube *kubeclien
...
+		return "", fmt.Errorf("failed to list debug pod: %q", err)
+	}
+
+	if len(podList.Items) < 1 {
+		return "", fmt.Errorf("failed to find debug pod, list by selector returned no results")
+	}
+
+	var podName string
+	for _, pod := range podList.Items {
+		if strings.Contains(pod.Spec.NodeName, "aks-nodepool") {
```
cameronmeissner, line 138: we should only be using the debug pod present on the original aks nodepool's single node, rather than potentially using a pod on an ephemeral node from other testing happening simultaneously

was also thinking it might be helpful to bump the number of nodes in the original nodepool to 2 or 3 to make high volumes of parallel tests more robust

e2e/cluster.go
```diff
@@ -6,30 +6,64 @@ import (
...
+			nil)
+
 		if err != nil {
-			return fmt.Errorf("failed to start aks cluster deletion: %q", err)
+			return fmt.Errorf("failed to create AB E2E RG %q: %q", resourceGroupName, err)
 		}
+	}
 
-		_, err = poller.PollUntilDone(ctx, nil)
+	if !needNewCluster {
```
alexeldeib, line 36: nit - bit confusing to check for needNewCluster then fetch the cluster...suppose the condition names could be more unique but not a big deal

e2e/retry.go
```diff
@@ -0,0 +1,27 @@
+package e2e_test
+
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
+func doWithRetry(funcToRetry func() error, numRetries int, waitInterval time.Duration, t *testing.T) error {
```
alexeldeib, line 9: very similar to wait.Poll

alexeldeib (approved):

e2e/retry.go
cameronmeissner: lol toucheeeee

e2e/cluster.go
cameronmeissner: agreed, refactored this to make things a bit more understandable

juan-lee (approved):

PR #2936, opened 2023-03-28
chore: fuzz GetNodeBootstrapping instead of internal AgentBaker functions

cameronmeissner:
**What type of PR is this?**

/kind cleanup

To support privatizing internal AgentBaker APIs, we need to only fuzz the intentionally public-facing AgentBaker API: GetNodeBootstrapping

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


fuzz/api/main.go
```diff
@@ -14,9 +15,15 @@ func Fuzz(data []byte) int {
 		return -1
 	}
 
-	// should not panic
-	baker := agent.InitializeTemplateGenerator()
-	baker.GetNodeBootstrappingCmd(&config)
+	baker, err := agent.NewAgentBaker()
+	if err != nil {
+		return -2
```
alexeldeib, line 20: I don't think this actually does anything? it deprioritizes invalid input but only -1/0/1 make a difference

fuzz/api/main.go
```diff
@@ -14,9 +15,15 @@ func Fuzz(data []byte) int {
...
+	}
+
+	_, err = baker.GetNodeBootstrapping(context.Background(), &config)
+	if err != nil {
+		return -3
```
alexeldeib, line 25: same as above

fuzz/api/main.go
UtheMan, line 20: what's the meaning of -2, -3  (and other) return codes? Is this some internal thing for fuzzing lib?

fuzz/api/main.go
```diff
@@ -14,9 +15,15 @@ func Fuzz(data []byte) int {
...
+	}
 
-	return 1
+	return 0
```
alexeldeib, line 28: this should remain 1 to prioritize valid input as "interesting"

cameronmeissner: oops

fuzz/api/main.go
cameronmeissner: oops

fuzz/api/main.go
cameronmeissner: oops

fuzz/api/main.go
alexeldeib: https://github.com/dvyukov/go-fuzz#usage as ref

fuzz/api/main.go
cameronmeissner: should actually just be -1/0/1

alexeldeib (approved):

alexeldeib (approved):

PR #2954, opened 2023-03-30
chore: clarify properties for windows e2e

junjiezhang1997:
**What type of PR is this?**
/kind cleanup

1. remove unnecessary properties (duplicate properties in nodebootstrapping_template.json and properties for linux) for windows e2e.
2. specify distro and node image version to avoid ambiguity.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/e2e-windows.yaml
+++ b/.pipelines/e2e-windows.yaml
@@ -41,6 +41,8 @@ stages:
       windowsImage: 2019-containerd
       windowsOSSKU: Windows2019
       windowsVMSize: Standard_D2s_v3
+      windowsDistro: aks-windows-2019-containerd
+      windowsNodeImageVersion: AKSWindows-2019-containerd-test-2023.02.07
       storageAccount: akswinstore2019
 
 - stage: test_2022_containerd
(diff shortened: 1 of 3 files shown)


.pipelines/e2e-windows.yaml
```diff
@@ -41,6 +41,8 @@ stages:
       windowsImage: 2019-containerd
       windowsOSSKU: Windows2019
       windowsVMSize: Standard_D2s_v3
+      windowsDistro: aks-windows-2019-containerd
+      windowsNodeImageVersion: AKSWindows-2019-containerd-test-2023.02.07
```
ShiqianTao, line 45: Why is the value `2023.02.07`?

junjiezhang1997: Currently we use the test image referred in Line 139 of e2e/windows/e2e-scenario.sh and it was built by ourselves on 2023/02/07. 

In the future, after moving our vhd sub to ame sub, we can use prod image and sync with the latest image.

.pipelines/e2e-windows.yaml
```diff
@@ -41,6 +41,8 @@ stages:
       windowsImage: 2019-containerd
       windowsOSSKU: Windows2019
       windowsVMSize: Standard_D2s_v3
+      windowsDistro: aks-windows-2019-containerd
```
AbelHu, line 44: Can you help to add a README for how to generate and update the test images, how to generate the template, the brief introduction of windows e2e, etc.?

junjiezhang1997: Could I create a new task rather than in this pr to do it? I think it may take some time.

AbelHu (approved):
LGTM

PR #2956, opened 2023-03-30
chore: test az login in coveralls action

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.github/workflows/check-coverage.yml
+++ b/.github/workflows/check-coverage.yml
@@ -64,12 +64,9 @@ jobs:
             go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}' -deps .  | grep agent | tr '\n' ',' | sed 's/,$//' > packages.txt
             go test -coverpkg=`cat packages.txt` -c -o e2e.exe -timeout 30m -covermode count
             GOCOVERDIR=covdatafiles ./e2e.exe
-            go tool covdata textfmt -i=covdatafiles -o e2e.cov
       - name: Generate coverage report
         run: |
-            mkdir -p covdatafiles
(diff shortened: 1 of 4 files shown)


.github/workflows/check-coverage.yml
```diff
@@ -21,6 +32,14 @@ jobs:
...
         run: |
           sed '/mocks.go/d' coverage_raw.out > coverage.out
+      - name: Az CLI login
+        uses: azure/login@v1
+        with:
+          client-id: ${{ secrets.GO_E2E_CLIENT_ID }}
+          tenant-id: ${{ secrets.GO_E2E_TENANT_ID }}
+          subscription-id: ${{ secrets.GO_E2E_SUBSCRIPTION_ID }}
+      - name: Run e2e tests
+        run: cd e2e; go test -timeout 30m -v -run Test_All ./...
```
cameronmeissner, line 42: is there a way we can reference e2e.yaml which is already used by the ADO pipeline to run the tests? This would be the 3rd instance of this run command spread across different files (already one in e2e.yaml and one in e2e-local-go.sh)

cameronmeissner: hmmmm, I'm also thinking it could be useful to implement a 'dry-run' mode for AB E2E that wouldn't actually interact with Azure very much (if at all) so we could keep these coverage checks reasonably light-weight and localized - would also make sense since really the coverage that comes from these E2E tests is only directly dependent on the NodeBootstrappingConfiguration which gets fed into the API (dry-run could basically just feed the nbc into agentbaker and verify no panics/errors occur, similar to fuzz)

PR #2960, opened 2023-03-30
style: fixing all lines over 111 chars

AlisonB319:
fixing all lines as directed from lint lll

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/bakerapi.go
```diff
@@ -11,9 +11,12 @@ import (
 )
 
 type AgentBaker interface {
-	GetNodeBootstrapping(ctx context.Context, config *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error)
-	GetLatestSigImageConfig(sigConfig datamodel.SIGConfig, region string, distro datamodel.Distro) (*datamodel.SigImageConfig, error)
-	GetDistroSigImageConfig(sigConfig datamodel.SIGConfig, region string) (map[datamodel.Distro]datamodel.SigImageConfig, error)
+	GetNodeBootstrapping(ctx context.Context, config *datamodel.NodeBootstrappingConfiguration) (
```
alexeldeib, line 14: hmmm...what's the actual longest failing line here?

I don't really love this line splitting in all cases, what do you think about raising the length limit on the linter instead?

AlisonB319: And also turns out the limit is acutally 120 and not 111  so I made them all shorter for no reason lol. The longest lines is (technically) 3803 with [EncodedTestCert](https://github.com/AlisonB319/AgentBaker/blob/master/pkg/agent/baker_test.go#L1481) lol but most of the lines that exceed 120 are around 140-180

AlisonB319: I think I am going to make a new PR with a better lll implementation.

PR #2963, opened 2023-03-31
refactor: increase line length linter to 150

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -170,6 +170,16 @@ linters-settings:
     # Otherwise, only methods that take `*testing.T`, `*testing.B`, and `testing.TB` as arguments are checked.
     # Default: false
     all: true
+  
+  lll:
+    # Max line length, lines longer will be reported.
+    # '\t' is counted as 1 character by default, and can be changed with the tab-width option.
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -170,6 +170,16 @@ linters-settings:
...
     all: true
+  
+  lll:
+    # Max line length, lines longer will be reported.
+    # '\t' is counted as 1 character by default, and can be changed with the tab-width option.
+    # Default: 120.
+    line-length: 150
+    # Tab width in spaces.
+    # Default: 1
+    tab-width: 1
```
alexeldeib, line 181: is that actually correct? I would imagine 2 or 4?

alexeldeib: oh duh. I see the comment above now. I guess I'm a spaces-over-tabs person but we don't enforce this anywhere anyway which is a bit annoying (something we could maybe fix with gitattributes or similar?).

alexeldeib (approved):

PR #2964, opened 2023-03-31
refactor: fix lint errors for line length linter

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -643,7 +648,8 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			parameters := getParameters(config, "baker", "1.0")
 			// get variable cloudInit
 			variables := getCustomDataVariables(config)
-			containerdConfigTemplate := template.Must(template.New("kubenet").Funcs(getBakerFuncMap(config, parameters, variables)).Parse(containerdConfigTemplateString))
+			containerdConfigTemplate := template.Must(template.New("kubenet").Funcs(getBakerFuncMap(config, parameters, variables)).Parse(
```
alexeldeib, line 651: nit - might be easier to read for this one to break out the vars separately?

pkg/agent/variables.go
```diff
@@ -12,8 +12,6 @@ import (
 	"github.com/blang/semver"
 )
 
-var dockerShimFlags = []string{"--cni-bin-dir", "--cni-cache-dir", "--cni-conf-dir", "--docker-endpoint", "--image-pull-progress-deadline", "--network-plugin", "--network-plugin-mtu"}
```
alexeldeib, line 15: hmm, what happened here?

if this is because you can't have a const slice, i'd prefer a func wrapper to return the locally defined func for the "const-ness" personally

pkg/agent/baker.go
```diff
@@ -273,7 +273,10 @@ func validateAndSetLinuxNodeBootstrappingConfiguration(config *datamodel.NodeBoo
 		kubeletFlags := config.KubeletConfig
 		delete(kubeletFlags, "--dynamic-config-dir")
 		delete(kubeletFlags, "--non-masquerade-cidr")
-		if profile != nil && profile.KubernetesConfig != nil && profile.KubernetesConfig.ContainerRuntime != "" && profile.KubernetesConfig.ContainerRuntime == "containerd" {
+		if profile != nil && profile.KubernetesConfig != nil && profile.KubernetesConfig.ContainerRuntime != "" &&
+			profile.KubernetesConfig.ContainerRuntime == "containerd" {
+			dockerShimFlags := []string{"--cni-bin-dir", "--cni-cache-dir", "--cni-conf-dir", "--docker-endpoint",
```
alexeldeib, line 278: can probably expand this as multiline but also see later comment?

pkg/agent/baker_test.go
```diff
@@ -359,196 +366,206 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
...
-				"--authentication-token-webhook":      "true",
-				"--authorization-mode":                "Webhook",
-				"--eviction-hard":                     "memory.available<750Mi,nodefs.available<10%,nodefs.inodesFree<5%",
-				"--feature-gates":                     "RotateKubeletServerCertificate=true,a=b,PodPriority=true,x=y",
-				"--system-reserved":                   "cpu=2,memory=1Gi",
-				"--kube-reserved":                     "cpu=100m,memory=1638Mi",
-				"--dynamic-config-dir":                "",
-			}
-		}, nil),
+				config.KubeletConfig = map[string]string{}
```
alexeldeib, line 375: this diff is a wee bit confusing -- what actually changed here besides the //nolint?

I don't see anything at a glance but git seems to disagree

alexeldeib: specifically around ordering of test scenarios, the new line breaks make sense

alexeldeib: got it, all the func() have been shifted which is confusing. it looks fine in your branch

pkg/agent/datamodel/helper.go
```diff
@@ -22,7 +22,9 @@ func ValidateDNSPrefix(dnsName string) error {
 		return err
 	}
 	if !re.MatchString(dnsName) {
-		return errors.Errorf("DNSPrefix '%s' is invalid. The DNSPrefix must contain between 3 and 45 characters and can contain only letters, numbers, and hyphens.  It must start with a letter and must end with a letter or a number. (length was %d)", dnsName, len(dnsName))
+		return errors.Errorf("DNSPrefix '%s' is invalid. The DNSPrefix must contain between 3 and 45 characters"+
```
alexeldeib, line 25: nit: extract the error template to a multiline string outside the function body?

pkg/agent/variables.go
AlisonB319: This is because lint-pr complained that dockerShimFlags was a global variable. It was only used in `validateAndSetLinuxNodeBootstrappingConfiguration ` so I added the variable inside the function instead. Are you saying that you would rather have a getter function that returns the flags?

pkg/agent/datamodel/helper_test.go
```diff
@@ -24,27 +24,33 @@ func TestValidateDNSPrefix(t *testing.T) {
 		{
 			"empty string",
 			"",
-			errors.New("DNSPrefix '' is invalid. The DNSPrefix must contain between 3 and 45 characters and can contain only letters, numbers, and hyphens.  It must start with a letter and must end with a letter or a number. (length was 0)"),
+			errors.New("DNSPrefix '' is invalid. The DNSPrefix must contain between 3 and 45 characters and can contain only letters," +
```
alexeldeib, line 27: this feels a little silly, we are contorting test code with valid long errors to satisfy linting.

I'm focused less on perfectly following the rules and more catching real issues // consistency.

I guess we already adjusted the line length -- what do you think about this one?

pkg/agent/datamodel/helper.go
AlisonB319: golint will probably complain of a global variable if I take it outside of the function body. But I could define it as a variable inside the function?

pkg/agent/datamodel/helper_test.go
AlisonB319: I don't mind changing it back and just adding an //nolint here.

pkg/agent/datamodel/helper.go
AlisonB319: Also can I make a multiline string in go using ` when there are variables in the string?

PR #2965, opened 2023-03-31
feat: add cluster config scenario support to abe2e

cameronmeissner:
**What type of PR is this?**

/kind feature

implements the ground-level framework and logic for adding ab e2e scenarios that are based around cluster configuration rather than VHD selection

MOVED TO A DIFFERENT BRANCH TO KEEP THINGS NEAT: https://github.com/Azure/AgentBaker/pull/2982

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/suite_test.go
```diff
@@ -25,93 +24,83 @@ func Test_All(t *testing.T) {
...
 	}
 
-	kube, err := getClusterKubeClient(ctx, cloud, suiteConfig)
-	if err != nil {
-		t.Fatal(err)
-	}
+	for _, scenario := range scenarioTable {
+		scenario := scenario
+
+		chosenCluster := chooseCompatibleCluster(scenario, clusters)
```
alexeldeib, line 46: hmm. we have a lot of setup inside this loop now -- the execution of each case blocks on the creation of the cluster, and beginning execution of the later cases blocks on cluster creation for all previous steps.

can we simplify or extract the setup somehow?

we could move it inside the actual t.Run(), at risk of causing conflicts, but we could filter out 409s.

alexeldeib: separate to that, wonder if we can wrap some of this into a setup function to keep the top level loop cleaner

e2e/suite_test.go
```diff
@@ -25,93 +24,83 @@ func Test_All(t *testing.T) {
...
+		kube, err := getClusterKubeClient(ctx, cloud, suiteConfig.resourceGroupName, clusterName)
+		if err != nil {
+			t.Fatal(err)
+		}
 
-	t.Logf("dumping cluster parameters to local directory: %s", clusterParamsDir)
-	if err := dumpFileMapToDir(clusterParamsDir, clusterParams); err != nil {
-		t.Error("error dumping cluster parameters", err)
-	}
+		if err := ensureDebugDaemonset(ctx, kube); err != nil {
```
alexeldeib, line 75: should we extract this into a cluster setup function?

e2e/suite_test.go
```diff
@@ -25,93 +24,83 @@ func Test_All(t *testing.T) {
...
 		nbc := copied.(*datamodel.NodeBootstrappingConfiguration)
 
-		if tc.bootstrapConfigMutator != nil {
-			tc.bootstrapConfigMutator(t, nbc)
+		if scenario.ScenarioConfig.BootstrapConfigMutator != nil {
+			scenario.ScenarioConfig.BootstrapConfigMutator(t, nbc)
 		}
 
-		t.Run(caseName, func(t *testing.T) {
+		t.Run(scenario.Name, func(t *testing.T) {
```
alexeldeib, line 100: our run function is also getting a little long -- maybe extract this all to a `runScenario` or similar helper?

e2e/scenario/clusterconfiguration.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Selectors
+
+func NetworkPluginKubenetSelector(cluster *armcontainerservice.ManagedCluster) bool {
+	return *cluster.Properties.NetworkProfile.NetworkPlugin == armcontainerservice.NetworkPluginKubenet
+}
+
+// Mutators
+
+func NetworkPluginKubenetMutator(cluster *armcontainerservice.ManagedCluster) {
```
alexeldeib, line 16: nice

alexeldeib: perhaps a few nil checks?

e2e/suite_test.go
cameronmeissner: yeees, was already thinking about that

e2e/suite_test.go
```diff
@@ -25,93 +24,83 @@ func Test_All(t *testing.T) {
...
+			t.Fatal(err)
+		}
 
-	baseConfig, err := getBaseBootstrappingConfig(ctx, t, cloud, suiteConfig, clusterParams)
-	if err != nil {
-		t.Fatal(err)
-	}
+		clusterParams, err := pollExtractClusterParameters(ctx, t, kube)
+		if err != nil {
+			t.Fatal(err)
```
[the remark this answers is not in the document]
cameronmeissner: right, maybe it makes the most sense to gather a set of "candidate" clusters that satisfy the scenario's cluster selector instead of naively selecting the first satisfactory one - we'd then only fatally error if all candidate clusters returned an error during this operation

e2e/suite_test.go
cameronmeissner: will do

e2e/suite_test.go
cameronmeissner: definitely agree with wrapping setup into a separate function, will play with moving these bits around to improve parallelization

cameronmeissner: have made some progress on the parallelization stuff, moved that to separate branch which I'll follow up with soon (mainly to keep this PR manageable)

PR #2968, opened 2023-04-03
refactor: refactor functions over 100 lines and complexity over 20

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -322,6 +324,7 @@ func validateAndSetWindowsNodeBootstrappingConfiguration(config *datamodel.NodeB
 // getContainerServiceFuncMap returns all functions used in template generation.
 /* These funcs are a thin wrapper for template generation operations,
 all business logic is implemented in the underlying func. */
+//nolint:funlen,gocognit,nolintlint
```
AlisonB319, line 327: Added nolintlint here because there would be an "unused" error for nolint:funlen without it. Which I think is a bug on golangci side since it is very much used. There is an option to add - unused:true inside the golangci.yaml but I don't think we should totally suppress it. 

On line 140 - https://gitdab.com/Russtopia/xs/commit/5c826f7a5f1791e82f3e98e126bb0a2faa6cd130 - this repository just suppresses the specific message. They also think this is due to a bug in golangci-lint

pkg/agent/datamodel/mocks.go
```diff
@@ -197,7 +197,7 @@ var (
 			WindowsTelemetryGUID:                 "fb801154-36b9-41bc-89c2-f4d4f05472b0",
 			CNIPluginsDownloadURL:                "https://acs-mirror.azureedge.net/cni/cni-plugins-amd64-v0.7.6.tgz",
 			VnetCNILinuxPluginsDownloadURL:       "https://acs-mirror.azureedge.net/azure-cni/v1.1.3/binaries/azure-vnet-cni-linux-amd64-v1.1.3.tgz",
-			VnetCNIWindowsPluginsDownloadURL:     "https://acs-mirror.azureedge.net/azure-cni/v1.1.3/binaries/azure-vnet-cni-singletenancy-windows-amd64-v1.1.3.zip",
+			VnetCNIWindowsPluginsDownloadURL:     "https://acs-mirror.azureedge.net/azure-cni/v1.1.3/binaries/azure-vnet-cni-singletenancy-windows-amd64-v1.1.3.zip", //nolint:lll
```
cameronmeissner, line 200: wondering if it would be better to have lll ignore this file entirely since it's only used for mocks

pkg/agent/datamodel/types_test.go
```diff
@@ -53,6 +53,7 @@ func TestGetCustomEnvironmentJSON(t *testing.T) {
 	}
 }
 
+//nolint:funlen,nolintlint
```
cameronmeissner, line 56: maaaaybe also worth to have funlen ignore *_test.go files?

curious what others think

pkg/agent/params.go
```diff
@@ -114,63 +179,8 @@ func assignKubernetesParameters(properties *datamodel.Properties, parametersMap
...
-				addValue(parametersMap, "kubeBinariesSASURL", k8sComponents.WindowsPackageURL)
-
-				addValue(parametersMap, "windowsContainerdURL", kubernetesConfig.WindowsContainerdURL)
-				addValue(parametersMap, "kubeServiceCidr", kubernetesConfig.ServiceCIDR)
-				addValue(parametersMap, "kubeBinariesVersion", k8sVersion)
-				addValue(parametersMap, "windowsTelemetryGUID", cloudSpecConfig.KubernetesSpecConfig.WindowsTelemetryGUID)
-				addValue(parametersMap, "windowsSdnPluginURL", kubernetesConfig.WindowsSdnPluginURL)
-			}
-		}
+		parametersMap = assignKubernetesParametersfromKubernetesConfig(kubernetesConfig, properties, parametersMap, cloudSpecConfig,
```
cameronmeissner, line 182: noice

pkg/agent/utils.go
```diff
@@ -362,12 +362,7 @@ func GetTLSBootstrapTokenForKubeConfig(tlsBootstrapToken *string) string {
 	return *tlsBootstrapToken
 }
 
-// GetKubeletConfigFileContent converts kubelet flags we set to a file, and return the json content.
-func GetKubeletConfigFileContent(kc map[string]string, customKc *datamodel.CustomKubeletConfig) string {
-	if kc == nil {
-		return ""
-	}
-	// translate simple values.
+func GetKubletConfigFromAKSKubeletConfiguration(kc map[string]string, customKc *datamodel.CustomKubeletConfig) *datamodel.AKSKubeletConfiguration {
```
cameronmeissner, line 365: opinionated nit: maybe rename to GetAKSKubeletConfiguration or similar

pkg/agent/utils.go
```diff
@@ -362,12 +362,7 @@ func GetTLSBootstrapTokenForKubeConfig(tlsBootstrapToken *string) string {
 	return *tlsBootstrapToken
 }
 
-// GetKubeletConfigFileContent converts kubelet flags we set to a file, and return the json content.
-func GetKubeletConfigFileContent(kc map[string]string, customKc *datamodel.CustomKubeletConfig) string {
-	if kc == nil {
```
cameronmeissner, line 367: do we still need this nil check?

pkg/agent/utils.go
```diff
@@ -397,41 +392,12 @@ func GetKubeletConfigFileContent(kc map[string]string, customKc *datamodel.Custo
...
-
-	// system reserve and kube reserve.
-	// looks like "cpu=100m,memory=1638Mi".
-	kubeletConfig.SystemReserved = strKeyValToMap(kc["--system-reserved"], ",", "=")
-	kubeletConfig.KubeReserved = strKeyValToMap(kc["--kube-reserved"], ",", "=")
-
-	// Settings from customKubeletConfig, only take if it's set.
+//nolint:gocognit
+func SetKubletConfigFileFromCustomKubeletConfig(customKc *datamodel.CustomKubeletConfig,
+	kubeletConfig *datamodel.AKSKubeletConfiguration) *datamodel.AKSKubeletConfiguration {
```
cameronmeissner, line 400: nit: don't need to explicitly return anything here if we're just mutating kubeletConfig

pkg/agent/utils.go
cameronmeissner: oh woops i can't read diffs

pkg/agent/utils.go
```diff
@@ -397,41 +392,12 @@ func GetKubeletConfigFileContent(kc map[string]string, customKc *datamodel.Custo
...
-	kubeletConfig.FeatureGates = strKeyValToMapBool(kc["--feature-gates"], ",", "=")
-
-	// system reserve and kube reserve.
-	// looks like "cpu=100m,memory=1638Mi".
-	kubeletConfig.SystemReserved = strKeyValToMap(kc["--system-reserved"], ",", "=")
-	kubeletConfig.KubeReserved = strKeyValToMap(kc["--kube-reserved"], ",", "=")
-
-	// Settings from customKubeletConfig, only take if it's set.
+//nolint:gocognit
+func SetKubletConfigFileFromCustomKubeletConfig(customKc *datamodel.CustomKubeletConfig,
```
cameronmeissner, line 399: i'd rename to SetCustomKubeletConfig or similar

pkg/agent/utils.go
cameronmeissner, line 365: customKc doesn't seem to be used here

cameronmeissner (approved):
overall lgtm, thanks for picking this up!

will defer to @alexeldeib for final sign-off on some of the linting exceptions

pkg/agent/datamodel/types_test.go
AlisonB319: I agree, and we actually already had it in our exclude-rules - 
```
- path: "_test\\.go"
      linters:
        - bodyclose
        - dupl
        - funlen
        - goconst
        - gosec
        - noctx
        - wrapcheck
```
I had just removed it during CLI testing! I removed the //nolints :)

pkg/agent/datamodel/mocks.go
AlisonB319: This is the only //nolint:lll in the file, so I think having one is fine, but if it seems to be a problem later on we can add it to the excludes in .golangci.yaml?

e2e/template_test.go
```diff
@@ -459,6 +459,8 @@ spec:
         app: *name
     spec:
       hostNetwork: true
+      nodeSelector:
```
alexeldeib, line 462: rebase from Camerons PR?

PR #2971, opened 2023-04-04
feat: specify the k8s version to support windows cse validation

wanqingfu:
**What type of PR is this?**
Support to specify the k8s version since some feature may need the latest k8s version to validate.

To validate latest windows cse package on the latest k8s version.

To achieve this
- The resource group name will contain the version string and the cluster will be created with the specified k8s version.
- The windows vmss deployment will be associated with the set k8s version by updating `WindowsPackageURL` of `k8scomponents` in the template `property-windows-template.json` which is for generating windows cse related files i.e. CustomData file and CSECmd file.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -31,11 +31,17 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
         export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
-        export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
-        if [[ -n "$WINDOWS_PACKAGE_URL" ]]; then
-          export windowsPackageURL="$WINDOWS_PACKAGE_URL"
+        if [[ -n "${WINDOWS_PACKAGE_URL}" ]]; then
+          export windowsPackageURL=$WINDOWS_PACKAGE_URL
+          export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
+          if [[ "${WINDOWS_PACKAGE_VERSION}" != "${KUBERNETES_VERSION}" ]]; then
+            echo "Windows package version $WINDOWS_PACKAGE_VERSION does not match with kubernetes version $KUBERNETES_VERSION. Invalid test scenario."
+            exit 1
+          fi
+        else
+          export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
+          export WINDOWS_PACKAGE_VERSION=$KUBERNETES_VERSION
         fi
-        export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
         export K8S_VERSION=${WINDOWS_PACKAGE_VERSION//./}
         export WINDOWS_E2E_STORAGE_CONTAINER=${{ parameters.storageAccount }}-$K8S_VERSION
         mkdir -p $WINDOWS_E2E_IMAGE
@@ -77,11 +83,17 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_DISTRO=${{ parameters.windowsDistro }}
         export WINDOWS_NODE_IMAGE_VERSION=${{ parameters.windowsNodeImageVersion }}
-        export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
-        if [[ -n "$WINDOWS_PACKAGE_URL" ]]; then
-          export windowsPackageURL="$WINDOWS_PACKAGE_URL"
+        if [[ -n "${WINDOWS_PACKAGE_URL}" ]]; then
+          export windowsPackageURL=$WINDOWS_PACKAGE_URL
+          export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
+          if [[ "${WINDOWS_PACKAGE_VERSION}" != "${KUBERNETES_VERSION}" ]]; then
+            echo "Windows package version $WINDOWS_PACKAGE_VERSION does not match with kubernetes version $KUBERNETES_VERSION. Invalid test scenario."
+            exit 1
+          fi
+        else
+          export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
+          export WINDOWS_PACKAGE_VERSION=$KUBERNETES_VERSION
         fi
-        export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
         export K8S_VERSION=${WINDOWS_PACKAGE_VERSION//./}
         mkdir -p $WINDOWS_E2E_IMAGE
         cp -a $(Pipeline.Workspace)/${{ parameters.windowsImage }}-clusterConfig/* $WINDOWS_E2E_IMAGE
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -31,8 +31,13 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
         export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
-        K=${KUBERNETES_VERSION//./}
-        export WINDOWS_E2E_STORAGE_CONTAINER=${{ parameters.storageAccount }}-$K
+        export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
+        if [[ -n "$WINDOWS_PACKAGE_URL" ]]; then
+          export windowsPackageURL="$WINDOWS_PACKAGE_URL"
+        fi
+        export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
+        export K8S_VERSION=${WINDOWS_PACKAGE_VERSION//./}
+        export WINDOWS_E2E_STORAGE_CONTAINER=${{ parameters.storageAccount }}-$K8S_VERSION
         mkdir -p $WINDOWS_E2E_IMAGE
         cp -r e2e/windows/* $WINDOWS_E2E_IMAGE
         go version
@@ -72,6 +77,12 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_DISTRO=${{ parameters.windowsDistro }}
         export WINDOWS_NODE_IMAGE_VERSION=${{ parameters.windowsNodeImageVersion }}
+        export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
+        if [[ -n "$WINDOWS_PACKAGE_URL" ]]; then
+          export windowsPackageURL="$WINDOWS_PACKAGE_URL"
+        fi
+        export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
+        export K8S_VERSION=${WINDOWS_PACKAGE_VERSION//./}
         mkdir -p $WINDOWS_E2E_IMAGE
         cp -a $(Pipeline.Workspace)/${{ parameters.windowsImage }}-clusterConfig/* $WINDOWS_E2E_IMAGE
         cd $WINDOWS_E2E_IMAGE
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -31,7 +31,8 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
         export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
-        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}
+        K=${KUBERNETES_VERSION//./}
+        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}$K
         mkdir -p $WINDOWS_E2E_IMAGE
         cp -r e2e/windows/* $WINDOWS_E2E_IMAGE
         go version
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -32,7 +32,7 @@ jobs:
         export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
         export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
         K=${KUBERNETES_VERSION//./}
-        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}$K
+        export WINDOWS_E2E_STORAGE_CONTAINER=${{ parameters.storageAccount }}-$K
         mkdir -p $WINDOWS_E2E_IMAGE
         cp -r e2e/windows/* $WINDOWS_E2E_IMAGE
         go version
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -44,7 +44,7 @@ jobs:
           fi
         else
           if [[ -z "${KUBERNETES_VERSION}" ]]; then
-            KUBERNETES_VERSION=$(az aks get-versions -l eastus --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
+            KUBERNETES_VERSION=$(az aks get-versions -l $LOCATION --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
             echo "Using default kubernetes version ${KUBERNETES_VERSION} for the windows package"
           fi
           export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
@@ -96,7 +96,7 @@ jobs:
           export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
         else
           if [[ -z "${KUBERNETES_VERSION}" ]]; then
-            KUBERNETES_VERSION=$(az aks get-versions -l eastus --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
+            KUBERNETES_VERSION=$(az aks get-versions -l $LOCATION --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
             echo "Using default kubernetes version ${KUBERNETES_VERSION} for the windows package"
           fi
           export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -34,11 +34,19 @@ jobs:
         if [[ -n "${WINDOWS_PACKAGE_URL}" ]]; then
           export windowsPackageURL=$WINDOWS_PACKAGE_URL
           export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
+          if [[ -z "${KUBERNETES_VERSION}" ]]; then
+            echo "You must set KUBERNETES_VERSION when setting WINDOWS_PACKAGE_URL"
+            exit 1
+          fi
           if [[ "${WINDOWS_PACKAGE_VERSION}" != "${KUBERNETES_VERSION}" ]]; then
             echo "Windows package version $WINDOWS_PACKAGE_VERSION does not match with kubernetes version $KUBERNETES_VERSION. Invalid test scenario."
             exit 1
           fi
         else
+          if [[ -z "${KUBERNETES_VERSION}" ]]; then
+            KUBERNETES_VERSION=$(az aks get-versions -l eastus --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
+            echo "Using default kubernetes version ${KUBERNETES_VERSION} for the windows package"
+          fi
           export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
           export WINDOWS_PACKAGE_VERSION=$KUBERNETES_VERSION
         fi
@@ -87,6 +95,10 @@ jobs:
           export windowsPackageURL=$WINDOWS_PACKAGE_URL
           export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
         else
+          if [[ -z "${KUBERNETES_VERSION}" ]]; then
+            KUBERNETES_VERSION=$(az aks get-versions -l eastus --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
+            echo "Using default kubernetes version ${KUBERNETES_VERSION} for the windows package"
+          fi
           export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
           export WINDOWS_PACKAGE_VERSION=$KUBERNETES_VERSION
         fi
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -86,10 +86,6 @@ jobs:
         if [[ -n "${WINDOWS_PACKAGE_URL}" ]]; then
           export windowsPackageURL=$WINDOWS_PACKAGE_URL
           export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
-          if [[ "${WINDOWS_PACKAGE_VERSION}" != "${KUBERNETES_VERSION}" ]]; then
-            echo "Windows package version $WINDOWS_PACKAGE_VERSION does not match with kubernetes version $KUBERNETES_VERSION. Invalid test scenario."
-            exit 1
-          fi
         else
           export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
           export WINDOWS_PACKAGE_VERSION=$KUBERNETES_VERSION
--- a/e2e/windows/e2e-create-windows-nodepool.sh
+++ b/e2e/windows/e2e-create-windows-nodepool.sh
@@ -6,10 +6,8 @@ source e2e-helper.sh
 
 log "Starting to create windows nodepool"
 
-if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
-    K8S_VERSION=$(echo $KUBERNETES_VERSION | tr '.' '-')
-    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-"$K8S_VERSION"
-fi
+K8S_VERSION=$(echo $KUBERNETES_VERSION | tr '.' '-')
+RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-"$K8S_VERSION"
 
 out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME | jq '.[].name')
 
--- a/e2e/windows/e2e-create-windows-nodepool.sh
+++ b/e2e/windows/e2e-create-windows-nodepool.sh
@@ -7,7 +7,8 @@ source e2e-helper.sh
 log "Starting to create windows nodepool"
 
 if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
-    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
+    K8S_VERSION=$(echo $KUBERNETES_VERSION | tr '.' '-')
+    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-"$K8S_VERSION"
 fi
 
 out=$(az aks nodepool list --cluster-name $CLUSTER_NAME -g $RESOURCE_GROUP_NAME | jq '.[].name')
--- a/e2e/windows/e2e-create-windows-nodepool.sh
+++ b/e2e/windows/e2e-create-windows-nodepool.sh
@@ -6,8 +6,11 @@ source e2e-helper.sh
 
 log "Starting to create windows nodepool"
 
-K8S_VERSION=$(echo $KUBERNETES_VERSION | tr '.' '-')
-RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-"$K8S_VERSION"
+if echo "$windowsPackageURL" | grep -q "hotfix"; then
+    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME-$WINDOWS_E2E_IMAGE-$K8S_VERSION-h"
+else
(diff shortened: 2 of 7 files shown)


e2e/windows/e2e-scenario.sh
```diff
@@ -268,8 +296,8 @@ else
     waitForDeleteStartTime=$(date +%s)
 
     # Only delete node and vmss since we reuse the resource group and cluster now
-    kubectl delete node $VMSS_INSTANCE_NAME
-    az vmss delete -g $(jq -r .group $SCENARIO_NAME-vmss.json) -n $(jq -r .vmss $SCENARIO_NAME-vmss.json)
+    # kubectl delete node $VMSS_INSTANCE_NAME
+    # az vmss delete -g $(jq -r .group $SCENARIO_NAME-vmss.json) -n $(jq -r .vmss $SCENARIO_NAME-vmss.json)
```
junjiezhang1997, line 300: Please add delete code back after testing.

e2e/windows/nodebootstrapping_template.json
```diff
@@ -8,11 +8,11 @@
         "properties": {
             "orchestratorProfile": {
                 "orchestratorType": "Kubernetes",
-                "orchestratorVersion": "1.24.0",
```
junjiezhang1997, line 11: I still think the change in nodebootstrapping_template.json is unnecessary. You can remove these code change and test again.

e2e/windows/e2e-scenario.sh
wanqingfu: Will do. Still testing different k8s versions.

e2e/windows/nodebootstrapping_template.json
wanqingfu: Removed and it worked.

e2e/windows/e2e-scenario.sh
wanqingfu: Added back after testing.

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -31,7 +31,8 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
         export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
-        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}
+        K=${KUBERNETES_VERSION//./}
+        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}$K
```
AbelHu, line 35: Suggest to create containers in existing storage accounts instead of creating storage accounts.

e2e/windows/e2e-create-windows-nodepool.sh
```diff
@@ -6,9 +6,8 @@ source e2e-helper.sh
 
 log "Starting to create windows nodepool"
 
-if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
-    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
-fi
+K8S_VERSION=$(echo $KUBERNETES_VERSION | tr '.' '-')
+RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-"$K8S_VERSION"
```
AbelHu, line 10: will this impact the existing pipelines/tests? I think that @junjiezhang1997 can explain why the old code uses v2.

e2e/windows/scenarios/windows/property-windows-template.json
```diff
@@ -61,7 +61,7 @@
     }
   },
   "K8sComponents": {
-    "WindowsPackageURL": "https://acs-mirror.azureedge.net/kubernetes/v1.23.12/windowszip/v1.23.12-1int.zip"
+    "WindowsPackageURL": "https://acs-mirror.azureedge.net/kubernetes/v${orchestratorVersion}/windowszip/v${orchestratorVersion}-1int.zip"
```
AbelHu, line 64: can we support to test a hotfix package?
v1.25.2-hotfix.20221006/windowszip/v1.25.2-hotfix.20221006-1int.zip

e2e/windows/e2e-scenario.sh
```diff
@@ -43,7 +43,8 @@ collect-logs() {
     set -x
 }
 
-RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
+K8S_VERSION=$(echo $KUBERNETES_VERSION | tr '.' '-')
+RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-"$K8S_VERSION"
 
 DEPLOYMENT_VMSS_NAME="$(mktemp -u winXXXXX | tr '[:upper:]' '[:lower:]')"
```
AbelHu, line 49: this is for @junjiezhang1997 : can we use a meaningful name `winXXXXX `?

e2e/windows/e2e-scenario.sh
```diff
@@ -159,14 +177,24 @@ jq --argjson JsonForVnet "$WINDOWS_VNET" \
     --arg ValueForCustomData "$CUSTOM_DATA" \
     --arg ValueForCSECmd "$CSE_CMD" \
     --arg ValueForVMSS "$DEPLOYMENT_VMSS_NAME" \
-    '.parameters += $JsonForVnet | .parameters += $JsonForLB | .parameters.galleries_AKSWindows_externalid.defaultValue=$ValueForImageExternalID | .resources[0] += $JsonForIdentity | .resources[0] += $JsonForSKU | .resources[0].properties.virtualMachineProfile.storageProfile+=$JsonForOSDisk | .resources[0].properties.virtualMachineProfile.networkProfile.networkInterfaceConfigurations[0] += $JsonForNetwork | .resources[0].properties.virtualMachineProfile.storageProfile.imageReference.id=$ValueForImageReference | .resources[0].properties.virtualMachineProfile.osProfile.adminPassword=$ValueForAdminPassword | .resources[0].properties.virtualMachineProfile.osProfile.customData=$ValueForCustomData | .resources[0].properties.virtualMachineProfile.extensionProfile.extensions[0].properties.settings.commandToExecute=$ValueForCSECmd | .parameters.virtualMachineScaleSets_akswin30_name.defaultValue=$ValueForVMSS' \
-    windows_vmss_template.json > $DEPLOYMENT_VMSS_NAME-deployment.json
+    '.parameters += $JsonForVnet | .parameters += $JsonForLB |
```
AbelHu, line 180: what is the change in this part? If it is only for format, please change it later in another PR.

e2e/windows/e2e-scenario.sh
```diff
@@ -125,8 +129,21 @@ tee $SCENARIO_NAME-vmss.json > /dev/null <<EOF
...
+'del(.KubeletConfig."--pod-manifest-path") | del(.KubeletConfig."--pod-max-pids") | 
+del(.KubeletConfig."--protect-kernel-defaults") | del(.KubeletConfig."--tls-cert-file") | 
+del(.KubeletConfig."--tls-private-key-file") | del(.KubeletConfig."--network-plugin") | 
+.ContainerService.properties.certificateProfile += {"clientCertificate": $clientCrt} | 
+.PrimaryScaleSetName=$vmssName' nodebootstrapping_config.json > $WINDOWS_E2E_IMAGE-nodebootstrapping_config_for_windows.json
+
+# Produce a finalized managed cluster config (nbc?) given 
+# 1. the node bootstrapping config from e2e-starter (kubeconfig)
+# 2. the vmss cse deployment config generated from this file (windows package \in k8s components)
+jq -s '.[0] * .[1]' $WINDOWS_E2E_IMAGE-nodebootstrapping_config_for_windows.json \
```
AbelHu, line 145: what is the change in this part? If it is only for format, please change it later in another PR.

e2e/windows/e2e-create-windows-nodepool.sh
junjiezhang1997: This won't impact existing pipelines/tests. v2 is used to avoid impacting previous pipelines and current "$K8S_VERSION" can help to avoid existing pipelines.

e2e/windows/e2e-scenario.sh
junjiezhang1997: Do you mean using a meaningful name for DEPLOYMENT_VMSS_NAME? We randomly generate the DEPLOYMENT_VMSS_NAME to avoid duplicate names between different pipelines.

e2e/windows/scenarios/windows/property-windows-template.json
wanqingfu: @junjiezhang1997 should we add a pipeline variable?

e2e/windows/e2e-scenario.sh
wanqingfu: Removed the "--network-plugin" from "KubeletConfig" in the json. 

A little bit of context:
Found that if the version in the `K8sComponents/WindowsPackageURL` is below 1.24.0 (e.g. 1.23.12, 1.23.15) in `property-windows-template.json`, the deployment node can join the test cluster. Otherwise the pipeline fails and the node cannot join the cluster. (e.g. 1.24.9, 1.24.10)

@junjiezhang1997 later helped me with this last week. I logged onto the deployment VMSS node and checked the `kubelet.err.log` which says `Error: failed to parse kubelet flag: unknown flag: --network-plugin
Usage:
  kubelet [flags]` This was apparently different from @junjiezhang1997 found on a normal `1.24.10` deployment node. There was no such tag in `CustomDataSetupScript.ps1` or `kubelet.err.log` he found.

So we decided to remove it from the json files and confirmed it worked for versions both below and beyond 1.24.0. (*by checking the k8s version of the deployment node $e.g. winXXXXX$ that joined the test cluster on the portal*)

e2e/windows/e2e-scenario.sh
wanqingfu: I have rolled back to the original version. For the scope of this PR reformatting is not necessary and hence reverted.

It is perhaps recommended to trim each line to have approximately 100 characters for better readability. (for example VS code has a plugin to do automatic trimming for C++)

.pipelines/templates/e2e-windows-template.yaml
wanqingfu: Storage account will be created only once if the RG does not exist for the k8s version e.g. it has been GCed. (if boolean indicator `$create_cluster` is true in line 88, `e2e-starter.sh`)

In subsequent testing on the same k8s version, SA won't be created assuming the RG is still there.

Or by saying existing storage accounts are you referring to something else more specifically?

wanqingfu: Used shared storage account in the VHD image release sub now.

e2e/windows/scenarios/windows/property-windows-template.json
wanqingfu: Added `WINDOWS_PACKAGE_URL` (see the latest file change)

e2e/windows/scenarios/windows/property-windows-template.json
```diff
@@ -3,7 +3,7 @@
     "properties": {
       "orchestratorProfile": {
         "orchestratorType": "Kubernetes",
-        "orchestratorVersion": "1.23.12",
+        "orchestratorVersion": "${orchestratorVersion}",
```
junjiezhang1997, line 6: Two questions here:
1. When we set "WindowsPackageURL" manually, the orchestractorVersion may be wrong here?
2. Have you checked the kubernetes version on Azure Portal when use hotfix package url?

e2e/windows/e2e-helper.sh
```diff
@@ -16,11 +16,23 @@ exec_on_host() {
...
-    account_key=$(az storage account keys list --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --query "[0].value")
-    az storage container create -n "tmp" --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --account-key $account_key
+    # check if the storage container exists and create one if not
+    exists=$(az storage container exists --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --name $WINDOWS_E2E_STORAGE_CONTAINER)
+    if [[ $exists == *false* ]]; then
+        az storage container create -n $WINDOWS_E2E_STORAGE_CONTAINER --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY
+        echo "Created storage container $WINDOWS_E2E_STORAGE_CONTAINER in $STORAGE_ACCOUNT_NAME"
+    else
+        # check if the storage container is empty and delete the blobs within one by one if not
+        blob_list=$(az storage blob list --account-name $STORAGE_ACCOUNT_NAME --container-name $WINDOWS_E2E_STORAGE_CONTAINER --account-key $MAPPED_ACCOUNT_KEY -o json | jq -r '.[] | .name')
```
junjiezhang1997, line 28: It seems that this part of code has not been tested in your shared pipeline links.
To test this part of code, I think you should select a kubernetes version which already has corresponding storage container, then delete the resource group and rerun the test pipeline to create new resource group.

e2e/windows/scenarios/windows/property-windows-template.json
wanqingfu: 1. Based on my experience, the kubernetes version on Azure Portal  (or `kubectl get nodes`) comes from the version in the windows package url. Not sure if this `orchestratorVersion` is needed.
2. No I haven't yet. I only checked the versions for non-hotfix which are the same as that in the package url even if the `orchestratorVersion` is different. Will double check on this.

If this is not necessary proven by experiments I think I can just remove this field? @junjiezhang1997

wanqingfu: For the pipeline using 1.22.1 hotfix, the version returned by `kubectl get nodes` is 1.22.1 as seen in the pipeline logs.
Same for 1.25.2.

e2e/windows/e2e-starter.sh
```diff
@@ -107,9 +109,11 @@ getMSIResourceID
 addJsonToFile "mcRGName" $MC_RESOURCE_GROUP_NAME
 addJsonToFile "clusterID" $CLUSTER_ID
 addJsonToFile "subID" $SUBSCRIPTION_ID
+addJsonToFile "orchestratorVersion" $KUBERNETES_VERSION
```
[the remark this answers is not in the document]
wanqingfu: That was for `percluster_template.json` earlier. But since it's not changed thinking I can safely remove this line. *Pending*.

e2e/windows/e2e-helper.sh
wanqingfu: The RG of 1.24.9 has already been deleted. Requeued the pipeline and if you check the logs it says 

`Deleted blob aks-nodepool1-XXXXXXXX-vmss-linux-file.zip from storage container akswinstore2019-1249 in the storage account akswinstorageaccount`

Same goes for 1.26.3. 
`Deleted blob aks-nodepool1-XXXXXXXX-vmss-linux-file.zip from storage container akswinstore2019-1263 in the storage account akswinstorageaccount`

e2e/windows/e2e-helper.sh
```diff
@@ -16,11 +16,23 @@ exec_on_host() {
     kubectl exec $(kubectl get pod -l app=debug -o jsonpath="{.items[0].metadata.name}") -- bash -c "nsenter -t 1 -m bash -c \"$1\"" > $2
 }
 
-create_storage_account() {
+create_storage_container() {
```
AbelHu, line 19: What would happen if two pipelines are queued with the same k8s version?

e2e/windows/e2e-scenario.sh
```diff
@@ -95,6 +96,13 @@ fi
 log "Upload cse packages done"
 
 log "Scenario is $SCENARIO_NAME"
+export orchestratorVersion=$KUBERNETES_VERSION
+export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${orchestratorVersion}/windowszip/v${orchestratorVersion}-1int.zip"
+if [[ -n "$WINDOWS_PACKAGE_URL" ]]; then
+    export windowsPackageURL="$WINDOWS_PACKAGE_URL"
```
AbelHu, line 102: log for debug

e2e/windows/e2e-scenario.sh
```diff
@@ -167,6 +177,7 @@ set +e
 az deployment group create --resource-group $MC_RESOURCE_GROUP_NAME \
          --template-file $DEPLOYMENT_VMSS_NAME-deployment.json || retval=$?
 set -e
+log "Group deployment creation completed"
```
AbelHu, line 180: Thanks for adding the log. But "Group deployment creation" is confused. Could you help to provide a better log, for example, 'Deploying Windows VMSS' by reference line 183?

e2e/windows/e2e-starter.sh
```diff
@@ -53,7 +54,7 @@ if [ "$create_cluster" == "true" ]; then
     clusterCreateStartTime=$(date +%s)
     retval=0
     
-    az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys --network-plugin azure -ojson || retval=$?
+    az aks create -g $RESOURCE_GROUP_NAME -n $CLUSTER_NAME --node-count 1 --generate-ssh-keys --network-plugin azure --kubernetes-version $KUBERNETES_VERSION -ojson || retval=$?
```
AbelHu, line 57: Would it still succeed if $KUBERNETES_VERSION is empty for the default test?

e2e/windows/nodebootstrapping_template.json
```diff
@@ -95,7 +95,7 @@
             "containerdDownloadURLBase": "https://storage.googleapis.com/cri-containerd-release/",
             "csiProxyDownloadURL": "https://acs-mirror.azureedge.net/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz",
             "windowsProvisioningScriptsPackageURL": "https://acs-mirror.azureedge.net/aks-engine/windows/provisioning/signedscripts-v0.2.2.zip",
-            "windowsPauseImageURL": "mcr.microsoft.com/oss/kubernetes/pause:1.4.0",
+            "windowsPauseImageURL": "mcr.microsoft.com/oss/kubernetes/pause:3.9",
```
AbelHu, line 98: May I know the reason to change the pause image version in this PR?

e2e/windows/scenarios/windows/property-windows-template.json
junjiezhang1997: Another problem needs to be considered is that your resource group name is named by KUBERNETES _VERSION, so maybe you need to clarify the resource group name with the hotfix name to avoid ambiguity?

e2e/windows/e2e-helper.sh
wanqingfu: Only the first pipeline will create a storage container because `$create_cluster = true`. Other pipelines will only download linux file from storage account. (tested a lot of times) @junjiezhang1997 maybe you can clarify on this?

e2e/windows/e2e-scenario.sh
wanqingfu: They are already in pipeline logs. Example:

`export orchestratorVersion=1.26.3
+ orchestratorVersion=1.26.3
+ export windowsPackageURL=https://acs-mirror.azureedge.net/kubernetes/v1.26.3/windowszip/v1.26.3-1int.zip
+ windowsPackageURL=https://acs-mirror.azureedge.net/kubernetes/v1.26.3/windowszip/v1.26.3-1int.zip`

e2e/windows/e2e-starter.sh
wanqingfu: Default $KUBERNETES_VERSION was set to 1.24.9 by @junjiezhang1997?

e2e/windows/nodebootstrapping_template.json
wanqingfu: When the node couldn't join the cluster for 1.24.0+ @junjiezhang1997 and I suspected reasons with the URL versions so that comes the changes of versions. But since we later found it was caused by deprecated flag `network-plugin` for k8s beyond 1.24 the changes are no longer needed.

Have reverted all the changes related to version numbers in the json files and confirmed the pipeline succeeded.

e2e/windows/e2e-helper.sh
junjiezhang1997: There are three scenarios:
1. resource group and container both do not exist: one pipeline will create the resource group and container, then upload linux files; the other pipeline will wait until linux files are uploaded, then download these files.
2. resource group does not exist, but container exists: one pipeline will create the resource group and clean up files in the container, then upload new linux files; the other pipeline will wait to download files.
3. resource group and container both exist: two pipelines will just go to download linux files from the container and go on testing steps.

e2e/windows/e2e-starter.sh
wanqingfu: Removed the specification. The linux cluster version goes with $KUBERNETES_VERSION defaulting to 1.24.9. The kubernetes version of windows deployment node originates from the version in the package url.

For example if using 1.25.2-hotfix for the windows package url, the deployment node's version is ... as seen on the portal.
`kubectl get nodes` returns
... (*pending*)

e2e/windows/scenarios/windows/property-windows-template.json
wanqingfu: Summary on the discussion with @junjiezhang1997 
- the deployment node that joined the test-cluster is expected to have k8s version equals to whatever comes from the `windowsPackageURL`.
- the resource group name contains this windows package version and an optional "hotfix" to disambiguate. $e.g.$ testing `1.25.2-hotfix` the RG ends with `1252-hotfix`.
- the k8s version of the linux `test-cluster` is `KUBERNETES_VERSION` - a pipeline UI variable added by @junjiezhang1997 and it defaults to `1.24.9`.

e2e/windows/nodebootstrapping_template.json
```diff
@@ -156,7 +156,6 @@
         "--cloud-provider": "azure",
         "--cluster-dns": "10.0.0.10",
         "--cluster-domain": "cluster.local",
-        "--dynamic-config-dir": "/var/lib/kubelet",
```
wanqingfu, line 159: Removed all things around dynamic kubelet config as it's deprecated. Logged onto v1.26.0 (or v1.26.3) failed deployment node and checked the `kubelet.err.log` saying `failed to set feature gates from initial flags-based : unrecognized feature gate: DynamicKubeletConfig`. Cse validation on these versions would fail if `DynamicKubeletConfig=false` is still there in the json file. 

Toyed around a bit with this and I decided to simply remove every occurence since it would still be compatible with earlier versions confirmed empirically.

FYI @junjiezhang1997

e2e/windows/scenarios/windows/property-windows-template.json
```diff
@@ -92,12 +92,11 @@
     "--cloud-provider": "external",
     "--enforce-node-allocatable": "\"\"\"\"",
     "--eviction-hard": "\"\"\"\"",
-    "--feature-gates": "CSIMigration=true,CSIMigrationAzureDisk=true,CSIMigrationAzureFile=true,DelegateFSGroupToCSIDriver=true,DynamicKubeletConfig=false",
```
wanqingfu, line 95: Same as above. This is deprecated for latest versions. See `baker.go` for reference. And since I found a lot of issues with this trying to validate $e.g.$ v1.25.6, v1.26.0, v1.26.3 cse packages I've determined to simply remove this as it's not used anyhow. FYI @junjiezhang1997

e2e/windows/e2e-starter.sh
```diff
@@ -107,9 +112,11 @@ getMSIResourceID
 addJsonToFile "mcRGName" $MC_RESOURCE_GROUP_NAME
 addJsonToFile "clusterID" $CLUSTER_ID
 addJsonToFile "subID" $SUBSCRIPTION_ID
+addJsonToFile "orchestratorVersion" $KUBERNETES_VERSION
```
wanqingfu, line 115: The `orchestratorVersion` in `property-windows-template.json` is the version for the ARM template. I do have seen some transient windows vmss deployment errors using `v1.23.12` with cse package v1.25.6+. 

I think we can use $KUBERNETES_VERSION (or the windows package version more precisely) for the windows vmss deployment to reduce the chance of deployment failure. @junjiezhang1997 ?

e2e/windows/e2e-scenario.sh
```diff
@@ -95,6 +99,11 @@ fi
 log "Upload cse packages done"
 
 log "Scenario is $SCENARIO_NAME"
+log "Windows package url is $windowsPackageURL"
+log "Windows package version is $WINDOWS_PACKAGE_VERSION"
+
+# Generate vmss cse deployment config for windows nodepool testing
+export orchestratorVersion=$WINDOWS_PACKAGE_VERSION
```
wanqingfu, line 106: The `orchestratorVersion` in `property-windows-template.json` is the version for the ARM template. I do have seen some transient windows vmss deployment errors using `v1.23.12` with cse package v1.25.6+. 

I think we can use the windows package version for the windows vmss deployment to reduce the chance of deployment failure. @junjiezhang1997 ?

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -31,13 +31,21 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_E2E_OSSKU=${{ parameters.windowsOSSKU }}
         export WINDOWS_E2E_VMSIZE=${{ parameters.windowsVMSize }}
-        export WINDOWS_E2E_STORAGE_ACCOUNT=${{ parameters.storageAccount }}
+        export windowsPackageURL="https://acs-mirror.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/windowszip/v${KUBERNETES_VERSION}-1int.zip"
+        if [[ -n "$WINDOWS_PACKAGE_URL" ]]; then
+          export windowsPackageURL="$WINDOWS_PACKAGE_URL"
+        fi
+        export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
```
AbelHu, line 38: I think that we should not use the version from the package url. By design, the version in the package url should always match the k8s version set in the pipeline variable. 
I think that we cannot test the hotfix without a workaround since all supported k8s versions do not have a hotfix package. It can be ignored since you have validated 1.25.2-hotfix package

e2e/windows/e2e-helper.sh
```diff
@@ -16,11 +16,23 @@ exec_on_host() {
...
     set +x
-    az storage account create -n $WINDOWS_E2E_STORAGE_ACCOUNT -g $RESOURCE_GROUP_NAME -l $LOCATION
-    account_key=$(az storage account keys list --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --query "[0].value")
-    az storage container create -n "tmp" --account-name $WINDOWS_E2E_STORAGE_ACCOUNT --account-key $account_key
+    # check if the storage container exists and create one if not
+    exists=$(az storage container exists --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --name $WINDOWS_E2E_STORAGE_CONTAINER)
```
AbelHu, line 22: Will the cluster for the specified k8s version be kept for 3 days as the current design? Who will delete the container created here and when?

wanqingfu: The current design will reclaim the resource group including the cluster in about less than 1 day (fewer than 3 days) by my observation. @junjiezhang1997 is more familiar with this and maybe can clarify?

junjiezhang1997: The container will not be deleted. The current design is to clean up files in the container every time new cluster is created.

.pipelines/templates/e2e-windows-template.yaml
wanqingfu: Are you saying we should just provide the k8s version set in the pipeline variable and remove the package url variable?

Or you are suggesting keeping the url variable 

 `$KUBENERTES_VERSION=1.24.9`, package = 1.25.2-hotfix (*workaround??*)

e2e/windows/e2e-create-windows-nodepool.sh
```diff
@@ -6,8 +6,10 @@ source e2e-helper.sh
 
 log "Starting to create windows nodepool"
 
-if [[ "$RESOURCE_GROUP_NAME" == *"windows"*  ]]; then
-    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
+if echo "$windowsPackageURL" | grep -q "hotfix"; then
+    RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME-$WINDOWS_E2E_IMAGE-$K8S_VERSION-h"
```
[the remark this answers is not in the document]
wanqingfu: @junjiezhang1997 previously suggested to disambiguate.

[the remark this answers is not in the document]
wanqingfu: Will there be a `1.25.2-hotfix` and `1.25.2` in the future? `1.25.2` is just an example.

.pipelines/templates/e2e-windows-template.yaml
wanqingfu: Added check on the versions before setting up the test cluster. Windows package version (extracted from the non-empty) url should be the same as the KUBERNETES_VERSION pipeline variable.

e2e/windows/e2e-scenario.sh
wanqingfu: @junjiezhang1997 found that the $orchestratorVersion in `property-windows-template` is used to decide if it's necessary to remove the dynamic kubelet config flag that's not supported by 1.24.0+ in `baker.go`.

If 1.11 < version < 1.24, `DynamicKubeletConfig=False` in feature gates when validating windows nodebootstrapping;
if version > 1.24, this flag is simply removed.

e2e/windows/e2e-create-windows-nodepool.sh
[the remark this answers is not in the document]
AbelHu: Sorry. I cannot get your point. Is my understanding right that there is no impact even if both hotfix and non-hotfix exist?

e2e/windows/e2e-helper.sh
AbelHu: Will the container be still there if the k8s version is not supported.

e2e/windows/e2e-create-windows-nodepool.sh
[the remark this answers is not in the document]
wanqingfu: yes. No impact. If they coexist the test clusters will be in different resource groups `1252` and `1252-h`. Just adding an "h" to differentiate. Nothing else.

e2e/windows/e2e-helper.sh
wanqingfu: It will I believe since it's stored in `akswinstorageaccount` in the vhd release sub that does not do cleaning. But I think it's out of the scope of this PR. Something like the backfill deletion I was and @junjiezhang1997 is working on?

e2e/windows/e2e-create-windows-nodepool.sh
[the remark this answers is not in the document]
wanqingfu: If there are two parallel pipelines. 
One: `1.25.2` the test cluster, windows nodepool will be in the RG that ends with `1252`. The windows vmss node will join the test cluster in this `1252` RG.
Another: `1.25.2-hotfix`. They will be in the RG that ends with `1252-h`. The windows vmss node will join the test cluster in this `1252-h` RG.

The current GC period is slightly less than 1 day in my observation and I ran sample scripts to clean up a dozen of RGs created around the same time so I think it should be fine.

e2e/windows/e2e-helper.sh
AbelHu: I think that you need to clean up it with adding the code in backfill deletion since these containers are created by this PR

e2e/windows/e2e-create-windows-nodepool.sh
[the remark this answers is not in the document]
AbelHu: Please use the same rg since they can use the same cluster. We can reduce the cost and avoid possible throttling in future.

e2e/windows/e2e-helper.sh
AbelHu: You have two options: 1. clean up the containers in this PR; 2. create a new task and do it in another pr

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -71,6 +83,18 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_DISTRO=${{ parameters.windowsDistro }}
         export WINDOWS_NODE_IMAGE_VERSION=${{ parameters.windowsNodeImageVersion }}
+        if [[ -n "${WINDOWS_PACKAGE_URL}" ]]; then
+          export windowsPackageURL=$WINDOWS_PACKAGE_URL
+          export WINDOWS_PACKAGE_VERSION=$(echo $windowsPackageURL | grep -oP '(?<=/v)\d+\.\d+\.\d+' | head -1)
+          if [[ "${WINDOWS_PACKAGE_VERSION}" != "${KUBERNETES_VERSION}" ]]; then
```
junjiezhang1997, line 89: I think the comparison of WINDOWS_PACKAGE_VERSION and KUBERNETES_VERSION here is not necessary, since the pipeline will fail at the previous step. You could test the invalid scenario in one pipeline.

wanqingfu: I'll remove

e2e/windows/e2e-helper.sh
wanqingfu: Go with 2. Created a task to do backfill and commenting/formatting/documenting.

e2e/windows/e2e-starter.sh
```diff
@@ -8,7 +8,7 @@ log "Starting e2e tests"
 
 # Create a resource group for the cluster
 log "Creating resource group"
-RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME"-"$WINDOWS_E2E_IMAGE"-v2
+RESOURCE_GROUP_NAME="$RESOURCE_GROUP_NAME-$WINDOWS_E2E_IMAGE-$K8S_VERSION"
```
AbelHu, line 11: Per the offline discussion, we should not set the default k8s version in pipeline variables so it should be empty by default.

wanqingfu: Made changes.

k8s version in pipeline variables defaults to empty "".

If windows package url is set, `KUBERNETES_VERSION` must be set otherwise it returns error. This is to avoid using default version like `1.25.6` if intending to test a windows package version $e.g.$ `1.26.0`. Then we check if the package version is the same as the k8s version in pipeline variable. If it differs we return in invalid scenario.

If windows package url is not set, `KUBERNETES_VERSION` will be the windows package version. And if it's empty we populate it with the default version returned by the command `1.25.6` now as of Apr 20, 2023.

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -31,7 +31,27 @@ jobs:
...
+            echo "You must set KUBERNETES_VERSION when setting WINDOWS_PACKAGE_URL"
+            exit 1
+          fi
+          if [[ "${WINDOWS_PACKAGE_VERSION}" != "${KUBERNETES_VERSION}" ]]; then
+            echo "Windows package version $WINDOWS_PACKAGE_VERSION does not match with kubernetes version $KUBERNETES_VERSION. Invalid test scenario."
+            exit 1
+          fi
+        else
+          if [[ -z "${KUBERNETES_VERSION}" ]]; then
+            KUBERNETES_VERSION=$(az aks get-versions -l eastus --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv)
```
AbelHu, line 47: Hard code the region

wanqingfu: ?

AbelHu (approved):

junjiezhang1997 (approved):

PR #2973, opened 2023-04-04
chore: ensure vmss deletion during abe2e

cameronmeissner:
**What type of PR is this?**

/kind bug

make sure vmss resources are deleted after abe2e runs even if the job is cancelled during go code execution

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.pipelines/e2e.yaml
```diff
@@ -27,11 +27,16 @@ jobs:
       cd e2e
       go test -timeout 30m -v -run Test_All ./...
     displayName: Run AgentBaker E2E
-  - publish: $(System.DefaultWorkingDirectory)/e2e/vm-logs
-    artifact: vm-logs
-    condition: always()
-  - publish: $(System.DefaultWorkingDirectory)/e2e/cluster-parameters
```
cameronmeissner, line 33: was planning on removing cluster param uploads during that other PR, doing it here instead

.pipelines/e2e.yaml
```diff
@@ -27,11 +27,22 @@ jobs:
...
+    artifact: scenario-logs
     condition: always()
-  - publish: $(System.DefaultWorkingDirectory)/e2e/cluster-parameters
-    artifact: cluster-parameters
+  - bash: |
+      set -x
+
+      vmssResourceIds=""
+      for vmssModel in e2e/scenario-logs/*/vmss.json; do
+        resourceId=$(cat ${vmssModel} | jq -r .id)
```
alexeldeib, line 38: can we just log the list of IDs instead of the full model?

alexeldeib (approved):

.pipelines/e2e.yaml
cameronmeissner: aahh i completely forgot to do this, good catch

PR #2974, opened 2023-04-04
fix: ensure e2e debug pods are scheduled onto aks nodepool

cameronmeissner:
**What type of PR is this?**

/kind bug

ensure abe2e debug pods are scheduled onto the correct nodepool

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/e2e/deploy.yaml
+++ b/e2e/deploy.yaml
@@ -14,14 +14,10 @@ spec:
       labels:
         app: *name
     spec:
-      # nodeSelector:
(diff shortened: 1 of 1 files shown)


cameronmeissner:
will follow-up with changing getDebugPodName once this change has propagated and all nodes are running the aks nodepool as expected

juan-lee (approved):

PR #2982, opened 2023-04-05
feat: add cluster config scenario support to abe2e

cameronmeissner:
**What type of PR is this?**

/kind feature

allows for abe2e scenarios based on different cluster configurations (such as network plugin kubenet vs. azure)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/cluster.go
```diff
@@ -132,3 +144,181 @@ func getClusterSubnetID(ctx context.Context, cloud *azureClient, location, resou
...
+
+		*clusters = append(*clusters, cluster)
+
+		chosenCluster = cluster
+		chosenKubeClient = kube
+		chosenSubnetID = subnetID
+		chosenClusterParams = clusterParams
+	} else {
+		for _, cluster := range viableClusters {
+			if _, err := ensureCluster(ctx, t, cloud, suiteConfig.location, suiteConfig.resourceGroupName, cluster, false); err != nil {
```
alexeldeib, line 229: this fork feels a bit like a code smell (calling ensureCluster in both code paths with isNew true/false but it already exists in one case)

can you share the thinking here?

e2e/scenario/clusterconfiguration.go
```diff
@@ -0,0 +1,23 @@
+package scenario
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+)
+
+// Selectors
+
+func NetworkPluginKubenetSelector(cluster *armcontainerservice.ManagedCluster) bool {
```
alexeldeib, line 10: fancy

e2e/scenario/init.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+	return table
+}
+
+// Is called internally by the scenario package to get each scenario's respective config as one long slice.
+// To add a sceneario, implement a new function in a separate file that returns a *Scenario and add
+// its return value to the slice returned by this function.
+func scenarios() []*Scenario {
+	return []*Scenario{
+		base(),
```
alexeldeib, line 22: nice

e2e/suite_test.go
```diff
@@ -26,192 +27,195 @@ func Test_All(t *testing.T) {
...
+	suiteConfig *suiteConfig,
+	scenario *scenario.Scenario,
+	chosenCluster *armcontainerservice.ManagedCluster,
+	nbc *datamodel.NodeBootstrappingConfiguration,
+	subnetID string,
+	publicKeyBytes []byte) (*armcompute.VirtualMachineScaleSet, func(), error) {
+	nodeBootstrapping := mustGetNodeBootstrapping(ctx, t, nbc)
+
+	vmssName := fmt.Sprintf("abtest%s", randomLowercaseString(r, 4))
+	t.Logf("vmss name: %q", vmssName)
```
alexeldeib, line 158: nit: I suggest avoiding using the test helpers explicitly in the deeper nested logic, use fmt.Errorf and function wrappers to return errors up the stack to the main test func, and we can start plumbing a propper logger down the stack (in func signatures)

this will help @utheman with integrationt test coverage as well

e2e/cluster.go
cameronmeissner: the original thinking was that ensureCluster could be used both to create a new cluster from scratch, and make sure that an existing cluster is in a usable state (MC resource group is there, provisioning state isn't "Failed", etc.) - I guess the smell comes from the fact that this function is probably doing too much without being more clear.

that boolean argument is really just there to short-circuit between "cluster already exists" and "cluster doesn't exist" scenarios, which I guess goes further to suggest that we should split that function up

alexeldeib: I think the confusing part is maybe the fact that we really have 3 states -- good, dirty, and non-existent.

I'm okay to merge/iterate but would like to break it up a bit more maybe

e2e/suite_test.go
cameronmeissner: removed from here - will be nice to have a separate logger implementation to use so we don't have to plumb 't' around everywhere

e2e/cluster.go
cameronmeissner: did some splitting

e2e/suite_test.go
alexeldeib: you can fmt.Println already or log.Println, t.Log is nothing fancy

t.Error/Fatal are just println/and t.fail/failnow()

alexeldeib (approved):

e2e/suite_test.go
```diff
@@ -26,192 +27,198 @@ func Test_All(t *testing.T) {
...
+	ctx context.Context,
+	t *testing.T,
+	r *mrand.Rand,
+	cloud *azureClient,
+	kube *kubeclient,
+	suiteConfig *suiteConfig,
+	scenario *scenario.Scenario,
+	chosenCluster *armcontainerservice.ManagedCluster,
+	nbc *datamodel.NodeBootstrappingConfiguration,
+	subnetID, caseLogsDir string) {
```
juan-lee, line 99: too many parameters?

e2e/suite_test.go
```diff
@@ -26,192 +27,198 @@ func Test_All(t *testing.T) {
...
+	ctx context.Context,
+	t *testing.T,
+	r *mrand.Rand,
+	cloud *azureClient,
+	suiteConfig *suiteConfig,
+	scenario *scenario.Scenario,
+	chosenCluster *armcontainerservice.ManagedCluster,
+	nbc *datamodel.NodeBootstrappingConfiguration,
+	subnetID string,
+	publicKeyBytes []byte) (*armcompute.VirtualMachineScaleSet, func(), error) {
```
juan-lee, line 154: too many parameters

e2e/vmss.go
```diff
@@ -48,16 +49,16 @@ func getNewRSAKeyPair(r *mrand.Rand) (privatePEMBytes []byte, publicKeyBytes []b
 	return
 }
 
-func createVMSSWithPayload(ctx context.Context, publicKeyBytes []byte, cloud *azureClient, location, name, subnetID, customData, cseCmd string, mutator func(*armcompute.VirtualMachineScaleSet)) (*armcompute.VirtualMachineScaleSet, error) {
-	model := getBaseVMSSModel(name, location, subnetID, string(publicKeyBytes), customData, cseCmd)
+func createVMSSWithPayload(ctx context.Context, publicKeyBytes []byte, cloud *azureClient, location, mcResourceGroupName, name, subnetID, customData, cseCmd string, mutator func(*armcompute.VirtualMachineScaleSet)) (*armcompute.VirtualMachineScaleSet, error) {
```
juan-lee, line 52: too many parameters

e2e/suite_test.go
```diff
@@ -26,192 +27,198 @@ func Test_All(t *testing.T) {
...
+
+	vmssModel, cleanupVMSS, err := bootstrapVMSS(ctx, t, r, cloud, suiteConfig, scenario, chosenCluster, nbc, subnetID, publicKeyBytes)
+	defer cleanupVMSS()
+	isCSEError := isVMExtensionProvisioningError(err)
+	vmssSucceeded := true
+	if err != nil {
+		vmssSucceeded = false
+		if isCSEError {
+			t.Error("VM was unable to be provisioned due to a CSE error, will still atempt to extract provisioning logs...", err)
+		} else {
```
juan-lee, line 114: You could avoid else here if you did `if !isCSEError { t.Fatal...` first.

e2e/scenario/types.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+)
+
+type ScenarioTable map[string]*Scenario
+
+type Scenario struct {
```
juan-lee, line 14: https://google.github.io/styleguide/go/decisions#repetition

juan-lee (changes requested):

e2e/suite_test.go
cameronmeissner: added a new struct to deal with this

juan-lee (approved):

PR #2988, opened 2023-04-05
fix: Add new distros for Edge Zones 2204 images

yunrugemicrosoft:
**What type of PR is this?**
/kind bug

Adding two new distros for Edge Zone SIG 2204 Gen1 and Gen2 images.

Fixes #

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


UtheMan (approved):

alexeldeib (approved):
I think just changing the casing and leaving them as consts/var was fine? 

this works though :)

PR #3003, opened 2023-04-07
fix: minor version runc bump to 1.1.5, remove multiple version usage for CVE fixes and new features

ganeshkumar5699:
**What type of PR is this?**
/kind feature

- Consolidate to a single runc version (vs 3 versions) and use same version for AMD and ARM
- Prevent caching older versions with CVEs (since older versions are unsupported)
- Upgrade by a runc minor version.
- Handle the new format of runc versions (with azure and Ubuntu version suffix)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
runc version for all new VHDs has been bumped to 1.1.5, from 1.0.3.
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -153,13 +153,10 @@ fi
...
 
-## for ubuntu-based images, cache multiple versions of runc
 if [[ $OS == $UBUNTU_OS_NAME ]]; then
-  RUNC_VERSIONS="
-  1.0.0-rc92
-  1.0.0-rc95
-  1.0.3
-  "
+  
+  RUNC_VERSIONS="1.1.5+azure-ubuntu-${UBUNTU_VERSION_CLEANED}u1_${CPU_ARCH}"
```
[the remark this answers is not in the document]
alexeldeib: no sorry -- I just mean for the logging in the file, so we don't have to hardcode it. e.g. `runc --version >> vhdfile`

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -247,7 +247,7 @@ ensureRunc() {
 
     TARGET_VERSION=${1:-""}
     if [[ -z ${TARGET_VERSION} ]]; then
-        TARGET_VERSION="1.0.3"
+        TARGET_VERSION="1.1.5+azure-ubuntu${UBUNTU_RELEASE}u1"
```
alexeldeib, line 250: does the check on L268 work with this version? 

the check below uses the runc version itself, not the apt package version.

alexeldeib: also L260?

vhdbuilder/packer/install-dependencies.sh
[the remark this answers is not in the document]
ganeshkumar5699: Got it. This is being logged on L153 and 154 already: 

```
INSTALLED_RUNC_VERSION=$(runc --version | head -n1 | sed 's/runc version //')
echo "  - runc version ${INSTALLED_RUNC_VERSION}" >> ${VHD_LOGS_FILEPATH}
```

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumar5699: L260 is a potentially unnecessary check, since previously the runc version and apt package versions were the same. So having that to prevent older versions being used is right.

Keeping that check for now, since it's a large jump in versions.

ganeshkumar5699: Will update check in L268 though

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -265,20 +259,22 @@ ensureRunc() {
...
-        RUNC_DEB_PATTERN="moby-runc_${TARGET_VERSION/-/\~}+azure-*_${CPU_ARCH}.deb"
+        RUNC_DEB_PATTERN="moby-runc_*.deb"
         RUNC_DEB_FILE=$(find ${RUNC_DOWNLOADS_DIR} -type f -iname "${RUNC_DEB_PATTERN}" | sort -V | tail -n1)
         if [[ -f "${RUNC_DEB_FILE}" ]]; then
             installDebPackageFromFile ${RUNC_DEB_FILE} || exit $ERR_RUNC_INSTALL_TIMEOUT
             return 0
         fi
     fi
-    apt_get_install 20 30 120 moby-runc=${TARGET_VERSION/-/\~}* --allow-downgrades || exit $ERR_RUNC_INSTALL_TIMEOUT
+    apt_get_install 20 30 120 moby-runc=${TARGET_VERSION} --allow-downgrades || exit $ERR_RUNC_INSTALL_TIMEOUT
```
alexeldeib, line 277: curious to validate this really works for all skus, we can find out in test all vhd build

alexeldeib (approved):
lgtm, if you didn't already, let's run a test build on all skus for validation

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -153,20 +153,10 @@ fi
...
-  "
-  if [[ $(isARM64) == 1 ]]; then
-    # RUNC versions of 1.0.3 later might not be available in Ubuntu AMD64/ARM64 repo at the same time
-    # so use different version set for different arch to avoid affecting each other during VHD build
-    RUNC_VERSIONS="
-    1.0.3
-    "
-  fi
+  
+  RUNC_VERSIONS="1.1.5+azure-ubuntu-${UBUNTU_RELEASE}u1_${CPU_ARCH}"
```
alexeldeib, line 158: isn't this entirely unused because it's installed in standalone containerd ?

ganeshkumar5699:
> lgtm, if you didn't already, let's run a test build on all skus for validation

Ran it already for all SKUs, with a slightly earlier commit, and it succeeded: 

https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=71578821&view=results

vhdbuilder/packer/install-dependencies.sh
ganeshkumar5699: it's used for downloading, and then installation is triggered through standalone containerd

ganeshkumar5699: Good point. That's called before these downloads, so it should not be necessary. Removed it and got green builds again (in test all VHDs pipeline).

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumar5699: Ran the test all VHD pipelines, and got green builds for all (mariner doesn't use this section btw)

PR #3004, opened 2023-04-07
fix: remove docker installation logic and checks in VHD build, remove unused 16.04 post VHD build test

ganeshkumar5699:
**What type of PR is this?**
/kind cleanup

- Removes unused docker specific installation checks/steps during VHD build
- Prevents the need to update components that are specific to Docker, and makes installation steps more readable
- Clean up unused 16.04 post VHD build check

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -189,23 +185,14 @@ if [[ $OS == $UBUNTU_OS_NAME && $(isARM64) != 1 ]]; then  # no ARM64 SKU with GP
...
+    exit 1
   fi
 fi
 
 if [ "${CONTAINER_RUNTIME:=}" == "containerd" ]; then
     systemctlEnableAndStart containerd-monitor.timer || exit $ERR_SYSTEMCTL_START_FAIL
-else
-    systemctlEnableAndStart docker-monitor.timer || exit $ERR_SYSTEMCTL_START_FAIL
 fi
-
```
cameronmeissner, line 208: maybe also fail hard here if container runtime isn't containerd?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -269,10 +256,7 @@ watcherStaticImg=${watcherBaseImg//\*/static}
 # can't use cliTool because crictl doesn't support retagging.
 if [[ "${CONTAINER_RUNTIME}" == "containerd" ]]; then
     retagContainerImage "ctr" ${watcherFullImg} ${watcherStaticImg}
-else
-    retagContainerImage "docker" ${watcherFullImg} ${watcherStaticImg}
 fi
-
```
cameronmeissner, line 275: diddo

cameronmeissner (approved):

vhdbuilder/packer/install-dependencies.sh
anujmaheshwari1: following from what Cam said, we seem to be using
`echo "Unsupported container runtime"
exit 1`

for cases where cri isnt containerd. maybe make a helper function that has those 2 lines and call it in all the else statements?

ganeshkumar5699: L145 already will fail, if runtime is not containerd. Having that check earlier prevents the need from repeating it right?

alexeldeib: should we just do this once at the very beginning of the build and stop bothering to check elsewhere?

cameronmeissner: > L145 already will fail, if runtime is not containerd. Having that check earlier prevents the need from repeating it right?

I probably just missed that :D

cameronmeissner (approved):

alexeldeib (approved):
nice!

PR #3008, opened 2023-04-07
chore: substitute calls to logging functions using testing.T with std…

UtheMan:
…lib log package

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/suite_test.go
```diff
@@ -147,11 +150,11 @@ func bootstrapVMSS(ctx context.Context, t *testing.T, r *mrand.Rand, opts *scena
 	}
 
 	vmssName := fmt.Sprintf("abtest%s", randomLowercaseString(r, 4))
-	t.Logf("vmss name: %q", vmssName)
+	log.Printf("vmss name: %q", vmssName)
 
 	cleanupVMSS := func() {
-		t.Log("deleting vmss", vmssName)
-		poller, err := opts.cloud.vmssClient.BeginDelete(ctx, *opts.chosenCluster.Properties.NodeResourceGroup, vmssName, nil)
+		log.Println("deleting vmss", vmssName)
```
cameronmeissner, line 156: can we be consistent on Println with multiple args vs. Printf? I'd personally lean towards Printf but that's just me

e2e/suite_test.go
```diff
@@ -160,7 +163,7 @@ func bootstrapVMSS(ctx context.Context, t *testing.T, r *mrand.Rand, opts *scena
 		if err != nil {
 			t.Error("error polling deleting vmss", vmssName, err)
 		}
-		t.Logf("finished deleting vmss %q", vmssName)
+		log.Printf"finished deleting vmss %q", vmssName)
```
cameronmeissner, line 166: seems like a syntax error somewhere around here caused the latest run to fail

e2e/suite_test.go
UtheMan: fixed, switched log.Println calls where there were extra args to log.Printf

e2e/suite_test.go
UtheMan: yeah missing parenthesis :'(

PR #3009, opened 2023-04-07
feat: add live vm validation to abe2e

cameronmeissner:
**What type of PR is this?**

/kind feature

adds support to run arbitrary commands on the live vm after bootstrapping and make assertions on the output

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/scenario/types.go
```diff
@@ -24,6 +24,15 @@ type Config struct {
 	BootstrapConfigMutator func(*testing.T, *datamodel.NodeBootstrappingConfiguration)
 	VMConfigMutator        func(*armcompute.VirtualMachineScaleSet)
 	Validator              func(context.Context, *testing.T, *ValidationInput) error
+	LiveVMValidators       []*LiveVMValidator
```
alexeldeib, line 27: nit: would be nice if our "validator interface" was either consistently a func or array of funcs

e2e/scenario/types.go
```diff
@@ -24,6 +24,15 @@ type Config struct {
...
+}
+
+type VMCommandOutputAsserterFn func(string, string) error
```
alexeldeib, line 30: I think we can label the var names on type definitions?

e2e/validation.go
```diff
@@ -0,0 +1,117 @@
...
+	if opts.scenario.LiveVMValidators != nil {
+		validators = append(validators, opts.scenario.LiveVMValidators...)
+	}
+
+	for _, validator := range validators {
+		desc := validator.Description
+		command := validator.Command
+		log.Printf("running live VM validator: %q", desc)
+
+		execResult, err := pollExecOnVM(ctx, opts.kube, privateIP, podName, sshPrivateKey, command)
```
alexeldeib, line 47: hmm. this makes sense to me. wondering if there is a way to simplify it a bit.

it almost feels like we could abstract the "run a command on the VM" part into a struct or higher level function, and "build" the validator, then "run" (execute/assert) it

I'm okay with this, definitely a net win overall, but curious about structure

e2e/validation.go
```diff
@@ -0,0 +1,117 @@
...
+			Command:     "cat /etc/default/kubelet",
+			Asserter: func(stdout, stderr string) error {
+				if strings.Contains(stdout, "--dynamic-config-dir") {
+					return fmt.Errorf("/etc/default/kubelet should not contain kubelet flag '--dynamic-config-dir', but does")
+				}
+				return nil
+			},
+		},
+		{
+			Description: "assert sysctls set by customdata",
```
alexeldeib, line 83: descriptions are nice :)

e2e/scenario/types.go
alexeldeib: maybe not really a big deal. I actually like the separate validator funcs for reuse.

e2e/pod.go
```diff
@@ -0,0 +1,81 @@
...
+		return "", fmt.Errorf("failed to list debug pod: %q", err)
+	}
+
+	if len(podList.Items) < 1 {
+		return "", fmt.Errorf("failed to find debug pod, list by selector returned no results")
+	}
+
+	var podName string
+	for _, pod := range podList.Items {
+		if strings.Contains(pod.Spec.NodeName, "aks-nodepool") {
```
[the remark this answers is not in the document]
cameronmeissner: ahh yess, this is what I meant to remove earlier after merging that other PR, thanks for catching this

e2e/pollers.go
```diff
@@ -54,3 +84,21 @@ func pollExtractVMLogs(ctx context.Context, t *testing.T, vmssName string, priva
...
+		if err != nil {
+			return false, err
+		}
+
+		return pod.Status.Phase == corev1.PodPhase("Running"), nil
+	})
+}
+
+func waitUntilPodDeleted(ctx context.Context, kube *kubeclient, podName string) error {
+	return wait.PollImmediateWithContext(ctx, 5*time.Second, 3*time.Minute, func(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
cameronmeissner: I think this was something added in the initial demo, can probably get away with at least lowering the timeout value or getting rid of the poller altogether

e2e/validation.go
cameronmeissner: yeeeeah I was kinda thinking the same thing, I like the idea of introducing a struct along the lines of "RemoteCommandOpts" that would contain the necessary config (private key, private IP, pod name, etc.) which could then implement a receiver that actually takes in a command and executes it

might be misunderstanding what you're saying tho

e2e/pollers.go
```diff
@@ -2,12 +2,42 @@ package e2e_test
...
+		// this denotes a retriable SSH failure
+		if res.exitCode == "255" {
+			return false, nil
+		}
+
+		execResult = res
+		return true, nil
+	})
+
+	if err != nil {
```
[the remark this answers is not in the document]
cameronmeissner: might be worth adding a comment there - the way execOnVm is implemented has it such that the only non-retriable error it throws is when there's a parsing error during exit code extraction from the stream error message, the other errors: "... encountered unexpected error when executing command on pod: ..." and "... unable to create new SPDY executor for pod exec: ..." are meant to be retried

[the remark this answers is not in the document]
cameronmeissner: might also be worth logging those errors out though at least so we know what's happening and such that they aren't completely swallowed up

e2e/validation.go
cameronmeissner: either way, maybe a good follow-up item

e2e/pollers.go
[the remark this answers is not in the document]
cameronmeissner: added some logging and a comment

e2e/exec.go
```diff
@@ -0,0 +1,172 @@
...
+		stdout, stderr bytes.Buffer
+		exitCode       string = "0"
+	)
+
+	err = exec.StreamWithContext(ctx, remotecommand.StreamOptions{
+		Stdout: &stdout,
+		Stderr: &stderr,
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "command terminated with exit code") {
```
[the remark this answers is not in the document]
cameronmeissner: yeah this seems to be the pattern people follow from a bit of googling, though I couldn't find another reference of someone needing to extract exit code specifically from stderr or the stream error message directly, though this does make sense since it seems like exit codes are independent from stderr entirely

also makes sense from what I've seen the contents of stderr actually are in previous testing

alexeldeib (approved):

e2e/exec.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		result[file] = execResult.stdout.String()
+	}
+	return result, nil
+}
+
+func extractClusterParameters(ctx context.Context, t *testing.T, kube *kubeclient) (map[string]string, error) {
+	commandList := map[string]string{
+		"/etc/kubernetes/azure.json":            "cat /etc/kubernetes/azure.json",
```
anujmaheshwari1, line 60: did we extract just these 3 files from the cluster? my memory on ab e2e is fogged.

e2e/validation.go
```diff
@@ -0,0 +1,117 @@
...
+			if err != nil {
+				return fmt.Errorf("failed validator assertion: %s", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func commonLiveVMValidators() []*scenario.LiveVMValidator {
```
anujmaheshwari1, line 70: truly expanding ab e2e! awesome stuff!

anujmaheshwari1 (approved):

PR #3023, opened 2023-04-12
chore: run e2e coverage

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/suite_test.go
```diff
@@ -15,10 +19,18 @@ import (
 	"github.com/barkimedes/go-deepcopy"
 )
 
+type getNodeBootstrapping func(ctx context.Context, nbc *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error)
```
cameronmeissner, line 22: maybe move this to another file entirely? could make a new one called nodebootstrapping.go or nbc.go

e2e/suite_test.go
```diff
@@ -15,10 +19,18 @@ import (
...
+
+var e2eMode string
```
cameronmeissner, line 24: can we move this to another file as well? maybe flags.go?

e2e/suite_test.go
```diff
@@ -91,19 +103,27 @@ func Test_All(t *testing.T) {
...
-			runScenario(ctx, t, r, opts)
+			runScenario(ctx, t, r, opts, bootstrappingGetter)
 		})
 	}
 }
 
-func runScenario(ctx context.Context, t *testing.T, r *mrand.Rand, opts *scenarioRunOpts) {
+func determineNodeBootstrappingGetter(mode string) getNodeBootstrapping {
+	if mode == "coverage" {
+		fmt.Println("using coverage mode")
```
cameronmeissner, line 113: nit: fmt.Println -> log.Println

e2e/suite_test.go
```diff
@@ -173,7 +193,25 @@ func bootstrapVMSS(ctx context.Context, t *testing.T, r *mrand.Rand, opts *scena
 	return vmssName, vmssModel, cleanupVMSS, nil
 }
 
-func getNodeBootstrapping(ctx context.Context, nbc *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error) {
+func getNodeBootstrappingForCoverage(ctx context.Context, nbc *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error) {
```
cameronmeissner, line 196: I'd also move getNodeBootstrapping{ForCoverage/ForValidation} to nbc/nodebootstrapping.go, then maybe have a function which wraps the 2 called getNodeBootstrappingConfiguration that calls one or the other based on `e2eMode` - that way we don't have to plumb around a function ptr

cameronmeissner, line 196: the implementation of this function specifically looks solid btw

e2e/suite_test.go
cameronmeissner: might be able to get away without this with the other changes I suggested, either way if you wanna keep it I'd rename to getNodeBootstrappingFn or getNodeBootstrappingFunction or something

e2e/suite_test.go
UtheMan: Moved getNodeBootstrapping, but e2eMode - I'd prefer to leave it here - it's only used in suite_test.go to get appropriate bootstrapping func and suite_test.go is also where the flag value is retrieved.

e2e/suite_test.go
UtheMan: Moved to another file

e2e/suite_test.go
cameronmeissner: makes sense, either way is fine with me

cameronmeissner (approved):

e2e/nodebootstrapping.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+	"encoding/json"
+	"github.com/Azure/agentbaker/pkg/agent"
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"log"
+	"os/exec"
+)
+
+type getNodeBootstrappingFn func(ctx context.Context, nbc *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error)
```
cameronmeissner, line 12: actually can we rename this type to nodeBootstrappingFn and then rename getNodeBootstrapping to getNodeBootstrappingFn since its returning a function now rather than node bootstrapping?

e2e/nodebootstrapping.go
```diff
@@ -0,0 +1,51 @@
...
+		return getNodeBootstrappingForValidation
+	}
+}
+
+func getNodeBootstrappingForCoverage(ctx context.Context, nbc *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error) {
+	payload, err := json.Marshal(nbc)
+	if err != nil {
+		log.Fatalf("failed to marshal nbc, error: %s", err)
+	}
+	cmd := exec.Command("curl", "-X", "POST", "-d", string(payload), "localhost:8080/getnodebootstrapdata")
```
cameronmeissner, line 28: can we use net/http to do this instead?

cameronmeissner (approved):

alexeldeib (approved):
[3:19 PM] Ace Eldeib




lgtm once tests pass, mind squashing commits? or going for the squash merge?

PR #3024, opened 2023-04-12
fix: do not add IPv6DualStack to Windows kube-proxy feature-gates when K8s >= 1.25.0

tyler-lloyd:
**What type of PR is this?**

/kind bug

`IPv6DualStack` is removed from the Kubernetes FeatureFlag set
in 1.25.0 and has been enabled by default since 1.21.0. There are
some other places in CSE that depend on the feature flag being part
of the config so this change is just to remove it from kube-proxy's
list of args so it doesn't crash at startup.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/datamodel/const.go
+++ b/pkg/agent/datamodel/const.go
@@ -181,7 +181,7 @@ const (
 	OSSKUMariner    = "Mariner"
 )
 
-// Feature Flags
+// Feature Flags.
 const (
 	BlockOutboundInternet = "BlockOutboundInternet"
 	CSERunInBackground    = "CSERunInBackground"
--- a/pkg/agent/datamodel/const.go
+++ b/pkg/agent/datamodel/const.go
@@ -180,3 +180,12 @@ const (
 	OSSKUCBLMariner = "CBLMariner"
 	OSSKUMariner    = "Mariner"
 )
+
+// Feature Flags
+const (
+	BlockOutboundInternet = "BlockOutboundInternet"
+	CSERunInBackground    = "CSERunInBackground"
+	EnableIPv6DualStack   = "EnableIPv6DualStack"
+	EnableIPv6Only        = "EnableIPv6Only"
+	EnableWinDSR          = "EnableWinDSR"
+)
--- a/pkg/agent/datamodel/types_test.go
+++ b/pkg/agent/datamodel/types_test.go
@@ -1664,6 +1664,18 @@ func TestGetKubeProxyFeatureGatesWindowsArguments(t *testing.T) {
 			expectedFeatureGates: "",
 		},
 		{
+			name: "IPV6 enabled and version has feature gate (>= 1.15 < 1.25)",
+			properties: &Properties{
+				FeatureFlags: &FeatureFlags{
+					EnableIPv6DualStack: true,
+				},
+				OrchestratorProfile: &OrchestratorProfile{
+					OrchestratorVersion: "1.24.12",
+				},
+			},
+			expectedFeatureGates: "\"IPv6DualStack=true\"",
+		},
+		{
 			name: "IPV6 enabled but version does not have feature gate",
 			properties: &Properties{
 				FeatureFlags: &FeatureFlags{
--- a/pkg/agent/datamodel/types_test.go
+++ b/pkg/agent/datamodel/types_test.go
@@ -1640,6 +1640,43 @@ func TestGetKubeProxyFeatureGatesWindowsArguments(t *testing.T) {
 			expectedFeatureGates: "\"IPv6DualStack=true\"",
 		},
 		{
+			name: "IPV6 enabled but version does not have feature gate (too old)",
+			properties: &Properties{
+				FeatureFlags: &FeatureFlags{
+					EnableIPv6DualStack: true,
+				},
+				OrchestratorProfile: &OrchestratorProfile{
+					OrchestratorVersion: "1.11.0",
+				},
+			},
+			expectedFeatureGates: "",
+		},
+		{
+			name: "IPV6 enabled but version does not have feature gate",
+			properties: &Properties{
+				FeatureFlags: &FeatureFlags{
+					EnableIPv6DualStack: true,
+				},
+				OrchestratorProfile: &OrchestratorProfile{
+					OrchestratorVersion: "1.25.0",
+				},
+			},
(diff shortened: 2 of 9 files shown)


staging/cse/windows/provisioningscripts/kubeproxystart.ps1
```diff
@@ -33,8 +70,23 @@ if ($KubeproxyFeatureGates -contains "WinDSR=true") {
     $global:KubeproxyArgList += @("--enable-dsr=true")
 }
 
-if ($KubeproxyFeatureGates.Count -ne 0) {
-    $global:KubeproxyArgList += @("--feature-gates=" + ($KubeproxyFeatureGates -join ","))
+$featureGateArgs = ""
+foreach ($feature in $KubeproxyFeatureGates) {
+    # IPv6DualStack feature gate should not be passed to kube-proxy in >= 1.25.0
+    # https://github.com/kubernetes/kubernetes/blob/ef70d260f3d036fc22b30538576bbf6b36329995/pkg/features/kube_features.go#L945
+    if (($feature -like "IPv6DualStack=*") -and ((Compare-SemanticVersion -Version1 $KubernetesVersion -Version2 "1.25.0") -ge 0)) {
```
tyler-lloyd, line 77: to be honest we may not need the sem ver check since IPv6DualStack has been enabled by default since 1.21.

AbelHu, line 77: could you use `([version]$KubernetesVersion).CompareTo([version]"1.25.0") -ge 0`?

staging/cse/windows/provisioningscripts/kubeproxystart.ps1
```diff
@@ -33,8 +70,23 @@ if ($KubeproxyFeatureGates -contains "WinDSR=true") {
...
+        continue
```
AbelHu, line 78: Can we remove `featureGates["IPv6DualStack"] = true` in `GetKubeProxyFeatureGatesWindowsArguments` in pkg/agent/datamodel/types.go?

staging/cse/windows/provisioningscripts/kubeproxystart.ps1
tyler-lloyd: the problem using the built in `version` type is it default to `-1` for any version categories that aren't there. So comparing "1.25" and "1.25.0" doesn't work. I would expect them to be equal but pwsh thinks "1.25" is less than "1.25.0"

```
PS /home/tyler> ([version]"1.25").CompareTo([version]"1.25.0")
-1
```

This may not matter here because OrchestratorVersion always contains the patch version but I just wanted to be safe.

staging/cse/windows/provisioningscripts/kubeproxystart.ps1
tyler-lloyd: ah actually that might be better. I was thinking we were checking the `IPv6DualStack=true` feature gate somewhere else but it looks like we are really using `IsIPv6DualStackFeatureEnabled` and then setting `$global:IsDualStackEnabled` based on that.

tyler-lloyd: For now I'm going to leave this bit in kubeproxystart.ps1 to remove it from the list in addition to removing from types.go. Since we need this change relatively quickly, I want to be able to use a new CSE version and will remove this in a follow up PR so that it is fully controlled by types.go

pkg/agent/datamodel/const.go
```diff
@@ -180,3 +180,12 @@ const (
 	OSSKUCBLMariner = "CBLMariner"
 	OSSKUMariner    = "Mariner"
 )
+
+// Feature Flags.
+const (
```
alexeldeib, line 185: thanks!

alexeldeib (approved):
const/datamodel changes look okay to me

we aren't using a number of those, probably candidates for cleanup

pkg/agent/datamodel/types_test.go
```diff
@@ -1635,6 +1635,43 @@ func TestGetKubeProxyFeatureGatesWindowsArguments(t *testing.T) {
...
+				FeatureFlags: &FeatureFlags{
+					EnableIPv6DualStack: true,
+				},
+				OrchestratorProfile: &OrchestratorProfile{
+					OrchestratorVersion: "1.11.0",
+				},
+			},
+			expectedFeatureGates: "",
+		},
+		{
```
AbelHu, line 1650: Is a test case with >1.15.0 <1.25.0 missed?

staging/cse/windows/provisioningscripts/windowsnodereset.ps1
```diff
@@ -15,7 +15,12 @@ $global:NetworkMode = "L2Bridge"
 $global:NetworkPlugin = $Global:ClusterConfiguration.Cni.Name
 $global:ContainerRuntime = $Global:ClusterConfiguration.Cri.Name
 $UseContainerD = ($global:ContainerRuntime -eq "containerd")
-$IsDualStackEnabled = $Global:ClusterConfiguration.Kubernetes.Kubeproxy.FeatureGates -contains "IPv6DualStack=true"
+# if dual-stack is enabled, the clusterCidr will have an IPv6 CIDR in the comma separated list
+# we can split the entire string by ":" to get a count of how many ":" there are. If there are
+# at least 3 groups (which means there are at least 2 ":") then we know there is an IPv6 CIDR
+# in the list. We cannot just rely on `ClusterCidr -like "*::*" because there are IPv6 CIDRs that
+# don't have "::", e.g. fe80:0:0:0:0:0:0:0/64
+$IsDualStackEnabled = ($Global:ClusterConfiguration.Kubernetes.Network.ClusterCidr -split ":").Count -ge 3
```
AbelHu, line 24: Q: Do we still need `featureGates["IPv6DualStack"] = true` with this change? I think that IPv6DualStack is not used by any feature except dual-stack

tyler-lloyd: probably not, but I'll keep it for now just in case

AbelHu (approved):

AbelHu:
@tyler-lloyd please check the failed test case to confirm whether it is related to this PR

PR #3028, opened 2023-04-13

HDYA:

### DIFF
--- a/vhdbuilder/packer/kube-proxy-images.json
+++ b/vhdbuilder/packer/kube-proxy-images.json
@@ -6,12 +6,16 @@
         "downloadURL": "mcr.microsoft.com/oss/kubernetes/kube-proxy:v*",
         "amd64OnlyVersions": [],
         "multiArchVersions": [
-          "1.24.9-hotfix.20230208.2",
           "1.24.10.3",
+          "1.24.9-hotfix.20230208.2",
           "1.25.5-hotfix.20230208",
+          "1.25.5-hotfix.20230407.1",
           "1.25.6",
+          "1.25.6-hotfix.20230407.1",
           "1.26.0",
-          "1.26.3.2"
+          "1.26.0-hotfix.20230407.1",
+          "1.26.3.2",
+          "1.26.3.4"
         ]
       }
     ]
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -203,6 +203,8 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/autoscaler/cluster-proportional-autoscaler:*",
       "amd64OnlyVersions": [
+        "1.7.1-hotfix.20200403",
+        "1.7.1-hotfix.20200403.1",
         "1.8.3"
       ],
       "multiArchVersions": [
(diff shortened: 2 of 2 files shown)


vhdbuilder/packer/kube-proxy-images.json
```diff
@@ -6,12 +6,16 @@
...
           "1.24.10.3",
+          "1.24.9-hotfix.20230208.2",
           "1.25.5-hotfix.20230208",
+          "1.25.5-hotfix.20230407.1",
           "1.25.6",
+          "1.25.6-hotfix.20230407.1",
           "1.26.0",
-          "1.26.3.2"
+          "1.26.0-hotfix.20230407.1",
+          "1.26.3.2",
```
HDYA, line 17: currently in RP we are using 1.26.3.2 without baking it into VHD, but 1.26.3.4 is already available.

ganeshkumar5699 (commented):
Do the older versions have CVEs? In that case, can we remove those versions from being cached in the VHD?

HDYA:
> @HDYA Do the older versions have CVEs? In that case, can we remove those versions from being cached in the VHD?

I don't think so, if new VHD only contains patched version, when the VHD version gets bumped, RP still reference old versions,
suddenly all new nodes lose old version of images from cache, for some customers their clusters may not be able to pull the image.

ganeshkumar5699:
> > @HDYA Do the older versions have CVEs? In that case, can we remove those versions from being cached in the VHD?
> 
> @ganeshkumar5699
> 
> I think this would be debatable, if new VHD only contains patched version, when the VHD version gets bumped, RP still reference old versions, suddenly all new nodes lose old versions of images from cache, Then, for some customers, their clusters may not be able to pull the image.

> If RP and Agentbaker release are not fully in sync (which would happen), then there will be a brief period with discrepancy where running the pod, would need to pull in a few changed layers between image versions. I imagine the latency difference won't be significant vs caching vulnerable images on the VHD. 

Are there cases where pulling layers from mcr would be blocked?

HDYA:
> > > @HDYA Do the older versions have CVEs? In that case, can we remove those versions from being cached in the VHD?
> > 
> > 
> > @ganeshkumar5699
> > I think this would be debatable, if new VHD only contains patched version, when the VHD version gets bumped, RP still reference old versions, suddenly all new nodes lose old versions of images from cache, Then, for some customers, their clusters may not be able to pull the image.
> 
> If RP and Agentbaker release are not fully in sync (which would happen), then there will be a brief period with discrepancy where running the pod, would need to pull in a few changed layers between image versions. I imagine the latency difference won't be significant vs caching vulnerable images on the VHD.
> 
> Are there cases where pulling certain layers from mcr would be blocked?

replied offline.
main concern is isolated environment and may cause incidents for customers not setting firewall correctly.

However, this should be another topic, I think we should not let it block our current patch process.

andyliuliming (approved):

alexeldeib (approved):

alexeldeib:
> main concern is isolated environment and may cause incidents for customers not setting firewall correctly.

I want to be very clear -- caching images at all is an optimization. We require MCR connectivity and node provisioning will fail before you ever get to the point of pulling images (kubelet won't even be started).

+1 to merge and iterate, Ganesh we can use our own scan results as a test case for CVEs here.

PR #3036, opened 2023-04-13
fix: replace password-related settings properly

tobiasb-ms:
**What type of PR is this?**
/kind bug

We have code that modifies `/etc/default/useradd` `/etc/login.defs` to bring them into compliance with CIS benchmarks. However, the script does some weird things to those files and in the case of `mariner` actually fails to set the `INACTIVE` setting in `/etc/useradd`.

This change fixes the script to modify settings in place or append them, taking into account the fact that a file may not have a trailing new-line.

I also normalized the tab length in the file, which was different in different functions.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Here's the mariner `/etc/default/useradd` before this change. You can see that `INACTIVE` isn't actually set and `CREATE_MAIL_SPOOL` is corrupted.
```
# useradd defaults file
GROUP=100
HOME=/home
-1
EXPIRE=
SHELL=/bin/bash
SKEL=/etc/skel
CREATE_MAIL_SPOOL=noINACTIVE=30
```

Here it is with these changes:
```
# useradd defaults file
GROUP=100
HOME=/home
INACTIVE=30
EXPIRE=
SHELL=/bin/bash
SKEL=/etc/skel
CREATE_MAIL_SPOOL=no
```

`/etc/login.defs` is a much larger file, as is `/etc/default/useradd` on ubuntu. You can see both files from before/after on mariner and ubuntu in this zip file:
[cis-fix-password-settings.zip](https://github.com/Azure/AgentBaker/files/11227106/cis-fix-password-settings.zip)

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cis.sh
+++ b/parts/linux/cloud-init/artifacts/cis.sh
@@ -1,19 +1,19 @@
 #!/bin/bash
 
 assignRootPW() {
-  if grep '^root:[!*]:' /etc/shadow; then
-    VERSION=$(grep DISTRIB_RELEASE /etc/*-release| cut -f 2 -d "=")
-    SALT=$(openssl rand -base64 5)
-    SECRET=$(openssl rand -base64 37)
-    CMD="import crypt, getpass, pwd; print(crypt.crypt('$SECRET', '\$6\$$SALT\$'))"
-    if [[ "${VERSION}" == "22.04" ]]; then
-      HASH=$(python3 -c "$CMD")
-    else
-      HASH=$(python -c "$CMD")
-    fi
+    if grep '^root:[!*]:' /etc/shadow; then
+        VERSION=$(grep DISTRIB_RELEASE /etc/*-release | cut -f 2 -d "=")
+        SALT=$(openssl rand -base64 5)
+        SECRET=$(openssl rand -base64 37)
+        CMD="import crypt, getpass, pwd; print(crypt.crypt('$SECRET', '\$6\$$SALT\$'))"
+        if [[ "${VERSION}" == "22.04" ]]; then
+            HASH=$(python3 -c "$CMD")
+        else
+            HASH=$(python -c "$CMD")
+        fi
 
-    echo 'root:'$HASH | /usr/sbin/chpasswd -e || exit $ERR_CIS_ASSIGN_FILE_PERMISSION
-  fi
+        echo 'root:'$HASH | /usr/sbin/chpasswd -e || exit $ERR_CIS_ASSIGN_ROOT_PW
+    fi
 }
 
 assignFilePermissions() {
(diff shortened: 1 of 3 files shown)


alexeldeib:
if you want to assert the contents of the files look right after the VHD build, you could add some more bash in https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/test/linux-vhd-content-test.sh#L352

parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -1,19 +1,19 @@
...
-    else
-      HASH=$(python -c "$CMD")
-    fi
+    if grep '^root:[!*]:' /etc/shadow; then
+        VERSION=$(grep DISTRIB_RELEASE /etc/*-release | cut -f 2 -d "=")
+        SALT=$(openssl rand -base64 5)
+        SECRET=$(openssl rand -base64 37)
+        CMD="import crypt, getpass, pwd; print(crypt.crypt('$SECRET', '\$6\$$SALT\$'))"
+        if [[ "${VERSION}" == "22.04" ]]; then
+            HASH=$(python3 -c "$CMD")
```
SeanDougherty, line 10: I think we use python3 as well, we will want to hook into this conditional as well

image

parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -1,19 +1,19 @@
...
+        else
+            HASH=$(python -c "$CMD")
+        fi
 
-    echo 'root:'$HASH | /usr/sbin/chpasswd -e || exit $ERR_CIS_ASSIGN_FILE_PERMISSION
-  fi
+        echo 'root:'$HASH | /usr/sbin/chpasswd -e || exit $ERR_CIS_ASSIGN_FILE_PERMISSION
```
SeanDougherty, line 15: Should we be using [ERR_CIS_ASSIGN_ROOT_PW](https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_helpers.sh#L57) here?

parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -59,29 +59,74 @@ assignFilePermissions() {
...
+    local FILE=$3
+
+    # Search and replace/append.
+    if grep -E "$SEARCH_PATTERN" "$FILE" >/dev/null; then
+        sed -E -i "s|${SEARCH_PATTERN}|${SETTING_LINE}|g" "$FILE" || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
+    else
+        echo -e "\n${SETTING_LINE}" >>"$FILE"
+    fi
+
+    # At the end, there must be only line for the setting, and it must be what we want.
```
SeanDougherty, line 83: Nit: The wording here is a bit unclear

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -345,14 +345,152 @@ testCustomCATimerNotStarted() {
...
+  echo "$test: Checking general format of $settings_file"
+  grep -E -v -n -H '^\s*(#|$)' $settings_file | grep -E -v '^.+:.+:[A-Z_]+=[^[:cntrl:]]*$' >>/dev/stderr
+  if [ $? -eq 0 ]; then
+    err $test "File $settings_file is not valid; see above for bad lines."
   fi
+
+  # Look for the settings we specifically set in <repo-root>/parts/linux/cloud-init/artifacts/cis.sh
+  # and ensure they're set to the values we expect.
+  echo "$test: Checking specific settings in $settings_file"
+  testSetting $test $settings_file '=' INACTIVE 30
```
SeanDougherty, line 442: Optional Suggestion: could we reorder the last three arguments to 'setting' 'separator' 'expected value', so that it reads from left to right?

SeanDougherty (changes requested):
Just a few comments, overall very good!

parts/linux/cloud-init/artifacts/cis.sh
tobiasb-ms: almost certainly; I had only changed a formatting thing here but I'll make this change.

vhdbuilder/packer/test/linux-vhd-content-test.sh
tobiasb-ms: I change the function in general, but I think I have the ordering better now.

parts/linux/cloud-init/artifacts/cis.sh
tobiasb-ms: Good idea, but let's make that a separate checkin.

SeanDougherty (approved):

alexeldeib (approved):
overall lgtm -- since our run tests don't log the output currently, and bash is fickle, if you want to get a final sanity check, two things you could do (not a requirement):
- upload the file from the VHD build to the artifacts for a visual inspection.
  - example: we extract a list of all pulled container images after running a script here https://github.com/Azure/AgentBaker/blob/66cdfe9d8e7d6cb2379870b0c19c1bb926afe38e/vhdbuilder/packer/vhd-image-builder-mariner2-gen2.json#L484-L500
- update log extraction from agentbaker e2e https://github.com/Azure/AgentBaker/blob/master/e2e/exec.go#L64-L67
  - but that requires us bumping the VHD used to pick up your changes, bit more work, no real reason. I'd do the VHD artifact.

tobiasb-ms:
> overall lgtm -- since our run tests don't log the output currently, and bash is fickle, if you want to get a final sanity check, two things you could do (not a requirement):
> 
> * upload the file from the VHD build to the artifacts for a visual inspection.
>   
>   * example: we extract a list of all pulled container images after running a script here https://github.com/Azure/AgentBaker/blob/66cdfe9d8e7d6cb2379870b0c19c1bb926afe38e/vhdbuilder/packer/vhd-image-builder-mariner2-gen2.json#L484-L500
> * update log extraction from agentbaker e2e https://github.com/Azure/AgentBaker/blob/master/e2e/exec.go#L64-L67
>   
>   * but that requires us bumping the VHD used to pick up your changes, bit more work, no real reason. I'd do the VHD artifact.

Good ideas, but I'm comfortable without those files being uploaded to artifacts. Soon, these will have extra testing in the form of overall cis-related validation.

PR #3038, opened 2023-04-13
fix: explicitly activate nf_conntrack on mariner

cameronmeissner:
**What type of PR is this?**

/kind bug

Fixes a bug where custom linux os config would not respected for certain sysctls related to the nf_conntrack module on Mariner VHDs

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -29,6 +29,9 @@ if [[ $OS == $MARINER_OS_NAME ]]; then
   chmod 755 /opt
   chmod 755 /opt/azure
   chmod 644 ${VHD_LOGS_FILEPATH}
+
+  # explicitly activate nf_conntrack so custom sysctls can be properly set 
+  echo nf_conntrack >> /etc/modules-load.d/contrack.conf
```
hbeberman, line 34: I suggest moving this into a function in [tool_installs_mariner.sh](https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh) then calling it in this block of [install-dependencies.sh](https://github.com/Azure/AgentBaker/blob/4e596e6763a0407a839aff08ea357438e2125aad/vhdbuilder/packer/install-dependencies.sh#L89) to keep as many of the config fixes in one place as possible.

dmcilvaney (approved):

alexeldeib (approved):
can we follow up with live VM validation for this? looks like you already started offline

cc @hbeberman

alexeldeib:
oops didn't realize Henry and Daniel already commented :)

vhdbuilder/packer/pre-install-dependencies.sh
cameronmeissner: > can we follow up with live VM validation for this? looks like you already started offline
> 
> cc @hbeberman

already working on the PR for that :)

PR #3045, opened 2023-04-14
fix: update addon-resizer and cluster-proportional-autoscaler to fix CVEs

ganeshkumar5699:
**What type of PR is this?**
/kind update

These components are using an older version with CVEs, that has been flagged. Updating to fix that.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -201,7 +201,7 @@
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/autoscaler/cluster-proportional-autoscaler:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "1.8.5"
+        "1.8.6"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/components.json
```diff
@@ -2,11 +2,9 @@
   "ContainerImages": [
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/autoscaler/addon-resizer:*",
-      "amd64OnlyVersions": [
-        "1.8.5"
-      ],
+      "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "1.8.14"
+        "1.8.18"
```
alexeldeib, line 7: should we just delete these? do any addons use them?

alexeldeib (approved):
lgtm but deleting maybe better?

vhdbuilder/packer/components.json
ganeshkumar5699: These components are being referenced by addons (though different versions), so keeping the components for now.

ganeshkumar5699:
> lgtm but deleting maybe better?

Want to delete if possible too. Components do have references by certain addons, so keeping them for now.

PR #3056, opened 2023-04-17
fix: switch dumpAll() and surrounding funcs to value semantics and handle dereference panics

UtheMan:
…lExecOnVM

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/pollers.go
```diff
@@ -52,7 +52,7 @@ func pollExecOnVM(ctx context.Context, kube *kubeclient, vmPrivateIP, jumpboxPod
 	})
 
 	if err != nil {
-		return nil, err
+		return execResult, err
```
alexeldeib, line 55: or in the method?

alexeldeib (approved):

e2e/pollers.go
cameronmeissner: yeah, I'd personally rather keep the explicit nil here and add the check to the method, what do you think @UtheMan

UtheMan: We could do that, but I feel like it would clutter the method?

```go
		execResult, err := pollExecOnVM(ctx, opts.kube, privateIP, podName, sshPrivateKey, command)
		if err != nil {
                        if execResult != nil {
			execResult.dumpAll()
                        }
			return fmt.Errorf("unable to execute validator command %q: %s", command, err)
		}

		if execResult != nil &&  execResult.exitCode != "0" {
			execResult.dumpAll()
			return fmt.Errorf("validator command %q terminated with exit code %s", command, execResult.exitCode)
		}
```

would have to be something like this I think?

cameronmeissner: depends on whether we think there could be instances where execResult is nil while err is also nil

I was thinking we'd do something like

if r != nil {
  r.dumpStdout()
  r.dumpStderr()
}

in r.dumpAll(), while keeping the other stuff as is

cameronmeissner (approved):

e2e/exec.go
```diff
@@ -24,8 +24,10 @@ type podExecResult struct {
 }
 
 func (r *podExecResult) dumpAll() {
-	r.dumpStdout()
-	r.dumpStderr()
+	if r != nil {
```
juan-lee, line 27: Does this func need to be a pointer receiver? Does it mutate state?

UtheMan: Good point, we can modify it to have a value receiver.

cameronmeissner: yeah this makes sense, and also since the majority of this struct's fields are already pointers themselves so copies aren't too expensive

cameronmeissner: changed to value semantics

cameronmeissner (approved):

e2e/exec.go
```diff
@@ -76,9 +76,7 @@ func extractLogsFromVM(ctx context.Context, t *testing.T, vmssName string, sshPr
 		t.Logf("executing command on remote VM at %s of VMSS %s: %q", privateIP, vmssName, sourceCmd)
 
 		execResult, err := execOnVM(ctx, opts.kube, privateIP, podName, sshPrivateKey, sourceCmd)
-		if execResult != nil {
-			checkStdErr(execResult.stderr, t)
-		}
+		execResult.dumpStderr()
```
cameronmeissner, line 79: need to check execResult for nil here

e2e/exec.go
```diff
@@ -105,9 +103,7 @@ func extractClusterParameters(ctx context.Context, t *testing.T, kube *kubeclien
 		t.Logf("executing privileged command on pod %s/%s: %q", defaultNamespace, podName, sourceCmd)
 
 		execResult, err := execOnPrivilegedPod(ctx, kube, defaultNamespace, podName, sourceCmd)
-		if execResult != nil {
-			checkStdErr(execResult.stderr, t)
-		}
+		execResult.dumpStderr()
```
cameronmeissner, line 106: and here

e2e/validation.go
```diff
@@ -33,7 +33,9 @@ func runLiveVMValidators(ctx context.Context, t *testing.T, vmssName, sshPrivate
 
 		execResult, err := pollExecOnVM(ctx, opts.kube, privateIP, podName, sshPrivateKey, command)
 		if err != nil {
-			execResult.dumpAll()
+			if execResult != nil {
```
juan-lee, line 36: shouldn't this be removed all together?

cameronmeissner (approved):

juan-lee (approved):

PR #3061, opened 2023-04-18
fix: remove mariner 1.0 from vhd-content-test suite

SeanDougherty:
**What type of PR is this?**
/kind test

Removing VHD testing in 1.0 is not a regression in behavior since we weren't testing 1.0 before
this week. Going forward we're focusing on testing 2.0, and don't want to add workarounds within
the tests to make them support 1.0

N/A

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

1.0 not passing was uncovered after adding #3058

**Release note**:
```
none
```


hbeberman (approved):
LGTM.

Not needed for this PR, but this script should be refactored slightly to at least stash the script arguments into named variables so that these function calls are readable.

SeanDougherty:
Electing to remove 1.0 testing. Going forward we're focusing on testing 2.0, and don't want to add workarounds within
the tests to make them support 1.0

hbeberman (approved):

vhdbuilder/packer/test/run-test.sh
```diff
@@ -11,8 +11,8 @@ TEST_VM_ADMIN_PASSWORD="TestVM@$(date +%s)"
 set -x
 
 if [ "$OS_TYPE" == "Linux" ]; then
-  if [ "$IMG_SKU" == "20_04-lts-cvm" ]; then
-    echo "Skipping tests for CVM 20.04"
+  if [ "$IMG_SKU" == "20_04-lts-cvm" ] || [ "$IMG_SKU" == "cbl-mariner-1" ]; then
```
alexeldeib, line 14: don't we have v1 gen2 now as well? or is that covered by this?

alexeldeib (approved):

vhdbuilder/packer/test/run-test.sh
SeanDougherty: Good call out, I have sent up an update and will validate against the pipeline before merging

PR #3064, opened 2023-04-18
fix: cleanup k8s hyperkube check logic

ganeshkumar5699:
**What type of PR is this?**
/kind cleanup

Remove checks and helper function for installing k8s hyperkube, which was not used after k8s 1.18.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -415,16 +377,13 @@ cleanupRetaggedImages() {
...
     export KUBERNETES_VERSION
     export CLI_TOOL
     export -f retrycmd_if_failure
     export -f removeContainerImage
     export -f cleanUpImages
-    export -f cleanUpHyperkubeImages
     export -f cleanUpKubeProxyImages
-    bash -c cleanUpHyperkubeImages &
-    bash -c cleanUpKubeProxyImages &
+    bash -c cleanUpKubeProxyImages
```
juan-lee, line 386: are we intentionally removing & so cleanUpKubeProxyImages no longer runs in the background?

cameronmeissner (approved):
more of a rubber stamp than not, though seems to make sense :)

parts/linux/cloud-init/artifacts/cse_install.sh
ganeshkumar5699: Thanks for bringing this up. It could've slowed down CSE, so I reverted it back to the background run.

juan-lee (approved):

PR #3073, opened 2023-04-19
refactor: lint errors; strings, undefinied nums, global vars

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


apiserver/getnodebootstrapdata.go
```diff
@@ -19,8 +19,9 @@ const (
 
 // GetNodeBootstrapConfig endpoint for getting node bootstrapping data.
 func (api *APIServer) GetNodeBootstrapData(w http.ResponseWriter, r *http.Request) {
+	const timeout = 30 * time.Second
```
alexeldeib, line 22: nit: outside func definition for reuse in other handler?

pkg/agent/datamodel/const.go
```diff
@@ -124,12 +124,14 @@ const (
...
-// these sizes use a "converged" driver to support both cuda/grid workloads.
-// how do you figure this out? ask HPC or find out by trial and error.
-// installing vanilla cuda drivers will fail to install with opaque errors.
-// nvidia-bug-report.sh will is helpful, but usually it tells you the pci card id is incompatible.
-// That sends me to HPC folks.
-// see https://github.com/Azure/azhpc-extensions/blob/daaefd78df6f27012caf30f3b54c3bd6dc437652/NvidiaGPU/resources.json
+/* ConvergedGPUDriverSizes : these sizes use a "converged" driver to support both cuda/grid workloads.
+how do you figure this out? ask HPC or find out by trial and error.
+installing vanilla cuda drivers will fail to install with opaque errors.
+nvidia-bug-report.sh will is helpful, but usually it tells you the pci card id is incompatible.
```
alexeldeib, line 130:
```suggestion
nvidia-bug-report.sh may be helpful, but usually it tells you the pci card id is incompatible.
```

former typo on myself

pkg/agent/datamodel/sig_config.go
```diff
@@ -171,31 +96,74 @@ func (d Distro) IsContainerdDistro() bool {
...
+	for _, distro := range availableCBLMarinerDistros {
 		if d == distro {
 			return true
 		}
 	}
 	return false
 }
 func (d Distro) IsWindowsSIGDistro() bool {
-	for _, distro := range AvailableWindowsSIGDistros {
+	availableWindowsSIGDistros := []Distro{
```
alexeldeib, line 158: if you define these outside the func as un-exported, it still complains?

pkg/agent/datamodel/sig_config.go
```diff
@@ -285,9 +246,9 @@ var edgeZoneJSONContentsEmbedded string
 //go:embed mariner_v2_kata_gen2_tl_sig_version.json
 var marinerV2KataGen2TLJSONContentsEmbedded string
 
-var LinuxSIGImageVersion = getSIGVersionFromEmbeddedString(linuxVersionJSONContentsEmbedded)
-var EdgeZoneSIGImageVersion = getSIGVersionFromEmbeddedString(edgeZoneJSONContentsEmbedded)
-var CBLMarinerV2KataGen2TLSIGImageVersion = getSIGVersionFromEmbeddedString(marinerV2KataGen2TLJSONContentsEmbedded)
+var LinuxSIGImageVersion = getSIGVersionFromEmbeddedString(linuxVersionJSONContentsEmbedded)                         //nolint:gochecknoglobals
```
alexeldeib, line 249: nit - can we keep the nolint generally on the preceding lint where reasonable for readability?

pkg/agent/datamodel/types.go
```diff
@@ -823,7 +832,7 @@ func (p *Properties) GetClusterID() string {
 		} else if len(p.AgentPoolProfiles) > 0 {
 			h.Write([]byte(p.AgentPoolProfiles[0].Name))
 		}
-		r := rand.New(rand.NewSource(int64(h.Sum64())))
+		r := rand.New(rand.NewSource(int64(h.Sum64()))) //nolint:gosec // I think we want rand not crypto/rand here
```
alexeldeib, line 835: in theory I think we wanted to ban math/rand to avoid confusion but think we'd then need a wrapper which is out of scope

this isn't security sensitive either way

pkg/agent/datamodel/versions_test.go
```diff
@@ -368,7 +368,8 @@ func Test_GetValidPatchVersion(t *testing.T) {
 		t.Errorf("It is not the default Kubernetes version")
 	}
 
-	for version, enabled := range AllKubernetesWindowsSupportedVersions {
+	allKubernetesWindowsSupportedVersions := getAllKubernetesWindowsSupportedVersionsMap()
```
alexeldeib, line 371: cc @AbelHu but looks identical to me

pkg/agent/params.go
```diff
@@ -85,22 +85,23 @@ func assignKubernetesParametersFromAgentProfile(profile *datamodel.AgentPoolProf
...
-			if config.ContainerdVersion != "" {
-				addValue(parametersMap, "containerdVersion", config.ContainerdVersion)
-			}
-			if config.TeleportdPluginURL != "" {
-				addValue(parametersMap, "teleportdPluginURL", config.TeleportdPluginURL)
-			}
-			addValue(parametersMap, "containerdPackageURL", config.ContainerdPackageURL)
-		} else {
-			addValue(parametersMap, "cliTool", "docker")
+	if profile.KubernetesConfig == nil || profile.KubernetesConfig.ContainerRuntime == "" {
```
alexeldeib, line 88: isn't this incorrect? empty string needs to hit the docker else on L104

pkg/agent/params.go
```diff
@@ -85,22 +85,23 @@ func assignKubernetesParametersFromAgentProfile(profile *datamodel.AgentPoolProf
...
+		addValue(parametersMap, "cliTool", "ctr")
+		if config.ContainerdVersion != "" {
+			addValue(parametersMap, "containerdVersion", config.ContainerdVersion)
 		}
+		if config.TeleportdPluginURL != "" {
+			addValue(parametersMap, "teleportdPluginURL", config.TeleportdPluginURL)
+		}
+		addValue(parametersMap, "containerdPackageURL", config.ContainerdPackageURL)
+	} else {
+		addValue(parametersMap, "cliTool", "docker")
```
alexeldeib, line 104: feels like this gets missed for docker? can we assert on it somehow in unit test? 

should populate a bash variable in cse_cmd.sh, there are some tests in baker_test.go asserting on variable contents

pkg/agent/utils.go
```diff
@@ -99,6 +94,7 @@ func addSecret(m paramsMap, k string, v interface{}, encode bool) {
 		addValue(m, k, v)
 		return
 	}
+	keyvaultSecretPathRe := regexp.MustCompile(`^(/subscriptions/\S+/resourceGroups/\S+/providers/Microsoft.KeyVault/vaults/\S+)/secrets/([^/\s]+)(/(\S+))?$`) //nolint:lll
```
alexeldeib, line 97: i'm skeptical aks ever calls this but problem for another day

pkg/agent/utils.go
```diff
@@ -311,7 +315,7 @@ func getOrderedKubeletConfigFlagWithCustomConfigurationString(customConfig, defa
 
 	keys := []string{}
 	for key := range config {
-		if !datamodel.CommandLineOmittedKubeletConfigFlags[key] {
+		if !datamodel.GetCommandLineOmittedKubeletConfigFlags()[key] {
```
alexeldeib, line 318: probably a func call once before the loop makes sense

pkg/agent/baker.go
```diff
@@ -258,11 +258,10 @@ func normalizeResourceGroupNameForLabel(resourceGroupName string) string {
 	if strings.HasSuffix(truncated, "-") ||
 		strings.HasSuffix(truncated, "_") ||
 		strings.HasSuffix(truncated, ".") {
-		if len(truncated) > 62 {
+		if len(truncated) > maxLen-1 {
```
alexeldeib, line 261: I can't seem to find where this is defined, can we name is something clearer like maxResourceGroupLabelLength or just maxKubernetesLabelLength

pkg/agent/params.go
AlisonB319: `if profile.KubernetesConfig != nil && profile.KubernetesConfig.ContainerRuntime != "" {` does not have an else block. The 104 else - `addValue(parametersMap, "cliTool", "docker")` is for the if on line 94 `if profile.KubernetesConfig.ContainerRuntime == "containerd" {`

The old if would not run if KubernetesConfig was nil or if ContainerRuntime was empty. Is that not the expected behavior? I will look into the docker unit test as described in your below comment

pkg/agent/utils.go
AlisonB319: It seems to be running per key that is in config. Is that not what should happen?

pkg/agent/datamodel/sig_config.go
AlisonB319: Can you expand on what you mean by an unexported variable? From what I can tell, you would define the variable using a lower case letter but then need a function to set the value? - https://www.ardanlabs.com/blog/2014/03/exportedunexported-identifiers-in-go.html


```
type availableCBLMarinerDistros []Distro
func getAvailableCBLMarinerDistros( []Distro distro)  []Distro {
    return availableCBLMarinerDistros(distro) 
}

```

juan-lee (approved):

apiserver/routers.go
```diff
@@ -21,7 +20,7 @@ type Route struct {
 type Routes []Route
 
 // NewRouter returns a new router with defaults.
-func (api *APIServer) NewRouter(ctx context.Context) *mux.Router {
```
alexeldeib, line 24: should we remove it?

pkg/agent/datamodel/helper.go
```diff
@@ -47,7 +47,7 @@ func IsMIGNode(gpuInstanceProfile string) bool {
 // GetStorageAccountType returns the support managed disk storage tier for a give VM size.
 func GetStorageAccountType(sizeName string) (string, error) {
 	spl := strings.Split(sizeName, "_")
-	if len(spl) < 2 {
+	if len(spl) < 2 { //nolint:gomnd // less than 2 is invalid size
```
alexeldeib, line 50: nit: separate lines?

pkg/agent/datamodel/types.go
alexeldeib: also newline for comment please :)

pkg/agent/datamodel/helper.go
cameronmeissner: decided to fix the actual error instead

pkg/agent/datamodel/sig_config.go
cameronmeissner: > if you define these outside the func as un-exported, it still complains?

yeah it does unfortunately, I think I'll stick with the original format and add some nolint directives

pkg/agent/datamodel/versions_test.go
cameronmeissner: yeah this should be fine

pkg/agent/params.go
cameronmeissner: yeah this change looks correct to me

pkg/agent/params.go
cameronmeissner: added a couple unit tests around (what I believe is) the correct logic

apiserver/apiserver.go
```diff
@@ -44,8 +44,9 @@ func NewAPIServer(o *Options) (*APIServer, error) {
 // ListenAndServe wraps http.Server and provides context-based cancelation.
 func (api *APIServer) ListenAndServe(ctx context.Context) error {
 	svr := http.Server{
-		Addr:    api.Options.Addr,
-		Handler: api.NewRouter(ctx),
+		Addr:              api.Options.Addr,
+		Handler:           api.NewRouter(),
+		ReadHeaderTimeout: 0,
```
cameronmeissner, line 49: im assuming this is what it was doing before (deferring to ReadTimeout)

pkg/agent/baker_test.go
```diff
@@ -319,6 +319,25 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 
 			config.KubeletConfig = map[string]string{}
 		}, nil),
+		Entry("AKSUbuntu11604 with containerd", "AKSUbuntu1604+Containerd", "1.15.7", func(config *datamodel.NodeBootstrappingConfiguration) {
```
cameronmeissner, line 322: added these tests re: comments about refactored param logic

apiserver/apiserver.go
cameronmeissner: (prompted by gosec saying not explicitly setting it is bad practice)

cameronmeissner:
lol why is this triggering UVM builds...

apiserver/apiserver.go
alexeldeib: 0 is ignored

pkg/agent/datamodel/types.go
```diff
@@ -471,8 +473,8 @@ type WindowsProfile struct {
 	AlwaysPullWindowsPauseImage    *bool                      `json:"alwaysPullWindowsPauseImage,omitempty"`
 	ContainerdWindowsRuntimes      *ContainerdWindowsRuntimes `json:"containerdWindowsRuntimes,omitempty"`
 	WindowsCalicoPackageURL        string                     `json:"windowsCalicoPackageURL,omitempty"`
-	WindowsSecureTlsEnabled        *bool                      `json:"windowsSecureTlsEnabled,omitempty"`
-	WindowsGmsaPackageUrl          string                     `json:"windowsGmsaPackageUrl,omitempty"`
+	WindowsSecureTLSEnabled        *bool                      `json:"windowsSecureTlsEnabled,omitempty"`
```
alexeldeib, line 476: does this affect serialization anywhere to json?

cameronmeissner: it shouldn't since I made sure to only update the variable names rather than the json field mappings

cameronmeissner: added some testing around this to be certain

alexeldeib: potentially we should have sanity check with @AbelHu since this only touches windows, our testing for linux won't catch this well even if the logic is right

PR #3075, opened 2023-04-20
fix: remove cached metrics-server with CVEs

ganeshkumar5699:
**What type of PR is this?**
/kind fix

Remove unused metrics-server components with CVEs.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


nirnaymsft (changes requested):
I don't think we can remove v0.5.2 from this list, as Kubernetes version 1.22 will still be using it.

ganeshkumar5699:
> I don't think we can remove v0.5.2 from this list, as Kubernetes version 1.22 will still be using it.

k8s 1.22 is not being supported by AKS, so I assume it's ok to remove it?

ganeshkumar5699:
@nirnaymsft since you confirmed offline that my original change is still needed, could you please update your review? Thanks

nirnaymsft (approved):

alexeldeib (approved):

PR #3077, opened 2023-04-20
fix: update components.json with Metrics-server v0.6.3

nirnaymsft:
**What type of PR is this?**
/kind feature

Bump metrics-server patch version to v0.6.3

We're updating the patch versions used in RP as well as introducing a new version for k8s 1.24 preview.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


robbiezhang (approved):

vhdbuilder/packer/components.json
```diff
@@ -12,7 +12,8 @@
       "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v0.5.2",
-        "v0.6.1"
+        "v0.6.1",
```
ganeshkumar5699, line 15: Can these older versions be removed? They have CVEs and were flagged. I'd opened a PR yesterday for this, and wanted to reach out about that: https://github.com/Azure/AgentBaker/pull/3075

nirnaymsft: I can't remove v0.5.2 as this is used for Kubernetes version 1.22. i.e. old customer will still be using it.
I can see if I can remove v0.6.1, as after this change Kubernetes version 1.24 will use v0.6.3.

ganeshkumar5699: Caching components in the VHD is an optimization, for faster pod startup times. Since k8s 1.22 is currently out of support on AKS, caching v0.5.2 may not be necessary.

ganeshkumar5699 (approved):
Approving for v0.6.3 addition. Removal of older components with CVEs can be done in a separate PR.

cameronmeissner (approved):

alexeldeib (approved):

PR #3087, opened 2023-04-20
fix: ClusterFuzzLite batch fuzzing linter fixes

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/utils.go
+++ b/pkg/agent/utils.go
@@ -259,6 +259,9 @@ func IsKubernetesVersionGe(actualVersion, version string) bool {
 
 func getCustomDataFromJSON(jsonStr string) string {
 	var customDataObj map[string]string
+	jsonStr = strings.ReplaceAll(jsonStr, "\r", "\n")
+	jsonStr = strings.ReplaceAll(jsonStr, "\n", "\\n")
+	jsonStr = strings.ReplaceAll(jsonStr, "\t", "\\t")
 	err := json.Unmarshal([]byte(jsonStr), &customDataObj)
 	if err != nil {
 		panic(err)
--- a/pkg/agent/variables.go
+++ b/pkg/agent/variables.go
@@ -171,8 +171,19 @@ func isVHD(profile *datamodel.AgentPoolProfile) string {
 }
 
 func getOutBoundCmd(nbc *datamodel.NodeBootstrappingConfiguration, cloudSpecConfig *datamodel.AzureEnvironmentSpecConfig) string {
+	if nbc.ContainerService == nil {
+		return ""
+	}
 	cs := nbc.ContainerService
+
+	if cs.Properties == nil {
+		return ""
+	}
 	if cs.Properties.FeatureFlags.IsFeatureEnabled("BlockOutboundInternet") {
+		return ""
+	}
+
+	if cloudSpecConfig == nil {
 		return ""
 	}
 	registry := ""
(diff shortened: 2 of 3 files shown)


pkg/agent/variables.go
```diff
@@ -191,8 +202,14 @@ func getOutBoundCmd(nbc *datamodel.NodeBootstrappingConfiguration, cloudSpecConf
...
 	// so we need to use nc for the connectivity check.
-	clusterVersion, _ := semver.Make(cs.Properties.OrchestratorProfile.OrchestratorVersion)
-	minVersion, _ := semver.Make("1.18.0")
+	clusterVersion, err := semver.Make(cs.Properties.OrchestratorProfile.OrchestratorVersion)
+	if err != nil {
+		return err.Error()
+	}
+	minVersion, err := semver.Make("1.18.0")
+	if err != nil {
+		return err.Error()
```
alexeldeib, line 211: the return value here is really used during node provisioning, you can't return errors here

pkg/agent/variables.go
```diff
@@ -171,10 +171,21 @@ func isVHD(profile *datamodel.AgentPoolProfile) string {
 }
 
 func getOutBoundCmd(nbc *datamodel.NodeBootstrappingConfiguration, cloudSpecConfig *datamodel.AzureEnvironmentSpecConfig) string {
+	if nbc.ContainerService == nil {
+		return "nbc.ContainerService is nil"
```
alexeldeib, line 175: hrm...we may need a simpler way to fix these, we can't return error strings in any of these cases as-is

pkg/agent/utils.go
```diff
@@ -259,6 +259,9 @@ func IsKubernetesVersionGe(actualVersion, version string) bool {
 
 func getCustomDataFromJSON(jsonStr string) string {
 	var customDataObj map[string]string
+	jsonStr = strings.ReplaceAll(jsonStr, "\r", "\r\n")
```
alexeldeib, line 262: what's this one actually fixing?

AlisonB319: The errors were 
`panic: invalid character '\r' in string literal`
`panic: invalid character '\n' in string literal`
`panic: invalid character '\t' in string literal`

pkg/agent/variables.go
AlisonB319: I see, the linter was basically upset that the err wasn't checked. Would it be better to return an empty string?

alexeldeib:
I like the thinking but I might head a slightly different direction:
- for functions where fuzzing results in a panic because e.g. we don't do proper error checking, we need to adjust function signatures all the way up and down the stack to account for that. the goal is the root func cannot be misused.
- many funcs don't validate their input. we can fix that, but we need to avoid real errors in provisioning logic to satisfy the fuzzer. this again kind of points to strutural refactoring vs quick additions.

I think this is probably a longer/slower effort of identifying areas to refactor, to keep all the functionality and error cases the same but with cleaner structure

pkg/agent/variables.go
AlisonB319: Same with the comment above, any object with the potential to be nil needed to be checked. Same suggestion of returning an empty string?

pkg/agent/utils.go
AlisonB319: (I fixed `jsonStr = strings.ReplaceAll(jsonStr, "\r", "\r\n")` to `jsonStr = strings.ReplaceAll(jsonStr, "\r", "\n")` was trying to see if it just needed to be a different new line, but it wasn't happy)

cameronmeissner:
linting is currently passing, closing this for now.

PR #3088, opened 2023-04-20
fix: set network settings according to cis recommendations

tobiasb-ms:
**What type of PR is this?**
/kind bug

We are out of compliance with certain CIS recommendations related to network parameter settings. This change brings us into compliance with a couple changes.
First, it directly changes the file that is installed to `/etc/sysctl.d/60-CIS.conf` to set the parameters to the correct values.
Second, it updates our script that modifies files to comment out some settings in `/usr/lib/sysctl.d/50-default.conf` and `/lib/sysctl.d/50-default.conf`.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

You can find before and after versions of the files mentioned above for both mariner and ubuntu in the attached zip file.
[cis-fix-network-settings.zip](https://github.com/Azure/AgentBaker/files/11291052/cis-fix-network-settings.zip)

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cis.sh
+++ b/parts/linux/cloud-init/artifacts/cis.sh
@@ -124,10 +124,22 @@ setPWExpiration() {
     replaceOrAppendUserAdd INACTIVE 30
 }
 
+function fixRpFilter() {
+    # Mariner AKS CIS Benchmark: 3.3.7 Ensure Reverse Path Filtering is enabled
+    # Comment out rp_filter settings in /usr/lib/sysctl.d/50-default.conf and /lib/sysctl.d/50-default.conf
+    # We set these correctly in sysctl-d-60-CIS.conf, but they also need to not be present in the default file.
+    # These sed comands will comment out any line that starts with the settings we want to get rid of.
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.all.rp_filter/#net.ipv4.conf.all.rp_filter/g' /usr/lib/sysctl.d/50-default.conf
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.default.rp_filter/#net.ipv4.conf.default.rp_filter/g' /usr/lib/sysctl.d/50-default.conf
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.all.rp_filter/#net.ipv4.conf.all.rp_filter/g' /lib/sysctl.d/50-default.conf
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.default.rp_filter/#net.ipv4.conf.default.rp_filter/g' /lib/sysctl.d/50-default.conf
+}
+
 applyCIS() {
     setPWExpiration
     assignRootPW
     assignFilePermissions
+    fixRpFilter
 }
 
 applyCIS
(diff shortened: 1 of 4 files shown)


parts/linux/cloud-init/artifacts/sysctl-d-60-CIS.conf
```diff
@@ -1,24 +1,49 @@
...
+
+# Ubuntu CIS Benchmark: 3.3.1 Ensure IPv6 router advertisements are not accepted
+# Mariner AKS CIS Benchmark: 3.3.9 Ensure IPv6 router advertisements are not accepted
 net.ipv6.conf.all.accept_ra = 0
 net.ipv6.conf.default.accept_ra = 0
-# 3.3.2 Ensure IPv6 redirects are not accepted
-net.ipv6.conf.all.accept_redirects = 0
-net.ipv6.conf.default.accept_redirects = 0
+
+# Mariner AKS CIS Benchmark: 3.3.5 Ensure broadcast ICMP requests are ignored
```
corvus-callidus, line 34: Be cautious with the recommendation numbers for the Mariner benchmark. If there are any that were removed in a section, they'll be renumbered at publication time.

parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -124,10 +124,22 @@ setPWExpiration() {
...
 
+function fixRpFilter() {
+    # Mariner AKS CIS Benchmark: 3.3.7 Ensure Reverse Path Filtering is enabled
+    # Comment out rp_filter settings in /usr/lib/sysctl.d/50-default.conf and /lib/sysctl.d/50-default.conf
+    # We set these correctly in sysctl-d-60-CIS.conf, but they also need to not be present in the default file.
+    # These sed comands will comment out any line that starts with the settings we want to get rid of.
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.all.rp_filter/#net.ipv4.conf.all.rp_filter/g' /usr/lib/sysctl.d/50-default.conf
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.default.rp_filter/#net.ipv4.conf.default.rp_filter/g' /usr/lib/sysctl.d/50-default.conf
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.all.rp_filter/#net.ipv4.conf.all.rp_filter/g' /lib/sysctl.d/50-default.conf
+    sed -E -i 's/^[[:space:]]*net.ipv4.conf.default.rp_filter/#net.ipv4.conf.default.rp_filter/g' /lib/sysctl.d/50-default.conf
```
hbeberman, line 135: This is going to cause CIS scanning failures if systemd has been updated on the system. Our systemd package doesn't tag /usr/lib/sysctl.d/50-default.conf as a %config(noreplace) [though it probably should], so any upgrade to systemd is going to stomp the file entirely.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -412,6 +412,59 @@ testUserAdd() {
...
+        local escaped_setting=
+        echo "$test: Checking that setting $setting is commented out or not present in $settings_file"
+        escaped_setting=${setting//\./\\.}
+
+        # These settings must be commented-out or not present. So, filter out lines that are
+        # that are commented out and then any lines with the setting are errors.
+        # Disable shellcheck warning about using '$?' in an if statement because we don't want
+        # the return value later anyway and this is more clear.
+        local setting_lines=
+        setting_lines=$(grep -E -n -H "^[[:space:]]*${escaped_setting}" $settings_file)
```
SeanDougherty, line 451: Is the -n flag here to print out what line number the offending setting is on?

SeanDougherty (approved):

PR #3097, opened 2023-04-24
Remove executionMode from dualstack overlay conflist

ashvindeodhar:
**What type of PR is this?**

/kind cleanup

The conflist option of executionMode is not relevant in case of dualstack overlay. It is not used by dualstack overlay and as a result removing it to avoid any confusion as it's currently set to v4Swift.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


tyler-lloyd:
opened a PR with a branch from a non-forked repo and completed in https://github.com/Azure/AgentBaker/pull/3153

PR #3100, opened 2023-04-24
fix: set network settings according to cis recommendations

tobiasb-ms:
**What type of PR is this?**
/kind bug

We are out of compliance with certain CIS recommendations related to network parameter settings by changing and adding settings to the file that is installed to `/etc/sysctl.d/60-CIS.conf` to set the parameters to the correct values.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

You can find before and after versions of the files mentioned above for both mariner and ubuntu in the attached zip file.
[cis-fix-network-settings.zip](https://github.com/Azure/AgentBaker/files/11314684/cis-fix-network-settings.zip)

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/sysctl-d-60-CIS.conf
+++ b/parts/linux/cloud-init/artifacts/sysctl-d-60-CIS.conf
@@ -1,24 +1,49 @@
-# 3.1.2 Ensure packet redirect sending is disabled
+# Ubuntu CIS Benchmark: Ensure packet redirect sending is disabled
 net.ipv4.conf.all.send_redirects = 0
 net.ipv4.conf.default.send_redirects = 0
-# 3.2.1 Ensure source routed packets are not accepted 
+
+# Ubuntu CIS Benchmark: Ensure source routed packets are not accepted
+# Mariner AKS CIS Benchmark: Ensure source routed packets are not accepted
 net.ipv4.conf.all.accept_source_route = 0
 net.ipv4.conf.default.accept_source_route = 0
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/sysctl-d-60-CIS.conf
```diff
@@ -1,24 +1,49 @@
...
+# Ubuntu CIS Benchmark: Ensure IPv6 router advertisements are not accepted
+# Mariner AKS CIS Benchmark: Ensure IPv6 router advertisements are not accepted
 net.ipv6.conf.all.accept_ra = 0
 net.ipv6.conf.default.accept_ra = 0
-# 3.3.2 Ensure IPv6 redirects are not accepted
-net.ipv6.conf.all.accept_redirects = 0
-net.ipv6.conf.default.accept_redirects = 0
+
+# Mariner AKS CIS Benchmark: Ensure broadcast ICMP requests are ignored
+net.ipv4.icmp_echo_ignore_broadcasts = 1
```
alexeldeib, line 35: is this a behavior change for ubuntu? I believe azure blocks icmp from internet but I'd think this breaks ping node to node?

tobiasb-ms: Actually it isn't -- Ubuntu has this same setting.

alexeldeib (approved):
per offline discussion should be no change to ubuntu. lgtm

PR #3101, opened 2023-04-24
fix: change owner of crictl to root

tobiasb-ms:
**What type of PR is this?**
/kind bug

After downloading `crictl`, we install it by untarring it. However, since we're running as `root`, the normal default behavior of setting ownership of extracted files the current user does not apply, so we extract it with ownership from the tar file itself. That user and group are not guaranteed to exist (and in fact the user doesn't exist in this case). This creates a potential threat, because a user could be created with that same PID.

This change fixes that by setting the user and group of /usr/local/bin/crictl` to `root`. Note that other downloaded and installed files in this same script already do this -- this was an oversight.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

I also did some automatic reformatting of the file and normalized it to a four-space indentation (which is what most of the file and the other nearby files have already).

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_install.sh
+++ b/parts/linux/cloud-init/artifacts/cse_install.sh
@@ -30,29 +30,29 @@ cleanupContainerdDlFiles() {
 }
 
 installContainerRuntime() {
-if [ "${NEEDS_CONTAINERD}" == "true" ]; then
-    echo "in installContainerRuntime - KUBERNETES_VERSION = ${KUBERNETES_VERSION}"
-    wait_for_file 120 1 /opt/azure/manifest.json # no exit on failure is deliberate, we fallback below.
+    if [ "${NEEDS_CONTAINERD}" == "true" ]; then
+        echo "in installContainerRuntime - KUBERNETES_VERSION = ${KUBERNETES_VERSION}"
+        wait_for_file 120 1 /opt/azure/manifest.json # no exit on failure is deliberate, we fallback below.
 
-    local containerd_version
-    if [ -f "$MANIFEST_FILEPATH" ]; then
-        containerd_version="$(jq -r .containerd.edge "$MANIFEST_FILEPATH")"
+        local containerd_version
+        if [ -f "$MANIFEST_FILEPATH" ]; then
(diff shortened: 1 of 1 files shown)


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -30,29 +30,29 @@ cleanupContainerdDlFiles() {
 }
 
 installContainerRuntime() {
-if [ "${NEEDS_CONTAINERD}" == "true" ]; then
-    echo "in installContainerRuntime - KUBERNETES_VERSION = ${KUBERNETES_VERSION}"
-    wait_for_file 120 1 /opt/azure/manifest.json # no exit on failure is deliberate, we fallback below.
+    if [ "${NEEDS_CONTAINERD}" == "true" ]; then
```
alexeldeib, line 33: there's no actual diff here besides spacing, right? git seems to have choked on the line movement

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -186,13 +187,13 @@ installCNI() {
...
     fi
-    
+
     chown -R root:root $CNI_BIN_DIR
 }
 
 installAzureCNI() {
     CNI_TGZ_TMP=${VNET_CNI_PLUGINS_URL##*/} # Use bash builtin ## to remove all chars ("*") up to the final "/"
-    CNI_DIR_TMP=${CNI_TGZ_TMP%.tgz} # Use bash builtin % to remove the .tgz to look for a folder rather than tgz
+    CNI_DIR_TMP=${CNI_TGZ_TMP%.tgz}         # Use bash builtin % to remove the .tgz to look for a folder rather than tgz
```
alexeldeib, line 196: we really need a git stripspace on all files...

alexeldeib (approved):

alexeldeib:
need a `make` when ready :)

parts/linux/cloud-init/artifacts/cse_install.sh
tobiasb-ms: Thank you! I had completely forgotten to do this.

PR #3106, opened 2023-04-25
fix: ensure cron.allow exists and that it and cron.deny have appropri…

tobiasb-ms:
**What type of PR is this?**
/kind bug

Currently, we have an `/etc/cron.deny` file, but that actually opens up `cron` usage to any user not in that file. This is against CIS recommendations. We also find that permissions on the existing `/ect/cron.deny` are too permissions.

To fix this, we do two things.
1. Create an empty file `/etc/cron.allow`, which locks down `cron` usage.
2. Set appropriate permissions for both `/etc/cron.allow` and `/etc/cron.deny`.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cis.sh
+++ b/parts/linux/cloud-init/artifacts/cis.sh
@@ -61,6 +61,15 @@ assignFilePermissions() {
     for filepath in /etc/cron.hourly /etc/cron.daily /etc/cron.weekly /etc/cron.monthly /etc/cron.d; do
         chmod 0600 $filepath || exit $ERR_CIS_ASSIGN_FILE_PERMISSION
     done
+
+    # Docs: https://www.man7.org/linux/man-pages/man1/crontab.1.html
+    # If cron.allow exists, then cron.deny is ignored. To minimize who can use cron, we
+    # always want cron.allow and will default it to empty if it doesn't exist.
+    # We also need to set appropriate permissions on it.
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -61,6 +61,16 @@ assignFilePermissions() {
     for filepath in /etc/cron.hourly /etc/cron.daily /etc/cron.weekly /etc/cron.monthly /etc/cron.d; do
         chmod 0600 $filepath || exit $ERR_CIS_ASSIGN_FILE_PERMISSION
     done
+
+    # Docs: https://www.man7.org/linux/man-pages/man1/crontab.1.html
+    # If cron.allow exists, then cron.deny is ignored. To minimize who can use cron, we
+    # always want cron.allow and will default it to empty if it doesn't exist.
+    # Both need to have appropriate permissions.
+    touch /etc/cron.allow || exit $ERR_CIS_ASSIGN_FILE_PERMISSION
```
alexeldeib, line 69: so do we even need cron.deny?

alexeldeib (approved):
lgtm but curious why we need both

parts/linux/cloud-init/artifacts/cis.sh
tobiasb-ms: spoke offline and decided to drop `cron.deny`

PR #3112, opened 2023-04-26
chore: skip abe2e gpu scenarios in github actions

cameronmeissner:
**What type of PR is this?**

/kind flake

skip gpu tests in abe2e run by gh actions

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.github/workflows/check-coverage.yml
+++ b/.github/workflows/check-coverage.yml
@@ -64,6 +64,22 @@ jobs:
           GOCOVERDIR=covdatafiles ./baker start &
       - name: Run e2e tests
         run: |
+          export SCENARIOS_TO_RUN=ubuntu1804,\
+          ubuntu2204,\
+          marinerv1,\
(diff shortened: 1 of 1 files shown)


anujmaheshwari1 (approved):

alexeldeib (approved):
do we want to do this the opposite way, for the greater coverage here? I thought we preferred GH to ADO?

cameronmeissner:
> do we want to do this the opposite way, for the greater coverage here? I thought we preferred GH to ADO?

im open to keeping this as is, I've already phased the gpu tests out on the ADO side.

folks were actually experiencing timeouts when running _all_ scenarios on the gh actions (after the wasm additions) so maybe bumping the timeout here specifically could work as well

wondering if the gh action runners have less cores than the ADO agents

Issue #2815, opened 2023-02-28
Explicit configuration interface


maxwolffe:
Moving some private discussion I've had with @alexeldeib into a public ticket to increase bus factor. 

**Is your feature request related to a problem?/Why is this needed**
Our team uses a custom AKS image which has a few dependencies which are currently provided by AgentBaker's customnodedata:

We currently depend on the following files to be provided via [nodecustomdata](https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/nodecustomdata.yml):
/etc/default/kubelet
/var/lib/kubelet/bootstrap-kubeconfig 
/etc/kubernetes/certs/ca.crt

Within those files we depend on:
/etc/default/kubelet:
- node labels
- kubelet command line parameters

/var/libe/kubelet/bootstrap-kubeconfg:
- cluster-server:
```
name: localcluster
  cluster:
    certificate-authority: /etc/kubernetes/certs/ca.crt
    server: https://dev-azure-westus-xxx-cc782afe.hcp.westus.azmk8s
```
- token 
```
- name: kubelet-bootstrap
  user:
    token: "sbiizf.avcjfgfj5h3oni"
```

- ca.crt  - we depend on the whole file for the bootstrap kube-config.

We explicitly prevent the CustomScriptExtension from running by touching the `/opt/azure/containers/provision.complete` file which CSE checks prior to running. We don't want CSE to run because it does node level configuration which conflicts with our own. 

**Describe the solution you'd like in detail**
Ideally, the interface in AgentBaker we come up with:
- Allows us to test that these files are present prior to CSE running 
- Allows us to confirm that CSE does not run if the following file is already set: /opt/azure/containers/provision.complete 
 
There are a few options which stand out to me:

1. Write a test just confirming that those files are populated in nodedata
2. Write a new file interface (bootstrap.cfg) which either:
    a. Has those files included in it
    b. Has the fields we need only included
3. Some mechanism to fetch these dynamically from within the node, so we can completely remove the dependency on data provided by AgentBaker (though we’d shift the dependency to requiring that the fields are accessible somehow)

After discussions with @alexeldeib - there's a preference to *not* have the data provided via customdata. 

**Describe alternatives you've considered**

**Additional context**
We've had a number of incidents due to us not having a clear contract around node configuration, so hoping to work with y'all to get one defined. Thanks in advance!

juan-lee:
[...]

PR #3120, opened 2023-04-27
feat: add artifact upload for coverage e2e

UtheMan:
**What type of PR is this?**

Ensures no regression for #2815 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.github/workflows/check-coverage.yml
+++ b/.github/workflows/check-coverage.yml
@@ -67,6 +67,11 @@ jobs:
           echo "Running e2e tests"
           cd e2e
           go test -e2eMode="coverage" -timeout 60m -v -run Test_All ./
+      - name: Upload e2e artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: scenario-logs
+          path: e2e/scenario-logs
       - name: Generate coverage report
         run: |
           kill $(pgrep baker)


.github/workflows/check-coverage.yml
```diff
@@ -67,6 +67,11 @@ jobs:
           echo "Running e2e tests"
           cd e2e
           go test -e2eMode="coverage" -timeout 60m -v -run Test_All ./
+      - name: Upload e2e artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: scenario-logs
+          path: e2e/scenario-logs
```
cameronmeissner, line 74: something I was meaning to do on the ADO side - can we upload a new bundle that doesn't include scenario-logs/*/vmssId.txt? those text files are only there such that we can ensure the ephemeral VMSS are deleted (we have an 'ensure vmss deletion step in the ADO pipeline that reads the ids from those text files and does 'az delete --ids ...' on them

I'm also realizing now that we don't have the same sort of 'ensure vmss deletion' step on the GH action side...

we should really choose one and pivot towards that

cameronmeissner: might also be worth leaving those text files lying around so we can use them to manually delete stuff via az cli if needed....

I'll leave it up to you

cameronmeissner (approved):

.github/workflows/check-coverage.yml
cameronmeissner: (can also just go use the portal to delete... :D)

UtheMan: I will be looking into how i could combine ADO running e2e and gh action taking results of that and handling everything else

alexeldeib (approved):

PR #3125, opened 2023-04-27
fix: ensure sshd configuration is correct and sshd_config has proper permissions

tobiasb-ms:
**What type of PR is this?**
/kind bug

CIS recommends restricting access to ssh and ensuring restrictive permissions on `/etc/ssh/sshd_config` itself. This change prevents certain known users from `ssh`-ing into a machine and fixes the permissions on the file.

I also took the opportunity to shore up what `sshd_config` files are referenced by the mariner-specific vhd config files and how we choose which config to us for ubuntu configs.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/packer_source.sh
+++ b/vhdbuilder/packer/packer_source.sh
@@ -91,7 +91,7 @@ copyPackerFiles() {
   KUBELET_SERVICE_DEST=/etc/systemd/system/kubelet.service
   VHD_CLEANUP_SCRIPT_SRC=/home/packer/cleanup-vhd.sh
   VHD_CLEANUP_SCRIPT_DEST=/opt/azure/containers/cleanup-vhd.sh
-  
+
   CSE_REDACT_SRC=/home/packer/cse_redact_cloud_config.py
   CSE_REDACT_DEST=/opt/azure/containers/provision_redact_cloud_config.py
   cpAndMode $CSE_REDACT_SRC $CSE_REDACT_DEST 0744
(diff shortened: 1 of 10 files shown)


vhdbuilder/packer/packer_source.sh
```diff
@@ -186,10 +186,13 @@ copyPackerFiles() {
 
   NOTICE_SRC=/home/packer/NOTICE.txt
   NOTICE_DEST=/NOTICE.txt
-  if [[ ${UBUNTU_RELEASE} == "16.04" ]]; then
-    SSHD_CONFIG_SRC=/home/packer/sshd_config_1604
-  elif [[ ${UBUNTU_RELEASE} == "18.04" && ${ENABLE_FIPS,,} == "true" ]]; then
-    SSHD_CONFIG_SRC=/home/packer/sshd_config_1804_fips
+
+  if [[ $OS != $MARINER_OS_NAME ]]; then
```
alexeldeib, line 190: is this condition actually necessary?

vhdbuilder/packer/packer_source.sh
```diff
@@ -262,6 +265,8 @@ copyPackerFiles() {
 }
 
 cpAndMode() {
-  src=$1; dest=$2; mode=$3
+  src=$1
```
alexeldeib, line 268: nit - for the sake of reviews I appreciate avoiding mixing purely stylistic changes and functional changes. I'm happy to take style changes just makes reviewing real diffs harder.

alexeldeib (approved):
lgtm although I think that is an extra conditional, appreciate removing it

nice cleanup on the templates :)

vhdbuilder/packer/packer_source.sh
tobiasb-ms: No, but I do feel that it makes it more clear. Perhaps flipping the if would add even more clarity -- if ubuntu look at these things.

vhdbuilder/packer/packer_source.sh
tobiasb-ms: You're right of course and in my heart I agree. I'll leave this one since the damage is done but next time I'll do a separate PR.

vhdbuilder/packer/packer_source.sh
tobiasb-ms: No. I'll remove.

PR #3131, opened 2023-04-28
feat: allow skipping nvidia gpu driver install

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -91,7 +91,15 @@ if [[ ${ID} != "mariner" ]]; then
     logs_to_events "AKS.CSE.removeManDbAutoUpdateFlagFile" removeManDbAutoUpdateFlagFile
 fi
 
-if [[ "${GPU_NODE}" != "true" ]]; then
+export -f should_skip_nvidia_drivers
+skip_nvidia_driver_install=$(retrycmd_if_failure_no_stats 10 1 10 bash -c should_skip_nvidia_drivers)
+ret=$?
+if [[ "$ret" != "0" ]]; then
```
alexeldeib, line 97: should this be terminal? I think yes if we call it a real contract.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -329,4 +329,14 @@ logs_to_events() {
       return $ret
     fi
 }
+
+should_skip_nvidia_drivers() {
+    body=$(curl -fsSL -H "Metadata: true" --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01")
```
alexeldeib, line 334: kinda gross but not sure if there's a better way to trap the curl exit code while also getting the body parsed out?

cameronmeissner: so this is working off of tags present in the instance metadata?

cameronmeissner: yep, clarified offline

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -91,7 +91,15 @@ if [[ ${ID} != "mariner" ]]; then
     logs_to_events "AKS.CSE.removeManDbAutoUpdateFlagFile" removeManDbAutoUpdateFlagFile
 fi
 
-if [[ "${GPU_NODE}" != "true" ]]; then
+export -f should_skip_nvidia_drivers
+skip_nvidia_driver_install=$(retrycmd_if_failure_no_stats 10 1 10 bash -c should_skip_nvidia_drivers)
```
cameronmeissner, line 95: huh, haven't seen this pattern before, I guess this is the closest thing we have to "returning a bool"

probably better than trying to reuse exit codes and such

alexeldeib: yeah, this threw me for a loop...stdout vs exit codes and subshells is funky.

cameronmeissner: feel like the only other (probably less) sane way to do it would be to use a global var and have it should_skip_nvidia_drivers just mutate it

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -135,7 +143,7 @@ fi
 REBOOTREQUIRED=false
 
 echo $(date),$(hostname), "Start configuring GPU drivers"
-if [[ "${GPU_NODE}" = true ]]; then
+if [[ "${GPU_NODE}" = true ]] && [[ "${skip_nvidia_driver_install}" != "true" ]]; then
```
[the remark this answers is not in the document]
alexeldeib: yes: you're not using GPU VM sizes

that's an expected scenario

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: you mean having an explicit exit code for failure during this step?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -91,7 +91,15 @@ if [[ ${ID} != "mariner" ]]; then
...
-if [[ "${GPU_NODE}" != "true" ]]; then
+export -f should_skip_nvidia_drivers
+skip_nvidia_driver_install=$(retrycmd_if_failure_no_stats 10 1 10 bash -c should_skip_nvidia_drivers)
+ret=$?
+if [[ "$ret" != "0" ]]; then
+    echo "Failed to determine if nvidia driver install should be skipped"
+    exit $ERR_NVIDIA_DRIVER_INSTALL
+fi
+
+if [[ "${GPU_NODE}" != "true" ]] || [[ "${skip_nvidia_driver_install}" == "true" ]]; then
```
cameronmeissner, line 102: would we still want to cleanup in the case where GPU_NODE is true and skip_nvidia_driver_install is also true?

cameronmeissner: oh nvm i can't read

parts/linux/cloud-init/artifacts/cse_main.sh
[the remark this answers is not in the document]
cameronmeissner: ah okay that makes sense, since skip_nvidia_driver_install would default to false... duh

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: yes -- these files should never be left on disk if unused.

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: I might try moving the jq command separate from the curl retry, since that shouldn't fail, but I kind of like wrapping them together and that doesn't really solve the stdout/exit code issue...might just leave it.

alexeldeib:
nice. should try to run nvidia-smi and assert it fails.
```bash
+ echo Fri Apr 28 19:51:41 UTC 2023,abtestsbhw000000, 'Start configuring GPU drivers'
Fri Apr 28 19:51:41 UTC 2023,abtestsbhw000000, Start configuring GPU drivers
+ [[ true = true ]]
+ [[ true != \t\r\u\e ]]
++ date
++ hostname
+ echo Fri Apr 28 19:51:41 UTC 2023,abtestsbhw000000, 'End configuring GPU drivers'
Fri Apr 28 19:51:41 UTC 2023,abtestsbhw000000, End configuring GPU drivers
```

e2e/scenario/scenario_ubuntu2204-gpu-nodriver.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+)
+
+func ubuntu2204gpuNoDriver() *Scenario {
+	return &Scenario{
+		Name:        "ubuntu2204_gpuNoDriver",
```
cameronmeissner, line 11: can we rename to something like "ubuntu2204-gpu-nodriver" to keep format consistent scenario name formatting? 

also open to suggestions for changing the naming format altogether

alexeldeib: yep, was not 100% sure what was desired, will adust

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: should I call the `exit` at all, or let CSE keep going but install the drivers by default

I think fail, if we'll have users relying on this

pkg/agent/baker.go
```diff
@@ -1048,3 +1060,72 @@ root = "{{GetDataDir}}"{{- end}}
...
 `
+
+// this pains me, but to make it respect mutability of vmss tags,
+// we cannot use go templates at runtime.
+// CSE needs to be able to generate the full config, with all params,
+// with the tags pulled from wireserver. this is a hack to avoid
+// moving all the go templates to CSE -- we allow two options,
+// duplicate them in CSE base64-encoded, and pick the right one.
+// they're identical except for GPU runtime class.
+const containerdConfigNoGpuTemplateString = `version = 2
```
cameronmeissner, line 1071: I guess we can't just add extra parameterization on the existing template since whether or not we want to skip the driver installation is based on the instance data rather than the bootstrapping config right?

e2e/scenario/scenario_ubuntu2204-gpu-nodriver.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+)
+
+func ubuntu2204gpuNoDriver() *Scenario {
+	return &Scenario{
+		Name:        "ubuntu2204-gpu-nodriver",
+		Description: "Tests that a node using the Ubuntu 2204 VHD can be properly bootstrapped",
```
cameronmeissner, line 12:
```suggestion
		Description: "Tests that a GPU-enabled node using the Ubuntu 2204 VHD opting for skipping gpu driver installation can be properly bootstrapped",
```

pkg/agent/baker.go
alexeldeib: right -- we can move all of the parameterization itself into bash, but that's even more tedious.

i'd love to write a standalone python/go/or whatever helper for this honestly, just low priority

e2e/scenario/validators.go
```diff
@@ -34,3 +34,16 @@ func SysctlConfigValidator(customSysctls map[string]int) *LiveVMValidator {
 		},
 	}
 }
+
+func NvidiaSMINotInstalledValidater() *LiveVMValidator {
```
cameronmeissner, line 38: cool!

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: left this as-is for now, resolving unless you have thoughts

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: yeah I think hard failure makes more sense

e2e/scenario/validators.go
ganeshkumar5699: good check indeed!

pkg/agent/baker.go
```diff
@@ -673,6 +673,18 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return base64.StdEncoding.EncodeToString(b.Bytes())
 		},
+		"GetContainerdConfigNoGPUContent": func() string {
+			parameters := getParameters(config)
```
cameronmeissner, line 677: nit: consider extracting to a helper?

cameronmeissner (approved):

pkg/agent/baker.go
ganeshkumar5699: Is it possible to copy this from the original file instead of a full copy-paste? There's no GPU specific difference here right?

alexeldeib: there is -- I remove a large piece. see L1070 above. no default_runtime_class = "nvidia-container-runtime" or configuration for that at all

I could leave the nvidia container runtime but thee's still no great way to default it

cameronmeissner (approved):

PR #3139, opened 2023-05-01
fix: var expansion in ensureRunc

cameronmeissner:
**What type of PR is this?**

/kind bug

fixes variable expansion such that correct runc version is installed during VHD builds, and thus installContainerRuntime will behave as expected during CSE time

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


ganeshkumar5699 (commented):
lgtm. this function will be heavily refactored soon though

PR #3104, opened 2023-04-25
chore: add test checking for presence of some of the default files when using TLSBootstrapping

UtheMan:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #3144, opened 2023-05-02
test: Change test to check for kubelet.service file existence as well. 

maxwolffe:
**What type of PR is this?**
/kind test

This PR adds a little more context to the test added in https://github.com/Azure/AgentBaker/pull/3104 for issue #2815 

The issue #2815 describes the full set of currently required data for custom images which don't rely on the CSE script to execute. 

Those are:
- /etc/default/kubelet
- /var/lib/kubelet/bootstrap-kubeconfig 
- /etc/kubernetes/certs/ca.crt

The Databricks usecase also has a silly dependency on kubelet.service which I have removed and will roll out in the next month. I've added it as a required file for now, but happy to not if it's possible to otherwise prevent it from being removed in the near term? 

`/opt/azure/containers/kubelet.sh` also shouldn't be required, but currently is because it's referenced by the kubelet.service. 

None

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

First PR in this repo so apologies if I'm missing something / happy to make changes. 

I've confirmed that tests failed first, then passed with this change. 

I have a potentially dumb question about TLSBootstrapping being enabled. Under what conditions is it NOT enabled? 

Found this - https://github.com/Azure/AgentBaker/blob/a7df3d609eac4ad608ccd3e84ba193565c236243/pkg/agent/utils.go#L355 - but perhaps someone has context already? 

Thanks!

**Release note**:
```
none
```


alexeldeib (approved):
lgtm

maxwolffe:
@alexeldeib - updated the commit messages to pass linting - could you re-approve tests at your convenience?

Also - e2e tests failed on the last run due to az login failed - any other follow up I need to make there?

alexeldeib:
unfortunately we have some annoying rbac issues running the e2e/generated tests from forks today :(

the lint is fine after rebase, I'm about to brea kit again but will squash-merge and ignore the other bit for now.

alexeldeib:
> I have a potentially dumb question about TLSBootstrapping being enabled. Under what conditions is it NOT enabled?

tl;dr - always today. the token itself is generated by RP, so it's possible the field in the struct is empty. but that's never the case today anymore afaik.

PR #3169, opened 2023-05-08
chore: update gk v3.11.1 and addon/webhook 1.0.1

anlandu:
**What type of PR is this?**
/kind cleanup
GK v3.11.0 and addon/webhook v20230208.1 have [CVE-2022-41721](https://www.cve.org/CVERecord?id=CVE-2022-41721). Moving to patched versions, will remove images with CVEs after AKS tag update goes through to avoid failing AKS checkin VHD checks for other people.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


ganeshkumar5699 (approved):
gatekeeper is only run when enabled by this addon right? https://learn.microsoft.com/en-us/azure/governance/policy/concepts/policy-for-kubernetes

And how large is this image (approximate Mbs)?

I wonder if in the future, we can skip downloading these images in Agentbaker

PR #3180, opened 2023-05-10
chore: remove container runtime flag for k8s >= 1.27

anujmaheshwari1:
**What type of PR is this?**
k8s 1.27 removes the flag --container-runtime, which we pass to kubelet during invocation. This PR handles the change, where it uses 2 drop-in files, one with the base flags that are still valid, and one with the container-runtime flag.

The drop-in with container-runtime flag is added if k8s version < 1.27

Tested the change with RP e2e, succesful run: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=73087068&view=results

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -264,10 +264,22 @@ if [ "${NEEDS_CONTAINERD}" == "true" ]; then
...
 [Service]
-Environment="KUBELET_CONTAINERD_FLAGS=--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock --runtime-cgroups=/system.slice/containerd.service"
+Environment="KUBELET_CONTAINERD_FLAGS=--runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock --runtime-cgroups=/system.slice/containerd.service"
 EOF
+    
+    # if k8s version < 1.27.0, add the drop in for --container-runtime flag
+    if ! semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.27.0"; then
+        tee "/etc/systemd/system/kubelet.service.d/10-container-runtime-flag.conf" > /dev/null <<'EOF'
+[Service]
+Environment="KUBELET_CONTAINERD_FLAGS=--container-runtime=remote"
```
alexeldeib, line 280: you need a different var name here, and you need to add the new var name to kubelet.service.

anujmaheshwari1: good catch, updated

alexeldeib (approved):

PR #3183, opened 2023-05-11
cache cns v1.4.44_hotfix

camrynl:
**What type of PR is this?**

cache the newest fix for cns

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


wedaly (approved):

cameronmeissner (approved):

cameronmeissner:
@alexeldeib

anujmaheshwari1 (approved):

vhdbuilder/packer/components.json
```diff
@@ -35,8 +35,7 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cns:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.4.39",
-        "v1.4.44"
+        "v1.4.44_hotfix"
```
ganeshkumar5699, line 38: Why is this not in semver format? Typically we add the hotfix version (like .0) as the suffix and not the "hotfix" string

rbtr: CNS is built out of a monorepo, and the version is not really semver. The release tag on the repo has already moved past 1.4.45, we need to indicate that this is a single fix on the 1.4.44 revision, and I don't think 1.4.44.1 was a viable option, thus the _hotfix.

PR #3199, opened 2023-05-16
fix: dns timeout to respect resolve.conf 5s timeout with retries

alexeldeib:
**What type of PR is this?**

default resolv.conf timeout is 5 sec with 3 retries, meaning a single nslookup takes 15 sec on the first dns server.

if a user has multiple DNS servers, and the first one is down, we will never try the 2nd+ one. in theory they could have a bunch (3+) but this is a quick fix.

we also appear to do sequential v4/v6 queries with nslookup and wait for both, which is not really necessary, but not sure how to easily issue both.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -321,7 +321,7 @@ if ! [[ ${API_SERVER_NAME} =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
        API_SERVER_DNS_RETRIES=200
     fi
     if [[ "${ENABLE_HOSTS_CONFIG_AGENT}" != "true" ]]; then
-        RES=$(logs_to_events "AKS.CSE.apiserverNslookup" "retrycmd_if_failure ${API_SERVER_DNS_RETRIES} 1 10 nslookup ${API_SERVER_NAME}")
+        RES=$(logs_to_events "AKS.CSE.apiserverNslookup" "retrycmd_if_failure ${API_SERVER_DNS_RETRIES} 1 20 nslookup -timeout=5 -retry=0 ${API_SERVER_NAME}")
```
ganeshkumar5699, line 324: Do we know how many DNS servers on average, an AKS cluster/API server has?

ganeshkumar5699 (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
alexeldeib: max 3 respected fully, so this hits all of them

5 sec retry, 1x per server (no retries), with 3 servers -> 15 sec

PR #3206, opened 2023-05-17
fix: avoid nvidia latency issues by enabling nvidia-persistenced systemd service

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/nvidia-persistenced.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=NVIDIA Persistence Daemon
+Wants=syslog.target
+
+[Service]
+Type=forking
```
marwanad, line 6: 🔑

parts/linux/cloud-init/artifacts/nvidia-persistenced.service
```diff
@@ -0,0 +1,13 @@
...
+ExecStart=/usr/bin/nvidia-persistenced --verbose
+ExecStopPost=/bin/rm -rf /var/run/nvidia-persistenced
+PIDFile=/var/run/nvidia-persistenced/nvidia-persistenced.pid
```
alexeldeib, line 9: self-nit: is the pid file usable if it's deleted by ExecStopPost? which happens first?

ganeshkumar5699 (approved):

alexeldeib:
actually...I should do this inside https://github.com/Azure/aks-gpu/blob/main/install.sh

then it will always apply, and it'll be a little simpler.

parts/linux/cloud-init/artifacts/nvidia-persistenced.service
alexeldeib: I got rid of this, it's not in upstream template any more

alexeldeib:
superceded by https://github.com/Azure/AgentBaker/pull/3209

PR #3218, opened 2023-05-19
feat(SIG): Community Image Gallery Support for VHD Builder, and Karpe…

Bryce-Soghigian:
…nter VHD Pipeline

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/init-variables.sh
```diff
@@ -119,7 +125,7 @@ if [[ "$MODE" == "linuxVhdMode" || "$MODE" == "windowsVhdMode" ]]; then
 	id=$(az sig show --resource-group ${AZURE_RESOURCE_GROUP_NAME} --gallery-name ${SIG_GALLERY_NAME}) || id=""
 	if [ -z "$id" ]; then
 		echo "Creating gallery ${SIG_GALLERY_NAME} in the resource group ${AZURE_RESOURCE_GROUP_NAME} location ${AZURE_LOCATION}"
-		az sig create --resource-group ${AZURE_RESOURCE_GROUP_NAME} --gallery-name ${SIG_GALLERY_NAME} --location ${AZURE_LOCATION}
+		az sig create --resource-group ${AZURE_RESOURCE_GROUP_NAME} --gallery-name ${SIG_GALLERY_NAME} --location ${AZURE_LOCATION} ${GALLERY_PERMISSIONS_COMMAND} ${}
```
alexeldeib, line 128: I don't think this will do what you want. this will not make the final image a community gallery image. this is merely the intermediate SIG that packer builds to.

you need to change the publishing logic in RP, which re-hydrates our prod SIG from a blob. there's a lot of weird steps today...

vhdbuilder/packer/init-variables.sh
```diff
@@ -8,6 +8,12 @@ SUBSCRIPTION_ID="${SUBSCRIPTION_ID:-$(az account show -o json --query="id" | tr
 CREATE_TIME="$(date +%s)"
 STORAGE_ACCOUNT_NAME="aksimages${CREATE_TIME}$RANDOM"
 
+GALLERY_PERMISSIONS_COMMAND = "--permissions ${IMAGE_GALLERY_PERMISSIONS}"
+# if IMAGE_GALLERY_TYPE is not set, default to SharedImageGallery 
+if [[ "${IMAGE_GALLERY_PERMISSIONS}" != "community" || -z "${IMAGE_GALLERY_PERMISSIONS}" ]]; then
```
cameronmeissner, line 13: nit:
```suggestion
if [[ "${IMAGE_GALLERY_PERMISSIONS}" != "community" ]] || [[ -z "${IMAGE_GALLERY_PERMISSIONS}" ]]; then
```

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -73,6 +73,7 @@ steps:
         -e IMG_SKU=${IMG_SKU} \
         -e IMG_VERSION=${IMG_VERSION} \
         -e MODE=$(MODE) \
+        -e IMAGE_GALLERY_PERMISSIONS=${IMAGE_GALLERY_PERMISSIONS} \
```
cameronmeissner, line 76: are you planning on setting this as a pipeline var?

vhdbuilder/packer/init-variables.sh
cameronmeissner: yeah... problem is that our actual SIG publish happens in the weekly release pipeline rather than during the build here

current flow is

AgentBaker/packer VHD build -> intermediate SIG -> intermediate blob (<something>.vhd) -> another intermediate blob (prefetch optimized) -> (for AGC do blob copy to correct tenants/subs) -> managed image -> final prod SIG

PR #3221, opened 2023-05-22
fix: redirect curl verbose output to file, cat on failure

anujmaheshwari1:
**What type of PR is this?**
We use curl -v to get verbose output on curl execution. However, it also generates a lot of noise and makes debugging VHD logs difficult. This PR pipes the curl -v output to a file and only displays it on failure.

Successful test run: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=73732770&view=results

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -143,7 +144,7 @@ retrycmd_get_tarball() {
         if [ $i -eq $tar_retries ]; then
             return 1
         else
-            timeout 60 curl -fsSLv $url -o $tarball
+            timeout 60 curl -fsSLv $url -o $tarball 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && cat $CURL_OUTPUT
```
anujmaheshwari1, line 147: 2>&1 - to combine stderr and stdout
tee $CURL_OUTPUT >/dev/null - to only write output to file and not console because it shows up in VHD logs
grep -E "^(curl:.*)|([eE]rr.*)$" - regex to check for err/Err and curl:
cat the file if the above regex matches.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -156,7 +157,7 @@ retrycmd_curl_file() {
         if [ $i -eq $curl_retries ]; then
             return 1
         else
-            timeout $timeout curl -fsSLv $url -o $filepath
+            timeout $timeout curl -fsSLv $url -o $filepath 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && cat $CURL_OUTPUT
```
cameronmeissner, line 160: wonder if it would be worth it to add a helper function specifically for curl commands so we don't have to worry about setting up logging correctly everywhere we use it

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
anujmaheshwari1: fair point, however, our usage of curl varies w.r.t args and exit codes. for eg, we have an exit code for curl'ing wasm shim binaries for a timeout but none for curling tarballs and curl files. true though, that
`tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && cat $CURL_OUTPUT` can be common and parsed out but not sure if worth the effort?

parts/linux/cloud-init/artifacts/cse_helpers.sh
anujmaheshwari1: updated!

alexeldeib (approved):
you can also break out of the loop as in retrycmd_if_failure_no_stats but I like the explicit vs implicit

PR #3223, opened 2023-05-23
fix: add dnf_update to pre-install for Mariner to address kernel upgr…

SeanDougherty:
…ade issues

**What type of PR is this?**

Adding dnf_update to pre-install dependencies, which is before the reboot, so that the kernel is properly updated.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

[Pipeline runs](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=73678641&view=results)
[alternate pipeline run](https://dev.azure.com/mariner-org/mariner/_build/results?buildId=365325&view=results)

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/pre-install-dependencies.sh
+++ b/vhdbuilder/packer/pre-install-dependencies.sh
@@ -50,6 +50,9 @@ systemctlEnableAndStart sync-container-logs.service || exit 1
 
 # First handle Mariner + FIPS
 if [[ ${OS} == ${MARINER_OS_NAME} ]]; then
+  dnf_makecache || exit $ERR_APT_UPDATE_TIMEOUT
+  dnf_update || exit $ERR_APT_DIST_UPGRADE_TIMEOUT
+
   if [[ "${ENABLE_FIPS,,}" == "true" ]]; then
     # This is FIPS install for Mariner and has nothing to do with Ubuntu Advantage
     echo "Install FIPS for Mariner SKU"


hbeberman (approved):

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -50,6 +50,9 @@ systemctlEnableAndStart sync-container-logs.service || exit 1
 
 # First handle Mariner + FIPS
 if [[ ${OS} == ${MARINER_OS_NAME} ]]; then
+  dnf_makecache || exit $ERR_APT_UPDATE_TIMEOUT
```
anujmaheshwari1, line 53: I wonder if this should be after the fips install is done?

SeanDougherty: + @dmcilvaney @rmhsawyer for their insight.

For context:
We are placing a call to dnf_makecache and dnf_update in pre-install-dependencies to trigger updating to the latest kernel before the reboot.

Question:
Should we wait until FIPS is enabled and the FIPS related grubby and dracut-fips packages are installed before updating?

SeanDougherty: @dmcilvaney or @rmhsawyer to comment

Is there a risk to installing an updated kernel before installingFIPS _and_ before rebooting into said newly updated kernel?

dmcilvaney: Generally all of Mariner's kernels and openssl are fips 'compliant' (certified means a specific version of the .rpm). So long as the grub cmdline is set correctly, and the dracut-fips package remains installed I don't see any issues with how we handle the kernel etc. packages.

dmcilvaney (approved):

anujmaheshwari1 (approved):
lgtm!

PR #3224, opened 2023-05-23
fix: mask nfs-server

tobiasb-ms:
**What type of PR is this?**
/kind bug

CIS recommends that we not install `nfs-server` and that if we must we mask it by default. On ubuntu systems, it isn't installed but on mariner systems it's required as a dependency but we don't want to run it, and it's disabled by default. This changes masks `nfs-server` if it's installed.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):

alexeldeib:
@andyzhangx for a quick sanity check as well?

andyzhangx (approved):

andyzhangx:
this change lgtm, and we also have nfs e2e test in aks rp, that should also avoid the regression

PR #3225, opened 2023-05-23
feat: add audit of installed Mariner packages to track unexpected dep…

SeanDougherty:
…endency additions

**What type of PR is this?**
/kind feat

This PR adds a check for Mariner builds after installing our dependencies to check that we are only installing packages we know need to be part of the container host image. If an unexpected package is found installed on the Mariner image, the likely cause is a new dependency for one of the packages used by the container host. If the package is suitable to be added, then it can be appended to the package allow list. If however, the package seems to be dubious, this gives us signal investigate further.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner:
feel free to reopen if still needed

PR #3235, opened 2023-05-24
fix: add reboot into correct kernel for mariner

alexeldeib:
this primarily affects the mariner kata image.
we accidentally remove the kernel in post-install without a reboot.

**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/test/run-test.sh
+++ b/vhdbuilder/packer/test/run-test.sh
@@ -103,12 +103,29 @@ if [ "$OS_TYPE" == "Linux" ]; then
   fi
 
   SCRIPT_PATH="$CDIR/$LINUX_SCRIPT_PATH"
-  for i in $(seq 1 3); do
-    ret=$(az vm run-command invoke --command-id RunShellScript \
+  retries=3
+  for i in $(seq 1 $retries); do
+    output=$(az vm run-command invoke --command-id RunShellScript \
       --name $VM_NAME \
       --resource-group $RESOURCE_GROUP_NAME \
       --scripts @$SCRIPT_PATH \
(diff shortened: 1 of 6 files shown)


hbeberman (approved):

cameronmeissner (approved):

anujmaheshwari1 (approved):

alexeldeib:
issue is primarily due to calling enableMarinerKata and then cleaning up kernels somewhat indiscriminately such that we reboot into a non-existing (?) kernel.

we never actually had a reboot even before https://github.com/Azure/AgentBaker/commit/27592e5ebdef0cbd46e0da840ff9a86d57df1d8c, but the cleanup added by https://github.com/Azure/AgentBaker/commit/c0612b842b6a060db124793cd29a169386fbc9df needs one to do the right thing (we hope)

https://github.com/Azure/AgentBaker/blob/4a7ba42d7abe6b55d069e9c2b3f10da2a958b0b9/vhdbuilder/packer/install-dependencies.sh#L102 -> enable kata

https://github.com/Azure/AgentBaker/blob/4a7ba42d7abe6b55d069e9c2b3f10da2a958b0b9/vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh#L128-L135 -> swapping grub cfg (applied to _next_ boot, but no reboot occurs)

https://github.com/Azure/AgentBaker/blob/4a7ba42d7abe6b55d069e9c2b3f10da2a958b0b9/vhdbuilder/packer/post-install-dependencies.sh#L36-L39 -> kernel cleanup

PR #3240, opened 2023-05-24
fix: use vm size that supports nested virt for mariner kata post content tests

anujmaheshwari1:
**What type of PR is this?**
This PR updates the post content VHD tests to use vm size Standard_D4ds_v5 to enable tests for Mariner Kata.

 like 1

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/test/run-test.sh
```diff
@@ -76,7 +76,15 @@ else
...
     TARGET_COMMAND_STRING+="--size Standard_D2pds_v5"
-  elif [[ "${OS_TYPE}" == "Linux" && "${ENABLE_TRUSTED_LAUNCH}" == "True" ]]; then
+  elif [[ "${FEATURE_FLAGS,,}" == "kata" ]]; then
+    TARGET_COMMAND_STRING="--size Standard_D4ds_v5"
+  fi
+
+  if [[ "${OS_TYPE}" == "Linux" && "${ENABLE_TRUSTED_LAUNCH}" == "True" ]]; then
+    if [[ -n "$TARGET_COMMAND_STRING" ]]; then
+      # To take care of Mariner Kata TL images
+      TARGET_COMMAND_STRING+=" "
```
cameronmeissner, line 86: why do we ned this?

ganeshkumar5699, line 86: Just to understand: why is this being appended for Mariner Kata TL specifically?

vhdbuilder/packer/test/run-test.sh
```diff
@@ -76,7 +76,15 @@ else
...
+    fi
     TARGET_COMMAND_STRING+="--security-type TrustedLaunch --enable-secure-boot true --enable-vtpm true"
```
ganeshkumar5699, line 88: Could the extra whitespace be added here, at the start instead?

cameronmeissner (approved):

ganeshkumar5699 (approved):

vhdbuilder/packer/test/run-test.sh
anujmaheshwari1: kata + trusted launch is a valid combination, just not something we build today, need a whitespace before --security-type

anujmaheshwari1: addressed above

vhdbuilder/packer/test/run-test.sh
anujmaheshwari1: not my preference, it will be a random whitespace, in the above way, its clear why its added + its only for kata + TL

PR #3243, opened 2023-05-25
fix: change condition of uploading artifacts

junjiezhang1997:
**What type of PR is this?**
/kind fix

Change condition of uploading artifacts to avoid uploading artifacts when step fails.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/e2e-windows-template.yaml
+++ b/.pipelines/templates/e2e-windows-template.yaml
@@ -63,7 +63,7 @@ jobs:
       displayName: Setup Test Cluster
     - publish: $(System.DefaultWorkingDirectory)/${{ parameters.windowsImage }}
       artifact: ${{ parameters.windowsImage }}-clusterConfig
-      condition: always()
+      condition: succeeded()
 
 - job: Generate_Matrix
   steps:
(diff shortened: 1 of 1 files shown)


.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -111,4 +111,4 @@ jobs:
         MAPPED_ACCOUNT_KEY: $(STORAGE_ACCOUNT_KEY)
     - publish: $(System.DefaultWorkingDirectory)/${{ parameters.windowsImage }}/$(SCENARIO_NAME)-logs
       artifact: ${{ parameters.windowsImage }}-$(SCENARIO_NAME)-logs
-      condition: always()
\ No newline at end of file
+      condition: succeeded()
```
AbelHu, line 114: Does it mean that we will not upload logs if it failed? I thought that the logs would help us for troubleshooting.

junjiezhang1997: Thanks for the comment. I will keep the condition of cse logs as always() for debug.

AbelHu (approved):
LGTM. Thanks for the PR

PR #3262, opened 2023-05-31
fix: mask systemd-timesyncd in Mariner

SeanDougherty:
**What type of PR is this?**
/kind fix

We should mask timesyncd, chronyd fills the role of time synchronization for the AzureLinux container host.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

[pipeline run](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=74122990&view=results)

**Release note**:
```
none
```


hbeberman (approved):

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -89,6 +89,16 @@ disableDNFAutomatic() {
...
 }
 
+disableTimesyncd() {
+    # Disable and Mask timesyncd to prevent it from interfering with chronyd's work
+    systemctl stop systemd-timesyncd || exit 1
+    systemctl disable systemd-timesyncd || exit 1
+    systemctl mask systemd-timesyncd || exit 1
+    
+    # Before we return, make sure that chronyd is running
+    systemctlEnableAndStart chronyd || exit $ERR_SYSTEMCTL_START_FAIL
```
alexeldeib, line 99: isn't this already happening elsewhere?

alexeldeib (approved):

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
SeanDougherty: I see the disabling and stopping in tools_installs_ubuntu.sh and in init-aks-custom-cloud.sh

Do you know if init-aks-custom-cloud.sh runs on Mariner?

To another end, Ubuntu Marketplace already masks timesyncd so these stop and disable steps might be no-op for Ubuntu. 

Meanwhile, Mariner does not mask timesyncd. So we are adding in the masking to tool_installs_mariner... If init-aks-custom-cloud.sh runs on all VMs I could instead add this masking step to that script so that this maintains parity of some sort across all SKUs... Let me know what you think

alexeldeib: was thinking of https://github.com/Azure/AgentBaker/blob/636b2725ab6564b54d440039fc82d9079e87e8d2/vhdbuilder/packer/install-dependencies.sh#L77

we could indeed align these a bit more, but defer to you on the quick fix

SeanDougherty: It looks like that fn is heavily weighted to be Ubuntu specific, rewriting chrony.conf, etc... I'd be game to tackle bringing the Mariner and Ubuntu approach into alignment, but do agree that this quick fix is the way to go for now and we can visit the parity approach at a later date

PR #3264, opened 2023-06-01
fix: hold mshv-linuxloader to prevent mshv boot failures

alexeldeib:
**What type of PR is this?**

fix: hold mshv-linuxloader to prevent mshv boot failures

currently a dnf update triggers an upgrade of
mshv-linuxloader which breaks next boot of the VM.

avoid upgrading that package in the interim to avoid breakage.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib:
I opted to keep the reboot to force similar failures earlier. I also opted to leave the earlier dnf updates to deliberately trigger the behavior and ensure we avoid it.

the --exclude is kind of a heavy hammer given it runs on all azure linux skus, but I'm okay with it if others are.

hbeberman:
> the --exclude is kind of a heavy hammer given it runs on all azure linux skus, but I'm okay with it if others are.

No objection from me here, the Kata SKU is the only one with that package and it'll just be a no-op on everything else.

hbeberman (approved):

PR #3282, opened 2023-06-06
chore: set vnet resource group name based on pool name during vhd builds

cameronmeissner:
**What type of PR is this?**

so we can determine the vnet resource group name to be used within packer templates based on the specified 1ES hosted pool resource being used for the build

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -41,6 +41,7 @@ steps:
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
       -w /go/src/github.com/Azure/AgentBaker \
+      -e POOL_NAME=$(PROD_POOL_NAME) \
```
cameronmeissner, line 44: looks like POOL_NAME was recently change to PROD_POOL_NAME within windows yaml

vhdbuilder/packer/init-variables.sh
```diff
@@ -369,6 +369,13 @@ if [ "$OS_TYPE" == "Windows" ]; then
 	fi
 fi
 
+vnet_rg_name=""
+if [[ "${pool_name}" == *nodesigprod* ]]; then
+	vnet_rg_name="nodesigprod-agent-pool"
```
cameronmeissner, line 374: this seems to be the right format for windows/ubuntu pool types based on previous PRs? @junjiezhang1997 is this correct?

since those are in different subs I can't just go and check I don't think

alexeldeib (approved):

vhdbuilder/packer/init-variables.sh
junjiezhang1997: It's correct. Thanks for the PR!

junjiezhang1997 (approved):

PR #3288, opened 2023-06-07
chore: modify mariner and ubuntu sysctl ab E2E to verify new sysctl values

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/scenario/validators.go
```diff
@@ -23,17 +23,23 @@ func DirectoryValidator(path string, files []string) *LiveVMValidator {
...
+	keysToCheck := make([]string, len(customSysctls))
+	for k, _ := range customSysctls {
+		keysToCheck = append(keysToCheck, k)
+	}
+	// regex used in sed command to remove extra spaces between two numerical values, used to verify correct values for
+	// sysctls that have string values, e.g. net.ipv4.ip_local_port_range, which would be printed with extra spaces
 	return &LiveVMValidator{
 		Description: "assert sysctl settings",
-		Command:     "sysctl -a",
+		Command:     fmt.Sprintf("sysctl %s | sed -E 's/([0-9])\\s+([0-9])/\\1 \\2/g'", strings.Join(keysToCheck, " ")),
```
cameronmeissner, line 35: very annoying that we need this lol, though looks we do indeed if we use `sysctl`

e2e/scenario/scenario_ubuntu2204-custom-sysctls.go
```diff
@@ -35,7 +41,8 @@ func ubuntu2204CustomSysctls() *Scenario {
 				}
 			},
 			LiveVMValidators: []*LiveVMValidator{
-				SysctlConfigValidator(customSysctls),
```
cameronmeissner, line 38: can we potentially just use a map[string]string and do type conversion as needed so we only need to call this a single time? each validator equates to another command that needs to be executed on the remote vm, so the fewer the better in terms of flakiness

e2e/scenario/validators.go
cameronmeissner: ugh, the same issue seems to occur when `cat`ing out the files themselves as well...

UtheMan: I wonder why that happens, couldn't find an answer

e2e/scenario/modifiers.go
```diff
@@ -0,0 +1,11 @@
+package scenario
+
+import "strconv"
+
+func stringToInt32(s string) int32 {
+	i, err := strconv.ParseInt(s, 10, 32)
+	if err != nil {
+		panic(err)
```
UtheMan, line 8: Allows me to keep the int32 conversion just like it was in the past. Since its the test code anyway (that converts values that user added to their scenario), I figured its not a big deal if we panic here - at least you'll know fast that your scenario config is incorrect.

cameronmeissner: makes sense

e2e/scenario/validators.go
alexeldeib: `tr -s ' '` (squashes whitespace) may help

`cat -A` if that doesn't work, to check for special chars before trimming.

alexeldeib (approved):

cameronmeissner (approved):

PR #3293, opened 2023-06-07
refactor: check cloud-init status at the beginning of cse

anujmaheshwari1:
**What type of PR is this?**
We do a wait_for_file check for each file uploaded through custom data to make sure it is completely written to disk.
The same can be simply achieved by checking cloud init completion status very early in the provisioning script, thereby removing the need for EOF markers, that signify that the file wrote completely and using wait_for_file.
This also makes toggling self contained VHD behavior easier.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -1,13 +1,8 @@
...
-for i in $(seq 1 1200); do
-grep -Fq "EOF" /opt/azure/containers/provision.sh && break;
-if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi;
-done;
+cloud-init status --wait  > /dev/null 2>&1;
+[ $? -ne 0 ] && echo 'cloud-init failed' >> /var/log/azure/cluster-provision-cse-output.log && exit 1;
+echo "cloud-init succeeded" >> /var/log/azure/cluster-provision-cse-output.log;
 {{if IsAKSCustomCloud}}
-for i in $(seq 1 1200); do
-grep -Fq "EOF" {{GetInitAKSCustomCloudFilepath}} && break;
```
alexeldeib, line 8: can we confirm when this was moved to VHD? and potentially base64 encode/write during CSE if needed?

alexeldeib: oh nvm. you're only removing the wait here, not removing cloud init

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -484,7 +484,7 @@ testCoreDumpSettings() {
   #   A section heading -- this file is only supposed to have '[Coredump]'
   #   Settings, which take the form 'NAME=VALUE', where values can be empty or strings, and strings
   #   is pretty loose (more or less any printable character).
-  testSettingFileFormat $test $settings_file '^(#|$)' '^\[Coredump\]$' '^[A-Za-z_]+=[^[:cntrl:]]*$'
```
alexeldeib, line 487: please revert this. the casing is correct pre-diff, looks like merge/rebase issue.

alexeldeib (approved):
lgtm but please revert the vhd test changes

vhdbuilder/packer/test/linux-vhd-content-test.sh
anujmaheshwari1: done, thank you for catching this, bad rebase!

parts/linux/cloud-init/artifacts/cse_cmd.sh
anujmaheshwari1: yes, just removing the wait, not from cloud init.

PR #3300, opened 2023-06-08
feat: add memoryManagerPolicy to custom kubelet config, combine custom kubelet config and custom sysctl e2e tests

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/e2e/scenario/validators.go
+++ b/e2e/scenario/validators.go
@@ -1,6 +1,7 @@
 package scenario
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 )
@@ -24,7 +25,7 @@ func DirectoryValidator(path string, files []string) *LiveVMValidator {
 }
 
 func SysctlConfigValidator(customSysctls map[string]string) *LiveVMValidator {
-	keysToCheck := make([]string, len(customSysctls))
+	keysToCheck := make([]string, 0)
 	for k, _ := range customSysctls {
 		keysToCheck = append(keysToCheck, k)
 	}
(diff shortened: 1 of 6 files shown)


e2e/scenario/validators.go
```diff
@@ -47,6 +48,34 @@ func SysctlConfigValidator(customSysctls map[string]string) *LiveVMValidator {
...
 }
 
+func KubeletConfigValidator(kubeletConfigParams map[string]string) *LiveVMValidator {
+	keysToCheck := make([]string, 0)
+	for k, _ := range kubeletConfigParams {
+		keysToCheck = append(keysToCheck, k)
+	}
+	return &LiveVMValidator{
+		Description: "assert kubelet config settings",
+		Command:     fmt.Sprintf("jq `{%s}` /etc/default/kubeleconfig.json", strings.Join(keysToCheck, ",")),
```
alexeldeib, line 58: what are you trying to achieve here? just spit out the file? `jq . /etc/default/kubeletconfig.json` or simply `cat /etc/default/kubeletconfig.json` ? 

backticks embedded inside double quotes will result in the contents of the backticks being executed as a command.

the brackets seem pointless/purely syntax error

the keys in the map do not match the keys in the top level of the file

even if they did, you do not prepend them with a dot for proper field access.

```bash
root@Ubuntu-2204-jammy-amd64-base ~ # echo '{"foo": "bar", "memory-manager-policy": "foo"  }' | jq `memory-manager-policy`
{
  "foo": "bar",
  "memory-manager-policy": "foo"
}
memory-manager-policy: command not found
root@Ubuntu-2204-jammy-amd64-base ~ # echo '{"foo": "bar", "memory-manager-policy": "foo"  }' | jq `{memory-manager-policy}`
{memory-manager-policy}: command not found
{
  "foo": "bar",
  "memory-manager-policy": "foo"
}
root@Ubuntu-2204-jammy-amd64-base ~ # echo '{"foo": "bar", "memory-manager-policy": "foo"  }' | jq '{memory-manager-policy}'
jq: error: syntax error, unexpected '-', expecting '}' (Unix shell quoting issues?) at <top-level>, line 1:
{memory-manager-policy}
jq: 1 compile error
```

e2e/scenario/validators.go
```diff
@@ -47,6 +48,34 @@ func SysctlConfigValidator(customSysctls map[string]string) *LiveVMValidator {
 	}
 }
 
+func KubeletConfigValidator(kubeletConfigParams map[string]string) *LiveVMValidator {
+	keysToCheck := make([]string, 0)
+	for k, _ := range kubeletConfigParams {
+		keysToCheck = append(keysToCheck, k)
```
alexeldeib, line 54: I don't think you want this.

e2e/scenario/validators.go
UtheMan: The goal was to print out keys that are added in custom config from kubelet config json that's on the node. When node-shelled into the node, it worked as expected. But when the command is triggered during test there is an error. 

Then that command outputs a json in stdout that I can unmarshal to check values

e2e/scenario/validators.go
UtheMan: Hmm, its so I can generate that json file in stdout with jq

cameronmeissner:
feel free to reopen if still needed

PR #3304, opened 2023-06-09
feat: add initial ulimit support (memLock and noFiles), add e2e test for ulimit config

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -100,6 +100,17 @@ logs_to_events "AKS.CSE.disableSystemdResolved" disableSystemdResolved
 
 logs_to_events "AKS.CSE.configureAdminUser" configureAdminUser
 
+
+if [[ "${SHOULD_CONFIG_MEMLOCK}" == "true" ]]; then
+  # TODO - add error handling
+  echo "* hard memlock ${MEM_LOCK_VAL}" >> /etc/security/limits.conf || exit $ERR_UPDATE_CA_CERTS
```
alexeldeib, line 106: do you know what the default content of this file is? particularly on mariner @hbeberman?

you can find out in e2e, but wondering if simply append makes sense vs a drop in for /etc/security/limits.d  as a seperate file with our constructions?

does dupe later in same file take higher precedence? or error?

alexeldeib: can we move the timing of this closer to sysctls with other custom node config, but before kubelet start?

pkg/agent/baker.go
```diff
@@ -435,6 +435,28 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return 0
 		},
+		"ShouldConfigMemLock": func() bool {
+			return profile.CustomLinuxOSConfig != nil &&
```
alexeldeib, line 439: potentially add a helper on the top level profile for GetUlimitConfig which handles the nil checks?

pkg/agent/datamodel/types.go
```diff
@@ -681,6 +682,11 @@ type SysctlConfig struct {
 	VMVfsCachePressure             *int32 `json:"vmVfsCachePressure,omitempty"`
 }
 
+type UlimitConfig struct {
+	MaxLockedMemory string `json:"memlock,omitempty"`
```
alexeldeib, line 686: your json tags match the OS settings but I don't see you actually using this for serialization anywhere

I would change these because we do serve an HTTP API so on the wire between RP <-> AB this is how RP will serialize it.

you can have a separate type for on disk serialization if you later decide to use that approach

pkg/agent/datamodel/types.go
```diff
@@ -681,6 +682,11 @@ type SysctlConfig struct {
...
+	NoFile          *int32 `json:"nofile,omitempty"`
```
alexeldeib, line 687:
```suggestion
	MaxLockedMemory string `json:"maxLockedMemory ,omitempty"`
	NoFile          *int32 `json:"noFile,omitempty"`
```

pkg/agent/baker.go
```diff
@@ -435,6 +435,36 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return 0
 		},
+		"ShouldConfigContainerdUlimits": func() bool {
+			return profile.GetCustomLinuxOSConfig().GetUlimitConfig() != nil
```
cameronmeissner, line 439: could `GetCustomLinuxOSConfig` ever return nil?

UtheMan: Yeah,

```go
func (a *AgentPoolProfile) GetCustomLinuxOSConfig() *CustomLinuxOSConfig {
	if a == nil {
		return nil
	}
	return a.CustomLinuxOSConfig
}
```
but if it does then the whole chain will simply return nil, its written in the protobuf getter style.

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -87,6 +87,10 @@ ENABLE_UNATTENDED_UPGRADES="{{EnableUnattendedUpgrade}}"
 ENSURE_NO_DUPE_PROMISCUOUS_BRIDGE="{{ and NeedsContainerd IsKubenet (not HasCalicoNetworkPolicy) }}"
 SHOULD_CONFIG_SWAP_FILE="{{ShouldConfigSwapFile}}"
 SHOULD_CONFIG_TRANSPARENT_HUGE_PAGE="{{ShouldConfigTransparentHugePage}}"
+SHOULD_CONFIG_CONTAINERD_ULIMITS="{{ShouldConfigContainerdUlimits}}"
+CONTAINERD_ULIMITS="{{GetContainerdUlimitString}}"
+MEM_LOCK_VAL="{{GetMemLockValue}}"
```
alexeldeib, line 92: both *_val unused?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -142,6 +142,18 @@ configureCustomCaCertificate() {
     systemctl restart containerd
 }
 
+configureContainerdUlimits() {
+  CONTAINERD_ULIMIT_DROP_IN_FILE_PATH="/etc/systemd/system/containerd.service.d/set_ulimits.conf"
+  touch "${CONTAINERD_ULIMIT_DROP_IN_FILE_PATH}"
+  chmod 0600 "${CONTAINERD_ULIMIT_DROP_IN_FILE_PATH}"
+  tee "${CONTAINERD_ULIMIT_DROP_IN_FILE_PATH}" > /dev/null <<EOF
+$(echo "$CONTAINERD_ULIMITS" | tr ' ' '\n')
```
alexeldeib, line 150: you're already formatting it in go, why do extra formatting here?

or why not template it entirely? see containerd config for example if special chars is hard (use b64)

pkg/agent/baker.go
```diff
@@ -435,6 +435,36 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+			sb.WriteString("[Service]\n")
+			if ulimitConfig.MaxLockedMemory != "" {
+				sb.WriteString(fmt.Sprintf("LimitMEMLOCK=%s\n", ulimitConfig.MaxLockedMemory))
+			}
+			if ulimitConfig.NoFile != nil {
+				sb.WriteString(fmt.Sprintf("LimitNOFILE=%d\n", *ulimitConfig.NoFile))
+			}
+			return sb.String()
+		},
+		"GetMemLockValue": func() string {
```
alexeldeib, line 456: unused?

parts/linux/cloud-init/artifacts/cse_config.sh
UtheMan: I formatted it here because new lines were not getting transported over as expected, will try with b64

alexeldeib (approved):

PR #3307, opened 2023-06-09
fix: false negative when testing pamD settings on ubuntu

tobiasb-ms:
**What type of PR is this?**
/kind bug

The PamD tests were only ever meant to be run against Mariner 2.0. Somewhere along the way, that guard was deleted and the tests as checked in run against all os/version combinations. This change makes them only run against Mariner 2.0.
**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -521,24 +521,35 @@ testNfsServerService() {
 # Tests that the pam.d settings are set correctly, per the function
 # addFailLockDir in <repo-root>/parts/linux/cloud-init/artifacts/cis.sh.
 testPamDSettings() {
+  local os_sku="${1}"
+  local os_version="${2}"
   local test="testPamDSettings"
   local settings_file=/etc/security/faillock.conf
   echo "$test:Start"
 
-  # Existence and format check. The man page https://www.man7.org/linux/man-pages/man5/faillock.conf.5.html
-  # describes the following format for each line:
-  #   Comments start with '#'.
-  #   Blank lines are ignored.
-  #   Lines are of in two forms:
-  #       'setting = value', where settings are lower-case and include '_'
-  #       'setting'
-  #   Whitespace at beginning and end of line, along with around the '=' is ignored.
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -521,24 +521,35 @@ testNfsServerService() {
...
-  #   Whitespace at beginning and end of line, along with around the '=' is ignored.
-  testSettingFileFormat $test $settings_file '^(#|$)' '^[[:space:]]*$' '^[[:space:]]*[a-z_][[:space:]]*' '^[[:space:]]*[a-z_]+[[:space::]]*=[^[:cntrl:]]*$'
-
-  # Look for the setting we specifically set in <repo-root>/parts/linux/cloud-init/artifacts/cis.sh
-  # and ensure it's set to the values we expect.
-  echo "$test: Checking specific settings in $settings_file"
-  testSetting $test $settings_file 'dir' '^[[:space:]]*dir[[:space:]]*=' '=' '/var/log/faillock'
+  # We only want to run this test on Mariner 2.0
+  # So if it's anything else, report that we're skipping the test and bail.
+  if [[ "${os_sku}" != "CBLMariner" || "${os_version}" != "2.0" ]]; then
```
alexeldeib, line 532: am I crazy or are they both listed as CBLMariner in /etc/os-release?

https://github.com/Azure/AgentBaker/blob/166c545f12241d5ee6160d6328c61d672f47e91d/vhdbuilder/release-notes/AKSCBLMarinerV2/gen2/latest.txt#L659-L661

https://github.com/Azure/AgentBaker/blob/166c545f12241d5ee6160d6328c61d672f47e91d/vhdbuilder/release-notes/AKSCBLMariner/gen1/latest.txt#L429-L431

alexeldeib: oh I get it. skip v1 only but do Ubuntu.

alexeldeib (approved):

alexeldeib:
ssh issues on arm64

vhdbuilder/packer/test/linux-vhd-content-test.sh
tobiasb-ms: Yes, I think they're both `CBL-Mariner`, but then the version check makes sure we're at V2.

PR #3309, opened 2023-06-10
test: comment out misbehaving test

tobiasb-ms:
**What type of PR is this?**
/kind test

The test `testImagesRetagged` in `linux-vhd-content-test.sh` has, for a long time we think, misbehaved in a couple ways.
1. It sends error output to `stdout` instead of `stderr`.
2. In an error scenario, it does an `exit 1` instead of a `return 1`.

This in turn causes two problems:
1. The calling code -- `run-tests.sh`, using `az vm run-command invoke` -- never looks at the exit status of the test script (and in fact doesn't have access to it). It only looks at `stderr`, which isn't written to. So the test failure isn't being flagged as a failure.
2. Because the test does an `exit`, the entire test script is stopped, which means the rest of the tests aren't run.

This change just comments out `testImagesRetagged` so we have full test coverage. Note that doing this found two recently added tests that failed, but since the tests hadn't run in the pipeline we hadn't noticed. ([Both ](https://github.com/Azure/AgentBaker/pull/3285) of [them](https://github.com/Azure/AgentBaker/pull/3307) were test issues, thankfully.)

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):
should we add a comment in code similar to what you have in the PR?

code comments -> git blame -> github PR overview without in-repo history for me (mostly because it's hard to reassociate if merge title/PR number won't match)

appreciate the diligence here!

PR #3314, opened 2023-06-12
refactor: move sysctl out of customdata, into cse

anujmaheshwari1:
**What type of PR is this?**
This PR moves sysctl configs out of customdata and into CSE to isolate variables for self contained VHD. It also adds unit tests to ensure defaults are being set correctly(to avoid a regression like in January 2023).

image

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -908,6 +908,14 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"ShouldDisableSSH": func() bool {
 			return config.SSHStatus == datamodel.SSHOff
 		},
+		"GetSysctlContent": func() string {
+			sysctlTemplate := template.Must(template.New("sysctl").Parse(sysctlTemplateString))
+			var b bytes.Buffer
+			if err := sysctlTemplate.Execute(&b, profile); err != nil {
+				panic(fmt.Errorf("failed to execute sysctl template: %s", err))
```
cameronmeissner, line 915: is panicking really what we should be doing in this case?

cameronmeissner: I see, I guess we do panic in other spots too

cameronmeissner (approved):

pkg/agent/baker.go
anujmaheshwari1: yeah good point, I used the commit from previous iteration, let me just return the error.

anujmaheshwari1: yeah, probably not, updated func to return error

pkg/agent/baker.go
```diff
@@ -908,6 +908,14 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"ShouldDisableSSH": func() bool {
 			return config.SSHStatus == datamodel.SSHOff
 		},
+		"GetSysctlContent": func() (string, error) {
+			sysctlTemplate := template.Must(template.New("sysctl").Parse(sysctlTemplateString))
```
alexeldeib, line 912: this func already returns an error so may as well return it back instead of template.Must()

pkg/agent/baker.go
```diff
@@ -946,6 +954,129 @@ func isMariner(osSku string) bool {
...
 
+const sysctlTemplateString = `# This is a partial workaround to this upstream Kubernetes issue:
+# https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731
+net.ipv4.tcp_retries2=8
+net.core.message_burst=80
+net.core.message_cost=40
+{{- if .CustomLinuxOSConfig}}
+  {{ if .CustomLinuxOSConfig.Sysctls}}
+    {{ if .CustomLinuxOSConfig.Sysctls.NetCoreSomaxconn -}}
+      net.core.somaxconn={{.CustomLinuxOSConfig.Sysctls.NetCoreSomaxconn}}
```
alexeldeib, line 965: I doubt this indentation works and in checking the decoded customdata from UT I think it's messed up

this needs to be de-dented all the way, or else you need to use hyphens `{{- foo -}}` to trim whitespace

anujmaheshwari1: dedented it.

pkg/agent/baker.go
```diff
@@ -946,6 +954,129 @@ func isMariner(osSku string) bool {
...
 }
 
+const sysctlTemplateString = `# This is a partial workaround to this upstream Kubernetes issue:
+# https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731
+net.ipv4.tcp_retries2=8
+net.core.message_burst=80
+net.core.message_cost=40
+{{- if .CustomLinuxOSConfig}}
+  {{ if .CustomLinuxOSConfig.Sysctls}}
+    {{ if .CustomLinuxOSConfig.Sysctls.NetCoreSomaxconn -}}
```
alexeldeib, line 964:
```suggestion
{{ with $sysctls :=. CustomLinuxOSConfig.Sysctls }}
{{ if $sysctls.NetCoreSomaxconn }}
{{ else }}
```

maybe? unsure

alexeldeib: maybe will fail nil check too

pkg/agent/baker.go
anujmaheshwari1: done.

pkg/agent/baker.go
anujmaheshwari1: dedented to one line.

alexeldeib (approved):

PR #3329, opened 2023-06-16
feat: generate image bom for Windows VHDs

AbelHu:
**What type of PR is this?**
/kind feature

generate image bom for Windows VHDs

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/list-images.ps1
+++ b/vhdbuilder/packer/list-images.ps1
@@ -1,0 +1,91 @@
+<#
+    .SYNOPSIS
+        Produces a JSON image BOM for a Windows VHD
+
+    .DESCRIPTION
+        Produces a JSON image BOM for a Windows VHD
+#>
+$ErrorActionPreference = "Stop"
+
+$imageBomJsonFilePath = "c:\image-bom.json"
+$bomList = @()
+
+# starting containerd for printing containerD info, the same way as we pre-pull containerD images in configure-windows-vhd.ps1
+Start-Job -Name containerd -ScriptBlock { containerd.exe }
+$imageList=$(ctr.exe -n k8s.io image ls | select -Skip 1)
+foreach($image in $imageList) {
+    $splitResult=($image -split '\s+')
+    if ($splitResult[0].StartsWith("sha256:")) {
+        # Get repoDigests from sha256
+        # Example:
+        # sha256:1fb25eb608ab558cf66b7546fd598b2dc81b6678a380fcdb188da780e107f4ab
+        # application/vnd.docker.distribution.manifest.list.v2+json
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/list-images.ps1
```diff
@@ -0,0 +1,91 @@
...
+        # GiB
+        # windows/amd64
+        # io.cri-containerd.image=managed
+        $repoTags=$splitResult[0]
+        $id=$splitResult[2]
+
+        $isExist=$false
+        foreach($bom in $bomList) {
+            if ($bom.id -eq $id) {
+                $bom.repoTags += $repoTags
```
alexeldeib, line 68: we don't add this for linux I think. we only have the repo digest? I believe I checked with Gaurav and he didn't need the image id (hash of layers), only the manifest sha (which you have already).

may want to double check with Gaurav

alexeldeib (approved):
lgtm.  two minor comment you may want to check with Gaurav/FIM folks -- 

Linux has very few cases where the same image has multiple tags, so the sha/tag are mostly 1:1. We have one case and it hasn't caused issues, but if _most_ windows images are tagged twice FIM folks may need to handle it in their pipelines.

other comment in code

vhdbuilder/packer/list-images.ps1
alexeldeib: as ref https://github.com/Azure/AgentBaker/blob/7a0ef4ff0090e9581ccfc97e5c3b4477db1863ad/vhdbuilder/release-notes/AKSUbuntu/gen2/2204containerd/latest-image-list.json#L509-L518

alexeldeib: sorry -- not add it as part of repoTags. it's already there as `id`.

AbelHu:
> lgtm. two minor comment you may want to check with Gaurav/FIM folks --
> 
> Linux has very few cases where the same image has multiple tags, so the sha/tag are mostly 1:1. We have one case and it hasn't caused issues, but if _most_ windows images are tagged twice it may need to be handled in ingestion pipelines.
> 
> other comment in code

Thanks for the feedback, @alexeldeib. I think that we can remove sha256tag as you said since the sha256tag is id.

Is retagMCRImagesForChina not used by Linux VHDs? It seems like that it is needed for AzureChinaCloud. Please free to correct me.

vhdbuilder/packer/list-images.ps1
AbelHu: Thanks for the feedback. Add the code to ignore repoTags when it contains id.

PR #3330, opened 2023-06-16
fix: change nofile ulimit to string

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -448,8 +448,8 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			if ulimitConfig.MaxLockedMemory != "" {
 				sb.WriteString(fmt.Sprintf("LimitMEMLOCK=%s\n", ulimitConfig.MaxLockedMemory))
 			}
-			if ulimitConfig.NoFile != nil {
-				sb.WriteString(fmt.Sprintf("LimitNOFILE=%d\n", *ulimitConfig.NoFile))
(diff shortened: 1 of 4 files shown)


pkg/agent/baker.go
```diff
@@ -448,8 +448,8 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			if ulimitConfig.MaxLockedMemory != "" {
 				sb.WriteString(fmt.Sprintf("LimitMEMLOCK=%s\n", ulimitConfig.MaxLockedMemory))
 			}
-			if ulimitConfig.NoFile != nil {
-				sb.WriteString(fmt.Sprintf("LimitNOFILE=%d\n", *ulimitConfig.NoFile))
+			if ulimitConfig.NoFile != "" {
```
cameronmeissner, line 451: unrelated, but do we have tests in baker_test.go for this?

cameronmeissner (approved):

pkg/agent/baker.go
cameronmeissner: mainly for ulimit stuff in general I mean

UtheMan: Not right now, only e2e. The only custom linux os config tests i could find were for ubuntu 16.04 (probably some old edge case?), so I didn't add any yet for my change. Could add more of them in the future when more options will be added

PR #3334, opened 2023-06-17
feat: add abe2e support to retain vms for debugging

cameronmeissner:
**What type of PR is this?**

/kind feature

add support to retain vms for debugging to abe2e

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib:
is the thinking you'd start a scenario with this expecting it to fail? or should we attempt to keep them for some amount of time by default if the scenario fails? how would you access it -- need to create a bastion still?

cameronmeissner:
> is the thinking you'd start a scenario with this expecting it to fail? or should we attempt to keep them for some amount of time by default if the scenario fails? how would you access it -- need to create a bastion still?

yep exactly, I mainly did this in response to use-cases encountered by @UtheMan and @anujmaheshwari1 where they expected some sort of failure (I believe) and needed to manually access the vm

and yeah, we'd use bastion with the SSH key written to disk (as I've tested)

cameronmeissner:
I'd be open to the idea of keeping the VMs around for a certain period of time assuming the scenario fails (maybe only if the live VM validation/bootstrapping bits fail?) - will give this some more thought

UtheMan:
In my case, my current workflow for AB changes looks a bit like TDD flow:

- add/modify existing e2e test case, expecting it to fail
- node-shell into that failed vmss, figure out where the thing i want to add needs to go (where in the file system, do i need to restart daemon etc)
- Change AB code to make my test case pass, as I'm doing that I will most likely node-shell into the node a couple more times to figure out why what I expected to happen didn't happen (could use logs for that, but it's often easier for me to just node-shell and check files/systemd units/whatever myself)
- Eventually test case is working as expected and passing as expected, then there's no longer need for node-shell

cameronmeissner:
> as I'm doing that I will most likely node-shell into the node a couple more times to figure out why what I expected to happen didn't happen 

am assuming by node-shell you mean bastion + ssh

UtheMan:
> am assuming by node-shell you mean bastion + ssh

Ah yeah, depends on what kind of stuff im debugging. If the node is not getting added to cluster properly then yeah. But e.g. when adding custom node config changes node would provision correctly, but my CSE changes were in wrong place so ulimits weren't getting added as expected. Then I could use node-shell since node was up and running.

anujmaheshwari1:
I second what Mikolaj said. I dont necessarily expect a test to fail when I run it, but if it does, I have to change code to not delete the vmss and then run it again to debug what is going on. This functionality makes it easier to do it in one go, if I expect a failure. 

Also, in other cases where I might want to node-shell to a successfully bootstrapped cluster to fetch some files, run some commands, etc, this is useful as well!

cameronmeissner:
maybe we can merge this and loop back around for auto-retaining if a given scenario fails?

ganeshkumar5699:
Like Anuj's use case, I would also find it useful to look at node logs or run a particular pod on the node after AB e2e succeeds, so the option to keep VMSS instances even on success would be better (vs manual code change for that).

anujmaheshwari1 (approved):

PR #3339, opened 2023-06-19
chore: add disable custom data field to NBC

anujmaheshwari1:
**What type of PR is this?**
This PR introduces a field DisableCustomData to NBC, which when set to true will not check for cloud init completion. Currently, the default is false, that is, it will always check for cloud init completion. This, in tandem with RP toggle change will allow not setting custom data.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -1,7 +1,10 @@
-echo $(date),$(hostname) > /var/log/azure/cluster-provision-cse-output.log;
-cloud-init status --wait  > /dev/null 2>&1;
-[ $? -ne 0 ] && echo 'cloud-init failed' >> /var/log/azure/cluster-provision-cse-output.log && exit 1;
-echo "cloud-init succeeded" >> /var/log/azure/cluster-provision-cse-output.log;
+PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log";
+echo $(date),$(hostname) > ${PROVISION_OUTPUT};
+{{if not ShouldDisableCustomData}}
```
UtheMan, line 3: idea: flip it around to be 

{{ if ShouldEnableCustomData}}

i feel like it will be easier to read without the double negative?

cameronmeissner: I agree, bool names should ideally never contain 'negatives'

anujmaheshwari1: @alexeldeib what is your opinion on the name "ShouldEnableCustomData" for the function for the sake of readability?
Mikolaj, Cameron and I had a side discussion offline, I would like the variable name to remain DisableCustomData since the crux of the matter here is disabling custom data + I want to make sure it aligns with RP toggle, but I would be open to renaming the function to be better readable with the bash script. 

also welcome other suggestions!

alexeldeib: changing the function in baker.go and leaving the datamodel sounds fine to me. I agree the bash is wordy

anujmaheshwari1: done!

UtheMan (approved):

e2e/suite_test.go
```diff
@@ -103,14 +104,17 @@ func runScenario(ctx context.Context, t *testing.T, r *mrand.Rand, opts *scenari
...
 	vmssName := getVmssName(r)
 	log.Printf("vmss name: %q", vmssName)
 
 	vmssSucceeded := true
-	vmssModel, cleanupVMSS, err := bootstrapVMSS(ctx, t, r, vmssName, opts, publicKeyBytes)
-	if cleanupVMSS != nil {
-		defer cleanupVMSS()
-	}
+	vmssModel, _, err := bootstrapVMSS(ctx, t, r, vmssName, opts, publicKeyBytes)
+	// if cleanupVMSS != nil {
```
cameronmeissner, line 115: :(

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -1,7 +1,10 @@
-echo $(date),$(hostname) > /var/log/azure/cluster-provision-cse-output.log;
-cloud-init status --wait  > /dev/null 2>&1;
-[ $? -ne 0 ] && echo 'cloud-init failed' >> /var/log/azure/cluster-provision-cse-output.log && exit 1;
-echo "cloud-init succeeded" >> /var/log/azure/cluster-provision-cse-output.log;
+PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log";
+echo $(date),$(hostname) > ${PROVISION_OUTPUT};
+{{if ShouldEnableCustomData}}
+cloud-init status --wait > /dev/null 2>&1;
```
cameronmeissner, line 4: could we also add a test to baker_test.go to disable custom data and ensure that `cloud-init status --wait > /dev/null 2>&1` isn't present within cse?

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_cmd.sh
anujmaheshwari1: done :)

alexeldeib (approved):

cameronmeissner (approved):

PR #3348, opened 2023-06-21
Revert "test: clone the branch from the build rather main for testing (#3342)"

tobiasb-ms:
This reverts commit 2bbf1058c89e5c5ff2f314fcb9bda13cd79da66e.

**What type of PR is this?**
/kind bug
/kind test
/kind regression

The aforementioned commit checks out the git branch passed to use by the build pipeline in our tests. However, it doesn't work in all cases (notably pull request branches), so reverting it while figuring out the right solution.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):
arm64 should pass now if you want a sanity check, but lgtm!

PR #3349, opened 2023-06-21
chore: move /etc/default/kubelet to CSE

anujmaheshwari1:
**What type of PR is this?**
This PR adds the file /etc/default/kubelet to CSE, while letting it also reside in custom data(for databricks). This duplication helps progress work for self contained VHD.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -366,6 +366,17 @@ ensureDHCPv6() {
 }
 
 ensureKubelet() {
+    KUBELET_DEFAULT_FILE=/etc/default/kubelet
+    mkdir -p /etc/default
+    echo "KUBELET_FLAGS=${KUBELET_FLAGS}" >> "${KUBELET_DEFAULT_FILE}"
```
alexeldeib, line 371: this actually duplicates all values in the file fyi because it's all appending. but it works since they just get sourced twice

alexeldeib: this has the cloud init wait before it right?

alexeldeib (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
anujmaheshwari1: yes it does, cloud init wait at the very beginning https://github.com/Azure/AgentBaker/blob/26dedd5c9270e0f77ed1303676ed31eb9e8ac114/parts/linux/cloud-init/artifacts/cse_cmd.sh#L2

and then later setting vars https://github.com/Azure/AgentBaker/blob/26dedd5c9270e0f77ed1303676ed31eb9e8ac114/parts/linux/cloud-init/artifacts/cse_cmd.sh#L133

anujmaheshwari1: hmm, if I change the first redirect from >> to >, that solves the problem of duplication. @alexeldeib preference?

anujmaheshwari1: updated PR to use > instead of >> to overwrite contents of /etc/default/kubelet to avoid duplication.

PR #3352, opened 2023-06-22
test: use git branch from build for tests

tobiasb-ms:
DRAFT -- IGNORE

**What type of PR is this?**
/kind test

In `linux-vhd-content-test.sh`, we clone `AgentBaker` to source a couple helper files. However, we clone `master`, which means that we can't test updating those files, and we can't iterate on adding tests that we would want to keep in other files in repo.

This change takes the git branch to clone on the command-line of the test script, passing it in through the pipeline in `run-tests.sh`.

Note that this is the second attempt at this. The [first](https://github.com/Azure/AgentBaker/pull/3342) used a simpler method that doesn't work for pull request branches.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -1,16 +1,62 @@
 #!/bin/bash
-git clone https://github.com/Azure/AgentBaker.git 2>/dev/null
-source ./AgentBaker/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh 2>/dev/null
-source ./AgentBaker/parts/linux/cloud-init/artifacts/cse_helpers.sh 2>/dev/null
 COMPONENTS_FILEPATH=/opt/azure/components.json
 KUBE_PROXY_IMAGES_FILEPATH=/opt/azure/kube-proxy-images.json
 MANIFEST_FILEPATH=/opt/azure/manifest.json
 VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1,16 +1,62 @@
...
+COMMAND="git checkout --quiet ${LOCAL_GIT_BRANCH}"
+if ! ${COMMAND}; then
+  err 'git-clone' "Failed to checkout local branch '${LOCAL_GIT_BRANCH}'"
+  err 'git-clone' "Used command '${COMMAND}'"
+  exit 1
+fi
+if ! popd; then
+  err 'git-clone' "Failed to popd out of AgentBaker repo -- this seems impossible"
+  err 'git-clone' "Current directory is $(pwd)"
+  err 'git-clone' "pushd stack is $(dirs -p)"
```
alexeldeib, line 54: dir -p.... TIL

alexeldeib (approved):
lgtm, thanks for round 2 :)

alexeldeib:
guess we can reuse arm64 for validation. but seems promising :D

vhdbuilder/packer/test/linux-vhd-content-test.sh
sirpentor: Yeah -- I learned the other day that `dirs` will list the stack used by `pushd` and `popd`. the more you know, the farther you go!

PR #3357, opened 2023-06-25
fix: clean up created storage container in cse validation

wanqingfu:
**What type of PR is this?**
Clean up created storage container in cse validation

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/e2e/windows/e2e-helper.sh
+++ b/e2e/windows/e2e-helper.sh
@@ -16,8 +16,34 @@ exec_on_host() {
     kubectl exec $(kubectl get pod -l app=debug -o jsonpath="{.items[0].metadata.name}") -- bash -c "nsenter -t 1 -m bash -c \"$1\"" > $2
 }
 
+backfill_clean_storage_container() {
+    set +x
+    # Get supported kubernetes versions
+    versions=$(az aks get-versions --location $LOCATION --query "orchestrators[].orchestratorVersion" -o tsv)
+    k8s_versions=${versions//./}
+
+    # Get container names e.g. akswinstore2022-1256 (for this $container_version would be "1256")
+    container_names=$(az storage container list --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --query "[?starts_with(name, 'akswinstore')].name" -o tsv)
+    # Check if the container version is still supported and delete the container if not
+    for container_name in $container_names; do
+        container_version=$(echo $container_name | cut -d '-' -f 2- | awk '{print tolower($0)}')
+        echo "container version is $container_version"
+        if [[ $k8s_versions == *"$container_version"* ]]; then
+            echo "The version $container_version is available in the $LOCATION region."
+        else
+            echo "The version $container_version is not available in the $LOCATION region."
+            echo "Deleting the container."
+            az storage container delete --name $container_name --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY
+            echo "Deletion completed."
+        fi
+    done
+
+    set -x
+}
+
 create_storage_container() {
     set +x
+
     # check if the storage container exists and create one if not
     exists=$(az storage container exists --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --name $WINDOWS_E2E_STORAGE_CONTAINER)
     if [[ $exists == *false* ]]; then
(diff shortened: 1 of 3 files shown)


e2e/windows/scenarios/windows/property-windows-template.json
```diff
@@ -92,7 +92,7 @@
     "--cloud-provider": "external",
     "--enforce-node-allocatable": "\"\"\"\"",
     "--eviction-hard": "\"\"\"\"",
-    "--feature-gates": "CSIMigration=true,CSIMigrationAzureDisk=true,CSIMigrationAzureFile=true,DelegateFSGroupToCSIDriver=true,DynamicKubeletConfig=false",
```
AbelHu, line 95: Will it fail the tests with k8s < v1.27?

e2e/windows/e2e-helper.sh
```diff
@@ -16,8 +16,22 @@ exec_on_host() {
     kubectl exec $(kubectl get pod -l app=debug -o jsonpath="{.items[0].metadata.name}") -- bash -c "nsenter -t 1 -m bash -c \"$1\"" > $2
 }
 
+backfill_clean_storage_container() {
+    set +x
+    # list outdated storage containers created over 1 week ago
```
AbelHu, line 21: @junjiezhang1997 This will only delete those containers created over 1 week ago so it should not impact our tests/builds, is my understanding right?

e2e/windows/scenarios/windows/property-windows-template.json
wanqingfu: It will not. I've tried 1.24.10, 1.25.6 and attached pipelines in the channel.

e2e/windows/e2e-helper.sh
junjiezhang1997: I don't think so. The containers are reused by different pipelines. In my opinion, we should delete container whose kubernetes version is not supported any more.

wanqingfu: You are saying there may be problems running concurrent builds right? E.g. if pipeline A runs this function to clean something say the container for 1.24.9 that was created over a week ago, another concurrent pipeline B thought the container existed only to find it was later deleted by pipeline A?

How should we pass obsolete kubernetes version? Pipeline variables? Hard-coded?

Would it be better if we add a variable like "BACKFILL_CONTAINER" such that only those pipelines with this switch turned on will clean up - no conflict any more?

junjiezhang1997: We don't need to pass obsolete kubernetes versions. We can use az cli to get supported kubernetes versions and check whether the version of container is still supported.

For example: az aks get-versions -l eastus --query "orchestrators[?default==\`true\`].orchestratorVersion" -otsv

junjiezhang1997: The variable like "BACKFILL_CONTAINER" will not work. As long as a container with supported version is deleted and the cluster with the same supported version is not deleted yet, a new pipeline will fail to get linux file from the deleted container.

e2e/windows/e2e-helper.sh
```diff
@@ -16,8 +16,34 @@ exec_on_host() {
...
+backfill_clean_storage_container() {
+    set +x
+    # Get supported kubernetes versions
+    versions=$(az aks get-versions --location $LOCATION --query "orchestrators[].orchestratorVersion" -o tsv)
+    k8s_versions=${versions//./}
+
+    # Get container names
+    container_names=$(az storage container list --account-name $STORAGE_ACCOUNT_NAME --account-key $MAPPED_ACCOUNT_KEY --query "[?starts_with(name, 'akswinstore')].name" -o tsv)
+    # Check if the container version is still supported and delete the container if not
+    for container_name in $container_names; do
```
wanqingfu, line 28: Is this what you would prefer or? @junjiezhang1997 

pipeline log
```
container version is 1237
The version 1237 is not available in the eastus region.
Deleting the container.
{
  "deleted": true
}
Deletion completed.
```

AbelHu (approved):
lint test failed but I think that it is ok if you squash merge this PR. Please always only delete those resources in plan next time to avoid delete all resources and block others. Since those containers are only for test, the impact is minimal but it will cause outage if it happens in production.

e2e/windows/e2e-helper.sh
```diff
@@ -16,8 +16,34 @@ exec_on_host() {
...
+        container_version=$(echo $container_name | cut -d '-' -f 2- | awk '{print tolower($0)}')
```
AbelHu, line 29: It is better to add a comment with a container name as an example

junjiezhang1997 (approved):

PR #3361, opened 2023-06-26
fix: remove managed_image_resource_group_name and managed_image_name and publish SIG only

wanqingfu:
**What type of PR is this?**
publish to shared image gallery only after packer v1.4.0 

For some contexts see this https://github.com/hashicorp/packer-plugin-azure/issues/41.

`azure-arm: ERROR: -> PropertyChangeNotAllowed : Changing property 'sourceVirtualMachine' is not allowed.`

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/packer.mk
+++ b/packer.mk
@@ -63,7 +63,7 @@ run-packer: az-login
 	@packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer | tee -a packer-output)
 
 run-packer-windows: az-login
-	@packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer-windows | tee -a packer-output)
+	@packer init ./vhdbuilder/packer/packer-plugin.pkr.hcl && packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer-windows | tee -a packer-output)
 
 cleanup: az-login
 	@./vhdbuilder/packer/cleanup.sh
--- a/vhdbuilder/packer/packer-plugin.pkr.hcl
+++ b/vhdbuilder/packer/packer-plugin.pkr.hcl
@@ -1,0 +1,8 @@
+packer {
+  required_plugins {
+    azure = {
+      version = "= 1.4.3"
+      source  = "github.com/hashicorp/azure"
+    }
+  }
+}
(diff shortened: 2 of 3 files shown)


packer.mk
```diff
@@ -68,7 +68,7 @@ run-packer: az-login
 	@packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer | tee -a packer-output)
 
 run-packer-windows: az-login
-	@packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer-windows | tee -a packer-output)
+	@packer init -upgrade ./vhdbuilder/packer/packer-plugin.pkr.hcl && packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer-windows | tee -a packer-output)
```
wanqingfu, line 71: Ref: https://github.com/hashicorp/packer-plugin-azure https://developer.hashicorp.com/packer/docs/configure

vhdbuilder/packer/packer-plugin.pkr.hcl
```diff
@@ -0,0 +1,8 @@
+packer {
+  required_plugins {
+    azure = {
+      version = ">= 1.4.3"
```
wanqingfu, line 4: Ref https://github.com/hashicorp/packer-plugin-azure

AbelHu, line 4: can we specify a version here to avoid any possible issue in the latest version?

vhdbuilder/packer/packer-plugin.pkr.hcl
```diff
@@ -0,0 +1,8 @@
+packer {
+  required_plugins {
+    azure = {
+      version = "= 1.4.3"
```
AbelHu, line 4: It seems like that packer version installed by default is `Packer v1.8.1`. Should we need this file to add v1.4.3 as the required?

wanqingfu: I used to use `>=1.4.3` but you asked me to use a fixed static version `1.4.3` in case of failure. This is the version that guarantees to work.

wanqingfu: btw this is the packer plugin version not the packer version.

AbelHu (approved):
LGTM. Thanks for the PR.

PR #3385, opened 2023-07-04
feat: Provide backwards compatibility with existing Ubuntu-based path…

eklingmsft:
**What type of PR is this?**
/kind feature

Provide backwards compatibility with Ubuntu based nodes for mounting of ca-certificates.crt file into pods. /etc/ssl/certs/ is already symlinked to /etc/pki/tls/certs on Mariner hosts.

Current practice is to mount host private CAs into pods once certificate bundle is built. In a partial migration to Mariner-based nodes the host mount path ideally should stay consistent to avoid having to create duplicate specs/node selectors based on OS. 

Fixes #5785

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/init-aks-custom-cloud-mariner.sh
```diff
@@ -13,6 +13,8 @@ IFS=$IFS_backup
 
 cp /root/AzureCACertificates/*.crt /etc/pki/ca-trust/source/anchors/
 /usr/bin/update-ca-trust
+# Provide backwards compatibility with existing Ubuntu-based path of /etc/ssl/certs/ca-certificates.crt
+ln -s /etc/pki/tls/certs/ca-bundle.crt /etc/pki/tls/certs/ca-certificates.crt
```
hbeberman, line 17: Both `/etc/pki/tls/certs/ca-certificates.crt` and `/etc/pki/tls/certs/ca-bundle.crt` should already be present on the host as they're both provided by the `ca-certificates-shared` package. 

If this isn't the case from observation on custom clouds we'll need to dig in a bit further as to why it's not there.

PR #3392, opened 2023-07-06
fix: add reserved port for vnet if custom node config port range exceeds 65330

UtheMan:
**What type of PR is this?**

https://learn.microsoft.com/en-us/azure/virtual-network/virtual-networks-faq#what-protocols-can-i-use-within-vnets

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/scenario/scenario_marinerv2-custom-sysctls.go
```diff
@@ -8,10 +8,11 @@ import (
 
 func marinerv2CustomSysctls() *Scenario {
 	customSysctls := map[string]string{
-		"net.ipv4.ip_local_port_range":       "32768 65535",
+		"net.ipv4.ip_local_port_range":       "32768 62535",
```
cameronmeissner, line 11: lol I think that fix I made a while ago to change the sshkey file name actually worked, this just reminded me

pkg/agent/baker.go
```diff
@@ -1068,6 +1080,10 @@ net.ipv4.tcp_tw_reuse={{if $s.NetIpv4TcpTwReuse}}1{{else}}0{{end}}
 {{- end}}
 {{- if $s.NetIpv4IpLocalPortRange}}
 net.ipv4.ip_local_port_range={{$s.NetIpv4IpLocalPortRange}}
+{{$rangeEnd := getPortRangeEndValue $s.NetIpv4IpLocalPortRange " " 1}}
+{{ if ge $rangeEnd 65330}}
+net.ipv4.ip_local_reserved_ports=65330
```
cameronmeissner, line 1085: can we also add something to baker_test.go for this?

pkg/agent/baker.go
```diff
@@ -945,6 +948,15 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 	}
 }
 
+func getPortRangeEndValue(str, delimiter string, index int) int {
```
cameronmeissner, line 951: can the port range ever be anything else than "A B"? if not then we probably don't need delimiter/index args here

UtheMan: Good point, added it at first cause I tried to make it more generic, in the end it ended up like this :D

cameronmeissner: maybe also worth trimming from the left and right side as well, though not sure if that would ever actually help

pkg/agent/baker.go
```diff
@@ -928,7 +929,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return config.SSHStatus == datamodel.SSHOff
 		},
 		"GetSysctlContent": func() (string, error) {
-			sysctlTemplate, err := template.New("sysctl").Parse(sysctlTemplateString)
+			templateFuncMap := make(template.FuncMap)
+			templateFuncMap["getPortRangeEndValue"] = getPortRangeEndValue
```
alexeldeib, line 933: why define this as a template func and not call the go func directly? we do this in many places for e.g. gpu driver version, you don't need extra indirection -- that's what the map of template funcs already provides.

alexeldeib: oh jk, it's not the same template

alexeldeib (approved):

cameronmeissner (approved):

PR #3395, opened 2023-07-06
fix: fix cse

cvvz:
**What type of PR is this?**
/kind bug

`$?` returns 0 forever without this change, thus we can never get the failure output for "curl" command.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


andyzhangx:
@anujmaheshwari1 could you take a look? this change lgtm, thanks.

PR #3370, opened 2023-06-28
feat: add ubuntu minimal gen1/gen2 VHD build support

anujmaheshwari1:
**What type of PR is this?**
This PR adds Ubuntu Minimal Gen1/Gen2 VHD build support to our pipelines. Green build https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=75675653&view=results

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #3396, opened 2023-07-06
chore: add sig config for ubuntu minimal

anujmaheshwari1:
**What type of PR is this?**
This PR adds sig config for Ubuntu Minimal VHD.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/datamodel/sig_config.go
+++ b/pkg/agent/datamodel/sig_config.go
@@ -93,6 +93,8 @@ var AvailableUbuntu2204Distros = []Distro{
 	AKSUbuntuContainerd2204TLGen2,
 	AKSUbuntuEdgeZoneContainerd2204,
 	AKSUbuntuEdgeZoneContainerd2204Gen2,
+	AKSUbuntuMinimalContainerd2204,
+	AKSUbuntuMinimalContainerd2204Gen2,
 }
 
 //nolint:gochecknoglobals
@@ -122,6 +124,8 @@ var AvailableContainerdDistros = []Distro{
 	AKSUbuntuContainerd2204TLGen2,
 	AKSUbuntuEdgeZoneContainerd2204,
 	AKSUbuntuEdgeZoneContainerd2204Gen2,
+	AKSUbuntuMinimalContainerd2204,
+	AKSUbuntuMinimalContainerd2204Gen2,
 }
 
 //nolint:gochecknoglobals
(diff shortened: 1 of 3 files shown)


pkg/agent/datamodel/sig_config.go
```diff
@@ -447,6 +452,20 @@ var (
 		Version:       LinuxSIGImageVersion,
 	}
 
+	SIGUbuntuMinimalContainerd2204ImageConfigTemplate = SigImageConfigTemplate{
+		ResourceGroup: AKSUbuntuResourceGroup,
+		Gallery:       AKSUbuntuGalleryName,
+		Definition:    "2204minimalcontainerd",
+		Version:       "202306.30.0",
```
cameronmeissner, line 459: can we have a separate const for minimal versions?

pkg/agent/datamodel/sig_config.go
```diff
@@ -447,6 +452,20 @@ var (
 		Version:       LinuxSIGImageVersion,
 	}
 
+	SIGUbuntuMinimalContainerd2204ImageConfigTemplate = SigImageConfigTemplate{
+		ResourceGroup: AKSUbuntuResourceGroup,
+		Gallery:       AKSUbuntuGalleryName,
+		Definition:    "2204minimalcontainerd",
```
cameronmeissner, line 458: this reminds me, did you make sure to update the set SKU_NAME logic in the vhd builds?

pkg/agent/datamodel/sig_config.go
anujmaheshwari1: I had thought so but its gonna be testing only after which the version should be same as linuxSIGimage version, hence I did not bother creating a different json file.

pkg/agent/datamodel/sig_config.go
anujmaheshwari1: I did :) https://github.com/Azure/AgentBaker/pull/3370/files#r1254790224

pkg/agent/datamodel/sig_config.go
cameronmeissner: makes sense, though I was mainly just talking about a const in the go file here, no need to deal with the json embedding - either way is fine tho

pkg/agent/datamodel/sig_config.go
cameronmeissner: noice I missed that lol

pkg/agent/datamodel/sig_config.go
anujmaheshwari1: will let it stay as is if it isnt blocking. will ofcourse revisit once its actively being used/built.

cameronmeissner (approved):

PR #3400, opened 2023-07-07
feat: Enable cgroupv2 support for Azure Linux

henryli001:
**What type of PR is this?**

Currently in AKS, the condition to turn on cgroupv2 is entirely dependent on whether the distro is Ubuntu 2204 or not. Modify this condition by also including Azure Linux, which will be onboarded to cgroupv2 starting from Kubernetes version 1.27.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


hbeberman:
The changes in this PR should be the last in the sequence for enabling cgroupv2 since we want to make sure the cgroup enabled images are available and selected for nodepools that are deploying with `SystemdCgroup = true`

pkg/agent/utils.go
```diff
@@ -263,6 +263,16 @@ func IsKubernetesVersionGe(actualVersion, version string) bool {
 	return v1.GE(v2)
 }
 
+// IsAzureLinuxCgroupv2 returns true if the VHD distro is Azure Linux and the Kubernetes version is greater than or equal to 1.27.0
+func IsAzureLinuxCgroupv2(config *datamodel.NodeBootstrappingConfiguration) bool {
+	if config.AgentPoolProfile.Distro.IsCBLMarinerDistro() || isMariner(config.OSSKU) {
```
hbeberman, line 268: No need to check distro since isMariner(config.OSSKU) **should** already be true for every Mariner/AzureLinux distro.

Key word being should, this PR https://github.com/Azure/AgentBaker/pull/3399/files needs to go in first then you can drop the IsCBLMarinerDistro check.

pkg/agent/utils.go
```diff
@@ -263,6 +263,16 @@ func IsKubernetesVersionGe(actualVersion, version string) bool {
...
+		if IsKubernetesVersionGe(config.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion, "1.27.0") {
```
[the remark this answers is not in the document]
hbeberman: Yeah, the images aren't in yet, we may end up retargeting this change to 1.28 for safety.

pkg/agent/utils.go
```diff
@@ -263,6 +263,17 @@ func IsKubernetesVersionGe(actualVersion, version string) bool {
 	return v1.GE(v2)
 }
 
+// IsAzureLinuxCgroupv2 returns true if the VHD distro is Azure Linux and the Kubernetes
+// version is greater than or equal to 1.28.0.
+func IsAzureLinuxCgroupv2(config *datamodel.NodeBootstrappingConfiguration) bool {
```
cameronmeissner, line 268: could we theoretically just do something like `osSku == datamodel.OSSKUAzureLinux`? since that'll be the only SKU with cgroupsv2?

this func could theoretically return true if we're on a Mariner distro in k8s 1.28 (AzureLinux is the default for this case, but could cx still specify a MarinerV2cgroupsv1 image on 1.28?)

henryli001: No just checking this condition alone will still include CgroupV1. This is because datamodel.OSSKUAzureLinux is not a one-to-one-match to Azure Linux 2.0 CgroupV2 in RP, it actually points to the overall Azure Linux 2.0. The only way to differentiate between Azure Linux 2.0 CgroupV1 and Azure Linux 2.0 CgroupV2 is to check the k8s version until the day when Azure Linux 2.0 CgroupV1 is entirely deprecated.

cameronmeissner (approved):

pkg/agent/utils.go
cameronmeissner: discussed offline, using a static check for now

henryli001: After internal discussion, the decision is to re-implement this logic by following what was done for Ubuntu 2204 where I'll create a list that contains all Azure Linux CgorupV2 distros and use this to determine whether IsCgroupV2 is true or false.

PR #3402, opened 2023-07-07
fix: set UseDomains to true for systemd-networkd on Azure Linux

SeanDougherty:
**What type of PR is this?**
/kind bug

Adding UseDomains=true to systemd-networkd's 99-dhcp-en.network config file to populate the search domain for /etc/resolve.conf

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


hbeberman (approved):

alexeldeib (approved):

alexeldeib:
thanks! TIL we already had a drop in there. was that actually the reason for the issue - it would be true, but we wiped it? or is it really set the other way in the base image too?

need a rebase but lgtm.

PR #3449, opened 2023-07-21
fix: repairing cgroup telemetry for memory and pressure metrics

Antonio-Villarreal:
**What type of PR is this?**
Repairs cgroup telemetry for memory and pressure metrics

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cgroup-memory-telemetry.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+set -o nounset
+set -o pipefail
+
```
alexeldeib, line 5: might be worth doing something like this to avoid stray issues:

```
find /var/log/azure/Microsoft.Azure.Extensions.CustomScript/events/ -mtime +5 -type f -delete
```

this would delete events older than 5 days, which would mean we only clean up after 1440 events have been backlogged (24*60 / 5 * 5)

parts/linux/cloud-init/artifacts/cgroup-pressure-telemetry.sh
```diff
@@ -0,0 +1,355 @@
+#!/bin/bash
+
+set -o nounset
+set -o pipefail
+
+EVENTS_LOGGING_DIR=/var/log/azure/Microsoft.Azure.Extensions.CustomScript/events/
```
alexeldeib, line 6: https://github.com/Azure/AgentBaker/pull/3449/files#r1273705680

ganeshkumar5699 (approved):

PR #3451, opened 2023-07-21
fix: dynamically generate sas tokens

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -70,7 +88,7 @@ sas=$(az disk grant-access --ids $disk_resource_id --duration-in-seconds 3600 --
 
 echo "Uploading $disk_resource_id to ${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"
 
-azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd${CLASSIC_SAS_TOKEN}" --recursive=true
+azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd?${storage_sas_token}" --recursive=true
```
AbelHu, line 91: CLASSIC_SAS_TOKEN is used by Windows. So it seems like that this change will block Windows to publish the official VHD.

AbelHu (dismissed):
It seems like that this PR will block Windows to publish the official VHDs.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
AbelHu: It seems like that this PR will generate the SAS token dynamically.

AbelHu: If the sas token is leaked, I think that we always need to rotate the storage account key. So what is the differnce with/without this change?

wanqingfu: FYI I think it's passed from pipeline variable `STORAGE_ACCT_SAS_TOKEN` to `CLASSIC_SAS_TOKEN` in `.builder-release-template-windows.yaml`?

alexeldeib: discussed offline, but the point is not to prevent the leak, it's to reduce manual effort in maintaining pipeline secrets.

theoretically we can use MSI + generated SAS in all places and never rotate secrets.

will cleanup CLASSIC_SAS_TOKEN and test a bit more

ganeshkumar5699 (approved):

PR #3453, opened 2023-07-24
fix: set umask in a way that cis audits recognize

tobiasb-ms:
**What type of PR is this?**
/kind bug

For CIS compliance, we need `umask` set to `027`. And, due to the CIS tooling, we need to not set it to anything else anywhere in `/etc/profile.d/`. This had initially been fixed in CBL-Mariner, but that was reverted because it broke some customers. Since we know that for AKS, we already technically have it set to `027` via `/etc/profile`, it's safe to make a change to work with the CIS tooling.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cis.sh
+++ b/parts/linux/cloud-init/artifacts/cis.sh
@@ -157,8 +157,22 @@ configureCoreDump() {
     replaceOrAppendCoreDump ProcessSizeMax 0
 }
 
-fixDefaultUmaskForAccountCreation() {
+fixUmaskSettings() {
+    # CIS requires the default UMASK for account creation to be set to 027, so change that in /etc/login.defs.
     replaceOrAppendLoginDefs UMASK 027
+
+    # It also requires that nothing in etc/profile.d sets umask to anything less restrictive than that.
+    # Mariner sets umask directly in /etc/profile after sourcing everything in /etc/profile.d. But it also has /etc/profile.d/umask.sh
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -157,8 +157,24 @@ configureCoreDump() {
...
+    # the package that supplies it, it would just be copied over again.
+    # This is complicated by an oddity/bug in the auditing script cis uses, which will flag line in a file with the work umask in the file name
+    # that doesn't set umask correctly. So we can't just comment out all the lines or have any comments that explain what we're doing.
+    # So since we can't delete the file, we just overwrite it with the correct umask setting. This duplicates what /etc/profile does, but
+    # it does no harm and works with the tools.
+    local umask_sh="/etc/profile.d/umask.sh"
+    if [[ "${OS}" == "${MARINER_OS_NAME}" && "${OS_VERSION}" == "2.0" && -f "${umask_sh}" ]]; then
+        echo "umask 027" >${umask_sh}
+        # umask is already set up to as 027, but we want this to be world-readable.
+        chmod 644 ${umask_sh}
```
alexeldeib, line 176: confused here - the file is on disk and we're configuring the umask here. if it's 027 in the base image, shouldn't this already work? or is the chmod wrong in base image too? 

if it's because of the umask here, same thing - doesn't the file already exist?

alexeldeib (approved):

hbeberman (approved):

PR #3455, opened 2023-07-25
feat(private egress): add SecurityProfile to datamodel

karataliu:
**What type of PR is this?**
/kind feature

To support private egress feature change in agentbaker. We plan to make it in three parts:
Part 1 (this PR), adds data model, and test baseline
Part 2, update provision script to log value from data model, verify test basline change
Part 3, update real logic in provision script

**Requirements**:
  
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/types.go
```diff
@@ -2106,3 +2107,26 @@ func (a *AgentPoolProfile) IsDisableWindowsOutboundNat() bool {
...
 }
+
+// SecurityProfile begin.
+type SecurityProfile struct {
+	PrivateEgress *PrivateEgress `json:"privateEgress,omitempty"`
+}
+
+type PrivateEgress struct {
+	Enabled      bool   `json:"enabled"`
+	ProxyAddress string `json:"proxyAddress"`
```
AbelHu, line 2118: Will it be possible to support multiple addresses?

karataliu: the proxy is managed by AKS. for current design, it's always single.

pkg/agent/datamodel/types.go
```diff
@@ -2106,3 +2107,26 @@ func (a *AgentPoolProfile) IsDisableWindowsOutboundNat() bool {
...
+}
+
+func (s *SecurityProfile) GetProxyAddress() string {
+	if s == nil {
```
AbelHu, line 2122: NIT: I prefer to use 
```
if s != nil && s.PrivateEgress != nil && s.PrivateEgress.Enabled {
  return privateEgress.ProxyAddress
}
return ""
```

pkg/agent/baker_test.go
```diff
@@ -1023,7 +1023,14 @@ oom_score = 0
 				Expect(sysctlContent).To(ContainSubstring("net.ipv4.neigh.default.gc_thresh2=8192"))
 				Expect(sysctlContent).To(ContainSubstring("net.ipv4.neigh.default.gc_thresh3=16384"))
 				Expect(sysctlContent).To(ContainSubstring("net.ipv4.ip_local_reserved_ports=65330"))
-			}))
+			}),
+		Entry("AKSUbuntu2204 with SecurityProfile", "AKSUbuntu2204+SecurityProfile", "1.26.0",
```
karataliu, line 1027: This is to add a baseline for now, all SecurityProfile related change can use this snapshot. Same for windows part later.

pkg/agent/baker_test.go
```diff
@@ -1317,7 +1324,14 @@ var _ = Describe("Assert generated customData and cseCmd for Windows", func() {
 		Entry("AKSWindows2019 with k8s version 1.19 + FIPS", "AKSWindows2019+K8S119+FIPS", "1.19.0",
 			func(config *datamodel.NodeBootstrappingConfiguration) {
 				config.FIPSEnabled = true
-			}))
+			}),
+		Entry("AKSWindows2019 with SecurityProfile", "AKSWindows2019+SecurityProfile", "1.26.0",
```
AbelHu, line 1328: Prefer to use AKSWindows2022 since the default Windows OS is 2022 in k8s 1.25+

karataliu: Currently all cases are on windows2019, and it's hardcoded:

Distro:              datamodel.Distro("aks-windows-2019"),

WindowsSku:                    "aks-2019-datacenter-core-smalldisk-2011",


Do you suggest to override the SKU in this case, or update for all cases later?

pkg/agent/datamodel/types.go
karataliu: updated.

pkg/agent/baker_test.go
AbelHu: I see. You can still use Windows2019. We will update it later.

pkg/agent/datamodel/types.go
AbelHu: @karataliu  Could it be possible to support multiple addresses in future? We hit similar issues in other features and we have to add a new property later. It is better to use a string array to avoid such issue. How do you think?

AbelHu (approved):
LGTM. Thanks for the PR

pkg/agent/datamodel/types.go
karataliu: this is service managed address, so suppose won't be multiple.

pkg/agent/datamodel/types.go
```diff
@@ -2106,3 +2107,22 @@ func (a *AgentPoolProfile) IsDisableWindowsOutboundNat() bool {
...
 		*a.AgentPoolWindowsProfile.DisableOutboundNat
 }
+
+// SecurityProfile begin.
+type SecurityProfile struct {
+	PrivateEgress *PrivateEgress `json:"privateEgress,omitempty"`
+}
+
+type PrivateEgress struct {
+	Enabled      bool   `json:"enabled"`
```
bingosummer, line 2117: Add `ContainerRegistryFQDN string` so that we can use it in `getOutBoundCmd`.
https://github.com/Azure/AgentBaker/blob/master/pkg/agent/variables.go#L173

bingosummer (changes requested):

pkg/agent/datamodel/types.go
karataliu: added ContainerRegistryServer

bingosummer (approved):

anujmaheshwari1 (approved):

karataliu:
created https://github.com/Azure/AgentBaker/pull/3464 from branch instead of fork to trigger more test run.

karataliu:
close in favor of #3464

PR #3461, opened 2023-07-25
fix: typos in testCronPermissions

tobiasb-ms:
**What type of PR is this?**
/kind test

When I introduced tests for permissions on `cron`-related files, I had a typo in the array name for a set of files to check, which resulted in not actually checking the permissions on those files. Turns out the script that set the permissions worked properly, so there wasn't a bug -- this is a test-only issue.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


alexeldeib (approved):
give an s, take an s

hbeberman (approved):

PR #3471, opened 2023-07-26
feat: add check outbound connectivity script

alyssa1303:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/check-outbound-network.sh
```diff
@@ -0,0 +1,182 @@
...
+# Set the number of times to retry before logging an error
+retries=3
+
+# Set the delay between retries in seconds
+delay=5
+
+for url in ${urlLists[@]};
+do
+    # Check DNS 
+    nslookup $url > /dev/null
```
phealy, line 114: I'd suggest using dig instead of nslookup here, as dig is a more flexible tool with better trace output.

PR #3479, opened 2023-07-27
feat: add private blob storage download for k8s

alexeldeib:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder.yaml
+++ b/.pipelines/.vsts-vhd-builder.yaml
@@ -53,6 +53,25 @@ stages:
             echo "##vso[task.setvariable variable=SKU_NAME]$SKU_NAME"
             echo "Set SKU_NAME to $SKU_NAME"
           displayName: Set SKU Name
+          bash: |
+            echo "Logging into Azure with agent VM MSI..."
+            mkdir -p private/usr/local/bin
+            az login --identity
+            for version in "${HOTFIX_VERSIONS}"; do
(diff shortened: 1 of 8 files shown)


.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -53,6 +53,25 @@ stages:
...
+              --auth-mod login \
+              --account-name ${PRIVATE_STORAGE_ACCOUNT} \
+              --container "kubernetes" \
+              --name $(printf "%s/binaries/node/linux/amd64/kubelet" ${version}) \
+              --file private/usr/local/bin/kubelet-${patch_no_v}
+              chown root:root private/usr/local/bin/kubelet-${patch_no_v}
+              chmod a+x private/usr/local/bin/kubelet-${patch_no_v}
+            done
+            tar -C private -czvf private.tar.gz .
+            tar -tzvf ./private.tar.gz
```
AbelHu, line 73: how can we use private.tar.gz to differentiate different versions?

alexeldeib:
In this example, it contains all versions, as usr/local/bin/kubelet-1.27.2
or whatever

So when we unpack it, it overwrites any normal binaries with the same non
hotfix version in /usr/local/bin

The download is a for loop with the file output containing the version, NOT
downloading a tar.gz directly

This was just my quick approach, but it worked out okay, better ideas
welcome

PR #3483, opened 2023-07-27
test: remove dependency on diff from tests

tobiasb-ms:
**What type of PR is this?**
/kind test
/kind regression

A recently introduced test used the `diff` utility, which isn't installed by default on our mariner images.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

In addition to normal checkin pipelines, I also ran a mariner-specific pipeline to test with the most recent build of mariner: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=398811&view=results

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -479,23 +479,30 @@ testNetworkSettings() {
 # Ensures that the content /etc/profile.d/umask.sh is correct, per code in
 # <repo-root>/parts/linux/cloud-init/artifacts/cis.sh
 testUmaskSettings() {
-  local test="testUmaskSettings"
-  local settings_file=/etc/profile.d/umask.sh
-  local expected_settings_file_content='umask 027'
-  echo "$test:Start"
+    local test="testUmaskSettings"
+    local settings_file=/etc/profile.d/umask.sh
+    local expected_settings_file_content='umask 027'
+    echo "$test:Start"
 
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -479,23 +479,30 @@ testNetworkSettings() {
...
+    echo "$test:Start"
+
+    # If the settings file exists, it must just be a single line that sets umask properly.
+    if [[ -f "${settings_file}" ]]; then
+        echo "${test}: Checking that the contents of ${settings_file} is exactly '${expected_settings_file_content}'"
+
+        # Command substitution (like file_contents=$(cat "${settings_file}")) strips trailing newlines, so we use mapfile instead.
+        # This creates an array of the lines in the file, and then we join them back together by expanding the array into a single string.
+        local file_contents_array=()
+        mapfile <"${settings_file}" file_contents_array
```
alexeldeib, line 494: do you even need mapfile vs cat here for simple string?

I also suspect `git diff --no-index` may work?

alexeldeib (approved):

vhdbuilder/packer/test/linux-vhd-content-test.sh
tobiasb-ms: We need to get trailing whitespace, which command-substitution with `cat` will trim. `git diff` is a good idea, though it feels weird to use `git` to do a non-git thing if that makes sense?

PR #3538, opened 2023-08-15
feat: add telemetry for CNI conflist creation

wedaly:
**What type of PR is this?**

/kind feature

Nodes remain in NotReady state until a CNI conflist file is created in /etc/cni/net.d Depending on the network configuration, the conflist may be created by containerd itself (based on cni.conf_template configuration) or by a daemonset init container.

To measure CNI conflist creation delay, add a new systemd service that waits for the CNI conflist file to get created, then sends an event via walinuxagent with the conflist create time.

If no CNI conflist is created within a timeout, log an event saying that the CNI conflist was not created. This allows us to verify that telemetry is working even if something else prevents the CNI conflist from being created.

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
add cni-telemetry systemd service to report CNI conflist creation time
```


parts/linux/cloud-init/artifacts/cni-telemetry.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Emit CNI telemetry
+
+[Service]
+Type=simple
+RemainAfterExit=true
+Environment="PYTHONUNBUFFERED=1"
+ExecStart=/usr/bin/python3 /opt/scripts/cni-telemetry.py
```
wedaly, line 8: I'm open to rewriting this in bash for consistency with other scripts if people feel strongly about it (I just happen to be more comfortable with Python, so I could get something up-and-running faster this way).

parts/linux/cloud-init/artifacts/cni-telemetry.py
```diff
@@ -0,0 +1,102 @@
...
+        emit_event({
+            "CNIConflistFound": False,
+            "CNIConflistCreateTimestamp": None,
+        })
+    else:
+        print("Emitting event that CNI conflist found with timestamp")
+        cni_ts = datetime.utcfromtimestamp(ts).isoformat(sep=" ", timespec="milliseconds")
+        emit_event({
+            "CNIConflistFound": True,
+            "CNIConflistCreateTimestamp": cni_ts,
```
wedaly, line 101: Is there a reliable way to get the time that the node started? Ideally I'd like to include the duration from node start to cni_ts so we can report on the delay without needing to join with any other data source, but I wasn't sure where to pull that from (maybe current time minus /proc/uptime would be accurate enough?)

wedaly: thinking about this more, I think what I really care about is the time from when kubelet starts to when the CNI is created, which we're [already capturing in cse_start.sh](https://github.com/Azure/AgentBaker/blob/9c7c539087928bbc1a4dbce8e84d403bbe88ac6c/parts/linux/cloud-init/artifacts/cse_start.sh#L19). So I'll just leave this as-is for now.

PR #3539, opened 2023-08-16
feat: remove mirror enablement, change mirror download to prod URL, update version

ganeshkumar5699:
**What type of PR is this?**
/kind feature

This updates the acr-mirror (for artifact streaming) download URL to the production version and uses a new version of the mirror that has the mirror service disabled by default (need explicit enablement later).

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -160,15 +160,14 @@ done
 
 installAndConfigureArtifactStreaming() {
   # download acr-mirror proxy
-  MIRROR_PROXY_VERSION='19'
+  MIRROR_PROXY_VERSION='7'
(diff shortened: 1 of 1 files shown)


UtheMan (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -160,15 +160,14 @@ done
 
 installAndConfigureArtifactStreaming() {
   # download acr-mirror proxy
-  MIRROR_PROXY_VERSION='19'
+  MIRROR_PROXY_VERSION='7'
```
UtheMan, line 163: Where is that version coming from? How do we know which one to use?

PR #3378, opened 2023-06-30
Enable kata-cc image curation on top of mariner 2.0 base image

miz060:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Move kata-cc image curation logic inside AgentBaker so that the Mariner kata-cc release flow can be accomplished all within AKS.

**Which issue(s) this PR fixes**:
Fixes #

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [x] adds unit tests
- [x] tested upgrade from previous version

**Special notes for your reviewer**:
1. [Kata test pipeline build](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=75728579&view=logs&j=2ed7bee1-13c6-5ac8-633f-26820f0b06a6&t=25048989-8415-5618-db0f-a93111045e0d) manually verified with AKS node cluster.
2.  All mariner/azure linux build: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=81213905&view=results

**Release note**:
```
none
```

PR #3541, opened 2023-08-16
Add CustomizedImageKata distro to enable Kata provisioning for custom header BYOI

miz060:
**What type of PR is this?**
/kind feature

Add CustomizedImageKata distro to enable Kata provisioning for custom header BYOI.

Fixes #

**Requirements**:
  
- [x] includes documentation
- [x] adds unit tests

1. Passed manual tests using aksdev using https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=79932221&view=results

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -1264,6 +1264,31 @@ root = "{{GetDataDir}}"{{- end}}
     {{- if IsKata }}
     [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata]
       runtime_type = "io.containerd.kata.v2"
+    [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata]
```
cameronmeissner, line 1267: is this something that should be added separately?

anujmaheshwari1: yep, is this valid for regular kata images too?

miz060: Yes, the same change would apply to regular kata release image as well.
You are right. This change has been moved to #3378 where it would be more fitting.

pkg/agent/baker.go
```diff
@@ -480,7 +480,7 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return profile.Distro.IsKataDistro()
 		},
 		"IsCustomImage": func() bool {
```
cameronmeissner, line 482: can we add something in baker_test.go to make sure this piece of the func map gets called correctly?

basically a test where the node bootstrapping config specifies a custom kata image

miz060: Good point! Added a test case.

cameronmeissner (approved):

hbeberman (approved):

anujmaheshwari1 (approved):

PR #3573, opened 2023-08-25
chore: cache busybox in VHD

wedaly:
**What type of PR is this?**

/kind cleanup

In AKS kubernetes version 1.25 and later, kube-proxy init container runs busybox:1.35. We used to use the kube-proxy image, but had to switch when upstream started building from distroless.

The busybox image is 4.27 MB. This is downloaded by almost every AKS node on startup, so I think we should cache it in the VHD.

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
cache busybox 1.35 in the VHD for kube-proxy init container
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -32,6 +32,13 @@
       ]
     },
     {
+      "downloadURL": "mcr.microsoft.com/mirror/docker/library/busybox:*",
+      "amd64OnlyVersions": [],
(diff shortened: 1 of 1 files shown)


wedaly:
same as #3572 from a branch in this repo instead of a fork

ganeshkumar5699 (approved):
Lgtm

PR #3647, opened 2023-09-27
feat: add artifact streaming CSE changes for 22.04

ganeshkumarashok:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -340,6 +345,15 @@ ensureTeleportd() {
     systemctlEnableAndStart teleportd || exit $ERR_SYSTEMCTL_START_FAIL
 }
 
+ensureArtifactStreaming(){
+    touch /etc/default/artifact-streaming.test
+    echo "in ensureArtifactStreaming"
+    if [[ "${UBUNTU_RELEASE}" == "22.04" && ( "${CPU_ARCH}" == "amd64" ) ]]; then
```
cameronmeissner, line 351:
```suggestion
    if [[ "${UBUNTU_RELEASE}" == "22.04" ]]  && [[ "${CPU_ARCH}" == "amd64" ]]; then
```

e2e/template.go
```diff
@@ -351,6 +351,7 @@ func baseTemplate() *datamodel.NodeBootstrappingConfiguration {
 		EnableNvidia:                   false,
 		EnableACRTeleportPlugin:        false,
 		TeleportdPluginURL:             "",
+		EnableArtifactStreaming:        true,
```
cameronmeissner, line 354: is this something we want enabled by default for all e2e tests? I'd maybe lean more towards defaulting to false and having specific scenarios for this to be set to true within (though we do have a fair number of scenarios now... I've actually been thinking about adding the concept of "suites" which would be pre-configured scenario-sets designed to test certain things)

cameronmeissner: compromise could also be that we start with false for now and later once this has rolled out we can default it to true

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -306,6 +306,11 @@ ensureContainerd() {
   if [ "${TELEPORT_ENABLED}" == "true" ]; then
     ensureTeleportd
   fi
+  if [ "${ARTIFACT_STREAMING_ENABLED}" == "true" ]; then
+    echo "ARTIFACT STREAMING is enabled, running test function for artifact streaming"
+    ensureArtifactStreaming
```
cameronmeissner, line 311:
```suggestion
    logs_to_events ensureArtifactStreaming
```

maybe we want to log this as an actual CSE step?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -340,6 +345,15 @@ ensureTeleportd() {
     systemctlEnableAndStart teleportd || exit $ERR_SYSTEMCTL_START_FAIL
 }
 
+ensureArtifactStreaming(){
+    touch /etc/default/artifact-streaming.test
+    echo "in ensureArtifactStreaming"
+    if [[ "${UBUNTU_RELEASE}" == "22.04" && ( "${CPU_ARCH}" == "amd64" ) ]]; then
+        sudo /opt/acr/tools/overlaybd/install.sh || exit $ERR_ARTIFACT_STREAMING_DOWNLOAD_INSTALL
```
cameronmeissner, line 352: could maybe even call `logs_to_events` on these lines with "AKS.CSE.ensureArtifactStreaming.overlaybdInstall" or something, some with enable-http-auth.sh

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -340,6 +345,15 @@ ensureTeleportd() {
     systemctlEnableAndStart teleportd || exit $ERR_SYSTEMCTL_START_FAIL
 }
 
+ensureArtifactStreaming(){
```
cameronmeissner, line 348:
```suggestion
ensureArtifactStreaming() {
```

ganeshkumar5699:
This PR is currently in the testing phase - not ready for review

e2e/template.go
ganeshkumar5699: Agreed. This is for ab e2e testing now -- PR is not yet ready for review.

PR #3653, opened 2023-09-28
feat: add EnableSecureTLSBootstrapping to bootstrap config

cameronmeissner:
**What type of PR is this?**

/kind feature

This is the initial PR for secure TLS bootstrap integration work. It simply adds a new property to the node bootstrapping configuration datamodel which specifies whether or not secure TLS bootstrapping is enabled by RP. Subsequent PRs will use this value to implement the secure TLS bootstrap-specific logic.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/types.go
```diff
@@ -1645,19 +1645,23 @@ type NodeBootstrappingConfiguration struct {
...
-	SIGConfig                      SIGConfig
-	IsARM64                        bool
-	CustomCATrustConfig            *CustomCATrustConfig
-	DisableUnattendedUpgrades      bool
-	SSHStatus                      SSHStatus
-	DisableCustomData              bool
+	// EnableSecureTLSBootstraping - when this feature is enabled we don't hard-code TLS bootstrap tokens at all,
+	// instead we create a modified bootstrap kubeconfig which points towards the STLS bootstrap client-go
+	// credential plugin installed on the VHD, which will be responsible for generating TLS bootstrap tokens on the fly
+	EnableSecureTLSBootstrapping bool
```
cameronmeissner, line 1651: this will be set to false until enabled on the RP-side

pkg/agent/baker.go
```diff
@@ -379,7 +379,10 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return IsKubeletConfigFileEnabled(cs, profile, config.EnableKubeletConfigFile)
 		},
 		"IsKubeletClientTLSBootstrappingEnabled": func() bool {
-			return IsKubeletClientTLSBootstrappingEnabled(config.KubeletClientTLSBootstrapToken)
+			return config.EnableSecureTLSBootstrapping || IsKubeletClientTLSBootstrappingEnabled(config.KubeletClientTLSBootstrapToken)
```
cameronmeissner, line 382: since we're OR'ing, EnableSecureTLSBootstrapping being false won't affect the overall value

UtheMan: clueless question - whats the difference between config having EnableSecureTLSBootstrapping set to true and IsKubeletClientTLSBootstrappingEnabled returning true?

UtheMan (approved):

pkg/agent/baker.go
```diff
@@ -379,7 +379,10 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+		},
+		"EnableSecureTLSBootstrapping": func() bool {
+			return config.EnableSecureTLSBootstrapping
```
anujmaheshwari1, line 385: do we want a nil check here?

cameronmeissner: it's a bool not a *bool, so default will be false

pkg/agent/baker.go
cameronmeissner: basically, `EnableSecureTLSBootstrap` will be true when we are doing TLS bootstrapping but we're NOT using a hard-coded token, `IsKubeletClientTLSBootstrappingEnabled` will be true when the node bootstrapping config we get from RP includes a hard-coded token, also implying that we're doing TLS bootstrapping.

so `IsKubeletClientTLSBootstrappingEnabled` overall will return true when we're doing _some_ form of TLS bootstrapping, whether that be the secure way or with a hard-coded token

pkg/agent/baker.go
anujmaheshwari1: oops my bad, didnt see it wasnt *bool

pkg/agent/baker.go
UtheMan: Thanks, makes sense - might be a good idea to add this as a comment somewhere maybe for the future reference?

cameronmeissner: did some renaming to make this a bit more clear, also added comments :)

UtheMan (approved):

maxwolffe:
@alexeldeib , @cameronmeissner - Hey friends - when are y'all planning on releasing this?

DB is planning on moving off of our custom AKS image (described in some detail here - https://github.com/Azure/AgentBaker/issues/2815) - but we don't plan on being able to be fully off of it until the end of December. 

I'm wondering if there's a way to explicitly exclude our clusters or subscriptions (within the MSFT tenant) from this feature until we've completed that migration?

maxwolffe:
cc @juan-lee , @yangl900

PR #3658, opened 2023-10-02
feat: add EnableArtifactStreaming to NBC and add helper code

ganeshkumar5699:
**What type of PR is this?**
/kind feature

This adds the `EnableArtifactStreaming` field to the NodeBootstrappingConfig so it can be enabled on AKS nodes during provisioning. Additional code for scaffolding is also part of this. With just this PR, enabling it should be a no-op.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -150,6 +150,7 @@ KUBENET_TEMPLATE="{{GetKubenetTemplate}}"
 CONTAINERD_CONFIG_CONTENT="{{GetContainerdConfigContent}}"
 CONTAINERD_CONFIG_NO_GPU_CONTENT="{{GetContainerdConfigNoGPUContent}}"
 IS_KATA="{{IsKata}}"
+ARTIFACT_STREAMING_ENABLED="{{ArtifactStreamingEnabled}}"
```
anujmaheshwari1, line 153: if we follow the other fields, this should have a prefix like "Is", "Has" or "Get".
For ArtifactStreaming, maybe "IsArtifactStreamingEnabled" is appropriate?

pkg/agent/baker.go
```diff
@@ -953,6 +953,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetPrivateEgressProxyAddress": func() string {
 			return config.ContainerService.Properties.SecurityProfile.GetProxyAddress()
 		},
+		"ArtifactStreamingEnabled": func() bool {
```
anujmaheshwari1, line 956: we should add a UT in baker_test.go?

parts/linux/cloud-init/artifacts/cse_cmd.sh
ganeshkumar5699: There is a mix of fields, like `FIPSEnabled` and others which have such a prefix. I think either one is fine for NBC. What do you think?

pkg/agent/baker.go
ganeshkumar5699: Will be adding it after the behavior change in the next PR. If `ArtifactStreamingEnabled` is true, then it should have a particular containerd config. If not true, then the standard one.

parts/linux/cloud-init/artifacts/cse_cmd.sh
cameronmeissner: I personally like `ENABLE_ARTIFACT_STREAMING` a bit better, along with `EnableArtifactStreaming`, which is what you already have :)

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_cmd.sh
ganeshkumar5699: Discussed and we want to keep the state, so switched to a combination of both.

PR #3664, opened 2023-10-03
style: remove unused variable and rename function

AlisonB319:
**What type of PR is this?**

/kind cleanup
/kind failing-test

**What this PR does / why we need it**: Fixes linting

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/mocks.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // CreateMockContainerService returns a mock container service for testing purposes.
-func CreateMockContainerService(containerServiceName, orchestratorVersion string, masterCount, agentCount int, certs bool) *ContainerService {
```
AlisonB319, line 12: CreateMockContainerService is not being used anywhere, removing these from the function should be okay.

pkg/agent/bakerapi.go
```diff
@@ -23,7 +23,7 @@ func NewAgentBaker() (AgentBaker, error) {
 
 type agentBakerImpl struct{}
 
-func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context,
+func (agentBaker *agentBakerImpl) GetNodeBootstrapping(_ context.Context,
```
cameronmeissner, line 26: lol, we probably should be using ctx somehow...

pkg/agent/datamodel/mocks.go
cameronmeissner: can we remove the func altogether if it's not being used?

pkg/agent/bakerapi.go
cameronmeissner: hmmmmm https://github.com/Azure/AgentBaker/blob/10134a39958e3ee9a04e76644c1a0632d163ae99/apiserver/getnodebootstrapdata.go#L23C1-L25C16

pkg/agent/datamodel/mocks.go
AlisonB319: Do you think this mock for a container service will be used again? Or does it seem outdated?

pkg/agent/bakerapi.go
AlisonB319: Should we be logging the ctx or is it not being passed to anything atm when it should be?

cameronmeissner: honestly I think it would be best to leave `ctx` as it is and add a TODO note to figure out what to do with it, as it stands the timeout we set in the calling func isn't having any effect when it really should be...

I'd be open to further discussion around this, maybe we legitimately don't need it (which would kinda make sense since GetNodeBootstrapping doesn't make any outbound RPCs or anything)

cameronmeissner: @anujmaheshwari1 @ganesshkumar @UtheMan

PR #3672, opened 2023-10-04
fix: run abe2e tests in custom location

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/template.go
```diff
@@ -10,15 +10,15 @@ import (
 // TODO(ace): minimize the actual required defaults.
 // this is what we previously used for bash e2e from e2e/nodebootstrapping_template.json.
 // which itself was extracted from baker_test.go logic, which was inherited from aks-engine.
-func baseTemplate() *datamodel.NodeBootstrappingConfiguration {
+func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
 	var (
 		trueConst  = true
 		falseConst = false
 	)
 	return &datamodel.NodeBootstrappingConfiguration{
```
cameronmeissner, line 18: we need to do some work on getting this actually completely correct depending on k8s version, cluster fqdn, etc.

e2e/cluster.go
```diff
@@ -18,6 +18,14 @@ const (
 	managedClusterResourceType = "Microsoft.ContainerService/managedClusters"
 )
 
+// This map is used during cluster creation to check what VM size should
+// be used to create the single default agentpool used for running cluster essentials
+// and the jumpbox pods/resources.
+var locationToDefaultClusterAgentPoolVMSize = map[string]string{
+	"eastus":         "Standard_DS2_v2",
+	"southcentralus": "Standard_ND96isr_H100_v5",
```
ganeshkumar5699, line 26: We shouldn't default to this SKU for southcentralus because there's very limited capacity and it's expensive. 
Having a variable to update the SKU once and have it be handled everywhere would be nice though

cameronmeissner: I guess we ned to figure out which default we _can_ us here, it seems like these expensive h100-based SKUs are the only ones to choose from there?

ganeshkumar5699: We don't have quota for this but we'll have to get quota for a cheaper one.

ganeshkumar5699 (approved):

PR #3690, opened 2023-10-10
feat: add support for secure TLS bootstrapping to CSE

cameronmeissner:
**What type of PR is this?**

/kind feature

Adds support for secure TLS bootstrapping on agent nodes - for now we only install the plugin at runtime since we're using an alpha version of the client-go plugin. Will add a step to the VHD build in a subsequent PR.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/10-tlsbootstrap.conf
cameronmeissner, line 1: removed this in favor of encoding directly into nodecustomdata.yaml so we can use go templates - this doesn't matter anyways since we'll end up actually using what comes from CSE.

parts/linux/cloud-init/artifacts/10-tlsbootstrap.conf
```diff
@@ -1,2 +1,2 @@
 [Service]
-Environment="KUBELET_TLS_BOOTSTRAP_FLAGS=--kubeconfig /var/lib/kubelet/kubeconfig --bootstrap-kubeconfig /var/lib/kubelet/bootstrap-kubeconfig"
+Environment="KUBELET_TLS_BOOTSTRAP_FLAGS=--kubeconfig /var/lib/kubelet/kubeconfig --bootstrap-kubeconfig /var/lib/kubelet/bootstrap-kubeconfig"
```
cameronmeissner, line 2: this is unchanged, showing an empty diff

PR #3702, opened 2023-10-12
chore: add repo depot for mariner in AGC

UtheMan:
**What type of PR is this?**
Adds repo depot endpoints for mariner in air-gapped cloud environment for package updates.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/init-aks-custom-cloud-mariner.sh
+++ b/parts/linux/cloud-init/artifacts/init-aks-custom-cloud-mariner.sh
@@ -17,11 +17,15 @@ cp /root/AzureCACertificates/*.crt /etc/pki/ca-trust/source/anchors/
 cloud-init status --wait
 
 marinerRepoDepotEndpoint="$(echo "${REPO_DEPOT_ENDPOINT}" | sed 's/\/ubuntu//')"
-for f in /etc/yum.repos.d/*.repo
-do
-    sed -i -e "s|https://packages.microsoft.com|${marinerRepoDepotEndpoint}/mariner/packages.microsoft.com|" $f
-    echo "## REPO - $f - MODIFIED"
-done
+if [[ "$marinerRepoDepotEndpoint" == "" ]]; then
+  >&2 echo "repo depot endpoint empty while running custom-cloud init script"
+else
+  for f in /etc/yum.repos.d/*.repo
+  do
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/init-aks-custom-cloud-mariner.sh
```diff
@@ -16,7 +16,12 @@ cp /root/AzureCACertificates/*.crt /etc/pki/ca-trust/source/anchors/
 
 cloud-init status --wait
 
-# TODO - Set the repoDepotEndpoint in a .repo file if package update becomes necessary
+marinerRepoDepotEndpoint="$(echo "${REPO_DEPOT_ENDPOINT}" | sed 's/\/ubuntu//')"
```
cameronmeissner, line 19: could we add an error for if this is empty in custom cloud case or something? might be a good sanity check

ganeshkumar5699 (approved):
Lgtm. If you can add the PR description too, that'd be nice for future reference :)

parts/linux/cloud-init/artifacts/init-aks-custom-cloud-mariner.sh
UtheMan: @hbeberman from your side, what would be the expected behavior here if that happens? Wrap this around an if-block and don't update repoDepot endpoints? Or throw some sort of error message?

hbeberman: I think throwing an error would be preferrable, you don't want to leave the VMs in a state where they assume they'll be getting automatic package updates, but the repos aren't actually pointing to the correct mirrors.

parts/linux/cloud-init/artifacts/init-aks-custom-cloud-mariner.sh
```diff
@@ -16,7 +16,16 @@ cp /root/AzureCACertificates/*.crt /etc/pki/ca-trust/source/anchors/
...
+if [[ "$marinerRepoDepotEndpoint" == "" ]]; then
```
UtheMan, line 20: wasn't sure if the fact that its empty is important enough to exit from the entire script. So far its just a warning that gets printed to logs.

hbeberman (approved):

PR #3715, opened 2023-10-18
cleanup: remove NEEDS_DOCKER_LOGIN from new bootstrap scripts

anujmaheshwari1:
**What type of PR is this?**
This is the first PR in retiring variables in cse_cmd.sh that are no longer going to be required for self contained VHDs. 
This PR removes NEEDS_DOCKER_LOGIN, which is not a necessity anymore since we only support containerd.
The new files bootstrap_<name>.sh will be the new scripts that will be used for bootstrapping with self contained VHDs.
These files will currently not be in used, but will be iteratively updated as we retire older fields.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/self-contained/bootstrap_cmd.sh
+++ b/self-contained/bootstrap_cmd.sh
@@ -1,0 +1,156 @@
+PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log";
+echo $(date),$(hostname) > ${PROVISION_OUTPUT};
+{{if ShouldEnableCustomData}}
+cloud-init status --wait > /dev/null 2>&1;
+[ $? -ne 0 ] && echo 'cloud-init failed' >> ${PROVISION_OUTPUT} && exit 1;
+echo "cloud-init succeeded" >> ${PROVISION_OUTPUT};
+{{end}}
+{{if IsAKSCustomCloud}}
+REPO_DEPOT_ENDPOINT="{{AKSCustomCloudRepoDepotEndpoint}}"
+{{GetInitAKSCustomCloudFilepath}} >> /var/log/azure/cluster-provision.log 2>&1;
+{{end}}
+ADMINUSER={{GetParameter "linuxAdminUsername"}}
+MOBY_VERSION={{GetParameter "mobyVersion"}}
+TENANT_ID={{GetVariable "tenantID"}}
+KUBERNETES_VERSION={{GetParameter "kubernetesVersion"}}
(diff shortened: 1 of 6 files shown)


self-contained/bootstrap_cmd.sh
```diff
@@ -0,0 +1,156 @@
...
+SHOULD_CONFIGURE_HTTP_PROXY_CA="{{ShouldConfigureHTTPProxyCA}}"
+HTTP_PROXY_TRUSTED_CA="{{GetHTTPProxyCA}}"
+SHOULD_CONFIGURE_CUSTOM_CA_TRUST="{{ShouldConfigureCustomCATrust}}"
+CUSTOM_CA_TRUST_COUNT="{{len GetCustomCATrustConfigCerts}}"
+{{range $i, $cert := GetCustomCATrustConfigCerts}}
+CUSTOM_CA_CERT_{{$i}}="{{$cert}}"
+{{end}}
+IS_KRUSTLET="{{IsKrustlet}}"
+GPU_NEEDS_FABRIC_MANAGER="{{GPUNeedsFabricManager}}"
+#NEEDS_DOCKER_LOGIN="{{and IsDockerContainerRuntime HasPrivateAzureRegistryServer}}" This field is no longer required for the new contract since Docker is out of support and its value depends on Container Runtime = Docker
```
anujmaheshwari1, line 86: there was just one usage in bootstrap_main.sh as follows:

https://github.com/Azure/AgentBaker/blob/278e87a73ccab798fcf1cc9c298f52fbf83d74fd/parts/linux/cloud-init/artifacts/cse_main.sh#L194

```
if [ "${NEEDS_DOCKER_LOGIN}" == "true" ]; then
    set +x
    docker login -u $SERVICE_PRINCIPAL_CLIENT_ID -p $SERVICE_PRINCIPAL_CLIENT_SECRET "${AZURE_PRIVATE_REGISTRY_SERVER}"
    set -x
fi
```

Devinwong (approved):

self-contained/bootstrap_cmd.sh
cameronmeissner: yeah we should be able to remove it I think - are all these bootstrap_*.sh files mostly copy-pasta from parts/linux/artifacts?

cameronmeissner (approved):

PR #3728, opened 2023-10-23
feat: exporting Node Bootstrapping Contract for external use

Devinwong:
**What type of PR is this?**

Fixes #

**Release note**:
```
none
```


pkg/nbcontracthelper/helper.go
```diff
@@ -0,0 +1,11 @@
+package nbcontracthelper
+
+const (
+	defaultLinuxUser = "azureuser"
+)
+
+func GetDefaultLinuxAdminUsername() string {
+	// var abc nbc.ApiServerConfig
```
anujmaheshwari1, line 8: what is var abc?

Devinwong: sorry this was generated by copilot. I should have removed it 🤣

anujmaheshwari1 (approved):

pkg/proto/readme.md
```diff
@@ -0,0 +1,8 @@
+Notes for protobuf
+
+1. Run this command to compile .proto to .pb.go file.
+```/AgentBaker/hack/tools/bin/buf generate -o .```
```
cameronmeissner, line 4: nit: add a target to root-level Makfile instead?

pkg/proto/readme.md
```diff
@@ -0,0 +1,8 @@
...
+
+1. If you haven't installed buf yet, run the following command, assumed /usr/local/bin is where your Go binary locates.
+
+```GO111MODULE=on GOBIN=/usr/local/bin go install github.com/bufbuild/buf/cmd/buf@v1.27.2```
```
anujmaheshwari1, line 8: Devin, we already have a make target to create the buf binary, https://github.com/Azure/AgentBaker/blob/c6fb221ae595a481af6380d777969916e5367634/hack/tools/Makefile#L25

pkg/proto/readme.md
Devinwong: Updated the makefile. Do you mean that?

pkg/proto/readme.md
Devinwong: updated to use that binary.

Makefile
```diff
@@ -96,6 +100,8 @@ generate: bootstrap
 	popd \
 	)
 	GENERATE_TEST_DATA="true" go test ./pkg/agent...
+	@echo "compiling proto files in ./pkg/proto/ to pb.go files"
```
cameronmeissner, line 103: do we actually need to do this on every `make generate`? I might lean towards keeping it as a separate target that we only run when changing those protos

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 )
 
 require (
+	github.com/Azure/AgentBaker/pkg/proto/gen/nbcontractv1 v0.0.0-20231024012513-889e5d49d683 // indirect
```
cameronmeissner, line 22: a bit confused why we're now requiring this

Makefile
Devinwong: Makes sense. So we will run "make compile-proto-files" manually when we think needed

go.mod
Devinwong: You are right. This is not needed. Ran "go mod tidy" to clean it

cameronmeissner (approved):

PR #3733, opened 2023-10-24
fix: check if systemd-timesyncd is dead, only stop+disable if not

anujmaheshwari1:
**What type of PR is this?**
Originally, we always stopped and disabled systemd-timesyncd, however, in the latest builds, we are getting the following errors:

```
timeout 120 systemctl stop systemd-timesyncd
==> azure-arm: Failed to stop systemd-timesyncd.service: Unit systemd-timesyncd.service not loaded
```

This PR first checks if that service is dead, if yes, it skips the stop and disable.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -108,10 +108,15 @@ installBpftrace() {
 }
 
 disableNtpAndTimesyncdInstallChrony() {
-    # Disable systemd-timesyncd
-    systemctl_stop 20 30 120 systemd-timesyncd || exit $ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
-    systemctl disable systemd-timesyncd || exit $ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
-
+    # Disable systemd-timesyncd if present
+    status=$(systemctl show -p SubState --value systemd-timesyncd)
+    if [ $status == 'dead' ]; then
```
anujmaheshwari1, line 113: we already do this for ntp, see L120-127

cameronmeissner: we should probably factor this out to some common func we can call whenever we want to mess with a service we _think_ is running, im more than fine with this for now though

cameronmeissner (approved):

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -108,10 +108,15 @@ installBpftrace() {
 }
 
 disableNtpAndTimesyncdInstallChrony() {
-    # Disable systemd-timesyncd
-    systemctl_stop 20 30 120 systemd-timesyncd || exit $ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
-    systemctl disable systemd-timesyncd || exit $ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
-
+    # Disable systemd-timesyncd if present
+    status=$(systemctl show -p SubState --value systemd-timesyncd)
```
phealy, line 112: I would suggest checking like this:

```shell
if (dpkg-query -W systemd-timesyncd); then
  systemctl_stop 20 30 120 systemd-timesyncd || exit $ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
  systemctl disable systemd-timesyncd || exit $ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT
else
  echo "systemd-timesyncd is removed, no need to disable"
fi
```
```

PR #3736, opened 2023-10-25
feat: disable background pull for artifact streaming and increase cache size 

ganeshkumar5699:
**What type of PR is this?**
/kind feature
/kind bug

Increasing cache size is primarily for very large container images and ML workloads, so overlaybd takes longer to garbage collect image chunks.

Due to a bug, background pull of images has issues. This changes the background pull value in the overlaybd config to false. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -348,6 +348,8 @@ ensureArtifactStreaming() {
   systemctl start acr-mirror.service
   sudo /opt/acr/tools/overlaybd/install.sh
   sudo /opt/acr/tools/overlaybd/enable-http-auth.sh
+  sudo /opt/acr/tools/overlaybd/config.sh download.enable false
+  sudo /opt/acr/tools/overlaybd/config.sh cacheConfig.cacheSizeGB 32
```
ganeshkumar5699, line 352: Even if storage space in node < 32 Gb, this would not error out, which is ok since this is the cache size after which there will be garbage collection of partial image layers.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -348,6 +348,8 @@ ensureArtifactStreaming() {
   systemctl start acr-mirror.service
   sudo /opt/acr/tools/overlaybd/install.sh
   sudo /opt/acr/tools/overlaybd/enable-http-auth.sh
+  sudo /opt/acr/tools/overlaybd/config.sh download.enable false
```
ganeshkumar5699, line 351: This is temporarily being set to false till the issue is fixed.

ganeshkumar5699:
This has the description of the overlaybd config: https://github.com/containerd/overlaybd

parts/linux/cloud-init/artifacts/cse_config.sh
ganeshkumar5699: This increase is primarily for large ML models

cameronmeissner (approved):

PR #3749, opened 2023-10-26
feat: set read_ahead_size for nfs mount

cvvz:
**What type of PR is this?**
/kind feature

We could set `read_aead_size` to a large value (default is 128) to improve the read performance of NFS.
Please refer to 
1. https://learn.microsoft.com/en-us/azure/azure-netapp-files/performance-linux-nfs-read-ahead#how-to-persistently-set-read-ahead-for-nfs-mounts
2. https://github.com/kubernetes-sigs/blob-csi-driver/pull/759

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
set `read_aead_size` to 15380
```


cvvz:
cc @andyzhangx

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -106,6 +106,13 @@ net.ipv6.conf.all.forwarding = 1
 net.bridge.bridge-nf-call-iptables = 1
 EOF
 
+echo "set read ahead size to 15380 KB"
```
andyzhangx, line 109: this is the Ubuntu image, what about Mariner support?

andyzhangx:
and pls add more comments in the PR description why we need to reset this read_ahead_size in vhd image, thanks.

vhdbuilder/packer/install-dependencies.sh
cvvz: I think that both Ubuntu and Mariner will execute this script, but I'm not 100% sure.

andyzhangx:
@ganeshkumar5699 could you take a look at this PR? thanks.

ganeshkumar5699:
A few questions: 
- How much is the performance improved by? 
- The link in the PR description mentions it's improved for NetApp NFS - are other systems impacted by this similarly?
- The linked doc also mentions Ubuntu 18.04 was the one impacted by this reduced performance. Is it the case for Ubuntu 22.04 and Ubuntu 20.04 as well?

andyzhangx:
> A few questions:
> 
> * How much is the performance improved by?
> * The link in the PR description mentions it's improved for NetApp NFS - are other systems impacted by this similarly?
> * The linked doc also mentions Ubuntu 18.04 was the one impacted by this reduced performance. Is it the case for Ubuntu 22.04 and Ubuntu 20.04 as well?

@ganeshkumar5699 with this read_ahead_size change, the nfs read perf could be improved by ~5x, and this is a common NFS setting (applied for all NFS mounts though there is already doc NetApp NFS, azure file team are also adding similar setting doc). And this setting applies for both 18.04, 22.04, etc.

andyzhangx (approved):

ganeshkumar5699:
PR lgtm. Would also check with @juan-lee for merge criteria at this time.

ganeshkumarashok (approved):
Lgtm, thanks for the patience

andyzhangx:
@cvvz could you fix the commit message lint error? thanks.

cvvz:
Replace this PR with https://github.com/Azure/AgentBaker/pull/3963. PATL @andyzhangx

PR #3777, opened 2023-11-03
feat(managed prometheus): update azure monitor metrics november release

bragi92:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


rashmichandrashekar (approved):

vishiy (approved):

anujmaheshwari1 (changes requested):
why are we adding 3 versions and removing just 1?
we have just come out of reducing our disk space, can we please retire older versions?

bragi92:
hey @anujmaheshwari1, with the previous version those images did not exist. We've now enabled a new feature where these images are also getting deployed to the AKS cluster and hence I'm only removing one version while adding those 3.

We want to keep the current release (that's going to start on Sunday) and the (current - 1) release version in the vhd and hence I'm not removing them as by the time we make a PR next month would be when most of the clusters would've received the update.

bragi92:
@anujmaheshwari1 Please have a look when you can. We've already merged in the changes in the AKS RP and the image has started rollout (its in staging right now).

rashmichandrashekar:
> @anujmaheshwari1 Please have a look when you can. We've already merged in the changes in the AKS RP and the image has started rollout (its in staging right now).

@anujmaheshwari1 - The new images being added also have a lesser size than the main image that existed previously(~785MB). For ex - these 2 images 6.8.1-main-11-03-2023-c061beb4-targetallocator and 6.8.1-main-11-03-2023-c061beb4-cfg have sizes of 151MB and 226MB which is lesser then the main image we have had previously. We have had 2 previous versions of the metrics image in the past accounting for rollout. Can we please allow this merge since the new images are not quite big and can be removed soon after the previous rollout completes?

anujmaheshwari1 (approved):

PR #3788, opened 2023-11-06
chore: cache new CNS

rbtr:
**What type of PR is this?**

/kind cleanup

Updates cached CNS version in 1.4 branch to 1.4.50

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -41,6 +41,7 @@
       "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v1.4.44.5",
+        "v1.4.50",
         "v1.5.11"
       ]
     },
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/components.json
```diff
@@ -41,6 +41,7 @@
       "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v1.4.44.5",
```
juan-lee, line 43: Can we remove the old one if it is superseded?

rbtr: @juan-lee the old one is still present in production, we can remove it after this new version rolls out

cameronmeissner: should be safe to remove since this is just what we cache on the VHD during build-time _moving forward,_ what's currently in production won't be affected

rbtr: but you could migrate to this VHD before getting the updated addon version, no? in that case all your nodes are going to need to pull the image from MCR at boot

cameronmeissner (approved):

PR #3794, opened 2023-11-07
feat: update cse script package changes for aks windows gpu feature

calvin197:
**What type of PR is this?**
kind feature

We need this PR to install nvidia gpu driver for Windows nodes.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,244 @@
...
+  
+function Get-Setup {
+    param(
+        [Parameter(Mandatory = $true)]
+        [PSCustomObject]$DriverUrlConfig
+    )
+
+    [OutputType([hashtable])]
+    
+    $GpuDriverURL = $DriverUrlConfig.GpuDriverURL
```
AbelHu, line 99: what is the endpoint of the url? We need to assure that they can be accessed in different clouds with custom's firewall settings.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,244 @@
...
+                Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_DOWNLOAD_FAILURE -ErrorMessage $ErrorMsg
+            }
+            else {
+                Start-Sleep -Seconds ($RetryCount * 2 + 1)
+                $RetryCount++
+            }
+        }
+    } while ($Loop)
+}
+function Get-UsingWebClient {
```
AbelHu, line 199: We use DownloadFileOverHttp so we can avoid download when the file is cached in the vhd.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,244 @@
+function Start-InstallGPUDriver {
+    param(
+        [Parameter(Mandatory = $true)]
+        [bool]$EnableInstall,
+        [Parameter(Mandatory = $true)]
+        [PSCustomObject]$DriverUrlConfig
+    )
+  
+    if (-not $EnableInstall) {
+        Write-ConsoleLog "ConfigGPUDriverIfNeeded is false. GPU driver installation skipped as per configuration."
```
AbelHu, line 10: I think that we need to investigate the log issue and then use wirte-log instead of Write-ConsoleLog

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: The url can be set differently for different clouds. Currently in the rp PR draft, they are all set to be the same urls. Those urls don't work on the private clouds but we have the flexibility to configure them.
RP PR: [https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/pullrequest/8878419?_a=files&path=/resourceprovider/server/microsoft.com/datamodel/model/aksengine_azenvtypes.go](https://github.com/Azure/AgentBaker/pull/url)

Chris and I are currently working on creating a pipeline to release the drivers to acs mirror. Then we will change the url on the rp side to make sure they work on private clouds.

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Write-Log uses Write-Output instead of Write-Host. It will write to the output. If i use Write-Output, the $RebootNeeded (return value captured from Start-InstallGPUDriver) will capture all the logs along with other output. It will cause 2 issues: 

1) The logs will not be printed unless we handle them specially.
Something like this:
`$RebootNeeded = Start-InstallGPUDriver -EnableInstall $global:ConfigGPUDriverIfNeeded -DriverUrlConfig $DriverUrlConfig | Tee-Object -Variable "LogMessages"
`
2) The return value of Start-InstallGPUDriver will be more than a bool, which caused strange behavior in my standalone test.  (i.e. Start-InstallGPUDriver return false. But the return value also includes log message. Hence, the $RebootNeeded is interpreted as true, which causes misbehaviors.)

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: But currently we don't cache the file. Are there plans in the roadmap to implement caching soon?

AbelHu: I think that we should cache all packages in the disk since it will avoid download failure and have better perf

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Since Write-Log is used frequently in many helper functions, to avoid the same issue in the future, I will make code changes to accommodate the use of Write-Log.

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Given that the combined size of the two drivers is approximately 1 GB, would it be acceptable to cache them in all VHDs?

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+    }
+    catch {
+        $ErrorMsg = "Failed to query the SKU information."
+        Write-Log $ErrorMsg
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_SKU_INFO_NOT_FOUND -ErrorMessage $ErrorMsg
+    }
+      
+    $Setup = @{
+        RebootNeeded = ($vmSize -ne $null -and $Compute.vmSize -match "_NV")
+        Target       = "$RootDir\..\$fileName"
```
AbelHu, line 120: Should we delete this file?

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+    }
+  
+    Write-Log "ConfigGPUDriverIfNeeded is true. GPU driver installation started as per configuration."
+  
+    $RootDir = "C:\AzureData\Windows"
+  
+    try {
+        $FatalError = @()
+  
+        $LogFolder = "$RootDir\.."
```
AbelHu, line 21: Please also include the changes for log collection in this PR and validate whether the changes work as expected and it does not break current log collection.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+  
+        # Get the SetupTarget based on the input
+        $Setup = Get-Setup -DriverConfig $DriverConfig
+        $SetupTarget = $Setup.Target
+        Write-Log "Setup complete"
+        $IsSignatureValid = VerifySignature $SetupTarget 
+        if ($IsSignatureValid -eq $false) {
+            $ErrorMsg = "Signature embedded in $($SetupTarget) is not valid."
+            Write-Log $ErrorMsg
+            Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED -ErrorMessage $ErrorMsg
```
AbelHu, line 33: Please use different error code for different failures so we can know the failure reason without cse logs.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+        catch [System.TimeoutException] {
+            $ErrorMsg = "Timeout $Timeout s exceeded. Stopping the installation process. Reboot for another attempt."
+            Write-Log $ErrorMsg
+            Stop-Process -InputObject $p -Force
+            Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_TIMEOUT -ErrorMessage $ErrorMsg
+        }
+        catch {
+            $ErrorMsg = "Exception: $($_.ToString())"
+            Write-Log $ErrorMsg # the status file may get over-written when the agent re-attempts this step
+            Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED -ErrorMessage $ErrorMsg
```
AbelHu, line 75: Use a different error code for the exception

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+        }
+      
+    }
+    catch {
+        $FatalError += $_
+        $errorCount = $FatalError.Count
```
AbelHu, line 81: Suggest to add ut to cover these lines. Otherwise it may not work as expected when the code is executed.

staging/cse/windows/nvidiagpudriverfunc.ps1
AbelHu, line 120: $RootDir is defined in Start-InstallGPUDriver. Recommend to pass the path as a parameter.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+        $ErrorMsg = "A fatal error occurred. Number of errors: $errorCount. Error details: $($_ | Out-String)"
+        Write-Log $ErrorMsg
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED -ErrorMessage $ErrorMsg
+    }
+}
+  
+function Get-Setup {
```
AbelHu, line 88: Get-Setup is a little confused. It contains two parts: construct the setup object and download the driver. `Prepare-Installation` or something else may be better.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+        Target       = "$RootDir\..\$fileName"
+    }
+
+    Write-Log "Downloading from $GpuDriverURL to $($Setup.Target)"
+    DownloadFileOverHttp -Url $GpuDriverURL -DestinationPath $Setup.Target -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_DOWNLOAD_FAILURE
+  
+    return $Setup
+}
+  
+function Get-VmData {
```
AbelHu, line 129: We have a function `Retry-Command`. You can reference the existing code, for example

    try {
        $response = Retry-Command -Command "Invoke-RestMethod" -Args @{Uri=$uri; Method="Get"; ContentType="application/json"; Headers=$headers} -Retries 5 -RetryDelaySeconds 10
    } catch {
        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GET_SUBNET_PREFIX -ErrorMessage "Error getting subnet prefix. Error: $_"
    }

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+                Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED -ErrorMessage $ErrorMsg
+            }
+  
+            if ($Setup.RebootNeeded -or $p.ExitCode -eq 1) {
+                Write-Log "Reboot is needed for this GPU Driver..."
+                $global:RebootNeeded = $true
+            }
+        }
+        catch [System.TimeoutException] {
+            $ErrorMsg = "Timeout $Timeout s exceeded. Stopping the installation process. Reboot for another attempt."
```
AbelHu, line 67: s => seconds?

AbelHu: Windows cannot return the error message to the users so `Reboot for another attempt` does not work

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+        $vmSize = $Compute.vmSize
+    }
+    catch {
+        $ErrorMsg = "Failed to query the SKU information."
+        Write-Log $ErrorMsg
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_SKU_INFO_NOT_FOUND -ErrorMessage $ErrorMsg
+    }
+      
+    $Setup = @{
+        RebootNeeded = ($vmSize -ne $null -and $Compute.vmSize -match "_NV")
```
AbelHu, line 119: "XXX_nvfads" also matches "_NV"

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
...
+  
+        $LogFolder = "$RootDir\.."
+  
+        Write-Log "Attempting to install Nvidia driver..."
+  
+        # Get the SetupTarget based on the input
+        $Setup = Get-Setup -DriverConfig $DriverConfig
+        $SetupTarget = $Setup.Target
+        Write-Log "Setup complete"
+        $IsSignatureValid = VerifySignature $SetupTarget
```
AbelHu, line 29: It seems like that the target is MSI or exe. Will the final official package be same as the package used for testing now?

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,178 @@
+function Start-InstallGPUDriver {
+    param(
+        [Parameter(Mandatory = $true)]
+        [bool]$EnableInstall,
+        [Parameter(Mandatory = $true)]
+        [PSCustomObject]$DriverConfig
```
AbelHu, line 6: Can we use string URL instead of object as the parameter? It will help us to reduce the cse length in https://github.com/Azure/AgentBaker/pull/3820/files#diff-fdbab787063757b4c74fc431a241ee84d3f02448e4f60b4b22f731c917832fbb

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: renamed Get-Setup to Prepare-Installation

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: added Remove-InstallerFile to remove installer after successful installation

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: refactored to use existing Retry-Command

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Refactored to use Prepare-Installation

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: added unique error code

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: added unique error code

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Removed this catch block

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Is it a typs of VM that has GPU? If not, rp will block the request. We can only reach this line when the vm size is either NV, ND, or NC series VMs.

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: Yes, they will be the same package.

staging/cse/windows/provisioningscripts/loggenerator.ps1
```diff
@@ -87,6 +87,15 @@ $kLogFiles | Foreach-Object {
     Create-SymbolLinkFile -SrcFile (Join-Path "C:\k\" $_) -DestFile (Join-Path $aksLogFolder $_)
 }
 
+$nvidiaInstallLogFolder="C:\AzureData\NvidiaInstallLog"
+if (Test-Path $nvidiaInstallLogFolder) {
+    $logFiles=Get-ChildItem (Join-Path $nvidiaInstallLogFolder *.log)
+    $logFiles | Foreach-Object {
+        $fileName = [IO.Path]::GetFileName($_)
+        Create-SymbolLinkFile -SrcFile (Join-Path $nvidiaInstallLogFolder $fileName) -DestFile (Join-Path $aksLogFolder $fileName)
```
AbelHu, line 95: Does the gpu install log file have a unique prefix so there is no conflict when creating symbol links in $aksLogFolder?

staging/cse/windows/debug/collect-windows-logs.ps1
```diff
@@ -309,6 +309,12 @@ if ($tempFile -ne "") {
   $paths += $tempFile
 }
 
+#Collect Nvidia log files
+$tempFile=(CollectLogsFromDirectory -path "C:\AzureData\NvidiaInstallLog" -targetFileName "NvidiaInstallLog-$($timeStamp).zip")
```
AbelHu, line 313: Are those gpu log files not used when running this script? Compress-Archive will fail if the log files are being written.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,169 @@
...
+    if (-not $EnableInstall) {
+        Write-Log "ConfigGPUDriverIfNeeded is false. GPU driver installation skipped as per configuration."
+        return
+    }
+  
+    Write-Log "ConfigGPUDriverIfNeeded is true. GPU driver installation started as per configuration."
+  
+    if ([string]::IsNullOrEmpty($GpuDriverURL)) {
+        $ErrorMsg = "DriverURL is not properly specified."
+        Write-Log $ErrorMsg
```
AbelHu, line 19: Remove logging before calling Set-ExitCode since Set-ExitCode will always log the error message before exit with the ExitCode.

AbelHu: NOTE: Same in other lines.

parts/windows/windowscsehelper.ps1
```diff
@@ -54,6 +54,16 @@ $global:WINDOWS_CSE_ERROR_NO_CUSTOM_DATA_BIN=49 # Return this error code in csec
 $global:WINDOWS_CSE_ERROR_NO_CSE_RESULT_LOG=50 # Return this error code in csecmd.ps1 when C:\AzureData\CSEResult.log does not exist
 $global:WINDOWS_CSE_ERROR_COPY_LOG_COLLECTION_SCRIPTS=51
 $global:WINDOWS_CSE_ERROR_RESIZE_OS_DRIVE=52
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED=53
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_TIMEOUT=54
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_VM_SIZE_NOT_SUPPORTED=55
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_URL_NOT_SET=56
+$global:WINDOWS_CSE_ERROR_SKU_INFO_NOT_FOUND=57
```
AbelHu, line 61: Please update it to `$global:WINDOWS_CSE_ERROR_GPU_SKU_INFO_NOT_FOUND=57` so it is clear to indicate the error is thrown by gpu script.

parts/windows/windowscsehelper.ps1
```diff
@@ -54,6 +54,16 @@ $global:WINDOWS_CSE_ERROR_NO_CUSTOM_DATA_BIN=49 # Return this error code in csec
 $global:WINDOWS_CSE_ERROR_NO_CSE_RESULT_LOG=50 # Return this error code in csecmd.ps1 when C:\AzureData\CSEResult.log does not exist
 $global:WINDOWS_CSE_ERROR_COPY_LOG_COLLECTION_SCRIPTS=51
 $global:WINDOWS_CSE_ERROR_RESIZE_OS_DRIVE=52
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED=53
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_TIMEOUT=54
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_VM_SIZE_NOT_SUPPORTED=55
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_URL_NOT_SET=56
+$global:WINDOWS_CSE_ERROR_SKU_INFO_NOT_FOUND=57
+$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_DOWNLOAD_FAILURE=58
+$global:WINDOWS_CSE_ERROR_INVALID_SIGNATURE=59
```
AbelHu, line 63: Please update it to `$global:WINDOWS_CSE_ERROR_GPU_DRIVER_INVALID_SIGNATURE=59` so it is clear to indicate the error is thrown by gpu script.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,169 @@
...
+            Remove-InstallerFile -InstallerPath $Target
+        } 
+        else {
+            $ErrorMsg = "GPU Driver Installation Failed! Code: $($p.ExitCode)"
+            Write-Log $ErrorMsg
+            Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_FAILED -ErrorMessage $ErrorMsg
+            return
+        }
+
+        # check if reboot is needed
```
AbelHu, line 79: Suggest to pass $p.ExitCode to the function Handle-InstallationResult and only set $global:RebootNeeded inside Handle-InstallationResult .

The logical should be more clear.
Fail if $gpuInstallationExitCode is neither 0 nor 1
Reboot if $p.ExitCode is 1
Reeboot if $p.ExitCode is 0 and vmsize is  NVseries

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,169 @@
...
+        $Compute = Get-VmData
+        $vmSize = $Compute.vmSize
+    }
+    catch {
+        $ErrorMsg = "Failed to query the SKU information."
+        Write-Log $ErrorMsg
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_SKU_INFO_NOT_FOUND -ErrorMessage $ErrorMsg
+    }
+
+    $IsNVSeries = $vmSize -ne $null -and $vmSize -match "_NV"
```
AbelHu, line 107: https://learn.microsoft.com/en-us/azure/virtual-machines/nv-series
https://learn.microsoft.com/en-us/azure/virtual-machines/nvv3-series
Could we use `Standard_NV` instead of `_NV`?

staging/cse/windows/provisioningscripts/loggenerator.ps1
calvin197: Yes, they all have the prefix "LOG."
image

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,165 @@
...
+        Write-Log "ConfigGPUDriverIfNeeded is false. GPU driver installation skipped as per configuration."
+        return
+    }
+  
+    Write-Log "ConfigGPUDriverIfNeeded is true. GPU driver installation started as per configuration."
+  
+    if ([string]::IsNullOrEmpty($GpuDriverURL)) {
+        $ErrorMsg = "DriverURL is not properly specified."
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_URL_NOT_SET -ErrorMessage $ErrorMsg
+        return
```
AbelHu, line 20: Set-ExitCode will throw an exception so you do not need to add `return`

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function Handle-InstallationResult {
+    param(
+        [Parameter(Mandatory = $true)]
+        [int]$ErrorCode
+    )
+
+    # check if vm size is nv series. if so, set RebootNeeded to be true
+    try {
```
AbelHu, line 80: I think that we need to check $ErrorCode at first. If $ErrorCode is neither 0 nor 1, we should exit; If $ErrorCode is 1, we just need to set $global:RebootNeeded to $true. If $ErrorCode is 0, get vm size and then take action.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,165 @@
...
+        $p = Start-Process -FilePath $Target -ArgumentList $Arguments -PassThru
+        
+        $Timeout = 10 * 60 # in seconds. 10 minutes for timeout of the installation
+
+        # This is for testability. Start-Process mock returns a hashtable.
+        if (-not ($p -is [hashtable])) {
+            Wait-Process -InputObject $p -Timeout $Timeout -ErrorAction Stop
+        }
+
+        $ExitCode = $p.ExitCode
```
AbelHu, line 63: Handle-InstallationResult -ErrorCode $p.ExitCode

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,165 @@
...
+  
+    if (-not $EnableInstall) {
+        Write-Log "ConfigGPUDriverIfNeeded is false. GPU driver installation skipped as per configuration."
+        return
+    }
+  
+    Write-Log "ConfigGPUDriverIfNeeded is true. GPU driver installation started as per configuration."
+  
+    if ([string]::IsNullOrEmpty($GpuDriverURL)) {
+        $ErrorMsg = "DriverURL is not properly specified."
```
AbelHu, line 18: Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_URL_NOT_SET -ErrorMessage "DriverURL is not properly specified."
One line should be enough

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: The Mock Set-ExitCode looks like this: 

```
        Mock Set-ExitCode -MockWith {
            Param(
                $ExitCode,
                $ErrorMessage
            )
            Write-Host "Set-ExitCode $ExitCode $ErrorMessage"
        } -Verifiable
```
        
If I don't add return after Set-ExitCode, the code won't stop running during UT.

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: refactored to do early return and only get vm data when it's needed

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: refactored to make them one liner

staging/cse/windows/nvidiagpudriverfunc.ps1
calvin197: added throw in the mock function. removed return after Set-ExitCode.

staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,154 @@
...
+    Write-Log "Downloading from $GpuDriverURL to $Target"
+    DownloadFileOverHttp -Url $GpuDriverURL -DestinationPath $Target -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_DOWNLOAD_FAILURE
+    Write-Log "Installer download complete"
+    
+    $IsSignatureValid = VerifySignature $Target 
+    if ($IsSignatureValid -eq $false) {
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INVALID_SIGNATURE -ErrorMessage "Signature embedded in $($Target) is not valid."
+    }
+    else {
+        Write-Log "Signature embedded in $($Target) is valid."
```
AbelHu, line 40: Duplicate logs since line 134 has logged it.

calvin197: I just realized $IsSignatureValid will also capture VerifySignature's Write-Output content. And its value will never be false. To fix the issue, I will move the Set-ExitCode inside the VerifySignature.

AbelHu (approved):
LGTM. Thanks for the PR.

PR #3796, opened 2023-11-07
refactor: always set AZURE_ENVIRONMENT_FILEPATH on windows when custo…

cameronmeissner:
**What type of PR is this?**

Always set AZURE_ENVIRONMENT_FILEPATH on windows node when custom cloud is configured for AGC.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -81,9 +81,6 @@ if ($global:NetworkPlugin -eq "azure") {
     # Restart Kubeproxy, which would wait, until the network is created
     # This was fixed in 1.15, workaround still needed for 1.14 https://github.com/kubernetes/kubernetes/pull/78612
     Restart-Service Kubeproxy -Force
-
-    # Set env file for Azure Stack
-    $env:AZURE_ENVIRONMENT_FILEPATH = "c:\k\azurestackcloud.json"
```
AbelHu, line 86: If this is removed, kubelet will not be able to get endpoints for custom clouds.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -332,6 +332,7 @@ try
     # In case of AKSCustomCloud, customer cloud env will be loaded from azurestackcloud.json 
     {{if IsAKSCustomCloud}}
     $azureStackConfigFile = [io.path]::Combine($global:KubeDir, "azurestackcloud.json")
+    $env:AZURE_ENVIRONMENT_FILEPATH = $azureStackConfigFile
```
AbelHu, line 335: Which components will use this environment variable in running AKS Windows CSE script in node provisioning?

cameronmeissner: gotcha, I guess these are separate envs then

cameronmeissner: we want to make sure this set after CSE finishes so the new secure TLS bootstrap client-go credential plugin will be able to use it in order to request a TLS bootstrap token to join the cluster. The main idea is that instead of using a hard-coded TLS bootstrap token, we'll instead point kubelet towards this new credential plugin (which is just a go binary) via bootstrap-kubeconfig in order to generate a dynamic and short-lived TLS bootstrap token instead

staging/cse/windows/provisioningscripts/kubeletstart.ps1
cameronmeissner: reverted this change

AbelHu: @cameronmeissner If you plan to support custom clouds with other networkplugin, maybe you need to move `$env:AZURE_ENVIRONMENT_FILEPATH = "c:\k\azurestackcloud.json"` to outside of if.

parts/windows/kuberneteswindowssetup.ps1
AbelHu: I think that $env:AZURE_ENVIRONMENT_FILEPATH only can be used in current powershell session used by CSE. Please test whether it works as expected in the new secure TLS bootstrap client-go credential plugin

AbelHu: Setting Environment Variable Persistently
To set the environment persistently so they should remain even when close the session, PowerShell uses [System.Environment] class with the SetEnvironmentVariable method for the environment variable to set it persistently.

[System.Environment]::SetEnvironmentVariable('ResourceGroup','AZ_Resource_Group')
PS C:\> $env:ResourceGroup
AZ_Resource_Group

staging/cse/windows/provisioningscripts/kubeletstart.ps1
AbelHu: @cameronmeissner I think that you can remove it if a persistent environment variable `AZURE_ENVIRONMENT_FILEPATH` exists

cameronmeissner: > @cameronmeissner I think that you can remove it if a persistent environment variable `AZURE_ENVIRONMENT_FILEPATH` exists

right, this is what I was thinking - I figured we could just always set it irregardless of network plugin, that way whenever a particular plugin needs it it'll just be there, as well as for the secure bootstrap client

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -332,6 +332,7 @@ try
     # In case of AKSCustomCloud, customer cloud env will be loaded from azurestackcloud.json 
     {{if IsAKSCustomCloud}}
     $azureStackConfigFile = [io.path]::Combine($global:KubeDir, "azurestackcloud.json")
+    [Environment]::SetEnvironmentVariable("AZURE_ENVIRONMENT_FILEPATH", $azureStackConfigFile, [EnvironmentVariableTarget]::Machine)
```
AbelHu, line 335: Could we run some test in custom cloud to validate this test?

PR #3799, opened 2023-11-07
feat: optimize resize-osdrive to improve node provisioning time

thecloudtaylor:
**What type of PR is this?**
/kind feature

Resize-OSDrive can be slow due to Get-PartitionSupportedSize, replacing that call with diskpart

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

While inspecting node provision time – Resize-OSDrive was taking over 2min in some cases, the Windows Storage team informed us that Get-PartitionSupportedSize can be a slow call as it not only gets the maximum size but also the minimum size which requires a call into defrag. This change replaces that call and uses diskpart for the resize operation.

Unit tested against 2019 and 2022, generation 1 and generation 2, across 7 sku sizes. Specific unit test used a CSE script of just the old/new resize function against the 30GB (smalldisk) OS version. All versions/sku's passed. I did not test integrated with the AKS CSE scripts (not sure how to do that). This unit test only observes small improvements due to the nature of the base image but proves the core functionality is stable.

@AbelHu also collected some data observing a significant improvement. For WS2019 an avg of 89.2 sec down to 8.85 sec, and WS2022 an avg of 79.4 sec down to 8.6 sec, and WS2022 Gen 2 an avg of 122.8 sec down to 13.4 sec.

**Release note**:
```
none
```

### DIFF
--- a/staging/cse/windows/configfunc.ps1
+++ b/staging/cse/windows/configfunc.ps1
@@ -11,15 +11,21 @@ function Set-TelemetrySetting
 }
 
 # Resize the system partition to the max available size. Azure can resize a managed disk, but the VM still needs to extend the partition boundary
+# This approach was recommended by the Windows Storage team to avoid performance delay when calling Get-PartitionSupportedSize
 function Resize-OSDrive
 {
     try {
         $osDrive = ((Get-WmiObject Win32_OperatingSystem -ErrorAction Stop).SystemDrive).TrimEnd(":")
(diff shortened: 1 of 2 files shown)


thecloudtaylor:
will resubmit clean.

thecloudtaylor:
replaced with https://github.com/Azure/AgentBaker/pull/3804

PR #3805, opened 2023-11-08
chore: update CNS, CNI, and dropgz

rbtr:
**What type of PR is this?**
/kind cleanup

adds the latest CNS, CNI, and dropgz (of 1.4 family)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -40,18 +40,16 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cns:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.4.44.5",
-        "v1.4.50",
-        "v1.5.11"
+        "v1.4.51",
+        "v1.5.15"
       ]
     },
     {
       "downloadURL": "mcr.microsoft.com/containernetworking/cni-dropgz:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v0.0.4.1",
-        "v0.0.11",
-        "v0.0.15"
+        "v0.0.4.2",
+        "v0.1.1"
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/components.json
```diff
@@ -40,16 +40,18 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cns:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.4.44.5",
         "v1.4.50",
-        "v1.5.11"
+        "v1.4.51",
```
paulgmiller, line 44: we're dropping 1.4.50 because 51 is already mostly shoved out i assume?

But keeping 1.5.11 because .14 is still baking?

paulgmiller (approved):

vhdbuilder/packer/components.json
rbtr: dropping 1.4.44.5 because 1.4.50 should be mostly out by the time this goes out, and 1.4.51 is vNext.
keeping 1.5.11 as it's the current release on 1.28+/dualstack, we'll bring 1.5.14 as vNext after freeze

cameronmeissner:
seems to be blocked by an issue this introduces in VHD builds

rbtr:
I think an artifact updated here wasn't published correctly on my end, working on it

vhdbuilder/packer/components.json
juan-lee: Can we only have the latest versions of 1.4.x and 1.5.x?

juan-lee (approved):

vhdbuilder/packer/components.json
```diff
@@ -314,17 +312,21 @@
...
     {
       "fileName": "azure-vnet-cni-swift-linux-amd64-v*",
       "downloadLocation": "/opt/cni/downloads",
       "downloadURL": "https://acs-mirror.azureedge.net/azure-cni/v*/binaries",
       "versions": [
+        "1.4.43.1",
+        "1.4.51",
         "1.5.11",
-        "1.4.43.1"
+        "1.5.14"
```
juan-lee, line 329: should we just have 1.5.14 and 1.4.51?

UtheMan (approved):

ganeshkumar5699 (approved):

rbtr:
seems like the ARM check is broken? I don't see anything in the log that looks related to this change?

PR #3820, opened 2023-11-10
feat: use nvidia gpu driver install script in windows cse

calvin197:
**What type of PR is this?**
kind feature

We need this PR to for the AKS Windows GPU Feature

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


staging/cse/windows/nvidiagpudriverfunc.ps1
```diff
@@ -0,0 +1,161 @@
...
+
+function Remove-InstallerFile {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$InstallerPath
+    )
+  
+    Write-Log "Attempting to remove installer file at $InstallerPath..."
+  
+    try {
```
calvin197, line 154: To do: No need to try catch

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -95,6 +98,16 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_DISTRO=${{ parameters.windowsDistro }}
         export WINDOWS_NODE_IMAGE_VERSION=${{ parameters.windowsNodeImageVersion }}
+        export CONFIG_GPU_DRIVER_IF_NEEDED=false
+        export GPU_SUFFIX=""
+        if [[ "${{ parameters.windowsGpuDriverSuffix }}" == "-grid" ]]; then
+          export WINDOWS_GPU_DRIVER_URL="https://download.microsoft.com/download/9/d/6/9d6f3611-ff0c-43bc-8958-fe7fb0ded78d/537.13_grid_win10_win11_server2019_server2022_dch_64bit_international_azure_swl.exe"
```
chengliangli0918, line 104: use the driver binaries uploaded to https://acs-mirror.azureedge.net/aks/windows/gpudrivers/grid

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -95,6 +98,16 @@ jobs:
...
+          export CONFIG_GPU_DRIVER_IF_NEEDED=true
+        fi
+        if [[ "${{ parameters.windowsGpuDriverSuffix }}" == "-cuda" ]]; then
+          export WINDOWS_GPU_DRIVER_URL="https://download.microsoft.com/download/a/c/a/aca6122c-3565-4f55-9d4d-56a1cecc6ba6/473.47-data-center-tesla-desktop-winserver-2019-2016-international.exe"
```
chengliangli0918, line 108: use the driver binaries uploaded to https://acs-mirror.azureedge.net/aks/windows/gpudrivers/cuda

.pipelines/e2e-windows.yaml
```diff
@@ -72,4 +72,32 @@ stages:
...
+
+- stage: test_2022_containerd_gen2_gpu_grid
+  dependsOn: []
+  condition: eq('${{ parameters.test2022containerdgen2 }}', True)
+  jobs:
+  - template: ./templates/e2e-windows-template.yaml
+    parameters:
+      windowsImage: 2022-containerd-gen2
+      windowsOSSKU: Windows2022
+      windowsVMSize: Standard_NV6
```
chengliangli0918, line 85: Standard_NV6 is V1 generation vm_size, it cannot be used together with gen2 distro aks-windows-2022-containerd-gen2

.pipelines/e2e-windows.yaml
```diff
@@ -72,4 +72,32 @@ stages:
...
+
+- stage: test_2022_containerd_gen2_gpu_cuda
+  dependsOn: []
+  condition: eq('${{ parameters.test2022containerdgen2 }}', True)
+  jobs:
+  - template: ./templates/e2e-windows-template.yaml
+    parameters:
+      windowsImage: 2022-containerd-gen2
+      windowsOSSKU: Windows2022
+      windowsVMSize: Standard_nc6_promo
```
chengliangli0918, line 99: Standard_NC6_prom is V1 generation vm_size, it cannot be used together with gen2 distro aks-windows-2022-containerd-gen2

chengliangli0918: use Standard_NC6s_v3 instead

.pipelines/e2e-windows.yaml
chengliangli0918: use Standard_NV6ads_A10_v5 instead.

AbelHu:
Please add tags `windows` `VHD`

.pipelines/e2e-windows.yaml
```diff
@@ -72,4 +80,32 @@ stages:
...
+  dependsOn: []
+  condition: eq('${{ parameters.test2022containerdgen2gpugrid }}', True)
+  jobs:
+  - template: ./templates/e2e-windows-template.yaml
+    parameters:
+      windowsImage: 2022-containerd-gen2
+      windowsOSSKU: Windows2022
+      windowsVMSize: Standard_NV6ads_A10_v5
+      windowsDistro: aks-windows-2022-containerd-gen2
+      windowsNodeImageVersion: AKSWindows-2022-containerd-gen2-test-2023.02.07
```
AbelHu, line 95: could we use latest instead of a specified version to avoid updating it from time to time?

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -14,6 +14,9 @@ parameters:
   - name: windowsNodeImageVersion
     type: string
     default: AKSWindows-2019-containerd-test-2023.02.07
+  - name: windowsGpuDriverSuffix
+    type: string
+    default:
```
AbelHu, line 19: Is default empty?

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -95,6 +99,17 @@ jobs:
         export WINDOWS_E2E_IMAGE=${{ parameters.windowsImage }}
         export WINDOWS_DISTRO=${{ parameters.windowsDistro }}
         export WINDOWS_NODE_IMAGE_VERSION=${{ parameters.windowsNodeImageVersion }}
+        export WINDOWS_GPU_DRIVER_SUFFIX=${{ parameters.windowsGpuDriverSuffix }}
+        if [[ "${WINDOWS_GPU_DRIVER_SUFFIX}" == "-grid" ]]; then
+            export WINDOWS_GPU_DRIVER_URL="https://acs-mirror.azureedge.net/aks/windows/gpudrivers/grid/537.13_grid_win10_win11_server2019_server2022_dch_64bit_international_azure_swl.exe"
```
AbelHu, line 104: Can we update the driver versions from a configuration file instead of this pipeline?

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -170,13 +170,19 @@ $global:EnableHostsConfigAgent = [System.Convert]::ToBoolean("{{ EnableHostsConf
 # These scripts are used by cse
 $global:CSEScriptsPackageUrl = "{{GetVariable "windowsCSEScriptsPackageURL" }}";
 
+# The windows nvidia gpu driver related urls are used by windows cse
```
AbelHu, line 173: urls => url?

staging/cse/windows/README
```diff
@@ -55,6 +55,9 @@ popd
 # AKS Windows CSE Scripts Package
 All files except *.test.ps1 and README will be published in AKS Windows CSE Scripts Package.
 
+## v0.0.35
+feat: update cse script package changes for aks windows gpu feature #3794
```
AbelHu, line 59: v0.0.35 should also include this commit https://github.com/Azure/AgentBaker/commit/1c875831b2c7119c233510e8e60e374ad4d74bd2

`feat: optimize resize-osdrive replacing get-partitionsupportedsize with diskpart #3804`

AbelHu: With checking the history of the folder `staging/cse/windows`
https://github.com/Azure/AgentBaker/commits/master/staging/cse/windows

.pipelines/e2e-windows.yaml
```diff
@@ -72,4 +80,32 @@ stages:
...
+
+- stage: test_2022_containerd_gen2_gpu_grid
+  dependsOn: []
+  condition: eq('${{ parameters.test2022containerdgen2gpugrid }}', True)
+  jobs:
+  - template: ./templates/e2e-windows-template.yaml
+    parameters:
+      windowsImage: 2022-containerd-gen2
+      windowsOSSKU: Windows2022
+      windowsVMSize: Standard_NV6ads_A10_v5
```
AbelHu, line 93: It may be better to set vmsize via variables since we may hit the quota issue.

.pipelines/e2e-windows.yaml
calvin197: I am following the existing pattern for non-GPU Windows E2E. Maybe @junjiezhang1997 can chime in to see how we can achieve that, perhaps in a separate PR?

.pipelines/e2e-windows.yaml
calvin197: I agree it could hit the quota issue. Is there an existing pattern that we're choosing available vm size dynamically in e2e testing?

.pipelines/templates/e2e-windows-template.yaml
calvin197: Can I create two params in e2e-windows-template.yaml, then set it and pass it from e2e-windows.yaml?

calvin197:
> Please add tags `windows` `VHD`

calvin197:
> Please add tags `windows` `VHD`

I am not sure what kind of tags you are referring to. Could you please clarify?

.pipelines/e2e-windows.yaml
junjiezhang1997: This variable is not used to deploy test vmss now. Actually, we obtain the latest image to deploy test vmss in https://github.com/Azure/AgentBaker/blob/master/e2e/windows/e2e-scenario.sh#L143.

.pipelines/templates/e2e-windows-template.yaml
AbelHu: It seems like that this is required and we set it in e2e-windows.yaml so I think that we should not set a default value (empty).

.pipelines/templates/e2e-windows-template.yaml
calvin197: Or do you mean that we can create a centralized template file to update the driver urls in only one place? Something like https://github.com/Azure/azhpc-extensions/blob/master/NvidiaGPU/resources.json used by the vm extension?

.pipelines/templates/e2e-windows-template.yaml
calvin197: It is not required for non-gpu scenario to set a value in e2e-windows.yaml. We don't want to force non-gpu scenario to set a value for gpu related stuffs in e2e-windows.yaml so I set a default value of an empty string. The empty string will be passed down so that we don't append a suffix to those non-gpu scenarios artifacts. 

For example, line 71 in e2e-windows-template.yaml:
      `artifact: ${{ parameters.windowsImage }}${{ parameters.windowsGpuDriverSuffix }}-clusterConfig`

.pipelines/e2e-windows.yaml
junjiezhang1997: The variable windowsNodeImageVersion is not used to deploy test vmss now. We use https://github.com/Azure/AgentBaker/blob/master/e2e/windows/e2e-scenario.sh#L143 to obtain the latest image to deploy test vmss. So I think gpu case should already use latest image now. We can check it in Azure Portal when running the test.

e2e/windows/e2e-starter.sh
```diff
@@ -8,7 +8,7 @@ log "Starting e2e tests"
 
 # Create a resource group for the cluster
 log "Creating resource group"
-E2E_RESOURCE_GROUP_NAME="$AZURE_E2E_RESOURCE_GROUP_NAME-$WINDOWS_E2E_IMAGE-$K8S_VERSION"
+E2E_RESOURCE_GROUP_NAME="$AZURE_E2E_RESOURCE_GROUP_NAME-$WINDOWS_E2E_IMAGE$WINDOWS_GPU_DRIVER_SUFFIX-$K8S_VERSION"
```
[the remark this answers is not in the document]
calvin197: Just added the suffix to the cse logs filename. 

todo: check the new test run artifact name to verify the change.

.pipelines/e2e-windows.yaml
calvin197: I just verified with @junjiezhang1997 that the test vm is indeed using the latest image.

e2e/windows/e2e-starter.sh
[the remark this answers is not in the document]
calvin197: verified the gpu log name is now appended with the gpu suffix, and it also does not affect the non gpu log filename.

.pipelines/e2e-windows.yaml
junjiezhang1997: This variable is not used by test vm anymore. Create a task to remove it later.

.pipelines/templates/e2e-windows-template.yaml
AbelHu: I think that we should define the gpu drivers in agentbaker to avoid failures caused by external changes.

.pipelines/e2e-windows.yaml
calvin197: added pipeline params to choose available vm size. 
todo: remove e2e resource and rerun gpu scenarios to ensure the vm size is. correctly populated from the pipeline params.

.pipelines/templates/e2e-windows-template.yaml
junjiezhang1997: I think the main concern here is how to align the driver versions for test with the driver versions in prod environment. We should avoid to update the urls manually every time the new driver versions are released.

AbelHu (approved):
LGTM. Thanks for the PR. Please address those unresolved comments in next PR.

.pipelines/templates/e2e-windows-template.yaml
calvin197: Just added a task to track this issue. This issue will be resolved later. 

https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/25951813

.pipelines/e2e-windows.yaml
junjiezhang1997: Seems like you haven't use the pipeline params in this yaml file.

PR #3826, opened 2023-11-14
chore: add basic skeleton for contract parser

anujmaheshwari1:
**What type of PR is this?**
This PR adds a simple skeleton of what the Go binary on a self-contained VHD will look like.
It will take an input which will be a serialized JSON, deserialize it to contract struct, and generate the final bootstrap trigger script. 
Input, deserialization and templating to follow in coming PRs.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/parser/parser.go
```diff
@@ -0,0 +1,33 @@
...
+	"bytes"
+	_ "embed"
+	"log"
+	"text/template"
+)
+
+var (
+	//go:embed cse_cmd.sh.gtpl
+	bootstrapTrigger         string
+	bootstrapTriggerTemplate = template.Must(template.New("triggerBootstrapScript").Parse(bootstrapTrigger)) //nolint:gochecknoglobals
```
anujmaheshwari1, line 13: the only way to use go:embed, as Karpenter does https://github.com/Azure/karpenter/blob/057ac97ff34fe57195e00ea9e8cb70148656b447/pkg/providers/imagefamily/bootstrap/aksbootstrap.go#L225 is to use global variables.

from official Golang docs:

`the variable declared with go:embed must be a package-level variable. The directive allows you to embed files or directories into your Go program during the build process, and this embedding happens at the package level.`

if there is unanimous pushback on suppressing the golint, we can go back to using the more verbose method. can also revisit this when this is closer to production.

pkg/parser/parser.go
```diff
@@ -0,0 +1,33 @@
...
+func executeBootstrapTemplate() (string, error) {
+	var buffer bytes.Buffer
+	if err := bootstrapTriggerTemplate.Execute(&buffer, nil); err != nil {
+		return "", err
+	}
+	return buffer.String(), nil
+}
+
+// this function will eventually take a pointer to the bootstrap contract struct.
+// it will then emplate out the variables into the final bootstrap trigger script.
```
cameronmeissner, line 25: typo in `emplate`

pkg/parser/parser.go
cameronmeissner, line 13:
```suggestion
       // nolint:gochecknoglobals
	bootstrapTriggerTemplate = template.Must(template.New("triggerBootstrapScript").Parse(bootstrapTrigger))
```

cameronmeissner (approved):

pkg/parser/parser.go
anujmaheshwari1: oops thanks, fixed!

Devinwong (approved):

Issue #3810, opened 2023-11-09
Not usable anymore outside MSFT or standalone project


mjudeikis:
**What happened**:

After the introduction of https://github.com/Azure/AgentBaker/pull/3792 this project is not usable outside Microsoft infrastructure. We use this project to do security scans of Azure image structure and configurations, run tests and overlay configuration for some of our BYO nodes. 

Now to build windows nodes you need access to MSFT managed identity and private packages. 

**What you expected to happen**:

Opensource stay open :) 

**How to reproduce it**:

**Anything else we need to know?**:

**Environment**:
- AgentBaker version:
- Kubernetes version (use `kubectl version`):
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

AbelHu:
We need to test or build windows vhds for some urgent fixes. Those URLs will be public finally so it will not block you.

mjudeikis:
The blocking now is that without MSI you can't even run builds anymore. It's required variable now to be able to run packer build.

mjudeikis:
Would something like this would work? 
https://github.com/Azure/AgentBaker/pull/3818
Use MSI only for private packages pull. This way we could run same build flow and it still works. As not despite the fact packages will be public at some point, builds are not runnable anymore as MSI identity takes over in env variables.

mjudeikis:
@junjiezhang1997  let me know if you have any questions

junjiezhang1997:
@mjudeikis Sorry for the late reply. I'm just preparing the pull request to fix it.

junjiezhang1997:
@mjudeikis I tried the method you mentioned in #3818, but our pipeline failed with error "Identity not found" when trying to download private packages with msi. Also I tried to set the default value of msi string as empty and it also failed because it's not a valid msi string. I think I need more time to find the solution. Maybe you could set the variable "AZURE_MSI_RESOURCE_STRING" in your pipeline to mitigate this issue?

mjudeikis:
I think this means we need to start using MSI for these builds as packer will try to inject this into the build VM, so this must be valid MSI. WIll need to check how hard is to re-engineer our pipelines for this. Will let you know in a day or so.

junjiezhang1997:
@mjudeikis I think #3827 could solve this issue. Could you help test my branch on your pipelines? Thanks!

mjudeikis:
Checking now

mjudeikis:
[...]

PR #3827, opened 2023-11-14
fix: move msi to be optional

junjiezhang1997:
**What type of PR is this?**
/kind fix

Move msi to be optional.
#3810 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/init-variables.sh
+++ b/vhdbuilder/packer/init-variables.sh
@@ -206,7 +206,7 @@ windows_private_packages_url=""
 
 windows_msi_resource_string=()
 if [ -n "${WINDOWS_MSI_RESOURCE_STRING}" ]; then
-	windows_msi_resource_string+=${WINDOWS_MSI_RESOURCE_STRING}
+	windows_msi_resource_string+=(${WINDOWS_MSI_RESOURCE_STRING})
 fi
 
 # shellcheck disable=SC2236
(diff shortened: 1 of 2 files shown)


mjudeikis:
Worked like a charm :)
/approve
/lgtm

vhdbuilder/packer/init-variables.sh
```diff
@@ -203,7 +203,12 @@ WINDOWS_IMAGE_URL=""
 windows_servercore_image_url=""
 windows_nanoserver_image_url=""
 windows_private_packages_url=""
-windows_msi_resource_string=${WINDOWS_MSI_RESOURCE_STRING}
+
+windows_msi_resource_string=()
```
ShiqianTao, line 207: Why an empty string is `()` instead of `""`?

junjiezhang1997: This variable is an array so the initialization is ().

junjiezhang1997: Maybe I should rename it to avoid confusion,

AbelHu: Or you can construct the array in line 209 since we only use one msi for now.

junjiezhang1997: The condition could set msi as an optional choice for creating windows vm, which could help solve the issue mentioned in pr description.

junjiezhang1997: Also rename this variable to windows_msi_resource_strings to avoid confusion.

AbelHu (approved):
LGTM. Thanks for the PR.

mjudeikis:
Thanks for the quick turnaround! OSS for the win :)

PR #3830, opened 2023-11-14
feat: add 10 variables to the bootstrap template

Devinwong:
**What type of PR is this?**
/kind feature

add 10 variables to the template

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

terminal outputs looks like this.

> ADMINUSER="azureuser"
PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log"
MOBY_VERSION=""
TENANT_ID=""
KUBERNETES_VERSION="1.26.0"
HYPERKUBE_URL="mcr.microsoft.com/oss/kubernetes/"
KUBE_BINARY_URL=""
CUSTOM_KUBE_BINARY_URL="https://acs-mirror.azureedge.net/kubernetes/v1.26.0/binaries/kubernetes-node-linux-amd64.tar.gz"
KUBEPROXY_URL=""
DISABLE_SSH="false"
IS_CUSTOM_CLOUD=false

**Release note**:
```
none
```


pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -1,3 +1,10 @@
-#!/bin/bash
-echo "hello"
-#This is a simple placeholder script, but will eventually resemble what cse_cmd.sh looks today
\ No newline at end of file
+ADMINUSER = "{{.LinuxAdminUsername}}"
```
cameronmeissner, line 1: can't have spaces between var name and `=`'s

pkg/parser/parser.go
```diff
@@ -24,10 +27,38 @@ func executeBootstrapTemplate() (string, error) {
...
+	json.Unmarshal(inputJson, &inputContract)
+	triggerBootstrapScript, err := executeBootstrapTemplate(&inputContract)
+	if err != nil {
+		log.Printf("Failed to execute the template: %v", err)
+	}
+	log.Println("output env vars:")
 	log.Println(triggerBootstrapScript)
 }
+
+func getBaseTemplate() *nbcontractv1.Configuration {
```
anujmaheshwari1, line 53: can we put this in a different file like how ab e2e does it? https://github.com/Azure/AgentBaker/blob/master/e2e/template.go

also, lets try to use as much of the actual variable values as we can, i.e what we expect from the bootstrapper.

eg, instead of test k8s version, lets use an actual k8s version, similarly for URLs, you can refer to template.go that I linked.

this will make it easier to test once we are ready

pkg/parser/parser.go
```diff
@@ -24,10 +27,40 @@ func executeBootstrapTemplate() (string, error) {
...
+		log.Printf("Failed to marshal the nbcontractv1 to json: %v", err)
 	}
+	log.Println("Input Json: ")
+	log.Println(string(inputJSON))
+
+	// inputJson above will be provided by bootstrappers. We are using getBaseTemplate() for dev/test purpose for now.
+	// We can further move it to other file for unit tests later.
 
+	inputContract := nbcontractv1.Configuration{}
+	json.Unmarshal(inputJSON, &inputContract)
```
cameronmeissner, line 41: missing err assignment?

pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -1,3 +1,10 @@
...
+PROVISION_OUTPUT = "{{.ProvisionOutput}}"
+REPO_DEPOT_ENDPOINT ="{{.RepoDepotEndpoint}}"
```
anujmaheshwari1, line 3: lets try to imitate the actual cse_cmd.sh as much as possible, the repo depot endpoint had a conditional on custom cloud, we should include that too

pkg/parser/parser.go
Devinwong: Sounds good.
Here are what I could find from the template.go. 

PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log"
REPO_DEPOT_ENDPOINT=""
MOBY_VERSION=""
TENANT_ID=""
KUBERNETES_VERSION="1.26.0"
HYPERKUBE_URL="mcr.microsoft.com/oss/kubernetes/"
KUBE_BINARY_URL=""
CUSTOM_KUBE_BINARY_URL="https://acs-mirror.azureedge.net/kubernetes/v1.26.0/binaries/kubernetes-node-linux-amd64.tar.gz"
KUBEPROXY_URL="

pkg/parser/parser.go
```diff
@@ -3,8 +3,11 @@ package parser
 import (
 	"bytes"
 	_ "embed"
+	"encoding/json"
```
anujmaheshwari1, line 6: can you try the Golang JSON serializer with an enum field? I think it serializes the enum to its numeric value instead, might be worth checking if it deserializes it back correctly though. SSHStatus might be a good one

pkg/parser/parser.go
Devinwong: fixed

pkg/parser/cse_cmd.sh.gtpl
Devinwong: @anujmaheshwari1 I added a {{if}} to the REPO_DEPOT_ENDPOINT. 
And I add a comment saying that {{/* need to properly add {{GetInitAKSCustomCloudFilepath}} >> /var/log/azure/cluster-provision.log 2>&1; */}}

pkg/parser/parser.go
Devinwong: added DISABLE_SSH. PTAL

pkg/parser/parser.go
```diff
@@ -24,10 +32,44 @@ func executeBootstrapTemplate() (string, error) {
...
+	triggerBootstrapScript, err := executeBootstrapTemplate(&inputContract)
 	if err != nil {
 		log.Printf("Failed to execute the template: %v", err)
 	}
 
+	log.Println("output env vars:")
 	log.Println(triggerBootstrapScript)
 }
+
+func getBoolStr(state nbcontractv1.FeatureState) string {
```
Devinwong, line 61: need to add a function to convert FeatureState to bool string

pkg/parser/parser.go
```diff
@@ -24,10 +32,44 @@ func executeBootstrapTemplate() (string, error) {
...
+	if state == nbcontractv1.FeatureState_FEATURE_STATE_ENABLED {
+		return "true"
+	}
+
+	return "false"
+}
+
+func getInverseBoolStr(state nbcontractv1.FeatureState) string {
```
Devinwong, line 69: Some env vars need inverse bool str, for example, disable_ssh. Let me know if there is other way to do this better.

pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -1,3 +1,17 @@
...
+{{/* need to properly add {{GetInitAKSCustomCloudFilepath}} >> /var/log/azure/cluster-provision.log 2>&1; */}}
+{{end}}
+MOBY_VERSION="{{.MobyVersion}}"
+TENANT_ID="{{.TenantId}}"
+KUBERNETES_VERSION="{{.KubernetesVersion}}"
+HYPERKUBE_URL="{{.HyperkubeUrl}}"
+KUBE_BINARY_URL="{{.KubeBinaryUrl}}"
+CUSTOM_KUBE_BINARY_URL="{{.CustomKubeBinaryUrl}}"
+KUBEPROXY_URL="{{.KubeproxyUrl}}"
+
```
anujmaheshwari1, line 14: extra line for variables between? we can remove it for now and order it with cse_cmd.sh later

pkg/parser/parser.go
```diff
@@ -3,19 +3,27 @@ package parser
...
+	nbcontractv1 "github.com/Azure/agentbaker/pkg/proto/nbcontract/v1"
 )
 
 var (
 	//go:embed cse_cmd.sh.gtpl
 	bootstrapTrigger         string
-	bootstrapTriggerTemplate = template.Must(template.New("triggerBootstrapScript").Parse(bootstrapTrigger)) //nolint:gochecknoglobals
+	bootstrapTriggerTemplate = template.Must(template.New("triggerBootstrapScript").Funcs(funcMap).Parse(bootstrapTrigger)) //nolint:gochecknoglobals
+	//nolint:gochecknoglobals
+	funcMap = template.FuncMap{
```
anujmaheshwari1, line 18: these do not need to be global vars, we can do something like this if we do want to use func map:
https://github.com/Azure/AgentBaker/blob/bb5ee284ab17e8b610a6e7fc77e08519c67c1532/pkg/agent/baker.go#L183

pkg/parser/parser.go
anujmaheshwari1: nit: we should considering moving out these funcs to a different file, helper.go or types.go to increase modularity

pkg/parser/parser.go
anujmaheshwari1: hmm i am not fond of this but we can iterate on this going forward, wont block the PR on this.

anujmaheshwari1 (approved):
left some suggestions which I think we should incorporate, but wont block PR in favor of iterating on them in subsequent PRs.

pkg/parser/parser.go
Devinwong: updated

pkg/parser/parser.go
Devinwong: Good point. updated

pkg/parser/cse_cmd.sh.gtpl
Devinwong: updated

PR #3839, opened 2023-11-16
chore: require ab e2e run to be successful if e2e dir is changed

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/e2e.yaml
+++ b/.pipelines/e2e.yaml
@@ -7,7 +7,9 @@ pr:
     - 'official/*'
   paths:
     include:
-      e2e
+      - 'e2e'
+    exclude:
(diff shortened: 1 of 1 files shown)


.pipelines/e2e.yaml
```diff
@@ -5,6 +5,11 @@ pr:
     include:
     - main
     - 'official/*'
+  paths:
+    include:
+    - 'e2e'
```
cameronmeissner, line 10: shouldn't need the quotes I don't think

cameronmeissner:
already addressed

PR #3841, opened 2023-11-17
test: add e2e scenarios for more gpu skus

lilypan26:
**What type of PR is this?**

/kind test

Optionally add one or more of the following kinds if applicable:
/kind api-change
/kind deprecation
/kind failing-test
/kind flake
/kind regression
-->

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/scenario/images.go
```diff
@@ -10,3 +10,9 @@ var DefaultImageVersionIDs = map[string]string{
 	"marinerv2-arm64":    "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/CBLMarinerV2Gen2Arm64/versions/1.1699146990.16638",
 	"azurelinuxv2-arm64": "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/AzureLinuxV2Gen2Arm64/versions/1.1699146999.31863",
 }
+
+var DefaultGPUSeriesSKU = map[string]string{
```
cameronmeissner, line 14: nit: `DefaultGPUSeriesVMSizes`?

e2e/scenario/init.go
```diff
@@ -54,4 +54,10 @@ func scenarios() []*Scenario {
 		ubuntu2204CustomCATrust(),
 		ubuntu2204ArtifactStreaming(),
 	}
+
+	for vmSeries, _ := range DefaultGPUSeriesSKU {
+		scenarios = append(scenarios, ubuntu2204gpu(vmSeries))
```
cameronmeissner, line 59: why not have a separate function, maybe called `gpuScenarios()`, which also returns a `[]*Scenario` - then we just concat the output of that with this slice using ellipsis operator

cameronmeissner: we can do that concatenation in the InitiScenarioTable func, and keep this func the same I think

cameronmeissner: something like `scenarios := scenarios() ... scenarios = append(scenarios, gpuScenarios()...)`

e2e/scenario/scenario_ubuntu2204-gpu.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+)
+
+// Returns config for the 'gpu' E2E scenario
+func ubuntu2204gpu(vmSeries string) *Scenario {
+	return &Scenario{
+		Name:        "ubuntu2204-gpu-" + vmSeries,
```
cameronmeissner, line 12: nit: fmt.Sprintf instead?

e2e/scenario/scenario_ubuntu2204-gpu.go
```diff
@@ -0,0 +1,34 @@
...
+		Description: "Tests that a GPU-enabled node using an Ubuntu 2204 VHD can be properly bootstrapped",
```
cameronmeissner, line 13: should we also call out the vm size in the description?

e2e/scenario/images.go
lilypan26: sounds good, do you think the file name should be changed? wasn't sure what a better name would be

ganeshkumar5699: I wonder if a separate file for this would be better - like gpuvmsizes.go or it can also be moved to e2e/const.go

cameronmeissner: maybe we could just do `vm.go` or similar

e2e/scenario/vm.go
```diff
@@ -0,0 +1,7 @@
+package scenario
+
+var DefaultGPUSeriesVMSizes = map[string]string{
```
cameronmeissner, line 3: I decided to try going the `go:embed` route for the default SIG image versions map, wondering if we should also do the same here - can potential follow-up on that in a separate PR as well

e2e/scenario/init.go
```diff
@@ -55,3 +58,13 @@ func scenarios() []*Scenario {
 		ubuntu2204ArtifactStreaming(),
 	}
 }
+
+func gpuAllVMSizeScenarios() []*Scenario {
+	scenarios := []*Scenario{}
```
cameronmeissner, line 63: nit: `var scenarios []*Scenario`

e2e/scenario/init.go
```diff
@@ -7,7 +7,10 @@ import (
 // Initializes and returns the set of scenarios comprising the E2E suite in table-form.
 func InitScenarioTable(include, exclude map[string]bool) Table {
 	table := Table{}
-	for _, scenario := range scenarios() {
+	scenarios := scenarios()
+	scenarios = append(scenarios, gpuAllVMSizeScenarios()...)
```
cameronmeissner, line 11: nit: `scenarios := append(scenarios(), gpuAllVMSizeScenarios()...)

e2e/scenario/init.go
```diff
@@ -55,3 +58,13 @@ func scenarios() []*Scenario {
 		ubuntu2204ArtifactStreaming(),
 	}
 }
+
+func gpuAllVMSizeScenarios() []*Scenario {
```
cameronmeissner, line 62: nit: `gpuVMSizeScenarios()`?

cameronmeissner (approved):

e2e/scenario/vm.go
lilypan26: deferring for now

PR #3844, opened 2023-11-17
chore: rename artifact files in windows vhd pipelines

junjiezhang1997:
**What type of PR is this?**
/kind improvement

rename artifact files in windows vhd pipelines.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -130,13 +130,14 @@ steps:
     displayName: Run VHD cache test
 
 # We can upload release notes for check-in pr and sig mode to validate whether it is expected.
-  - task: PublishPipelineArtifact@0
-    inputs:
-        artifactName: 'vhd-release-notes-${{ parameters.artifactName }}'
-        targetPath: 'release-notes.txt'
   
   - bash: |
-      WINDOWS_VERSION="$(grep -a "OS Version" < release-notes.txt | cut -d ":" -f 2 | tr -d "[:space:]")"
+      if [[ ${{ parameters.artifactName }} =~ "2019" ]]; then
+        WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2019_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
+      fi
+      if [[ ${{ parameters.artifactName }} =~ "2022" ]]; then
+        WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2022_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
+      fi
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/list-images.ps1
```diff
@@ -10,7 +10,7 @@ $windowsImageVersion = $env:WindowsImageVersion
 
 $ErrorActionPreference = "Stop"
 
-$imageBomJsonFilePath = "c:\image-bom.json"
+$imageBomJsonFilePath = "c:\image-list.json"
```
AbelHu, line 13: This file will exist inside every VHD. I think that we should not change this file name to keep it to be consistent with Linux. We only need to rename it before uploading it to artifacts.

junjiezhang1997: Thanks for the comment. Just find it and the empty file of image-list.json in artifacts should be caused by this change.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -130,23 +130,38 @@ steps:
...
+  - bash: |
+      if [[ ${{ parameters.artifactName }} =~ "2019" ]]; then
+        WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2019_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
+      elif [[ ${{ parameters.artifactName }} =~ "2022" ]]; then
+        WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2022_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
+      else
+        echo "Current distro is not supported to get image version. You need to update related code."
+        exit 1
+      fi
+      IMAGE_VERSION="$WINDOWS_VERSION.${BUILD_DATE}"
```
AbelHu, line 143: It seems like that `${BUILD_DATE}` is empty
image

junjiezhang1997: Modify the code for obtaining BUILD_DATE.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -50,7 +50,7 @@ steps:
       if [[ "${USE_RELEASE_DATE}" = "False" ]]; then \
         echo "use current date as build date";  BUILD_DATE=$(date +"%y%m%d"); \
       else \
-        echo "use release date as build date"; branchName=$(Build.SourceBranchName); BUILD_DATE=${branchName: -6}; \
+        echo "use release date as build date"; BUILD_DATE=${RELEASE_DATE}; \
```
AbelHu, line 53: It is better to validate whether `${RELEASE_DATE}` is valid, for example, not null, not empty, the format (`yymmdd`).

AbelHu (approved):
LGTM. Thanks for the PR.

PR #3848, opened 2023-11-21
chore: support windows vhd build for 23H2

junjiezhang1997:
**What type of PR is this?**
/kind feature

Support windows vhd build for 23H2.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/e2e-windows.yaml
+++ b/.pipelines/e2e-windows.yaml
@@ -49,12 +49,12 @@ parameters:
   displayName: Windows NV Series VM Size (gen2)
   type: string
   default: Standard_NV6ads_A10_v5
-- name: test23H2containerd
-  displayName: Test 23H2 containerd
+- name: test23H2
+  displayName: Test 23H2
   type: boolean
   default: False
-- name: test23H2containerdgen2
-  displayName: Test 23H2 containerd Gen 2
+- name: test23H2gen2
+  displayName: Test 23H2 Gen 2
   type: boolean
   default: False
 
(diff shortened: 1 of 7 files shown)


.pipelines/e2e-windows.yaml
```diff
@@ -49,6 +49,14 @@ parameters:
   displayName: Windows NV Series VM Size (gen2)
   type: string
   default: Standard_NV6ads_A10_v5
+- name: test23H2containerd
```
ShiqianTao, line 52: I prefer to remove "containerd" that same as other varibales

junjiezhang1997: Fixed. Miss here before.

.pipelines/e2e-windows.yaml
```diff
@@ -116,4 +124,30 @@ stages:
...
+  - template: ./templates/e2e-windows-template.yaml
+    parameters:
+      windowsImage: 23H2
+      windowsOSSKU: WindowsAnnual
+      windowsVMSize: Standard_D2_v2
+      windowsDistro: aks-windows-23H2
+      windowsNodeImageVersion: AKSWindows-23H2-test-2023.02.07
+      storageAccount: akswinstore23H2
+
+- stage: test_23H2_gen2
```
AbelHu, line 142: Not an issue. Maybe we can keep only one test (default gen1 and later change it to gen2) to save the cost

AbelHu (approved):
LGTM. Thanks for the PR

.pipelines/e2e-windows.yaml
junjiezhang1997: Create a task to track this issue.

PR #3856, opened 2023-11-27
chore: add httpproxyconfig to the contract

anujmaheshwari1:
**What type of PR is this?**
This is the first PR to start formalizing the bootstrap contract. Taking cue from AKS-RP protos structuring, the idea is to have different proto files for broader message types so as to increase readability and have proper segregation of fields.

HTTPProxyConfig is user input, and hence, it needs to be part of the bootstrap contract.

In the current cse_cmd.sh, there are 7 variables that reference/use HTTPProxyConfig:

The following 3 are conditionals that simply depend on the actual proxy values, and whether or not they are populated. These do not need to be in the contract, and can either be computed by the parser on the VHD, or be expanded directly in the bootstrap scripts.

SHOULD_CONFIGURE_HTTP_PROXY="{{ShouldConfigureHTTPProxy}}"
SHOULD_CONFIGURE_HTTP_PROXY_CA="{{ShouldConfigureHTTPProxyCA}}"
PROXY_VARS="{{GetProxyVariables}}"

The following 4 are actual values of the proxy fields and hence need to be in the contract.

HTTP_PROXY_TRUSTED_CA="{{GetHTTPProxyCA}}"
HTTP_PROXY_URLS="{{GetHTTPProxy}}"
HTTPS_PROXY_URLS="{{GetHTTPSProxy}}"
NO_PROXY_URLS="{{GetNoProxy}}"

Please ignore the changes to `config.proto`, since its the one final file we are adding the contract to and it is changing state from its current form. The file to review is `httpproxyconfig.proto` and its corresponding import in `config.proto`

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


anujmaheshwari1:
Please ignore the changes to `config.proto`, since its the one final file we are adding the contract to and it is changing state from its current form. The file to review is `httpproxyconfig.proto` and its corresponding import in `config.proto`

pkg/proto/nbcontract/v1/httpproxyconfig.proto
tallaxes, line 1: A couple of general comments that will likely apply to everything going forward, so good to get out of the way early:
- Not convinced we need a separate file for each substructure - but can revisit later
- Do we need *Config suffix for each message? Seems like it should be clear from the context ...
- What is the appropriate naming convention here? I see we are switching from snake to capital camel case here - is this deliberate?

pkg/proto/nbcontract/v1/httpproxyconfig.proto
```diff
@@ -0,0 +1,16 @@
...
+  optional string HTTPProxy = 1;
+
+  // HTTPSProxy represents the proxy endpoint using HTTPS, that the traffic should go through for making HTTPS connections outside the cluster
+  optional string HTTPSProxy = 2;
+
+  // NoProxy represents a list of endpoints including IPs, domains, or other network CIDRs that should not be accessed through the proxy
+  repeated string NoProxy = 3;
+
+  // TrustedCA repesents alternative CA certificate content
+  optional string TrustedCA = 4;
```
tallaxes, line 15: May want to keep "Proxy" somewhere in the name, for consistency with the rest. (The alternative might be to drop "Proxy" everywhere in this context, but seems it would be less readable / clear ...)

pkg/proto/nbcontract/v1/httpproxyconfig.proto
```diff
@@ -0,0 +1,16 @@
...
+
+message HTTPProxyConfig {
+  // HTTPProxy represents the proxy endpoint/URL using HTTP, that the traffic should go through for making HTTP connections outside the cluster
+  optional string HTTPProxy = 1;
+
+  // HTTPSProxy represents the proxy endpoint using HTTPS, that the traffic should go through for making HTTPS connections outside the cluster
+  optional string HTTPSProxy = 2;
+
+  // NoProxy represents a list of endpoints including IPs, domains, or other network CIDRs that should not be accessed through the proxy
+  repeated string NoProxy = 3;
```
tallaxes, line 12: Part of me is wondering if naming it something like "NoProxyEntry" would make it clear it is not a full NO_PROXY string with separators ...

tallaxes (approved):
LGTM, some minor comments inline

Devinwong (approved):

pkg/proto/nbcontract/v1/httpproxyconfig.proto
anujmaheshwari1: 1) Like I said, just better structuring + following how RP does it. https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/protos/hcp/types/agentpool/v1
2) Well, I think the suffix *Config may be useful sometimes where the name itself may be confusing. If we go with HTTPProxy, it overlaps with one of the actual fields itself, HTTPProxy. But the same is not necessarily true for other messages like CustomCloudEnvironment or CustomNodeOS, Config just makes it "fuller"
3) Again, I followed RP pattern but it turns out RP itself isnt consistent, it has a mix of both camel case and snake case. I am going with capital camel case unless there is a specific need for snake case.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
anujmaheshwari1: fair point, let me add Proxy prefix.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
anujmaheshwari1: inclined to keep as is, since there are comments highlighting the exact use, but can change if you have a strong preference for the name.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
anujmaheshwari1: done.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
tallaxes: For 1, and especially 3: While RP is useful for reference, I don't think we have to be following RP closely, should be doing whatever makes more sense in this new context. The field naming, for example, should follow whatever is most common for this context. I think protobuf itself usually uses / encourages snake case for field names. Don't have a strong opinion here, but wanted to point out we don't have to follow RP, and that this is worth a little bit of thought, since whatever we decide will be used everywhere.

anujmaheshwari1: changed to camel case, followed docs from protobuf and google and that seems to be the convention!

PR #3861, opened 2023-11-28
feat: add custom search domain to bootstrap contract

anujmaheshwari1:
**What type of PR is this?**
This PR adds custom search domain to the bootstrap contract. I havent found much information on this particular attribute, but looking at the code on how the fields are consumed in bootstrap scripts, these fields seem to be required if set.

The 4 environment variables referencing these in cse_cmd.sh are:
HAS_CUSTOM_SEARCH_DOMAIN="{{HasCustomSearchDomain}}"

^ the above is not needed, can be computed on the VHD since it simply checks for the existence of the actual fields.

CUSTOM_SEARCH_DOMAIN_NAME="{{GetSearchDomainName}}"
CUSTOM_SEARCH_REALM_USER="{{GetSearchDomainRealmUser}}"
CUSTOM_SEARCH_REALM_PASSWORD="{{GetSearchDomainRealmPassword}}"

^ these 3 are inputs from bootstrapper/user and need to be present in the contract.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/config.proto
```diff
@@ -63,29 +64,29 @@ message Configuration {
...
-  bool needs_cgorupv2 = 60; // we can calculate this on the VHD, will remove from final contract
+  string azure_private_registry_server = 55;
+  string private_egress_proxy_address = 56;
+  FeatureState artifact_streaming_enabled = 57;
+  bool is_kata = 58; // we can calculate this on the VHD, will remove from final contract
+  bool needs_cgorupv2 = 59; // we can calculate this on the VHD, will remove from final contract
 
   // Here begins the final contract
-  nbcontract.v1.HTTPProxyConfig HTTPProxyConfig = 61;
+  nbcontract.v1.HTTPProxyConfig http_proxy_config = 60;
```
anujmaheshwari1, line 88: changed httpproxyconfig to snake case as per offline discussions with Alex

pkg/proto/nbcontract/v1/customsearchdomain.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+message CustomSearchDomain {
+  // Did not find much information on what this really does, but can confirm from the usage in cse_config.sh that these fields are indeed required if set on bootstrapper
+  // They are expanded in the file https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
+  // Given that there is a conditional in the script on whether or not these fields are present, these are marked optional, however, in RP datamodel, these arent marked optional as part of Linux Profile
+  optional string name = 1;
+  optional string realm_user = 2;
+  optional string real_password = 3;
```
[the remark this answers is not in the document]
anujmaheshwari1: good catch, fixed

pkg/proto/nbcontract/v1/config.proto
cameronmeissner: seems like we've generally begun moving towards snake case ubiquitously?

cameronmeissner (approved):

Devinwong (approved):

pkg/proto/nbcontract/v1/customsearchdomain.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+message CustomSearchDomain {
```
tallaxes, line 4: I don't know enough about this space, but find the "custom search domain" here a little confusing. Yes, looking at the script, it does add extra `dns-search` to the interface, but it looks like it also joins the node to the AD domain/realm, which seem a much more important/impactful action, that I would never have guessed from the name. Should we revisit the name? (Maybe tracking down the documented AKS/RP functionality or data model helps?)

pkg/proto/nbcontract/v1/customsearchdomain.proto
```diff
@@ -0,0 +1,11 @@
...
+  // Did not find much information on what this really does, but can confirm from the usage in cse_config.sh that these fields are indeed required if set on bootstrapper
+  // They are expanded in the file https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
+  // Given that there is a conditional in the script on whether or not these fields are present, these are marked optional, however, in RP datamodel, these arent marked optional as part of Linux Profile
+  optional string name = 1;
+  optional string realm_user = 2;
+  optional string realm_password = 3;
```
tallaxes, line 10: From the script it seems like all these would be required - if used at all. So maybe it is upper level field that needs to be optional, and these individual fields would be required?

pkg/proto/nbcontract/v1/customsearchdomain.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+message CustomSearchDomain {
+  // Did not find much information on what this really does, but can confirm from the usage in cse_config.sh that these fields are indeed required if set on bootstrapper
+  // They are expanded in the file https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
+  // Given that there is a conditional in the script on whether or not these fields are present, these are marked optional, however, in RP datamodel, these arent marked optional as part of Linux Profile
+  optional string name = 1;
```
tallaxes, line 8: Even though it may be clear from the context, a more descriptive field name (domain_name? realm_name?)  would be appropriate, I think ...

pkg/proto/nbcontract/v1/customsearchdomain.proto
anujmaheshwari1: from AKS_RP datamodel, its the same https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/protos/hcp/types/managedcluster/profiles/v1/linuxprofile.proto&version=GBmaster&line=33&lineEnd=34&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

I wasnt able to find documentation in our wiki and the blame on this goes back to 3 years when AB was created. I can try digging up more.

pkg/proto/nbcontract/v1/customsearchdomain.proto
anujmaheshwari1: oh that makes 100% sense, oversight on my part, let me fix this

anujmaheshwari1: addressed.

pkg/proto/nbcontract/v1/customsearchdomain.proto
anujmaheshwari1: changed.

PR #3862, opened 2023-11-28
design: add gpuconfig

Devinwong:
**What type of PR is this?**
/kind design
Adding GpuConfig to the contract

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+import "pkg/proto/nbcontract/v1/enum.proto";
+
+message GpuConfig {
+  // nvidia_state represents the Nvidia GPU state
+  FeatureState status = 1;
+
+  // FeatureState is_mig_node = 2; removed since it only depends on gpu_instance_profile
```
Devinwong, line 10: @anujmaheshwari1 I think we can keep the removed variables with comments for now. That will be easier to track what are removed and why it could be removed. Then we can have a clean-up at the end of development stage. Thoughts?

pkg/proto/nbcontract/v1/nodepoolprofile.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+message NodePoolProfile {
+  // vm_size represents the size of this node pool
+  string vm_size = 1;
+}
```
tallaxes, line 7: This does not quite make sense to me. For one thing, I assume vm_size is actually the VM SKU (like Standard_D2_v5), rather than the size of the node pool. For another, it is not clear to me why VHD/bootstrap would need to know either (and it could likely find out the VM SKU by itself, if it is actually needed)

Devinwong: Confirmed that we will be able to get vm_size from IMDS. Removed this as suggested. :)

pkg/proto/nbcontract/v1/nodepoolprofile.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+message NodePoolProfile {
```
anujmaheshwari1, line 4: not sure if this is the best name, and whether we need a separate proto for it? do we expect to have more fields here? if not, then vm size can just be in the main message of the contract

Devinwong: @anujmaheshwari1 There will be more fields in the nodepoolprofile. But since they are not part of the GPUConfig, I didn't put them there now. The original name from AKS-RP is Agentpoolprofile. I just think NodePool is easier to understand to me. Or you think Agentpool is a more official name?

anujmaheshwari1: Agentpool is a term very specific to AKS, I think we should make this specific to a node, rather than a particular bootstrapper, what do you think?

Devinwong: As we synced, let's start with NodePoolProfile and see if we can come up with a better name

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+import "pkg/proto/nbcontract/v1/enum.proto";
+
+message GpuConfig {
```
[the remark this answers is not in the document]
Devinwong: done

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
...
+  // nvidia_state represents the Nvidia GPU state
+  FeatureState status = 1;
+
+  // FeatureState is_mig_node = 2; removed since it only depends on gpu_instance_profile
+
+  //config_gpu_driver_if_needed represents if bootstrap process should install and configure the GPU driver when necessary. Configuration includes appropriate set up of components like the fabric manager where applicable.
+  FeatureState config_gpu_driver = 2;
+
+  //enable_gpu_device_plugin_if_needed represents if special config is needed for MIG GPUs that use GPU dedicated VHDs and enable the device plugin (for all GPU dedicated VHDs)
+  FeatureState enable_gpu_device_plugin = 3;
```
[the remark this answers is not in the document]
Devinwong: Removing "enable" makes sense. Done.

pkg/proto/nbcontract/v1/enum.proto
tallaxes, line 1: Should have a more descriptive file name (or just keep in the main file)

pkg/proto/nbcontract/v1/nodepoolprofile.proto
tallaxes: I agree with looking hard for opportunities to get rid of the message. Hard to judge without knowing which other fields may come in there. But having vm_size under NodePoolProfile is already confusing, since we are starting to support mixed SKU NodePools.

pkg/proto/nbcontract/v1/gpuconfig.proto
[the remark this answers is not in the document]
tallaxes: Given that FeatureState already supports (and defaults to?) "FEATURE_STATE_UNSPECIFIED", I find "optional FeatureState" confusing, but maybe there are valid use cases for it?

Related: This is another case where the GpuConfig as a whole is optional, and all individual fields are optional. Again, maybe valid, but good to check if intentional.

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+import "pkg/proto/nbcontract/v1/enum.proto";
+
+message GpuConfig {
+  // nvidia_state represents the Nvidia GPU state
```
tallaxes, line 7: I don't quite understand the meaning of this, at least not from the comment. The desired state of Nvidia GPU on the node, to be set by the bootstrap process?

Related: Do we have a single place that captures all of the different GPU settings we support, with their meaning and interaction? Still concisely, but longer form than field comments? (It would also capture the relationship between these and the NVidia operator ...)

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
...
+  optional FeatureState nvidia_state = 1;
+
+  // FeatureState is_mig_node = 2; removed since it only depends on gpu_instance_profile
+
+  //config_gpu_driver represents if bootstrap process should install and configure the GPU driver when necessary. Configuration includes appropriate set up of components like the fabric manager where applicable.
```
tallaxes, line 12:
```suggestion
  // config_gpu_driver specifies whether bootstrap process should install and configure the GPU driver when necessary. Configuration includes appropriate set up of components like the fabric manager where applicable.
```

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
...
+  optional FeatureState config_gpu_driver = 2;
+
+  //gpu_device_plugin represents if special config is needed for MIG GPUs that use GPU dedicated VHDs and enable the device plugin (for all GPU dedicated VHDs)
```
tallaxes, line 15:
```suggestion
  // gpu_device_plugin specifies whether special config is needed for MIG GPUs that use GPU dedicated VHDs and enable the device plugin (for all GPU dedicated VHDs)
```

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
...
+  optional FeatureState gpu_device_plugin = 3; 
+
+  // FeatureState fabric_manager_status = 5;  removed. consider just extracting vm_size since both GPUNeedsFabricManager and getGPUDriverVersion are just depending ont vm_size.
+  // string gpu_driver_version = 6; removed. this is dependent only on VM size, can be calculated on the VHD
+
+  // gpu_instance_profile represents the GPU instance profile.
```
tallaxes, line 21: Maybe a little more detail here?

tallaxes (changes requested):

pkg/proto/nbcontract/v1/enum.proto
Devinwong: @tallaxes  My intention was just to put all the enum in a single file. We will have more enums later on. But I am open to move it back to the main file.

pkg/proto/nbcontract/v1/config.proto
```diff
@@ -86,6 +87,7 @@ message Configuration {
 
   // Here begins the final contract
   nbcontract.v1.HTTPProxyConfig HTTPProxyConfig = 61;
+  optional nbcontract.v1.GpuConfig gpu_config = 62;
```
ganeshkumarashok, line 90: nit: GpuConfig  -> GPUConfig, since GPU is an acronym

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
...
+  // FeatureState is_mig_node = 2; removed since it only depends on gpu_instance_profile
+
+  //config_gpu_driver represents if bootstrap process should install and configure the GPU driver when necessary. Configuration includes appropriate set up of components like the fabric manager where applicable.
+  optional FeatureState config_gpu_driver = 2;
+
+  //gpu_device_plugin represents if special config is needed for MIG GPUs that use GPU dedicated VHDs and enable the device plugin (for all GPU dedicated VHDs)
+  optional FeatureState gpu_device_plugin = 3; 
+
+  // FeatureState fabric_manager_status = 5;  removed. consider just extracting vm_size since both GPUNeedsFabricManager and getGPUDriverVersion are just depending ont vm_size.
+  // string gpu_driver_version = 6; removed. this is dependent only on VM size, can be calculated on the VHD
```
Bryce-Soghigian, line 19: Ah is this currently calculated in newest vhds?

pkg/proto/nbcontract/v1/gpuconfig.proto
ganeshkumarashok: For the related question: there's a PRD for GPU operator which captures constraints and certain config relationships (limited though). It would be a good to-do for us to add.

pkg/proto/nbcontract/v1/nodepoolprofile.proto
ganeshkumarashok: Regarding NodePool vs AgentPool - I think both terms are AKS-specific. NodePool is more descriptive though

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -0,0 +1,23 @@
...
+
+  // FeatureState is_mig_node = 2; removed since it only depends on gpu_instance_profile
+
+  //config_gpu_driver represents if bootstrap process should install and configure the GPU driver when necessary. Configuration includes appropriate set up of components like the fabric manager where applicable.
+  optional FeatureState config_gpu_driver = 2;
+
+  //gpu_device_plugin represents if special config is needed for MIG GPUs that use GPU dedicated VHDs and enable the device plugin (for all GPU dedicated VHDs)
+  optional FeatureState gpu_device_plugin = 3; 
+
+  // FeatureState fabric_manager_status = 5;  removed. consider just extracting vm_size since both GPUNeedsFabricManager and getGPUDriverVersion are just depending ont vm_size.
```
ganeshkumarashok, line 18:
```suggestion
  // FeatureState fabric_manager_status = 5;  removed. consider just extracting vm_size since both GPUNeedsFabricManager and getGPUDriverVersion depend only on the vm_size.
```

pkg/proto/nbcontract/v1/gpuconfig.proto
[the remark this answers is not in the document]
ganeshkumarashok: nit: GpuConfig -> GPUConfig

pkg/proto/nbcontract/v1/gpuconfig.proto
Devinwong: yes, with vm_size added in the contract

pkg/proto/nbcontract/v1/gpuconfig.proto
Devinwong: @ganeshkumarashok That will be awesome. We will add the detailed descriptions back once they're ready.

pkg/proto/nbcontract/v1/gpuconfig.proto
Devinwong: @ganeshkumarashok  Please also provide more details about this. Thanks

pkg/proto/nbcontract/v1/gpuconfig.proto
Devinwong: Thanks for catching this. We will eventually remove all the comments here. For now they are here for us to track where they go more easily.

ganeshkumarashok (approved):

PR #3863, opened 2023-11-28
chore: cache secrets-store-provider-azure v1.5.0

jiashun0011:
**What type of PR is this?**
/kind feature

- regularly upgrade. cache secrets-store-csi-driver v1.4.0.
- regularly upgrade. cache secrets-store-provider-azure v1.5.0.

Fixes #

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
chore: cache secrets-store-provider-azure v1.5.0
```


vhdbuilder/packer/components.json
```diff
@@ -72,14 +72,16 @@
       "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v1.3.4",
-        "v1.3.4-1"
+        "v1.3.4-1",
```
anujmaheshwari1, line 75: we are only adding, not retiring older versions. 
our VHDs are at capacity and we are having build issues, can we please retire older versions if we are adding newer ones?

andyzhangx: if old image is still used in aks-rp, we should keep it until new image is used in aks-rp

andyzhangx: then you should remove `v1.3.4` also since it's not used any more

```
{{- else if eq .component "secrets-store-csi-driver" -}}
  {{- if semverCompare ">=1.24.0" .version -}}v1.3.4-1
  {{- else -}}v1.3.0.3
  {{- end -}}
```

vhdbuilder/packer/generate-windows-vhd-configuration.ps1
```diff
@@ -88,8 +88,8 @@ $global:imagesToPull += @(
     "mcr.microsoft.com/oss/kubernetes-csi/azurefile-csi:v1.29.1-windows-hp",
     "mcr.microsoft.com/oss/kubernetes-csi/azurefile-csi:v1.29.2-windows-hp",
     # Addon of Azure secrets store. Owner: ZeroMagic (Ji'an Liu)
-    "mcr.microsoft.com/oss/kubernetes-csi/secrets-store/driver:v1.3.4",
```
andyzhangx, line 91: seems these two images are still used on Windows now?

andyzhangx (approved):

AbelHu (approved):
LGTM. Thanks for the PR.

AbelHu (changes requested):
Please do not use the forked repo to create the PR.

PR #3865, opened 2023-11-28
fix: update gpu driver image to support h100s

lilypan26:
**What type of PR is this?**

/kind bug

Optionally add one or more of the following kinds if applicable:
/kind api-change
/kind deprecation
/kind failing-test
/kind flake
/kind regression
-->

Update gpu driver image to include support for h100s from fix in aks-gpu: https://github.com/Azure/aks-gpu/pull/23

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker_test.go
```diff
@@ -1753,6 +1753,6 @@ var _ = Describe("getAKSGPUImageSHA", func() {
 		Expect(getAKSGPUImageSHA("standard_nv6ads_a10_v5")).To(Equal("sha-20ffa2"))
 	})
 	It("should use newest AKSGPUCudaSHA with non grid SKU", func() {
-		Expect(getAKSGPUImageSHA("standard_nc6_v3")).To(Equal("sha-e8873b"))
+		Expect(getAKSGPUImageSHA("standard_nc6_v3")).To(Equal("sha-16fd35"))
```
cameronmeissner, line 1756: oh, can we use the consts here as well?

cameronmeissner (approved):

pkg/agent/baker_test.go
UtheMan: Are we sure we want it tho, see my last pr

lilypan26: hmm I think it makes sense to use the consts. Similar to the [getGPUDriverVersion tests](https://github.com/Azure/AgentBaker/blob/master/pkg/agent/baker_test.go#L1735) above though, is there a reason we don't use the driver version const variables for each version?

PR #3871, opened 2023-11-29
chore: support to test Windows 2025

junjiezhang1997:
**What type of PR is this?**
/kind improvement

1. Modify Get-WindowsVersion to support the test of Windows 2025.
2. Add new function Get-WindowsPauseVersion for pasue image.
3. Add related unit test.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/windows/windowscsehelper.ps1
```diff
@@ -275,7 +275,21 @@ function Get-WindowsVersion {
     switch ($buildNumber) {
         "17763" { return "1809" }
         "20348" { return "ltsc2022" }
-        "25398" { return "ltsc2022" } # Only for test of partner team now. Will remove it further.
+        "25398" { return "ltsc2022" }
```
AbelHu, line 278: I think that we should not return ltsc2022 for 23H2 since we have separated the code for pause image

staging/cse/windows/containerdfunc.ps1
```diff
@@ -158,7 +158,7 @@ function Install-Containerd {
   $formatedbin = $(($CNIBinDir).Replace("\", "/"))
   $formatedconf = $(($CNIConfDir).Replace("\", "/"))
   $sandboxIsolation = 0
-  $windowsVersion = Get-WindowsVersion
+  $windowsVersion = Get-WindowsPauseVersion
```
AbelHu, line 161: FYI: It will not work until you update cse script package version.

.github/workflows/validate-windows-ut.yml
```diff
@@ -12,7 +12,13 @@ jobs:
         shell: powershell
         run: |
           Invoke-Pester parts/windows/*.tests.ps1 -Passthru
-      - name: Perform a Pester test from the staging/cse/windows/*.tests.ps1 file
-        shell: powershell
-        run: |
-          Invoke-Pester staging/cse/windows/*.tests.ps1 -Passthru
\ No newline at end of file
+  pester-test-2:
```
AbelHu, line 15: What is the purpose to separate them into two stages?

junjiezhang1997: To avoid the case that we meet before: the failure in the first run will not be shown due toe the succeed in the second run.

staging/cse/windows/containerdfunc.ps1
junjiezhang1997: I see. I will build a new cse script package after the code is merged.

.github/workflows/validate-windows-ut.yml
junjiezhang1997: I think I get something mixed here. Change it back and test again.

staging/cse/windows/containerdfunc.ps1
junjiezhang1997: Should I move the change that returns "23H2" for 23H2 in Get-WindowsVersion to the pr of updating cse script package to avoid blocking issue?

AbelHu (approved):
LGTM. Thanks for the PR.

PR #3885, opened 2023-11-30
fix: clean up AB linting errors

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/bakerapi.go
+++ b/pkg/agent/bakerapi.go
@@ -23,7 +23,7 @@ func NewAgentBaker() (AgentBaker, error) {
 
 type agentBakerImpl struct{}
 
-//nolint:revive // ctx is not used, but may be in the future
+//nolint:revive, nolintlint // ctx is not used, but may be in the future
 func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context,
 	config *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error) {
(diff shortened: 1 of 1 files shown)


pkg/agent/datamodel/mocks.go
```diff
@@ -3,77 +3,6 @@
 
 package datamodel
 
-import (
-	"github.com/Azure/go-autorest/autorest/to"
-	"github.com/google/uuid"
-)
-
-// CreateMockContainerService returns a mock container service for testing purposes.
-func CreateMockContainerService(containerServiceName, orchestratorVersion string, masterCount, agentCount int, certs bool) *ContainerService {
```
cameronmeissner, line 12: we're sure we don't have any further usage of this?

AlisonB319: Yep there are no other calls to it in the repo

pkg/agent/bakerapi.go
```diff
@@ -23,6 +23,7 @@ func NewAgentBaker() (AgentBaker, error) {
 
 type agentBakerImpl struct{}
 
+//nolint:revive // ctx is not used, but may be in the future
```
AlisonB319: directive `//nolint:revive // ctx is not used, but may be in the future` is unused for linter "revive" (nolintlint) 

is a bug with golangci https://github.com/golangci/golangci-lint/issues/3228

PR #3886, opened 2023-11-30
feat: add build ID tag to abe2e vmss models for tracking

cameronmeissner:
**What type of PR is this?**
/kind feature

adds build numbers to abe2e vmss models for easier tracking/debugging

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/vmss.go
```diff
@@ -54,7 +54,14 @@ func bootstrapVMSS(ctx context.Context, t *testing.T, r *mrand.Rand, vmssName st
 }
 
 func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName string, publicKeyBytes []byte, opts *scenarioRunOpts) (*armcompute.VirtualMachineScaleSet, error) {
-	model := getBaseVMSSModel(vmssName, opts.suiteConfig.location, opts.suiteConfig.subscription, *opts.clusterConfig.cluster.Properties.NodeResourceGroup, opts.clusterConfig.subnetId, string(publicKeyBytes), customData, cseCmd)
+	model := getBaseVMSSModel(vmssName, string(publicKeyBytes), customData, cseCmd, opts)
+
+	if opts.suiteConfig.buildID != "" {
```
UtheMan, line 59: can this be nil?

e2e/vmss.go
```diff
@@ -198,9 +205,9 @@ func getVmssName(r *mrand.Rand) string {
 	return fmt.Sprintf(vmssNameTemplate, randomLowercaseString(r, 4))
 }
 
-func getBaseVMSSModel(name, location, subscription, mcResourceGroupName, subnetID, sshPublicKey, customData, cseCmd string) armcompute.VirtualMachineScaleSet {
+func getBaseVMSSModel(name, sshPublicKey, customData, cseCmd string, opts *scenarioRunOpts) armcompute.VirtualMachineScaleSet {
 	return armcompute.VirtualMachineScaleSet{
-		Location: to.Ptr(location),
+		Location: to.Ptr(opts.suiteConfig.location),
```
UtheMan, line 210: Similar question - can opts be nil? Might be worth adding some safeguards to those accesses?

e2e/vmss.go
cameronmeissner: it should be able to nil since we construct it before running all scenarios, though maybe in a separate PR we can add proto-like getters to ensure we safely access these fields

e2e/vmss.go
cameronmeissner: again, shouldn't ever be nil since we construct it before running each scenario, though wouldn't be against adding safety around this in the future

e2e/vmss.go
UtheMan: I love proto like getters.

PR #3889, opened 2023-12-05
chore: remove comments from custom data while generating it

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker.go
```diff
@@ -161,6 +161,17 @@ func (t *TemplateGenerator) getSingleLine(textFilename string, profile interface
 	if err != nil {
 		return "", fmt.Errorf("yaml file %s does not exist", textFilename)
 	}
+	lines := strings.Split(string(b), "\n")
+	contentToKeep := []string{}
+	for _, line := range lines {
+		trimmedToCheck := strings.TrimSpace(line)
+		if strings.HasPrefix(trimmedToCheck, "# ") || strings.HasPrefix(trimmedToCheck, "##") {
```
cameronmeissner, line 168: can we just check if `strings.HasPrefix(trimmedToCheck, "#")`?

I'd also refactor to a separate func if possible

UtheMan: Some bash lines start with "#", for example `#!/bin/bash`
Can move to separate func

pkg/agent/utils.go
```diff
@@ -177,6 +177,15 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
 		panic(fmt.Sprintf("BUG: %s", err.Error()))
 	}
 	// translate the parameters.
+	lines := strings.Split(string(b), "\n")
```
cameronmeissner, line 180: prefer separate func here as well if possible

pkg/agent/baker.go
cameronmeissner: are there others that start with the same? maybe we can just exclude that?

open to either approach though, I see what you're saying...

UtheMan: Not sure honestly, but I think there are others, for example `#>` that appears in custom data

pkg/agent/utils.go
```diff
@@ -194,6 +195,33 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
 	return getBase64EncodedGzippedCustomScriptFromStr(csStr)
 }
 
+func removeComments(b []byte) []byte {
+	var contentToKeep []string
```
juan-lee, line 199: should we use a string builder instead?

pkg/agent/utils.go
```diff
@@ -194,6 +195,33 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
...
+	lines := strings.Split(string(b), "\n")
```
juan-lee, line 200: should we be using bufio ScanLines or similar?

pkg/agent/utils.go
```diff
@@ -194,6 +195,33 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
...
+	for _, line := range lines {
+		lineNoWhitespace := strings.TrimSpace(line)
+		if isCommentInBeginningOfLine(lineNoWhitespace) {
+			// ignore entire line that is a comment
+			continue
+		}
+		lastHashIndex := strings.LastIndex(lineNoWhitespace, "#")
```
juan-lee, line 207: do we need to do this or will Split figure this out for us?

e2e/e2e-local.sh
```diff
@@ -11,6 +11,9 @@ set -euxo pipefail
 export SUBSCRIPTION_ID
 export LOCATION
 export AZURE_TENANT_ID
+export SCENARIOS_TO_RUN="ubuntu2204"
```
UtheMan, line 14: oops, need to remove that

pkg/agent/utils.go
cameronmeissner: this will add _some_ latency to the call from RP to get node bootstrapping right?

I'd imagine the difference in terms of latency between the 2 methods is most likely negligible however

pkg/agent/utils.go
```diff
@@ -194,6 +195,40 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
...
+		if end > len(str) || start > end {
+			return ""
+		}
+		return str[start:end]
+	}
+	return getSlice(lastHashIndex-1, lastHashIndex+1, trimmedToCheck) != "<#" && getSlice(lastHashIndex, lastHashIndex+2, trimmedToCheck) == "# "
+}
+
+func isCommentInBeginningOfLine(trimmedToCheck string) bool {
+	return strings.HasPrefix(trimmedToCheck, "# ") || strings.HasPrefix(trimmedToCheck, "##")
```
cameronmeissner, line 229: woudn't this miss comments that start with '#' and immediately proceed with a non-space character?

e.g. "#foo bar baz"

pkg/agent/utils.go
```diff
@@ -194,6 +195,40 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
...
+			line = line[:lastHashIndex]
+		}
+		contentWithoutComments = append(contentWithoutComments, line)
+	}
+	return []byte(strings.Join(contentWithoutComments, "\n"))
+}
+
+// Trying to avoid using a regex. There are certain patterns we ignore just to be on the safe side. This is enough to get rid of most of the obvious comments.
+func isCommentAtTheEndOfLine(lastHashIndex int, trimmedToCheck string) bool {
+	getSlice := func(start, end int, str string) string {
```
cameronmeissner, line 219: why make this func anonymous?

pkg/agent/utils.go
```diff
@@ -194,6 +195,40 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
...
+func removeComments(b []byte) []byte {
+	var contentWithoutComments []string
+	lines := strings.Split(string(b), "\n")
+	for _, line := range lines {
+		lineNoWhitespace := strings.TrimSpace(line)
+		if isCommentInBeginningOfLine(lineNoWhitespace) {
+			// ignore entire line that is a comment
+			continue
+		}
+		lastHashIndex := strings.LastIndex(line, "#")
```
cameronmeissner, line 207: for readability, I'd have isCommentAtEndOfLine just take in the whole line, the same way isCommentInBeginningOfLine does

that way we can just do:

```
if !startsWithComment(line) && !endsWithComment(line) {
content = append(content, line)
}
```


i'd also rename:
- IsCommentInBeginningOfLine -> startsWithComment
- isCommentAtTheEndOfLine -> endsWithComment

pkg/agent/utils.go
```diff
@@ -194,6 +195,40 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
...
+		}
+		contentWithoutComments = append(contentWithoutComments, line)
+	}
+	return []byte(strings.Join(contentWithoutComments, "\n"))
+}
+
+// Trying to avoid using a regex. There are certain patterns we ignore just to be on the safe side. This is enough to get rid of most of the obvious comments.
+func isCommentAtTheEndOfLine(lastHashIndex int, trimmedToCheck string) bool {
+	getSlice := func(start, end int, str string) string {
+		if end > len(str) || start > end {
```
cameronmeissner, line 220: does the index-range operator not already take care of this?

pkg/agent/utils.go
```diff
@@ -194,6 +195,40 @@ func getBase64EncodedGzippedCustomScript(csFilename string, config *datamodel.No
 	return getBase64EncodedGzippedCustomScriptFromStr(csStr)
 }
 
+func removeComments(b []byte) []byte {
```
cameronmeissner, line 198: I know it's a pain, but we should probably write some basic unit tests for this

pkg/agent/utils.go
UtheMan: it would, but i wasn't aiming to remove every possible comment, this was supposed to be a quick fix :'( but if we want to make it more "thorough" we could try. At that point we might need a regex though?

pkg/agent/utils.go
UtheMan: We could only be sure once we run a benchmark, as my friend Bill K says

pkg/agent/utils.go
UtheMan: just to avoid letting people using it elsewhere, I just didn't feel like it's useful enough elsewhere to make it into a regular function.

pkg/agent/utils.go
UtheMan: true, will add some

pkg/agent/utils.go
UtheMan: This function just checks if specific patterns appear in the line, to access it "safely" i added this getSlice function. So I cannot easily replicate this with for-range. Or did you mean something else?

cameronmeissner: oh, yeah I guess this operator isn't smart enough, just tested locally

pkg/agent/utils.go
cameronmeissner: fair enough

pkg/agent/utils_test.go
```diff
@@ -686,3 +686,30 @@ var _ = Describe("Assert datamodel.CSEStatus can be used to parse output JSON",
 		Expect(err).NotTo(BeNil())
 	})
 })
+
+var _ = Describe("Test removeComments", func() {
+	It("Should leave lines without comments unchanged", func() {
+		input := []byte("#!/bin/bash\n\nCC_SERVICE_IN_TMP=/opt/azure/containers/cc-proxy.service.in\nCC_SOCKET_IN_TMP=/opt/azure/containers/cc-proxy.socket.in\nCNI_CONFIG_DIR=\"/etc/cni/net.d\"\nCNI_BIN_DIR=\"/opt/cni/bin\"\nCNI_DOWNLOADS_DIR=\"/opt/cni/downloads\"\nCRICTL_DOWNLOAD_DIR=\"/opt/crictl/downloads\"\nCRICTL_BIN_DIR=\"/usr/local/bin\"\nCONTAINERD_DOWNLOADS_DIR=\"/opt/containerd/downloads\"\nRUNC_DOWNLOADS_DIR=\"/opt/runc/downloads\"\nK8S_DOWNLOADS_DIR=\"/opt/kubernetes/downloads\"\nUBUNTU_RELEASE=$(lsb_release -r -s)\nSECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_DOWNLOAD_DIR=\"/opt/azure/tlsbootstrap\"\nSECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_VERSION=\"v0.1.0-alpha.2\"\nTELEPORTD_PLUGIN_DOWNLOAD_DIR=\"/opt/teleportd/downloads\"\nTELEPORTD_PLUGIN_BIN_DIR=\"/usr/local/bin\"\nCONTAINERD_WASM_VERSIONS=\"v0.3.0 v0.5.1 v0.8.0\"\nMANIFEST_FILEPATH=\"/opt/azure/manifest.json\"\nMAN_DB_AUTO_UPDATE_FLAG_FILEPATH=\"/var/lib/man-db/auto-update\"\nCURL_OUTPUT=/tmp/curl_verbose.out")
```
cameronmeissner, line 692: what if we used the testdata files for this instead? that way we'd get full coverage theoretically

UtheMan: will try doing that

pkg/agent/utils.go
UtheMan: We're gonna have 2 problems now

cameronmeissner (approved):

juan-lee (approved):

PR #3907, opened 2023-12-06
fix: temporarily disable abe2e garbage collection, tune path triggers

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/e2e.yaml
+++ b/.pipelines/e2e.yaml
@@ -13,6 +13,8 @@ pr:
     - pkg/agent
     exclude:
     - e2e/windows
+    - pkg/agent/datamodel/sig_config*.go # SIG config changes
+    - pkg/agent/datamodel/*.json # SIG version changes
     
 variables:
 - group: ab-e2e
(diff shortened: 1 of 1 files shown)


.pipelines/e2e.yaml
```diff
@@ -41,19 +43,5 @@ jobs:
...
-      vmssResourceIds=""
-      for vmssModel in e2e/scenario-logs/*/vmssId.txt; do
-        resourceId=$(cat ${vmssModel})
-        vmssResourceIds="${vmssResourceIds} ${resourceId}"
-      done
-
-      if [ -n "${vmssResourceIds// }" ]; then
-        az resource delete --ids ${vmssResourceIds}
-      fi
-    displayName: ensure vmss deletion
```
cameronmeissner, line 56: pretty sure this isn't actually doing anything anyways...

.pipelines/e2e.yaml
```diff
@@ -13,6 +13,8 @@ pr:
     - pkg/agent
     exclude:
     - e2e/windows
+    - pkg/agent/datamodel/sig_config*.go # SIG config changes
```
anujmaheshwari1, line 16: might also want to add sig_config_test.go here

cameronmeissner: wildcard should take care of it right?

anujmaheshwari1 (approved):

PR #3911, opened 2023-12-07
chore: improve windows backfill cleanup script

junjiezhang1997:
**What type of PR is this?**
/kind feature

1. Add "wait --deleted" for delete operations.
2. Add clean up for test sig image versions.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/backfill-cleanup.sh
+++ b/vhdbuilder/packer/backfill-cleanup.sh
@@ -1,12 +1,12 @@
 #!/bin/bash -x
 
+if [[ "${MODE}" != "windowsVhdMode" ]]; then
+  exit 0
+fi
+
 if [[ ${SUBSCRIPTION_ID} == ${PROD_SUBSCRIPTION_ID} ]]; then
   echo "Shouldn't do backfill clean up in production subscription."
   exit 1
-fi
-
-if [[ "${MODE}" != "windowsVhdMode" ]]; then
-  exit 0
 fi
 
 EXPIRATION_IN_HOURS=168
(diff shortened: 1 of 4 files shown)


vhdbuilder/packer/backfill-cleanup.sh
```diff
@@ -1,18 +1,22 @@
 #!/bin/bash -x
 
+if [[ ${SUBSCRIPTION_ID} == ${PROD_SUBSCRIPTION_ID} ]]; then
```
AbelHu, line 3: Does this expect that Linux does not set `PROD_SUBSCRIPTION_ID` so PROD_SUBSCRIPTION_ID is empty and if is false?
Suggest to check `windowsVhdMode` before others.

vhdbuilder/packer/cleanup.sh
```diff
@@ -69,6 +69,7 @@ if [[ "${MODE}" == "windowsVhdMode" && "$SIG_FOR_PRODUCTION" == "True" ]]; then
         az sig image-version show -e $version -i ${SIG_IMAGE_NAME} -r ${SIG_GALLERY_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id
         echo "Deleting sig image-version ${version} ${SIG_IMAGE_NAME} from gallery ${SIG_GALLERY_NAME} rg ${AZURE_RESOURCE_GROUP_NAME}"
         az sig image-version delete -e $version -i ${SIG_IMAGE_NAME} -r ${SIG_GALLERY_NAME} -g ${AZURE_RESOURCE_GROUP_NAME}
+        az sig image-version wait --deleted --timeout 1800 -e $version -i ${SIG_IMAGE_NAME} -r ${SIG_GALLERY_NAME} -g ${AZURE_RESOURCE_GROUP_NAME}
```
AbelHu, line 72: This will increase the build duration. How long will this increase?

vhdbuilder/packer/backfill-cleanup.sh
junjiezhang1997: Currently backfill cleanup is only used by Windows pipeline. I could check `windowsVhdMode` to avoid such issue.

vhdbuilder/packer/cleanup.sh
junjiezhang1997: In my test, the cleanup step only needs several minutes. This step is to improve the successful rate of deleting sig images to reduce the resources that are not cleaned up.

junjiezhang1997: Remove this part to save vhd build time.

AbelHu (approved):
LGTM. Thanks for the PR.

PR #3914, opened 2023-12-09
feat: introduce replication factor in publishing info

anujmaheshwari1:
**What type of PR is this?**
This PR introduces a replication factor(higher the factor, lower the number of replicas) for all SKUs. This information will be consumed in the release scripts to determine the exact number of replicas needed for a particular SKU.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/generate-vhd-publishing-info.sh
```diff
@@ -87,7 +110,8 @@ if [ "${OS_NAME,,}" == "linux" ]; then
     "offer_name": "$OFFER_NAME",
     "hyperv_generation": "${HYPERV_GENERATION}",
     "image_architecture": "${IMAGE_ARCH}",
-    "image_version": "${IMAGE_VERSION}"
+    "image_version": "${IMAGE_VERSION}",
+    "replication_inverse": "${REPLICATION_INVERSE}"
```
cameronmeissner, line 114: would it be possible to compute this at release-time instead?

vhdbuilder/packer/generate-vhd-publishing-info.sh
```diff
@@ -11,6 +11,29 @@ required_env_vars=(
...
 )
 
+# Higher the replication_inverse, lower is the usage and number of replicas
+set -x
+REPLICATION_INVERSE=1
+feature_set=("fips" "gpu" "arm64" "cvm" "tl" "1804")
+if [ "${OFFER_NAME,,}" != "ubuntu" ]; then
+    # Since Ubuntu is our most used SKU as compared to Windows/Mariner/AzLinux, we dont need the same number of replicas for all.
+    # Starting off with half replicas.
+    REPLICATION_INVERSE=$((REPLICATION_INVERSE * 2))
```
cameronmeissner, line 21: yeah, will need confirmation from @AbelHu on this

vhdbuilder/packer/generate-vhd-publishing-info.sh
```diff
@@ -11,6 +11,29 @@ required_env_vars=(
...
+fi
+
+if [ "${HYPERV_GENERATION,,}" == "v1" ]; then
+    # Gen2 SKUs are more used as compared to Gen1 SKUs, therefore Gen1 SKUs do not warrant the same number of replicas
+    REPLICATION_INVERSE=$((REPLICATION_INVERSE * 2))
+fi
+
+for feature in "${feature_set[@]}"; do
```
cameronmeissner, line 29: I see... I guess we wouldn't have this in context of release scripts

cameronmeissner: it looks like this only actually works for kata/gpu SKUs, could maybe just manually check *"gpu"* and *"kata"* and then move all this over?

granted, I'm leaning this way just to avoid breaking up the change between both repos and to minimize what we include in publishing info - curious what your thoughts are on this

anujmaheshwari1: this also increases the inverse for arm64, TL, CVM

feature_set=("fips" "gpu" "arm64" "cvm" "tl" "1804")
declared on L17, you actually gave me a good one I missed, kata.

vhdbuilder/packer/generate-vhd-publishing-info.sh
anujmaheshwari1: yep, will take this PR to Henry, Charlie, and Abel

vhdbuilder/packer/generate-vhd-publishing-info.sh
anujmaheshwari1: it is possible, yes. I dont have a strong preference of doing it here or there, I just thought its simpler to have the calc done here and just use the inverse from publishing info in RP.

But it is doable in RP as well, sure.

anujmaheshwari1: one reason I did do it here is testability. its easier to build VHDs and see the replication inverse calculated here instead of behind ev2 logs :)

vhdbuilder/packer/generate-vhd-publishing-info.sh
cameronmeissner: oh, I completely missed that lol, thought this was using the value of the FEATURE_FLAGS var from the environment - but yeah in that case should definitely be doable at release time, just something to consider

vhdbuilder/packer/generate-vhd-publishing-info.sh
cameronmeissner: this is true

cameronmeissner: doing it here is fine, lets just make sure we have good reasons we've all agreed upon when adding stuff to publishing info moving forward

juan-lee (approved):

cameronmeissner (approved):

hbeberman (approved):

vhdbuilder/packer/generate-vhd-publishing-info.sh
anujmaheshwari1: got approval from Henry and Charlie, this PR just adds to Windows publishing info but doesnt consume it, Windows team can pick it if they want to alter replicas.

junjiezhang1997 (approved):

anujmaheshwari1:
https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=84288353&view=results

Successful test run, cross checked that replication inverses are correct.

PR #3915, opened 2023-12-11
fix: use containerd v1.7.9 when building 23H2 vhd

junjiezhang1997:
**What type of PR is this?**
/kind fix

1. use containerd v1.7.9 when building 23H2 vhd.
2. use "unknown" instead of throwing error when windows version is not set.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -140,7 +140,7 @@ steps:
         WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2019_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
       elif [[ ${{ parameters.artifactName }} =~ "2022" ]]; then
         WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2022_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
-      elif [[ ${{ parameters.artfiactName }} =~ "23H2" ]]; then
+      elif [[ ${{ parameters.artifactName }} =~ "23H2" ]]; then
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -236,6 +236,10 @@ function Install-ContainerD {
     # and the containerd to managed customer containers after provisioning the vm is not necessary
     # the one used here, considering containerd version/package is configurable, and the first one
     # is expected to override the later one
+    $buildNumber = (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").CurrentBuild
```
AbelHu, line 239: Would it not work if v1.6.21 is installed inside 23h2 vhd and v1.7.9 is installed in provisioning 23h2 nodes?

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -142,7 +142,7 @@ steps:
         WINDOWS_VERSION="$(cat vhdbuilder/packer/windows-image.env | grep -a "WINDOWS_2022_BASE_IMAGE_VERSION" | cut -d "=" -f 2 | cut -d "." -f 1,2)"
       else
         echo "Current distro is not supported to get image version. You need to update related code."
-        exit 1
+        WINDOWS_VERSION="unknown"
```
AbelHu, line 145: What is the expected result with this change? It will fail the build before this change

junjiezhang1997: The artifact file will have "unknown" in the file name. https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=84214453&view=artifacts&pathAsName=false&type=publishedArtifacts

I want to add the support for 23H2 when the market place images are ready. Otherwise, I'm afraid that a fake 23H2 image version could make confusion.

junjiezhang1997: Added the code to support 23H2 here and keep "exit 1" for not supported distro.

PR #3920, opened 2023-12-12
fix: check the path of nvidiaInstallLogFolder

ShiqianTao:
**What type of PR is this?**
/kind bug

check the path of nvidiaInstallLogFolder, otherwise, it will panic if the node doesn't enable gpu.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


AbelHu (approved):
LGTM. Thanks for the PR.

PR #3926, opened 2023-12-13
fix: remove invalid registry keys

AbelHu:
**What type of PR is this?**
/kind bug

 remove invalid registry keys

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -662,19 +662,7 @@ function Update-Registry {
         }
         Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name FwPerfImprovementChange -Value 1 -Type DWORD
 
-        Write-Log "Enable 7 fixes in 2023-11B"
-        $currentValue=(Get-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name PortExclusionChange -ErrorAction Ignore)
-        if (![string]::IsNullOrEmpty($currentValue)) {
-            Write-Log "The current value of PortExclusionChange is $currentValue"
-        }
-        Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name PortExclusionChange -Value 1 -Type DWORD
-
-        $currentValue=(Get-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name NamespaceExcludedUdpPorts -ErrorAction Ignore)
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -669,12 +669,6 @@ function Update-Registry {
         }
         Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name PortExclusionChange -Value 1 -Type DWORD
```
ShiqianTao, line 670: You may want to remove it `PortExclusionChange`

ShiqianTao (approved):

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -662,19 +662,7 @@ function Update-Registry {
...
-        }
-        Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name PortExclusionChange -Value 1 -Type DWORD
-
-        $currentValue=(Get-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name NamespaceExcludedUdpPorts -ErrorAction Ignore)
-        if (![string]::IsNullOrEmpty($currentValue)) {
-            Write-Log "The current value of NamespaceExcludedUdpPorts is $currentValue"
-        }
-        Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name NamespaceExcludedUdpPorts -Value 1 -Type DWORD
-
+        Write-Log "Enable 6 fixes in 2023-11B"
```
ShiqianTao, line 665: 5?

ShiqianTao (changes requested):

vhdbuilder/packer/configure-windows-vhd.ps1
mzhang439: wondering if you want to just Disable the fix (set the reg PortExclusionChange to 0 and still keep the NamespaceExcludedUdpPorts reg as it is) instead of removing it. NamespaceExcludedUdpPorts reg will not be active with PortExclusionChange== 0.

That case, you don't need to add the whole block back but only to flip it from 0 to 1 if the fix is needed later.
In addition, for those cx who really need the fix, they just need to set PortExclusionChange== 1

AbelHu: 4. Those two keys are for one bug. The old value 7 is also wrong.

ShiqianTao (approved):

PR #3929, opened 2023-12-14
feat: add custom linux os config to contract

anujmaheshwari1:
**What type of PR is this?**
This PR adds custom linux os config to the bootstrap contract.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/customlinuxosconfig.proto
```diff
@@ -0,0 +1,55 @@
...
+    optional int32 FsFileMax = 22;
+    optional int32 FsAioMaxNr = 23;
+    optional int32 FsNrOpen = 24;
+    optional int32 KernelThreadsMax = 25;
+    optional int32 VMMaxMapCount = 26;
+    optional int32 VMSwappiness = 27;
+    optional int32 VMVfsCachePressure = 28;
+  }
+  
+  message UlimitConfig {
```
UtheMan, line 52: This is currently not used in aks-rp, however long term plan is to eventually use it and expand it with a lot more params (possibly allow users to specify arbitrary ulimits). Not sure if its worth it to include it at this point?

pkg/proto/nbcontract/v1/customlinuxosconfig.proto
```diff
@@ -0,0 +1,55 @@
...
+    optional string transparent_defrag = 3;
+  
+    // Sysctl settings for Linux agent nodes
+    optional SysctlConfig sysctl_config = 4;
+  
+    // Ulimit settings for Linux agent nodes
+    optional UlimitConfig ulimit_config = 5;
+  }
+
+message SysctlConfig {
```
UtheMan, line 21: Similar to ulimits, the list of config options we provide might change in the future. There was some talk about allowing people to pass arbitrary fields, but no decision was made on that yet.

UtheMan:
Other than my comments about (very likely) future changes to available config options, I think LGTM.

UtheMan (approved):

pkg/proto/nbcontract/v1/customlinuxosconfig.proto
tallaxes, line 21: As discussed out-of-band, let's capture (maybe in the contract design doc?) the trade-offs between explicit fields for each value vs. more generic approach, for when we revisit this (and other relevant sets of fields.)

tallaxes (approved):

cameronmeissner (approved):

PR #3934, opened 2023-12-18
feat: add retrying to e2e tests

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

--- PASS: Test_All (22.54s)
    --- PASS: Test_All/azurelinuxv2-custom-sysctls (121.00s)
    --- PASS: Test_All/marinerv2-custom-sysctls (121.33s)
    --- PASS: Test_All/ubuntu2204-gpu-nodriver (149.20s)
    --- PASS: Test_All/marinerv2-azurecni (149.83s)
    --- PASS: Test_All/ubuntu2204 (150.58s)
    --- PASS: Test_All/ubuntu2204-custom-ca-trust (151.67s)
    --- PASS: Test_All/marinerv2-gpu (208.61s)
    --- PASS: Test_All/azurelinuxv2-gpu (208.72s)
    --- PASS: Test_All/azurelinuxv2 (118.50s)
    --- PASS: Test_All/ubuntu2204-custom-sysctls (149.35s)
    --- PASS: Test_All/azurelinuxv2-arm64 (148.18s)
    --- PASS: Test_All/ubuntu1804-gpu (323.90s)
    --- PASS: Test_All/ubuntu2204-artifact-streaming (177.61s)
    --- PASS: Test_All/marinerv2 (117.43s)
    --- PASS: Test_All/azurelinuxv2-gpu-azurecni (209.91s)
    --- PASS: Test_All/ubuntu1804 (180.73s)
    --- PASS: Test_All/ubuntu2204-arm64 (148.70s)
    --- PASS: Test_All/marinerv2-arm64 (143.99s)
    --- PASS: Test_All/ubuntu2204-gpu-a100 (295.07s)
    --- PASS: Test_All/azurelinuxv2-wasm (517.96s)
    --- PASS: Test_All/azurelinuxv2-azurecni (117.46s)
    --- PASS: Test_All/marinerv2-wasm (467.22s)
    --- PASS: Test_All/marinerv2-gpu-azurecni (209.33s)
    --- PASS: Test_All/ubuntu2204-gpu-ncv3 (325.57s)
    --- PASS: Test_All/ubuntu2204-wasm (515.88s)
    --- PASS: Test_All/ubuntu1804-azurecni (208.75s)
    --- PASS: Test_All/ubuntu2204-gpu-a10 (299.95s)
    --- PASS: Test_All/ubuntu1804-gpu-azurecni (331.61s)
PASS
ok      github.com/Azure/agentbakere2e  774.750s

**Release note**:
```
none
```


e2e/pollers.go
```diff
@@ -230,3 +236,49 @@ func waitUntilPodDeleted(ctx context.Context, kube *kubeclient, podName string)
...
+		}
+		break
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error on VMSS operation, all retries failed %q: %v", vmssName, err)
+	}
+
+	var vmssResp T
+	for i := 0; i < vmssRetryAttempts; i++ {
+		vmssResp, err = poller.PollUntilDone(ctx, pollerOpts)
```
cameronmeissner, line 269: I don't think retrying the polling here makes sense right? I believe the underlying poller returned by the ARM API will poll correctly until a terminal state has been reached - we need to check that terminal state and retry the actual operation itself if that terminal state isn't a success

cameronmeissner: now that I think about it I don't believe we actually need to retry the actual invocation of the op to get the poller object - the ARM client retry config should take care of that. We really just care about retrying in the case where the terminal state reported by the poller obj is non-success. So we'd wanna do something like:


```suggestion
for retry in retries {
  poller, err := op()
  if err != nil {
    fail
  }

  resp, err := poller.PollUntilDone()
  if err != nil {
    if errors.As(...) && requestError.ServiceError != nil {
        maybe check the code
        sleep()
        continue
    }
    fail
  }
  
  return success
}

```

e2e/pollers.go
```diff
@@ -230,3 +236,49 @@ func waitUntilPodDeleted(ctx context.Context, kube *kubeclient, podName string)
...
+
+type Poller[T any] interface {
+	PollUntilDone(ctx context.Context, options *runtime.PollUntilDoneOptions) (T, error)
+}
+
+func pollVMSSOperation[T any](ctx context.Context, vmssName string, pollerOpts *runtime.PollUntilDoneOptions, vmssOperation func() (Poller[T], error)) (*T, error) {
+	var requestError azure.RequestError
+	var err error
+	var poller Poller[T]
+	for i := 0; i < vmssRetryAttempts; i++ {
```
cameronmeissner, line 248: can we just use the `wait.PollImmediateWithContext` paradigm here instead like we do for all the others?

cameronmeissner: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediateWithContext

e2e/pollers.go
```diff
@@ -230,3 +235,36 @@ func waitUntilPodDeleted(ctx context.Context, kube *kubeclient, podName string)
...
+}
+
+func pollVMSSOperation[T any](ctx context.Context, vmssName string, pollerOpts *runtime.PollUntilDoneOptions, vmssOperation func() (Poller[T], error)) (*T, error) {
+	var requestError azure.RequestError
+	var vmssResp T
+
+	pollErr := wait.PollImmediateWithContext(ctx, vmssOperationPollInterval, vmssOperationPollingTimeout, func(ctx context.Context) (bool, error) {
+		poller, err := vmssOperation()
+		if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
```
cameronmeissner, line 250: we should only need to check the error details as a result of the `PollUntilDone` below I believe since we added the new config to the arm client

also, in the case where it's a service error, we should be checking that requestErrror.ServiceError.Code == <retriable code>, and in that case:
`return false, nil` to have the operation be retried

cameronmeissner: off the bat, we probably wanna retry:

`AllocationFailed`
`InternalExecutionError`
`StorageFailure/SocketException`

e2e/pollers.go
```diff
@@ -230,3 +235,44 @@ func waitUntilPodDeleted(ctx context.Context, kube *kubeclient, podName string)
...
+	var vmssResp T
+	var requestError azure.RequestError
+
+	poller, err := vmssOperation()
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return nil, fmt.Errorf("unable to complete VMSS operation in allotted time of %s: %w", createVMSSPollingTimeout.String(), err)
+		}
+		return nil, err
+	}
```
AlisonB319, line 253: I took 247-253 outside of the PollImmediateWithContext, since there is nothing that needs to retry here. I am also a supporter of only having the necessary bits inside "loops" for future readability. But if you think there would be a case where _poller, err := vmssOperation()_
needs a retry, I can put it back under PollImmediateWithContext

AlisonB319: This is actually not correct. The poller could fail but then the operation needs to be re-called in order for retry to work. This has been reverted

e2e/vmss.go
```diff
@@ -82,27 +79,19 @@ func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName str
...
-		*opts.clusterConfig.cluster.Properties.NodeResourceGroup,
-		vmssName,
-		model,
-		nil,
-	)
-	if err != nil {
-		return nil, err
+	vmssOperation := func() (Poller[armcompute.VirtualMachineScaleSetsClientCreateOrUpdateResponse], error) {
+		return opts.cloud.vmssClient.BeginCreateOrUpdate(
+			createVMSSCtx,
```
cameronmeissner, line 84: use base ctx

e2e/vmss.go
```diff
@@ -82,27 +79,19 @@ func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName str
...
+			vmssName,
+			model,
+			nil,
+		)
 	}
-
-	vmssResp, err := pollerResp.PollUntilDone(createVMSSCtx, &runtime.PollUntilDoneOptions{
-		Frequency: createVMSSPollingInterval,
-	})
+	vmssResp, err := pollVMSSOperation(ctx, vmssName, &runtime.PollUntilDoneOptions{Frequency: createVMSSPollingInterval}, vmssOperation)
```
cameronmeissner, line 91: use createVMSSCtx

cameronmeissner (approved):

PR #3939, opened 2023-12-26
feat: cache and use Linux private packages for kubctl and kubelet.

daxakp:
**What type of PR is this?**
feature

This PR downloads and saves private packages for kubelet and kubectl. This packages will be used at run-time when needed. 
The PR adds two build-time variables as below:

Build Variables:

    name: PRIVATE_PACKAGES_URL
    value: comma separated list of packages, for example: https://private-acs-mirror.azureedge.net/kubernetes/v1.25.6-hotfix.20230612/binaries/kubernetes-node-linux-amd64.tar.gz,https://private-acs-mirror.azureedge.net/kubernetes/v1.28.0/binaries/kubernetes-node-linux-amd64.tar.gz
    
    name: AZURE_MSI_RESOURCE_ID
    value: MSI resource ID that is created already and has sufficient roles assigned to be able to access the above private packages from the packer VM.

- Kubernetes packages are cached on disk at: /opt/kubernetes/downloads/private-packages
   At run-time, given a Kube version, the cached version if exists is used by default.
   For example, 
   kubernetes package from https://private-acs-mirror.azureedge.net/kubernetes/v1.28.0/binaries/kubernetes-node-linux-amd64.tar.gz is saved as /opt/kubernetes/downloads/private-packages/v1.28.0.tar.gz

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests
- [X] tested upgrade from previous version

**Release note**:
```
none
```


.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -35,15 +35,15 @@ parameters:
   type: boolean
   default: false
 - name: buildMarinerV2gen1fips
-  displayName: Build MarinerV2 Gen1 FIPS
```
cameronmeissner, line 38: stray spacing changes?

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -106,4 +106,3 @@ stages:
         - template: ./templates/.builder-release-template.yaml
           parameters:
             artifactName: marinerv2-gen2
-
```
cameronmeissner, line 109: lets try to avoid spacing/newline changes to minimize the size of the change-set

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -105,6 +105,9 @@ steps:
         -e SGX_INSTALL=${SGX_INSTALL} \
         -e ENABLE_CGROUPV2=${ENABLE_CGROUPV2} \
         -e IMAGE_VERSION=${IMAGE_VERSION} \
+        -e PRIVATE_PACKAGES_URL="${PRIVATE_PACKAGES_URL}" \
```
cameronmeissner, line 108: nit: rename to `PRIVATE_PACKAGE_URLS`?

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -180,14 +189,24 @@ steps:
...
       -e ARM64_OS_DISK_SNAPSHOT_NAME=${ARM64_OS_DISK_SNAPSHOT_NAME} \
+      -e PRIVATE_PACKAGES_URL="${PRIVATE_PACKAGES_URL}" \
+      -e PRIVATE_KUBE_PROXY_IMAGES="${PRIVATE_KUBE_PROXY_IMAGES}" \
+      -e LINUX_MSI_RESOURCE_ID=${AZURE_MSI_RESOURCE_ID} \
       ${CONTAINER_IMAGE} make -f packer.mk cleanup
     displayName: Clean Up Packer Generated Resources
     condition: always()
   - bash: |
         echo MODE=$(MODE) && \
+        echo "container image: ${CONTAINER_IMAGE}" && \
```
cameronmeissner, line 200: can we remove this debugging code?

e2e/scenario/base_vhd_catalog.json
```diff
@@ -13,6 +13,10 @@
         "gen2containerd": {
             "artifactName": "2204-gen2-containerd",
             "resourceId": "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/2204Gen2/versions/1.1699020621.21398"
+        },
+        "gen2containerdprivatekubepkg": {
```
cameronmeissner, line 17: we shouldn't need a new VHD here, would just reuse ubuntu 2204 and supply the custom package URLs

e2e/scenario/scenario_ubuntu1804.go
```diff
@@ -9,6 +13,11 @@ func (t *Template) ubuntu1804() *Scenario {
 			ClusterSelector: NetworkPluginKubenetSelector,
 			ClusterMutator:  NetworkPluginKubenetMutator,
 			VHDSelector:     t.Ubuntu1804Gen2Containerd,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
```
cameronmeissner, line 16: this scenario should just use what's already in the NBC, any new functionality should be tested separately in a new scenario(s)

cameronmeissner: granted that template is pretty out of date :)

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -240,9 +241,32 @@ extractKubeBinaries() {
...
+    CACHED_PKG="${K8S_CACHE_DIR}/${K8S_TGZ_TMP}"
+
+    if [[ -f "${CACHED_PKG}" ]]; then
+        echo "cached package ${CACHED_PKG} is found, will use that"
+    else
+        echo "cached package ${CACHED_PKG} not found"
+        return 1
+    fi
+
+    retrycmd_get_tarball 120 5 "${CACHED_PKG}" ${KUBE_BINARY_URL} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
```
cameronmeissner, line 258: should we have a new exit code for failures when downloading private kube binaries?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -151,7 +150,7 @@ if [[ "${GPU_NODE}" = true ]] && [[ "${skip_nvidia_driver_install}" != "true" ]]
 [Service]
 Environment="MIG_STRATEGY=--mig-strategy single"
 ExecStart=
-ExecStart=/usr/local/nvidia/bin/nvidia-device-plugin $MIG_STRATEGY    
+ExecStart=/usr/local/nvidia/bin/nvidia-device-plugin $MIG_STRATEGY
```
cameronmeissner, line 153: be careful about trimming spaces here

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -104,7 +104,6 @@ logs_to_events "AKS.CSE.configureAdminUser" configureAdminUser
 
 VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
 if [ -f $VHD_LOGS_FILEPATH ]; then
-    echo "detected golden image pre-install"
```
cameronmeissner, line 107: why remove this?

self-contained/bootstrap_cmd.sh
```diff
@@ -16,6 +16,7 @@ KUBERNETES_VERSION={{GetParameter "kubernetesVersion"}}
 HYPERKUBE_URL={{GetParameter "kubernetesHyperkubeSpec"}}
 KUBE_BINARY_URL={{GetParameter "kubeBinaryURL"}}
 CUSTOM_KUBE_BINARY_URL={{GetParameter "customKubeBinaryURL"}}
+PRIVATE_KUBE_BINARY_URL="{{GetLinuxPrivatePackageURL}}"
```
cameronmeissner, line 19: @anujmaheshwari1 to confirm whether this should be part of the bootstrap contract (assuming it has to be)

vhdbuilder/packer/init-variables.sh
```diff
@@ -211,6 +211,15 @@ if [ -n "${WINDOWS_MSI_RESOURCE_STRING}" ]; then
 	windows_msi_resource_strings+=(${WINDOWS_MSI_RESOURCE_STRING})
 fi
 
+linux_msi_resource_ids=()
```
cameronmeissner, line 214: does this need to be an array? I think it would only ever hold one value for linux

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -395,11 +396,96 @@ for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
...
+  else
+    # save kube pkg with version number from the url path, this convention is used to find the cached package at run-time
+    K8S_TGZ_NAME=$(echo "$KUBE_PRIVATE_BINARY_URL" | grep -o -P '(?<=\/kubernetes\/).*(?=\/binaries\/)').tar.gz
+  fi
+
+  # use azcopy with MSI instead of curl to download packages
+  getAzCopyCurrentPath
+
+  export AZCOPY_AUTO_LOGIN_TYPE="MSI"
+  export AZCOPY_MSI_RESOURCE_STRING="$LINUX_MSI_RESOURCE_IDS"
```
cameronmeissner, line 464: I wonder if we actually need to specify the linux MSI here, I'd think azcopy would be smart enough to infer the resource ID of the MSI that's actually attached to the VM

cameronmeissner: ah, I guess we do since this should be a user-assigned MSI...

https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-authorize-azure-active-directory#authorize-by-using-a-user-assigned-managed-identity

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -395,11 +396,96 @@ for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
...
   pullContainerImage ${cliTool} ${CONTAINER_IMAGE}
   ctr --namespace k8s.io run --rm ${CONTAINER_IMAGE} checkTask /bin/sh -c "iptables --version" | grep -v nf_tables && echo "kube-proxy contains no nf_tables"
-  
+
   # shellcheck disable=SC2181
   echo "  - ${CONTAINER_IMAGE}" >>${VHD_LOGS_FILEPATH}
 done
 
+# download and setup azcopy to use to download private packages with MSI auth
+getAzCopyCurrentPath() {
```
cameronmeissner, line 405: probably better to move these funcs to a separate file, such as vhdbuilder/packer/scripts/tools_installs.sh

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -395,11 +396,96 @@ for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
...
+
+  # use azcopy with MSI instead of curl to download packages
+  getAzCopyCurrentPath
+
+  export AZCOPY_AUTO_LOGIN_TYPE="MSI"
+  export AZCOPY_MSI_RESOURCE_STRING="$LINUX_MSI_RESOURCE_IDS"
+
+  cachedpkg="${K8S_CACHE_DIR}/${K8S_TGZ_NAME}"
+  echo "download private package ${KUBE_PRIVATE_BINARY_URL} and store as ${cachedpkg}"
+  if ! ./azcopy copy "${KUBE_PRIVATE_BINARY_URL}" "${cachedpkg}"; then
```
cameronmeissner, line 468: I guess I might be a bit confused: during the VHD build we install the private packages from the private package URL, if supplied, using azcopy. However, at runtime during CSE, if RP provides a private package URL, we download that using curl if the binaries aren't already cached. We couldn't we just avoid azcopy during the build process and use curl like we do at runtime?

parts/linux/cloud-init/artifacts/cse_main.sh
daxakp: ⁠did not get this comment, don't see any concatenation/append to the command, so removing space should be ok? going to undo changes for spaces/trims but just trying to understand the issue.

.pipelines/templates/.builder-release-template.yaml
daxakp: Although i don't have a strong preference and private_packages_urls makes more sense to me, decided the the current private_packages_url to keep it consistent with current Windows variable naming for the same.

vhdbuilder/packer/init-variables.sh
daxakp: It created an issue for windows when it was not an array when MSI was not specified, i haven't come across the same for Linux but no harm in keeping it as an array - consistent and expadable?

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: clarified offline, we should only ever actually be downloading/copying stuff during build-time - CSE side simply checks to see if the VHD _should_ have private packages cached, and if not would fail provisioning

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: I think it is no different than any other package and for internal only usage once in a while - should be obvious when failed? Do we track it codes/failures via any other automation?

vhdbuilder/packer/install-dependencies.sh
daxakp: *That is how it is implemented, to download packages only at build time. But instead of failing to provision when the packages are not cached, will use the default for that version.

e2e/scenario/base_vhd_catalog.json
daxakp: Using the pre-built image here to test provisioning with VHDs with private packages.

self-contained/bootstrap_cmd.sh
daxakp: added back, thanks

parts/linux/cloud-init/artifacts/cse_main.sh
daxakp: added back, thanks

vhdbuilder/packer/install-dependencies.sh
daxakp: done, thanks

e2e/scenario/base_vhd_catalog.json
```diff
@@ -13,6 +13,10 @@
...
+            "artifactName": "2204-gen2-containerd",
```
cameronmeissner, line 18: we can remove the artifact name here since we'll just opt towards using the pinned version for this scenario in all cases

e2e/template.go
```diff
@@ -257,6 +257,7 @@ func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
 			PodInfraContainerImageURL: "mcr.microsoft.com/oss/kubernetes/pause:3.6",
 			HyperkubeImageURL:         "mcr.microsoft.com/oss/kubernetes/",
 			WindowsPackageURL:         "windowspackage",
+			LinuxPrivatePackageURL:    "linuxPrivatepackagesURL",
```
cameronmeissner, line 260: shouldn't this be empty in the base template and only set during scenarios where we actually use a VHD with private pkgs installed?

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner: we track CSE failures by exit code in our node SIG dashboard - I think it might be useful to be able to differentiate failures on normal and private cases

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -251,13 +274,18 @@ installKubeletKubectlAndKubeProxy() {
         # kube binaries used by AKS and Kubernetes upstream.
         # TODO(mainred): let's see if necessary to auto-detect the path of kubelet
         logs_to_events "AKS.CSE.installKubeletKubectlAndKubeProxy.extractKubeBinaries" extractKubeBinaries ${KUBERNETES_VERSION} ${CUSTOM_KUBE_BINARY_DOWNLOAD_URL}
+        install_default_if_missing=false
+    elif [[ ! -z ${PRIVATE_KUBE_BINARY_DOWNLOAD_URL} ]]; then
+        # remove the kubelet binaries to make sure the only binary left is from the PRIVATE_KUBE_BINARY_DOWNLOAD_URL
```
cameronmeissner, line 279: this comment is a bit misleading since we aren't actually ever going to download from PRIVATE_KUBE_BINARY_DOWNLOAD_URL according to the current implementation of extractPrivateKubeBinaries

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -251,13 +274,18 @@ installKubeletKubectlAndKubeProxy() {
...
+    fi
 
-    else
-        if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" ]]; then
-            #TODO: remove the condition check on KUBE_BINARY_URL once RP change is released
-            if (($(echo ${KUBERNETES_VERSION} | cut -d"." -f2) >= 17)) && [ -n "${KUBE_BINARY_URL}" ]; then
-                logs_to_events "AKS.CSE.installKubeletKubectlAndKubeProxy.extractKubeBinaries" extractKubeBinaries ${KUBERNETES_VERSION} ${KUBE_BINARY_URL}
-            fi
+    # if the custom url is not specified and the required kubectl/kubelet-version via private url is not installed, install using the default url/package
+    if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" && "$install_default_if_missing" == true ]]; then
```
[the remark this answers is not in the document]
cameronmeissner: e.g. `if [[ ! -f "blah" ]] && [[ "$install_default_if_missing" == true ]]`

pkg/agent/datamodel/types.go
```diff
@@ -1606,6 +1606,10 @@ type K8sComponents struct {
 	// Full path to the Windows package (windowszip) to use.
 	// For example: https://acs-mirror.azureedge.net/kubernetes/v1.17.8/windowszip/v1.17.8-1int.zip.
 	WindowsPackageURL string
+
+	// Full path to the Linux package (tar.gz) to use.
+	// For example: url=https://acs-mirror.azureedge.net/kubernetes/v1.25.6-hotfix.20230612/binaries/v1.25.6-hotfix.20230612.tar.gz
+	LinuxPrivatePackageURL string
```
cameronmeissner, line 1612: I guess I'm a bit confused why we really need this and can't just use something like a boolean flag `usePrivateKubePackages` since the URL effectively isn't actually used other than to tell CSE to first check that there are private packages cached on the VHD and use them if they're there

is it mainly because it would require changes on the RP side?

daxakp: It is because private package name will come from RP side, the kubernetes_version could be 1.28.3 for example, but the package could be v1.28.3-hotfix.20231103.tgz or v1.28.3.tgz etc.

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: updated with ERR_PRIVATE_K8S_PKG_ERR=203

e2e/template.go
daxakp: set it to empty string

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: updated, thanks

vhdbuilder/scripts/linux/tool_installs.sh
```diff
@@ -45,4 +45,28 @@ disableSystemdIptables() {
...
+}
+
+# download and setup azcopy to use to download private packages with MSI auth
+getAzCopyCurrentPath() {
+  if [[ -f ./azcopy ]]; then
+    echo "./azcopy already exists"
+  else
+    echo "get azcopy at \"${PWD}\"...start"
+    # Download and extract
+    azcopydownloadurl="https://aka.ms/downloadazcopy-v10-linux"
```
bcho, line 57: This binaries are not being published with checksum, it might be a bit risky to just download from a redirected url and use (even it's from aka.ms) as it opens risks on supply chain attack. I would suggest we download from the canonical urls for now and compare with pre-calculated sha256 checksums.

I also sent a follow up comment in azcopy's issue: https://github.com/Azure/azure-storage-azcopy/issues/2411

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -229,20 +230,45 @@ installAzureCNI() {
...
+
+    if [[ -f "${cached_pkg}" ]]; then
+        echo "cached package ${cached_pkg} is found, will use that"
+    else
+        echo "cached package ${cached_pkg} not found"
+        return 1
+    fi
+
+    # remove the current kubelet binaries before extracting new binaries from the cached package
+    rm -rf /usr/local/bin/kubelet-* /usr/local/bin/kubectl-*
```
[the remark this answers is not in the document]
daxakp: These binaries look like 
/usr/local/bin/kubelet-<version> / /usr/local/bin/kubelet-1.28.0 for example.
At build time, all the configured versions are downloaded from the public URL, at run-time, the required version is kept and all the other ones are deleted (in the caller - mv /usr/local/bin/kubelet-<version> /usr/local/bin/kubelet).
In case of private and custom url, we delete all the current version to overwrite with the cached one at run-time.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -229,20 +230,45 @@ installAzureCNI() {
...
+    local cached_pkg="${K8S_CACHE_DIR}/${k8s_tgz_tmp}"
+
+    if [[ -f "${cached_pkg}" ]]; then
+        echo "cached package ${cached_pkg} is found, will use that"
+    else
+        echo "cached package ${cached_pkg} not found"
+        return 1
+    fi
+
+    # remove the current kubelet binaries before extracting new binaries from the cached package
```
bcho, line 258: this comment should include kubectl as well?

vhdbuilder/scripts/linux/tool_installs.sh
daxakp: updated per suggestion, thank you!

daxakp: @AbelHu Abel, this will apply to the azcopy installation for windows as well, i haven't updated that code, just fyi

e2e/scenario/vhd.go
```diff
@@ -164,6 +165,11 @@ func (c *VHDCatalog) Ubuntu2204Gen2Containerd() VHD {
 	return c.Ubuntu2204.Gen2Containerd
 }
 
+// Returns the gen2containerdprivatekubepkg catalog entry.
+func (c *VHDCatalog) Ubuntu2204Gen2ContainerdPrivateKubePkg() VHD {
+	return c.Ubuntu2204.Gen2Arm64ContainerdPrivateKubePkg
```
cameronmeissner, line 170: is this actually an ARM64-based VHD?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -251,13 +278,17 @@ installKubeletKubectlAndKubeProxy() {
...
+        logs_to_events "AKS.CSE.installKubeletKubectlAndKubeProxy.extractPrivateKubeBinaries" extractPrivateKubeBinaries ${KUBERNETES_VERSION} ${PRIVATE_KUBE_BINARY_DOWNLOAD_URL}
+    fi
 
-    else
-        if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" ]]; then
-            #TODO: remove the condition check on KUBE_BINARY_URL once RP change is released
-            if (($(echo ${KUBERNETES_VERSION} | cut -d"." -f2) >= 17)) && [ -n "${KUBE_BINARY_URL}" ]; then
-                logs_to_events "AKS.CSE.installKubeletKubectlAndKubeProxy.extractKubeBinaries" extractKubeBinaries ${KUBERNETES_VERSION} ${KUBE_BINARY_URL}
-            fi
+    # if the custom url is not specified and the required kubectl/kubelet-version via private url is not installed, install using the default url/package
```
cameronmeissner, line 287: should we also explicitly check if `/usr/local/bin/kubelet-${KUBERNETES_VERSION}` is missing as well? would there ever be cases where the kubectl binary exists but the kubelet binary is missing?

vhdbuilder/packer/init-variables.sh
```diff
@@ -211,6 +211,14 @@ if [ -n "${WINDOWS_MSI_RESOURCE_STRING}" ]; then
...
+if [ -n "${LINUX_MSI_RESOURCE_ID}" ]; then
+	echo "LINUX_MSI_RESOURCE_ID is set in pipeline variables: ${LINUX_MSI_RESOURCE_ID}"
+	linux_msi_resource_ids+=(${LINUX_MSI_RESOURCE_ID})
+fi
+
+private_packages_url=""
```
cameronmeissner, line 220: nit: move declaration to L384 right before we append `PRIVATE_PACKAGES_URL` to it?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -395,11 +396,37 @@ for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
...
   ctr --namespace k8s.io run --rm ${CONTAINER_IMAGE} checkTask /bin/sh -c "iptables --version" | grep -v nf_tables && echo "kube-proxy contains no nf_tables"
-  
+
   # shellcheck disable=SC2181
   echo "  - ${CONTAINER_IMAGE}" >>${VHD_LOGS_FILEPATH}
 done
 
+# download kubernetes package from the given URL using MSI for auth for azcopy
+# if it is a kube-proxy package, extract image from the downloaded package
+cacheKubePackageFromPrivateUrl() {
```
cameronmeissner, line 406: can we move this func to tools_installs.sh as well?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -413,17 +440,29 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
   sed -i 's/After=network-online.target/After=multi-user.target/g' /lib/systemd/system/motd-news.service
 fi
 
+# use the private_packages_url to download and cache packages
+if [[ -n ${PRIVATE_PACKAGES_URL} ]]; then
+  IFS=',' read -ra PRIVATE_URLS <<< "${PRIVATE_PACKAGES_URL}"
```
cameronmeissner, line 445: we should probably reset IFS after overriding it here,

IFS_backup=$IFS; IFS=','
....
IFS=$IFS_backup

pkg/agent/datamodel/types.go
cameronmeissner: can you add a note here to mention that this URL is _only_ used at build-time outside of CSE, and this is only used as a marker to check whether private pkgs are cached at runtime?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -395,11 +396,37 @@ for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
...
+  local k8s_tgz_name
+  k8s_tgz_name=$(echo "$kube_private_binary_url" | grep -o -P '(?<=\/kubernetes\/).*(?=\/binaries\/)').tar.gz
+
+  # use azcopy with MSI instead of curl to download packages
+  getAzCopyCurrentPath
+
+  export AZCOPY_AUTO_LOGIN_TYPE="MSI"
+  export AZCOPY_MSI_RESOURCE_STRING="$LINUX_MSI_RESOURCE_IDS"
+
+  cachedpkg="${K8S_CACHE_DIR}/${k8s_tgz_name}"
```
cameronmeissner, line 423: nit: `cached_pkg`?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -413,17 +440,29 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
...
-# regular version >= v1.17.0 or hotfixes >= 20211009 has arm64 binaries. 
+# regular version >= v1.17.0 or hotfixes >= 20211009 has arm64 binaries.
 KUBE_BINARY_VERSIONS="$(jq -r .kubernetes.versions[] manifest.json)"
 
 for PATCHED_KUBE_BINARY_VERSION in ${KUBE_BINARY_VERSIONS}; do
   KUBERNETES_VERSION=$(echo ${PATCHED_KUBE_BINARY_VERSION} | cut -d"_" -f1 | cut -d"-" -f1 | cut -d"." -f1,2,3)
   extractKubeBinaries $KUBERNETES_VERSION "https://acs-mirror.azureedge.net/kubernetes/v${PATCHED_KUBE_BINARY_VERSION}/binaries/kubernetes-node-linux-${CPU_ARCH}.tar.gz"
 done
 
+rm -f ./azcopy # cleanup immediately after usage will retrun in two downloads
```
cameronmeissner, line 466: typo: return?

vhdbuilder/scripts/linux/tool_installs.sh
```diff
@@ -45,4 +45,34 @@ disableSystemdIptables() {
...
+}
+
+# download and setup azcopy to use to download private packages with MSI auth
+getAzCopyCurrentPath() {
+  if [[ -f ./azcopy ]]; then
+    echo "./azcopy already exists"
+  else
+    echo "get azcopy at \"${PWD}\"...start"
+    # Download and extract
+    azcopydownloadurl="https://azcopyvnext.azureedge.net/releases/release-10.22.1-20231220/azcopy_linux_amd64_10.22.1.tar.gz"
```
cameronmeissner, line 57: nit: use snake/camel casing here

vhdbuilder/scripts/linux/tool_installs.sh
```diff
@@ -45,4 +45,34 @@ disableSystemdIptables() {
...
+    azcopysha256="7549424d56ab2d8b4033c84c2a9bb167dc2dcbb23998acd7fffb37bc1a71a267"
```
cameronmeissner, line 58: nit: localize `azcopy_sha256`/`azcopy_download_url`?

vhdbuilder/scripts/linux/tool_installs.sh
```diff
@@ -45,4 +45,34 @@ disableSystemdIptables() {
...
+    if [[ $(isARM64) == 1 ]]; then
+      azcopydownloadurl="https://azcopyvnext.azureedge.net/releases/release-10.22.1-20231220/azcopy_linux_arm64_10.22.1.tar.gz"
+      azcopysha256="4db9a4b48abc7775f1a5d6d928afc42361dcc57bbfcde23ac82e4c419a0dc8fc"
+    fi
+
+    downloadedpkg="downloadazcopy"
+    pkgprefix="azcopy_linux_"
+
+    wget "$azcopydownloadurl" -O "$downloadedpkg" &&
```
cameronmeissner, line 67: we should be exiting on a new exit code here if the azcopy download fails, you can check how we do it for bpftrace above

I'd also wrap the wget in a `retrycmd_if_failure`

vhdbuilder/packer/install-dependencies.sh
daxakp: It is not a tool, like azcopy is? Just using azcopy with MSI env. vars so i think it does not belong to tools. Leave it here?

vhdbuilder/packer/install-dependencies.sh
daxakp: since IFS and read are on the same line, IFS value before and after this line remains the same, don't have to backup and restore - something like - Prefixing a command with parameter assignment affects the environment of the command being executed.?
Or am i missing anything?

e2e/scenario/vhd.go
daxakp: Either will work for this test but correcting the naming.

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: Should fail before landing here in this case but will add a check for kubelet as well.

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -16,6 +16,7 @@ KUBERNETES_VERSION={{GetParameter "kubernetesVersion"}}
 HYPERKUBE_URL={{GetParameter "kubernetesHyperkubeSpec"}}
 KUBE_BINARY_URL={{GetParameter "kubeBinaryURL"}}
 CUSTOM_KUBE_BINARY_URL={{GetParameter "customKubeBinaryURL"}}
+PRIVATE_KUBE_BINARY_URL="{{GetLinuxPrivatePackageURL}}"
```
anujmaheshwari1, line 19: @Bryce-Soghigian @tallaxes for a new addition to ce_cmd.sh

for @daxakp, is this URL something that needs to be provided through RP? can we not hardcode/program it into the VHD?

we are trying to get away from having URLs in the contract as much as possible.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -229,20 +230,46 @@ installAzureCNI() {
...
+    local k8s_tgz_tmp=${kube_binary_url##*/}
+    retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${k8s_tgz_tmp}" ${kube_binary_url} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
+    tar --transform="s|.*|&-${k8s_version}|" --show-transformed-names -xzvf "$K8S_DOWNLOADS_DIR/${k8s_tgz_tmp}" \
         --strip-components=3 -C /usr/local/bin kubernetes/node/bin/kubelet kubernetes/node/bin/kubectl
-    rm -f "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}"
+    rm -f "$K8S_DOWNLOADS_DIR/${k8s_tgz_tmp}"
 }
 
-installKubeletKubectlAndKubeProxy() {
+extractPrivateKubeBinaries() {
```
anujmaheshwari1, line 244: do we need a separate func for this? a lot of the code is same as `extractKubeBinaries`. 

since currently, PRIVATE_KUBE_BINARY_URL is an input from cse_cmd.sh, we can filter the download criteria based on whether that variable is set, right?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -229,20 +230,46 @@ installAzureCNI() {
 }
 
 extractKubeBinaries() {
-    K8S_VERSION=$1
-    KUBE_BINARY_URL=$2
+    local k8s_version="$1"
```
anujmaheshwari1, line 233: nit: any particular reason to lower case the vars? I dont have a strong preference but i'd like the format to be consistent with the other variables used in the func like K8S_DOWNLOADS_DIR, and the exit code.

parts/linux/cloud-init/artifacts/cse_cmd.sh
bcho: is that possible to reuse the `CUSTOM_KUBE_BINARY_URL`?

daxakp: Yes, @anujmaheshwari1, it needs to be provided from RP, to decide which cached package to use from the VHD at run-time.

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: It can be written multiple ways, but i think the separate function is better - it clearly explains how custom, private and default paths are handled.

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: My understanding is that the local variables in lower case is the convention - K8S_DOWNLOADS_DIR and exit codes are global and/or exported.

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: ah, I guess you're right, just tested this locally

parts/linux/cloud-init/artifacts/cse_cmd.sh
daxakp: @bcho, others can confirm but as i understand, custom_url is not used except rarely for internal testing? Private_url naming etc. is in sync with windows naming, potentially that can be used instead of custom_url and can remove custom_url.

ganeshkumarashok:
would it be useful to do validation on the private package URLs?

parts/linux/cloud-init/artifacts/cse_install.sh
daxakp: @anujmaheshwari1 
This is how the common function will look, two separate functions are better, easy to understand what is done in which case and less error prone - but pls confirm if you prefer this version or suggest what you are thinking here? 

extractKubeBinaries() {
    local k8s_version="$1"
    local kube_binary_url="$2"
    local is_private_url="$3"

    local k8s_tgz_tmp=${kube_binary_url##*/}

    local err=$ERR_K8S_DOWNLOAD_TIMEOUT
    if [[ $is_private_url == true ]]; then
        k8s_tgz_tmp="${K8S_CACHE_DIR}/${k8s_tgz_tmp}"
        if [[ -f "${k8s_tgz_tmp}" ]]; then
            echo "cached package ${k8s_tgz_tmp} is found, will use that"
        else
            echo "cached package ${k8s_tgz_tmp} not found"
            return 1
        fi

        # remove the current kubelet and kubectl binaries before extracting new binaries from the cached package
        rm -rf /usr/local/bin/kubelet-* /usr/local/bin/kubectl-*
        err=$ERR_PRIVATE_K8S_PKG_ERR
    fi

    retrycmd_get_tarball 120 5 "${k8s_tgz_tmp}" ${kube_binary_url} || exit "$err"
    tar --transform="s|.*|&-${k8s_version}|" --show-transformed-names -xzvf "${k8s_tgz_tmp}" \
        --strip-components=3 -C /usr/local/bin kubernetes/node/bin/kubelet kubernetes/node/bin/kubectl

    if [[ $is_private_url == false ]]; then
        rm -f "$K8S_DOWNLOADS_DIR/${k8s_tgz_tmp}"
    fi
}

Bryce-Soghigian (approved):
Approved on the karpenter side, shouldn't be any breaking changes since the url passed in is an optional parameter and functions fine when we pass in an empty string. 

Please wait on node sig approval as well before merging :)

parts/linux/cloud-init/artifacts/cse_install.sh
anujmaheshwari1: i think this works! thank you for consolidating it!

anujmaheshwari1 (approved):

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -10,6 +10,7 @@ CRICTL_BIN_DIR="/usr/local/bin"
 CONTAINERD_DOWNLOADS_DIR="/opt/containerd/downloads"
 RUNC_DOWNLOADS_DIR="/opt/runc/downloads"
 K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
+K8S_CACHE_DIR="/opt/kubernetes/downloads/private-packages"
```
cameronmeissner, line 13: nit: K8S_PRIVATE_PACKAGES_CACHE_DIR?

PR #3950, opened 2024-01-04
feat: update CUDA driver version to 535.54.03

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/baker_test.go
```diff
@@ -1737,14 +1737,14 @@ var _ = Describe("getGPUDriverVersion", func() {
...
-		Expect(getGPUDriverVersion("standard_nc6_v3")).To(Equal(datamodel.Nvidia525CudaDriverVersion))
+		Expect(getGPUDriverVersion("standard_nc6_v3")).To(Equal(datamodel.Nvidia535CudaDriverVersion))
 	})
 	It("should use 535 grid with nv v5", func() {
 		Expect(getGPUDriverVersion("standard_nv6ads_a10_v5")).To(Equal(datamodel.Nvidia535GridDriverVersion))
 		Expect(getGPUDriverVersion("Standard_nv36adms_A10_V5")).To(Equal(datamodel.Nvidia535GridDriverVersion))
 	})
-	It("should use 525 cuda with nv v1 (although we don't know if that works)", func() {
-		Expect(getGPUDriverVersion("standard_nv6")).To(Equal(datamodel.Nvidia525CudaDriverVersion))
+	It("should use 535 cuda with nv v1 (although we don't know if that works)", func() {
```
Bryce-Soghigian, line 1746: "Although we don't know if that works" seems mans couldn't get GPU Quota

pkg/agent/datamodel/const.go
```diff
@@ -119,17 +119,15 @@ const (
...
-	Nvidia510GridDriverVersion = "grid-510.73.08"
+	Nvidia535CudaDriverVersion = "cuda-535.54.03"
 	Nvidia535GridDriverVersion = "grid-535.54.03"
 )
 
 // These SHAs will change once we update aks-gpu images in aks-gpu repository. We do that fairly rarely at this time
 // So for now these will be kept here like this.
 const (
 	AKSGPUGridSHA = "sha-20ffa2"
-	AKSGPUCudaSHA = "sha-16fd35"
```
cameronmeissner, line 132: should we also be updating this in the build scripts?

UtheMan: updated, will run a test build

pkg/agent/baker_test.go
UtheMan: That's a question to our ancestors

UtheMan: Checked - those SKUs are retired now, will keep this test just for safety and will add a note about retirement.

UtheMan: https://learn.microsoft.com/en-us/azure/virtual-machines/nv-series?source=recommendations

ganeshkumarashok (approved):
Lgtm if driver version and device plugin installation works when testing for H100 as well

PR #3969, opened 2024-01-12
refactor: remove containerd-monitor from vhd

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/const.go
```diff
@@ -60,8 +60,6 @@ const (
 	kubernetesKubeletMonitorSystemdService    = "linux/cloud-init/artifacts/kubelet-monitor.service"
 	kubernetesDockerMonitorSystemdTimer       = "linux/cloud-init/artifacts/docker-monitor.timer"
```
cameronmeissner, line 61: we can probably also get rid of the docker monitor stuff as well, maybe a good follow-up item

cameronmeissner (approved):

PR #3971, opened 2024-01-12
feat: prefetch script generation during VHD builds

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -834,6 +834,21 @@ testPam() {
   return $retval
 }
 
+testPrefetchScripts() {
+  local test="testPrefetchScripts"
+  local cni_prefetch_script="/opt/azure/containers/cni-prefetch.sh"
```
anujmaheshwari1, line 839: nit: consider adding path suffix to var name to make more sense in the echo statement that follows, maybe "cni_prefetch_script_path"

vhdbuilder/prefetch/pkg/cni/types.go
```diff
@@ -0,0 +1,17 @@
...
+
+// CNIContainerImage represents a container image used for CNI-related functionality.
+type CNIContainerImage struct {
+	// FullyQualifiedTag is the fully qualified tag identifier of the image (e.g. mcr.microsoft.com/containernetworking/cni-dropgz:v0.0.4.2).
+	FullyQualifiedTag string
+	// Binaries is a list of absolute paths to binaries which should be prefetched from within the container image.
+	Binaries []string
+}
+
+// PrefetchTemplateArgs represnets the arguments required by the prefetch script template.
```
anujmaheshwari1, line 11: nit: represents

anujmaheshwari1 (approved):
lgtm pending Will's review!

vhdbuilder/prefetch/pkg/cni/cni.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+var (
+	//go:embed cni-prefetch.sh.gtpl
+	prefetchScript         string
+	prefetchScriptTemplate = template.Must(template.New("cniprefetch").Parse(prefetchScript))
+
+	// this represents the CNI-related container images and their respective binaries
+	// which are in-scope for prefetch optimization
+	prefetchList = CNIPrefetchList{
```
wedaly, line 21: any chance we could put this in components.json? Worried people won't know about this and will forget to update it as things change.

vhdbuilder/prefetch/pkg/cni/cni.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Generate generates and saves the CNI prefetch script to disk based on the specified component list
+// and destination path. This can later be generalized to an interface if we need to prefetch
+// other non-CNI related images/binaries.
+func Generate(components *component.List, dest string) error {
+	var templateArgs PrefetchTemplateArgs
+	for _, image := range components.ContainerImages {
+		if binaries, shouldPrefetch := prefetchList[image.DownloadURL]; shouldPrefetch {
+			// only prefetch mutli-arch versions for now
+			for _, tag := range image.MultiArchVersions {
```
wedaly, line 46: Could we make the list of binaries specific to a version? It's possible that binaries are added/removed/renamed across versions.

vhdbuilder/prefetch/pkg/cni/cni-prefetch.sh.gtpl
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -eux
```
wedaly, line 2: if someone specifies a binary that doesn't exist in the image, will CI fail on the PR? If not, is there a way we could make that happen? (maybe run the prefetch script as part of CI and verify that it exits with status zero?)

vhdbuilder/prefetch/pkg/cni/cni.go
```diff
@@ -0,0 +1,72 @@
...
+			"opt/cni/bin/loopback",
+			"opt/cni/bin/portmap",
+			"opt/cni/bin/tuning",
+		},
+	}
+)
+
+// Generate generates and saves the CNI prefetch script to disk based on the specified component list
+// and destination path. This can later be generalized to an interface if we need to prefetch
+// other non-CNI related images/binaries.
```
wedaly, line 40: apart from the binaries listed in `prefetchList`, is there anything CNI specific here? I wonder if it's worth removing the "cni" names in pkgs/structs/functions to make it easier if we ever want to use this for other components.

cameronmeissner: yeah that's fair, my idea was to put different high-level things we want to prefetch, such as CNI in this case, into separate pkgs that are all self-contained and provide respective `Generate` functions which get called by the main pkg - though you're probably right in that this pkg could be further generalized to "prefetching binaries from a container image"

vhdbuilder/prefetch/pkg/cni/cni-prefetch.sh.gtpl
cameronmeissner: yep, Alison is working on a separate PR for the driver script which will include running the script itself as a part of the content tests - might just consolidate tests into that PR

vhdbuilder/prefetch/pkg/cni/cni.go
cameronmeissner: fair point, will also try to address "prefetchable" binaries per container image version as well with this shift

vhdbuilder/prefetch/pkg/cni/cni-prefetch.sh.gtpl
cameronmeissner: went ahead and added a base-level execution test

vhdbuilder/prefetch/pkg/cni/cni.go
cameronmeissner: fixed

vhdbuilder/prefetch/pkg/container/container.go
```diff
@@ -0,0 +1,68 @@
...
+	return &list, nil
+}
+
+// Generate generates and saves the container image prefetch script to disk based on the specified component list and destination path.
+func Generate(components *ComponentList, dest string) error {
+	var args TemplateArgs
+	for _, image := range components.Images {
+		if len(image.PrefetchOptimizations) > 0 {
+			for _, optimization := range image.PrefetchOptimizations {
+				args.Images = append(args.Images, TemplateImage{
```
wedaly, line 44: could you add validation here that `optimization.Tag` exists in either `image.MultiArchVersions` or `image.AMD64OnlyVersions`? Want to catch cases where someone accidentally mistypes the version.

vhdbuilder/prefetch/pkg/container/container.go
```diff
@@ -0,0 +1,68 @@
...
+					FullyQualifiedTag: fmt.Sprintf("%s%s", strings.TrimSuffix(image.DownloadURL, "*"), optimization.Tag),
```
wedaly, line 45: Does anything validate that `DownloadURL` always ends with "*"? (guessing the answer is "yes" since everything in components.json seem to have it)

vhdbuilder/prefetch/pkg/container/container.go
cameronmeissner: good call, added

vhdbuilder/prefetch/pkg/container/container.go
cameronmeissner: I actually originally thought that the cue schema did something along these lines, but this seems to not be the case - added an appropriate failure mode

technically if it wasn't formatted in this way, the container image pull steps within install-dependencies.sh would most likely fail: https://github.com/Azure/AgentBaker/blob/cameissner/cni-prefetch/vhdbuilder/packer/install-dependencies.sh#L237-L267

wedaly (approved):

PR #3988, opened 2024-01-23
feat: get DU tool to Window VHD

ShiqianTao:
**What type of PR is this?**
/kind feature

Download du.exe tool to Windows VHD and this tool can be used for debugging AKS Windows node.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -205,6 +205,19 @@ function Get-FilesToCacheOnVHD {
     }
 }
 
+function Get-ToolsToVHD {
+    # Rely on the completion of Get-FilesToCacheOnVHD
+    $cacheDir = "c:\akse-cache\tools"
+
+    $toolsDir = "c:\aks-tools"
+    if (!(Test-Path -Path $toolsDir)) {
+        New-Item -ItemType Directory -Path $toolsDir | Out-Null
+    }
+
+    Write-Log "Getting DU (Windows Disk Usage)"
+    Expand-Archive -Path "$cacheDir\DU.zip" -DestinationPath "$toolsDir\DU" -Force
+}
+
 function Get-PrivatePackagesToCacheOnVHD {
     if (![string]::IsNullOrEmpty($env:WindowsPrivatePackagesURL)) {
         Write-Log "Caching private packages on VHD"
@@ -813,6 +826,7 @@ try{
             Update-Registry
             Get-ContainerImages
             Get-FilesToCacheOnVHD
+            Get-ToolsToVHD # Rely on the completion of Get-FilesToCacheOnVHD
             Get-PrivatePackagesToCacheOnVHD
             Remove-Item -Path c:\windows-vhd-configuration.ps1
             (New-Guid).Guid | Out-File -FilePath 'c:\vhd-id.txt'
--- a/vhdbuilder/packer/test/windows-vhd-content-test.ps1
+++ b/vhdbuilder/packer/test/windows-vhd-content-test.ps1
@@ -506,6 +506,19 @@ function Test-WindowsDefenderPlatformUpdate {
     }
 }
 
+function Test-ToolsToCacheOnVHD {
+    $toolsDir = "c:\aks-tools"
+    $toolsList = @("DU\du.exe", "DU\du64.exe", "DU\du64a.exe")
+
+    foreach ($tool in $toolsList) {
+        $toolPath = Join-Path -Path $toolsDir -ChildPath $tool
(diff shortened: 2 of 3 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -232,6 +232,26 @@ function Get-PrivatePackagesToCacheOnVHD {
     }
 }
 
+function Get-ToolsToVHD {
+    $cacheDir = "c:\toolscache"
```
AbelHu, line 236: Keep same pattern to use a temp folder in $env:TEMP instead of creating a temp folder in the root?

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -232,6 +232,26 @@ function Get-PrivatePackagesToCacheOnVHD {
...
+    if (!(Test-Path -Path $cacheDir)) {
+        New-Item -ItemType Directory -Path $cacheDir | Out-Null
+    }
+
+    $toolsDir = "c:\tools" # discuss
```
AbelHu, line 241: could we use aks as prefix? NOTE: Remove the comment.

vhdbuilder/packer/test/windows-vhd-content-test.ps1
```diff
@@ -506,6 +506,19 @@ function Test-WindowsDefenderPlatformUpdate {
     }
 }
 
+function Test-ToolsToCacheOnVHD {
+    $toolsDir = "c:\tools"
```
AbelHu, line 510: Could you also help to update vhdbuilder/packer/test/windows-files-check.ps1 to validate the signature of the new binaries?

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -232,6 +232,26 @@ function Get-PrivatePackagesToCacheOnVHD {
...
+    if (!(Test-Path -Path $toolsDir)) {
+        New-Item -ItemType Directory -Path $toolsDir | Out-Null
+    }
+
+    Write-Log "Caching du.exe"
+    $URL = "https://download.sysinternals.com/files/DU.zip"
```
AbelHu, line 247: It should be better to put the URL and file list in vhdbuilder/packer/generate-windows-vhd-configuration.ps1

ShiqianTao: Updated

vhdbuilder/packer/configure-windows-vhd.ps1
ShiqianTao: Updated

vhdbuilder/packer/configure-windows-vhd.ps1
ShiqianTao: Put it into the map of vhdbuilder/packer/generate-windows-vhd-configuration.ps1 so it will use the folder `c:\akse-cache`

vhdbuilder/packer/test/windows-vhd-content-test.ps1
ShiqianTao: Confirmed, it will validate

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -205,6 +205,20 @@ function Get-FilesToCacheOnVHD {
     }
 }
 
+function Get-ToolsToVHD {
+    # Rely on the completion of Get-FilesToCacheOnVHD
+    $cacheDir = "c:\akse-cache\tools"
```
AbelHu, line 210: akse-cache cannot be used since it will be deleted in AKS Windows CSE script

AbelHu: is it only used for temporary caching? I think that we should not do that for this purpose.

AbelHu: If we use a temp folder in $env:temp, we do not need to clean up it since it will be cleaned up automatically.

AbelHu: It seems like that you are using the current code logic to download the files in akse-cache and validate the signature. It is good for me.
My only suggestion is to expand the archive to the folder `c:\aks-tools\du`

ShiqianTao: Updated, thank you

AbelHu (approved):
LGTM. Thanks for the PR. Please merge the PR after all tests pass.

PR #3996, opened 2024-01-25

HDYA:


vhdbuilder/packer/components.json
```diff
@@ -52,20 +52,6 @@
       "multiArchVersions": [
         "v0.0.4.2",
         "v0.1.1"
-      ],
-      "prefetchOptimizations": [
```
cameronmeissner, line 56: is a script being used to make these updates? if so we'll need to change it to preserve `prefetchOptimizations`

cameronmeissner (changes requested):
is a script being used to make these updates? if so we'll need to change it to preserve prefetchOptimizations

vhdbuilder/packer/components.json
cameronmeissner, line 56: this shouldn't be getting modified here

PR #4007, opened 2024-01-26
feat: add AKS check network script to VHD

alyssa1303:
**What type of PR is this?**
/kind feature

Adds AKS Check Network script to perform basic troubleshooting by checking outbound connections from AKS node to required endpoints.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

Script was tested using AgentBaker E2E instruction and manually created a cluster to run the image. Example result of test output when there are blocked endpoints.
image

**Release note**:
```
none
```


vhdbuilder/packer/packer_source.sh
```diff
@@ -99,6 +99,10 @@ copyPackerFiles() {
   AKS_LOGROTATE_TIMER_DROPIN_DEST=/etc/systemd/system/logrotate.timer.d/override.conf
   AKS_LOGROTATE_CONF_SRC=/home/packer/rsyslog
   AKS_LOGROTATE_CONF_DEST=/etc/logrotate.d/rsyslog
+  AKS_CHECK_NETWORK_SRC=/home/packer/aks-check-network.sh.sh
+  AKS_CHECK_NETWORK_DEST=/usr/local/bin/aks-check-network.sh.sh
```
cameronmeissner, line 103: can probably put this in `/opt/azure/containers/aks-check-network.sh` instead

cameronmeissner: also seems `.sh` is repeated twice here?

vhdbuilder/packer/packer_source.sh
```diff
@@ -99,6 +99,10 @@ copyPackerFiles() {
   AKS_LOGROTATE_TIMER_DROPIN_DEST=/etc/systemd/system/logrotate.timer.d/override.conf
   AKS_LOGROTATE_CONF_SRC=/home/packer/rsyslog
   AKS_LOGROTATE_CONF_DEST=/etc/logrotate.d/rsyslog
+  AKS_CHECK_NETWORK_SRC=/home/packer/aks-check-network.sh.sh
```
cameronmeissner, line 102: nit: `AKS_CHECK_NETWORK_SCRIPT_SRC` / `AKS_CHECK_NETWORK_SCRIPT_DEST`

parts/linux/cloud-init/artifacts/aks-check-network.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=Check outbound connectivity from node to certain endpoints
+
+[Service]
+Type=oneshot
+ExecStart=/bin/bash /usr/local/bin/aks-check-network.sh
```
cameronmeissner, line 6: we should probably specify network as a dependency

cameronmeissner: `After=network-online.target`

cameronmeissner: we'll also need to explicitly enable the unit as well - we should be able to do this during the VHD build itself (pre-install-dependenceis.sh)

cameronmeissner: actually, it seems like we'd need the apiserver fqdn to actually invoke the script... so I guess we can't necessarily enable the unit during build time, would need to be at runtime during CSE instead

parts/linux/cloud-init/artifacts/aks-check-network.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
+# 
+# AKS Check Network
+#
+# This script is used to check network connectivity from the node to certain required AKS endpoints 
+# and log the results to the events directory. For now, this script has to be triggered manually to
+# collect the log. In the future, we will run it periodically to check and alert any issue.
```
cameronmeissner, line 7: we can setup a systemd timer to run this periodically later if we want

vhdbuilder/packer/packer_source.sh
alyssa1303: Fixed! thanks for pointing that out! I moved it to `/opt/azure/containers` too.

vhdbuilder/packer/packer_source.sh
alyssa1303: updated!

parts/linux/cloud-init/artifacts/aks-check-network.sh
alyssa1303: yes! that's what I plan to do in the future. Still need to decide how often we want to run it

parts/linux/cloud-init/artifacts/aks-check-network.service
alyssa1303: Added dependency for network. I made the apiserver fqdn optional, so the script won't throw an error. But I think we still want to enable it at runtime so we can check connection to apiserver.

anson627 (approved):

anson627:
it would be nice if you can document how you tested the script

parts/linux/cloud-init/artifacts/aks-check-network.sh
```diff
@@ -0,0 +1,206 @@
...
+AZURE_CONFIG_PATH="/etc/kubernetes/azure.json"
+AKS_CA_CERT_PATH="/etc/kubernetes/certs/apiserver.crt"
+AKS_CERT_PATH="/etc/kubernetes/certs/client.crt"
+AKS_KEY_PATH="/etc/kubernetes/certs/client.key"
+RESOLV_CONFIG_PATH="/etc/resolv.conf"
+SYSTEMD_RESOLV_CONFIG_PATH="/run/systemd/resolve/resolv.conf"
+
+ARM_ENDPOINT="management.azure.com"
+METADATA_ENDPOINT="http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://${ARM_ENDPOINT}/"
+API_VERSION="2023-11-01"
```
anujmaheshwari1, line 22: nit: are we going to periodically update this API version? if not, we can just hardcode like its done for metadata_endpoint, otherwise, we should think of a mechanism to update this through some automation. can be a follow up, not blocking for this PR.

parts/linux/cloud-init/artifacts/aks-check-network.sh
```diff
@@ -0,0 +1,206 @@
...
+AKS_ENDPOINT="https://${ARM_ENDPOINT}/providers/Microsoft.ContainerService/operations?api-version=${API_VERSION}"
+APISERVER_ENDPOINT="https://${APISERVER_FQDN}/healthz"
+ACS_BINARY_ENDPOINT="acs-mirror.azureedge.net/azure-cni/v1.4.43/binaries/azure-vnet-cni-linux-amd64-v1.4.43.tgz"
```
anujmaheshwari1, line 25: again, this particular CNI version is controlled by a toggle in RP, if that toggle is updated, whats the plan to update it here?

parts/linux/cloud-init/artifacts/aks-check-network.sh
alyssa1303: we might but that won't be necessary for now so I removed `API_VERSION` variable according to your suggestion.

parts/linux/cloud-init/artifacts/aks-check-network.sh
alyssa1303: The reason why I used a specific endpoint is to get a 200 status code when using curl because this is a repository endpoint so just doing a generic path like this `curl https://acs-mirror.azureedge.net` will return a 400 status code saying bad request error. However, since the purpose of this test is to check if connection is blocked by any firewall/nsg, 400 is a good indication that there's connection so I guess we should be fine with 400 status code. Thus, I updated the endpoint to remove the specific version, so we don't have to deal with updating version constantly or if the binary is taken down.

anujmaheshwari1 (approved):

PR #4008, opened 2024-01-26
chore: new policy addon and gk versions

anlandu:
**What type of PR is this?**
/kind cleanup

update policy addon and gatekeeper versions

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/components.json
+++ b/vhdbuilder/packer/components.json
@@ -304,6 +304,7 @@
       "downloadURL": "mcr.microsoft.com/oss/open-policy-agent/gatekeeper:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
+        "v3.14.0",
         "v3.13.4",
         "v3.13.3"
       ]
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/components.json
```diff
@@ -304,6 +304,7 @@
       "downloadURL": "mcr.microsoft.com/oss/open-policy-agent/gatekeeper:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
+        "v3.14.0",
         "v3.13.4",
         "v3.13.3"
```
cameronmeissner, line 309: do we need to keep the oldest version of these images cached even after this addition?

anlandu: After the update to the new image version has rolled out (hoping to make the Feb AKS release) I'll circle back with a PR removing the old one.

anlandu: Went ahead and removed 3.13.4 and 1.2.2 since those didn't get deployed due to the code freeze, hope that helps

PR #4018, opened 2024-01-30
fix: add trivy back to artifacts

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


ganeshkumarashok (approved):
Lgtm after you have confirmed that the artifacts are all being created in the VHD build

ganeshkumarashok:
We may want to consider alternate methods that don't need the trivy DB to be downloaded on/during the VHD build, because of space constraints.

PR #4027, opened 2024-02-01
feat: adding in windows dropgz to pre-cache list

pjohnst5:
**What type of PR is this?**
/kind feature

This pre-cache's Windows dropgz for AKS nodes
This Windows dropgz is only used in Overlay nodes in AKS for now, but asked by Abel to start caching it


AbelHu (approved):
LGTM. Thanks for the PR.

PR #4059, opened 2024-02-09
chore: cse reduce retries and increase wait time on apiserver validation

chmill-zz:
**What type of PR is this?**

cse wait time

increase the nslookup timeout to 15s and reduce the retries to 20 (40 private)

the increase is due to more than one customer using custom dns where the first dns doesn't resolve but the second or third does, waiting for a bit longer shows success

the retry reduction is to avoid exiting with 124 and getting the expected 52/53. 

So we're going from 5 * 100 +100 or 600 seconds to 15 * 20 + 20 or  320 seconds. The extra + is from the 1s sleep between calls. The p99 for success in the past 14d is ~5s

We're still going to wait 5min before giving up pending i can math. 

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner (approved):

levimm:
the `-timeout` won't help when dns server returns NXDOMAIN. That means it will only wait for 20s now if fqdn is not propagated. It will cause quite a lot cse issue I think.

cameronmeissner:
we went ahead with Lima's change instead, I think we can probably close this

PR #4070, opened 2024-02-14
feat: e2e test for containerd egress path

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/scenario/scenario_ubuntu2204-standalone-containerd.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+type Manifest struct {
+	Containerd struct {
+		Edge string `json:"edge"`
+	} `json:"containerd"`
+}
+
+func getCurrentManifestContainerdVersion() string {
```
cameronmeissner, line 57: might be worth moving this to a common location so others can modify as necessary since it could be useful in other scenarios

cameronmeissner: we could also just grab the manifest from the running node itself since it'll be there - I feel like that would be better since then we wouldn't have to rely on the manifest being in a particular location within the repo for this to work properly

this is probably okay for now but ^ that might be a good follow-up item (would basically just incorporate grabbing the manifest and checking the version into the live VM validator)

e2e/scenario/scenario_ubuntu2204-standalone-containerd.go
```diff
@@ -0,0 +1,72 @@
...
+				nbc.ContainerdPackageURL = "https://packages.microsoft.com/ubuntu/22.04/prod/pool/main/m/moby-containerd/moby-containerd_1.6.9+azure-ubuntu22.04u1_amd64.deb"
+			},
+			LiveVMValidators: []*LiveVMValidator{
+				containerdVersionValidator("1.6.9"),
+			},
+		},
+	}
+}
+
+func (t *Template) ubuntu2204ContainerdVersion() *Scenario {
```
cameronmeissner, line 31: we usually try to separate individual scenarios into their own files for clarity

e2e/scenario/scenario_ubuntu2204-standalone-containerd.go
```diff
@@ -0,0 +1,72 @@
...
+		Name:        "ubuntu2204ContainerdVersion",
+		Description: "Tests that a node using the Ubuntu 2204 VHD with a containerd version can be properly bootstrapped",
+		Config: Config{
+			ClusterSelector: NetworkPluginKubenetSelector,
+			ClusterMutator:  NetworkPluginKubenetMutator,
+			VHDSelector:     t.Ubuntu2204Gen2Containerd,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-ubuntu-containerd-22.04-gen2"
+				nbc.AgentPoolProfile.Distro = "aks-ubuntu-containerd-22.04-gen2"
+				nbc.ContainerdVersion = "1.6.9"
```
cameronmeissner, line 42: sorry a little confused, what exactly is this scenario testing?

e2e/scenario/scenario_ubuntu2204-standalone-containerd.go
```diff
@@ -0,0 +1,72 @@
...
+	manifestData, err := os.ReadFile("../parts/linux/cloud-init/artifacts/manifest.json")
+	if err != nil {
+		return ""
+	}
+	manifestDataStr := string(manifestData)
+	manifestDataStr = strings.TrimRight(manifestDataStr, "#EOF \n\r\t")
+	manifestData = []byte(manifestDataStr)
+
+	var manifest Manifest
+	err = json.Unmarshal([]byte(manifestData), &manifest)
```
cameronmeissner, line 67:
```suggestion
        if err = json.Unmarshal([]byte(manifestData), &manifest); err != nil {
              panic(err)
         }
```

we should probably panic here instead of just silencing the error and creating an invalid validator

e2e/scenario/scenario_ubuntu2204-standalone-containerd.go
AlisonB319: nbc.ContainerdVersion = "1.6.9"

there are technically different path options when ContainerdVersion is specified vs ContainerdPackageURL. But they are very similar, which is why I put them in the same file

AlisonB319: Added a clarifying comment at the top

e2e/scenario/vhd.go
```diff
@@ -97,6 +97,28 @@ func mustGetVHDCatalogFromEmbeddedJSON(rawJSON string) VHDCatalog {
 	return catalog
 }
 
+type Manifest struct {
+	Containerd struct {
+		Edge string `json:"edge"`
+	} `json:"containerd"`
+}
+
+func getVHDManifest() (Manifest, error) {
```
cameronmeissner, line 106: nit: lets use `*Manifest` here instead

e2e/scenario/vhd.go
```diff
@@ -97,6 +97,28 @@ func mustGetVHDCatalogFromEmbeddedJSON(rawJSON string) VHDCatalog {
...
+	if err != nil {
+		return Manifest{}, err
+	}
+	manifestDataStr := string(manifestData)
+	manifestDataStr = strings.TrimRight(manifestDataStr, "#EOF \n\r\t")
+	manifestData = []byte(manifestDataStr)
+
+	manifest := Manifest{}
+	if err = json.Unmarshal([]byte(manifestData), &manifest); err != nil {
+		panic(err)
```
cameronmeissner, line 117: lets just return the error here and panic in the parent call

e2e/scenario/scenario_ubuntu2204-standalone-containerd.go
```diff
@@ -0,0 +1,58 @@
...
+				containerdVersionValidator("1.6.9"),
+			},
+		},
+	}
+}
+
+func (t *Template) ubuntu2204ContainerdVersion() *Scenario {
+	return &Scenario{
+		Name:        "ubuntu2204ContainerdVersion",
+		Description: "Tests that a node using the Ubuntu 2204 VHD with a containerd version can be properly bootstrapped",
```
cameronmeissner, line 34: can we make the descriptions a bit more clear? e.g. "tests that node using an Ubuntu2204 VHD and the ContainerdVersion override bootstraps with the correct manifest containerd version and ignores the override" or similar

cameronmeissner (approved):

PR #4072, opened 2024-02-15
fix: use fixed retry timeout in cse nslookup step

levimm:
**What type of PR is this?**
/kind flake

Two issues to fix:
1. Sometimes cx using their own dns server would failed to resolve the fqdn because dns propagation takes more than 3 minutes. The `-timeout=5` doesn't help in this case since NXDOMAIN would return immediately.
2. When cx has multiple dns server defined, if one of them couldn't get connected, we want to use longer timeout 15 so we have chances to try 2nd or 3rd dns servers.

Instead of using fixed retry count and doing calculation for the sleep and timeout, this pr uses fixed retry timeout so we don't need to care about setting different sleep interval or nslookup timeout settings for these two issues.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
use fixed timeout 5minutes for public dns record(10 minutes for private dns record) in cse nslookup step.
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -136,6 +136,23 @@ retrycmd_if_failure() {
...
+    wait_sleep=$1; timeout=$2; total_timeout=$3; record=$4
+    start_time=$(date +%s)
+    while true; do
+        nslookup -timeout=$timeout -retry=0 $record && break || \
+        current_time=$(date +%s)
+        # Check if the total_timeout has been reached
+        if [ $((current_time - start_time)) -ge $total_timeout ]; then
+            echo "Total timeout $total_timeout reached, nslookup -timeout=$timeout -retry=0 $record failed"
+            return 1
+        else
```
cameronmeissner, line 149: nit: redundant else

PR #4088, opened 2024-02-20
feat: add systemd unit for Mariner live patching

hongbin:
**What type of PR is this?**

Add systemd unit for Mariner live patching

/kind feature

Mariner also needs live patching support as Ubuntu.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

The Ubuntu live patching PR (merged): https://github.com/Azure/AgentBaker/pull/3643

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -301,7 +301,7 @@ if [ "${ENSURE_NO_DUPE_PROMISCUOUS_BRIDGE}" == "true" ]; then
     logs_to_events "AKS.CSE.ensureNoDupOnPromiscuBridge" ensureNoDupOnPromiscuBridge
 fi
 
-if [[ $OS == $UBUNTU_OS_NAME ]]; then
+if [[ $OS == $UBUNTU_OS_NAME || $OS == $MARINER_OS_NAME ]]; then
```
cameronmeissner, line 304: nit: separate clauses into their own set of `[]`'s

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if ! dnf_update; then
+    echo "dnf_update failed"
+    exit 1
+fi
+
+# update current timestamp
+$KUBECTL annotate --overwrite node ${node_name} kubernetes.azure.com/live-patching-current-timestamp=${golden_timestamp}
+
+echo package update completed successfully
```
cameronmeissner, line 49: nit: echo "package update completed successfully"

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
```diff
@@ -0,0 +1,49 @@
...
+set -o nounset
+set -e
+
+# source dnf_update
+source /opt/azure/containers/provision_source_distro.sh
+
+KUBECTL="/usr/local/bin/kubectl --kubeconfig /var/lib/kubelet/kubeconfig"
+
+# At startup, we need to wait for kubelet to finish TLS bootstrapping to create the kubeconfig file.
+while [ ! -f /var/lib/kubelet/kubeconfig ]; do
```
cameronmeissner, line 12: should we add a timeout here?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -134,6 +134,28 @@ write_files:
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "migPartitionScript"}}
 
+{{if IsMariner}}
```
cameronmeissner, line 137: so is the plan to distribute this unit/timer to older VHDs currently in the wild?

adding customdata payloads here will have it such that certain operations (such as SP reset) will cause VMs running with older node image versions (and intended customdata/cse) to receive these payloads when the VMs are reimaged - is that something we want?

generally if we add something directly do the VHD like you're doing with these payloads, we don't put into CSE/customdata.

vhdbuilder/packer/packer_source.sh
```diff
@@ -111,6 +111,8 @@ copyPackerFiles() {
   KUBELET_SERVICE_DEST=/etc/systemd/system/kubelet.service
   USU_SH_SRC=/home/packer/ubuntu-snapshot-update.sh
   USU_SH_DEST=/opt/azure/containers/ubuntu-snapshot-update.sh
+  MPU_SH_SRC=/home/packer/mariner-package-update.sh
+  MPU_SH_DEST=/opt/azure/containers/mariner-package-update.sh
   USU_SERVICE_SRC=/home/packer/snapshot-update.service
```
cameronmeissner, line 116: should we rename `USU_SERVICE_SRC/DEST` to `SNAPSHOT_UPDATE_SRC/DEST` for clarity since we're using it in both ubuntu + mariner cases after this change?

vhdbuilder/packer/vhd-image-builder-mariner-arm64.json
```diff
@@ -509,6 +509,21 @@
...
+    },
+    {
+      "type": "file",
+      "source": "parts/linux/cloud-init/artifacts/mariner/package-update.service",
+      "destination": "/home/packer/snapshot-update.service"
+    },
+    {
+      "type": "file",
+      "source": "parts/linux/cloud-init/artifacts/mariner/package-update.timer",
+      "destination": "/home/packer/snapshot-update.timer"
```
cameronmeissner, line 525: if we're putting this on the VHD directly then we _could_ just be able to enable the timer at build time so it automatically starts on subsequent boots (e.g. CSE), though probably fine to leave as is with `ensureSnapshotUpdate` in cse_main.sh

parts/linux/cloud-init/artifacts/cse_main.sh
hongbin: Done

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
hongbin: Done

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
hongbin: Done. Timeout will be 100 * 3s = 300s

parts/linux/cloud-init/nodecustomdata.yml
hongbin: The requirement is to have the unit/timer backfilling into old clusters. The final goal is to rollout this feature to all clusters, including existing/new clusters.

So yes. This is what I want.

vhdbuilder/packer/packer_source.sh
```diff
@@ -316,6 +318,9 @@ copyPackerFiles() {
   fi
   if [[ $OS == $MARINER_OS_NAME ]]; then
     cpAndMode $CONTAINERD_SERVICE_SRC $CONTAINERD_SERVICE_DEST 644
+    cpAndMode $MPU_SH_SRC $MPU_SH_DEST 544
+    cpAndMode $USU_SERVICE_SRC $USU_SERVICE_DEST 644
```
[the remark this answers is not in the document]
hongbin: Done

vhdbuilder/packer/vhd-image-builder-mariner-arm64.json
hongbin: Since we also leverage nodecustomdata, I stay with ensureSnapshotUpdate approach.

parts/linux/cloud-init/nodecustomdata.yml
yewmsft: @cameronmeissner , This systemd unit is triggered by an internal RP API. So with this systemd unit alone, there will be noop on the nodes. Our plan is to use vm extension(npd) to push these units to older images.  Before this is done, we will only trigger the RP API when the target agent pool with specific node image versions (> the next release version). So even the old images gets these units with a reimage, there will be noop on the older images nodes. We will only remove the filtering when the systemd unit is available everywhere.

yewmsft (approved):

cameronmeissner (approved):

cameronmeissner:
lgtm, though might be good to have @anujmaheshwari1 take a look as well

PR #4123, opened 2024-02-28
Updating the new bootstrap contract as needed

Devinwong:
**What type of PR is this?**

Add one of the following kinds:
/kind feature

Updating the new bootstrap contract as needed. This is still under development.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/akscustomcloudconfig.pb.go
tallaxes, line 1: What .proto does this correspond to? Leftover after rename?

pkg/proto/nbcontract/v1/config.proto
```diff
@@ -15,21 +15,24 @@ import "pkg/proto/nbcontract/v1/kubeletconfig.proto";
...
-  // only required until Secure TLS bootstrapping in place
-  // would use kubelet identity after that.
-  string tls_bootstrap_token = 2;
+  TLSBootstrappingConfig tls_bootstrapping_config = 2;
+  
 
   // cluster/user config
   string kubernetes_version = 3; // Q: can this be auto-detected? Or is this part of specifying the desired node version?
   KubeBinaryConfig kube_binary_config = 4;
+  optional CustomCloudConfig custom_cloud_config = 5;
```
tallaxes, line 35: At some point lets review for consistent use (or not) of *Config and _config suffixes

pkg/proto/nbcontract/v1/config.proto
```diff
@@ -56,10 +58,9 @@ message Configuration {
   bool is_sgx_node = 25;
   TeleportConfig teleport_config = 26;
   FeatureState ssh_status = 27;
-  FeatureState unattended_upgrade_status = 28;
+  optional FeatureState unattended_upgrade_status = 28;
```
tallaxes, line 61: Curios about rationale / criteria for using optional with Feature State ...

pkg/proto/nbcontract/v1/akscustomcloudconfig.pb.go
Devinwong: Oh right. Thanks for catching this. This is a leftover.

pkg/proto/nbcontract/v1/customcatrustconfig.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+package nbcontract.v1;
+import "pkg/proto/nbcontract/v1/featurestate.proto";
+
+message CustomCaTrustConfig {
+    FeatureState status = 1;
+    repeated string custom_ca_certs = 2;
+  }
```
tallaxes, line 8: Please check on best practices for casing abbreviations in this context (should it be CustomCATrustConfig?), and using plurals for repeated fields (should it be "cert" or "certs"). The current change may be right, just to double check.

pkg/proto/nbcontract/v1/config.proto
Devinwong: Sounds good. For now I try to add _config suffix for sub-groups. But can definitely review this once the contract is more stable.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
```diff
@@ -10,7 +10,7 @@ message HTTPProxyConfig {
   repeated string https_proxy_urls = 2;
 
   // NoProxy represents a list of endpoints including IPs, domains, or other network CIDRs that should not be accessed through the proxy
-  repeated string no_proxy = 3;
+  repeated string no_proxy_urls = 3;
```
tallaxes, line 13: NO_PROXY are not URLs

pkg/proto/nbcontract/v1/identityconfig.proto
```diff
@@ -12,6 +12,6 @@ message IdentityConfig {
 enum IdentityType {
   IDENTITY_TYPE_UNSPECIFIED = 0;
   IDENTITY_TYPE_SERVICE_PRINCIPAL = 1;
-  IDENTITY_TYPE_SYSTEM_IDENTITY = 2;
-  IDENTITY_TYPE_USER_IDENTITY = 3;
+  IDENTITY_TYPE_USER_IDENTITY = 2;
```
tallaxes, line 15: Is it best practice to prefix these with IDENTITY_TYPE_? Or would this be clear / already namespaces because of "IdentityType" enum?

pkg/proto/nbcontract/v1/config.proto
Devinwong: Feature state has an unspecified state. If it's not specified in the contract, it will be set as this state. In the VHD go binary, it will mostly handle this as "false".
I put this as optional because I think bootstrapper doesn't necessarily need to set this. If this is not set, it will be in unspecified state.
We will do a full review for all variables to see if they should be optional.
Does this sound make sense to you?

pkg/proto/nbcontract/v1/kubeletconfig.proto
```diff
@@ -1,15 +1,18 @@
 syntax = "proto3";
 package nbcontract.v1;
+import "pkg/proto/nbcontract/v1/featurestate.proto";
 
-message KubeletConfiguration {
+message KubeletConfig {
   // TODO(ace): remove these/make api defensible
-  repeated string kubelet_flags = 1;
+  map<string, string> kubelet_flags = 1;
```
tallaxes, line 7: Not disagreeing, but flagging this as an impactful design decision. Order might actually matter here, need to review kubelet flags to make a final call. (Also need a decision on supporting generic flags vs. breaking them out into individual fields.)

pkg/proto/nbcontract/v1/httpproxyconfig.proto
Devinwong: But it's NO_PROXY_URLS in the cse_cmd.sh
https://github.com/Azure/AgentBaker/blob/102df0a193cb00042749092c107185873964d7fd/parts/linux/cloud-init/artifacts/cse_cmd.sh#L115

pkg/proto/nbcontract/v1/kubeletconfig.proto
```diff
@@ -1,15 +1,18 @@
...
   map<string, string> kubelet_node_labels = 2;
   repeated Taint taints = 3;
   repeated Taint startup_taints = 4;
   KubeletDiskType kubelet_disk_type = 5;
-  bool kubelet_config_file_enabled = 6;
+  FeatureState kubelet_config_file_status = 6;
   string kubelet_config_file_content = 7;
```
tallaxes, line 13: Need a consistent pattern for this ... I presume "optional string kube_config_file_content" is not good enough for some reason?

pkg/proto/nbcontract/v1/kubeletconfig.proto
```diff
@@ -1,15 +1,18 @@
...
+  string kubelet_client_content = 8;
```
tallaxes, line 14: Maybe a better (more descriptive) name for this?

pkg/proto/nbcontract/v1/identityconfig.proto
Devinwong: hmm. This is good point. In Karpenter now we are using it as
_identityConfig.IdentityType = nbcontractv1.IdentityType_IDENTITY_TYPE_USER_IDENTITY_
So I think you are right. We can remove the prefix. I will review other enums too.

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+package nbcontract.v1;
+import "pkg/proto/nbcontract/v1/featurestate.proto";
+
+message TLSBootstrappingConfig {
+    // discuss whether this should be featureState or bool
+    // understand how defaulting works for enum, because we want default to be false
+    FeatureState tls_bootstrapping_status = 1;
+    FeatureState secure_tls_bootstrapping_status = 2;
```
tallaxes, line 9: Would something like bootstrapping mode enum make more sense? These are mutually exclusive, aren't they?

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -0,0 +1,12 @@
...
+    string tls_bootstrap_token = 3; // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
```
tallaxes, line 10: This is a good example of something currently required that we know will go away. Would be good to develop an evolution pattern for this scenario.

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -0,0 +1,12 @@
...
+    string custom_secure_tls_bootstrap_appserver_appid = 4;
```
tallaxes, line 11: This would be optional I presume?

pkg/proto/nbcontract/v1/httpproxyconfig.proto
tallaxes: Can you check on the actual use in the VHD? Could be a naming "bug" in the current use, that we can fix in the contract ...

pkg/proto/nbcontract/v1/config.proto
tallaxes: Need to think some more about it, but wondering if in this case we need explicit "unspecified" value ...

pkg/proto/nbcontract/v1/customcatrustconfig.proto
Devinwong: Here are what I tried.
If we put _CustomCaTrustConfig custom_ca_trust_config = 33;_ in config.proto, 
in the compiled .go, it will be _CustomCaTrustConfig           *CustomCaTrustConfig_
The type and name are consistent.

I don't have a strong preference on this. We can still do _CustomCATrustConfig_.
p.s., _**Field definitions in the .proto files must use lower_case_underscore_separated_names**_ is suggested by Google.

Thoughts?

Devinwong: Btw, plural form of repeated field is suggested, ie. certs.

pkg/proto/nbcontract/v1/config.proto
Devinwong: Oh I know why we needed to keep the prefix.
For example, if I put

`enum NetworkPluginType {
  UNSPECIFIED = 0;
  NONE = 1;
  AZURE = 2;
  KUBENET = 3;
}

enum NetworkPolicyType {
  UNSPECIFIED = 0;
  NONE = 1;
  AZURE = 2;
  CALICO = 3;
}`


It will complain this.
`symbol "nbcontract.v1.AZURE" already defined at pkg/proto/nbcontract/v1/networkconfig.proto:32:3; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum`

pkg/proto/nbcontract/v1/identityconfig.proto
Devinwong: Oh I know why we needed to keep the prefix.
For example, if I put

`enum NetworkPluginType {
UNSPECIFIED = 0;
NONE = 1;
AZURE = 2;
KUBENET = 3;
}

enum NetworkPolicyType {
UNSPECIFIED = 0;
NONE = 1;
AZURE = 2;
CALICO = 3;
}`

It will complain this.
`symbol "nbcontract.v1.AZURE" already defined at pkg/proto/nbcontract/v1/networkconfig.proto:32:3; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum`

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: Sorry, are you referring to kubelet_config_file_status or string kube_config_file_content, or both? Could you please elaborate on consistent pattern here?

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: Does kubelet_client_key look more descriptive?

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
Devinwong: Both variables are in the cse_cmd.sh, so I put them as is.
I did a quick search and find that there is a condition in cse_config.sh that
`if [ "${ENABLE_SECURE_TLS_BOOTSTRAPPING}" == "true" ] || [ "${ENABLE_TLS_BOOTSTRAPPING}" == "true" ]; then`
So it seems they are not mutually exclusive.
I left a comment to follow up with this in the next milestone when we are extracting the logics. We will further follow up with the feature owner to understand more.

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
Devinwong: Will do, as explained in the previous comment.

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
Devinwong: We will further follow up with the feature owner to understand more about the TLS stuff.

pkg/proto/nbcontract/v1/config.proto
Devinwong: hmm. Per protobuff doc,
Do Include an Unspecified Value in an Enum

pkg/proto/nbcontract/v1/httpproxyconfig.proto
Devinwong: yep I did. In cse_config.sh, it writes the NO_PROXY_URLS to /etc/systemd/system.conf.d/proxy.conf.
And usually the proxy.conf looks like this,
`{
  "proxies": {
    "http-proxy": "http://proxy.example.com:3128",
    "https-proxy": "https://proxy.example.com:3129",
    "no-proxy": "*.test.example.com,.example.org,127.0.0.0/8"
  }
}`
no-proxy should be a list of URLs

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: Okay I didn't realize it's in alphabetical order. I will check with the team to understand why it's needed.
Assuming it's needed, then I think you are right, I should change it back to a list of string.
What I thought was that as the format of kubelet_flags are key-value pairs, using string map should be more suitable.

pkg/proto/nbcontract/v1/config.proto
tallaxes: Sounds good. Let's keep it this way and see if it feels right / we understand the tradeoffs.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
tallaxes: Nope, those entries in NO_PROXY are not [URLs](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax). :-) Domains, domain suffixes (with or without regex), IP addresses, CIDR blocks etc. - the full list is complicated - not URLs.

pkg/proto/nbcontract/v1/kubeletconfig.proto
tallaxes: Alphabetical order? When kubelet gets these as command line arguments, the order would matter - though probalby only if there are duplicates, which would not be the case with map. Modeling this is as map is probably fine - especially considering KubeletConfig exists, and does essentially the same. Alternatively, we could consider breaking them out - though evolving this would be challenging.

pkg/proto/nbcontract/v1/kubeletconfig.proto
tallaxes: I mean for a pair together - FeatureState indicating if something is present, paired with associated data. Elsewhere you have used separate messages for this scenario, I think.

pkg/proto/nbcontract/v1/kubeletconfig.proto
tallaxes: Yes - assuming that's what it is :-)

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
tallaxes: > Both variables are in the cse_cmd.sh, so I put them as is.

That should not be a constraint, you can generate cse_cmd.sh correctly even if Go data structure is shaped differently (e.g. using an expression in the template). But keeping both closely aligned may be ok at this state - if we plan to revisit it later.

> So it seems they are not mutually exclusive.

I am fairly certain they are - in a sense they will never be both true at the same time. There are things that need to happen in either case, thus the logic in the script.

pkg/proto/nbcontract/v1/customcatrustconfig.proto
tallaxes: > I don't have a strong preference on this. We can still do CustomCATrustConfig.

Thats what Golang style for [initialisms](https://google.github.io/styleguide/go/decisions.html#initialisms) in identifiers prescribes. (This is often checked by linters too ...)

pkg/proto/nbcontract/v1/httpproxyconfig.proto
Devinwong: I see. Didn't realize the difference. I will use no_proxy_entries. Does this sound good to you?
Will update other scripts as well when we modify them later.

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: In this line, it seems it's doing an alphabetical ordering to the keys. https://github.com/Azure/AgentBaker/blob/df3496f1dac695cc13693096cb6fa6ac3a88a4c5/pkg/agent/utils.go#L373
I didn't compile it to see the results though.
As for breaking them out, I think that will be considered in our next step.

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: Oh you are referring to the status, not the content. Actually I should put a comment here that we will remove this status because it's only checking if the file content is present. I've noted it the design doc but let me add a comment here too.

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: yes it is

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
Devinwong: Okay. Will review the TLS things together in this coming milestone.

pkg/proto/nbcontract/v1/httpproxyconfig.proto
tallaxes: Yes, `no_proxy_entries` would work - as long as we decide to break them into parts, which I think is worth trying. (Otherwise I would just keep it as a single `no_proxy` string ... `http_proxy`, `https_proxy` and `no_proxy` are well-known environment variable names ...)

Which brings me to this (and apologies for not noticing it earlier): for `http_proxy` and `https_proxy`, neither "repeated", nor "_urls" (plural) suffix make sense. It needs to be a single URL for each (unless I am missing something major). So either just `http_proxy` and `https_proxy` (preferred, IMO), or `http_proxy_url` and `https_proxy_url`. In either case, not repeated.

pkg/proto/nbcontract/v1/kubeletconfig.proto
tallaxes: Ok, I think we can take as yet another datapoint showing (perhaps counterintuitively?) that order does not actually matter. For actual sorting, I suspect AgentBaker does it just to make the resulting CSE deterministic, maybe even just for testability?

tallaxes (approved):
LGTM, with some remaining feedback (mostly around *_proxy fields).

pkg/proto/nbcontract/v1/identityconfig.proto
tallaxes: protobuf spec [for enums](https://protobuf.dev/reference/go/go-generated/#enum) says constants generated for enums _within a message_ are prefixed with the message name automatically. Worth a try?

pkg/proto/nbcontract/v1/httpproxyconfig.proto
Devinwong: That's a good catch! It doesn't make sense to have multiple http proxy. I changed the contract variable names to `http_proxy` and `https_proxy`. Will further update the bootstrap scripts later on.

pkg/proto/nbcontract/v1/identityconfig.proto
Devinwong: I tried this.

```
message NetworkConfig {
  NetworkPluginType network_plugin = 1;
  NetworkPolicyType network_policy = 2;
  NetworkModeType network_mode = 3;
  
  enum NetworkPluginType {
    UNSPECIFIED = 0;
    NONE = 1;
    AZURE = 2;
    KUBENET = 3;
  }

  enum NetworkPolicyType {
    UNSPECIFIED = 0;
    NONE = 1;
    AZURE = 2;
    CALICO = 3;
  }
}
```

It still complains 
```
pkg/proto/nbcontract/v1/networkconfig.proto:24:5:symbol "nbcontract.v1.NetworkConfig.UNSPECIFIED" already defined at pkg/proto/nbcontract/v1/networkconfig.proto:17:5; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum
pkg/proto/nbcontract/v1/networkconfig.proto:25:5:symbol "nbcontract.v1.NetworkConfig.NONE" already defined at pkg/proto/nbcontract/v1/networkconfig.proto:18:5; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum
pkg/proto/nbcontract/v1/networkconfig.proto:26:5:symbol "nbcontract.v1.NetworkConfig.AZURE" already defined at pkg/proto/nbcontract/v1/networkconfig.proto:19:5; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum
make: *** [Makefile:90: compile-proto-files] Error 100
```

pkg/proto/nbcontract/v1/kubeletconfig.proto
Devinwong: Could be. Let me check with the team to see if anyone knows.

Devinwong: Team thinks it could be just for human readability to know what flags are present. Let's keep it as is and we will review this in this milestone.

tallaxes: Sure. 

(Surprised testability is not the reason. Doesn't AB have tests where it compares what gets generated with expected output - including CustomData, like [this one](https://github.com/Azure/AgentBaker/blob/9d9d6a659c7af00c5d8bc6816d66634ac7a58410/pkg/agent/testdata/AKSUbuntu2204%2BSecureTLSBoostrapping/CustomData#L263)? Comparing the results won't work unless the order in the output is deterministic - the same every time - and sorting is a way to enforce it ...)

Devinwong: Your point could be another reason. I didn't (know how to) run the test yet though.

lilypan26 (approved):

pkg/proto/nbcontract/v1/httpproxyconfig.proto
```diff
@@ -3,14 +3,14 @@ package nbcontract.v1;
...
+  // HTTPSProxy represents the proxy endpoint/URL using HTTPS, that the traffic should go through for making HTTPS connections outside the cluster
+  string https_proxy = 2;
 
-  // NoProxy represents a list of endpoints including IPs, domains, or other network CIDRs that should not be accessed through the proxy
-  repeated string no_proxy = 3;
+  // NoProxy represents a list of entries including IPs, domains, or other network CIDRs that should not be accessed through the proxy
+  repeated string no_proxy_entries = 3;
 
   // TrustedCA repesents alternative CA certificate content
   optional string proxy_trusted_ca = 4;
```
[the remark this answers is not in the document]
Devinwong: In windows, there is a feature where you can leave the http proxy url there but turn off the proxy setting.
image
But that's windows.
In Iinux, it seems we usually just remove the http proxy url to disable it, other than having a switch to turn it off.

I did quick a search in baker.go, there are 3 functions related to http proxy status, which are
```
		"ShouldConfigureHTTPProxy": func() bool {
			return config.HTTPProxyConfig != nil && (config.HTTPProxyConfig.HTTPProxy != nil || config.HTTPProxyConfig.HTTPSProxy != nil)
		},
		"HasHTTPProxy": func() bool {
			return config.HTTPProxyConfig != nil && config.HTTPProxyConfig.HTTPProxy != nil
		},
		"HasHTTPSProxy": func() bool {
			return config.HTTPProxyConfig != nil && config.HTTPProxyConfig.HTTPSProxy != nil
		},
```

So actually they are just checking if the config is present, then http proxy is on.
So I think in the coming milestone, we can remove the status as you pointed out. 

Similar thing happens to CA.
```
"ShouldConfigureHTTPProxyCA": func() bool {
			return config.HTTPProxyConfig != nil && config.HTTPProxyConfig.TrustedCA != nil
		},
```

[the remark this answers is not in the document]
Devinwong: Left comments to track this later.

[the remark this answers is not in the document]
lilypan26: Okay yep, thought the same thing!

PR #4126, opened 2024-02-29
perf: Parallelize downloadContainerdWasmShims function / fix download error bug

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR changes the execution of downloadContainerdWasmShims to make use of parallel processing. It increases the speed of the function by 5 minutes on average.

This is one of a variety of incremental performance improvements that will ultimately increase the speed of VHD Build times.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -102,12 +103,16 @@ downloadContainerdWasmShims() {
...
+            retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-wws-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-wws-v1" 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT) &
+            pids+=($!)
             chmod 755 "$containerd_wasm_filepath/containerd-shim-spin-${binary_version}-v1"
             chmod 755 "$containerd_wasm_filepath/containerd-shim-slight-${binary_version}-v1"
-            chmod 755 "$containerd_wasm_filepath/containerd-shim-wws-${binary_version}-v1"
         fi
     done
+    wait ${pids[@]}
+    for shim_version in $CONTAINERD_WASM_VERSIONS; do
+        chmod 755 "$containerd_wasm_filepath/containerd-shim-wws-${binary_version}-v1"
```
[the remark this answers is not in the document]
zachary-bailey: @lilypan26 Hey! Thank you for the tip on binary_version. I initially took it out because I read the variable would still be in scope, but didn't take into account that its value needs to change as the for loop iterates so it can set permissions on each version.

For the permissions part, the first two downloads (spin / slight) are running sequentially and complete in about a second, so they will definitely be downloaded by the time the program reaches the code that sets permissions for their directories. For the last permission, I had to ensure that the 3 background processes (downloading shim_wws x 3) were complete before reaching that code, otherwise it could attempt to set permissions on a directory that doesn't exist on the VHD yet. I currently only have wws parallelized because downloading the first two and setting their permissions only takes about a second for each binary version.

If you think it makes more sense to parallelize everything for speed or stylistic reasons, please let me know! And thanks again.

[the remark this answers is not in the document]
lilypan26: ah i see, then yea for safety we should only access the directory after we are sure it is created. 

Although for spin and slight, even if they are much faster than wws i think it wouldn't hurt to parallelize those are well and then move all of the chmods to the for loop afterwards. This might be more consistent and readable, but up to you. If not, i would at least add a comment where you are downloading wws and explain that you are parallelizing wws download and only changing the permissions after you are sure the file is there.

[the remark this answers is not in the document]
zachary-bailey: @lilypan26 I implemented your suggestions, thanks!

That being said, I noticed a flakey little error that has been happening for some time now (before my changes) regarding setting permissions. Im looking into it then will submit another PR.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -93,19 +93,22 @@ downloadSecureTLSBootstrapKubeletExecPlugin() {
...
             retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-spin-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-spin-v1" 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT)
             retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-slight-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-slight-v1" 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT)
-            retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-wws-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-wws-v1" 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT)
+            if [ "$shim_version" = "v0.8.0" ]; then
+                # Only download the wws shim for v0.8.0, wws shim for v0.3.0 and v0.5.1 fail
```
ganeshkumarashok, line 106: Can we remove those two versions in a separate PR? And share it with @Mossaka for a review?

ganeshkumarashok: Separate PR for parallelization vs removal of component downloads that 404. 
The latter might need a cleanup of the containerd config too.

ganeshkumarashok: Deleting the versions that fail during the download would be cleaner as those versions will not be used

zachary-bailey:
@Mossaka Good morning, 

I am the engineer working with @ganeshkumarashok to decrease VHD build times in our pipelines. 

Can you review this PR and let me know what you think? I removed v0-3-0 and v0-5-1 in baker.go, baker_test.go, and the cse_install.sh file. This is meant to be a fix until the appropriate versions can be put in. Additionally, I am running the downloadContainerdWasmShims function in parallel now in order to increase speed. It passed in the VHD Build Test pipeline with no errors. 

The way the function and configs are currently written, the function fails for 5 minutes while it attempts to download wws-shims for v0-3-0 and v0-5-1. Please let me know if you have an input or suggestions! Thanks.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -16,7 +16,7 @@ SECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_DOWNLOAD_DIR="/opt/azure/tlsbootstrap"
 SECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_VERSION="v0.1.0-alpha.2"
 TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
-CONTAINERD_WASM_VERSIONS="v0.3.0 v0.5.1 v0.8.0"
+CONTAINERD_WASM_VERSIONS="v0.8.0"
```
Mossaka, line 19: Could we have a follow-up PR to upgrade the shim to the [latest version](https://github.com/deislabs/containerd-wasm-shims/releases/tag/v0.11.1)?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -16,7 +16,7 @@ SECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_DOWNLOAD_DIR="/opt/azure/tlsbootstrap"
 SECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_VERSION="v0.1.0-alpha.2"
 TELEPORTD_PLUGIN_DOWNLOAD_DIR="/opt/teleportd/downloads"
 TELEPORTD_PLUGIN_BIN_DIR="/usr/local/bin"
-CONTAINERD_WASM_VERSIONS="v0.3.0 v0.5.1 v0.8.0"
```
Mossaka, line 19: Do we have any users on the older versions of the shim, say 0.3.0? By removing the older versions, are we going to break their existing workflows? If so, should we be more cautious about deprecating older versions and probably notife customers beforehand? 

I am fine with deprecation in general but I prefer putting a notification to customers first.

Mossaka (commented):
Added a few comments.

parts/linux/cloud-init/artifacts/cse_install.sh
zachary-bailey: Do you want v0-8-0 switched to v0-11-1? Or do you want us to bump all versions up like below:

v0-3-0 ==> vX-X-X
v0-5-1 ==> vX-X-X
v0-8-0 ==> vX-X-X

If so, which versions would map to the older ones? 

We are just looking to get the failing curl command to wws for v0-3-0 and v0-5-1 out of the code, as its adding 5 minutes to our build time due to retries.

zachary-bailey: Thank you for the review

parts/linux/cloud-init/artifacts/cse_install.sh
ganeshkumarashok: These older versions are getting a 404 on downloads during the VHD builds. Customers won't be affected/already must be affected in that case right?

ganeshkumarashok: Only case I can think of is users migrating from an older version working earlier to a new nodepool/VHD image that has the newer version. What would the expected behavior for that be, before this change (as that path existed, because the older version was not available for a while)?

ganeshkumarashok: For the notification to users, adding a note in the AKS release notes should be sufficient right?

zachary-bailey: The spin and slight are downloading successfully for v0-3-0 and v0-5-1, but not wws, which makes me think that wws isn't available for those.
 
Is it possible they are still using spin / slight for v0-3-0 and v0-5-1 without wws? @Mossaka 

If not, then I don't think any customers would have been able to use those versions for quite some time.

Mossaka: > Is it possible they are still using spin / slight for v0-3-0 and v0-5-1 without wws?

Yes, we should remove wws from v0.3.0 and v0.5.0 because it wasn't introduced for these versions

Mossaka: maybe what you need is a condition logic saying that if the version if 0-3-0 or 0-5-0, don't install wws shim

zachary-bailey: Alrighty thanks!

zachary-bailey: @Mossaka Good morning! I have implemented your suggestions, adding the v3 and v5 versions back in the baker.go/baker_test.go files and creating the conditional in cse_install. Would you mind giving this PR another look?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -91,20 +91,32 @@ downloadSecureTLSBootstrapKubeletExecPlugin() {
...
-            chmod 755 "$containerd_wasm_filepath/containerd-shim-spin-${binary_version}-v1"
-            chmod 755 "$containerd_wasm_filepath/containerd-shim-slight-${binary_version}-v1"
+            retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-spin-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-spin-v1" 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT) &
+            wasmShimPids+=($!)
+            retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-slight-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-slight-v1" 2>&1 | tee $CURL_OUTPUT >/dev/null | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT) &
+            wasmShimPids+=($!)
+            if [ "$shim_version" = "v0.8.0" ]; then
```
cameronmeissner, line 108: nit: let's try to be consistent with `==` when comparing strings

ganeshkumarashok:
Agentbaker e2e docs for testing further: https://github.com/Azure/AgentBaker/tree/master/e2e

ganeshkumarashok (approved):
Approved but would also get an approval from @Mossaka

Mossaka (approved):
LGTM

PR #4132, opened 2024-03-04
fix: install missing kubenet ebtables drop rule

wedaly:
**What type of PR is this?**

/kind bug

In kubenet, the script ensure-no-dup.sh is supposed to install ebtables rules to prevent traffic to service IPs from being duplicated when the bridge is in promiscuous mode.

The script reads the pod CIDR from the CNI conflist. However, this broke when the CNI conflist format changed to support kubenet dual-stack in commit 869183619373769ae24e130afeef3f0dd865c6ef

Since the field `ipam.subnet` no longer exists in the CNI conflist, the jq expression was returning "null", which was rejected by ebtables. The result is that when the client and server pod are on the same node, and the client sends traffic to the service VIP, the server will see duplicate packets (the same issue described here: https://github.com/kubernetes/kubernetes/issues/25793)

This commit fixes the issue by updating the jq expression to read `.ipam.ranges[0][0].subnet` instead.

This addresses the immediate issue for IPv4, but longer-term we need to:
1. Test whether IPv6 is also seeing this issue (likely yes)
2. Consider configuring the bridge plugin to use hairpinMode=true, promiscMode=false and removing the ensure-no-dup script entirely.

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
install missing kubenet ebtables drop rule to fix duplicate packets between pods on the same node when using service cluster IP
```


parts/linux/cloud-init/artifacts/ensure-no-dup.sh
```diff
@@ -35,8 +35,8 @@ if [[ -z "${bridgeIP}" ]]; then
     exit 1
 fi
 
-podSubnetAddr=$(cat /etc/cni/net.d/10-containerd-net.conflist  | jq -r ".plugins[] | select(.type == \"bridge\") | .ipam.subnet")
-if [[ -z "${podSubnetAddr}" ]]; then
+podSubnetAddr=$(cat /etc/cni/net.d/10-containerd-net.conflist  | jq -r ".plugins[] | select(.type == \"bridge\") | .ipam.ranges[0][0].subnet")
+if [[ -z "${podSubnetAddr}" || "${podSubnetAddr}" == 'null' ]]; then
```
paulgmiller, line 39: comment about the parsing above and why we get null sometimes.

parts/linux/cloud-init/artifacts/ensure-no-dup.sh
```diff
@@ -35,8 +35,8 @@ if [[ -z "${bridgeIP}" ]]; then
     exit 1
 fi
 
-podSubnetAddr=$(cat /etc/cni/net.d/10-containerd-net.conflist  | jq -r ".plugins[] | select(.type == \"bridge\") | .ipam.subnet")
-if [[ -z "${podSubnetAddr}" ]]; then
```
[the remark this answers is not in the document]
wedaly: unfortunately, no. The way it works today is that cloud-controller manager assigns the node a pod CIDR. Kublet/containerd watch the node, and once it sees the pod CIDR it generates the conflist. The ensure-no-dup.sh script runs in a loop waiting for the conflist to appear so it can read the pod CIDR.

I think longer-term I'd want to see if we can remove this script entirely. From what I saw in K8s PRs from 2017, the bridge was set to promiscuous mode as a workaround for a kernel bug with hairpin mode.

paulgmiller (approved):
Okay wit this but also okay ripping out and publishing a github issue and encoruaging peopel to move to overlay instead.

parts/linux/cloud-init/artifacts/ensure-no-dup.sh
wedaly: comment added!

ganeshkumarashok:
What test can verify this functionality so we are sure this change doesn't cause issues and catch the current issue too?

ganeshkumarashok:
Is it possible to add such a test in agentbaker e2e?

wedaly:
good call, added E2E tests to validate that the ebtables rules are added. I put it in ubuntu2204 and mariner scenarios, since those use kubenet, but please let me know if you'd prefer me to create new scenarios @ganeshkumarashok

ganeshkumarashok (approved):

PR #4136, opened 2024-03-05
feat: Add airgap nsg to abe2e

AlisonB319:
**What type of PR is this?**
Adding a new nodepool type option where airgap networking settings are installed after the cluster creates. New e2e tests that specify 'airgap: True' will use the airgap nodepool.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


e2e/cluster.go
```diff
@@ -198,6 +201,24 @@ func getClusterSubnetID(ctx context.Context, cloud *azureClient, location, mcRes
 	return "", fmt.Errorf("failed to find aks vnet")
 }
 
+func getClusterVnetName(ctx context.Context, cloud *azureClient, mcResourceGroupName string) (string, error) {
```
cameronmeissner, line 204: nit: `getClusterVNetName`

e2e/cluster.go
```diff
@@ -198,6 +201,24 @@ func getClusterSubnetID(ctx context.Context, cloud *azureClient, location, mcRes
...
+	pager := cloud.vnetClient.NewListPager(mcResourceGroupName, nil)
+
+	for pager.More() {
+		nextResult, err := pager.NextPage(ctx)
+		if err != nil {
+			return "", fmt.Errorf("failed to advance page: %w", err)
+		}
+		for _, v := range nextResult.Value {
+			if v == nil {
+				return "", fmt.Errorf("aks vnet id was empty")
```
cameronmeissner, line 214: seems like a lot of code duplication with `getClusterSubnetID`, can we maybe refactor towards returning a common struct which contains both vnet name and subnet ID since they're both based on the same vnet?

e2e/cluster.go
```diff
@@ -226,12 +247,20 @@ func getInitialClusterConfigs(ctx context.Context, cloud *azureClient, resourceG
...
 
-				log.Printf("found agentbaker e2e cluster %q in provisioning state %q", *resource.Name, *cluster.Properties.ProvisioningState)
-				configs = append(configs, clusterConfig{cluster: &cluster.ManagedCluster})
+				clusterConfig := clusterConfig{cluster: &cluster.ManagedCluster}
+				isAirgap, err := isNetworkSecurityGroupAirgap(cloud, *clusterConfig.cluster.Properties.NodeResourceGroup)
+				if err != nil {
+					return nil, fmt.Errorf("failed to verify if aks subnet is for an airgap cluster: %w", err)
+				}
+
+				if isAirgap {
```
cameronmeissner, line 256: cant just set `clusterConfig.isAirgapCluster = isAirgap` no?

e2e/cluster.go
```diff
@@ -244,19 +273,31 @@ func hasViableConfig(scenario *scenario.Scenario, clusterConfigs []clusterConfig
...
-	cloud *azureClient,
-	suiteConfig *suite.Config,
-	scenarios scenario.Table,
-	clusterConfigs *[]clusterConfig) error {
+func hasViableAirgapConfig(scenario *scenario.Scenario, clusterConfigs []clusterConfig) bool {
+	if !scenario.Airgap {
+		return true // config not needed for scenario
+	}
+	for _, config := range clusterConfigs {
+		if config.isAirgapCluster && scenario.Config.ClusterSelector(config.cluster) {
```
cameronmeissner, line 281: ideally would like to keep things consistent with how we select on cluster properties - maybe we should refactor towards having cluster selectors/mutators work on cluster configs rather than individual cluster models

cameronmeissner: also can't we just merge the airgap logic here into the existing `hasViableConfig`?  this seems like a lot of duplication

e2e/network.go
```diff
@@ -0,0 +1,158 @@
...
+
+var (
+	nsgName           string = "abe2e-airgap-securityGroup"
+	defaultSubnetName string = "aks-subnet"
+)
+
+func cloudGapSecurityGroup(location, kubernetesEndpont string) armnetwork.SecurityGroup {
+	securityRules := []*armnetwork.SecurityRule{}
+
+	securityRules = append(securityRules, getSecurityRule("allow-mcr-microsoft-com", "204.79.197.219", 100))
```
cameronmeissner, line 21: nit: use composite literal instead of excessive appends

e2e/network.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
+)
+
+var (
+	nsgName           string = "abe2e-airgap-securityGroup"
+	defaultSubnetName string = "aks-subnet"
+)
+
+func cloudGapSecurityGroup(location, kubernetesEndpont string) armnetwork.SecurityGroup {
```
cameronmeissner, line 18: cloudGap or airGap?

e2e/cluster.go
AlisonB319: The createMissingClusters logic makes things a bit confusing which is why I made the hasViableAirgapConfig - but let me take another look at it since I wrote this section last week when I was figuring a lot of stuff out

e2e/network.go
cameronmeissner, line 1: we should probably have a network package which exposes airgap-related functionality as well as any other network-related functionality we might need later - for now I'd prefer to rename this to `airgap.go` since most all of this logic is airgap-specific

e2e/scenario/scenario_ubuntu2204-airgap.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+)
+
+func (t *Template) ubuntu2204AirGap() *Scenario {
+	return &Scenario{
+		Name:        "ubuntu2204-airgap",
+		Description: "Tests that a node using the Ubuntu 2204 VHD and is airgap can be properly bootstrapped",
+		Config: Config{
+			ClusterSelector: NetworkPluginKubenetSelector,
```
cameronmeissner, line 12: yeah, ideally we would have airgap cluster selectors/mutators - a good follow-up item

e2e/cluster.go
cameronmeissner: fair enough, as I said before I'd like to do a lot of refactoring here some day :)

e2e/network.go
AlisonB319: airgap - definitely was blind to this name since it was the first function I started with lol

e2e/airgap.go
```diff
@@ -0,0 +1,150 @@
...
+			SourcePortRange:          to.Ptr("*"),
+			DestinationAddressPrefix: to.Ptr(destinationAddressPrefix),
+			DestinationPortRange:     to.Ptr("*"),
+			Priority:                 to.Ptr[int32](priority),
+		},
+	}
+}
+
+func addAirgapNetworkSettings(ctx context.Context, cloud *azureClient, suiteConfig *suite.Config, clusterConfig clusterConfig) error {
+	fmt.Printf("Adding network settings for airgap cluster %s in rg %s\n", *clusterConfig.cluster.Name, *clusterConfig.cluster.Properties.NodeResourceGroup)
```
cameronmeissner, line 80: nit: let's at least use the `log` pkg here instead

e2e/vmss.go
```diff
@@ -23,8 +23,13 @@ const (
 	vmssNameTemplate                         = "abtest%s"
 	listVMSSNetworkInterfaceURLTemplate      = "https://management.azure.com/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/virtualMachineScaleSets/%s/virtualMachines/%d/networkInterfaces?api-version=2018-10-01"
 	loadBalancerBackendAddressPoolIDTemplate = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/loadBalancers/kubernetes/backendAddressPools/aksOutboundBackendPool"
+	maxRetries                               = 3
 )
 
+type RetryVmssOperations struct {
```
cameronmeissner, line 29: can we call this something like `VMSSOperationRetrier` or similar?

.pipelines/e2e.yaml
```diff
@@ -25,6 +25,7 @@ pool:
 
 jobs:
 - job: Run_AgentBaker_E2E
+  timeoutInMinutes: 90
```
cameronmeissner, line 28: we already specify a timeout below, do we need this as well? we also want to make sure to keep the local timeout the same (e.g. what's in e2e-local.sh)

cameronmeissner: if that's the case can we just get rid of the timeout from the `go test ... ` cmd? should be fine if we're allowing more than 60 minutes for these suites, and less places to have to update the better

e2e/scenario/scenario_ubuntu2204-airgap.go
cameronmeissner: can you create a work item for this?

e2e/pollers.go
```diff
@@ -27,14 +27,14 @@ const (
...
+	waitUntilPodRunningPollInterval         = 10 * time.Second
+	waitUntilPodDeletedPollInterval         = 10 * time.Second
 	waitUntilClusterNotCreatingPollInterval = 10 * time.Second
-	waitUntilNodeReadyPollingInterval       = 5 * time.Second
+	waitUntilNodeReadyPollingInterval       = 10 * time.Second
 
 	// Polling timeouts
 	vmssClientCreateVMSSPollingTimeout     = 10 * time.Minute
-	deleteVMSSPollingTimeout               = 5 * time.Minute
+	deleteVMSSPollingTimeout               = 10 * time.Minute
```
cameronmeissner, line 37: im not sure increasing polling timeouts is what we wanna do here, ideally we'd poll for _less_ time so we can have a chance of retrying before the test times out

e2e/vmss.go
```diff
@@ -44,15 +49,16 @@ func bootstrapVMSS(ctx context.Context, t *testing.T, r *mrand.Rand, vmssName st
...
+	vmssModel, err := rOpts.createVMSSWithPayload(ctx, nodeBootstrapping.CustomData, nodeBootstrapping.CSE, vmssName, publicKeyBytes, opts)
 	if err != nil {
 		return nil, cleanupVMSS, fmt.Errorf("unable to create VMSS with payload: %w", err)
 	}
 
 	return vmssModel, cleanupVMSS, nil
 }
 
-func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName string, publicKeyBytes []byte, opts *scenarioRunOpts) (*armcompute.VirtualMachineScaleSet, error) {
+func (rOpts VMSSOperationRetrier) createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName string, publicKeyBytes []byte, opts *scenarioRunOpts) (*armcompute.VirtualMachineScaleSet, error) {
```
cameronmeissner, line 61: nit: receiver var should be `r`

e2e/pollers.go
cameronmeissner: probably fine for now since this is for deletion, though this entire "polling" framework/unit we've made here seriously needs to be revisited at some point

.pipelines/e2e.yaml
AlisonB319: It seems the default timeout for go test is 10 minutes, so I think we need to keep both values. But I can make a variable so that it uses only one value

e2e/pollers.go
AlisonB319: Yeah I landed on 10, just because even with more retries it was just getting context exceeded.

.pipelines/e2e.yaml
cameronmeissner: ah didn't realize there was a default timeout, makes sense

either way is fine then

cameronmeissner (approved):

e2e/scenario/scenario_ubuntu2204-airgap.go
AlisonB319: https://dev.azure.com/msazure/CloudNativeCompute/_workitems/edit/27242881

PR #4145, opened 2024-03-08
feat: add outbound networking type to NodeBootstrappingConfiguration

AlisonB319:
**What type of PR is this?**
Adding outbound networking type to NodeBootstrappingConfiguration so that it can be used in variables.go so that an egress call is not made when outbound = block or none

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/variables.go
```diff
@@ -178,6 +178,10 @@ func getOutBoundCmd(nbc *datamodel.NodeBootstrappingConfiguration, cloudSpecConf
 		return ""
 	}
 
+	if nbc.OutboundType == "none" || nbc.OutboundType == "block" {
```
cameronmeissner, line 181: string is fine but wondering if it would be better to use an enum type or something, e.g.

package datamodel

...

type OutboundType int

const (
    OutboundTypeNone = iota
    OutboundTypeBlock
)

cameronmeissner: actually using string is easier since we can more easily denote RP having not set it at all, so I'm fine with that

cameronmeissner: though you could make it work by having the unspecified type be the initial `iota`, such as:

https://github.com/Azure/AgentBaker/blob/master/pkg/agent/datamodel/types.go#L1680-L1686

pkg/agent/datamodel/types.go
```diff
@@ -122,6 +122,14 @@ const (
 	WasmWasi WorkloadRuntime = "WasmWasi"
 )
 
+// OutboundType describes the options for outbound internet access.
+type OutboundType int
+
+const (
+	OutboundTypeNone OutboundType = iota
```
cameronmeissner, line 129: if you wanna use the enum then start with"

`OutboundTypeUnspecified = iota` instead

cameronmeissner: (since this will have value 0 and the zero-value of int is also 0)

pkg/agent/datamodel/types.go
```diff
@@ -122,6 +122,14 @@ const (
 	WasmWasi WorkloadRuntime = "WasmWasi"
 )
 
+// OutboundType describes the options for outbound internet access.
+type OutboundType string
+
+const (
+	OutboundTypeNone  string = "none"
```
cameronmeissner, line 129: you should use `OutboundType` as the type here since you define it right above

OutboundType == string, though string != OutboundType

pkg/agent/datamodel/types.go
```diff
@@ -1675,6 +1683,7 @@ type NodeBootstrappingConfiguration struct {
 	DisableUnattendedUpgrades              bool
 	SSHStatus                              SSHStatus
 	DisableCustomData                      bool
+	OutboundType                           string
```
cameronmeissner, line 1686: this should also be `OutboundType` - RP will be able to use this type as well

pkg/agent/variables.go
```diff
@@ -178,6 +178,10 @@ func getOutBoundCmd(nbc *datamodel.NodeBootstrappingConfiguration, cloudSpecConf
 		return ""
 	}
 
+	if strings.ToLower(nbc.OutboundType) == datamodel.OutboundTypeBlock || strings.ToLower(nbc.OutboundType) == datamodel.OutboundTypeNone {
```
cameronmeissner, line 181: probably better to just use `strings.EqualFold(nbc.OutboundType, datamodel.OutboundTypeBlock)`, etc.

pkg/agent/datamodel/types.go
AlisonB319: Right - since 

nbc.OutboundType is going to receive a string - it might be better to get rid of the OutboundType all together and just use strings - since I think the check would have to look like 

strings.EqualFold(nbc.OutboundType, string(datamodel.OutboundTypeBlock)) || strings.EqualFold(nbc.OutboundType, string(datamodel.OutboundTypeNone))

pkg/agent/datamodel/types.go
AlisonB319: I think OutboundType should stay a string, because the pro of going with a string allows for RP to pass in a new string without us needing to add anything on the rp side. If we want to use the OutboundType it probably makes more since to go back to the iota

cameronmeissner: I agree, was just confused since you were previously defining the type but not using it

cameronmeissner (approved):

PR #4147, opened 2024-03-10
feat: add node graceful shutdown kubelet config

weichch:
**What type of PR is this?**

/kind feature
/kind api-change

This PR adds node graceful shutdown Kubelet config items and exposes those for client use, so that we can allow customers to enable this feature on their nodes.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner:
you'll want to recreate this PR without using a fork so required checks can run - I'm gonna add a note about this to the readme so folks don't continue to make the same mistake

pkg/agent/datamodel/types.go
```diff
@@ -2033,6 +2035,22 @@ type AKSKubeletConfiguration struct {
 	Default: []
 	+optional. */
 	AllowedUnsafeSysctls []string `json:"allowedUnsafeSysctls,omitempty"`
+	// shutdownGracePeriod specifies the total duration that the node should delay the
+	// shutdown and total grace period for pod termination during a node shutdown.
+	// Default: "0s"
+	// +featureGate=GracefulNodeShutdown
```
cameronmeissner, line 2041: we'll need to make sure that RP is setteing the `GracefulNodeShutdown` feature gate

weichch: Feature gate is by default enabled in k8s 1.21+, and we can't edit nodes to have this config. Do we still need to enable the feature gate in this case?

weichch:
I'm closing this PR as per comment. Will recreate it from another branch within the repo.

PR #4148, opened 2024-03-11
feat: set --node-ip for Windows kubelet in k8s v1.29+

AbelHu:
**What type of PR is this?**
/kind feature

set --node-ip for Windows kubelet in k8s v1.29+
kubelet in k8s v1.29+ cannot get the node ip by itself. Reference: [kubelet: cloud-provider external addresses by aojea · Pull Request #121028 · kubernetes/kubernetes (github.com)](https://github.com/kubernetes/kubernetes/pull/121028/files#diff-ef3b90010597098b0d1f6ae71dcd256d14cffd6998159dcf1308b2c783557d75)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


staging/cse/windows/azurecnifunc.ps1
```diff
@@ -416,6 +417,22 @@ function New-ExternalHnsNetwork
         {
             $managementIP = $netIP.IPAddress
```
paulgmiller, line 418: this was just logged and not used before right? Do we need a more strict filter than just first adapter with an ip?

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -416,6 +417,22 @@ function New-ExternalHnsNetwork
         {
             $managementIP = $netIP.IPAddress
             $adapterName = $na.Name
+
+            Write-Log "Get node IPv4 address assigned to the adapter $($na.Name): $($managementIP)"
+            $nodeIPs += $managementIP
+
+            if ($IsDualStackEnabled) {
```
paulgmiller, line 424: kind of curious how we disercn this but ran out of time to go look it up.

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -416,6 +417,22 @@ function New-ExternalHnsNetwork
...
             $adapterName = $na.Name
+
+            Write-Log "Get node IPv4 address assigned to the adapter $($na.Name): $($managementIP)"
+            $nodeIPs += $managementIP
+
+            if ($IsDualStackEnabled) {
+                $netIPv6s = Get-NetIPAddress -ifIndex $na.ifIndex -AddressFamily IPv6 -ErrorAction SilentlyContinue -ErrorVariable netIPErr
+                foreach($ipv6 in $netIPv6s)
+                {
+                    if ($ipv6.PrefixOrigin -ne "WellKnown")
```
paulgmiller, line 428: why well known?  add a comment why this is your filter,

paulgmiller: ipv6 is also just taking the first nic with an ip right?

paulgmiller (approved):
A little worried about nic selection eventually failing if there are multiple nics. (multitenancy) this isn't a big deal for now and doesn't happen on windows yet but lettign some mulitenancy folks know. Hopefully we just have a hostNetwork cloud node manage by then.

neaggarwMS:
Multitenancy is coming to Windows, this is going to break that scenario.

staging/cse/windows/azurecnifunc.ps1
AbelHu: On an Azure Windows VM, there are two IPv6 IP addresses. Below is an example. It is same in an Azure Linux VM.
```
ifIndex IPAddress                                       PrefixLength PrefixOrigin SuffixOrigin AddressState PolicyStore
------- ---------                                       ------------ ------------ ------------ ------------ -----------
6       fe80::97bd:baf7:2853:f73d%6                               64 WellKnown    Link         Preferred    ActiveStore
6       2404:f800:8000:122::4                                    128 Dhcp         Dhcp         Preferred    ActiveStore
```
From the found docuements. `fe80:` with `WellKnown` is the link-local address so we should ignore it.

IPv6 link-local is a special type of unicast address that is auto-configured on any interface using a combination of the link-local prefix FE80::/10 (first 10 bits equal to 1111 1110 10) and the MAC address of the interface. 

https://learn.microsoft.com/en-us/dotnet/api/system.net.networkinformation.prefixorigin?view=net-8.0
<html>
<body>

WellKnown | 2 | The prefix is a well-known prefix. Well-known prefixes are specified in standard-track Request for Comments (RFC) documents and assigned by the Internet Assigned Numbers Authority (Iana) or an address registry. Such prefixes are reserved for special purposes.
-- | -- | --



</body>
</html>

staging/cse/windows/azurecnifunc.ps1
AbelHu: From https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/, it supports `comma-separated dual-stack IP addresse`. I think that we should add both ipv4 and ipv6 addresses since we support dualstack.

<html>
<body>

--node-ip string
--
  | IP address (or comma-separated dual-stack IP addresses) of the node. If unset, kubelet will use the node's default IPv4 address, if any, or its default IPv6 address if it has no IPv4 addresses. You can pass "::" to make it prefer the default IPv6 address rather than the default IPv4 address.


</body>
</html>

staging/cse/windows/azurecnifunc.ps1
AbelHu: It is used to create the network `ext` in https://github.com/Azure/AgentBaker/blob/ed90c97f0ab41ce8c237ca113114bcfe44ca3283/staging/cse/windows/azurecnifunc.ps1#L471C6-L471C30.
I think that it was provided by Windows team and was added in supporting accelerated network three years ago

AbelHu:
> Multitenancy is coming to Windows, this is going to break that scenario.

I think that the logic will be updated in adding multitenancy support on Windows.

AbelHu:
> A little worried about nic selection eventually failing if there are multiple nics. (multitenancy) this isn't a big deal for now and doesn't happen on windows yet but lettign some mulitenancy folks know. Hopefully we just have a hostNetwork cloud node manage by then.

Thank you for the comments. I have replied them. The current nic selection should work for current scenarios. I think that we need to update the code at that time when new scenario is introduced.

staging/cse/windows/azurecnifunc.ps1
AbelHu: Added the comment for the code.

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -416,6 +417,36 @@ function New-ExternalHnsNetwork
...
+                    # the link-local prefix FE80::/10 (first 10 bits equal to 1111 1110 10) and the MAC address of the interface.
+                    #
+                    # https://learn.microsoft.com/en-us/dotnet/api/system.net.networkinformation.prefixorigin?view=net-8.0
+                    # WellKnown | 2 | The prefix is a well-known prefix. Well-known prefixes are specified in standard-track Request for 
+                    # Comments (RFC) documents and assigned by the Internet Assigned Numbers Authority (Iana) or an address registry. Such 
+                    # prefixes are reserved for special purposes. -- | -- | --
+                    if ($ipv6.PrefixOrigin -ne "WellKnown")
+                    {
+                        Write-Log "Get node IPv6 address assigned to the adapter $($na.Name): $($ipv6.IPAddress)"
+                        $nodeIPs += $ipv6.IPAddress
```
feiskyer, line 445: should we only get primary IP and skip all secondary IPs here if they are configured?

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -416,6 +417,36 @@ function New-ExternalHnsNetwork
         {
             $managementIP = $netIP.IPAddress
             $adapterName = $na.Name
+
+            Write-Log "Get node IPv4 address assigned to the adapter $($na.Name): $($managementIP)"
+            $nodeIPs += $managementIP
```
feiskyer, line 422: just curious, would primary IP always the first one when secondary IPs are configured?

staging/cse/windows/azurecnifunc.ps1
AbelHu: I added break in this if block. But I do not know whether the first IP is ok. Currently I only can find two IPv6 ip addresses in one Azure Windows VM, one is WellKnown and another is Dhcp.

staging/cse/windows/azurecnifunc.ps1
AbelHu: I think that currently a AKS Windows VM does not have the second IP. So based on this, picking up the first or the last are same.

staging/cse/windows/azurecnifunc.ps1
AbelHu: @feiskyer I do not know whether it is possible to set multiple IPv6 addresses in one NIC. Do you have any idea?

staging/cse/windows/azurecnifunc.ps1
AbelHu: > econdary IPs are configured?

@feiskyer Same question: Could we assign more than one IPv4 address in one physical NIC?

feiskyer: the IPs are only configured on VM, but not configured in the NIC. So it would be OK here.

staging/cse/windows/azurecnifunc.ps1
feiskyer: there is only one IP for our scenario, so it should be ok here.

feiskyer (approved):

ShiqianTao (approved):

PR #4149, opened 2024-03-11
feat: add node graceful shutdown kubelet config

weichch:
**What type of PR is this?**

/kind feature
/kind api-change

This PR adds node graceful shutdown Kubelet config items and exposes those for client use, so that we can allow customers to enable this feature on their nodes.

This PR is recreated from the previous PR (#4147) from fork as per comment.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


weichch:
As per mentioned in #4147, the feature gate is [by default enabled](https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/#how-do-i-use-it) in all k8s versions we currently support.

cameronmeissner (approved):
approved pending agentbaker e2es

cameronmeissner (approved):

pkg/agent/datamodel/types.go
```diff
@@ -2033,6 +2035,22 @@ type AKSKubeletConfiguration struct {
...
+	// shutdownGracePeriodCriticalPods specifies the duration used to terminate critical
+	// pods during a node shutdown. This should be less than shutdownGracePeriod.
+	// For example, if shutdownGracePeriod=30s, and shutdownGracePeriodCriticalPods=10s,
+	// during a node shutdown the first 20 seconds would be reserved for gracefully
+	// terminating normal pods, and the last 10 seconds would be reserved for terminating
+	// critical pods.
+	// Default: "0s"
+	// +featureGate=GracefulNodeShutdown
+	// +optional
+	ShutdownGracePeriodCriticalPods Duration `json:"shutdownGracePeriodCriticalPods,omitempty"`
```
UtheMan, line 2053: Where do we validate whether `ShutdownGracePeriodCriticalPods` is lower than `ShutdownGracePeriod`?

weichch: Validation is going to be in RP, similar to `ImageGcHighThreshold` and `ImageGcLowThreshold`.

pkg/agent/utils.go
```diff
@@ -497,6 +497,12 @@ func setCustomKubeletConfig(customKc *datamodel.CustomKubeletConfig,
 		if customKc.PodMaxPids != nil {
 			kubeletConfig.PodPidsLimit = to.Int64Ptr(int64(*customKc.PodMaxPids))
 		}
+		if customKc.ShutdownGracePeriod != nil {
```
weichch, line 500: I also wanted to add some defaulting logic based on Kubernetes version. What do you think if we add `IsKubernetesVersionGe(xxx)` in this method? @UtheMan

weichch:
I'm putting this on hold now as there might be some investigations I need to do *before* this can be merged.

cameronmeissner:
@weichch closing for now due to inactivity, feel free to reopen when needed

PR #4150, opened 2024-03-11
chore: update tls bootstrap config in self-contained contract

lilypan26:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,12 +1,9 @@
 syntax = "proto3";
 package nbcontract.v1;
 import "pkg/proto/nbcontract/v1/featurestate.proto";
-
+ 
 message TLSBootstrappingConfig {
-    // discuss whether this should be featureState or bool
-    // understand how defaulting works for enum, because we want default to be false
-    FeatureState tls_bootstrapping_status = 1; // need to review the relationship between tls_bootstrapping_status and secure_tls_bootstrapping_status. For example, if they are multually exclusive, then we should use oneof
```
lilypan26, line 8: this value can be calculated by `if tls_bootstrap_token != ""`

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,12 +1,9 @@
...
-    FeatureState secure_tls_bootstrapping_status = 2;
-    string tls_bootstrap_token = 3; // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
-    string custom_secure_tls_bootstrap_appserver_appid = 4; // will follow up with the feature owner to see if this should be optional
+    bool enable_secure_tls_bootstrapping = 1;
+    optional string tls_bootstrap_token = 2; // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
+    optional string custom_secure_tls_bootstrap_appserver_appid = 3; // only used with secure tls bootstrapping
```
lilypan26, line 8: optional since it is only used when secure tls is enabled

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,12 +1,9 @@
...
+ 
 message TLSBootstrappingConfig {
-    // discuss whether this should be featureState or bool
-    // understand how defaulting works for enum, because we want default to be false
-    FeatureState tls_bootstrapping_status = 1; // need to review the relationship between tls_bootstrapping_status and secure_tls_bootstrapping_status. For example, if they are multually exclusive, then we should use oneof
-    FeatureState secure_tls_bootstrapping_status = 2;
-    string tls_bootstrap_token = 3; // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
-    string custom_secure_tls_bootstrap_appserver_appid = 4; // will follow up with the feature owner to see if this should be optional
+    bool enable_secure_tls_bootstrapping = 1;
+    optional string tls_bootstrap_token = 2; // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
```
lilypan26, line 7: optional since it is only used when tls bootstrapping is enabled

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,12 +1,9 @@
...
-
+ 
 message TLSBootstrappingConfig {
-    // discuss whether this should be featureState or bool
-    // understand how defaulting works for enum, because we want default to be false
-    FeatureState tls_bootstrapping_status = 1; // need to review the relationship between tls_bootstrapping_status and secure_tls_bootstrapping_status. For example, if they are multually exclusive, then we should use oneof
-    FeatureState secure_tls_bootstrapping_status = 2;
-    string tls_bootstrap_token = 3; // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
-    string custom_secure_tls_bootstrap_appserver_appid = 4; // will follow up with the feature owner to see if this should be optional
+    bool enable_secure_tls_bootstrapping = 1;
```
Devinwong, line 6: so you confirmed with the owner (Cameron, right?) that we only need a single variable to specify the status. Correct? That means, if enable_secure_tls_bootstrapping == false, then it's the original "insecure" tls bootstrapping.
Another, is there any reason we cannot use FeatureState secure_tls_bootstrapping_status? We've changed most of the enable/disable variables to _status.

Devinwong: What will happen if both tls_bootstrap_token has value and enable_secure_tls_bootstrapping is true? 
Or could you let me know the 4 combinations of tls_bootstrap_token = empty or not, and enable_secure_tls_bootstrapping == true or false?
I am a bit confused actually.

lilypan26: Yes I confirmed with Cameron that we only need a single variable as long we are able to populate both variables, `ENABLE_TLS_BOOTSTRAPPING` and `ENABLE_SECURE_TLS_BOOTSTRAPPING` in `cse_cmd`. It is NOT implied however, that if `enable_secure_tls_bootstrapping == true`, that the "not secure" tls bootstrapping is used. It is possible that neither is used.  Therefore, the 4 combinations are:

bootstrap_token = "" && enable_secure_tls_bootstrapping = false   -->   no tls boostrapping
bootstrap_token != "" && enable_secure_tls_bootstrapping = false   -->   normal tls boostrapping
bootstrap_token != "" && enable_secure_tls_bootstrapping = true   -->   secure tls boostrapping
bootstrap_token = "" && enable_secure_tls_bootstrapping = true   -->   secure tls boostrapping

lilypan26: Also as referring to my previous comment, the reason why `enable_secure_tls_bootstrapping == false` does not imply that normal tls bootstrapping is used is because in AB we currently calculate whether to use normal tls bootstrapping simply by checking whether the token == "". Therefore in the helper functions we can populate `ENABLE_TLS_BOOTSTRAPPING` simply by checking whether the token is empty or token. If both `ENABLE_TLS_BOOTSTRAPPING` and `ENABLE_SECURE_TLS_BOOTSTRAPPING` is true, that is possible and secure tls will be used but that is an implementation detail for cameron in AB.

Devinwong: Thanks for the truth table. Hopefully Cameron documented it somewhere. (Did he?)
You are right. We should use bool instead of FeatureState, as the default value for bool will be false in protobuff.

Devinwong (approved):

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
cameronmeissner: docs are coming soon (TM)

cameronmeissner: but yes Lily is correct :)

cameronmeissner (approved):

tallaxes (approved):
Approving, but we need to decide on a consistent use of `optional`. As discussed, absence of optional does not make a field required. Adding `optional` does allow to capture additional state / [explicit presence](https://protobuf.dev/programming-guides/field_presence), where the API also stores whether or not a field has been set. This has some use cases, but we need to decide whether and where these are applicable in this contract.

Devinwong:
> Approving, but we need to decide on a consistent use of `optional`. As discussed, absence of optional does not make a field required. Adding `optional` does allow to capture additional state / [explicit presence](https://protobuf.dev/programming-guides/field_presence), where the API also stores whether or not a field has been set. This has some use cases, but we need to decide whether and where these are applicable in this contract.

Let me try to clarify this with 2 examples, for proto3 which we are using now.

1. No presence (**without** `optional` label)
If this variable is unset, the receiver (in our case, Bootstrappers) will get a default value based on its type.
For example, if a `bool` variable is unset, the receiver will get `false`. I checked the doc and tested in codes.
For a string variable, the default value for an unset case is an empty string.

2. Explicit presence (**with** `optional` label)
If this variable is unset, the receiver will get a nil value so resolving template will fail. Here are the error messages tested in codes.
`{"level":"ERROR","time":"2024-03-12T19:57:34.889Z","logger":"controller","message":"Reconciler error","commit":"ce51bfa-dirty","controller":"nodeclaim.lifecycle","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"general-purpose-dvfj4"},"namespace":"","name":"general-purpose-dvfj4","reconcileID":"cf3e14fd-0b72-4afd-8e26-734582106f77","error":"launching nodeclaim, creating instance, getting launch template: getting launch templates, error getting AKS bootstrap script: error getting custom data nbcontract from node bootstrap variables: error executing custom data NbContract template: template: customdata:84:36: executing \"customdata\" at <derefBool .EnsureNoDupePromiscuousBridge>: error calling derefBool: runtime error: invalid memory address or nil pointer dereference"}`


In my opinion, unless there are special case needs _explicit presence_ that we may find out later, we are likely to use _no presence_ , which is without `optional` label, for most of the variables.

Do you guys think this is making sense?

tallaxes:
> Explicit presence (with optional label) ... If this variable is unset, the receiver will get a nil value so resolving template will fail. Here are the error messages tested in codes.

When using `optional` (explicit presence), the receiver should be checking whether the value is present first, using the generated "hazzer" method. To quote protobuf docs:

> The generated message interface includes methods to query presence of fields. For example, the field foo has a corresponding has_foo method. (The specific name follows the same language-specific naming convention as the field accessors.) These methods are sometimes referred to as “hazzers” within the protobuf implementation.

Devinwong (approved):

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
lilypan26: Removed optional, since no need to distinguish between unspecified and default value

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
lilypan26: Removed optional, since no need to distinguish between unspecified and default value

lilypan26:
> > Explicit presence (with optional label) ... If this variable is unset, the receiver will get a nil value so resolving template will fail. Here are the error messages tested in codes.
> 
> When using `optional` (explicit presence), the receiver should be checking whether the value is present first, using the generated "hazzer" method. To quote protobuf docs:
> 
> > The generated message interface includes methods to query presence of fields. For example, the field foo has a corresponding has_foo method. (The specific name follows the same language-specific naming convention as the field accessors.) These methods are sometimes referred to as “hazzers” within the protobuf implementation.

Discussed offline, will be uses protobuf generated getters which are nil safe

PR #4155, opened 2024-03-12
bump: containerd to 1.7.14

lilypan26:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,12 +1,9 @@
 syntax = "proto3";
 package nbcontract.v1;
 import "pkg/proto/nbcontract/v1/featurestate.proto";
-
+ 
 message TLSBootstrappingConfig {
```
cameronmeissner, line 5: looks like you forgot to checkout master before creating your new branch :P

cameronmeissner:
containerd changes lgtm I think, lets wait on abe2e's (if they would actually not flake...) and test builds first though

pkg/proto/nbcontract/v1/tlsbootstrappingconfig.proto
lilypan26: oops good catch, fixed

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -130,7 +130,7 @@ installStandaloneContainerd() {
     #if there is no containerd_version input from RP, use hardcoded version
     if [[ -z ${CONTAINERD_VERSION} ]]; then
         # pin 18.04 to 1.7.1
-        CONTAINERD_VERSION="1.7.7"
+        CONTAINERD_VERSION="1.7.14"
```
cameronmeissner, line 133: lets also verify that we want to keep 1804 pinned to 1.7.1

schemas/manifest.cue
```diff
@@ -58,7 +58,7 @@
         "pinned": {
             "1804": "1.7.1-1" // default in 1804 vhds.
         }
-        "edge": "1.7.7-1",  // edge is default in vhd.
+        "edge": "1.7.14-1",  // edge is default in vhd.
```
cameronmeissner, line 61: for future reference, we only ever need to update this manifest.cue - `make generate` as well as the VHD build itself will use this file to `export` to manifest.json in cloud-init/artifacts

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
cameronmeissner: confirmed offline that we'll keep 1804 where it is

schemas/manifest.cue
lilypan26: yep, anuj informed me a while back, sorry should've reported back

cameronmeissner (approved):
let's also make a note of updating the default abe2e VHDs to pick up the bump

PR #4189, opened 2024-03-19
feat: remove the dependency on managedcluster.Properties.AgentPoolProfiles

AbelHu:
/kind cleanup

**What type of PR is this?**

Remove the dependency on managedcluster.Properties.AgentPoolProfiles

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner (approved):
there's some other spots we need to take care of as well, for example: https://github.com/Azure/AgentBaker/blob/master/pkg/agent/datamodel/types.go#L903-L913

junjiezhang1997 (approved):

AbelHu:
> there's some other spots we need to take care of as well, for example: https://github.com/Azure/AgentBaker/blob/master/pkg/agent/datamodel/types.go#L903-L913

@cameronmeissner From https://learn.microsoft.com/en-us/azure-stack/user/kubernetes-aks-engine-custom-vnet?view=azs-2311#constraints-when-creating-a-custom-virtual-network, all agentpools should have the same result for `IsCustomVNET`
```
The control plane nodes pool and the agent nodes pool must be in the same virtual network. You can deploy your nodes into different subnets within the same virtual network.
```

ShiqianTao (approved):

PR #4193, opened 2024-03-20
chore: clean up windows client secret used in building vhd

junjiezhang1997:
**What type of PR is this?**
/kind cleanup

clean up windows client secret used in building vhd.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -182,9 +182,6 @@ stages:
             docker run --rm \
             -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
             -w /go/src/github.com/Azure/AgentBaker \
-            -e CLIENT_ID=${AZURE_BUILD_CLIENT_ID} \
-            -e CLIENT_SECRET="$(AZURE_BUILD_CLIENT_SECRET)" \
-            -e TENANT_ID=${AZURE_BUILD_TENANT_ID} \
             -e SUBSCRIPTION_ID="${AZURE_BUILD_SUBSCRIPTION_ID}" \
             -e PROD_SUBSCRIPTION_ID=${AZURE_PROD_SUBSCRIPTION_ID} \
             -e AZURE_RESOURCE_GROUP_NAME=${AZURE_BUILD_RESOURCE_GROUP_NAME} \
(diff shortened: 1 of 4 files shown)


.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -182,9 +182,6 @@ stages:
             docker run --rm \
             -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
             -w /go/src/github.com/Azure/AgentBaker \
-            -e CLIENT_ID=${AZURE_BUILD_CLIENT_ID} \
```
cameronmeissner, line 185: are you giving packer an MSI now instead? if so how are you accomplishing this?

I've wanted to get this working for our builds for a long time now, but my understanding is that packer doesn't support cross-tenant managed identities

junjiezhang1997: Yes. I'm giving packer an MSI now. Referring to https://developer.hashicorp.com/packer/integrations/hashicorp/azure#managed-identity, I remove the `client_id`, `client_secret` and `tenant_id`. So packer will attempt to use the Managed Identity and subscription of the VM that Packer is running on, which is related to our 1es hosted pool.

vhdbuilder/packer/init-variables.sh
```diff
@@ -44,15 +44,17 @@ echo "CAPTURED_SIG_VERSION set to: ${CAPTURED_SIG_VERSION}"
...
-if [ -a "${SP_JSON}" ]; then
-	echo "Existing credentials file found."
-	exit 0
-elif [ -z "${CLIENT_ID}" ]; then
-	echo "Service principal not found! Generating one @ ${SP_JSON}"
-	az ad sp create-for-rbac -n aks-images-packer${CREATE_TIME} -o json > ${SP_JSON}
-	CLIENT_ID=$(jq -r .appId ${SP_JSON})
-	CLIENT_SECRET=$(jq -r .password ${SP_JSON})
-	TENANT_ID=$(jq -r .tenant ${SP_JSON})
+if [ "$MODE" == "linuxVhdMode" ]; then
```
AbelHu, line 47: Can we assure that we have only two modes, linuxVhdMode and windowsVhdMode? If not, I think that it is better to use `"$MODE" != "windowsVhdMode"`

AbelHu: @cameronmeissner, it seems like that we only use `linuxVhdMode` and `windowsVhdMode` now and we can remove those legacy code for `default` and `sigMode`. Could you help to confirm it?

AbelHu (approved):
LGTM. Thanks for the PR.

PR #4194, opened 2024-03-21
test: assure that CSEResultFilePath is generated by Windows CSE

AbelHu:
**What type of PR is this?**
/kind test

1. add comment for CSEResultFilePath generated by Windows CSE
2. add a test to assure that CSEResultFilePath is generated by Windows CSE

Fixes #

**Requirements**:
  
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/e2e/windows/e2e-scenario.sh
+++ b/e2e/windows/e2e-scenario.sh
@@ -41,6 +41,19 @@ collect-logs() {
             err "Failed in deleting cse logs in remote storage. Error code is $retval."
         fi
     fi
+
+    array=(azcopy_*)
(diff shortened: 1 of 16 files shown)


e2e/windows/e2e-scenario.sh
```diff
@@ -41,6 +41,19 @@ collect-logs() {
             err "Failed in deleting cse logs in remote storage. Error code is $retval."
         fi
     fi
+
+    array=(azcopy_*)
```
ShiqianTao, line 45: Suggest you wrap this code to a function

AbelHu: Good suggestion. I think that we can do it later since it needs more time to do testing and there are many places in using the similar logic.
I think that we can use some functions, `azcopy-get`, `azcopy-delete`, etc.

AbelHu: @ShiqianTao Created the ADO task.

ShiqianTao (approved):

PR #4197, opened 2024-03-21
chore: bump build_vhd to 2204_gen2_containerd

zachary-bailey:
**What type of PR is this?**

/kind chore

This PR changes 1 of the 3 VHDs used for the AB PR gate (1804_gen2_containerd => 2204_gen2_containerd).

This PR changes the PR gate so that when making a pull request, the pull request builds 2204_gen2_containerd instead of 1804_gen2_containerd, which is an older version that is used less by our customers.


cameronmeissner:
thanks! we'll keep 1804 coverage via the gpu build, thus I think this should be fine

cameronmeissner (approved):

ganeshkumarashok (approved):

PR #4206, opened 2024-03-25
feat: toggles in agentbakersvc

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/toggles/types.go
```diff
@@ -66,16 +66,20 @@ func New() *Toggles {
...
+		}
 	}
 	return map[string]string{}
 }
 
 // getString attempts to resolve the named string toggle against the specified Entity.
 func (t *Toggles) getString(name string, entity *Entity) string {
-	if toggle, ok := t.Strings[name]; ok {
-		return toggle(entity)
+	if t != nil && t.Strings != nil {
```
bravebeaver, line 79: could t be nil here?

apiserver/apiserver.go
```diff
@@ -6,15 +6,21 @@ import (
...
+type OptionConfigurator func(opts *Options)
+
 // Options holds the options for the api server.
 type Options struct {
-	Addr string
+	Addr    string
+	Toggles *toggles.Toggles
 }
 
 func (o *Options) validate() error {
```
bravebeaver, line 26: this is not exactly related to this PR but just check my own golang sanity:
- line 44 would throw panic if o is nil, 
- shall we check for o == nil here?

cameronmeissner: So in golang you can actually call ptr receivers on nil ptrs - the receiver var will simply be nil within the method itself. We already have a nil check within the validate method, thus calling o.validate won't panic if o is nil, we'll get the error saying options are invalid instead

pkg/agent/toggles/types.go
cameronmeissner: in theory yes, since we need to make this an exported type for RP integration, someone could theoretically do:

var tgls *Toggles
m := tgls.getMap("something", EntityFromWhatever(...)) - this would panic without the above checks

from the perspective of usage within the actual agentbaker client implementation, we should be safeguarded against nil's since we only export the constructor and not the underlying type, which means clients would only be able to interact with the type via the AgentBaker interface after calling the corresponding constructor -  and we make sure to properly instantiate an empty set of toggles in said constructor

pkg/agent/datamodel/sig_config.go
```diff
@@ -268,6 +268,10 @@ func (d Distro) IsWindowsPIRDistro() bool {
 	return false
 }
 
+func (d Distro) IsWindowsDistro() bool {
+	return d.IsWindowsSIGDistro() || d.IsWindowsPIRDistro()
```
anujmaheshwari1, line 272: is PIRDistro check still valid?

cameronmeissner: in theory it could be for older clusters? figured I'd include it just to be on the safe side

pkg/agent/toggles/types.go
Devinwong: If I were implementing this, I would return an error when t=nil, so that the client knows while ensuring panic safe. But I don't have concern with this implementation either. ✔️

Devinwong (approved):

pkg/agent/toggles/types.go
cameronmeissner: fair enough, this is something I thought about as well, though I think it would probably be better for now to just log something out instead of throwing an error since we'd still want to fulfill the request I think

anujmaheshwari1 (approved):

PR #4208, opened 2024-03-25
Update nbc contract: identityconfig, vmtype, nodepoolprofile

lilypan26:
**What type of PR is this?**

Updating node bootstrap contract as needed. Updated IdentityConfig and added vmType enum. Removed nodepoolprofile config.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/nodepoolprofile.proto
```diff
@@ -1,9 +0,0 @@
-syntax = "proto3";
-package nbcontract.v1;
-
-message NodePoolProfileConfig {
-
-  // vm_size represents the size of this node pool
-  // Let's keep it for now but continue to review if this can be removed from contract as this can be retrieved from IMDS
-  string vm_size = 1;
```
lilypan26, line 8: Not seeing this used anywhere for populating cse_cmd, not sure why it was initially added?

pkg/proto/nbcontract/v1/vmtype.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+enum VmType {
+  VM_TYPE_STANDARD = 0;
```
lilypan26, line 5: Not sure if defaulting to standard is the best move here, but this is what AB currently does unless vmss is explicity specified.

pkg/proto/nbcontract/v1/identityconfig.proto
```diff
@@ -2,16 +2,8 @@ syntax = "proto3";
 package nbcontract.v1;
 
 message IdentityConfig {
-  IdentityType identity_type = 1;
```
lilypan26, line 5: Removing, since not seeing this used anywhere

pkg/proto/nbcontract/v1/vmtype.proto
Devinwong: Agreed. I think it should be good as long as AB is doing this too.

pkg/proto/nbcontract/v1/identityconfig.proto
Devinwong: Makes sense to me too. If later on we have other identity type to support, we can add it back.

pkg/proto/nbcontract/v1/vmtype.proto
Devinwong: BTW, protobuff official doc suggests to set the first enum=0 as unspecified to handle evolution scenario.
You may want to consider unspecified = 0, standard = 1, etc. And in our go binary parser, if it's unspecified, we will still set it to standard. Does that sound make sense to you?
https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum

tallaxes: Do we know where (and what for) this is used?

pkg/proto/nbcontract/v1/identityconfig.proto
```diff
@@ -2,16 +2,8 @@ syntax = "proto3";
...
-
-enum IdentityType {
-  IDENTITY_TYPE_UNSPECIFIED = 0;
-  IDENTITY_TYPE_SERVICE_PRINCIPAL = 1;
-  IDENTITY_TYPE_USER_IDENTITY = 2;
-  // IDENTITY_TYPE_SYSTEM_IDENTITY = 3; not supported yet
+  string service_principal_id = 1;
+  string service_principal_secret = 2;
+  string assigned_identity_id = 3;  //could be user or system assigned, depending on the type
+  string use_managed_identity_extension = 4;
```
tallaxes, line 8: This should be bool, I think

tallaxes, line 8: Wondering if this is only used for generating azure config file? If yes - it might make sense to collect all the settings that affect it in one place ...

pkg/proto/nbcontract/v1/vmtype.proto
lilypan26: the only place I found is that is it is logged into the azure json file: https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_config.sh#L192C5-L192C28

i'm not sure how this file is used however

tallaxes: This file is used by cloud-provder-azure clients, the structure is documented here: https://cloud-provider-azure.sigs.k8s.io/install/configs/. But:
* Cataloguing all of the uses of this _on the node_ would be great. Currently it is likely used by kubelet (to authenticate to ACRs, via in-tree auth helpers which use cloud-provider-azure clients) and probably CSI drivers. They may be other uses I am not aware of.
* I am not sure why any of these need vmType ... (kubelet ACR auth definitely should not need it?)
* From the contract perspective, it may still be a good idea to collect everything that goes to azure.json in one place (unless we have a clear indication some settings are used elsewhere as well)

pkg/proto/nbcontract/v1/nodepoolprofile.proto
lilypan26: Added back as it can be used to compute many other contract variables which can then be removed.

pkg/proto/nbcontract/v1/vmtype.proto
lilypan26: Sure I can start noting down where azure.json is used and what variables in the contract are stored inside.

When you say collect them in one place, do you mean grouping all the variables in azure.json into one config in the contract?

pkg/proto/nbcontract/v1/identityconfig.proto
lilypan26: good catch, should be a bool

lilypan26: Grouped all the auth config fields and cluster config fields used to generate azure config file as listed here: https://cloud-provider-azure.sigs.k8s.io/install/configs/ into AuthConfig and ClusterConfig.

pkg/proto/nbcontract/v1/clusterconfig.proto
```diff
@@ -0,0 +1,26 @@
...
+package nbcontract.v1;
+
+import "pkg/proto/nbcontract/v1/loadbalancerconfig.proto";
+import "pkg/proto/nbcontract/v1/vmtype.proto";
+
+// Cluster Config fields stored in azure.json used by cloud-provider-azure
+message ClusterConfig {
+  string resource_group = 1;
+  string location = 2;
+  VmType vm_type = 3;
```
Devinwong, line 11: suggest putting the VmType enum in this proto file as we are doing so for others

pkg/proto/nbcontract/v1/clusterconfig.proto
```diff
@@ -0,0 +1,26 @@
...
+// Cluster Config fields stored in azure.json used by cloud-provider-azure
+message ClusterConfig {
+  string resource_group = 1;
+  string location = 2;
+  VmType vm_type = 3;  
+  string primary_availability_set = 4;
+  string primary_scale_set = 5;
+  ClusterNetworkConfig virtual_network_config = 6;
+  bool use_instance_metadata = 7;
+  LoadBalancerConfig load_balancer_config = 8;
```
Devinwong, line 16: As you are putting ClusterNetworkConfig definition in this proto. Maybe we should also put LoadBalancerConfig definition this proto too for consistency?

pkg/proto/nbcontract/v1/clusterconfig.proto
```diff
@@ -0,0 +1,26 @@
...
+}
+
```
Devinwong, line 18: nit: extra line

pkg/proto/nbcontract/v1/clusterconfig.proto
lilypan26: moved

Devinwong (approved):

Devinwong:
One unit test failed. It should be because the contract is changed. In case you didn't know yet, you can run it locally for convenience with **go test `go list ./... | grep -v e2e` -coverprofile coverage_raw.out -covermode count**

pkg/proto/nbcontract/v1/authconfig.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+package nbcontract.v1;
+
+// Auth Config fields stored in azure.json used by cloud-provider-azure
+message AuthConfig {
+  string target_cloud = 1; // set to cloud, default to AzurePublicCloud
+  string tenant_id = 2;
+  string subscription_id = 3;
+  string service_principal_id = 4; // set to aadClientId
+  string service_principal_secret = 5; // set to aadClientSecret
```
tallaxes, line 10: At least these (and possibly others) can be unspecified. What pattern did we settle on for handling optional strings?

lilypan26: I think we decided that unless we need to change the default value of just an empty string or need to handle the nil case specially, then there is no need to specify optional. In this case, not specifying it should have an empty string value so i excluded the optional.

tallaxes (approved):

PR #4154, opened 2024-03-12
chore: enable msi for private container images

junjiezhang1997:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
enable msi for private container images.

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #4213, opened 2024-03-27
feat: add system startup task to expand os volume

ShiqianTao:
**What type of PR is this?**
/kind feature

Before it, CSE extension is responsible for expanding os volume, but other extensions may run predates CSE, result in extension installation failure due to insufficient free space.
After it, VHD registers system startup task to expand os volume which run predates all extensions.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -242,13 +242,40 @@ function Get-ToolsToVHD {
     # Rely on the completion of Get-FilesToCacheOnVHD
     $cacheDir = "c:\akse-cache\tools"
 
-    $toolsDir = "c:\aks-tools"
-    if (!(Test-Path -Path $toolsDir)) {
-        New-Item -ItemType Directory -Path $toolsDir | Out-Null
+    if (!(Test-Path -Path $global:aksToolsDir)) {
+        New-Item -ItemType Directory -Path $global:aksToolsDir -Force | Out-Null
     }
 
     Write-Log "Getting DU (Windows Disk Usage)"
-    Expand-Archive -Path "$cacheDir\DU.zip" -DestinationPath "$toolsDir\DU" -Force
+    Expand-Archive -Path "$cacheDir\DU.zip" -DestinationPath "$global:aksToolsDir\DU" -Force
+}
+
+function Register-ExpandVolumeTask {
+    if (!(Test-Path -Path $global:aksToolsDir)) {
+        New-Item -ItemType Directory -Path $global:aksToolsDir -Force | Out-Null
+    }
+
+    # Leverage existing folder 'c:\aks-tools' to store the task scripts
+    $taskScript = @'
+        $osDrive = ((Get-WmiObject Win32_OperatingSystem -ErrorAction Stop).SystemDrive).TrimEnd(":")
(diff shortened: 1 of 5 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -788,6 +788,34 @@ function Get-SystemDriveDiskInfo {
     }
 }
 
+function Register-ExpandVolumeTask {
+    $toolsDir = "c:\aks-tools\expand-volume"
+    if (!(Test-Path -Path $toolsDir)) {
+        New-Item -ItemType Directory -Path $toolsDir | Out-Null
+    }
+    $taskScript = 
+@'
```
AbelHu, line 797: Can we cache this script in the vhd directly?

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -788,6 +788,34 @@ function Get-SystemDriveDiskInfo {
...
+$expandVolumeLogPath = 'c:\aks-tools\expand-volume\expand-volume.log'
+$diskpartScriptPath = 'c:\aks-tools\expand-volume\diskpart.script'
+if (-not (Test-Path -Path $diskpartScriptPath)) { # Only run once
```
AbelHu, line 800: Should we check the log file instead of the script itself?

vhdbuilder/packer/configure-windows-vhd.ps1
ShiqianTao: Discussed offline:
- cache files should have signature
- we just write it and remove it

vhdbuilder/packer/configure-windows-vhd.ps1
ShiqianTao: remove the task directly

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -218,6 +218,33 @@ function Get-ToolsToVHD {
     Expand-Archive -Path "$cacheDir\DU.zip" -DestinationPath "$toolsDir\DU" -Force
 }
 
+function Register-ExpandVolumeTask {
+    $toolsDir = "c:\aks-tools"
```
AbelHu, line 222: define this in [vhdbuilder/packer/generate-windows-vhd-configuration.ps1](https://github.com/Azure/AgentBaker/pull/4154/files#diff-1a71ca93f12803bce028e28b3600b14d149c348b3c829b2a7fe122b0e017fe25) as @junjiezhang1997 defined `$global:aksTempDir = "c:\akstemp"`?

ShiqianTao: I use the existing folder `c:\aks-tools`, details in line 212. Add this code to decouple function `Get-ToolsToVHD`.
Would you like uses a new folder?

AbelHu: I think that we can use the existing tool folder. Suggest to define "c:\aks-tools" as a global variable.

ShiqianTao: Updated.

AbelHu (approved):
LGTM. Thanks for the PR.

PR #4215, opened 2024-03-27
Removing go-bindata for embed

element-of-surprise:
**What type of PR is this?**

/kind cleanup

It removes go-bindata for the built in embed.FS

This deletes the`template` package for the `parts` package.  Due to how embed works, it makes more sense to have a new package above the actual templates instead of another directory.

This removes complexity with two makefiles, removes the need to install a binary and run said generator. Instead, everything is moved into the standard compile phase.

gobind hasn't been updated in 4 years, which could be a sign of stability, however the master branch has been failing to build for 2 years.  

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

Might want to check my MAKEFILE changes with a skeptical eye, as I don't do makefiles since like the early 2000s.

I modified one of the templates to remove ^M, which is a line ending for windows.

**Release note**:
```
none
```


pkg/templates/templates_generated.go
cameronmeissner, line 1: ah so this removes templates_generated.go - wasn't aware this was being generated by go-bindata

cameronmeissner:
discussed offline, e2e's seemingly look good

cameronmeissner (approved):

PR #4217, opened 2024-03-27
optimizing NB contract including CustomLinuxOSConfig and others

Devinwong:
**What type of PR is this?**
/kind feature

optimizing NB contract including CustomLinuxOSConfig and others

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/kubeletconfig.proto
```diff
@@ -10,13 +10,10 @@ message KubeletConfig {
   repeated Taint startup_taints = 4;
   KubeletDiskType kubelet_disk_type = 5;
 
-  // kubelet_config_file_status specifies if the kubelet_config_file_content should be applied. 
-  FeatureState kubelet_config_file_status = 6;
```
lilypan26, line 14: context for why this can be removed? is it always true? always false?

Devinwong: Though I removed it from the contract, it's computable in the cse.cmd.sh. Please see https://github.com/Devinwong/karpenter-provider-azure/blob/3b7f1b27926597d3d3a55a80a561ee118951c86d/pkg/providers/imagefamily/bootstrap/cse_cmd_nbcontract.sh.gtpl#L99

lilypan26 (approved):

pkg/proto/nbcontract/v1/config.proto
```diff
@@ -95,25 +95,26 @@ message Configuration {
...
-  bool needs_cgroupv2 = 60; // we can calculate this on the VHD, will remove from final contract
+
+  // enable_artifact_streaming specifies whether artifact streaming is enabled or disabled on the VM node
+  bool enable_artifact_streaming = 58;
+
+  // is_kata specifies whether the node is a Kata node
+  bool is_kata = 59; 
+
+  // needs_cgroupv2 specifies whether the node needs cgroupv2
+  bool needs_cgroupv2 = 60;
```
tallaxes, line 117: Hope this goes away soon ... Should this be made optional to allow for this evolution?

pkg/proto/nbcontract/v1/customlinuxosconfig.proto
```diff
@@ -4,22 +4,23 @@ package nbcontract.v1;
 // Custom Linux Node OS Config
 message CustomLinuxOSConfig {
     // The size in MB of a swap file that will be created on each node
-    optional int32 swap_file_size = 1;
+    int32 swap_file_size = 1;
```
tallaxes, line 7: What's the reasoning behind making this (and others) non optional? This means client would always have to specify a value, right?

pkg/proto/nbcontract/v1/config.proto
Devinwong: I left a comment somewhere I couldn't find now, that we should be able to remove this once we actually work on the go binary. So whether it's optional or not should be fine.

pkg/proto/nbcontract/v1/customlinuxosconfig.proto
Devinwong: It means if client doesn't set a value, the default value will be 0, which is expected.
By default we set variable to non-optional unless we need another `nil` state. Does that sound make sense?

tallaxes: "expected" - 0 means the swap will be off? Let's document this?

But here is a harder case:
```go
    // Valid values are "always", "defer", "defer+madvise", "madvise" and "never"
    string transparent_hugepage_support = 2;
```
Which is it if the value is not specified? We may need document the meaning in each case explicitly.

In the spirit of the "contract", it should be very clear what happens for all variants of inputs

pkg/proto/nbcontract/v1/config.proto
tallaxes: It seems to have been on the original line
```go
  bool needs_cgroupv2 = 60; // we can calculate this on the VHD, will remove from final contract
```
Maybe let's keep the comment so that we don't forget?

Devinwong: will do

pkg/proto/nbcontract/v1/customlinuxosconfig.proto
Devinwong: In the cse_config.sh, value `0` will set the swap_size to be 0 instead of turning swap off, (swapon still applies) I don't see any swapoff in the scripts. Let me write this in the comment.

As for transparent_hugepage_support and transparent_defrag, if it's not set, it will do nothing. So it will just use the existing values in the VHD. So the next question is, what is the built-in values in the VHD. I will need to figure this out.

tallaxes: Both of these may be examples where we need to hit the right balance between "staying close to existing semantics / implementation (to minimize disruption)" and "doing the right thing in the contract". Need to be careful not to force some of the existing logic to do something in the contract that does not make sense. There may still be space for adapting the "right thing" to "what scripts need" on the VHD side.

Not saying the above are clear cases of this, probably a bit more investigation is needed. And not going to block anything on this right now.

Whatever we do, it should be very clear for contract consumers which behavior will be requested / affected by which settings.

tallaxes (approved):

pkg/proto/nbcontract/v1/customlinuxosconfig.proto
Devinwong: I double checked the default values in the VHD. For `transparent_hugepage_support`, it's `always`. For `transparent_defrag`, it's `madvise`. I think it makes sense to me if these values are not set, it will use the default values in the VHD. And I added comments to describe the default values.
Totally agreed on "doing the right thing" over "what scripts need"! And I will keep in mind that _it should be very clear for contract consumers which behavior will be requested / affected by which settings._
As for swap_file_size = 0 case, I further checked the difference between in Linux setting `swap_file_size = 0` vs `not setting swapon` and I realized we need to add `enable_swap_config` back that I thought it could be removed earlier. In AgentBaker, the logic of checking if enable_swap_config should be true is like this.
`if (!kubeletConfig.FailSwapOn && CustomLinuxOSConfig.SwapFileSizeMB > 0) then ConfigureSwapFile().`
So I think only the swapfile size isn't enough to tell if enable_swap_config should be true. It also depends on if the VHD's FailSwapOn is false (by default FailSwapOn is true therefore no swapfile should be configured).

PR #4219, opened 2024-03-28

HDYA:


vhdbuilder/packer/components.json
```diff
@@ -132,22 +124,6 @@
       "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v3.24.6"
-      ],
-      "prefetchOptimizations": [
```
andyliuliming, line 136: I think you need to update the fields in bumping

andyliuliming (changes requested):

vhdbuilder/packer/components.json
HDYA: drafting PR, thanks

HDYA: changes merged

PR #4228, opened 2024-03-29
feat: adding api to get k8s version in manifest.json

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/bakerapi.go
```diff
@@ -130,3 +140,23 @@ func (agentBaker *agentBakerImpl) GetDistroSigImageConfig(
 
 	return allDistros, nil
 }
+
+func (agentBaker *agentBakerImpl) GetCachedVHDImages() ([]string, error) {
+	pwd, err := os.Getwd()
```
cameronmeissner, line 145: we can probably just hard-code a file path here instead of always resolving cwd, since that shouldn't ever change at runtime

pkg/agent/bakerapi.go
```diff
@@ -130,3 +140,23 @@ func (agentBaker *agentBakerImpl) GetDistroSigImageConfig(
...
+	if err != nil {
+		return nil, fmt.Errorf("error getting current working directory: %w", err)
+	}
+	manifestFilePath := filepath.Join(pwd, "../../parts/linux/cloud-init/artifacts/manifest.json")
+
+	data, err := os.ReadFile(manifestFilePath)
```
cameronmeissner, line 151: could we just read the contents of the file once at startup since that also wouldn't change at all during runtime?

apiserver/getcachedvhdimages.go
```diff
@@ -0,0 +1,41 @@
...
+	agent "github.com/Azure/agentbaker/pkg/agent"
+)
+
+const (
+	// RoutePathGetCachedVHDImages the route path to get cached vhd images.
+	RoutePathGetCachedVHDImages string = "/getcachedvhdimages"
+)
+
+// GetCachedVHDImages endpoint for getting cached VHD images.
+func (api *APIServer) GetCachedVHDImages(w http.ResponseWriter, r *http.Request) {
```
cameronmeissner, line 18: I'd be curious to see on the RP side how this would get used

im also wondering if we'd ever run into problems where we cache certain things on certain SKUs but not others - in that case just returning contents of the manifest wouldn't quite work

cameronmeissner, line 18: might also rename to make it clear these are container images we're talking about

pkg/agent/bakerapi.go
```diff
@@ -130,3 +140,23 @@ func (agentBaker *agentBakerImpl) GetDistroSigImageConfig(
 
 	return allDistros, nil
 }
+
+func (agentBaker *agentBakerImpl) GetCachedVHDImages() ([]string, error) {
+	pwd, err := os.Getwd()
+	if err != nil {
+		return nil, fmt.Errorf("error getting current working directory: %w", err)
+	}
+	manifestFilePath := filepath.Join(pwd, "../../parts/linux/cloud-init/artifacts/manifest.json")
```
cameronmeissner, line 149: actually, if we care about container images we should be reading from components.json instead

apiserver/getcachedvhdimages.go
```diff
@@ -0,0 +1,41 @@
...
+	agentBaker, err := agent.NewAgentBaker()
```
cameronmeissner, line 19: sad that we always construct new instances of baker...

pkg/agent/bakerapi.go
cameronmeissner: embedding might be useful here

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +15,12 @@ const (
 	AzurePublicCloudSigSubscription string = "109a5e88-712a-48ae-9078-9ca8b3c81345" // AKS VHD
 )
 
+func init() {
+	_, filename, _, _ := runtime.Caller(0)
+	manifestFilePath := "../../../parts/linux/cloud-init/artifacts/manifest.json"
+	getCachedK8sVersionFromManifest(path.Join(path.Dir(filename), manifestFilePath))
```
AlisonB319, line 21: When you run `go test` it will calculate the relative file path from where you call go test. 

manifestFilePath worked as intended when called inside `pkg/agent/` but it would fail if you ran `go test` in `pkg/agent/datamodel`. So the `_, filename, _, _ := runtime.Caller(0)` is a workaround to make sure the relative path is from within sig_config.go

PR #4233, opened 2024-04-01
Self-contained: Update GPU and HTTPProxy Config in contract

lilypan26:
**What type of PR is this?**

Updating some remaining configs in new contract. Removed unnecessary fields from GPU config and HTTP Proxy config.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -4,23 +4,12 @@ package nbcontract.v1;
...
-
-  // FeatureState fabric_manager_status = 5;  removed. consider just extracting vm_size since both GPUNeedsFabricManager and getGPUDriverVersion depend only on the vm_size.
-  // string gpu_driver_version = 6; removed. this is dependent only on VM size, can be calculated on the VHD
+  FeatureState gpu_device_plugin = 2; 
 
   // GpuInstanceProfile represents the GPU instance profile.
-  optional string gpu_instance_profile = 4;
-
-  // GpuImageSha represents AKS-GPU image's SHA
-  optional string gpu_image_sha = 5;
```
Devinwong, line 25: Why gpu_image_sha is not needed?

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -4,23 +4,12 @@ package nbcontract.v1;
...
 
 message GPUConfig {
-  // NvidiaState represents the Nvidia GPU state
-  optional FeatureState nvidia_state = 1;
-
-  // FeatureState is_mig_node = 2; removed since it only depends on gpu_instance_profile
-
   // ConfigGpuDriver specifies whether bootstrap process should install and configure the GPU driver when necessary. Configuration includes appropriate set up of components like the fabric manager where applicable.
-  optional FeatureState config_gpu_driver = 2;
+  FeatureState config_gpu_driver = 1;
```
Devinwong, line 8: Do we need FeatureState for config_gpu_driver and gpu_device_plugin? Or simply bool is enough?
FeatureState was used here before the FeatureState vs bool conversation happened. We can make a decision here.

pkg/proto/nbcontract/v1/gpuconfig.proto
lilypan26: This can be computed from the vmsize like so: 
```
func GetAKSGPUImageSHA(size string) string {
	if useGridDrivers(size) {
		return AKSGPUGridSHA
	}
	return AKSGPUCudaSHA
}
```
The image SHAs can be stored on the VHD since Karpenter currently just copy and pastes this from AB anyway.

pkg/proto/nbcontract/v1/gpuconfig.proto
lilypan26: Yes good catch, I think these can be bools.

pkg/proto/nbcontract/v1/gpuconfig.proto
```diff
@@ -4,23 +4,12 @@ package nbcontract.v1;
 import "pkg/proto/nbcontract/v1/featurestate.proto";
 
 message GPUConfig {
-  // NvidiaState represents the Nvidia GPU state
-  optional FeatureState nvidia_state = 1;
```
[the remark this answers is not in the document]
Devinwong: That's awesome

Devinwong (approved):
lgtm

tallaxes (approved):

PR #4236, opened 2024-04-02
feat: add azure gpu helpers to AB

lilypan26:
**What type of PR is this?**

Move crphelper gpu helper functions from RP to AB so they can be ingested by Karpenter as a single source of truth instead of having to maintain a [separate copy](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/utils/gpu.go).

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/const.go
```diff
@@ -130,6 +130,101 @@ const (
...
+	"standard_nc24":  true,
+	"standard_nc24r": true,
+	// M60
+	"standard_nv6":      true,
+	"standard_nv12":     true,
+	"standard_nv12s_v3": true,
+	"standard_nv24":     true,
+	"standard_nv24s_v3": true,
+	"standard_nv24r":    true,
+	"standard_nv48s_v3": true,
```
Bryce-Soghigian, line 150: Thoughts on using apimachinery sets rather than maps here? Since all map entires are true probably makes more sense

pkg/agent/datamodel/helper.go
```diff
@@ -48,6 +48,32 @@ func IsMIGNode(gpuInstanceProfile string) bool {
 	return gpuInstanceProfile != ""
 }
 
+// IsNvidiaEnabledSKU determines if an VM SKU has nvidia driver support.
+func IsNvidiaEnabledSKU(vmSize string) bool {
+	// Trim the optional _Promo suffix.
+	vmSize = strings.ToLower(vmSize)
+	vmSize = strings.TrimSuffix(vmSize, "_promo")
+	return NvidiaEnabledSKUs[vmSize]
+}
```
Bryce-Soghigian, line 57: Datamodel is very opinionated. Not sure how the new bootstrapping contract work is progressing but it might make sense to pull this out into a separate module. People who may not be interested in agentbaker's data model might be interested in gpu skus.

pkg/agent/datamodel/const.go
```diff
@@ -130,6 +130,101 @@ const (
 	AKSGPUCudaSHA = "sha-ff213d"
 )
 
+/*
```
Bryce-Soghigian, line 133: It would also be good to introduce the helpers for GPU SHA like GETAKSGPUIMAGESHA, https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/utils/gpu.go#L33

pkg/agent/datamodel/helper.go
Bryce-Soghigian, line 57: Can you also bring over the test coverage from karpenter for all of these functions? Its really barebones but still useful a bit https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/utils/gpu_test.go

pkg/agent/datamodel/const.go
```diff
@@ -130,6 +130,101 @@ const (
...
+	"standard_nd96ams_v4":        true,
+	"standard_nd96ams_a100_v4":   true,
+	"standard_nd96amsrf_a100_v4": true,
+	"standard_nd96amsf_a100_v4":  true,
+	// H100
+	"standard_nd96isr_h100_v5":  true,
+	"standard_nd96is_h100_v5":   true,
+	"standard_nd96isrf_h100_v5": true,
+	"standard_nd96isf_h100_v5":  true,
+	// NC
```
Bryce-Soghigian, line 206: nit: I know these aren't your comments but might be useful to link to subfamily here or documentation for A100 that is available.

pkg/agent/datamodel/const.go
```diff
@@ -130,6 +130,101 @@ const (
...
+	"standard_nc40ads_h100_v5":  true,
+	"standard_nc80adis_h100_v5": true,
+}
+
+// List of GPU SKUs currently enabled and validated for Mariner. Will expand the support
+// to cover other SKUs available in Azure
+var Mariner_NvidiaEnabledSKUs = map[string]bool{
```
Bryce-Soghigian, line 213: nit: should these now be AzureLinux to reflect the rebranding effort from AZ Linux team?

Bryce-Soghigian (commented):
Left some comments, happy to see this driven into a shared place! 

 Next steps would to be to update some references to use agentbaker rather than their own lists: 
 - https://github.com/Azure/gpu-provisioner/blob/main/pkg/utils/gpu.go
 - https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/utils/gpu.go


We should also move the CudaDriverVersion and SHA getters into a similar library. No need to update Karpenter if you plan on removing gpu.go in the bootstrapping rework code. It makes sense that passing in vm size to the vhd should allow us to resolve things like "GPUNode" and other factors like that. 


Also nit on title, no one outside of aks will know what crphelper from RP to AB means

pkg/agent/datamodel/const.go
cameronmeissner: I guess we already have these in the agent package, though unexported...

pkg/agent/datamodel/const.go
cameronmeissner: im fine either way, it would require clients to also take a dependency on apimachinery if they wanted to use this directly

pkg/agent/datamodel/const.go
cameronmeissner: it's still confusing since we will have both until 1.27 is out of support

pkg/agent/datamodel/helper.go
cameronmeissner: part of me feels like this should be part of that skewer api

though maybe `pkg/agent/common` would be a good place to start?

cameronmeissner: im also fine with leaving it in the datamodel for now if that's easiest, but Bryce has a point - it also begs the question of what exactly we want to be exposing from the agentbaker module in general

pkg/agent/datamodel/const.go
Bryce-Soghigian: Right we should probably export them, im pretty sure i stole them from you initially

pkg/agent/datamodel/const.go
Bryce-Soghigian: Fair enough but the list is the same for support for both right?

pkg/agent/datamodel/const.go
Bryce-Soghigian: Either that or we could just use a map[string]struct{} so it uses zero bytes rather than taking api machinery dependency. There is a lot of useful deps in api machinery i would recommend adopting

lilypan26: Decided to try out apimachinery sets, the only usages in RP where they are directly accessing the map is to get a string formatted list of the skus for logging. I created a helper function for this so that RP shouldn't have to add a dependency. I made all the map/sets private since they shouldnt need to be public now.

pkg/agent/datamodel/helper.go
lilypan26: Created package `pkg/agent/common`

pkg/agent/datamodel/const.go
lilypan26: Yep I exported them and put them in the new `pkg/agent/common` package for now, but some helpers like `getAKSGPUImageSHA` will only be used for bootstrapping so I might move them to a separate library for self-contained later on.

pkg/agent/datamodel/helper.go
lilypan26: Moved over. Also merged with some tests from RP.

pkg/agent/datamodel/const.go
lilypan26: added links for sku families docs i could find

Bryce-Soghigian (approved):

pkg/agent/datamodel/const.go
cameronmeissner: I believe so, i'd be fine aligning towards azurelinux

cameronmeissner (approved):

ganeshkumarashok (commented):
So the plan is to have karpenter/RP use this library/common functions instead right? Have you tested it e2e from the consumption side?

pkg/agent/datamodel/helper.go
ganeshkumarashok: yes, also thinking if it's better for this to be part of the `skewer` API

ganeshkumarashok (approved):
For the future:
- We need to remember to bump the agentbaker version in all dependencies in when we update the SKUs here, once the RP integration is done. Matters more when this is time-sensitive. 
- If there's a way in which we prevent this additional step, that'll be good (since it's adding an extra step)
- Update the VHD automation pipeline as well

ganeshkumarashok (approved):

PR #4229, opened 2024-03-30
feat: Don't set feature gate TopologyManager and CustomCPUCFSQuotaPer…

gaopenghigh:
…iod, Get them from kubelet configs

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Currently, if TopologyManagerPolicy and CPUCFSQuotaPeriod are set in CustomKubeletConfigs, agent baker set feature gates "TopologyManager=true" and "CustomCPUCFSQuotaPeriod =true".
This PR remove the logic above because we set it in RP: https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/pullrequest/9808037

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #4240, opened 2024-04-03
feat: remove logic for TopologyManager feature gate

gaopenghigh:
**What type of PR is this?**

/kind feature

Per [Feature Gates (removed)](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/) , feature gate `TopologyManager` is default since 1.18, and was removed from 1.29. This PR remove logic for this feature gate.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/utils.go
```diff
@@ -473,8 +473,6 @@ func setCustomKubeletConfig(customKc *datamodel.CustomKubeletConfig,
 		}
 		if customKc.TopologyManagerPolicy != "" {
 			kubeletConfig.TopologyManagerPolicy = customKc.TopologyManagerPolicy
-			// enable TopologyManager feature gate is required for this configuration.
-			kubeletConfig.FeatureGates["TopologyManager"] = true
```
cameronmeissner, line 477: sorry I think you clarified this earlier but I can't find the messages - are we sure RP is currently setting this feature flag accordingly? otherwise this could introduce problems

cameronmeissner: ah, reread the PR description

so from reading this: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/, it seems like we haven't needed to set this feature gat since 1.18 since it was defaulted to true?

cameronmeissner (approved):

pkg/agent/utils.go
gaopenghigh: We don't need to set "TopologyManager" in agent baker, since it is default true for 1.18+ clusters.  So this PR is safe to merge and has no release dependency for RP.

Another PR (#4229 ) is to remove CustomCPUCFSQuotaPeriod logic from agent baker, this is dependent on RP and I'll update the PR once we have RP updated.

ShiqianTao (approved):

PR #351, opened 2020-09-23
Setup azure0 bridge when starting kubelet

zhedahht:
We have a bug that azure0 bridge is not initialized as excepted in some cases. We setup the bridge if it's not initialize before when starting kubelet, as suggested by partner teams.

PR #4241, opened 2024-04-03
feat: Update CustomCloudConfig proto file

Devinwong:
**What type of PR is this?**
/kind feature

Update CustomCloudConfig proto file to use enable_custom_cloud_config

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

- Rename enums for simplicity
- Update LoadBalancerConfig

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/customcloudconfig.proto
```diff
@@ -1,18 +1,18 @@
...
+    string init_file_path = 2;
 
     // RepoDepotEndpoint is the endpoint of the repo depot
-    optional string repo_depot_endpoint = 3;
+    string repo_depot_endpoint = 3;
 
     string target_environment = 4;
-    string custom_env_json_content = 5; // can be generated on the VHD, also rename
+
+    string custom_env_json_content = 5;
```
Devinwong, line 17: According to my investigation, this can't be computed thus adding it back.

pkg/proto/nbcontract/v1/networkconfig.proto
```diff
@@ -11,21 +10,21 @@ message NetworkConfig {
...
+  BRIDGE = 1;
+  TRANSPARENT = 2;
 }
 
 enum NetworkPluginType {
-  NETWORK_PLUGIN_TYPE_UNSPECIFIED = 0;
-  NETWORK_PLUGIN_TYPE_NONE = 1;
-  NETWORK_PLUGIN_TYPE_AZURE = 2;
-  NETWORK_PLUGIN_TYPE_KUBENET = 3;
+  NPT_UNSPECIFIED = 0;
```
lilypan26, line 19: is there a reason we are shortening all of the enum names? i feel like it makes the code less readable

pkg/parser/parser.go
```diff
@@ -13,7 +13,7 @@ import (
 var (
 	//go:embed cse_cmd.sh.gtpl
 	bootstrapTrigger         string
-	bootstrapTriggerTemplate = template.Must(template.New("triggerBootstrapScript").Funcs(getFuncMap()).Parse(bootstrapTrigger)) //nolint:gochecknoglobals
+	bootstrapTriggerTemplate = template.Must(template.New("triggerBootstrapScript").Parse(bootstrapTrigger)) //nolint:gochecknoglobals
```
lilypan26, line 16: for my own understanding since im not too clear on how the embedding works, but why can we get rid of the `.Funcs(getFuncMap())`?

pkg/proto/nbcontract/v1/networkconfig.proto
tallaxes, line 1: I am not at all sure shortening the prefixes works; here especially one has to get creative with NetworkPolicyType (NPOT) since NPT is already taken. Not pretty.

Other notes here:
* Can we drop TYPE in value names? And Type in enum names? (Or is any of this part of best practices for some good reason?)
* I am checking if NETWORK_MODE is still needed (suspect we could just assume "transparent")
* Surprised there is no NETWORK_POLICY=Cilium. Actually, it looks like NETWORK_POLICY is only used to determine bridge/transparent, so maybe we can get rid of it too (checking)

Devinwong: @tallaxes  @lilypan26  The reason I was trying to shorten it is because of simplicity.
For example, if I use an enum in the client, it looked like `nbcontractv1.NetworkPolicyType_NETWORK_POLICY_TYPE_CALICO`. After the simplicity, it looks like this now `nbcontractv1.NetworkPolicyType_NPOT_CALICO`, though still very long.

- We can drop TYPE. As the enum name becomes longer, dropping TYPE makes sense to me.
- The reason why we still need the prefix (e.g, NETWORK_MODE) is because protbuf compilation. For example, if I have the following definitions 
```
enum NetworkPluginType {
  NPT_UNSPECIFIED = 0;
  NPT_NONE = 1;
  AZURE = 2;
  NPT_KUBENET = 3;
}

enum NetworkPolicyType {
  NPOT_UNSPECIFIED = 0;
  NPOT_NONE = 1;
  AZURE = 2;
  NPOT_CALICO = 3;
}

message NetworkConfig {
  NetworkPluginType network_plugin = 1;
  NetworkPolicyType network_policy = 2;
  NetworkModeType network_mode = 3;
  string vnet_cni_plugins_url = 4;
  string cni_plugins_url = 5;
}
```

Note I removed the prefix from `AZURE` which exists in both `NetworkPluginType` and `NetworkPolicyType`.
Protobuf will complain `symbol "nbcontract.v1.AZURE" already defined at pkg/proto/nbcontract/v1/networkconfig.proto:21:3; protobuf uses C++ scoping rules for enum values, so they exist in the scope enclosing the enum`. That's why I put a shortened prefix. But if we prefer the longer form over shortened form, I am also fine. I don't have strong preference against either.

- NETWORK_POLICY and other network variables are still used in the bootstrap scripts for other computations (e.g., configureCNIIPTables) . So we still need to keep them😂. Cilium was found in the const.go in AgentBaker but seems not used in the bootstrap scripts. Will add it back if it's found useful.

pkg/proto/nbcontract/v1/networkconfig.proto
Devinwong: Please see my response in another comment :)

pkg/parser/parser.go
Devinwong: The reason I removed the getFuncMap was because it was using 2 functions depending on FeatureState, which I removed in this PR. Now I update the codes to keep the getFuncMap function returning an empty FuncMap.
Btw, this parser.go will be used for our upcoming parser which we haven't started the work yet. We will have a major development on parser soon. So anything in the parser.go or utilities.go are just temporary.

In case you need more context about go template, let me elaborate as follows.
Here we call `Funcs(getFuncMap())` so that the template `bootstrapTriggerTemplate` (which imports `cse_cmd.sh.gtpl`) can call some additional functions during parsing.
For example, if I add a `funcA` to the function map, then in the `cse_cmd.sh.gtpl` we can use it like `ENV_A= {{funcA .varA}}` where `funcA` takes an argument `.varA`. Hope this helps. 😊

pkg/proto/nbcontract/v1/networkconfig.proto
tallaxes: NETWORK_MODE/NETWORK_POLICY are only involved in bridge/transparent decision (in [configureCNIIPTables](https://github.com/Azure/AgentBaker/blob/a5e1c918a7c8dd0cf6a7e9fd8c8d8cfa1540eb41/parts/linux/cloud-init/artifacts/cse_config.sh#L270-L274)), so if we confirm transparent is always used (and as far as I can tell, the recent Azure CNI tarballs already have 10-azure.conflist with mode: transparent, so these sed commands don't do anything), we don't need them in contract. We could still set them to something on the VHD side, if we don't want to touch the scripts for now. The main point is that VHD right now does not really depend on, and does not need to know NETWORK_MODE/NETWORK_POLICY, so they don't need to be in the contract, and bootstrappers don't need to supply them (to be confirmed).

Devinwong: But NETWORK_POLICY is also written to /etc/default/kubelet here https://github.com/Azure/AgentBaker/blob/a5e1c918a7c8dd0cf6a7e9fd8c8d8cfa1540eb41/parts/linux/cloud-init/artifacts/cse_config.sh#L382C11-L382C25.  I don't know how kubelet is consuming it though.

Devinwong: > Should we just get rid of the abbreviations in front then? For example, I dont think `NPOT_` adds anything useful.

But for the `AZURE` case, we can only either do NPOT_ or NETWORK_POLICY_TYPE_. Otherwise protobuff will report compile error.

tallaxes: > But NETWORK_POLICY is also written to /etc/default/kubelet here https://github.com/Azure/AgentBaker/blob/a5e1c918a7c8dd0cf6a7e9fd8c8d8cfa1540eb41/parts/linux/cloud-init/artifacts/cse_config.sh#L382C11-L382C25. I don't know how kubelet is consuming it though.

I cannot think of a reason, but there may be one; would be great to track it down. This was introduced in https://github.com/Azure/AgentBaker/pull/3349, @anujmaheshwari1, any insight of what this could possibly be for?

tallaxes: kubelet should have nothing to do with network policy, that's the confusing part.

tallaxes: Mystery solved, looks like it was used by a script that is no longer there, https://github.com/Azure/AgentBaker/pull/351 (Thank you @wedaly for tracking this down!)

Devinwong: Updated the comments and removed TYPE from enum names.

pkg/proto/nbcontract/v1/networkconfig.proto
```diff
@@ -1,31 +1,38 @@
...
+  // NetworkPlugin is the network plugin to be used by the cluster. Options are NONE, AZURE, KUBENET.
+  NetworkPlugin network_plugin = 1;
+
+  // NetworkPolicy is the network policy to be used by the cluster.
+  // This is still needed to compute ENSURE_NO_DUPE_PROMISCUOUS_BRIDGE. 
+  // Other than that, it is not used by others. See the discussions here https://github.com/Azure/AgentBaker/pull/4241#discussion_r1554283228
+  NetworkPolicy network_policy = 2;
+
+  // NetworkMode is the network mode to be used by the cluster. Options are BRIDGE and TRANSPARENT.
+  NetworkMode network_mode = 3;
```
lilypan26, line 15: I believe the conclusion was that we can remove `network_mode`. It was always be `transparent` so we can just directly pass that to cse_cmd for now until we update the scripts, but it can be removed from contract.

Devinwong: Totally! Missed that conclusion. Removed network_mode from contract now

lilypan26 (approved):

tallaxes (approved):

PR #4246, opened 2024-04-05
perf: increase packer disk tier, add img_sku tag

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR increases the disk tier used on the Packer VMs from Standard_LRS to Premium_LRS. The Premium SSDs have notable performance increases and will allow the VM to take full advantage of future parallelization efforts. 

Additionally, this PR adds `img_sku` to the Azure Tags section of the builder template. This enables a developer to more easily find the packer resource group (VM, Disk, NIC) that is building any specific VHD in order to monitor performance metrics. 

Unsatisfactory VHD Build pipeline duration.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
+++ b/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
@@ -53,10 +53,12 @@
         "SkipLinuxAzSecPack": "true",
         "os": "Linux",
         "now": "{{user `create_time`}}",
-        "createdBy": "aks-vhd-pipeline"
+        "createdBy": "aks-vhd-pipeline",
+        "image_sku": "{{user `img_sku`}}"
       },
       "location": "{{user `location`}}",
       "vm_size": "{{user `vm_size`}}",
+      "managed_image_storage_account_type": "Premium_LRS",
       "managed_image_os_disk_snapshot_name": "{{user `arm64_os_disk_snapshot_name`}}",
       "skip_create_image": "true",
       "managed_image_resource_group_name": "{{user `resource_group_name`}}",
--- a/vhdbuilder/packer/vhd-image-builder-base.json
+++ b/vhdbuilder/packer/vhd-image-builder-base.json
@@ -55,10 +55,12 @@
         "SkipLinuxAzSecPack": "true",
         "os": "Linux",
         "now": "{{user `create_time`}}",
-        "createdBy": "aks-vhd-pipeline"
+        "createdBy": "aks-vhd-pipeline",
+        "image_sku": "{{user `img_sku`}}"
       },
       "location": "{{user `location`}}",
       "vm_size": "{{user `vm_size`}}",
+      "managed_image_storage_account_type": "Premium_LRS",
       "managed_image_resource_group_name": "{{user `resource_group_name`}}",
       "managed_image_name": "{{user `sig_image_name`}}-{{user `captured_sig_version`}}",
       "polling_duration_timeout": "1h",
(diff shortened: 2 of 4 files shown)


vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
```diff
@@ -53,10 +53,12 @@
         "SkipLinuxAzSecPack": "true",
         "os": "Linux",
         "now": "{{user `create_time`}}",
-        "createdBy": "aks-vhd-pipeline"
+        "createdBy": "aks-vhd-pipeline",
+        "image_sku": "{{user `img_sku`}}"
       },
       "location": "{{user `location`}}",
       "vm_size": "{{user `vm_size`}}",
+      "managed_image_storage_account_type": "Premium_LRS",
```
Devinwong, line 61: Just curious, did we see any perf improvement (like build time decreased?) after changing to Premium_LRS?

zachary-bailey: Yes, it has enabled performance increases throughout the build, particularly when pulling container images. Depending on the VHD it decreases the build time by about 5 minutes by itself. However, it enables the parallelization of pulling container images and kube-proxy/kubernetes binaries to execute much faster, so further optimizations build on the performance that this allows. 

Pulling container images originally took between 9:45 - 17:00. 

With the increased vm size, increased disk tier, and parallelization, it takes between 3:30 - 9:00 minutes the overwhelming majority of the time now.

Devinwong (approved):

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
anujmaheshwari1: oh, did upstream packer implement support for Premium_LRS? thats cool, it didnt until I guess an year ago?

anujmaheshwari1 (approved):

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
```diff
@@ -53,10 +53,12 @@
         "SkipLinuxAzSecPack": "true",
         "os": "Linux",
         "now": "{{user `create_time`}}",
-        "createdBy": "aks-vhd-pipeline"
+        "createdBy": "aks-vhd-pipeline",
+        "image_sku": "{{user `img_sku`}}"
```
cameronmeissner, line 57: do we need to set `img_sku` in `vhdbuilder/packer/settings.json`?

zachary-bailey: For the purposes I made this commit for (faster / smoother debugging and monitoring experience during builds), I do not believe so. 

The IMG_SKU variable is available in the environment at the time of the resource group deployment object creation, so I just tagged the deployment group with it. That way it can be found easier if you're wondering what is going on with one of your builds. 

That being said, we do set the img sku for Windows builds it seems, but not Linux. Linux only gets the sig_image_name, which is just the base image. Would it be useful in the node provisioning process? I can add it in this or a separate PR if you would like. @cameronmeissner

ganeshkumarashok (approved):

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
cameronmeissner: I just wanted to make sure this value is getting piped into the template correctly, though im assuming you've tested and it seems like it's good enough that it's in the environment - I can never remember if "user" variable space also takes from env, which it seems is the case

cameronmeissner (approved):

PR #4247, opened 2024-04-05
feat: STLS bootstrap integration with linux CSE

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -74,19 +74,19 @@ downloadCNI() {
     retrycmd_get_tarball 120 5 "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" ${CNI_PLUGINS_URL} || exit $ERR_CNI_DOWNLOAD_TIMEOUT
 }
 
-downloadSecureTLSBootstrapKubeletExecPlugin() {
-    local plugin_url="https://k8sreleases.blob.core.windows.net/aks-tls-bootstrap-client/${SECURE_TLS_BOOTSTRAP_KUBELET_EXEC_PLUGIN_VERSION}/linux/amd64/tls-bootstrap-client"
+downloadSecureTLSBootstrapClient() {
+    local client_url="https://k8sreleases.blob.core.windows.net/aks-tls-bootstrap-client/${SECURE_TLS_BOOTSTRAP_CLIENT_VERSION}/linux/amd64/tls-bootstrap-client"
```
phealy, line 78: This hostname is not currently included in [the list of required egress destinations](https://learn.microsoft.com/en-us/azure/aks/outbound-rules-control-egress) and thus may not be accessible for locked down customers. Can we get it somewhere that we're guaranteed to have access, instead? Perhaps we can package the tool as a container image on MCR and use ctr run with bind mounts to invoke it? This would also let us cache it easily.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -374,6 +381,25 @@ ensureDHCPv6() {
     retrycmd_if_failure 120 5 25 modprobe ip6_tables || exit $ERR_MODPROBE_FAIL
 }
 
+ensureKubeCAFile() {
+    KUBE_CA_FILE="/etc/kubernetes/certs/ca.crt"
+    mkdir -p "$(dirname "${KUBE_CA_FILE}")"
+    echo "${KUBE_CA_CRT}" | base64 -d > "${KUBE_CA_FILE}"
+    chmod 0600 "${KUBE_CA_FILE}"
```
phealy, line 388:
```suggestion
    chmod 0644 "${KUBE_CA_FILE}"
```

That's a public key, nothing confidential, might as well let it be accessible.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -374,6 +381,25 @@ ensureDHCPv6() {
...
+}
+
+configureSecureTLSBootstrap() {
+    SECURE_TLS_BOOTSTRAPPING_DROP_IN="/etc/systemd/system/secure-tls-bootstrap.service.d/10-securetlsbootstrap.conf"
+    touch "${SECURE_TLS_BOOTSTRAPPING_DROP_IN}"
+    chmod 0600 "${SECURE_TLS_BOOTSTRAPPING_DROP_IN}"
+
+    tee "${SECURE_TLS_BOOTSTRAPPING_DROP_IN}" > /dev/null <<EOF
```
phealy, line 396:
```suggestion
    cat "${SECURE_TLS_BOOTSTRAPPING_DROP_IN}" <<EOF
```

no need to tee and drop the output; just use cat.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -506,6 +493,23 @@ EOF
     systemctlEnableAndStart kubelet || exit $ERR_KUBELET_START_FAIL
 }
 
+ensureSecureTLSBootstrap() {
+    if [ -f "$KUBECONFIG_FILE" ]; then
+        return 0
+    fi
```
phealy, line 499: Should we be checking here to make sure the file is valid instead of just returning 0 if it exists? This could put us in an endless loop if the file is written but invalid.

parts/linux/cloud-init/artifacts/secure-tls-bootstrap.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Runs the secure TLS bootstrapping client binary to generate a kubelet client credential
+ConditionPathExists=/opt/azure/tlsbootstrap/tls-bootstrap-client
+Wants=network-online.target
+After=network-online.target
```
phealy, line 5: Should this have a `Before=kubelet.service`?

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner: the plan is to mirror it to acs-mirror once we get that far, which should give us availability in all environments afaik - for now we're just using upstream's staging environment for testing purposes

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: good call out, we can try to validate the kubeconfig here or potentially just skip this check altogether since the binary itself will take care of ensuring a valid kubeconfig exists already

parts/linux/cloud-init/artifacts/secure-tls-bootstrap.service
cameronmeissner: From reading docs it seems `Before=` is strictly for startup/shutdown ordering, so I think to answer my own question, it should be fine?

https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Before=

parts/linux/cloud-init/artifacts/cse_config.sh
phealy: if the binary is checking it, then I would just remove the check. :)

parts/linux/cloud-init/artifacts/secure-tls-bootstrap.service
cameronmeissner: added per offline discussion

cameronmeissner:
replaced by #6374

PR #4256, opened 2024-04-08
chore: cache azure-cni and azure-ipam

rbtr:
**What type of PR is this?**

/kind feature

Caches `azure-cni` and `azure-ipam`, which replace and deprecate `cni-dropgz` as the preferred containers for installing the CNI plugins in AKS.

https://github.com/Azure/azure-container-networking/issues/2333

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/components.json
+++ b/parts/linux/cloud-init/artifacts/components.json
@@ -37,11 +37,25 @@
       ]
     },
     {
+      "downloadURL": "mcr.microsoft.com/containernetworking/azure-cni:*",
+      "amd64OnlyVersions": [],
+      "multiArchVersions": [
+        "v1.5.28"
+      ],
+      "prefetchOptimizations": [
+        {
+          "version": "v1.5.28",
+          "binaries": [
+            "dropgz"
+          ]
+        }
+      ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -60,24 +74,31 @@
       ]
     },
     {
-      "downloadURL": "mcr.microsoft.com/containernetworking/cni-dropgz:*",
+      "downloadURL": "mcr.microsoft.com/containernetworking/azure-ipam:*",
```
juan-lee, line 77: any risk for this renaming? Anyone expecting the old name?

rbtr: this is an addition, it shows as a rename due to some quirk of git diff. this block has simply moved down: https://github.com/Azure/AgentBaker/pull/4256/files/aec8a803c11e8bd9b9fbc493eb27f212425245cd#diff-1af88d0f330c2b73c9a7dbd8c9ba3c649c0d192c40955f458e4050e62f0bbfd2R92
we still keep the old image name during this transition

PR #4257, opened 2024-04-09
chore: add functions to set/test Windows registry keys

AbelHu:
**What type of PR is this?**
/kind feature

 add functions to set/test Windows registry keys

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/vhdbuilder/packer/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/configure-windows-vhd.ps1
@@ -426,13 +426,117 @@ function Update-WindowsFeatures {
     }
 }
 
+function Enable-WindowsFixInFeatureManagement {
+    Param(
+      [Parameter(Mandatory = $true)][string]
+      $Name,
+      [Parameter(Mandatory = $false)][string]
+      $Value = "1",
+      [Parameter(Mandatory = $false)][string]
+      $Type = "DWORD"
+    )
+
+    $regPath=(Get-Item -Path "HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft" -ErrorAction Ignore)
+    if (!$regPath) {
+        Write-Log "Creating HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft"
+        New-Item -Path "HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft"
+    }
+
+    $regPath=(Get-Item -Path "HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement" -ErrorAction Ignore)
+    if (!$regPath) {
+        Write-Log "Creating HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement"
+        New-Item -Path "HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement"
+    }
+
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -426,13 +426,117 @@ function Update-WindowsFeatures {
...
+        Write-Log "The current value of $Name in FeatureManagement\Overrides is $currentValue"
+    }
+    Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Policies\Microsoft\FeatureManagement\Overrides" -Name $Name -Value $Value -Type $Type
+}
+
+function Enable-WindowsFixInHnsState {
+    Param(
+      [Parameter(Mandatory = $true)][string]
+      $Name,
+      [Parameter(Mandatory = $false)][string]
```
junjiezhang1997, line 468: According to https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-itemproperty?view=powershell-7.4#-type, the content types of ``-Type`` and ``-Value`` are ``RegistryValueKind`` and ``Object``. I'm not sure whether it could cause any issue if we hardcode their types as ``string`` here.

AbelHu: There is no difference between setting Name and Value via Set-ItemProperty and setting them via Enable-WindowsFixInFeatureManagement.
Set-ItemProperty will validate the input and raise an error if it is invalid.

vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -426,13 +426,117 @@ function Update-WindowsFeatures {
...
+      $Value = "1",
+      [Parameter(Mandatory = $false)][string]
+      $Type = "DWORD"
+    )
+    $regPath=(Get-Item -Path $Path -ErrorAction Ignore)
+    if (!$regPath) {
+        Write-Log "Creating $Path"
+        New-Item -Path $Path
```
ShiqianTao, line 524: Question: Do you need to create the upper-level path if the upper-level path does not exist? Like function `Enable-WindowsFixInFeatureManagement` and `Enable-WindowsFixInVfpExtParameters`.

ShiqianTao: I suggest having a generic function to create registry key path like below:
- merge similar code
- return early if exist

```
function Create-RegistryKeyPath {
    Param(
        [Parameter(Mandatory = $true)][string] $Path
    )

    $regPath=(Get-Item -Path $Path -ErrorAction Ignore)
    if (!$regPath) {
        # Create the upper level registry key path if it does not exist
        $upperLevelPath = [IO.Path]::GetDirectoryName($Path)
        Create-RegistryKeyPath -Path $upperLevelPath

        # Create the current registry key path
        Write-Log "Creating $Path"
        New-Item -Path $Path
    }
}
```

AbelHu: I prefer to handle these specific cases one by one instead of using this generic function to handle it as the function `Enable-WindowsFixInVfpExtParameters`.
And currently new registry keys are in FeatureManagement or hns\State. 
Let us do this improvement if it is needed in future.

ShiqianTao (approved):

junjiezhang1997 (approved):

PR #4274, opened 2024-04-10
documentation: polish NB contract and add more comments

Devinwong:
**What type of PR is this?**
/kind cleanup
/kind documentation

Polishing the contract and add more comments

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/clusterconfig.proto
```diff
@@ -20,18 +35,34 @@ message ClusterConfig {
...
+  LoadBalancerSku load_balancer_sku = 1;
+
+  // Specify if master node should be excluded from standard load balancer, default to true 
+  optional bool exclude_master_from_standard_load_balancer = 2;
+
+  // Maximum number of load balancer rules, default to 148
+  optional int32 max_load_balancer_rule_count = 3;
+
+  // Disable outbound SNAT (Source Network Address Translation) for load balancer, default to false
+  optional bool disable_outbound_snat = 4;
```
lilypan26, line 65: sorry should've asked in an earlier PR, but does this need to be optional? Just checked the getter on the bootstrapper and it seems we want to default to false in the nil case anyway so i dont think we need the nil check.

Devinwong: Thanks for catching this. Removed `optional`.

lilypan26 (approved):

PR #4281, opened 2024-04-11
feat: update window cse to install credental provider

mainred:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
...
+    $credentialProviderConfig = @"
+apiVersion: kubelet.config.k8s.io/v1
+kind: CredentialProviderConfig
+providers:
+  - name: acr-credential-provider
+    matchImages:
+      - "*.azurecr.io"
+      - "*.azurecr.cn"
+      - "*.azurecr.de"
+      - "*.azurecr.us"
```
andyzhangx, line 365: what about gov cloud support?

mainred: azurecr.us?
This is how we support matched images in in-tree code
```
	containerRegistryUrls = []string{"*.azurecr.io", "*.azurecr.cn", "*.azurecr.de", "*.azurecr.us"}
	acrRE                 = regexp.MustCompile(`.*\.azurecr\.io|.*\.azurecr\.cn|.*\.azurecr\.de|.*\.azurecr\.us`)
```
https://github.com/kubernetes/kubernetes/blob/release-1.29/pkg/credentialprovider/azure/azure_credentials.go#L53

mainred: This is in L370. It's for sure we can have CustomCloudContainerRegistryDNSSuffix for custom cloud, and the US clouds are configured as custom cloud.

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
     Write-Log "OpenSSH installed and configured successfully"
 }
 
+function Config-CredentialProvider {
```
mainred, line 345: This is not an externally exposed function, but is helpful for test.

staging/cse/windows/configfunc.tests.ps1
```diff
@@ -104,4 +104,73 @@ Describe 'Resize-OSDrive' {
...
+    BeforeEach {
+        $global:KubeDir = "$PSScriptRoot"
+        $CredentialProviderConfPATH=[Io.path]::Combine("$global:KubeDir", "credential-provider-config.yaml")
+    }
+
+    AfterEach {
+        Remove-Item -Path $CredentialProviderConfPATH
+    }
+
+    Context 'CustomCloudContainerRegistryDNSSuffix is empty' {
```
AbelHu, line 119: Could we store the test data as we did for azurecni https://github.com/Azure/AgentBaker/tree/master/staging/cse/windows/azurecnifunc.tests.suites

AbelHu: It is more readable.

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
...
+"@
+    $credentialProviderConfig | Out-File -encoding ASCII -filepath "$CredentialProviderConfPATH"
+}
+
+function Install-CredentialProvider {
+    Param(
+        [Parameter(Mandatory = $false)][string]
+        $CustomCloudContainerRegistryDNSSuffix
+    )
+    Logs-To-Event -TaskName "AKS.WindowsCSE.Install-CredentialProvider" -TaskMessage "Start to install credential provider"
```
AbelHu, line 389: Should we only log this event when it is enabled?

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
...
+
+    $KubeletConfigArgsStr=$global:KubeletConfigArgs -join " "
+    # Starting from Kubernetes 1.30, out of tree credential provider is enabled as a must. Otherwise, related kubelet flags will be set.
```
AbelHu, line 392: How can I understand `Otherwise`

staging/cse/windows/configfunc.tests.ps1
mainred: Sure, wanted to use cni one, but considering the scenario is not not much and the content is little, so I used here-string instead.
Let me update the test code.

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
...
+        Write-Log "Credential provider is enabled"
+
+        Write-Log "Create credential provider configuration file"
+        Config-CredentialProvider -CustomCloudContainerRegistryDNSSuffix $CustomCloudContainerRegistryDNSSuffix
+
+        $CredentialProviderBinDir = "c:\var\lib\kubelet\credential-provider"
+        Write-Log "Download credential provider binary from $global:CredentialProviderURL to $CredentialProviderBinDir"
+        $tempDir = New-TemporaryDirectory
+        $credentialproviderbinaryPackage = "$tempDir\credentialprovider.tar.gz"
+        DownloadFileOverHttp -Url $global:CredentialProviderURL -DestinationPath $credentialproviderbinaryPackage -ExitCode $global:WINDOWS_CSE_ERROR_DOWNLOAD_CREDEDNTIAL_PROVIDER
```
AbelHu, line 403: Suggest to add this new error code in the same PR in csehelper.ps1

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
...
+        tar -xzf $credentialproviderbinaryPackage -C $tempDir
+        Create-Directory -FullPath $CredentialProviderBinDir
+        cp "$tempDir\azure-acr-credential-provider.exe" "$CredentialProviderBinDir\acr-credential-provider.exe"
+        # acr-credential-provider.exe cannot be found by kubelet through provider name before the fix https://github.com/kubernetes/kubernetes/pull/120291
+        # so we copy the exe file to acr-credential-provider to make all 1.29 release work.
```
AbelHu, line 408: What will it happen for k8s version < v1.29?

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,77 @@ function Install-OpenSSH {
...
+        cp "$CredentialProviderBinDir\acr-credential-provider.exe" "$CredentialProviderBinDir\acr-credential-provider"
```
AbelHu, line 409: Will other steps fail? It may be better to use try/catch with an error code

staging/cse/windows/configfunc.tests.ps1
mainred: I just found it hard to handle the kubeDir var in the config yaml file. It's shared both in both test data and in the code.

staging/cse/windows/configfunc.ps1
mainred: Let me update the comment to be more readable.

staging/cse/windows/configfunc.ps1
mainred: Will update it in the following PR

staging/cse/windows/configfunc.ps1
mainred: 1.29 we don't set image-credential-provider-config and image-credential-provider-bin-dir kublet flags, which means we don't install credential provider

mainred: out of tree credential provider is enabled only after 1.29, and is turned on as a must in 1.30. for 1.29 it's optional

mainred: updated the comment to explain when out of tree credential provider is enabled.

staging/cse/windows/configfunc.ps1
mainred: Added try/catch block and a new error code WINDOWS_CSE_ERROR_INSTALL_CREDENTIAL_PROVIDERwill also be added in the following PR.

mainred:
ping @AbelHu for another look. 
One comment I did not address is not the testdata. Because the config file relies on kubedir var, also the content is small, less than 15 line, so I think it's good to keep as it is?

staging/cse/windows/configfunc.tests.ps1
mainred: Fixed by passing kubeDir  as a variable and use relative path for azure.json in testdata

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,92 @@ function Install-OpenSSH {
...
+      - "*.azurecr.cn"
+      - "*.azurecr.de"
+      - "*.azurecr.us"
+    defaultCacheDuration: "10m"
+    apiVersion: credentialprovider.kubelet.k8s.io/v1
+    args:
+      - $azureConfigFile
+"@
+
+    if ($CustomCloudContainerRegistryDNSSuffix) {
```
AbelHu, line 374: ![string]::IsNullOrEmpty($CustomCloudContainerRegistryDNSSuffix)?

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,92 @@ function Install-OpenSSH {
...
+function Install-CredentialProvider {
+    Param(
+        [Parameter(Mandatory=$true)][string]
+        $KubeDir,
+        [Parameter(Mandatory = $false)][string]
+        $CustomCloudContainerRegistryDNSSuffix
+    )
+
+    try {
+        $KubeletConfigArgsStr=$global:KubeletConfigArgs -join " "
```
AbelHu, line 404: Did you test it? Reference: https://github.com/Azure/AgentBaker/blob/master/pkg/agent/datamodel/types.go#L1490-L1528

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,92 @@ function Install-OpenSSH {
...
+            Logs-To-Event -TaskName "AKS.WindowsCSE.Install-CredentialProvider" -TaskMessage "Start to install out of tree credential provider"
+
+            Write-Log "Create credential provider configuration file"
+            Config-CredentialProvider -KubeDir $KubeDir -CustomCloudContainerRegistryDNSSuffix $CustomCloudContainerRegistryDNSSuffix
+
+            $CredentialProviderBinDir = "c:\var\lib\kubelet\credential-provider"
+            Write-Log "Download credential provider binary from $global:CredentialProviderURL to $CredentialProviderBinDir"
+            $tempDir = New-TemporaryDirectory
+            $credentialproviderbinaryPackage = "$tempDir\credentialprovider.tar.gz"
+            DownloadFileOverHttp -Url $global:CredentialProviderURL -DestinationPath $credentialproviderbinaryPackage -ExitCode $global:WINDOWS_CSE_ERROR_DOWNLOAD_CREDEDNTIAL_PROVIDER
```
AbelHu, line 417: Please add the new error codes in parts/windows/windowscsehelper.ps1 in this PR.

staging/cse/windows/debug/collect-windows-logs.ps1
```diff
@@ -77,6 +77,12 @@ if (Test-Path "c:\k\bootstrap-config") {
   $paths += "c:\k\bootstrap-config"
 }
 
+
+if (Test-Path "c:\k\credential-provider-config.yaml") {
```
AbelHu, line 81: Please also create a symbol link for this new file in https://github.com/Azure/AgentBaker/blob/master/staging/cse/windows/provisioningscripts/loggenerator.ps1

staging/cse/windows/configfunc.ps1
mainred: It has been included in https://github.com/Azure/AgentBaker/pull/4282/files#diff-3f509f6352b2fe049ee89937e5f390b0cb2479d64c64b91e108cf5a744b53be3R73

staging/cse/windows/configfunc.ps1
mainred: I tested the code by creating clusters when both these two flags are included and not.

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,92 @@ function Install-OpenSSH {
...
+        $KubeDir,
+        [Parameter(Mandatory = $false)][string]
+        $CustomCloudContainerRegistryDNSSuffix
+    )
+
+    try {
+        $KubeletConfigArgsStr=$global:KubeletConfigArgs -join " "
+        # Out of tree credential provider is turned on as a must after 1.30, and is optinal in 1.29, for cluster < 1.29, it's not enabled.
+        # And only when it's enabled, the credential provider flags are set.
+        if ($KubeletConfigArgsStr -Like "*image-credential-provider-config*" -And $KubeletConfigArgsStr -Like "*image-credential-provider-bin-dir*") {
```
AbelHu, line 407: Since neither is miss, kubelet will raise an error. I think that we should get the values from $global:KubeletConfigArgs, validate them and use the values directly.
Raise a specific error if the value is invalid.

parts/windows/windowscsehelper.ps1
```diff
@@ -68,6 +68,10 @@ $global:WINDOWS_CSE_ERROR_GPU_DRIVER_INSTALLATION_URL_NOT_EXE=61
 $global:WINDOWS_CSE_ERROR_UPDATING_KUBE_CLUSTER_CONFIG=62
 $global:WINDOWS_CSE_ERROR_GET_NODE_IPV6_IP=63
 $global:WINDOWS_CSE_ERROR_GET_CONTAINERD_VERSION=64
+$global:WINDOWS_CSE_ERROR_INSTALL_CREDENTIAL_PROVIDER = 65 # exit code for installing credential provider
+
+$global:WINDOWS_CSE_ERROR_DOWNLOAD_CREDEDNTIAL_PROVIDER=205 # exit code for downloading credential provider failure
```
AbelHu, line 73: Please follow the sequence to use 66,67

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -99,6 +99,7 @@ ERR_K8S_INSTALL_ERR=204 # Error installing or setting up kubernetes binaries on
 ERR_SYSTEMCTL_MASK_FAIL=2 # Service could not be masked by systemctl
 
 ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT=205 # Timeout waiting for credential provider downloads
+ERR_CREDENTIAL_PROVIDER_CONFIG=206 # Error checking credential provider flags
```
AbelHu, line 102: suggest to use a separate pr to check in linux error codes

mainred: removed

staging/cse/windows/configfunc.ps1
mainred: Achieved by checking flag values, and when either flag is not well set or only one var is not empty, we'll raise config error

mainred: This flag check is also protected by test cases.

AbelHu (approved):
LGTM. Thanks for the PR.

AbelHu (changes requested):
Please fix the failed ut

mainred:
> Please fix the failed ut

Removed the UT because it's not valid anymore after we check the _ensured flag name_

mainred:
ping @AbelHu  I think this PR is good to continue?

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,127 @@ function Install-OpenSSH {
...
+        $splitResult=($KubeletConfigArg -split "=")
+        if ($splitResult.Length -ne 2 -or [string]::IsNullOrEmpty($splitResult[1])){
+            Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CREDENTIAL_PROVIDER_CONFIG -ErrorMessage "Failed to get kubelet flag value from flag $KubeletConfigArg"
+        }
+        return $splitResult[1]
+    }
+    $credentialProviderConfigPath = ""
+    $credentialProviderBinDir = ""
+    ForEach ($kubeletConfigArg in $global:KubeletConfigArgs){
+        if ($kubeletConfigArg -like "--image-credential-provider-config*") {
```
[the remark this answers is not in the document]
mainred: Used --image-credential-provider-config=* instead.

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,127 @@ function Install-OpenSSH {
...
+            $credentialProviderConfigPath=get-KubeletFlagValue -KubeletConfigArg $kubeletConfigArg
+        }
+        if ($kubeletConfigArg -like "--image-credential-provider-bin-dir*") {
+            $credentialProviderBinDir=get-KubeletFlagValue -KubeletConfigArg $kubeletConfigArg
+        }
+    }
+    echo $credentialProviderConfigPath
```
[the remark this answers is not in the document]
mainred: Removed echo

staging/cse/windows/configfunc.ps1
```diff
@@ -342,6 +342,127 @@ function Install-OpenSSH {
...
+    echo $credentialProviderBinDir
+
+    # Both flags should be set to enable out of tree credential provider or not set at the same time to disable it.
+    if ([string]::IsNullOrEmpty($credentialProviderConfigPath) -xor [string]::IsNullOrEmpty($credentialProviderBinDir)) {
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_CREDENTIAL_PROVIDER_CONFIG -ErrorMessage "Not all credential provider flags are configured: --image-credential-provider-config=$credentialProviderConfigPath, --image-credential-provider-bin-dir=$credentialProviderBinDir"
+    }
+    return $credentialProviderConfigPath, $credentialProviderBinDir
```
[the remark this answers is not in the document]
mainred: There seems to be some latency in listing the new comments.
I used global var instead to fetch the config file and bin path.

staging/cse/windows/configfunc.ps1
mainred: Done

AbelHu (approved):
LGTM. Thanks for the PR.

PR #4282, opened 2024-04-11
feat: install credential provider support for windows and cache binary in VHD

mainred:
**What type of PR is this?**

install credential provider support for windows and cache binary in VHD

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


mainred:
This PR relies on https://github.com/Azure/AgentBaker/pull/4281

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -410,7 +410,7 @@ start_watch
 
 # Kubelet credential provider plugins
 CREDENTIAL_PROVIDER_VERSIONS="
-1.29.3
+1.29.2
```
mainred, line 413: windows binary of 1.29.3 is different from global and mooncake for cdn cache issue.
We use 1.29.2 instead, which has the same code as 1.29.3

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -292,7 +295,9 @@ try
     Get-LogCollectionScripts
     
     Write-KubeClusterConfig -MasterIP $MasterIP -KubeDnsServiceIp $KubeDnsServiceIp
-    
+
+    Install-CredentialProvider -KubeDir $global:KubeDir -CustomCloudContainerRegistryDNSSuffix {{if IsAKSCustomCloud}}"{{ AKSCustomCloudContainerRegistryDNSSuffix }}"{{else}}""{{end}}
```
AbelHu, line 299: Prefer to return empty when IsAKSCustomCloud is true in baker.go instead of add this condition check in this ps1 even if it is a template.

AbelHu (approved):
LGTM. Thanks for the PR.

parts/windows/kuberneteswindowssetup.ps1
mainred: Considerations here are:
1. AKSCustomCloudContainerRegistryDNSSuffix requires checking CustomCloudEnv empty first, and we have a lot of functions related to CustomCloud, so to keep the change minimal
```
		"AKSCustomCloudContainerRegistryDNSSuffix": func() string {
			return cs.Properties.CustomCloudEnv.ContainerRegistryDNSSuffix
		},
```
2. IsAKSCustomCloud is referenced several times already in window setup script, so it's safe to use

PR #4289, opened 2024-04-12
perf: parallelize container image and kube-proxy image pulls

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR implements parallelization for container image and kube-proxy image pulls. This decreases the amount of time it takes to pull container images by 1 minute and 43 seconds on average, or 29%. During testing it also decreased the variance of time required to complete all container image pulls through multiple runs by 2 minutes and 15 seconds, or 93%.

During testing, more than 8 concurrent pulls has not been observed. However, because VM size, disk tier, and networking resources could be improved in the future, I have set the `parallel_container_image_pull_limit` variable to the number of processor cores minus two. This was done in order to prevent potential issues with network, CPU, and disk resources while also allowing the number of parallel pulls to increase if more resources become available in the future.

Unsatisfactory build pipeline duration.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -295,9 +297,14 @@ for imageToBePulled in ${ContainerImages[*]}; do
 
   for version in ${versions}; do
     CONTAINER_IMAGE=$(string_replace $downloadURL $version)
-    pullContainerImage ${cliTool} ${CONTAINER_IMAGE}
+    pullContainerImage ${cliTool} ${CONTAINER_IMAGE} &
+    containerImagePids+=($!)
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
+    while [[ $(jobs -p | wc -l) -ge 13 ]]; do # Set maximum number of parallel pulls to 13
```
ganeshkumarashok, line 303: Can you add the data you used for determine 13 as the limit for parallel pulls?

ganeshkumarashok: And move this to a separate variable?

ganeshkumarashok: This needs to be explainable to someone who sees this in the future who doesn't have the context we do

ganeshkumarashok: Is it possible to calculate/estimate 13 based on the VM spec? Let's say we switch to a different VM size in the build pipeline, it'll be nice to adjust this dynamically,

ganeshkumarashok:
nit: remove branch name from PR title

ganeshkumarashok:
Can you also add the approx time improvement from this parallelization in the PR description?

ganeshkumarashok (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -278,6 +278,18 @@ string_replace() {
   echo ${1//\*/$2}
 }
 
+# Limit number of parallel pulls to 2 less than number of processor cores in order to prevent issues with network, CPU, and disk resources
+# Account for possibility that number of cores is 3 or less
+num_proc=$(nproc)
+if [[ $num_proc -gt 3 ]]; then
```
ganeshkumarashok, line 284: Do you want this for >= 3 or just > 3? I assume the former - so you can switch to `ge` instead of `gt`

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -449,11 +466,23 @@ rm -r /var/log/azure/Microsoft.Azure.Extensions.CustomScript || exit 1
 
 KUBE_PROXY_IMAGE_VERSIONS=$(jq -r '.containerdKubeProxyImages.ContainerImages[0].multiArchVersions[]' <"$THIS_DIR/kube-proxy-images.json")
 
+declare -a kube_proxy_pids=()
+
+for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
+  # use kube-proxy as well
```
ganeshkumarashok, line 472: remove this comment?

vhdbuilder/packer/install-dependencies.sh
zachary-bailey: If the number of processors is 3, it would result in 1, which is the same as the else statement. It would not make a difference either way that I can tell. Do you want me to change it?

vhdbuilder/packer/install-dependencies.sh
zachary-bailey: Haha I felt the same way! It has been there for a while I did not make it, I considered removing but wasn't sure why it was there so I didn't

zachary-bailey: I can remove if you would like, from both loops

ganeshkumarashok: That'll be neat, thanks!

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Fair, doesn't matter then

PR #4290, opened 2024-04-13
perf: run installBcc in a subshell

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR modifies the way BCC tools are installed on ab VHDs. The installBcc function is ran in a subshell so that the main script can continue to execute while bcc packages are downloaded / compiled / installed in the background. 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests


ganeshkumarashok:
What do you think about adding the tests for successful installation to this PR?

zachary-bailey:
> What do you think about adding the tests for successful installation to this PR?

@ganeshkumarashok Can do!

ganeshkumarashok:
Could you note the approx time saved in the PR description as well?

zachary-bailey:
> Could you note the approx time saved by parallelization, in the PR description as well?

Do you mean time saved by running it in the subshell? or marking time for both, in the other PR as well

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -258,11 +258,17 @@ ls -ltr /opt/gpu/* >> ${VHD_LOGS_FILEPATH}
 installBpftrace
 echo "  - $(bpftrace --version)" >> ${VHD_LOGS_FILEPATH}
 
-installBcc
-cat << EOF >> ${VHD_LOGS_FILEPATH}
-  - bcc-tools
-  - libbcc-examples
-EOF
+PARENT_DIR=$(pwd)
```
ganeshkumarashok, line 261: It should be the present directory right? Not parent directory?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -258,11 +258,17 @@ ls -ltr /opt/gpu/* >> ${VHD_LOGS_FILEPATH}
...
+
```
ganeshkumarashok, line 262: nit: add a comment about creating a subshell to install in the background? This is not a common pattern in Agentbaker so it'll be good to note for the future

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -258,11 +258,17 @@ ls -ltr /opt/gpu/* >> ${VHD_LOGS_FILEPATH}
...
+( 
+  cd $PARENT_DIR || { echo "Subshell in the wrong directory" >&2; exit 1; }
+
+  installBcc
+
+  exit $?
+) > /tmp/bcc.log 2>&1 &
```
ganeshkumarashok, line 269: /tmp directory may get deleted on reboots. Can you save it in `/var/log/bcc_installation.log` instead? Those logs will remain

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -858,6 +858,50 @@ testContainerImagePrefetchScript() {
...
+    not_executable=$((not_executable + 1))
+    notExecutableTools+=("$tool")
+    echo "Not Executable: $tool"
+  fi
+done
+
+echo "Total tools: $((not_executable + executable))"
+echo "Executable Tools: $executable"
+echo "Not Executable Tools: $not_executable"
+
```
ganeshkumarashok, line 881: Add a check for Azure Linux/Mariner 2.0 as well?

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -858,6 +858,50 @@ testContainerImagePrefetchScript() {
...
+if [ "$os" == "Ubuntu" ]; then
+  if [ "$executable" -ne 127 ] || [ "$not_executable" -ne 2 ]; then
```
ganeshkumarashok, line 883: We should not hardcode these number of tools since that may change based on the version or if we install other tools -- that might lead to manual changes necessary for this test later

ganeshkumarashok: Can we have a better basic verification of bcc that doesn't actually need counting of installed binaries or at least avoids hardcoded counts?

ganeshkumarashok: Is it possible to get expected counts programatically before this function is run, and then pass that as input parameters to this function? In that case you can count the expected files as well.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -858,6 +858,20 @@ testContainerImagePrefetchScript() {
...
 }
 
+testBccTools () {
+  result=0 
+  for line in '- bcc-tools' '- libbcc-examples'; do
+    if ! grep -F -x -e "$line" /opt/azure/vhd-install.complete; then
+      echo "$line is missing"
+      result=1
+    fi
+  done
```
ganeshkumarashok, line 868: Can be simplified - just return 1 inside the `then` condition if line is missing

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -858,6 +858,20 @@ testContainerImagePrefetchScript() {
...
+  if [ $result -eq 0 ]; then
```
ganeshkumarashok, line 869: This section can be removed and simplified

ganeshkumarashok (approved):

PR #4300, opened 2024-04-16
feat: support IMDS restriction for Linux nodes

norshtein:
**What type of PR is this?**
/kind feature

This PR adds support for IMDS restriction for Linux nodes

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```
feat: support IMDS restriction for Linux nodes
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -744,4 +744,80 @@ EOF
...
+    local insertRuleToMangleTable=${1:-false}
+    if [[ $insertRuleToMangleTable == true ]]; then
+        echo "Before inserting IMDS restriction rule to mangle table, checking whether the rule already exists..."
+        iptables -t mangle -S | grep -- '-d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker esnureIMDSRestriction for IMDS restriction feature" -j DROP'
+        if [[ $? -eq 0 ]]; then
+            echo "IMDS restriction rule already exists in mangle table, returning..."
+            return
+        fi
+        echo "Inserting IMDS restriction rule to mangle table..."
+        iptables -t mangle -I FORWARD 1 ! -s "$primaryNicIP" -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker esnureIMDSRestriction for IMDS restriction feature" -j DROP || exit $ERR_MANIPULATE_IPTABLES
```
cameronmeissner, line 782: lets use separate exit codes for each failure here, e.g. `ERR_INSERT_IMDS_RESTRICTION_RULE_INTO_MANGLE`, `ERR_INSERT_IMDS_RESTRICTION_RULE_INTO_FILTER` - same goes for deletion as well

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -199,6 +199,13 @@ if [ "${NEEDS_DOCKER_LOGIN}" == "true" ]; then
     set -x
 fi
 
+# Before setting up kubelet, ensure IMDS restriction iptables rules so that all Pods can get desired IMDS access
```
cameronmeissner, line 202: just for my understanding, is the only requirement here that we perform this step before we start the kubelet service?

pkg/agent/baker_test.go
```diff
@@ -1328,6 +1328,19 @@ oom_score = 0
 					},
 				}
 			}, nil),
+		Entry("AKSUbuntu2204 IMDSRestriction with enable restriction and insert to mangle table", "AKSUbuntu2204+IMDSRestrictionOnWithMangleTable", "1.24.2",
+			func(config *datamodel.NodeBootstrappingConfiguration) {
+				config.EnableIMDSRestriction = true
+				config.InsertIMDSRestrictionRuleToMangleTable = true
```
cameronmeissner, line 1334: can we add assertions on the output here? e.g. contents of the new vars added to cse_cmd.sh to make sure they're getting the correct corresponding values?

see up above for examples

pkg/agent/datamodel/types.go
```diff
@@ -1696,6 +1696,12 @@ type NodeBootstrappingConfiguration struct {
 	SSHStatus                              SSHStatus
 	DisableCustomData                      bool
 	OutboundType                           string
+	EnableIMDSRestriction                  bool
+	// InsertIMDSRestrictionRuleToMangleTable is only checked when EnableIMDSRestriction is true.
+	// When this is true, iptables rule will be inserted to `mangle` table. This is for Linux Cilium
+	// CNI, which will overwrite the `filter` table so that we can only insert to `mangle` table to avoid
+	// our added rule is overwritten by Cilium.
+	InsertIMDSRestrictionRuleToMangleTable bool
```
cameronmeissner, line 1704: so when we want to restrict IMDS but we're not using cilium, this will be false?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -744,4 +744,80 @@ EOF
     fi
 }
 
+getPrimaryNicIP() {
+    local sleepTime=1
+    local maxRetries=10
+    local i=0
+    local ip=""
+
+    while [[ $i -lt $maxRetries ]]; do
```
cameronmeissner, line 753: nit: we already have a helper which does something very similar which you can probably use instead: https://github.com/Azure/AgentBaker/blob/45392616eccd6b2bc9b8b9ed89416a1ba75715c4/parts/linux/cloud-init/artifacts/cse_helpers.sh#L128C1-L140C2

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -158,4 +158,6 @@ IS_KATA="{{IsKata}}"
 ARTIFACT_STREAMING_ENABLED="{{IsArtifactStreamingEnabled}}"
 SYSCTL_CONTENT="{{GetSysctlContent}}"
 PRIVATE_EGRESS_PROXY_ADDRESS="{{GetPrivateEgressProxyAddress}}"
+ENABLE_IMDS_RESTRICTION="{{EnableIMDSRestriction}}"
```
anujmaheshwari1, line 161: @Bryce-Soghigian for changes to cse_cmd.sh, although only additive
@lilypan26 @Devinwong for bootstrap contract

tallaxes: Do we have a way of producing an image (off PR?) that one can directly test with?

tallaxes: (Also please tag @Azure/karpenter rather than @Bryce-Soghigian directly. Thanks!)

parts/linux/cloud-init/artifacts/cse_config.sh
norshtein: Makes great sense, updated.

parts/linux/cloud-init/artifacts/cse_main.sh
norshtein: yep, only need to ensure the rule is added before starting kubelet.

pkg/agent/baker_test.go
norshtein: thanks for reminding! added validation.

pkg/agent/datamodel/types.go
norshtein: yep, exactly the case, rp will check and pass corresponding value.

parts/linux/cloud-init/artifacts/cse_config.sh
norshtein: Actually I also noticed the `retrycmd_if_failure` function, but it does not support return value via `echo "$ip"`, so I manually added some retry logic here.

One more point I just realized, do you have any suggestion on whether we should add retry for manipulating iptables rule in `ensureIMDSRestrictionRule` and `disableIMDSRestriction`?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -744,4 +744,80 @@ EOF
...
+        ip=$(curl -sSL -H "Metadata: true" "http://169.254.169.254/metadata/instance/network/interface?api-version=2021-02-01" | jq -r '.[].ipv4.ipAddress[0].privateIpAddress')
```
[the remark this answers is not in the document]
cameronmeissner: supported API versions are here: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#supported-api-versions, though I think this particular version is used in many examples within docs

should be fine for now - looks like they're currently releasing a new version for the first time in a couple years

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: probably a good question for @wedaly

cameronmeissner: `Actually I also noticed the retrycmd_if_failure function, but it does not support return value via echo "$ip", so I manually added some retry logic here.` fair enough... though you should be able to accomplish this using the existing function if you use a global var, though that is somewhat messy as well

I'm fine either way I suppose

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -158,4 +158,6 @@ IS_KATA="{{IsKata}}"
...
+INSERT_IMDS_RESTRICTION_RULE_TO_MANGLE_TABLE="{{InsertIMDSRestrictionRuleToMangleTable}}"
```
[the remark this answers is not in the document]
tallaxes: Nothing to add here, strongly agree with reducing the contract as much as possible. Depending on shell capabilities/needs, "explicit presence"/nillability on a single field at the contract level _could_ be translated into multiple shell/environment variables, if needed. But the simpler - the better.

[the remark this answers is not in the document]
norshtein: I think we can't only use 1 variable `INSERT_IMDS_RESTRICTION_RULE_TO_MANGLE_TABLE` handling all cases: when `INSERT_IMDS_RESTRICTION_RULE_TO_MANGLE_TABLE` is `false`, there're 2 cases:
1. IMDS restriction is disabled.
2. IMDS restriction is enabled, and we don't need to insert the rule to mangle table.

We can't distinguish above 2 cases when we only have 1 variable.

parts/linux/cloud-init/artifacts/cse_config.sh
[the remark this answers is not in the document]
norshtein: yes, I think the fixed API version should be fine just as Cameron pointed out.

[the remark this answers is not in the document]
norshtein: yes, I think the fixed API version should be fine just as Cameron pointed out.

norshtein:
Hi @cameronmeissner  @anujmaheshwari1 @Azure/karpenter , I have updated the PR based on your feedback, could you please have a look? 

btw, the linter failure is unrelated to this PR, seems something went wrong in linter rule?

cameronmeissner (approved):

norshtein:
All CI passed except the linter: https://github.com/Azure/AgentBaker/actions/runs/8966676633/job/24622682804?pr=4300 . The linter failure is irrelevant to this PR:

```
$ ./hack/tools/bin/golangci-lint run
WARN [linters_context] rowserrcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
WARN [linters_context] wastedassign is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649. 
pkg/agent/datamodel/types.go:1641:6: exported fields should be annotated with the "json" tag (musttag)
type GetLatestSigImageConfigRequest struct {
     ^
pkg/agent/datamodel/sig_config.go:281:6: exported fields should be annotated with the "json" tag (musttag)
type SigImageConfigTemplate struct {
     ^
pkg/agent/datamodel/types.go:1650:6: exported fields should be annotated with the "json" tag (musttag)
type NodeBootstrappingConfiguration struct {
     ^
pkg/agent/datamodel/types.go:1718:6: exported fields should be annotated with the "json" tag (musttag)
type NodeBootstrapping struct {
     ^
pkg/agent/baker_test.go:13:2: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
pkg/vhdbuilder/datamodel/component_configs_test.go:4:2: SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details. (staticcheck)
        "io/ioutil"
        ^
```

Going to merge this PR.

PR #4303, opened 2024-04-16
fix: Fix the k8s node name in live patching scripts

hongbin:
**What type of PR is this?**

/kind bug

Fix the k8s node name in live patching scripts. In before, we assume k8s node name is the hostname. However, this is a wrong assumption. The Azure cloud provider set k8s node name as the lower case of the hostname. As a result, if the hostname contains upper case letter, the live patching script won't locate the node properly.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
```diff
@@ -27,6 +27,9 @@ if [ -z "${node_name}" ]; then
     exit 1
 fi
 
+# Azure cloud provider assigns node name as the lowner case of the hostname
```
cameronmeissner, line 30: I was actually just grappling with this on some unrelated work that I'm doing, were you able to find any documentation which clearly states this?

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
hongbin: I couldn't find any documentation. Here is the code I could find:

https://github.com/kubernetes/component-helpers/blob/master/node/util/hostname.go
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/52bfaa6fd236200710cf2fab550a34a802a0fa86/pkg/provider/azure_instances_v1.go#L284

PR #4306, opened 2024-04-17
Alburgess/a bto r pns

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +15,30 @@ const (
...
+
+//nolint:gochecknoinits
+func init() {
+	CacheManifest()
+	CacheComponents()
+}
+
+func CacheManifest() {
+	_, filename, _, _ := runtime.Caller(0)
+	manifestFilePath := "../../../parts/linux/cloud-init/artifacts/manifest.json"
```
cameronmeissner, line 32: can't we just read from the relative path directly?

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,24 @@ func IndentString(original string, spaces int) string {
...
+			return data[:newlineIndex]
+		}
+	}
+	return data
+}
+
+func processDownloadURL(downloadURL string) string {
+	// example URL "downloadURL": "mcr.microsoft.com/oss/kubernetes/autoscaler/addon-resizer:*",
+	// getting the data between the last / and the last :
+	parts := strings.Split(downloadURL, "/")
```
cameronmeissner, line 111: we should probably be checking lens of the slices return from the `Split` calls to ensure that the URL is in the correct format while processing

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,24 @@ func IndentString(original string, spaces int) string {
...
+	if eofIndex != -1 { // #EOF found
+		newlineIndex := bytes.LastIndex(data[:eofIndex], []byte("\n"))
+		if newlineIndex != -1 {
+			return data[:newlineIndex]
+		}
+	}
+	return data
+}
+
+func processDownloadURL(downloadURL string) string {
```
cameronmeissner, line 108: nit: `processContainerImageTag` or similar

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,24 @@ func IndentString(original string, spaces int) string {
 	}
 	return out.String()
 }
+
+func trimEOF(data []byte) []byte {
```
cameronmeissner, line 97: can probably just use this instead: https://pkg.go.dev/bytes#ReplaceAll

pkg/agent/datamodel/sig_config.go
```diff
@@ -896,3 +923,112 @@ func withSubscription(subscriptionID string) SigImageConfigOpt {
...
+}
+
+type ProcessedManifest struct {
+	Versions  []string
+	Pinned    map[string]string
+	Edge      string
+	Installed map[string]string
+}
+
+func getCachedVersionsFromManifestJSON(manifestFilePath string) {
```
cameronmeissner, line 956: I'd probably prefer this to return the manifest and then within the init() block we actually do the package-level assignment to `CachedFromManifest` - mainly these functions become a bit easier to test independently

pkg/agent/bakerapi.go
```diff
@@ -174,3 +175,7 @@ func findSIGImageConfig(sigConfig datamodel.SIGAzureEnvironmentSpecConfig, distr
 
 	return nil
 }
+
+func (agentBaker *agentBakerImpl) GetCachedVersionsOnVHD() (map[string]datamodel.ProcessedManifest, map[string]datamodel.ProcessedComponents, error) {
+	return datamodel.CachedFromManifest, datamodel.CachedFromComponents, nil
```
cameronmeissner, line 180: we'd probably want to check whether these are actually instantiated and return an error if not, right now the error return value is unused

pkg/agent/datamodel/sig_config.go
```diff
@@ -896,3 +923,112 @@ func withSubscription(subscriptionID string) SigImageConfigOpt {
 		c.SubscriptionID = subscriptionID
 	}
 }
+
+type Manifest struct {
```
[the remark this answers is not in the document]
cameronmeissner: I'd also consider creating a separate `cache` package as well

pkg/agent/datamodel/sig_config.go
cameronmeissner: hoping @element-of-surprise's change can make it in so we could just take from the embedded FS instead...

pkg/agent/bakerapi_test.go
```diff
@@ -523,4 +526,22 @@ var _ = Describe("AgentBaker API implementation tests", func() {
...
+
+	Context("GetCachedVersionsOnVHD", func() {
+		It("should return cached VHD data", func() {
+			agentBaker, err := NewAgentBaker()
+			Expect(err).NotTo(HaveOccurred())
+
+			fromManifest, fromComponents, err := agentBaker.GetCachedVersionsOnVHD()
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(fromManifest["runc"].Installed["default"]).To(Equal("1.1.12"))
```
cameronmeissner, line 538: yeah... updating this would be annoying but we do something similar already with sig_config sorta, so not the end of the world - best bet would be to manually load in those files during the test and compare with the current state of the values rather than hard-coding here

pkg/agent/datamodel/sig_config.go
AlisonB319: We cannot add the relative path, when you run a go test, it assumes you are at whatever directory you are running 'go test' which causes problems where it thinkings the path is empty. Also we did explore the embedded path, but since the manifest.json and components.json are not in the same directory, it didn't work. Also manifest has the #EOF at the end that makes it tricky also.

pkg/agent/datamodel/helper.go
AlisonB319: Actually in the function I do 

lastPart := parts[len(parts)-1]
component := strings.Split(lastPart, ":")

so that I'm always looking at the last / to the final : "/addon-resizer:"

so instead of a / check, I think ensuring that component[0] exists is what will be important

pkg/agent/datamodel/sig_config.go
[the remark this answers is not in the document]
AlisonB319: In my one pager. Will Daly left the comment 

I know we've talked before about having the CSE fallback to whatever is cached on the VHD, which has its own drawbacks. But I feel like it's a better user experience for nodes to continue running outdated versions of components, rather than trying to block every case where the version might be out-of-date.

so if we want to consider doing this, processing the json into structs will be more useful. Defaulting to a not desired, but cache version would probably make the user experience better. But it could leave us open to some ambiguity of customers thinking they are using one version when in actuality the are getting another

apiserver/getcachedvhdimages.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+)
+
+const (
+	// RoutePathGetCachedVersionsOnVHD the route path to get cached vhd images.
+	RoutePathGetCachedVersionsOnVHD string = "/getcachedversionsonvhd"
+)
+
+type CachedOnVHD struct {
+	CachedFromManifest   map[string]datamodel.ProcessedManifest `json:"cached_from_manifest"`
```
cameronmeissner, line 19: nit: I think we usually go for camelCase when defining JSON keys

apiserver/getcachedvhdimages.go
```diff
@@ -0,0 +1,52 @@
...
+
+	result := CachedOnVHD{
+		CachedFromManifest:   cachedFromManifest,
+		CachedFromComponents: cachedFromComponents,
+	}
+
+	jsonResponse, err := json.Marshal(result)
+	if err != nil {
+		log.Println(err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
cameronmeissner, line 47: lol, I guess we do this in other endpoints, though we really should be returning a 5xx here instead

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,19 @@ func IndentString(original string, spaces int) string {
 	}
 	return out.String()
 }
+
+func processContainerImageTag(downloadURL string) (string, error) {
```
cameronmeissner, line 97: sorry I know I recommended this before, but can we do `processContainerImage` instead since these technically aren't tags, rather wild-carded repos

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,19 @@ func IndentString(original string, spaces int) string {
...
+	// example URL "downloadURL": "mcr.microsoft.com/oss/kubernetes/autoscaler/addon-resizer:*",
+	// getting the data between the last / and the last :
+	parts := strings.Split(downloadURL, "/")
+	if len(parts) == 0 || len(parts[len(parts)-1]) == 0 {
+		return "", errors.New("downloadURL is not in the expected format")
```
cameronmeissner, line 102:
```suggestion
		return "", fmt.Errorf("container image component URL is not in the expected format: %s", downloadURL)
```

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,19 @@ func IndentString(original string, spaces int) string {
...
+	}
+	lastPart := parts[len(parts)-1]
```
cameronmeissner, line 104: can we actually just do `components = strings.TrimSuffix(lastPart, ":*")` instead?

pkg/agent/datamodel/sig_config.go
cameronmeissner: yeah that was using raw embedding, though John's change would have it such that we can access all embedded files essentially as its own filesystem, where all paths are accessible - though as I mentioned that's not available as of yet, so this is fine for now

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,151 @@ const (
...
+	ContainerImages []struct {
+		DownloadURL           string   `json:"downloadURL"`
+		Amd64OnlyVersions     []string `json:"amd64OnlyVersions"`
+		MultiArchVersions     []string `json:"multiArchVersions"`
+		PrefetchOptimizations []struct {
+			Version  string   `json:"version"`
+			Binaries []string `json:"binaries"`
+		} `json:"prefetchOptimizations"`
+	} `json:"ContainerImages"`
+}
```
cameronmeissner, line 152:
```suggestion
}

```

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,151 @@ const (
...
+		panic(err)
+	}
+	var components Components
+	if err = json.Unmarshal(data, &components); err != nil {
+		panic(err)
+	}
+	return components
+}
+
+type Manifest struct {
```
cameronmeissner, line 113: I know I mentioned just doing raw JSON before but I think what you have is fine for now, can always revisit later

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,151 @@ const (
...
+
+type ProcessedManifest struct {
+	Versions  []string
+	Pinned    map[string]string
+	Edge      string
+	Installed map[string]string
+}
+
+type Components struct {
+	ContainerImages []struct {
```
cameronmeissner, line 143: since we won't do raw JSON we can just do something like this instead:
```
type Components struct {
    ContainerImages []ContainerImage `json:"containerImages"`
}

type ContainerImage struct {
	MultiArchVersions     []string `json:"multiArchVersions"`
	Amd64OnlyVersions     []string `json:"amd64OnlyVersions"`
	PrefetchOptimizations []PrefetchOptimization `json:"prefetchOptimizations"`
}

type PrefetchOptimization struct {
	Version  string `json:"version"`
	Binaries []string `json:"binaries"`
}
```

cameronmeissner: btw, do we also need to include these? https://github.com/Azure/AgentBaker/blob/ddad5fa23550576d200628a1b21bae6ab7d689a9/vhdbuilder/packer/components.json#L368C1-L408C4

apiserver/getcachedvhdimages.go
AlisonB319: We actually need to keep these capitalized so that they are properly exported. Looking through other files, seems like the JSON keys are all capitalized too.

cameronmeissner: err I meant the json tag

e.g. `cached_from_manifest` -> `cachedFromManifest`

and yes you must always export fields which are tagged with JSON so the json unmarshaller can actually access those fields, otherwise you'll get zero-values everywhere (I totally haven't made this mistake numerous times).

pkg/agent/bakerapi_test.go
```diff
@@ -523,4 +526,33 @@ var _ = Describe("AgentBaker API implementation tests", func() {
...
+			agentBaker, err := NewAgentBaker()
+			Expect(err).NotTo(HaveOccurred())
+
+			fromManifest, fromComponentsContainerImages, fromComponentsDownloadFiles, err := agentBaker.GetCachedVersionsOnVHD()
+			Expect(err).NotTo(HaveOccurred())
+
+			manifest := datamodel.CacheManifest()
+			component := datamodel.CacheComponents()
+
+			// The indices are hardcoded based on the current components.json.
```
cameronmeissner, line 541: would we ever need to update these?

pkg/agent/bakerapi.go
```diff
@@ -174,3 +175,11 @@ func findSIGImageConfig(sigConfig datamodel.SIGAzureEnvironmentSpecConfig, distr
 
 	return nil
 }
+
+func (agentBaker *agentBakerImpl) GetCachedVersionsOnVHD() (map[string]datamodel.ProcessedManifest, map[string]datamodel.ContainerImage,
```
cameronmeissner, line 179: consider having this return a `CachedOnVHD` instead?

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,151 @@ const (
...
+		panic(err)
+	}
+	var components Components
+	if err = json.Unmarshal(data, &components); err != nil {
+		panic(err)
+	}
+	return components
+}
+
+type Manifest struct {
```
cameronmeissner, line 111: this type along with `ProcessedManifest` feels redundant

why not just do 

```
type Dependency struct {
	Versions []string `json:"versions"`
         Installed map[string]string `json:"installed"`
         Pinned map[string]string `json:"pinned'`
         Edge string `json:"edge"`
}

type Manifest struct {
       Containerd Dependency `json:"containerd"`
       Runc Dependency `json:"runc"`
       NvidiaContainerRuntime `json:"nvidia-container-runtime"`
       NvidiaDrivers `json:"nvidia-drivers"`
       Kubernetes `json:"kubernetes"`
}
```

cameronmeissner: https://github.com/Azure/AgentBaker/blob/master/schemas/manifest.cue

cameronmeissner: it also make me sad since I'm trying to rework that cue schema for components/manifests to support the effort for configurable private builds; if you're gonna be picking up some of the security work then I might end up handing that off to you :)

for ref: https://github.com/Azure/AgentBaker/pull/4171

pkg/agent/bakerapi.go
anujmaheshwari1: was going to say the same, seems cleaner?

pkg/agent/bakerapi.go
```diff
@@ -174,3 +175,11 @@ func findSIGImageConfig(sigConfig datamodel.SIGAzureEnvironmentSpecConfig, distr
...
+	map[string]datamodel.DownloadFiles, error) {
+	if datamodel.CachedFromManifest == nil || datamodel.CachedFromComponentContainerImages == nil || datamodel.CachedFromComponentDownloadedFiles == nil {
```
anujmaheshwari1, line 181: should the conditions be && instead of ||?

it looks like right now, if any one of them is nil, we will return nil for all the others as well, that doesn't seem right?

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +93,24 @@ func IndentString(original string, spaces int) string {
...
+	component := strings.TrimSuffix(lastPart, ":*")
+	return component, nil
+}
+
+func processDownloadFile(downloadURL string) (string, error) {
+	// example URL "downloadURL": "https://acs-mirror.azureedge.net/cni-plugins/v*/binaries",
+	parts := strings.Split(downloadURL, "/")
+	if len(parts) == 0 || len(parts) != 6 {
```
anujmaheshwari1, line 112: can you please remind me why this split will always by 6? could this change in the future? having a hardcoded number in the API doesn't seem best but I want to understand the reasoning behind this to see if its unavoidable.

pkg/agent/bakerapi_test.go
AlisonB319: Only if someone adds a new binary to components.json and messes up the indexing. Shouldn't matter if they add to the end of the file

pkg/agent/bakerapi_test.go
```diff
@@ -21,6 +21,9 @@ var _ = Describe("AgentBaker API implementation tests", func() {
 	BeforeEach(func() {
 		toggles = agenttoggles.New()
 
+		datamodel.CacheManifest()
```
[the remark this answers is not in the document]
AlisonB319: it is not needed. I must have been doing something weird initially, because I definitely needed it last week.

pkg/agent/bakerapi.go
AlisonB319: I could do this 

```
       cached := datamodel.CachedOnVHD{
		CachedFromManifest:                 datamodel.CachedFromManifest,
		CachedFromComponentContainerImages: datamodel.CachedFromComponentContainerImages,
		CachedFromComponentDownloadedFiles: datamodel.CachedFromComponentDownloadedFiles,
	}

	if datamodel.CachedFromManifest == nil || datamodel.CachedFromComponentContainerImages == nil || datamodel.CachedFromComponentDownloadedFiles == nil {
		return cached, fmt.Errorf("cached versions are not available")
	}

	return cached, nil
```

that way if something exists it will be returned. But I think || makes sense, otherwise it will be a heavy conditional

pkg/agent/datamodel/helper.go
AlisonB319: https://acs-mirror.azureedge.net/cni-plugins/v*/binaries when split by "/" makes 6 indices 

[https:, "", acs-mirror.azureedge.net, cni-plugins, v*, binaries]

so I agree its not perfect, all the urls do follow this same pattern. If that was to change, we would need to adjust this function

pkg/agent/bakerapi_test.go
cameronmeissner: hmm, might want to add a note somewhere about that

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,136 @@ const (
...
+}
+
+type CachedOnVHD struct {
+	CachedFromManifest                 *Manifest                 `json:"cached_from_manifest"`
+	CachedFromComponentContainerImages map[string]ContainerImage `json:"cached_from_component_container_images"`
+	CachedFromComponentDownloadedFiles map[string]DownloadFiles  `json:"cached_from_component_downloaded_files"`
+}
+
+type Dependency struct {
+	FileName         string            `json:"fileName"`
```
cameronmeissner, line 104: feel free to omit FileName/DownloadLocation/DownloadURL if you're not using them

pkg/agent/datamodel/helper.go
cameronmeissner: maybe we could utilize this instead? https://pkg.go.dev/net/url

pkg/agent/datamodel/sig_config.go
AlisonB319: Do you think there's any value for passing these along - maybe not for this rp ask, but for anything down the line? Otherwise I can remove FileName, DownloadLocation, DownloadURL

pkg/agent/datamodel/helper.go
AlisonB319: I still had to utilize some string slicing, but maybe I was able to use url.Parse() to make it a bit cleaner

pkg/agent/datamodel/sig_config.go
cameronmeissner: I don't really think there's much value, I'd probably prefer omitting them

we can always add them back later if we think we need them

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +94,25 @@ func IndentString(original string, spaces int) string {
...
+	parts := strings.Split(downloadURL, "/")
+	if len(parts) == 0 || len(parts[len(parts)-1]) == 0 {
+		return "", fmt.Errorf("container image component URL is not in the expected format: %s", downloadURL)
+	}
+	lastPart := parts[len(parts)-1]
+	component := strings.TrimSuffix(lastPart, ":*")
+	return component, nil
+}
+
+func processDownloadFile(downloadURL string) (string, error) {
```
cameronmeissner, line 110: since this is re returning the name of the binary, can we rename to something like `getComponentNameFromURL`, and `getContainerImageNameFromURL`?

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +94,25 @@ func IndentString(original string, spaces int) string {
...
+	// example URL "downloadURL": "https://acs-mirror.azureedge.net/cni-plugins/v*/binaries",
+	url, err := url.Parse(downloadURL) // /cni-plugins/v*/binaries
+	if err != nil {
+		return "", fmt.Errorf("download file image URL is not in the expected format: %s", downloadURL)
+	}
+	urlSplit := strings.Split(url.Path, "/") // ["", cni-plugins, v*, binaries]
+	return urlSplit[1], nil
```
cameronmeissner, line 117: lets return an error here if there aren't at least 2 elements in the URL split, or if the segment we want is empty

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,136 @@ const (
...
+
+func processManifest(manifest Manifest) {
+	CachedFromManifest.Kubernetes = manifest.Kubernetes
+	CachedFromManifest.Runc = manifest.Runc
+	CachedFromManifest.Containerd = manifest.Containerd
+	CachedFromManifest.NvidiaContainerRuntime = manifest.NvidiaContainerRuntime
+	CachedFromManifest.NvidiaDrivers = manifest.NvidiaDrivers
+}
+
+func processComponents(components Components) {
```
cameronmeissner, line 55: I'd prefer this func return an error and then we panic on said errors directly within the `init()`

pkg/agent/bakerapi.go
```diff
@@ -174,3 +175,17 @@ func findSIGImageConfig(sigConfig datamodel.SIGAzureEnvironmentSpecConfig, distr
...
 }
+
+func (agentBaker *agentBakerImpl) GetCachedVersionsOnVHD() (datamodel.CachedOnVHD, error) {
+	cached := datamodel.CachedOnVHD{
+		CachedFromManifest:                 datamodel.CachedFromManifest,
+		CachedFromComponentContainerImages: datamodel.CachedFromComponentContainerImages,
+		CachedFromComponentDownloadedFiles: datamodel.CachedFromComponentDownloadedFiles,
+	}
+
+	if datamodel.CachedFromManifest == (&datamodel.Manifest{}) || datamodel.CachedFromComponentContainerImages == nil || datamodel.CachedFromComponentDownloadedFiles == nil {
```
cameronmeissner, line 186: since you use `make` to construct `CachedFromComponentContainerImages` and `CachedFromComponentDownloadFiles`, they would never be nil

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,136 @@ const (
...
+	return getCachedVersionsFromManifestJSON(path.Join(path.Dir(filename), manifestFilePath))
+}
+
+func CacheComponents() Components {
+	_, filename, _, _ := runtime.Caller(0)
+	componentsFilePath := "../../../vhdbuilder/packer/components.json"
+	return getCachedVersionsFromComponentsJSON(path.Join(path.Dir(filename), componentsFilePath))
+}
+
+func processManifest(manifest Manifest) {
```
cameronmeissner, line 47: also maybe rename to `initializeManifest`/`processComponents`?

pkg/agent/bakerapi.go
cameronmeissner: also probably better to initialize CachedFromManifest as nil since it's already a ptr type, and then simply check against that here instead of doing this. You could probably also do the something for the other two:

e.g.


```
if datamodel.CachedFromManifest == nil || datamodel.CachedFromComponentContainerImages == nil || datamodel.CachedFromComponentDownloadedFiles == nil || {
...
}

var (
	CachedFromComponentContainerImages = map[string]ContainerImage
	CachedFromComponentDownloadedFiles = map[string]DownloadFiles
	CachedFromManifest * Manifest
)

...
```

then initialize them directly within `processComponents` / `processManifest`

pkg/agent/datamodel/helper.go
```diff
@@ -93,3 +94,25 @@ func IndentString(original string, spaces int) string {
 	}
 	return out.String()
 }
+
+func processContainerImage(downloadURL string) (string, error) {
```
cameronmeissner, line 98: can we also add explicit unit tests for these?

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,136 @@ const (
...
+//nolint:gochecknoinits
+func init() {
+	manifest := CacheManifest()
+	processManifest(manifest)
+
+	components := CacheComponents()
+	processComponents(components)
+}
+
+func CacheManifest() Manifest {
```
cameronmeissner, line 35: nit `GetManifest` / `GetComponents`?

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,136 @@ const (
...
+
+type Manifest struct {
+	Containerd             Dependency `json:"containerd"`
+	Runc                   Dependency `json:"runc"`
+	NvidiaContainerRuntime Dependency `json:"nvidia-container-runtime"`
+	NvidiaDrivers          Dependency `json:"nvidia-drivers"`
+	Kubernetes             Dependency `json:"kubernetes"`
+}
+
+type Versions struct {
```
cameronmeissner, line 121: can we also move all these new types to `types.go`

pkg/agent/datamodel/sig_config.go
```diff
@@ -12,6 +16,136 @@ const (
...
+	}
+	var components Components
+	if err = json.Unmarshal(data, &components); err != nil {
+		panic(err)
+	}
+	return components
+}
+
+type CachedOnVHD struct {
+	CachedFromManifest                 *Manifest                 `json:"cached_from_manifest"`
```
cameronmeissner, line 98: `json:"cachedFromManifest"`
`json:"cachedFromComponentContainerImages"`
`json:"cachedFromComponentDownloadedFiles"`

pkg/agent/bakerapi.go
```diff
@@ -16,6 +16,7 @@ type AgentBaker interface {
 	GetNodeBootstrapping(ctx context.Context, config *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error)
 	GetLatestSigImageConfig(sigConfig datamodel.SIGConfig, distro datamodel.Distro, envInfo *datamodel.EnvironmentInfo) (*datamodel.SigImageConfig, error)
 	GetDistroSigImageConfig(sigConfig datamodel.SIGConfig, envInfo *datamodel.EnvironmentInfo) (map[datamodel.Distro]datamodel.SigImageConfig, error)
+	GetCachedVersionsOnVHD() (datamodel.CachedOnVHD, error)
```
cameronmeissner, line 19: maybe consider returning a ptr type here to be consistent with the rest of the API

cameronmeissner (approved):

PR #4332, opened 2024-04-23
perf: add async rg delete to packer template

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR adds the `async_resourcegroup_delete` boolean to the packer build templates. This change results in the packer tool asynchronously deleting the temporary packer resource group instead of waiting to confirm it is deleted at the end of the builds. 

The clean up packer generated resources step of the VHD builds already has a fail-safe in place to ensure temporary packer resource group deletion for each VHD. It always results in ResourceGroupNotFound because the packer tool ensures its deletion before this step in pipeline has been reached. By using this boolean, we can make use of that fail-safe and save roughly 15 seconds on build times if the process executes perfectly, and possibly more time if packer has issues confirming resource group deletion due to network or other issues.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
+++ b/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
@@ -57,7 +57,8 @@
       "skip_create_image": "true",
       "managed_image_resource_group_name": "{{user `resource_group_name`}}",
       "managed_image_name": "{{user `sig_image_name`}}-{{user `captured_sig_version`}}",
-      "user_assigned_managed_identities": "{{user `linux_msi_resource_ids`}}"
+      "user_assigned_managed_identities": "{{user `linux_msi_resource_ids`}}",
+      "async_resourcegroup_delete": "true"
     }
   ],
   "provisioners": [
--- a/vhdbuilder/packer/vhd-image-builder-base.json
+++ b/vhdbuilder/packer/vhd-image-builder-base.json
@@ -67,7 +67,8 @@
           "{{user `location`}}"
         ]
       },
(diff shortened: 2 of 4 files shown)


ganeshkumarashok:
Packer doc: https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm


async_resourcegroup_delete (bool) - If you want packer to delete the temporary resource group asynchronously set this value. It's a boolean value and defaults to false. **Important Setting this true means that your builds are faster, however any failed deletes are not reported.**

There's still a deletion check in the "clean up packer resources step right"?  Does enabling `aysnc_resource_group_delete` prevent those failed delete logs from showing up (curious based on that note in the docs)? If so, it's not worth enabling this for ~15s time reduction

zachary-bailey:
> Packer doc: https://developer.hashicorp.com/packer/integrations/hashicorp/azure/latest/components/builder/arm
> 
> async_resourcegroup_delete (bool) - If you want packer to delete the temporary resource group asynchronously set this value. It's a boolean value and defaults to false. **Important Setting this true means that your builds are faster, however any failed deletes are not reported.**
> 
> There's still a deletion check in the "clean up packer resources step right"? Does enabling `aysnc_resource_group_delete` prevent those failed delete logs from showing up (curious based on that note in the docs)? If so, it's not worth enabling this for ~15s time reduction

Yes, there is a deletion check in Clean Up Packer Generated Resources. 

I am not familiar with the failed delete logs. Do you mean in Azure in the azure portal, or a log on the VHD?

zachary-bailey:
Rejected due to possibility of losing deletion logs

PR #4335, opened 2024-04-23
Feat: Move helper function within the nbcontract package. Add validators to check contract payload.

Devinwong:
**What type of PR is this?**
/kind feature

Move configs to the top of the proto files for consistency.
Move helper function within the nbcontract package.
Add validators to check contract payload.

To see how a client will use it, please see this PR. https://github.com/Devinwong/karpenter-provider-azure/pull/11/files

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/helper.go
lilypan26, line 1: haha i moved these files over in a separate branch as well. I think a better place might be to keep these in the `self-contained` folder that Anuj made a while back. I am also thinking of moving the code for the go binary there as well. And also think it makes more sense to keep this folder dedicated to just the contract. What do you think?

pkg/proto/nbcontract/v1/clusterconfig.proto
```diff
@@ -49,6 +50,34 @@ message ClusterNetworkConfig {
 
   // Route table name
   string route_table = 5;
+
```
lilypan26, line 53: Although these are networking related, I separated the fields above because they are exclusively referenced in the `azure.json` file. This entire `ClusterConfig` contains fields that are only referenced in `azure.json`. So, in the future if we want to go with Alex's idea of separating the `azure.json` fields into its own JSON blob, i think it might be best to just keep the following fields in a separate `NetworkConfig`.

Devinwong: I see I see. wasn't aware of this. Then I agree with you. I will move them back. Thanks

lilypan26 (approved):

PR #4341, opened 2024-04-24
fix: don't collect network logs within aks-log-collector

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/aks-log-collector.sh
```diff
@@ -99,62 +99,6 @@ command -v crictl >/dev/null && {
...
-}
-command -v iptables >/dev/null && iptables -L -vn --line-numbers >collect/iptables.txt 2>&1
-command -v ip6tables >/dev/null && ip6tables -L -vn --line-numbers >collect/ip6tables.txt 2>&1
-command -v nft >/dev/null && nft -jn list ruleset >collect/nftables.json 2>&1
-command -v ss >/dev/null && {
-  ss -anoempiO --cgroup >collect/ss.txt 2>&1
-  ss -s >>collect/ss.txt 2>&1
-}
-
-# Collect network information from network namespaces
```
hunter32292, line 129: I only think you need to remove the ip_netns_commands collection, just the stuff using exec

cameronmeissner: discussed offline, the other bits are being removed due to OOM issue observed in customer environments as a result of the `nft` calls

hunter32292 (approved):

cameronmeissner:
fixed in separate PR

PR #4352, opened 2024-04-25
self-contained: Add scripts for go binary and tests

lilypan26:
**What type of PR is this?**

Move helper files for cse generation using new contract from Karpenter to AB. Add tests taken from baker_test.go to work with new parser and adjust scripts based on diffs from expected data.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/proto/nbcontract/v1/kubeletconfig.proto
```diff
@@ -17,14 +17,20 @@ message KubeletConfig {
   // The type of disk to use for the kubelet.
   KubeletDisk kubelet_disk_type = 5;
 
+  // Whether kubelet config should be enabled.
+  bool enable_kubelet_config_file = 6;
```
lilypan26, line 21: adding to contract because it cannot simply be determined by whether `kubelet_config_file_content` is specified. In AB, kubelet config file content is always populated with some default values, but that doesn't mean we want to use it. In RP, it is determined by a toggle that is defaulted to false. Also depends on other variables such as whether `CustomKubeletConfig` is set.

pkg/proto/nbcontract/v1/kubeletconfig.proto
```diff
@@ -17,14 +17,20 @@ message KubeletConfig {
...
   // Kubelet client private key
-  string kubelet_client_key = 7;
+  string kubelet_client_key = 8;
 
   // The content of the kubelet client certificate file.
-  string kubelet_client_cert_content = 8;
+  string kubelet_client_cert_content = 9;
+
+  // The path used to mount docker images, emptyDir volumes, and kubelet data.
+  string container_data_dir = 10;
```
lilypan26, line 33: Adding because it is used in containerd template and is needed when kubeletDiskType is temporary. Karpenter probably used it because they do not support this. But adding to keep consistency with AB.

pkg/proto/nbcontract/v1/kubebinaryconfig.proto
```diff
@@ -11,4 +11,10 @@ message KubeBinaryConfig {
    
   //privately cached kube binary url
   string private_kube_binary_url = 3;  
+  
+  //full path to the "pause" image. Used for --pod-infra-container-image.
+  string pod_infra_container_image_url = 4;
+
+  // Full path to the Linux credential provider (tar.gz) to use.
+  string linux_credential_provider_url = 5;
```
lilypan26, line 19: Addded recently to cse_cmd, Karpenter was out of date. Needed in contract.

pkg/proto/nbcontract/v1/kubebinaryconfig.proto
```diff
@@ -11,4 +11,10 @@ message KubeBinaryConfig {
    
   //privately cached kube binary url
   string private_kube_binary_url = 3;  
+
+  //full path to the "pause" image. Used for --pod-infra-container-image.
+  string pod_infra_container_image_url = 4;
```
lilypan26, line 16: used in containerd template. Karpenter hardcodes this, but could break things with RP, so we should add this to contract.

pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -1,15 +1,152 @@
...
+KUBELET_CONFIG_FILE_CONTENT="{{.KubeletConfig.GetKubeletConfigFileContent}}"
+SWAP_FILE_SIZE_MB="{{.CustomLinuxOsConfig.GetSwapFileSize}}"
+GPU_DRIVER_VERSION="{{getGpuDriverVersion .VmSize}}"
+GPU_IMAGE_SHA="{{getGpuImageSha .VmSize}}"
+GPU_INSTANCE_PROFILE="{{.GpuConfig.GetGpuInstanceProfile}}"
+CUSTOM_SEARCH_DOMAIN_NAME="{{.CustomSearchDomainConfig.GetDomainName}}"
+CUSTOM_SEARCH_REALM_USER="{{.CustomSearchDomainConfig.GetRealmUser}}"
+CUSTOM_SEARCH_REALM_PASSWORD="{{.CustomSearchDomainConfig.GetRealmPassword}}"
+MESSAGE_OF_THE_DAY="{{.GetMessageOfTheDay}}"
+HAS_KUBELET_DISK_TYPE="{{getHasKubeletDiskType .KubeletConfig}}"
```
lilypan26, line 138: was hardcoded to false, made it set based on config

pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -1,15 +1,152 @@
-ADMINUSER="{{.LinuxAdminUsername}}"
-PROVISION_OUTPUT="{{.ProvisionOutput}}"
-{{if .CustomCloudConfig.IsCustomCloud}}
-REPO_DEPOT_ENDPOINT="{{.RepoDepotEndpoint}}"
-{{/* need to properly add {{GetInitAKSCustomCloudFilepath}} >> /var/log/azure/cluster-provision.log 2>&1; */}}
+PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log";
+echo $(date),$(hostname) > ${PROVISION_OUTPUT};
+{{if true}}
```
lilypan26, line 3: this used to be `{{if ShouldEnableCustomData}}` but we'll be getting rid of this with self-contained.

pkg/parser/consts.go
```diff
@@ -0,0 +1,49 @@
...
+	customSearchDomainsCSEScriptFilepath = "/opt/azure/containers/setup-custom-search-domains.sh"
+	dhcpV6ServiceCSEScriptFilepath       = "/etc/systemd/system/dhcpv6.service"
+	dhcpV6ConfigCSEScriptFilepath        = "/opt/azure/containers/enable-dhcpv6.sh"
+	initAKSCustomCloudFilepath           = "/opt/azure/containers/init-aks-custom-cloud.sh"
+)
+
+const numInPair = 2
+
+const (
+	vmTypeStandard       = "standard"
```
Devinwong, line 35: Should we export these consts so that client can use them?

pkg/parser/parser.go
```diff
@@ -27,29 +28,23 @@ func executeBootstrapTemplate(inputContract *nbcontractv1.Configuration) (string
...
-
-	inputContract := nbcontractv1.Configuration{}
-	err = json.Unmarshal(inputJSON, &inputContract)
+func Parse(inputJSON []byte) string {
+	// Parse the JSON into a Person struct
+	var nbc nbcontractv1.Configuration
+	err := json.Unmarshal(inputJSON, &nbc)
 	if err != nil {
 		log.Printf("Failed to unmarshal the json to nbcontractv1: %v", err)
+		panic(err)
```
Devinwong, line 37: Should we not panic because the program will crash immediately? I think we can return error instead so that the caller can catch this error. Thoughts?
There is a post discussing panic vs error. https://stackoverflow.com/questions/44504354/should-i-use-panic-or-return-error

pkg/parser/cse_cmd.sh.gtpl
Devinwong: I am not very sure about this. Needs to check with Anuj. I just followed the same logic in the official karpenter. https://github.com/Azure/karpenter-provider-azure/blob/138a91aaf5cf450303b69c1d26b13ee30810d2ce/pkg/providers/imagefamily/bootstrap/cse_cmd.sh.gtpl#L12C1-L12C25

pkg/parser/utils.go
```diff
@@ -0,0 +1,281 @@
...
+	"strings"
+
+	"github.com/Azure/agentbaker/pkg/agent"
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	nbcontractv1 "github.com/Azure/agentbaker/pkg/proto/nbcontract/v1"
+	"github.com/blang/semver"
+)
+
+// GetLoadBalancerSKI returns the LoadBalancerSku enum based on the input string.
+func GetLoadBalancerSKU(sku string) nbcontractv1.LoadBalancerConfig_LoadBalancerSku {
```
Devinwong, line 34: Should we host this function in AgentBaker and export it so that client can use it? For example, Karpenter will use it to populate the nbconfig object.

pkg/parser/utils.go
```diff
@@ -0,0 +1,281 @@
...
+		return nbcontractv1.NetworkPlugin_NP_AZURE
+	} else if strings.EqualFold(networkPlugin, "kubenet") {
+		return nbcontractv1.NetworkPlugin_NP_KUBENET
+	}
+
+	return nbcontractv1.NetworkPlugin_NP_NONE
+}
+
+// GetNetworkPolicyType returns the NetworkPolicyType enum based on the input string.
+func GetNetworkPolicyType(networkPolicy string) nbcontractv1.NetworkPolicy {
```
Devinwong, line 56: Move to AB and export it?

pkg/parser/utils.go
```diff
@@ -0,0 +1,281 @@
...
+		return nbcontractv1.LoadBalancerConfig_STANDARD
+	} else if strings.EqualFold(sku, "Basic") {
+		return nbcontractv1.LoadBalancerConfig_BASIC
+	}
+
+	return nbcontractv1.LoadBalancerConfig_UNSPECIFIED
+}
+
+// GetNetworkPluginType returns the NetworkPluginType enum based on the input string.
+func GetNetworkPluginType(networkPlugin string) nbcontractv1.NetworkPlugin {
```
Devinwong, line 45: Move to AB and export it?

pkg/parser/utils.go
lilypan26: This is in agentbaker... and it is exported. But we can move it to a different package if that's what you're suggesting? I also wasn't sure where best to put it. I was thinking maybe the directory anuj created called `selfcontained`.

pkg/proto/nbcontract/v1/customcloudconfig.proto
```diff
@@ -2,18 +2,15 @@ syntax = "proto3";
 package nbcontract.v1;
 
 message CustomCloudConfig {
-  // IsAKSCustomCloud is a flag to indicate if it is AKS custom cloud
-  bool is_aks_custom_cloud = 1;
-
-  // InitFilePath is the path to the file that contains the init script
-  string init_file_path = 2;
+  // Name of custom cloud
+  string custom_cloud_env_name = 1;
```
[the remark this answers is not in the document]
Devinwong: is_aks_custom_cloud is quite messy per my shallow understand.
I see this in the cse_cmd
``` {{/* both CLOUD and ENVIRONMENT have special values when IsAKSCustomCloud == true */}}
{{/* CLOUD uses AzureStackCloud and seems to be used by kubelet, k8s cloud provider */}}
{{/* target environment seems to go to ARM SDK config */}}
{{/* not sure why separate/inconsistent? */}}
{{/* see GetCustomEnvironmentJSON for more weirdness. */}}
TARGET_CLOUD="{{- if IsAKSCustomCloud -}} AzureStackCloud {{- else -}} {{GetTargetEnvironment}} {{- end -}}"
TARGET_ENVIRONMENT="{{GetTargetEnvironment}}"
``` 

And in /aks-rp/protos/hcp/types/managedcluster/containerservice/v1/properties_methods.go, it also depends on p.CustomCloudEnv != nil. So seems not be able to compute from single custom_cloud_env_name?
I could be wrong though, because the logic here crosses over with others.
```// IsAKSCustomCloud checks if it's in AKS custom cloud
func (p *Properties) IsAKSCustomCloud() bool {
	return p.CustomCloudEnv != nil &&
		strings.EqualFold(p.CustomCloudEnv.Name, AKSCustomCloudName)
}
```

pkg/proto/nbcontract/v1/config.proto
```diff
@@ -94,7 +94,7 @@ message Configuration {
   optional bool is_vhd = 22;
 
   // Specifies whether SSH is enabled or disabled on the VM node
-  bool enable_ssh = 23;
+  optional bool enable_ssh = 23;
```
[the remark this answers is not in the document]
Devinwong: Great. Can you leave some comments in the code so that later on we know why it's optional? One example I made was the above one `is_vhd`.

pkg/proto/nbcontract/v1/authconfig.proto
```diff
@@ -3,24 +3,21 @@ package nbcontract.v1;
 
 // Auth Config fields stored in azure.json used by cloud-provider-azure
 message AuthConfig {
-  // Target cloud. Set to cloud. Default to AzurePublicCloud
-  string target_cloud = 1;
```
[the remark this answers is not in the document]
Devinwong: Again, this logic of this variable is messy along with the IsAKSCustomCloud. I didn't have confidence to remove this so I kept it here. Or we can double check with Anuj? (he may not know either)

pkg/parser/utils.go
Devinwong: It's in the parser folder so I thought this will be in the parser go binary. pkg/proto/nbcontract/v1/helper.go should be good because client just needs to import single package to have both nbcontract and corresponding helpers.

lilypan26: ah okay agreed, moved.

pkg/proto/nbcontract/v1/customcloudconfig.proto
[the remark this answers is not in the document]
lilypan26: I think they are just checking `p.CustomCloudEnv != nil`, because they are accessing `p.CustomCloudEnv.Name` in the next statement to avoid nil pointer dereference. There are no other checks on other fields in the config, so I think it is purely determined by the name.

pkg/parser/cse_cmd.sh.gtpl
lilypan26: The line you linked to is `{{if .IsAKSCustomCloud}}` which is different than `{{if ShouldEnableCustomData}}`. Karpenter doesn't have this line because they don't rely on AB's customdata scripts.

pkg/parser/parser.go
lilypan26: Agreed, return error instead.

pkg/proto/nbcontract/v1/config.proto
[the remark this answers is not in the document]
lilypan26: Added comment.

pkg/proto/nbcontract/v1/authconfig.proto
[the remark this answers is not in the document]
lilypan26: The logic I added is the same as that currently in AB and we know for sure that it works so I think we should try to stay as close to that as possible.

pkg/parser/consts.go
lilypan26: agreed, moved consts that might be needed outside of go binary to contract package

pkg/parser/cse_cmd.sh.gtpl
Devinwong: Ah I see. Ya it seems to me it's always true but I can't conclude it either.

pkg/proto/nbcontract/v1/customcloudconfig.proto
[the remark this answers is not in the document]
Devinwong: makes sense to me :)

Devinwong (approved):
Looks good to me

pkg/parser/containerd.toml.gtpl
```diff
@@ -0,0 +1,118 @@
+version = 2
```
anujmaheshwari1, line 1: what about the containerd config with no GPU?

pkg/proto/nbcontract/v1/utils.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return NetworkPolicy_NPO_NONE
+}
+
+// GetDefaultOutboundCommand returns a default outbound traffic command.
+func GetDefaultOutboundCommand() string {
+	return "curl -v --insecure --proxy-insecure https://mcr.microsoft.com/v2/"
+}
+
+func GetKubeletNodeLabels(agentPool *datamodel.AgentPoolProfile) map[string]string {
```
Devinwong, line 72: Just to confirm, this function is not really used in the current PR, right? Is it supposed to be used by AKS-RP later?

pkg/proto/nbcontract/v1/utils.go
```diff
@@ -0,0 +1,289 @@
...
+	} else {
+		connectivityCheckCommand = `nc -vz ` + registry + ` 443`
+	}
+
+	return connectivityCheckCommand
+}
+
+// GetOrderedKubeletConfigFlagString returns an ordered string of key/val pairs.
+// copied from AKS-Engine and filter out flags that already translated to config file.
+func GetKubeletConfigFlag(k map[string]string, cs *datamodel.ContainerService, profile *datamodel.AgentPoolProfile,
```
Devinwong, line 128: Just to confirm, this function is not really used in the current PR, right? Is it supposed to be used by AKS-RP later?

pkg/proto/nbcontract/v1/utils.go
```diff
@@ -0,0 +1,289 @@
...
+		return nil
+	}
+	// empty config is treated as nil.
+	if len(kubeletConfigurations.Config) == 0 {
+		return nil
+	}
+	return kubeletConfigurations.Config
+}
+
+func ValidateAndSetLinuxKubeletFlags(kubeletFlags map[string]string, cs *datamodel.ContainerService, profile *datamodel.AgentPoolProfile) {
```
Devinwong, line 192: Just to confirm, this function is not really used in the current PR, right? Is it supposed to be used by AKS-RP later?

pkg/parser/containerd.toml.gtpl
lilypan26: since the plan is to deprecate the skip-gpu-driver-install tag soon now that the api field is available, i don't think it is worth adding the same hacky way of doing it now since the api field will mostly likely be GA by the time RP is using self-contained anyway. But i can add it back if you think its best to keep consistency for now.

pkg/proto/nbcontract/v1/utils.go
lilypan26: see above

pkg/proto/nbcontract/v1/utils.go
lilypan26: see above.

pkg/parser/containerd.toml.gtpl
lilypan26: Oh i realize i need to update the `if` condition in the config though to work the skip gpu driver if we're going to get rid of the other config though. Just updated.

pkg/parser/cmd/main.go
```diff
@@ -1,12 +1,38 @@
...
+		log.Fatal(err)
+		return
+	}
+
+	cseCmd, err := parser.Parse(inputJSON)
+	if err != nil {
+		log.Fatal(err)
+		return
+	}
+	if err = os.WriteFile("cse_cmd.sh", []byte(cseCmd), 0600); err != nil {
```
anujmaheshwari1, line 34: nit: is 600 enough to give permission to execute the file?

pkg/parser/consts.go
```diff
@@ -0,0 +1,36 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package parser
+
+// cloud-init destination file references.
+const (
+	cseHelpersScriptFilepath             = "/opt/azure/containers/provision_source.sh"
```
anujmaheshwari1, line 21: we will need to talk about these in depth, fine for now, but we will likely have a different version of the scripts, that adhere more to an updated cse_cmd.sh env vars.

pkg/parser/cse_cmd.sh.gtpl
anujmaheshwari1: let us keep "ShouldEnableCustomData" for now, we will sync on it. we will need it for the transition period of going from non self to self

pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -1,15 +1,152 @@
...
+PRIMARY_AVAILABILITY_SET={{.ClusterConfig.GetPrimaryAvailabilitySet}}
+PRIMARY_SCALE_SET={{.ClusterConfig.GetPrimaryScaleSet}}
+SERVICE_PRINCIPAL_CLIENT_ID={{.AuthConfig.GetServicePrincipalId}}
+NETWORK_PLUGIN={{getStringFromNetworkPluginType .GetNetworkConfig.GetNetworkPlugin}}
+NETWORK_POLICY="{{getStringFromNetworkPolicyType .GetNetworkConfig.GetNetworkPolicy}}"
+VNET_CNI_PLUGINS_URL={{.GetNetworkConfig.GetVnetCniPluginsUrl}}
+CNI_PLUGINS_URL={{.GetNetworkConfig.GetCniPluginsUrl}}
+CLOUDPROVIDER_BACKOFF=true
+CLOUDPROVIDER_BACKOFF_MODE=v2
+CLOUDPROVIDER_BACKOFF_RETRIES=6
```
anujmaheshwari1, line 39: why are these hardcoded here?

lilypan26: we could add a bunch of getters to populate this instead... but i think maybe we should first decide how we want to handle the `azure.json` config from that discussion earlier since all of these fields reside inside that config. Once we decide that we should be able to remove these altogether.

pkg/parser/cse_cmd.sh.gtpl
lilypan26: Okay, added to contract.

pkg/parser/cse_cmd.sh.gtpl
Devinwong: If we want to remove these from the cse_cmd.sh, we can put them somewhere in the boot scripts. But so far we haven't made any changes to the boot scripts yet. Also agreed with Lily that we need to first decide about the azure.json.

pkg/parser/consts.go
Devinwong: Sounds good.
For now, our getter just returns this const. We can return different file paths based on other conditions if needed.
``` 
func getCSEHelpersFilepath() string {
	return cseHelpersScriptFilepath
}
```

anujmaheshwari1 (approved):

PR #4354, opened 2024-04-26
test: modify benchmarking workflow and publish build performance artifact

zachary-bailey:
What type of PR is this?

/test

What this PR does / why we need it:

In addition to publishing a pipeline artifact that contains build performance data, this PR changes the workflow for benchmarking the build process. Changes include removing the `start_watch` function, adding the `installJq` function, and outputting benchmarking data in JSON instead of formatted strings. At the end of each script, a JSON object will be output in the terminal with performance data for each section.

The `installJq` function was necessary because jq is not installed until the `installDeps` function is called in install-dependencies.sh, but it is needed when the build process starts in pre-install-dependencies.sh.

The pipeline artifact that will be published is a JSON file that will contain benchmarking data for the build process. In later changes, github actions will be used to evaluate this artifact and check for build time regression.

The formatting of the artifact can be checked by going to the PR gate build and checking the artifacts list.

Which issue(s) this PR fixes:

This will prevent vhd build times from incrementally increasing due to code changes.

**Note: When reviewing "Files changed", please collapse /pkg/agent/testdata to see all relevant file changes**

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests


zachary-bailey:
closing PR, new PR will be created

PR #4358, opened 2024-04-30
Execute ipv6 nftables before kubelet

camrynl:
**What type of PR is this?**
/kind design

This will allow nftables to execute bfore kubelet, which is needed for cilium dual stack clusters. In substitution for these nftable rules, we will attach an ebpf program to cilium clusters that will do the IP conversion instead. Cilium clusters will have an init container which will remove the nftable rules and start the bpf program. This change will make sure nftable rules are executed before init container.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
execute ipv6 nftable rules before kubelet
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/ipv6_nftables.service
+++ b/parts/linux/cloud-init/artifacts/ipv6_nftables.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Configure nftables rules for handling Azure SLB IPv6 health probe packets
+Before=kubelet.service
 
 [Service]
 Type=simple


tamilmani1989 (approved):

tamilmani1989:
cc @phealy @tyler-lloyd @wedaly

tyler-lloyd:
LGTM

tyler-lloyd (approved):

parts/linux/cloud-init/artifacts/ipv6_nftables.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=Configure nftables rules for handling Azure SLB IPv6 health probe packets
+Before=kubelet.service
```
wedaly, line 3: is there any scenario in which `ipv6_nftables.sh` could fail and prevent kubelet from starting? From a quick look at the script, it seems safe enough (just curl to IMDS, jq, then running `nft` to install the rules)

wedaly (approved):

cameronmeissner (approved):

PR #4362, opened 2024-05-02
Feat: update semver check. Add TestGetOutBoundCmd and contract compatibility tests.

Devinwong:
**What type of PR is this?**
/kind feature

1.  Update semver check. 
2.  Add TestGetOutBoundCmd and partial contract compatibility tests.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/parser/parser_test.go
```diff
@@ -1,4 +1,4 @@
-package parser_test
+package parser
```
Devinwong, line 1: Renamed the package name so that I can use the constants defined in package `parse`

pkg/parser/parser_test.go
```diff
@@ -170,17 +177,10 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
...
 
-		if generateTestData() {
-			if _, err = os.Stat(fmt.Sprintf("./testdata/%s", folder)); os.IsNotExist(err) {
-				e := os.MkdirAll(fmt.Sprintf("./testdata/%s", folder), 0755)
-				Expect(e).To(BeNil())
-			}
-			err = os.WriteFile(fmt.Sprintf("./testdata/%s/generatedCSECommand", folder), []byte(cseCmd), 0644)
-			Expect(err).To(BeNil())
-		}
+		generateTestDataIfRequested(folder, cseCmd)
```
Devinwong, line 183: @lilypan26  Refactored a bit your function so that I can re-use it in my test. 😊

pkg/parser/parser_test.go
Devinwong: changed package name back to parser_test as it's ginkgo's convention

lilypan26 (approved):

pkg/parser/consts.go
```diff
@@ -30,7 +30,15 @@ const (
 )
 
 const (
-	maxLBRuleCountDefault = 148
-	ipLocalReservedPorts  = 65330
-	MinArgs               = 2
+	maxLBRuleCountDefault               = 148
```
anujmaheshwari1, line 33: these are the sysctl values right?

pkg/parser/cse_cmd.sh.gtpl
```diff
@@ -90,7 +90,7 @@ OUTBOUND_COMMAND="{{.GetOutboundCommand}}"
 ENABLE_UNATTENDED_UPGRADES="{{.GetEnableUnattendedUpgrade}}"
 ENSURE_NO_DUPE_PROMISCUOUS_BRIDGE="{{getEnsureNoDupePromiscuousBridge .GetNetworkConfig}}"
 SHOULD_CONFIG_SWAP_FILE="{{getEnableSwapConfig .CustomLinuxOsConfig}}"
-SHOULD_CONFIG_TRANSPARENT_HUGE_PAGE="{{getShouldCOnfigTransparentHugePage .CustomLinuxOsConfig}}"
+SHOULD_CONFIG_TRANSPARENT_HUGE_PAGE="{{getShouldConfigTransparentHugePage .CustomLinuxOsConfig}}"
```
anujmaheshwari1, line 93: nice catch, thanks!

pkg/proto/nbcontract/v1/helper.go
```diff
@@ -110,33 +110,27 @@ func (nBCB *NBContractBuilder) ValidateNBContract() error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("invalid contract version from contract payload: %s. It should be a semantic version", nBCB.nodeBootstrapConfig.Version)
 	}
 
-	expectedMinor, err := nBCB.parseMinor(contractVersion)
-	if err != nil {
-		return err
+	if payloadContractVer.Major != expectedContractVer.Major {
+		return fmt.Errorf("contract major versions mismatch. Expecting %v, but got %v. Please update AgentBaker to the latest version",
```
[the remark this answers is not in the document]
Devinwong: the lint complained about this so I changed to lower case 😂
```
pkg/proto/nbcontract/v1/helper.go:125:10: ST1005: error strings should not be capitalized (stylecheck)
                return fmt.Errorf("Contract major versions mismatch. Expecting %v, but got %v. Please update AgentBaker to the latest version",

File: pkg/parser/consts.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (Devinwong) [reply pkg/parser/consts.go:33]: fixed

File: pkg/parser/parser_test.go [hunk 6]
```diff
@@ -328,6 +318,115 @@ oom_score = 0
 	)
 })
 
+var _ = Describe("Test NBContract compatibility from Json to CSE command", func() {
+	DescribeTable("for test case", func(folder string, validator outputValidator) {
+		nBCB := nbcontractv1.NewNBContractBuilder()
```
anujmaheshwari1, line 323: nit: not too fond of the naming, kinda confusing, but can revisit

pkg/proto/nbcontract/v1/utils_test.go
```diff
@@ -176,3 +176,98 @@ func Test_getKubeletNodeLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestGetOutBoundCmd(t *testing.T) {
```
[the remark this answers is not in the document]
Devinwong: Which one were you comparing with? This? https://github.com/Azure/AgentBaker/blob/e1d423e9731b222f30a1b469c170db58b88c5530/pkg/agent/variables.go#L177
It seems to me they are the same?

pkg/parser/consts.go
Devinwong: `maxLBRuleCountDefault` is for MAXIMUM_LOADBALANCER_RULE_COUNT and the others `default` are for sysctl

pkg/parser/parser_test.go
Devinwong: With this, the description in junit.xml looks smooth to me but definitely can revisit for a better one.
An example in junit.xml
`Test NBContract compatibility from Json to CSE command for test case with empty config. Parser Should provide default values to unset fields.`

pkg/proto/nbcontract/v1/helper.go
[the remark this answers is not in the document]
Devinwong: I guess it's because the error message format looks like "error: message". So lower case makes sense too.

pkg/proto/nbcontract/v1/utils_test.go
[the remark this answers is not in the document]
lilypan26: Oh I think it used to be different in an older version: https://github.com/Devinwong/karpenter-provider-azure/blob/d731ed0dfc4855d7fac165ce506c3fa1678d87f1/pkg/providers/imagefamily/bootstrap/helpers.go#L69

I think it was because you were trying to see if this could be computed on the VHD instead, which I also looked into, but we would probably need to add more variables to the contract. So I just updated it to be the same as the current one in the AB and decided it probably needs to be passed in from the bootstrapper.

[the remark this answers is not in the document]
Devinwong: I see.

PR #4364, opened 2024-05-02
refactor: fix + refactor JSON file loading within new caching API

cameronmeissner:
**What type of PR is this?**

currently we try to load in the contents of both manifest/components JSON files as part of initialization to support the new VHD cache api. this doesn't seem to be working with the current implementation, thus this PR migrates this logic towards using the new `parts` package to read these files instead, introduced in this PR: https://github.com/Azure/AgentBaker/pull/4215. 

to accomplish this, we need to include `components.json` within the emedded FS rooted at `parts/` - as a result this PR moves `components.json` to `parts/linux/cloud-init/artifacts/components.json`, and adds a new symlink to the same within `vhdbuilder/packer`

also includes some light refactoring of the caching API itself

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


pkg/agent/vhd/cache/cache.go
```diff
@@ -0,0 +1,134 @@
...
+
+const (
+	manifestFilePartPath   = "linux/cloud-init/artifacts/manifest.json"
+	componentsFilePartPath = "linux/cloud-init/artifacts/components.json"
+)
+
+//nolint:gochecknoinits
+func init() {
+	if err := initVHDCacheContent(); err != nil {
+		panic(err)
```
Devinwong, line 23: Should we use panic or just returning error here? Does the panic occur during VHD baking? It won't crash during node provisioning, right?

cameronmeissner: we panic since this is in an `init()` - https://tutorialedge.net/golang/the-go-init-function/

Devinwong: Just to clarify a bit, who is importing/using this cache package?

cameronmeissner: the `agent` pkg is the primary consumer, which contains the bakerapi code

Devinwong (approved):
lgtm

cameronmeissner:
> lgtm since this mostly just moves the logic around, I wonder if there was a better way to actual create a symlink though, non blocking!

I think in a subsequent PR we can just completely remove the symlink altogether and just move components.json over to parts/ - that will entail changing a bunch of other references and I wanted to keep this PR on the smaller side if possible

lilypan26 (approved):

pkg/agent/vhd/cache/cache.go
```diff
@@ -0,0 +1,140 @@
...
+	manifestContent = bytes.ReplaceAll(manifestContent, []byte("#EOF"), []byte(""))
+	var manifest Manifest
+	if err = json.Unmarshal(manifestContent, &manifest); err != nil {
+		return nil, fmt.Errorf("unmarshalling manifest.json file part content: %w", err)
+	}
+	return &manifest, nil
+}
+
+func getComponents() (*Components, error) {
+	componentsContent, err := parts.Templates.ReadFile(componentsFilePartPath)
```
cameronmeissner, line 100: example

PR #4367, opened 2024-05-03
perf: enable shallow replication

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR changes the replication method for VHDs from full replication with the packer tool to shallow replication with a direct SIG API call. Currently, we are using the packer tool to fully replicate all non-arm64 VHDs to the SIG. For arm64 builds, a snapshot is created at the beginning of the build, updated at the end of the build, and then converted to the SIG. 

This PR uses the arm64 publishing logic that is already in place to publish all VHDs to the SIG using shallow replication, with the exception of the confidential VM because it cannot be made into a snapshot.

In order to accomplish this, all references to arm64_os_disk* were changed to encompass all VHDs, all existing packer templates were modified to remove the `shared_image_gallery_destination` block, a new packer template was created to publish the confidential VM using the `shared_image_gallery_destination` block, and the convert-os-disk-snapshot-to-sig.sh file was modified to be used with all VHDs by assigning the architecture property dynamically and enabling shallow replication. Additionally, the deis/go-dev container needed to be reverted to v1.38.0.

Further details regarding these changes can be found in the project presentation.

Unsatisfactory VHD Build Times.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -141,7 +141,7 @@ parameters:
 
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
+  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
   
 pool:
   name: $(POOL_NAME)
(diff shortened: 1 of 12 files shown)


zachary-bailey:
Obsolete PR.

PR #4381, opened 2024-05-09
Feat: Add trivy back without adding to VHD space constraints 

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/storage-scan.sh
```diff
@@ -0,0 +1,28 @@
...
+mkdir -p "$(dirname "${STORAGE_REPORT_PATH}")"
+
+echo "----" >> $STORAGE_REPORT_PATH
+{ df -h >> "$STORAGE_REPORT_PATH" ; } 2>&1
+echo "----" >> $STORAGE_REPORT_PATH
+
+CUR_DIR=$(pwd)
+
+mkdir -p /mnt/sdb1
+mount /dev/sdb1 /mnt/sdb1
```
cameronmeissner, line 15: I know this is draft but I just wanted to mention this since I've dealt with it before: you can't rely on a particular device having a particular label within the context of Azure VMs - you need to determine which device is actually the OS device at runtime: https://learn.microsoft.com/en-us/troubleshoot/azure/virtual-machines/linux/troubleshoot-device-names-problems#discover-filesystem-uuids-by-using-blkid

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
```diff
@@ -609,15 +609,46 @@
...
     {
       "type": "file",
       "direction": "download",
       "source": "/opt/azure/containers/trivy-images-table.txt",
       "destination": "trivy-images-table.txt"
     },
+    {
+      "type": "file",
+      "source": "vhdbuilder/packer/storage-scan.sh",
+      "destination": "/home/packer/storage-scan.sh"
```
cameronmeissner, line 624: could we extend `trivy-scan.sh` to include both the trivy + storage scanning logic and potentially run them in parallel?

that way we simplify to:
- shell (execute)
- file download
- shell (rm dangling reports)

PR #4386, opened 2024-05-09
refactor: updated STLS bootstrap integration with linux CSE

cameronmeissner:
**What type of PR is this?**
/kind feature

adds support for the new design of secure TLS bootstrapping to linux CSE/customdata.

follow-up items:
- add start-kubelet.sh to the VHD during builds
- download the secure TLS bootstrap client during VHD builds

Some important things to note:
- with this change, we now start kubelet without waiting for the startup to complete in the secure TLS bootstrapping case. this is due to the fact that CSE will not have access to MSI tokens via IMDS because of how RP assigns kubelet identities to both VMSS and standalone VMs, thus we want kubelet to delegate to the secure bootstrap client after CSE has exited with code 0, and eventually after RP has attached the kubelet identity to the VM. this also means that there will now be a longer delay between when CSE exits successfully and when the node is able to register with the API server due to the time it takes CSE success to propagate back to RP and for RP to perform the PATCH on the VMSS/SAVM to assign it the kubelet identity.
- this change _does not_ effect the current state of vanilla TLS bootstrapping and will continue functioning the exact same way as it does now when we're not using secure bootstrapping
- this change will have it such that vanilla TLS bootstrapping is used as a fall-back in case secure TLS bootstrapping fails to bootstrap a signed kubelet client certificate after a pre-configured timeout

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -479,7 +455,13 @@ contexts:
...
 current-context: bootstrap-context
 EOF
-    else
+    fi
+
+    if [ "${ENABLE_SECURE_TLS_BOOTSTRAPPING}" == "true" ]; then
+        configureKubeletSecureTLSBootstrap
+    fi
+
+    if [ "${ENABLE_SECURE_TLS_BOOTSTRAPPING}" == "false" ] && [ "${ENABLE_TLS_BOOTSTRAPPING}" == "false" ]; then
```
jason1028kr, line 464: Is it possible that only the first variable is set to true, and we don't create this kubeconfig file to fallback to?

jason1028kr: ah `it's the only option available if we can't have kubelet bootstrap its own credential (which is determined by RP).` Makes sense, thank you for explaining. Lmk when you need approval.

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -118,7 +118,7 @@ NO_PROXY_URLS="{{GetNoProxy}}"
 PROXY_VARS="{{GetProxyVariables}}"
 ENABLE_TLS_BOOTSTRAPPING="{{EnableTLSBootstrapping}}"
 ENABLE_SECURE_TLS_BOOTSTRAPPING="{{EnableSecureTLSBootstrapping}}"
-CUSTOM_SECURE_TLS_BOOTSTRAP_AAD_SERVER_APP_ID="{{GetCustomSecureTLSBootstrapAADServerAppID}}"
+CUSTOM_SECURE_TLS_BOOTSTRAP_AAD_RESOURCE="{{GetCustomSecureTLSBootstrapAADResource}}"
```
anujmaheshwari1, line 121: cse_cmd.sh var naming change, want to loop in @Bryce-Soghigian @tallaxes here for viz, however, I dont think this variable is being used yet that the renaming would break them?

still tagging them.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -293,6 +293,27 @@ configureCNIIPTables() {
     fi
 }
 
+configureKubeletSecureTLSBootstrap() {
+    # default AAD resource here so we can minimze bootstrap contract surface
+    AAD_RESOURCE="6dae42f8-4368-4678-94ff-3960e28e3630"
```
Devinwong, line 298: If user doesn't set CUSTOM_SECURE_TLS_BOOTSTRAP_AAD_RESOURCE, AAD_RESOURCE is defaulted to 6dae42f8-4368-4678-94ff-3960e28e3630. Does it mean this AAD_RESOURCE is always available in every user's subscription?

parts/linux/cloud-init/artifacts/cse_cmd.sh
cameronmeissner: thanks, keep forgetting

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: yes, as far as I know - this is the app ID of AKS's AAD server, I've spoken with Baichao and Weinong about this:

https://azure.github.io/kubelogin/concepts/aks.html

cameronmeissner: though will be a good idea to confirm this is correct gov environments as well

parts/linux/cloud-init/artifacts/cse_cmd.sh
tallaxes: Ack, and appreciated. Not used yet, so should be fine - but only as long as not setting it at all is tolerated by the rest of the logic.

Devinwong (approved):

PR #4395, opened 2024-05-13
feat: add datamodel field for Windows next-gen networking

wedaly:
**What type of PR is this?**
/kind feature

NextGenNetworkingURL is the per-cloud URL for downloading Windows eBPF assets.
Next-gen networking is enabled only when NextGenNetworkingURL is non-nil.

Fixes #

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -800,6 +800,13 @@ func (a *AgentPoolProfile) GetCustomLinuxOSConfig() *CustomLinuxOSConfig {
 	return a.CustomLinuxOSConfig
 }
 
+func (a *AgentPoolProfile) GetAgentPoolWindowsProfile() *AgentPoolWindowsProfile {
+	if a == nil {
+		return nil
+	}
+	return a.AgentPoolWindowsProfile
+}
+
 // Properties represents the AKS cluster definition.
 type Properties struct {
 	ClusterID               string
@@ -2189,6 +2196,9 @@ func (err *CSEStatusParsingError) Error() string {
 
 type AgentPoolWindowsProfile struct {
 	DisableOutboundNat *bool `json:"disableOutboundNat,omitempty"`
+
+	// Windows next-gen networking uses Windows eBPF for the networking dataplane.
+	NextGenNetworkingURL *string `json:"nextGenNetworkingURL,omitempty"`
 }
 
 // IsDisableWindowsOutboundNat returns true if the Windows agent pool disable OutboundNAT.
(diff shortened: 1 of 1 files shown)


cameronmeissner (approved):
@AbelHu wanna take a look as well?

pkg/agent/datamodel/types.go
```diff
@@ -2173,6 +2180,10 @@ func (err *CSEStatusParsingError) Error() string {
 
 type AgentPoolWindowsProfile struct {
 	DisableOutboundNat *bool `json:"disableOutboundNat,omitempty"`
+
+	// Windows next-gen networking uses Windows eBPF for the networking dataplane.
+	EnableNextGenNetworking  bool   `json:"enableNextGenNetworking,omitempty"`
+	NextGenNetworkingURLBase string `json:"nextGenNetworkingURLBase,omitempty"`
```
AbelHu, line 2186: Does `Base` mean that it will pass the package with the version in another property or parameter?

wedaly: I was planning to hard-code the version in the CSE scripts for now. This ensures that the version of the CSE installation script matches with a specific version of the package. If we need to hotfix something we can go through the regular hotfix process for AgentBaker service.

We can always add a package version field later if necessary, but I think to start this will be faster to iterate and easier for Windows eBPF team (no need to worry about version drift between pkg verison and install scripts).

AbelHu: It needs a long time to release a new AgentBaker service instead of toggles. My suggestion is
  - change the name from `NextGenNetworkingURLBase` to `NextGenNetworkingURL`
  - reference the code to set `CSEScriptsPackageUrl` in https://github.com/Azure/AgentBaker/blob/master/parts/windows/kuberneteswindowssetup.ps1#L244-L254 
With this design, we can use toggle to pass a full URL to hotfix the cve issues.

wedaly: thanks, that makes sense. Updated to `NextGenNetworkingURL`

pkg/agent/datamodel/types.go
```diff
@@ -2173,6 +2180,9 @@ func (err *CSEStatusParsingError) Error() string {
 
 type AgentPoolWindowsProfile struct {
 	DisableOutboundNat *bool `json:"disableOutboundNat,omitempty"`
+
+	// Windows next-gen networking uses Windows eBPF for the networking dataplane.
+	NextGenNetworkingURL string `json:"nextGenNetworkingURL,omitempty"`
```
wedaly, line 2185: Using `string` here to match the pattern for other components (`WindowsCalicoPackageURL`, `WindowsGmsaPackageUrl`, `GpuDriverURL`)

wedaly: switched back to *string

AbelHu (approved):
LGTM. Thanks for the PR.

PR #4396, opened 2024-05-13
self-contained: add go binary to vhd

lilypan26:
**What type of PR is this?**

Add go binary to VHD and github action to ensure go binary is rebuilt upon appropriate changes (e.g. contract changes, cse changes, etc.)

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/vhd-image-builder-base.json
```diff
@@ -87,6 +87,11 @@
         "sudo chown -R $USER /opt/certs"
       ]
     },
+    {
+      "type": "file",
+      "source": "pkg/parser/bin/parser",
```
cameronmeissner, line 92: so if I were you I'd probably have it such that we build the binary during the VHD build itself - that way we don't have to host binaries within the repository, and we can also maintain the validity checks it seems you want via the VHD content tests

cameronmeissner: I kinda sorta did something similar for the prefetch script generation: https://github.com/Azure/AgentBaker/pull/3971

lilypan26: Anuj said this might be difficult because we need access to goproxy to build the go binary. Is there an easy way to add goproxy here?

cameronmeissner: I wouldn't think we'd need to use the goproxy since the parser doesn't need internal deps as far as I can see - though if we did need it then we could simply utilize the devinfra PAT/goproxy static token at build-time, similarly to how I added the same for the auto-vendor pipeline the other day

cameronmeissner: makes me wonder though, why are we building the parser as a separate main package with the agentbaker module? 

https://github.com/Azure/AgentBaker/blob/master/pkg/parser/cmd/main.go


I'd probably prefer for the parser to be in its own module which specifies agentbaker/nbcontract as dependencies

anujmaheshwari1: we do have an internal dependency though, we import the contract in parser, so it needs to pull it.

anujmaheshwari1: even if it were its own module, we'd need to import the whole thing through packer and onto the VM to compile it during build time, which I am not sure if its the best thing to do.

or are you suggestion something else here?

anujmaheshwari1: I think I understand what Cameron means, however, we'd still need the PAT

cameronmeissner: the contract is hosted as an agentbaker package through right? agentbaker is OSS so shouldn't need the goproxy afaik

if it were an RP module/pkg then it would require the proxy is my understanding

anujmaheshwari1: agentbaker is OSS but it still comes under `github.com/Azure/agentbaker`, why I am also certain it needs goproxy is because the compilation of the go binary failed when I hadn't updated my PAT, and it tried to import the package :)

pkg/nbcparser/go.mod
```diff
@@ -0,0 +1,31 @@
+module github.com/Azure/agentbaker/pkg/nbcparser
```
cameronmeissner, line 1: we can probably move this to the top-level and rename the module

e.g. `module github.com/Azure/agentbaker/nbcparser` since it's own module, and not an AgentBaker pkg

packer.mk
```diff
@@ -1,6 +1,6 @@
 SHELL=/bin/bash -o pipefail
 
-build-packer:
+build-packer: build-nbcparser-binary
```
cameronmeissner, line 3: noice

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -869,6 +869,18 @@ testBccTools () {
...
+testNBCParserBinary () {
+  local test="testNBCParserBinary"
+  local go_binary_path="/opt/azure/containers/nbcparser"
+
+  echo "$test: checking existence of nbcparser go binary at $go_binary_path"
+  if [ ! -f "$go_binary_path" ]; then
+    err "$test: nbcparser go binary does not exist at $go_binary_path"
+    return 1
+  fi
+  echo "$test: nbcparser go binary exists at $go_binary_path"
```
cameronmeissner, line 881: should we also invoke it and assert exit 0?

basically just call the help command like `nbcparser -h` and make sure it works

vhdbuilder/packer/vhd-image-builder-base.json
cameronmeissner: seems like no goproxy works, im assuming what happened in your case was you still had you GOPROXY vars setup using an invalid PAT, in which case go just blindly tried to use the proxy and thus it failed

vhdbuilder/packer/test/linux-vhd-content-test.sh
lilypan26: yea i was thinking about this, but i need a sample json with some test data in order to trigger it, so not sure if its work adding that to the vhd as well

djsly: we should, if its wrong compiled, it won't be of any use, we should ensure that it's working and ready to be executed.

nbcparser/go.mod
```diff
@@ -0,0 +1,31 @@
+module github.com/Azure/agentbaker/nbcparser
```
djsly, line 1: can we ensure we have Dependabot configured for AgentBaker to have those dips handled with automatic CVE updated.

lilypan26:
> what does `NBC` stand for ?

node bootstrapping config/contract

nbcparser/go.mod
lilypan26: dips? not sure i follow, can you elaborate?

djsly: deps (dependancies)

cameronmeissner (approved):

nbcparser/go.mod
lilypan26: Just double checked with Cameron that we already have a dependabot configured for AB

PR #4408, opened 2024-05-16
feat: STLS bootstrap integration with windows

cameronmeissner:
**What type of PR is this?**

This PR aims to add support for the secure TLS bootstrapping feature to windows nodes. For reference, see the analog PR for the linux side here: https://github.com/Azure/AgentBaker/pull/4386, as the logic is the exact same. 

With secure TLS bootstrapping, CSE will no longer always attempt to only use a bootstrap token so kubelet can bootstrap its own credentials. Instead, before the kubelet is started, a script will be invoked with calls a new component that is downloaded to the node at runtime (and eventually added to the VHD cache), the secure TLS bootstrapping client. This client will be invoked continuously until it succeeds, or until a deadline of 3 minutes (subject to change) is met.

In cases where the client succeeds, a valid kubeconfig referencing a fresh kubelet client certificate and key will be placed within KubeDir which will then be used by the kubelet to register itself with the API server. In cases where the bootstrap client fails after the 3 minute deadline, the bootstrap token passed by RP will be used as a fall-back so the kubelet can bootstrap its own certificate.

This PR is still a work-in-progress, though I'm now looking for preliminary reviews. I've gone as far as to test this within e2e environments and have seen windows nodes successfully join clusters without the usage of the bootstrap token.

NOTE: A couple changes have been made to the primary CSE scripts (outside of the CSE package) to enable the new functionality, namely that we now install calico _before_ the kubelet is started. Furthermore, we also now remove the temporary kubeconfig used to install calico after it has successfully been installed in both the secure TLS bootstrapping cases, as well as when using a bootstrap token.

From the rollout perspective, since this includes both changes to the primary CSE scripts as well as to the CSE package, I think we'll want to roll out the package changes first, and then once those are rolled out completely, proceed with rolling out the changes to the primary scripts.

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -167,6 +167,11 @@ $global:IsAzureCNIOverlayEnabled = {{if IsAzureCNIOverlayFeatureEnabled}}$true{{
 # Kubelet credential provider
 $global:CredentialProviderURL = "{{GetParameter "windowsCredentialProviderURL"}}"
 
+# Kubelet secure TLS bootstrapping settings and related vars
+$global:EnableSecureTLSBootstrapping = [System.Convert]::ToBoolean("{{EnableSecureTLSBootstrapping}}")
+$global:CustomSecureTLSBootstrapAADResource = "{{GetCustomSecureTLSBootstrapAADServerAppID}}"
+$global:SecureTLSBootstrapClientDownloadURL = "https://kubernetesreleases.blob.core.windows.net/aks-tls-bootstrap-client/client-v0.1.0-alpha.4/windows/amd64/tls-bootstrap-client.exe"
```
cameronmeissner, line 173: not yet caching

AbelHu: We need to use different endpoints per clouds. I think that we need to add a new property for this url and set it in rp

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -296,7 +301,11 @@ try
     Get-ProvisioningScripts
     Get-LogCollectionScripts
     
-    Write-KubeClusterConfig -MasterIP $MasterIP -KubeDnsServiceIp $KubeDnsServiceIp
+    Write-KubeClusterConfig `
+        -MasterIP $MasterIP `
+        -KubeDnsServiceIp $KubeDnsServiceIp `
+        -EnableSecureTLSBootstrapping $global:EnableSecureTLSBootstrapping `
+        -CustomSecureTLSBootstrapAADResource $global:CustomSecureTLSBootstrapAADResource
```
AbelHu, line 308: we only can change it after changing the function in cse script package and updating the cse script package version.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -411,7 +425,22 @@ try
     }
 
     New-ExternalHnsNetwork -IsDualStackEnabled $global:IsDualStackEnabled
+
+    # Install calico using the kubeconfig at KubeDir/config, if enabled
+    if ($global:WindowsCalicoPackageURL) {
```
AbelHu, line 430: we need to install calico after Check-APIServerConnectivity since it needs to access the api server

parts/windows/kuberneteswindowssetup.ps1
cameronmeissner: made a note about this in the description :)

parts/windows/kuberneteswindowssetup.ps1
cameronmeissner: yeah, this is also just for testing at the moment basically (upstream staging storage account)

would we need to set it differently for various clouds if we used acs-mirror however?

I think we also want to start putting stuff in MCR from what I've been hearing recently

parts/windows/kuberneteswindowssetup.ps1
AbelHu: It needs to be separated into two PRs.
 1. update the scripts in staging/cse/windows; 
 2. release a new cse script package
 3. update kuberneteswindowssetup.ps1 with the new cse script package version

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
```diff
@@ -0,0 +1,122 @@
+Param(
```
AbelHu, line 1: Please also update README in staging/cse/windows and update the script to publish aks windows cse script package in ado repo

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -48,6 +55,29 @@ if ($global:KubeBinariesVersion -lt "1.27.0") {
...
 
+# Run the secure TLS bootstrap script to generate a kubelet client certificate, if enabled
+if ($global:EnableSecureTLSBootstrapping) {
+    $aadResource = $global:DefaultSecureTLSBootstrapAADResource
+    if ($global:CustomSecureTLSBootstrapAADResource) {
+        $aadResource = $global:CustomSecureTLSBootstrapAADResource
+    }
+    & "c:\k\securetlsbootstrap.ps1" -KubeDir $global:KubeDir -APIServerFQDN $global:MasterIP -AADResource $aadResource
+    if (!$?) {
+        Write-Host "Secure TLS bootstrapping failed, will still try to start kubelet..."
```
AbelHu, line 66: what will happen if there is no valid cert/token?

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -48,6 +55,29 @@ if ($global:KubeBinariesVersion -lt "1.27.0") {
...
+    } else {
+        Write-Host "Secure TLS bootstrapping succeeded"
+        if ((Test-Path $global:KubeconfigPath) -and (Test-Path $global:BootstrapKubeconfigPath)) {
+            # Remove the bootstrap-kubeconfig flag from kubelet args, if present.
+            # This is needed for the secure TLS bootstrapping case since if --bootstrap-kubeconfig is specified
+            # alongside --rotate-certificates, kubelet will always treat what we give it as bootstrap credentials rather
+            # than valid kubelet client credentials at startup, and thus will try to bootstrap its own credentials.
+            Write-Host "No need for bootstrap-kubeconfig: removing bootstrap-kubeconfig arg from kubelet arg list and bootstrap-kubeconfig file from disk"
+            $KubeletArgList = @($KubeletArgList) -NotMatch "bootstrap-kubeconig"
+            Remove-Item $global:BootstrapKubeconfigPath
```
AbelHu, line 76: will it fail if it has been removed?

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
```diff
@@ -0,0 +1,122 @@
...
+}
+
+filter Timestamp { "$(Get-Date -Format o): $_" }
+
+function Write-Log($message) {
+    $msg = $message | Timestamp
+    Write-Output $msg
+}
+
+function As-Event {
```
AbelHu, line 44: could you wrapper this function to a common function to call it in non-cse scripts? We have a function `Logs-To-Event` to be called in cse scripts.

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function Bootstrap {
+    $now = CurrentUnixTimeSeconds
+    $deadline = $now + 180 # 3 minute deadline
+    Write-Log "Start secure TLS bootstrapping at time: $now"
+    Write-Log "Secure TLS bootstrapping deadline is: $deadline"
+    
+    while([DateTime]$now -lt [DateTime]$deadline) {
+        & $BootstrapClientPath bootstrap `
```
AbelHu, line 100: Will it be run once or running as a background service? It seems like that it is run once.

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -48,6 +55,29 @@ if ($global:KubeBinariesVersion -lt "1.27.0") {
     $KubeletArgList += @("--container-runtime=remote")
 }
 
+# Run the secure TLS bootstrap script to generate a kubelet client certificate, if enabled
+if ($global:EnableSecureTLSBootstrapping) {
+    $aadResource = $global:DefaultSecureTLSBootstrapAADResource
+    if ($global:CustomSecureTLSBootstrapAADResource) {
+        $aadResource = $global:CustomSecureTLSBootstrapAADResource
+    }
+    & "c:\k\securetlsbootstrap.ps1" -KubeDir $global:KubeDir -APIServerFQDN $global:MasterIP -AADResource $aadResource
```
AbelHu, line 64: kubeletstart.ps1 will be run when
- Node is provisioned
- Node is restarted
- after kubelet crashed
Is it expected to always call securetlsbootstrap.ps1 before starting kubelet?

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -353,6 +362,12 @@ try
         New-CsiProxyService -CsiProxyPackageUrl $global:CsiProxyUrl -KubeDir $global:KubeDir
     }
 
+    if ($global:EnableSecureTLSBootstrapping) {
+        $clientPath = [Io.path]::Combine("$global:KubeDir", "tls-bootstrap-client.exe")
+        Write-Log "Secure TLS Bootstrapping is enabled, downloading bootstrap client to $clientPath if needed..."
+        Get-SecureTLSBootstrapClient -BootstrapClientPath $clientPath -BootstrapClientDownloadUrl $global:SecureTLSBootstrapClientDownloadURL
+    }
+
     if ($global:TLSBootstrapToken) {
```
AbelHu, line 371: Could both EnableSecureTLSBootstrapping and TLSBootstrapToken be true?

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -353,6 +362,12 @@ try
         New-CsiProxyService -CsiProxyPackageUrl $global:CsiProxyUrl -KubeDir $global:KubeDir
     }
 
+    if ($global:EnableSecureTLSBootstrapping) {
```
AbelHu, line 365: It seems like that it only downloads the binary in cse in node provisioning. So it may still fail in running `tls-bootstrap-client.exe` after cse returns success. Is it by design?

staging/cse/windows/provisioningscripts/kubeletstart.ps1
cameronmeissner: if there's no valid cert after secure bootstrapping then we'd fall-back to the token supplied by RP

if the token itself is invalid, then the node wouldn't be able to join the cluster... though that's what the case is currently since we always use a bootstrap token as far as I know

after the first 2 rollout phases are completed, we plan on removing/disabling bootstrap tokens _only_ once we see high qos/good enough signals from the secure protocol

parts/windows/kuberneteswindowssetup.ps1
cameronmeissner: could we also just move the Check-APIServerConnectivity call up here before the calico installation? I guess the main difference there would be we do the connectivity check before installing/starting the kubelet service

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
cameronmeissner: fixed

parts/windows/kuberneteswindowssetup.ps1
cameronmeissner: also referenced here: https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication

parts/windows/kuberneteswindowssetup.ps1
AbelHu: Yes. I think that we can call Check-APIServerConnectivity and then install calico services

staging/cse/windows/provisioningscripts/kubeletstart.ps1
AbelHu: what will happen if securetlsbootstrap.ps1 found that the cert is expired but it failed in getting new cert?

AbelHu: And how will it handle the expired cert when kubelet is running?

cameronmeissner: kubelet is responsible for rotating its own certificate after its bootstrapped with an initial one (via the `--rotate-certificates` flag)

in cases where the bootstrap client sees that the kubeconfig references an expired certificate (or a certificate signed by an out-of-date cluster CA), it will initiate the bootstrap process

parts/windows/kuberneteswindowssetup.ps1
AbelHu: How do we know which actually works If both of them exist? Could we set $global:TLSBootstrapToken to false automatically when $global:EnableSecureTLSBootstrapping is true? I suppose that $global:EnableSecureTLSBootstrapping will be behind of a toggle so we can rollout it per sub or by region and easily rollout it back.

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
AbelHu: After it is merged but before you publish new cse script package, you need to update the script to publish aks windows cse script package in ado repo

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -48,6 +55,31 @@ if ($global:KubeBinariesVersion -lt "1.27.0") {
...
+    if ($?) {
+        if ((Test-Path $global:KubeconfigPath) -and (Test-Path $global:BootstrapKubeconfigPath)) {
```
AbelHu, line 66: When will it happen when $? is true but the expected files do not exist? Should we add some logs?

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
```diff
@@ -0,0 +1,123 @@
...
+    Param(
+        [Parameter(Mandatory = $true)][ScriptBlock]
+        $Command
+    )
+
+    $eventsFileName = [DateTimeOffset]::UtcNow.ToUnixTimeMilliseconds()
+
+    # execute the command and record how long it takes, as well as its exit code
+    $startTime = $(Get-Date -Format "yyyy-MM-dd HH:mm:ss.fff")
+    & $Command
```
AbelHu, line 54: maybe you can use
```
function As-Event {
    Param(
        [Parameter(Mandatory = $true)][ScriptBlock]
        $Command
    )
    try {
        Write-Host "Invoking command...";
        $Command.Invoke();
        Write-Host "Command is finished sucessfully";
    } catch {
        Write-Error $_
    }
}
As-Event -Command {
    ...
    throw "error";
}
```

staging/cse/windows/provisioningscripts/securetlsbootstrap.ps1
AbelHu: I think that you also need to run it in kuberneteswindowssetup.ps1 for some scenarios which requires cse to access the api server.

parts/windows/kuberneteswindowssetup.ps1
cameronmeissner: it's mentioned in the design doc, but we primarily plan on determining success rates based on the guest-agent telemetry events generated by securetlsbootstrap.ps1 (e.g. Logs-To-Events/As-Event)

cameronmeissner:
replacing with #6710

PR #4426, opened 2024-05-22
feat: include hostname and cluster FQDN in guest agent telemetry events

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -352,6 +353,12 @@ logs_to_events() {
     ret=$?
     local endTime=$(date +"%F %T.%3N")
 
+    msg=$(jq -n \
+        --arg Hostname "$(uname -n)" \
+        --arg ClusterFQDN "$API_SERVER_NAME" \
+        --arg Message "Completed: $*" \
```
djsly, line 359: Why are we printing all of the received arguments ? what are they ? should we instead give them proper names to make it more explicit in the code what is getting printed to accompany the "completed" status.

cameronmeissner: the issue is that this function is designed to take in arbitrary calls to arbitrary functions, thus we can't really be more specific here unless we were to come up with a standardized arg pattern that would need to be adhered to whenever this function is called

cameronmeissner:
discussed with @phealy offline, will probably end up closing this since we can just join on other CRP tables to get this info instead

PR #4330, opened 2024-04-23
chore: remove arm64 guard for fuse3

artunduman:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:
- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:
```
none
```

PR #4428, opened 2024-05-22
chore: add blobfuse2 support to 24.04

anujmaheshwari1:
**What type of PR is this?**
This PR adds blobfuse2 installation script for Ubuntu 24.04, previously, the scripts skipped it since the package wasn't available upstream, it is now.

https://packages.microsoft.com/ubuntu/24.04/prod/pool/main/b/blobfuse2/

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -41,16 +41,17 @@ installDeps() {
         BLOBFUSE2_VERSION="2.2.0"
     fi
 
-    # Currently skipping blobfuse install on 24.04 since packages are not available upstream, remove this once they are
-    if [[ "${OSVERSION}" != "24.04" ]]; then
-      pkg_list+=(blobfuse2=${BLOBFUSE2_VERSION})
+    # Sync with Andy Zhang on if blofuse2 for 22.04 can be bumped to 2.3.0 to maintain parity, right now its 2.2.1
```
andyzhangx, line 44: blobfuse 2.3.0 works well, I have upgraded to 2.3.0 in blob csi driver upstream project, thx

pkg/agent/testdata/RawUbuntuContainerd/line62.sh
```diff
@@ -39,12 +39,13 @@ installDeps() {
         BLOBFUSE2_VERSION="2.2.0"
     fi
 
-    if [[ "${OSVERSION}" != "24.04" ]]; then
```
artunduman, line 42: There is not `templates_generated.go` changes in this PR too
Do you all know if that's expected? Is that file no longer used?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -41,16 +41,17 @@ installDeps() {
...
     fi
 
-    if [[ $(isARM64) != 1 && "${OSVERSION}" != "24.04" ]]; then
+    pkg_list+=(blobfuse2=${BLOBFUSE2_VERSION})
+    if [[ $(isARM64) != 1 ]]; then
         # blobfuse2 is installed for all ubuntu versions, it is included in pkg_list
         # for 22.04, fuse3 is installed. for all others, fuse is installed
         # for 16.04, installed blobfuse1.3.7, for all others except 22.04, installed blobfuse1.4.5
-        if [[ "${OSVERSION}" == "22.04" ]]; then
+        if [[ "${OSVERSION}" == "22.04" || "${OSVERSION}" == "24.04" ]]; then
```
artunduman, line 54: Actually, I have this PR for removing the ARM64 guard https://github.com/Azure/AgentBaker/pull/4330/files

pkg/agent/testdata/RawUbuntuContainerd/line62.sh
artunduman: seems expected, as discussed offline!

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -41,16 +41,17 @@ installDeps() {
...
+    # Sync with Andy Zhang on if blofuse2 for 22.04 can be bumped to 2.3.0 to maintain parity, right now its 2.2.1
+    if [ "${OSVERSION}" == "24.04" ]; then
+        BLOBFUSE2_VERSION="2.3.0"
     fi
 
-    if [[ $(isARM64) != 1 && "${OSVERSION}" != "24.04" ]]; then
+    pkg_list+=(blobfuse2=${BLOBFUSE2_VERSION})
+    if [[ $(isARM64) != 1 ]]; then
         # blobfuse2 is installed for all ubuntu versions, it is included in pkg_list
         # for 22.04, fuse3 is installed. for all others, fuse is installed
```
artunduman, line 52: nit: maybe update this comment to include 24.04 too. Or just remove the comment altogether :)

artunduman (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
anujmaheshwari1: removed arm64 check!

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
anujmaheshwari1: thanks Andy, I updated it for 22.04 as well!

jason1028kr (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -34,27 +34,22 @@ installDeps() {
     local OSVERSION
     OSVERSION=$(grep DISTRIB_RELEASE /etc/*-release| cut -f 2 -d "=")
     BLOBFUSE_VERSION="1.4.5"
-    BLOBFUSE2_VERSION="2.2.1"
+    # Blobfuse2 has been upgraded to 2.3.0 in upstream, using this version for parity between 22.04 and 24.04
+    BLOBFUSE2_VERSION="2.3.0"
```
Devinwong, line 38: Doesn't need to be this PR but just early checking.
Now we are only working on the component versions management in manifest.json.
Later when we want to centralize the others, is this BLOBFUSE2_VERSION supposed to move to manifest.json/components.json?

anujmaheshwari1: very good point, it definitely can, however, like the other components there, this isn't updated as frequently. for eg, this version bump is happening after more than 1 year.

however, with how we are moving things toward a centralized location, I think it makes sense to move it, we should just float it by Andy Zhang and storage team as well since if it is updated, it'd be by them.

Devinwong: Makes sense. Once we complete the task, we should definitely let the stakeholders know.

Devinwong (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
djsly: or is this only installed with CSE so dynamically ? and it's not cached ?

djsly: @Devinwong  I have added the component to the wiki for tracking https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/678597/Managed-Component-List?anchor=os-packages

andyzhangx (approved):

PR #4433, opened 2024-05-23
test: update pam tests to allow us to use FedRAMP-specific settings

tobiasb-ms:
**What type of PR is this?**
/kind test

This change accomplishes two things.
First, it makes the `pam` tests less flaky by making sure we don't randomly produce passwords with too many of the same character or character class in a row. This is something we see every now and then.

Second, it lays groundwork for Azure Linux FIPS images to be FedRAMP hardened, which will change some requirements around password faillocks. This change adds a parameter to the tests (currently unused) to allow us to specify that once we harden the image.

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


cameronmeissner (approved):

djsly (approved):
thanks for the PR description

PR #4443, opened 2024-05-29
cleanup: use url from manifest.cue to download k8s

timmy-wright:
**What type of PR is this?**

/kind cleanup

Start getting all config information from manifest.json/cue instead of hardcoding it.

**Requirements**:
- [ X] uses [conventional commit messages](https://www.conventionalcommits.org/)
- [ ] adds unit tests

**Release note**:
```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -561,10 +561,14 @@ fi
 # Please do not use the .1 suffix, because that's only for the base image patches
 # regular version >= v1.17.0 or hotfixes >= 20211009 has arm64 binaries.
 KUBE_BINARY_VERSIONS="$(jq -r .kubernetes.versions[] manifest.json)"
+KUBE_URL="$(jq -r .kubernetes.downloadURL manifest.json)"
 
 for PATCHED_KUBE_BINARY_VERSION in ${KUBE_BINARY_VERSIONS}; do
   KUBERNETES_VERSION=$(echo ${PATCHED_KUBE_BINARY_VERSION} | cut -d"_" -f1 | cut -d"-" -f1 | cut -d"." -f1,2,3)
-  extractKubeBinaries $KUBERNETES_VERSION "https://acs-mirror.azureedge.net/kubernetes/v${PATCHED_KUBE_BINARY_VERSION}/binaries/kubernetes-node-linux-${CPU_ARCH}.tar.gz" false
+  KUBE_VERSION_URL="${KUBE_URL/\$\{PATCHED_KUBE_BINARY_VERSION\}/${PATCHED_KUBE_BINARY_VERSION}}"
+  KUBE_VERSION_URL="${KUBE_VERSION_URL/\${CPU_ARCH\}/${CPU_ARCH}}"
```
djsly, line 569: any reason why this cannot be combined with the previous line ?    when reading the code, I was wondering why you were setting the same VAR twice, till I realized you were augmenting it.

timmy-wright: It took me a bit to figure out the right shell syntax. I've modified it now to use "envsubst" as that seems much cleaner.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -561,10 +561,14 @@ fi
 # Please do not use the .1 suffix, because that's only for the base image patches
 # regular version >= v1.17.0 or hotfixes >= 20211009 has arm64 binaries.
 KUBE_BINARY_VERSIONS="$(jq -r .kubernetes.versions[] manifest.json)"
+KUBE_URL="$(jq -r .kubernetes.downloadURL manifest.json)"
```
cameronmeissner, line 564: just food for thought, we should have some validation which makes sure the manifest file includes exactly what we expect, since if there isn't an entry for `.kubernetes.downloadURL`,  this would break.

we can probably bake that validation directly into the cue schema itself, if it's not already there

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -561,10 +561,14 @@ fi
...
 
 for PATCHED_KUBE_BINARY_VERSION in ${KUBE_BINARY_VERSIONS}; do
-  KUBERNETES_VERSION=$(echo ${PATCHED_KUBE_BINARY_VERSION} | cut -d"_" -f1 | cut -d"-" -f1 | cut -d"." -f1,2,3)
-  extractKubeBinaries $KUBERNETES_VERSION "https://acs-mirror.azureedge.net/kubernetes/v${PATCHED_KUBE_BINARY_VERSION}/binaries/kubernetes-node-linux-${CPU_ARCH}.tar.gz" false
+  KUBERNETES_VERSION=$(echo "${PATCHED_KUBE_BINARY_VERSION}" | cut -d"_" -f1 | cut -d"-" -f1 | cut -d"." -f1,2,3)
+
```
cameronmeissner, line 568: super nit: lets get rid of the newlines here, we actually do run into CSE/customdata length limitations occasionally in AGC...

vhdbuilder/packer/install-dependencies.sh
timmy-wright: Good point.  Will add that.

vhdbuilder/packer/install-dependencies.sh
timmy-wright: super super nit - this script isn't run as part of CSE or CustomData, so that limit shouldn't apply ;) Unless there's something different in airgap?

djsly (approved):

vhdbuilder/packer/install-dependencies.sh
timmy-wright: I checked and the cue schema already requires it.

.github/workflows/check-shell.yml
```diff
@@ -12,3 +12,14 @@ jobs:
...
       name: Verify Shell scripts
+    - run: |
+        make unit-test-shell
+      name: Unit test Shell scripts
+    - name: Publish Test Results
+      uses: EnricoMi/publish-unit-test-result-action@v2
+      if: always()
+      with:
+        files: |
+          test-results/**/*.xml
```
djsly, line 23: I forgot does the glob pattern `**` also looks in the current dir ? Since `test-results/bats_shell.junit.xml` is not found in any subdirectory

bash_tests/cse_install.test.bats.sh
```diff
@@ -0,0 +1,23 @@
...
+  unmock rm
+}
+
+@test "test calling removeManDbAutoUpdateFlagFile" {
+  removeManDbAutoUpdateFlagFile
+
+  assert_called rm -f /var/lib/man-db/auto-update
+}
+
+@test "failing test" {
```
djsly, line 20: I'm guessing this is simply to test that the logic works and it's not tone committed ?

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: brain fart on my part :D

cameronmeissner (changes requested):
maybe we want to create a separate PR for the new shell unit testing stuff?

otherwise I think we decided to move towards components.json and away from manifest.cue/json

cameronmeissner:
closing for now, feel free to reopen

PR #4445, opened 2024-05-29
bug: add systemd dropon for chrony so it will restart

timmy-wright:
**What type of PR is this?**

/kind bug

Fixes #27616556

**Requirements**:
- [x] includes documentation
- [x] adds unit tests

* Have manually tested by creating VMs with the VHDs for mariner2, azurelinux2, ubuntu1804, and ubuntu2204, ssh'd to it, checked chrony was running, killed it, and checked it restarted.
* Have e2e tests verifying the contents of the systemd drop-in for those 3 distros
* have e2e tests replicating the automated tests.

**Release note**:
```
Add drop-in for systemd so Chrony will restart if it crashes
```

### DIFF
--- a/e2e/artifact/download.go
+++ b/e2e/artifact/download.go
@@ -80,8 +80,10 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 		ArtifactName: to.Ptr(artifactName),
 	})
 	if err != nil {
+		log.Printf(err.Error())
 		if isMissingArtifactError(err) {
 			log.Printf("unable to download publishing info %q, not found for build ID: %d", artifactName, opts.BuildID)
+			panic(err)
 		} else {
 			d.errChan <- fmt.Errorf("unable get artifact info for %q: %w", artifactName, err)
 		}
@@ -94,14 +96,14 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 	zipOut, err := os.Create(zipName)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to create zip archive %s: %w", zipName, err)
-		return
+		panic(err)
 	}
 
 	client := &http.Client{}
 	req, err := http.NewRequest(http.MethodGet, downloadURL, nil)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to create new HTTP request: %w", err)
-		return
+		panic(err)
 	}
 
 	req.Header.Add("Authorization", fmt.Sprintf("Basic %s", d.basicAuth))
@@ -109,22 +111,22 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 	resp, err := client.Do(req)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to perform HTTP request to download artifact: %w", err)
-		return
+		panic(err)
 	}
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
 		d.errChan <- fmt.Errorf("unable to download artifact from %s, received HTTP status code: %d", downloadURL, resp.StatusCode)
-		return
+		panic(err)
 	}
 
 	if _, err = io.Copy(zipOut, resp.Body); err != nil {
 		d.errChan <- fmt.Errorf("unable to copy artifact data to zip archive: %w", err)
-		return
+		panic(err)
 	}
 
 	if err = extractPublishingInfoFromZip(artifactName, zipName, opts.TargetDir); err != nil {
 		d.errChan <- fmt.Errorf("unable to extract publishing info from zip archive %s: %w", zipName, err)
-		return
+		panic(err)
 	}
 }
 
@@ -151,6 +153,7 @@ func (d *Downloader) getVHDPublishingInfoArtifactNames(ctx context.Context, opts
 
 	for _, artifact := range *artifacts {
 		if strings.HasPrefix(*artifact.Name, "publishing-info") {
+			log.Printf("Got publishing info artifact: %s", *artifact.Name)
 			artifactNames = append(artifactNames, *artifact.Name)
 		}
 	}
@@ -194,6 +197,6 @@ func extractPublishingInfoFromZip(artifactName, zipName, targetDir string) error
 
 func isMissingArtifactError(err error) bool {
 	// there doesn't seem to be a proper error type for this,
-	// thus we need to assert on thecontents of the error msg itself
+	// thus we need to assert on the contents of the error msg itself
 	return err != nil && (strings.Contains(err.Error(), "404 Not Found") || strings.Contains(err.Error(), "was not found for build"))
 }
--- a/e2e/artifact/download.go
+++ b/e2e/artifact/download.go
@@ -82,8 +82,7 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 	if err != nil {
 		log.Printf(err.Error())
 		if isMissingArtifactError(err) {
-			log.Printf("unable to download publishing info %q, not found for build ID: %d", artifactName, opts.BuildID)
-			panic(err)
+			log.Fatalf("unable to download publishing info %q, not found for build ID: %d", artifactName, opts.BuildID)
 		} else {
 			d.errChan <- fmt.Errorf("unable get artifact info for %q: %w", artifactName, err)
 		}
@@ -96,14 +95,14 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 	zipOut, err := os.Create(zipName)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to create zip archive %s: %w", zipName, err)
-		panic(err)
+		log.Fatalf("unable to create zip archive %s: %w", zipName, err)
 	}
 
 	client := &http.Client{}
 	req, err := http.NewRequest(http.MethodGet, downloadURL, nil)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to create new HTTP request: %w", err)
-		panic(err)
+		log.Fatalf("unable to create new HTTP request: %w", err)
 	}
 
 	req.Header.Add("Authorization", fmt.Sprintf("Basic %s", d.basicAuth))
@@ -111,22 +110,22 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 	resp, err := client.Do(req)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to perform HTTP request to download artifact: %w", err)
-		panic(err)
+		log.Fatalf("unable to perform HTTP request to download artifact: %w", err)
 	}
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
 		d.errChan <- fmt.Errorf("unable to download artifact from %s, received HTTP status code: %d", downloadURL, resp.StatusCode)
-		panic(err)
+		log.Fatalf("unable to download artifact from %s, received HTTP status code: %d", downloadURL, resp.StatusCode)
 	}
 
 	if _, err = io.Copy(zipOut, resp.Body); err != nil {
 		d.errChan <- fmt.Errorf("unable to copy artifact data to zip archive: %w", err)
-		panic(err)
+		log.Fatalf("unable to copy artifact data to zip archive: %w", err)
 	}
 
 	if err = extractPublishingInfoFromZip(artifactName, zipName, opts.TargetDir); err != nil {
 		d.errChan <- fmt.Errorf("unable to extract publishing info from zip archive %s: %w", zipName, err)
-		panic(err)
+		log.Fatalf("unable to extract publishing info from zip archive %s: %w", zipName, err)
 	}
 }
 
--- a/e2e/scenario/base_vhd_catalog.json
+++ b/e2e/scenario/base_vhd_catalog.json
@@ -2,7 +2,7 @@
     "ubuntu1804": {
         "gen2containerd": {
             "artifactName": "1804-gen2-containerd",
-            "resourceId": "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1716223055.6413"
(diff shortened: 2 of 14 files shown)


djsly (approved):
LGTM

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -77,12 +77,19 @@ else
   done
 fi
 
+mkdir -p /etc/systemd/system/chrony.service.d
+tee -a /etc/systemd/system/chrony.service.d/10-chrony-restarts.conf > /dev/null <<'EOF'
+Restart=always
```
cameronmeissner, line 82: I think we need a `[Service]` directive at the top here

cameronmeissner: also, we should probably make sure to only do this _after_ chrony has been installed and the unit has been setup, otherwise there's a risk of this configuration being clobbered - probably something good to test with agentbaker e2e's.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -77,12 +77,19 @@ else
   done
 fi
 
+mkdir -p /etc/systemd/system/chrony.service.d
+tee -a /etc/systemd/system/chrony.service.d/10-chrony-restarts.conf > /dev/null <<'EOF'
```
cameronmeissner, line 81: nit: can use `cat` without the need to redirect output to the null device

cameronmeissner:
best way to verify would be to to get on a live node with these changes and run `systemctl cat chrony` and verify the settings you're overriding are showing up via drop-in override

might also be a good idea to add an agentbakere2e validator to catch this as well: https://github.com/Azure/AgentBaker/blob/master/e2e/scenario/validators.go

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -77,12 +77,20 @@ else
   done
 fi
 
+mkdir -p /etc/systemd/system/chrony.service.d
+cat >> /etc/systemd/system/chrony.service.d/10-chrony-restarts.conf <<'EOF'
```
cameronmeissner, line 81: I think we can also edit the first EOF to be `EOF` instead of `'EOF'`

e2e/scenario/scenario.go
```diff
@@ -77,6 +77,10 @@ func (t *Template) scenarios() []*Scenario {
 		t.marinerv2AirGap(),
 		t.azurelinuxv2AirGap(),
 		t.marinerv2ARM64AirGap(),
+		t.ubuntu2204SystemdChronyDropin(),
```
djsly, line 80: could we look at ordering the scenarios in list group by OSFamily ? to make it easier to see what is being tested on which OS ?

e2e/scenario/scenario.go
```diff
@@ -77,6 +77,10 @@ func (t *Template) scenarios() []*Scenario {
...
+		t.ubuntu2204ChronyRestarts(),
```
djsly, line 81: @anujmaheshwari1  will we be duplicating ALL of the 2204 e2e scenarios with 2404 ? 

Should we look at improving those scenarios to be OS agnostic, looping them across all OS family /version

e2e/scenario/validators.go
```diff
@@ -103,6 +103,45 @@ func NonEmptyDirectoryValidator(dirName string) *LiveVMValidator {
...
+	steps := []string{
+		"echo Existing processes:",
+		fmt.Sprintf("pgrep --list-full %s", serviceName),
+		"echo Killing...",
+		// we use systemctl kill rather than kill -9 because container restrictions stop us sending a kill sig to a process
+		fmt.Sprintf("systemctl kill %s", serviceName),
+		"echo Sleeping...",
+		fmt.Sprintf("sleep %d", timeoutInSeconds),
+		"echo Checking processes have restarted...",
+		fmt.Sprintf("pgrep --list-full %s", serviceName),
```
djsly, line 129: we should ensure that is has a different PID

djsly (changes requested):

vhdbuilder/packer/install-dependencies.sh
timmy-wright: Done.

vhdbuilder/packer/install-dependencies.sh
timmy-wright: Done. Cat is far nicer!

e2e/scenario/scenario.go
timmy-wright: Good idea. I'll include this reordering in the PR.

e2e/scenario/scenario.go
timmy-wright: I was wondering this too. It should be feasible to say a scenario runs on several VHDs rather than specifying only one. On saying that, the test for chrony on Ubuntu1804 is slightly different to 2204. It wouldn't surprise me if there were differences for Ubuntu2404 as well.

e2e/scenario/validators.go
timmy-wright: That's a good idea. I was having issues using "pgrep" in this particular command line and have since changed to a different approach. Once I've got these tests actually working, I'll try to add the pid change check. Or just a check that it's inactive after killing it.

vhdbuilder/packer/install-dependencies.sh
timmy-wright: Agree. Have added e2e tests for ubuntu 1804, 2204, azure linuxv2 and marinerv2. Now to get them passing...

e2e/scenario/scenario.go
cameronmeissner: we have some scenarios which are responsible for only getting coverage on the particular SKU, rather than testing a particular scenario, though I'd be open to the idea of creating OS-agnostic scenarios

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -300,4 +300,16 @@ steps:
       artifactName: '${{ parameters.artifactName }}'
       targetPath: '$(Build.ArtifactStagingDirectory)'
     displayName: Publish artifacts
-    condition: always()
\ No newline at end of file
+    condition: always()
+  - task: PublishPipelineArtifact@1
+    displayName: Publish publishing-info so e2e tests can be run locally
```
cameronmeissner, line 305: what is this for? e2e's can already be run locally?

cameronmeissner: oh, I think this was because of the bug to remove these, that should be fixed now with a merge from master

e2e/scenario/scenario_azurelinuxv2-chrony-restarts.go
```diff
@@ -0,0 +1,44 @@
...
+		Description: "Tests that the systemd drop-in file for chrony is placed in the right place",
+		Config: Config{
+			ClusterSelector: NetworkPluginKubenetSelector,
+			ClusterMutator:  NetworkPluginKubenetMutator,
+			VHDSelector:     t.AzureLinuxV2Gen2,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-azurelinux-v2-gen2"
+				nbc.AgentPoolProfile.Distro = "aks-azurelinux-v2-gen2"
+			},
+			LiveVMValidators: []*LiveVMValidator{
```
cameronmeissner, line 19: can we combine these scenarios/validators?

e2e/scenario/validators.go
```diff
@@ -103,6 +103,72 @@ func NonEmptyDirectoryValidator(dirName string) *LiveVMValidator {
...
+	str = strings.Replace(str, "(", "OB", -1)
+	str = strings.Replace(str, ")", "CB", -1)
+	str = strings.Replace(str, "||", "OR", -1)
+	str = strings.Replace(str, "&&", "AND", -1)
+
+	return str
+}
+
+func serviceCanRestartValidator(serviceName string, restartTimeoutInSeconds int) *LiveVMValidator {
+	steps := []string{
```
cameronmeissner, line 132: maybe we should investigate whether we can just use dedicated scripts to accomplish this at some point

timmy-wright: Oh absolutely. I'd love to have this specified in a yaml file so we can add new ones easily. This code is a bit of a hack TBH.

e2e/scenario/scenario_azurelinuxv2-chrony-restarts.go
timmy-wright: I plan to - once I've managed to get them passing locally. Have been battling issues running the tests against custom VHDs and having them separate until I get the tests passing is useful.

.pipelines/templates/.builder-release-template.yaml
timmy-wright: Yup. Have removed this from the PR now.

e2e/scenario/scenario_azurelinuxv2-chrony-restarts.go
timmy-wright: Combined now.

e2e/scenario/scenario.go
timmy-wright: Resolving convo as it's a really a feature add for a different PR.

e2e/scenario/validators.go
timmy-wright: Added this check.

e2e/artifact/download.go
```diff
@@ -94,37 +96,37 @@ func (d *Downloader) downloadPublishingInfo(ctx context.Context, tempDir, artifa
 	zipOut, err := os.Create(zipName)
 	if err != nil {
 		d.errChan <- fmt.Errorf("unable to create zip archive %s: %w", zipName, err)
-		return
+		panic(err)
```
djsly, line 99: won't those panic prevent other test cases from the 2e2 suite to complete ?  Do we know which binary will exit from this panic call ?

If each testcases are separate coroutine and the suite will continue, this is fine

e2e/scenario/scenario.go
```diff
@@ -44,39 +44,50 @@ func GetScenariosForSuite(ctx context.Context, suiteConfig *suite.Config) (Table
...
+		t.azurelinuxv2(),
+		t.azurelinuxv2ARM64(),
+		t.azurelinuxv2gpu(),
+		t.azurelinuxv2CustomSysctls(),
+		t.azurelinuxv2Wasm(),
+		t.azurelinuxv2_azurecni(),
+		t.azurelinuxv2gpu_azurecni(),
+		t.azurelinuxv2ARM64AirGap(),
+		t.azurelinuxv2AirGap(),
+		t.azurelinuxv2ChronyRestarts(),
```
djsly, line 90: thanks a lot @timmy-wright  for this clean up  It makes it 110% more readable and we will easily be able to spot where test are missing and what is being tested where!

e2e/scenario/validators.go
```diff
@@ -103,6 +103,94 @@ func NonEmptyDirectoryValidator(dirName string) *LiveVMValidator {
...
+		// print the status of the service and then verify it is active.
+		fmt.Sprintf("(systemctl -n 5 status %s || true)", serviceName),
+		fmt.Sprintf("systemctl is-active %s", serviceName),
+
+		// get the PID of the service after restart, so we can check it's changed
+		fmt.Sprintf("POST_PID=`sudo pgrep %s`", serviceName),
+		"echo POST_PID: $POST_PID",
+
+		// verify the PID has changed.
+		"if [[ \"$INITIAL_PID\" == \"$POST_PID\" ]]; then exit 1; fi",
```
djsly, line 177:
```suggestion
		"if [[ \"$INITIAL_PID\" == \"$POST_PID\" ]]; then echo 'PID didn't change after restart, failing validator.'; exit 1; fi",
```

e2e/scenario/vhd.go
```diff
@@ -236,6 +240,8 @@ func (c *VHDCatalog) addEntryFromPublishingInfo(info artifact.VHDPublishingInfo)
 			c.CBLMarinerV2.Gen2Arm64.ResourceID = id
 		case vhdNameCBLMarinerV2Gen2:
 			c.CBLMarinerV2.Gen2.ResourceID = id
+		default:
+			panic(fmt.Sprintf("cannot assign VHD resource IDto VHD name \"%s\" as name not known.", name))
```
djsly, line 244:
```suggestion
			panic(fmt.Sprintf("cannot assign VHD resource ID to VHD name \"%s\" as name not known.", name))
```

e2e/suite_test.go
```diff
@@ -119,27 +119,27 @@ func runScenario(ctx context.Context, t *testing.T, r *mrand.Rand, opts *scenari
 	if err != nil {
 		vmssSucceeded = false
 		if !isVMExtensionProvisioningError(err) {
-			t.Fatalf("encountered an unknown error while creating VM: %v", err)
+			t.Fatalf("encountered an unknown error while creating VM %s: %v", vmssName, err)
```
djsly, line 122: FatalF will perform the same as `panic` no ? should we focus at levering one or the other ?

djsly (changes requested):

e2e/scenario/base_vhd_catalog.json
```diff
@@ -2,7 +2,7 @@
     "ubuntu1804": {
         "gen2containerd": {
             "artifactName": "1804-gen2-containerd",
-            "resourceId": "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1716223055.6413"
+            "resourceId": "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1717624756.7706"
```
UtheMan, line 5: Why was VHD update needed here? Just asking for context

UtheMan (approved):
lgtm for the most part, curious about Sylvains comments too though

e2e/scenario/base_vhd_catalog.json
anujmaheshwari1: because we add a new drop in file for chrony restart to the VHD

e2e/artifact/download.go
timmy-wright: This code path is only run when you run the e2e tests locally using a custom build. It will stop your tests running if the VHD for one of the tests doesn't exist. Because if the tests run with the wrong VHD, then you spend half a day trying to debug your tests on the wrong VHD. Which is super painful.

e2e/suite_test.go
timmy-wright: Changed.

djsly (approved):

djsly (approved):

PR #4450, opened 2024-05-30
fix: blob conversion in VHD builds using MSI

cameronmeissner:
**What type of PR is this?**

fixes recent changes made to use the agent pool's identity for converting built VHDs from gallery image versions to classic storage account blobs

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -196,6 +196,7 @@ steps:
         -e LOCATION="${AZURE_LOCATION}" \
         -e OS_TYPE="Linux" \
         -e CLASSIC_BLOB=${CLASSIC_BLOB} \
+        -e POOL_IDENTITY_RESOURCE_ID=$(POOL_IDENTITY_RESOURCE_ID) \
         -e SIG_GALLERY_NAME=${SIG_GALLERY_NAME} \
         -e SIG_IMAGE_NAME=${SIG_IMAGE_NAME} \
         -e CAPTURED_SIG_VERSION=${captured_sig_version} \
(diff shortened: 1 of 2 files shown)


djsly (approved):

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -67,8 +68,9 @@ echo "Converted $sig_resource_id to $disk_resource_id"
 # shellcheck disable=SC2102
 sas=$(az disk grant-access --ids $disk_resource_id --duration-in-seconds 3600 --query [accessSas] -o tsv)
 
-echo "Logging into azcopy with MSI"
-azcopy-preview login --identity
+echo "Using azcopy with pool identity: $POOL_IDENTITY_RESOURCE_ID"
+export AZCOPY_AUTO_LOGIN_TYPE="MSI"
```
AlisonB319, line 72: So now we don't need to login at all, azcopy will use the existing creds?

cameronmeissner: we don't need the login command when delegating to a managed identity: https://github.com/Azure/azure-storage-azcopy/wiki/Improved-login-support-for-AzCopy-commands-(with-in-memory-secret-store)#using-a-managed-identity

AlisonB319: thats pretty neat!

AlisonB319 (approved):

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -196,6 +196,7 @@ steps:
         -e LOCATION="${AZURE_LOCATION}" \
         -e OS_TYPE="Linux" \
         -e CLASSIC_BLOB=${CLASSIC_BLOB} \
+        -e POOL_IDENTITY_RESOURCE_ID=$(POOL_IDENTITY_RESOURCE_ID) \
```
djsly, line 199: where is the ENV coming from ? ADO pipeline Var ?

PR #4455, opened 2024-06-02
feat: harden and add release creation to VHD automation pipeline

cameronmeissner:
**What type of PR is this?**

enhances the VHD automation pipeline by adding the following new features:
- automation pipeline will now automatically create the SIG release using the created artifacts on behalf of the release manager, using the devassistant service account
- release managers no longer need to specify the node image version when running the automation pipeline, as this is now automatically inferred from the publishing-info artifacts published by the VHD build
- each of the 3 "steps" performed by the automation pipeline are now run within 3 separate stages in parallel, with the overall aim to reduce pipeline runtimes, as well as to make testing the pipeline a lot easier (previously it was annoying since certain steps would checkout different branches and then run subsequent logic, making it difficult to actually test everything e2e off a particular dev branch). having dedicated stages also makes retrying simply a matter of retrying the failed stage, rather than needing to create an entirely new build.
- added some light-weight retry logic around RPCs which commonly fail at runtime

this PR also removes the "cleanup go environment" stage which is currently run before autonotes to setup the golang environment. after digging a bit deeper into this, I found that we don't actually need this step, as the pool this pipeline runs on is utilizing agents that have go properly configured.

follow-up items:
- investigate failures with ado artifact tool, potentially refactor towards a Go implementation where we can natively utilize the azdevops go sdk: https://github.com/microsoft/azure-devops-go-api

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


.pipelines/.vsts-vhd-automation.yaml
```diff
@@ -5,61 +5,146 @@ pool:
...
+    condition: eq('${{ parameters.ReleaseNotes }}', true)
+    jobs:
+    - job: automation
+      timeoutInMinutes: 30
+      steps:
+      - bash: |
+          set -x
+          echo "Removing existing go environment"
+          sudo rm -r /usr/local/go
+          GOLANG_VERSION="1.22.2"
```
djsly, line 55: do we have a global var for this ?  there is a risk that we will forgot to bump this version when bumping the generic build version.

.pipelines/.vsts-vhd-automation.yaml
```diff
@@ -5,61 +5,146 @@ pool:
...
+          echo "Downloading ${GOLANG_VERSION}"
+          curl -O "https://dl.google.com/go/${GOLANG_VERSION}.linux-amd64.tar.gz"
```
djsly, line 57: There was a security item this sprint which forbidden access to the internet during build pipelines. 

1-  can we prebake this in the agent ? 
2- is there an hosted version on MAR ?

.pipelines/.vsts-vhd-automation.yaml
```diff
@@ -5,61 +5,146 @@ pool:
...
-  condition: eq('${{ parameters.ArtifactTrigger }}', true)
+          echo "unpacking go"
+          sudo mkdir -p /usr/local/go
+          sudo chown -R "$(whoami):$(whoami)" /usr/local/go 
+          sudo tar -xvf "${GOLANG_VERSION}.linux-amd64.tar.gz" -C /usr/local
+          rm "${GOLANG_VERSION}.linux-amd64.tar.gz"
+        displayName: 'clean up go environment'
+      - bash: |
+          az extension add -n azure-devops
+          echo $MAPPED_ADO_PAT | az devops login --organization=https://dev.azure.com/msazure --verbose
```
djsly, line 67: does the `--verbose` flag print the PAT ?

Also, I see that the PAT is create in a different bash/stage, if we have access to the ENV var, how come we are loosing the login to azuredevops performed above ?

.pipelines/.vsts-vhd-automation.yaml
```diff
@@ -5,61 +5,146 @@ pool:
...
+          
+          echo "Creating EV2 artifacts and corresponding SIG release"
+          /bin/bash vhdbuilder/scripts/automate_release.sh
+        env:
+          ADO_PAT: $(PAT-aksdevassistant)
+          AZURE_DEVOPS_EXT_PAT: $(PAT-aksdevassistant)
+          VHD_BUILD_ID: $(BUILD_ID)
+        displayName: 'create SIG release'
+
+# steps:
```
djsly, line 107: Dead code, remove once testing is complete.

vhdbuilder/scripts/automate_helpers.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/bash
 set -x
 
+retrycmd_if_failure() {
+    retries=$1; wait_sleep=$2; shift && shift
+    for i in $(seq 1 $retries); do
+        "${@}" && break || \
+        if [ $i -eq $retries ]; then
+            echo Executed \"$@\" $i times;
```
djsly, line 9: maybe augment the error line with `ERROR` or `FATAL` to maybe it stand out and allow for easier grep. 

We should also print the expected count: `${times}/{$retries}`

djsly: Or `ERROR: Exhausted all retries (${times}/{$retries}), forcing a failure`

vhdbuilder/scripts/automate_release_notes.sh
```diff
@@ -33,6 +33,10 @@ generate_release_notes() {
         else
             go run vhdbuilder/release-notes/autonotes/main.go --build $build_id --date $image_version --include ${included_skus%?}
         fi
+        if [ $? -ne 0 ]; then
```
djsly, line 36: with `set -e` configured, will this ever be reached ?

vhdbuilder/scripts/automate_release_notes.sh
```diff
@@ -33,6 +33,10 @@ generate_release_notes() {
...
+            git clean -df
```
djsly, line 37: Which previous git command are we cleaning after ? I cannot find the location nor rational of this clean logic.  Maybe it's worth adding a comment

vhdbuilder/scripts/automate_release.sh
```diff
@@ -0,0 +1,57 @@
...
+    echo "SIG release successfully created for VHD build with ID: $VHD_BUILD_ID"
+    echo "release URL: https://msazure.visualstudio.com/CloudNativeCompute/_releaseProgress?_a=release-pipeline-progress&releaseId=$RELEASE_ID"
+}
+
+if [ -z "$VHD_BUILD_ID" ]; then
+    echo "VHD_BUILD_ID must be set to run automated EV2 artifact + release trigger"
+    exit 1
+fi
+
+# retrycmd_if_failure 3 60 trigger_ev2_artifacts || exit $?
```
djsly, line 56: remove comment once confirmed

vhdbuilder/scripts/automate_release.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+set -euxo pipefail
+source vhdbuilder/scripts/automate_helpers.sh
+
+EV2_ARTIFACT_PIPELINE_ID="319030"
```
djsly, line 5: can we add a URL of the matching pipeline in ADO, for easier lookup

vhdbuilder/scripts/automate_release.sh
```diff
@@ -0,0 +1,57 @@
...
+        echo "EV2 artifact build failed for VHD build with ID: $VHD_BUILD_ID, failed build ID: $EV2_BUILD_ID"
+        return 1
+    fi
+    echo "EV2 artifacts successfully built for VHD build with ID: $VHD_BUILD_ID, EV2 build ID: $EV2_BUILD_ID"
+}
+
+create_release() {
+    echo "creating SIG release for VHD with build ID: $VHD_BUILD_ID"
+
+    EV2_BUILD_ID="94988139"
```
djsly, line 37: can we add a URL in comment to allow for quick lookup

djsly (changes requested):

.pipelines/.vsts-vhd-automation.yaml
cameronmeissner: let me look into this, I can't actually remember the reason why we ended up needing to do adhoc setup like we're doing now

.pipelines/.vsts-vhd-automation.yaml
cameronmeissner: the login in this step propagates to the following steps within the same job

cameronmeissner: and no, `--verbose` does not print any secrets

vhdbuilder/scripts/automate_helpers.sh
cameronmeissner: I just copy-pasta'd what we already have in our CSE scripts, will like into enhancing

vhdbuilder/scripts/automate_release_notes.sh
cameronmeissner: good catch

vhdbuilder/scripts/automate_release.sh
cameronmeissner: this is currently just for testing

.pipelines/.vsts-vhd-automation.yaml
cameronmeissner: I think we should consider using a 1ES-managed custom image - that way we can specify the dependencies we want, while also getting automatic updates

right now I believe we're using a static 2004 base image

cameronmeissner: we also might want to consider switching over to azure linux as well

cameronmeissner: so I was wrong, we are actually using a 2004 image that gets updated by CloudTest, just verified that the image already has the correct go version we need, and thus we're able to get rid of the step altogether

vhdbuilder/scripts/automate_release_notes.sh
cameronmeissner: fixed

vhdbuilder/scripts/automate_release.sh
cameronmeissner: im actually not sure adding a URL here makes sense since it's ADO-internal, and this is technically OSS...

vhdbuilder/scripts/automate_helpers.sh
cameronmeissner: fixed

vhdbuilder/scripts/automate_release.sh
djsly: good call. I keep forgetting this is Public...

vhdbuilder/scripts/automate_release.sh
```diff
@@ -32,7 +32,7 @@ trigger_ev2_artifacts() {
 
 create_release() {
     echo "creating SIG release for VHD with build ID: $VHD_BUILD_ID"
-    API_ENDPOINT="https://${ADO_ORG}.vsrm.visualstudio.com/${ADO_PROJECT}/_apis/Release/releases?api-version=5.1"
+    API_ENDPOINT="https://msazure.vsrm.visualstudio.com/CloudNativeCompute/_apis/Release/releases?api-version=5.1"
```
djsly: I think the idea was not to make those public ?

cameronmeissner: it's already all over the place lol, though I can get rid of them from here if we want

vhdbuilder/scripts/automate_release.sh
```diff
@@ -28,18 +32,39 @@ trigger_ev2_artifacts() {
 
 create_release() {
     echo "creating SIG release for VHD with build ID: $VHD_BUILD_ID"
-
+    API_ENDPOINT="https://${ADO_ORG}.vsrm.visualstudio.com/${ADO_PROJECT}/_apis/Release/releases?api-version=5.1"
     EV2_BUILD_ID="94988139"
+    EV2_BUILD_NAME="20240602.7"
+
+    # RELEASE_ID=$(az pipelines release create --detect true \
```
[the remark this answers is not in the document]
cameronmeissner: we run into this issue: https://github.com/Azure/azure-devops-cli-extension/issues/1257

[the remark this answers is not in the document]
cameronmeissner: added a comment in code to explain

vhdbuilder/scripts/automate_release.sh
```diff
@@ -0,0 +1,60 @@
...
+    EV2_BUILD_NUMBER=$(echo "$RESPONSE" | jq -r '.buildNumber')
+    STATUS="$(az pipelines runs show --id $EV2_BUILD_ID | jq -r '.status')"
+
+    while [ "${STATUS,,}" == "notstarted" ] || [ "${STATUS,,}" == "inprogress" ]; do
+        echo "EV2 artifact build $EV2_BUILD_ID is still in-progress..."
+        sleep 60
+        STATUS="$(az pipelines runs show --id $EV2_BUILD_ID | jq -r '.status')"
+    done
+
+    if [ "${STATUS,,}" != "completed" ]; then
```
anujmaheshwari1, line 29: status is completed even for failed builds:

eg https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=94897542&view=results

➜  ~ az pipelines runs show --id 94897542 | jq -r '.status'
completed

➜  ~ az pipelines runs show --id 94897542 | jq -r '.result'
failed

so we'd end up creating SIG releases even for failed ev2 artifact builds. although rare, I think we should still change this evaluation condition to avoid getting in this state ever.

something like:

if status != completed or result == failed
  exit with error
fi

vhdbuilder/scripts/automate_release.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "creating SIG release for VHD with build ID: $VHD_BUILD_ID"
+
+    # we do it this way for now since the devops CLI extension doesn't support what we need: https://github.com/Azure/azure-devops-cli-extension/issues/1257
+    echo "sending POST request to $RELEASE_API_ENDPOINT"
+    REQUEST_BODY="{'artifacts': [{'alias': '$RELEASE_EV2_ARTIFACTS_ALIAS_NAME', 'instanceReference': {'id': '$EV2_BUILD_ID', 'name': '$EV2_BUILD_NUMBER'}}], 'definitionId': $SIG_RELEASE_PIPELINE_ID}"
+    RESPONSE=$(curl -X POST -H "Authorization: Basic $(echo -n ":$ADO_PAT" | base64)" -H "Content-Type: application/json" -d "$REQUEST_BODY" "$RELEASE_API_ENDPOINT")
+    
+    RELEASE_ID=$(echo "$RESPONSE" | jq -r '.id')
+    echo "SIG release successfully created for VHD build with ID: $VHD_BUILD_ID"
+    echo "release URL: https://msazure.visualstudio.com/CloudNativeCompute/_releaseProgress?_a=release-pipeline-progress&releaseId=$RELEASE_ID"
```
anujmaheshwari1, line 46: nit: I wonder if its possible to parse out the release approval ev2 link as well from the API? that way, it reduces another click for the release manager.

vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -56,29 +47,19 @@ cut_official_branch() {
     # Image version format: YYYYMM.DD.revision
     # Official branch format: official/vYYYYMMDD
     # Official tag format: v0.YYYYMMDD.revision
-    parsed_image_version="$(echo -n "${new_image_version}" | head -c-1 | tr -d .)"
+    parsed_image_version="$(echo -n "${NEW_IMAGE_VERSION}" | head -c-1 | tr -d .)"
     official_branch_name="official/v${parsed_image_version}"
     official_tag="v0.${parsed_image_version}.0"
-    final_commit_hash=""
-    for build_id in $build_ids; do
```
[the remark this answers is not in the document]
cameronmeissner: I'll add a note about this

vhdbuilder/scripts/automate_release.sh
cameronmeissner: I like the way you think

vhdbuilder/scripts/automate_release.sh
cameronmeissner: lol I completely missed that

vhdbuilder/scripts/automate_release.sh
cameronmeissner: unfortunately it isn't included in the response JSON - I'll create a follow-up item to investigate if this is possible some other way

vhdbuilder/scripts/automate_release.sh
cameronmeissner: fixed

anujmaheshwari1 (approved):
lgtm, thanks for this amazing change that will make our lives easier by at least a few lesser clicks!

djsly (approved):

PR #4464, opened 2024-06-03
cleanup: add kube-proxy definition to component.json

jason1028kr:
**What type of PR is this?**
/kind cleanup

To cleanup vhd dependendies management

Fixes #

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```
none
```


schemas/manifest.cue
```diff
@@ -25,27 +25,41 @@
...
+}
+
+#kube_proxy_dep: {
+    downloadURL: string
+    amd64OnlyVersions: [...string]
+    multiArchVersions: [...string]
+}
+
+#kube_proxy: {
+    dockerKubeProxyImages: string | null
```
djsly, line 51: what's the difference between dockerkuzbeproxy and ContainerdKubeProxyInages

schemas/manifest.cue
```diff
@@ -25,27 +25,41 @@
...
 }
 
 // root object schema enforced against manifest.json
 #root: {
-	runc:                       #runc
-	containerd:                 #containerd
-	[string]:                   #dep
+    runc:                                       #runc
+    containerd:                                 #containerd
+    "kube-proxy":                               #kube_proxy
```
djsly, line 61: why is this one a string ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -448,9 +448,9 @@ rm -r /var/log/azure/Microsoft.Azure.Extensions.CustomScript || exit 1
 # So keeping as many versions as we can - those unsupported version can be removed when we don't have enough space
 # NOTE that we keep multiple files per k8s patch version as kubeproxy version is decided by CCP.
 
-# kube-proxy regular versions >=v1.17.0  hotfixes versions >= 20211009 are 'multi-arch'. All versions in kube-proxy-images.json are 'multi-arch' version now.
+# kube-proxy regular versions >=v1.17.0  hotfixes versions >= 20211009 are 'multi-arch'. All versions in manifest.json are 'multi-arch' version now.
 
-KUBE_PROXY_IMAGE_VERSIONS=$(jq -r '.containerdKubeProxyImages.ContainerImages[0].multiArchVersions[]' <"$THIS_DIR/kube-proxy-images.json")
+KUBE_PROXY_IMAGE_VERSIONS=$(jq -r '."kube-proxy".containerdKubeProxyImages.ContainerImages[0].multiArchVersions[]' <"$THIS_DIR/manifest.json")
```
djsly, line 453: maybe we dint need the double quote if they are removed from the key

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -348,7 +347,7 @@ testKubeProxyImagesPulled() {
   test="testKubeProxyImagesPulled"
   echo "$test:Start"
   containerRuntime=$1
-  containerdKubeProxyImages=$(jq .containerdKubeProxyImages <${KUBE_PROXY_IMAGES_FILEPATH})
+  containerdKubeProxyImages=$(jq .\"kube-proxy\".containerdKubeProxyImages <${MANIFEST_FILEPATH})
```
djsly, line 350: maybe we dont need the double quote if they are removed from the key

parts/linux/cloud-init/artifacts/manifest.json
```diff
@@ -21,6 +21,32 @@
             "default": "1.1.12"
         }
     },
+    "kube-proxy": {
```
djsly, line 24: why do we need to define the metadata in manifest.json and manifest.cue ? shouldnt this be in a single place ? and the `.json` file gets generated at build time ?

djsly (changes requested):
left inline comments

parts/linux/cloud-init/artifacts/manifest.json
jason1028kr: I believe this is the current behavior with `make generate`.  If we are going to use .cue file only for validation, I believe this will be addressed.

schemas/manifest.cue
jason1028kr: cue export command wasn't parsing the fiile correctly when dash is included in the key. I was using dash for consistency, but if you prefer I can change it to camelcase to remove the double quote. Let me know if you think that's preferred. 

Same issue with using jq with dashed key (below comments).

vhdbuilder/packer/install-dependencies.sh
jason1028kr: I had to wrap it around double quote due to jq issue with dashed key

jason1028kr:
Will change this pr to move things over to `component.json`

schemas/manifest.cue
djsly: I would opt to keep it consistent and remove the `-` to avoid quoting

cameronmeissner:
can we close this if we're gonna use components.cue/json instead?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -360,6 +360,29 @@
       "multiArchVersions": [
         "v0.7.1"
       ]
+    },
+    {
+      "downloadURL": "mcr.microsoft.com/oss/kubernetes/kube-proxy:v*",
+      "amd64OnlyVersions": [],
+      "multiArchVersions": [
```
djsly, line 367: what's the rule for those version to be all cached ?  we will need to find a way to ensure we only cached what is really needed.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -363,14 +363,38 @@ for imageToBePulled in ${ContainerImages[*]}; do
...
+# So keeping as many versions as we can - those unsupported version can be removed when we don't have enough space
+# NOTE that we keep multiple files per k8s patch version as kubeproxy version is decided by CCP.
+
+# kube-proxy regular versions >=v1.17.0  hotfixes versions >= 20211009 are 'multi-arch'. All versions in kube-proxy-images.json are 'multi-arch' version now.
+KUBE_PROXY=$(jq '.ContainerImages[] | select(.downloadURL | contains("kube-proxy"))' $COMPONENTS_FILEPATH)
+KUBE_PROXY_IMAGE=$(echo $KUBE_PROXY | jq -r .downloadURL)
+KUBE_PROXY_IMAGE_VERSIONS=$(echo $KUBE_PROXY | jq -r .multiArchVersions)
+for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
+  CONTAINER_IMAGE=$(string_replace $KUBE_PROXY_IMAGE $KUBE_PROXY_IMAGE_VERSION)
+  ctr --namespace k8s.io run --rm ${CONTAINER_IMAGE} checkTask /bin/sh -c "iptables --version" | grep -v nf_tables && echo "kube-proxy contains no nf_tables"
```
djsly, line 388: I dont understand what the purpose of this line.  Also why isn't this logic added at the top in the main version loop ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -363,14 +363,38 @@ for imageToBePulled in ${ContainerImages[*]}; do
     CONTAINER_IMAGE=$(string_replace $downloadURL $version)
     pullContainerImage ${cliTool} ${CONTAINER_IMAGE} &
     image_pids+=($!)
-    echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
+
+    # kube-proxy install's logged after checking for nf_tables
```
djsly, line 367:
```suggestion
    if [[ ${CONTAINER_IMAGE} != *"kube-proxy"* ]]; then
       ctr --namespace k8s.io run --rm ${CONTAINER_IMAGE} checkTask /bin/sh -c        "iptables --version" | grep -v nf_tables && echo "kube-proxy contains no        nf_tables"
       fi
       echo " - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
```

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -464,34 +488,7 @@ fi
...
-wait ${kube_proxy_pids[@]} # Wait for all parallel pulls to finish
-
-for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
-  CONTAINER_IMAGE="mcr.microsoft.com/oss/kubernetes/kube-proxy:v${KUBE_PROXY_IMAGE_VERSION}"
-  ctr --namespace k8s.io run --rm ${CONTAINER_IMAGE} checkTask /bin/sh -c "iptables --version" | grep -v nf_tables && echo "kube-proxy contains no nf_tables"
-
-  # shellcheck disable=SC2181
-  echo "  - ${CONTAINER_IMAGE}" >>${VHD_LOGS_FILEPATH}
-done
-stop_watch $capture_time "Configure Telemetry, Create Logging Directory, Kube-proxy" false
```
djsly, line 494: should we keep the Capture Time around the new Image Pull logic ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -464,34 +488,7 @@ fi
...
-
-KUBE_PROXY_IMAGE_VERSIONS=$(jq -r '.containerdKubeProxyImages.ContainerImages[0].multiArchVersions[]' <"$THIS_DIR/kube-proxy-images.json")
-
-declare -a kube_proxy_pids=()
-
-for KUBE_PROXY_IMAGE_VERSION in ${KUBE_PROXY_IMAGE_VERSIONS}; do
-  CONTAINER_IMAGE="mcr.microsoft.com/oss/kubernetes/kube-proxy:v${KUBE_PROXY_IMAGE_VERSION}"
-  pullContainerImage ${cliTool} ${CONTAINER_IMAGE} &
-  kube_proxy_pids+=($!)
-  while [[ $(jobs -p | wc -l) -ge $parallel_container_image_pull_limit ]]; do
```
djsly, line 481: should we add the parallel image Pull logic to the new main pulling logic ?

djsly (changes requested):

vhdbuilder/packer/install-dependencies.sh
jason1028kr: Checking with the team on the purpose. Since it assumes kube-proxy's installed, I am guessing this goes after wait statement

jason1028kr: I can remove this line, since it is not getting persisted in the vhd log file

vhdbuilder/packer/install-dependencies.sh
jason1028kr: Goal was to make kube-proxy part of the main containerImage loop, which already does parallel pull.

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -381,6 +381,29 @@
       "multiArchVersions": [
         "v0.7.1"
       ]
+    },
+    {
+      "downloadURL": "mcr.microsoft.com/oss/kubernetes/kube-proxy:*",
+      "amd64OnlyVersions": [],
+      "multiArchVersions": [
```
cameronmeissner, line 388: I know @Devinwong is working on a new schema, should we wait for that to be merged and then realign?

jason1028kr: yep! I'll close this one.

PR #4475, opened 2024-06-04
fix: add vhd-publishing-info.json to gitignore 

anujmaheshwari1:
**What type of PR is this?**
This is so that we never accidentally commit the json file manually or through automation, that can leak secrets.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ bin/
 .env
 .coverprofile
 coverage.txt
-publishing-info.json
+vhd-publishing-info.json
 
 test/junit/
 test/e2e/kubernetes/junit.xml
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -11,6 +11,7 @@ bin/
 .env
 .coverprofile
 coverage.txt
+vhd-publishing-info.json
```
cameronmeissner, line 14: will this cover any and all sub-dirs as well?

cameronmeissner (approved):

.gitignore
anujmaheshwari1: addressed

PR #4487, opened 2024-06-06
test: use resource tags to determine image version for e2e tests

r2k1:
**What type of PR is this?**

/kind test

Instead of using a hard-coded VHD image version, pull the latest image with the specified tag.

This change should simplify a few things:

1) Eliminate the need to manually update base_vhd_catalog.json with the latest image version used for testing.
2) Remove the dependency on ADO. Currently, an ADO PAT is required to fetch the image version. PAT has a limited lifetime, which can be annoying when running tests locally. This results in one less external dependency.
3) One step away from running e2e test after VHD builds.


e2e/const.go
```diff
@@ -1,4 +1,4 @@
-package e2e_test
+package e2e
```
timmy-wright, line 1: This was annoying me too. Glad you fixed it!

r2k1: For some reason it made Goland going nuts. Probably a bug, but I couldn't find other solution.

e2e/scenario/base_vhd_catalog.json
```diff
@@ -1,41 +1,48 @@
 {
     "ubuntu1804": {
         "gen2containerd": {
-            "artifactName": "1804-gen2-containerd",
-            "resourceId": "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2/versions/1.1716223055.6413"
+            "versionTagName": "branch",
+            "versionTagValue": "r2k1/e2e-update",
```
r2k1, line 5: alternatively, I can make a single variable for all images.

Set it default to master, with an option to override though ENV variable.
Like `TAG=buildId:12346`. (And with optional fallback to master if the version doesn't exist)

e2e/e2e-local.sh
```diff
@@ -2,16 +2,8 @@
...
-: "${BUILD_ID:=local}"
-: "${SUBSCRIPTION_ID:=8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8}" #Azure Container Service - Test Subscription
-: "${LOCATION:=eastus}"
-: "${AZURE_TENANT_ID:=72f988bf-86f1-41af-91ab-2d7cd011db47}"
 : "${TIMEOUT:=90m}"
 : "${PARALLEL:=100}"
 
-export SUBSCRIPTION_ID
-export LOCATION
-export AZURE_TENANT_ID
```
r2k1, line 14: I like my debugger. I want to run tests without shell script.

So, I moved it as defaults in go config

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+
+func findLatestImageWithTag(imageID, tagName, tagValue string) (VHDResourceID, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), time.Minute)
+	defer cancel()
+	image, err := parseImageID(imageID)
+	if err != nil {
+		return "", err
+	}
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
Devinwong, line 120: I guess this is the magic why we can get rid of the PAT. What pre-requisites do we need to get the cred from this function?

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+	// of an image version in a shared image gallery.
+	resourceID               VHDResourceID
+	buildResourceID          VHDResourceID
+	fetchBuildResourceIDOnce sync.Once
+	fetchResourceIDOnce      sync.Once
+}
+
+var (
+	VHDUbuntu1804Gen2Containerd = &VHD{
+		ImageID:         "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2",
```
Devinwong, line 30: nit: can we declare the template string `/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/` as a constant? For easier maintenance purpose.

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+func (v *VHD) ResourceID() VHDResourceID {
+	if v.BuildResourceID() != "" {
+		return v.BuildResourceID()
+	}
+	return v.NonBuildResourceID()
+}
+
+func (v *VHD) BuildResourceID() VHDResourceID {
+	v.fetchBuildResourceIDOnce.Do(func() {
+		v.buildResourceID, _ = findLatestImageWithTag(v.ImageID, "buildId", VHDBuildID)
```
Devinwong, line 95: Why do we still need to find the latest Image if a VHDBuildID is already given? Will 1 VHDBuildID has multiple built Images?

r2k1: it will find the latest image with `buildId=1233` tag.

The function will go through all versions to find it.

e2e/config/vhd.go
r2k1: I don't think it will improve maintainability. But it will add extra level of indirection.
When I navigate through the code, I like to be able to find the definition as soon as possible.
I also think it's easier to read when all things related to the image is in a single place and not scattered in the file.

e2e/config/vhd.go
r2k1: This part hasn't been changed much. These credentials are used for every Azure API call. Like accessing SIG Images, creating AKS clusters etc.

The PAT token was used only to fetch artifacts from ADO. To extract a VHD resource id from it.

e2e/config/vhd.go
r2k1: I just reused the function. Finding latest is important when VHDBuildID isn't specified and `branchs=refs/head/master` filter is used.

It's probably also possible to re-run a job and get more than a single VHD.

e2e/config/vhd.go
Devinwong: I was just thinking if one day we change the resource group or subscription, we can simply change 1 place instead of searching for all and replace. Anyway, just my own way. Feel free to keep yours🙂

Devinwong:
Do we have any passed test pipeline with this PR? I guess you do. Could you please link it too?

Devinwong (approved):

e2e/config/vhd.go
r2k1: I generally try to avoid this approach (In some cases it makes more sense though). Find and replace is trivial operation in modern editors.
But trying to mentally execute a piece of code is an extra burden every reader will need to perform.

r2k1:
I won't be able to merge the PR unless every check is green.

Here is an example of a tests with VHD_BUILD_ID

1) https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=96406242&view=results
No VHD image uses in tests has been build. So, no tests are run. 
2) https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=96406600&view=logs&j=df77c11a-d70e-5fda-f10b-e543dae019aa&t=42160a71-98d9-5950-cc5c-ad5fea7cc94c . All images were rebuilt, not tests are skipped.

e2e/e2e-local.sh
djsly: Could we create a task to ensure we create a Wiki page/ runbook on how to run e2e locally ?  since you have been successfully running them during this massive learning experience, It might be the best time to do a brain dump on all the tips and tricks you discovered when comes time to running e2e on your laptop!

e2e/scenario/scenario.go
```diff
@@ -1,101 +1,58 @@
...
+		azurelinuxv2CustomSysctls(),
+		azurelinuxv2Wasm(),
+		azurelinuxv2_azurecni(),
+		azurelinuxv2gpu_azurecni(),
+		azurelinuxv2ARM64AirGap(),
+		azurelinuxv2AirGap(),
+		azurelinuxv2ChronyRestarts(),
+
+		// block for gpu scenarios
+		ubuntu2204gpu("ubuntu2204-gpu-ncv3", "Standard_NC6s_v3"),
```
djsly, line 54: any reason why two 3 scenarios require 2 properties ?  should we align the logic between all scnenarios ?

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+
+var (
+	VHDUbuntu1804Gen2Containerd = &VHD{
+		ImageID:         "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/1804Gen2",
+		VersionTagName:  "branch",
+		VersionTagValue: "refs/heads/master",
+	}
+	VHDUbuntu2204Gen2Arm64Containerd = &VHD{
+		ImageID:         "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/2204Gen2Arm64",
+		VersionTagName:  "branch",
```
djsly, line 36: could you add a description as to what the expectation of finding a tag with name `branch` would mean ?

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (v *VHD) BuildResourceID() VHDResourceID {
+	v.fetchBuildResourceIDOnce.Do(func() {
+		v.buildResourceID, _ = findLatestImageWithTag(v.ImageID, "buildId", VHDBuildID)
+	})
+	return v.buildResourceID
+}
+
+func (v *VHD) NonBuildResourceID() VHDResourceID {
```
djsly, line 100: I'm not a fan of `NonBuild` but at the same time I cannot find a better term.  Could it be `getResourceIDFromImageTag`

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+// 2. Fetch resource id only once (multiple scenarios can share the VHD)
+// 3. Call fetch method only if tests is scheduled to run (it takes times)
+func (v *VHD) ResourceID() VHDResourceID {
+	if v.BuildResourceID() != "" {
+		return v.BuildResourceID()
+	}
+	return v.NonBuildResourceID()
+}
+
+func (v *VHD) BuildResourceID() VHDResourceID {
```
djsly, line 93: could this be renamed to `getResourceIDFromBuildID`

e2e/config/config.go
```diff
@@ -8,24 +8,23 @@ import (
...
 	KeepVMSS           = strings.EqualFold(os.Getenv("KEEP_VMSS"), "true")
-	Azure              = MustNewAzureClient(mustenv("SUBSCRIPTION_ID"))
+	Azure              = MustNewAzureClient(SubscriptionID)
 )
 
-func mustenv(env string) string {
-	result := os.Getenv(env)
-	if result == "" {
-		panic("missing environment variable: " + env)
+func envDefault(env string, defaultValue string) string {
```
djsly, line 22: I'm surprised Go didn't implement this! hehe. 

for the `val == ""` what if the ENV exists but it contains a default "" value which is valid.  

I have seen implementation using https://pkg.go.dev/os#LookupEnv. 

```
func getEnv(key, fallback string) string {
    if value, ok := os.LookupEnv(key); ok {
        return value
    }
    return fallback
}
```

e2e/config/azure.go
```diff
@@ -20,15 +20,16 @@ import (
...
+	Core                      *azcore.Client
+	VMSS                      *armcompute.VirtualMachineScaleSetsClient
+	VMSSVM                    *armcompute.VirtualMachineScaleSetVMsClient
+	VNet                      *armnetwork.VirtualNetworksClient
+	Resource                  *armresources.Client
+	ResourceGroup             *armresources.ResourceGroupsClient
+	AKS                       *armcontainerservice.ManagedClustersClient
+	SecurityGroup             *armnetwork.SecurityGroupsClient
+	Subnet                    *armnetwork.SubnetsClient
+	GalleryImageVersionClient *armcompute.GalleryImageVersionsClient
```
cameronmeissner, line 32: nit: `GalleryImageVersionsClient`

e2e/config/config.go
cameronmeissner: `LookupEnv` seems nice :)

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return str
+}
+
+// This is ugly, but I wanted to achieve multiple things:
+// 1. If buildID specified use the image from the build, skip scenario if not found
+// 2. Fetch resource id only once (multiple scenarios can share the VHD)
+// 3. Call fetch method only if tests is scheduled to run (it takes times)
+func (v *VHD) ResourceID() VHDResourceID {
+	if v.BuildResourceID() != "" {
```
cameronmeissner, line 87:
```suggestion
	if rid := v.BuildResourceID(); bid != "" {
	      return bid
	}
```

to avoid multiple calls to the same method

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+
+type imageID struct {
+	subscriptionID string
+	resourceGroup  string
+	galleryName    string
+	imageName      string
+}
+
+func parseImageID(resourceID string) (imageID, error) {
+	pattern := `(?i)^/subscriptions/([^/]+)/resourceGroups/([^/]+)/providers/Microsoft\.Compute/galleries/([^/]+)/images/([^/]+)$`
```
cameronmeissner, line 162: regex makes me sad, can we use this instead? https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID

e2e/config/vhd.go
cameronmeissner: maybe `TaggedResourceID`?

e2e/scenario/scenario.go
cameronmeissner: it's specific to GPU scenarios since they're capable of using different VM sizes

I'd be in favor of creating a `GPUScenarioConfig` struct or similar to make this a bit more readable

e2e/scenario/vhd.go
```diff
@@ -1,116 +1,12 @@
...
-	if err := json.Unmarshal([]byte(rawJSON), &catalog); err != nil {
-		panic(err)
-	}
-
-	return catalog
-}
-
 type Manifest struct {
 	Containerd struct {
 		Edge string `json:"edge"`
```
cameronmeissner, line 12: unrelated to these changes, but we should make a note that we'll need to update this based on the consolidation towards components.json

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+	pager := client.NewListByGalleryImagePager(image.resourceGroup, image.galleryName, image.imageName, nil)
+	var latestVersion *armcompute.GalleryImageVersion
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return "", fmt.Errorf("failed to get next page: %v", err)
+		}
+		versions := page.Value
+		for _, version := range versions {
+			tag, ok := version.Tags[tagName]
```
cameronmeissner, line 139: should we check that each `version` is non-nil before attempting to dereference `Tags`?

e2e/config/vhd.go
cameronmeissner, line 95: can we at least log the error value here instead of silencing?

e2e/config/vhd.go
```diff
@@ -0,0 +1,176 @@
...
+		VersionTagName:  "branch",
+		VersionTagValue: "refs/heads/master",
+	}
+	VHDUbuntu2204Gen2ContainerdPrivateKubePkg = &VHD{
+		resourceID: "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/2204Gen2/versions/1.1704411049.2812",
+	}
+	VHDAzureLinuxV2Gen2Arm64 = &VHD{
+		ImageID:         "/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.Compute/galleries/PackerSigGalleryEastUS/images/AzureLinuxV2Gen2Arm64",
+		VersionTagName:  "branch",
+		VersionTagValue: "refs/heads/master",
```
cameronmeissner, line 50: are we planning on letting users use their own branches here? I guess we accomplish that by keeping the build ID option around

e2e/scenario/scenario.go
r2k1: Not something I originally wrote. But I actually like it. There is a minor difference between scenarios and it's obvious from the arguments what's the difference between them.

e2e/config/config.go
r2k1: `LookupEnv ` is important if you want to know that ENV variable is set to an empty string.

In this case, if ENV set to an empty string the fallback won't be used and we will fail something down the line. Is it something you want to achieve?

e2e/config/vhd.go
r2k1: I tried initially. SIG resource id contains more sections than a typical resource. So post-processing with regexes or splitting by `/` is still required after the function call.

e2e/config/vhd.go
r2k1: if azure SDK returns a slice with empty pointers, then something is really wrong with the SDK and we should fix it.

e2e/config/vhd.go
r2k1: Yes, why not. Initially I wanted to use a PR branch name instead of buildID.
But didn't want to change too much and buildIDs seems to work just fine.

I'd just modify this file during development. Maybe there is a better way to configure branch name with ENV variables or git commands. I'll think how to make it flexible if I ever need to use it.

e2e/config/vhd.go
r2k1: This is ugly. But sometimes there is no image for the build. And it's ok. We just ignore such scenarios.

I'm not happy with the code but can't figure out a better way to achieve it.

e2e/config/vhd.go
anujmaheshwari1: I do somewhat agree with Devin here, at least the subscription ID, gallery name and image definition can be high level consts defined earlier in the file and then propagated downwards. for most of the cases, the first 2 don't even change across different definitions. 

but say in a case where we update the gallery name or move to a different subscriptions, this change is just harder to make.

e2e/config/config.go
r2k1: Sorry, I misread the original comment. I don't think we have a single case where `""` is a valid value.
In general, I think it's bad idea to support empty string.

How would like to clarify it? A comment?
The test will fail the someone does it.

e2e/config/vhd.go
r2k1: I think I found a better solution. I added 3 new config variables with defaults:

```
SIG_VERSION_TAG_NAME=branch
SIG_VERSION_TAG_VALUE=refs/heads/master
IGNORE_SCENARIOS_WITH_MISSING_VHD=false
```

So could remove the buildId logic from the code.

r2k1: As a side-effect you can specify the ENV variables once during development.

Trigger the build. And re-run e2e tests locally without any addition changes.

e2e/config/config.go
djsly: I would try to find a method name that provides all of the details 
Something like that: 
“getNonEmptyEnv”

e2e/config/vhd.go
r2k1: Not the hill I'm willing to die on. I removed the duplication.

But I think once in 3 years performing a single find and replace to change the subscription is trivial.
On other hand now it's not obvious anymore which image is used. And it's not searchable in the code anymore.

e2e/config/config.go
r2k1: `envDefault` =>`getEnvWithDefaultIfEmpty`?

e2e/config/vhd.go
r2k1: I think I found a better approach to it

e2e/config/vhd.go
r2k1: Made this configurable. With default to `branch=refs/heads/master`

.pipelines/e2e.yaml
```diff
@@ -42,14 +41,21 @@ jobs:
       export PATH="/usr/local/go/bin:$PATH"
       go version
 
-      export BUILD_ID=$(Build.BuildId)
+      if [ -z "${VHD_BUILD_ID}" ]; then
```
cameronmeissner, line 44: did you mean `-n` here?

e2e/config/vhd.go
```diff
@@ -0,0 +1,125 @@
...
+var (
+	VHDUbuntu1804Gen2Containerd               = newVHDResourceIDFetcher(imageGallery + "1804Gen2")
+	VHDUbuntu2204Gen2Arm64Containerd          = newVHDResourceIDFetcher(imageGallery + "2204Gen2Arm64")
+	VHDUbuntu2204Gen2Containerd               = newVHDResourceIDFetcher(imageGallery + "2204Gen2")
+	VHDAzureLinuxV2Gen2Arm64                  = newVHDResourceIDFetcher(imageGallery + "AzureLinuxV2Gen2Arm64")
+	VHDAzureLinuxV2Gen2                       = newVHDResourceIDFetcher(imageGallery + "AzureLinuxV2Gen2")
+	VHDCBLMarinerV2Gen2Arm64                  = newVHDResourceIDFetcher(imageGallery + "CBLMarinerV2Gen2Arm64")
+	VHDCBLMarinerV2Gen2                       = newVHDResourceIDFetcher(imageGallery + "CBLMarinerV2Gen2")
+	VHDUbuntu2204Gen2ContainerdPrivateKubePkg = func() (VHDResourceID, error) {
+		return imageGallery + "2204Gen2PrivateKubePkg", nil
```
cameronmeissner, line 26: `2204Gen2PrivateKubePkg` isn't a real image definition

e2e/scenario/vhd.go
cameronmeissner: yeah no worries, more-so just a note for myself and @Devinwong

e2e/suite_test.go
```diff
@@ -40,54 +33,74 @@ func Test_All(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	scenarios := scenario.AllScenarios()
 	if err := createMissingClusters(ctx, r, scenarios, &clusterConfigs); err != nil {
 		t.Fatal(err)
 	}
 
 	for _, e2eScenario := range scenarios {
-		e2eScenario := e2eScenario
```
cameronmeissner, line 48: I think we need the copy here since the loop counter is being used within a closure, otherwise you'll get closures that always act on the last element in the slice

I can see this behavior showing up in the most recent run off this branch: 

```
=== CONT  Test_All/ubuntu1804
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/marinerv2-custom-sysctls
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/ubuntu2204-gpu-a10
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/ubuntu2204-gpu-a100
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-chrony-restarts
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/ubuntu2204-gpu-ncv3
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-arm64-airgap
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-airgap
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-gpu-azurecni
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-azurecni
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-wasm
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-custom-sysctls
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-gpu
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/marinerv2-gpu
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/ubuntu2204-artifact-streaming
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/azurelinuxv2-arm64
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image
=== CONT  Test_All/ubuntu2204ContainerdVersion
    suite_test.go:60: skipping scenario "ubuntu2204-gpu-a10": could not find image

```

cameronmeissner: ah, this is interesting: https://go.dev/doc/faq#closures_and_goroutines

looks like this was fixed in 1.22, I guess you might as well bump us up while you're at it :)

e2e/suite_test.go
```diff
@@ -40,54 +33,74 @@ func Test_All(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	scenarios := scenario.AllScenarios()
 	if err := createMissingClusters(ctx, r, scenarios, &clusterConfigs); err != nil {
```
cameronmeissner, line 37: I guess if we do it this way then the suite will attempt to create testing clusters for scenarios that we might not actually be running

I'm not necessarily opposed to this since in theory that would speed up subsequent runs either from the same user if they decide to later run the scenarios they originally skipped, or if other users run those scenarios as well

e2e/scenario/vhd.go
Devinwong: Noted!

e2e/config/vhd.go
cameronmeissner: I think we can probably just have this function return a hard-coded resource ID for now to make things work

e2e/config/azure.go
cameronmeissner, line 32: where are we actually using this client? I see we also create one within `findLatestImageWithTag`, which seems to be the only place we're using any object of this type

e2e/config/vhd.go
```diff
@@ -0,0 +1,125 @@
...
+	str := string(id)
+	if strings.Contains(str, sep) && !strings.HasSuffix(str, sep) {
+		return strings.Split(str, sep)[1]
+	}
+	return str
+}
+
+var ErrNotFound = fmt.Errorf("not found")
+
+func findLatestImageWithTag(imageID, tagName, tagValue string) (VHDResourceID, error) {
```
cameronmeissner, line 61: what would happen if we haven't run the VHD build pipeline in MSFT for at least 3 days and there are no longer any images that have the master branch tag due to the GC?

I guess we'd need to always keep at least one set of images locked like we currently do today so we can guarantee that the test suite is always runnable by itself

cameronmeissner: one thing to note is that with the new duplication of AME images into MSFT galleries for RP e2e's, things will get a lot easier

once that effort is done (which it basically it is at this point, the galleries exist with actual images already), we can simplify all this logic to something like:

```
if tag name and value are specified:
  if tag is `buildId`
    take images from `PackerSigGalleryEastUS` that match the tag and value
 otherwise:
    take the latest images from `PackerSigGalleryEastUS` that match the tag and value
otherwise:
  take latest images from the new MSFT galleries containing copies of official AME images
```

the new galleries are named such as `AKSUbuntu`, `AKSAzureLinux`, `AKSCBLMariner`, and live within a separate DevInfra subscription within the MSFT tenant, which we'll most likely need to have access to via service principal

cameronmeissner (changes requested):
might want to wait on this until we can access the new MSFT galleries to use fall-backs in case tagged images don't exist

e2e/config/vhd.go
r2k1: I think something like this reflect better my intent:

```
if tag name and value are specified:
    take the latest images from `PackerSigGalleryEastUS` that match the tag and value. Skip tests if it doesn't exist
otherwise:
  take latest images from the new MSFT galleries containing copies of official AME images
```

e2e/suite_test.go
r2k1: I think I can do something similar to the way I load last version for SIG. Create a cluster in a lazy way and only once.

But I'd rather fix it in a separate PR. This one already grew beyond my intent.

e2e/suite_test.go
r2k1: ah, good catch. I run 1.22 locally and didn't realise ADO uses a different version.

e2e/config/vhd.go
r2k1: ah, good catch. Copilot "helped" me there.

.pipelines/e2e.yaml
r2k1: `-z` will also catch an empty string. I don't have a strong opinion on which one is better in this case. But it feels a slightly safer

r2k1:
Images are tagged through packer. If packer config hasn't changed, then no extra change should be necessary.

e2e/config/azure.go
r2k1: yep, this is the only usage of the client

e2e/config/vhd.go
r2k1: > what would happen if we haven't run the VHD build pipeline in MSFT for at least 3 days and there are no longer any images that have the master branch tag due to the GC?

How does it work currently? Wouldn't we have the same issue if GC deletes versions referenced in the e2e tests? What's the difference?

e2e/suite_test.go
r2k1: I wrongly assumed that passing it as a function argument would be enough to avoid the issue.
But instead, it should be copied before a goroutine is created. And it's not obvious when it happens in the tests.

e2e/config/vhd.go
cameronmeissner: today as a part of our "update agentbaker e2e VHD" process, we add a delete lock to the VHD's themselves to ensure they're protected from GC

e2e/config/vhd.go
```diff
@@ -0,0 +1,132 @@
...
+var (
+	VHDUbuntu1804Gen2Containerd               = newVHDResourceIDFetcher(imageGallery + "1804Gen2")
+	VHDUbuntu2204Gen2Arm64Containerd          = newVHDResourceIDFetcher(imageGallery + "2204Gen2Arm64")
+	VHDUbuntu2204Gen2Containerd               = newVHDResourceIDFetcher(imageGallery + "2204Gen2")
+	VHDAzureLinuxV2Gen2Arm64                  = newVHDResourceIDFetcher(imageGallery + "AzureLinuxV2Gen2Arm64")
+	VHDAzureLinuxV2Gen2                       = newVHDResourceIDFetcher(imageGallery + "AzureLinuxV2Gen2")
+	VHDCBLMarinerV2Gen2Arm64                  = newVHDResourceIDFetcher(imageGallery + "CBLMarinerV2Gen2Arm64")
+	VHDCBLMarinerV2Gen2                       = newVHDResourceIDFetcher(imageGallery + "CBLMarinerV2Gen2")
+	VHDUbuntu2204Gen2ContainerdPrivateKubePkg = func() (VHDResourceID, error) {
+		return imageGallery + "2204Gen2/versions/1.1704411049.2812", nil
```
cameronmeissner, line 27: can we add a comment here explaining that:

`2204Gen2/versions/1.1704411049.2812 is a particular 2204Gen2 image originally built with private packages, if we ever want to update this then we'd need to run a new VHD build using private package overrides"

cameronmeissner (approved):

cameronmeissner:
discussed offline that we can follow-up when we're ready to implementing fall-back logic using the new MSFT galleries, in the meantime we'll just be sure to keep at least one set of images built off of master locked and ready to go

djsly (approved):

djsly:
@r2k1 congrats on the merge!

PR #4496, opened 2024-06-11
feat: support Windows private k8s packages

AbelHu:
**What type of PR is this?**
/kind feature

support Windows private k8s packages

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/configure-windows-vhd.ps1
```diff
@@ -286,14 +286,24 @@ function Get-PrivatePackagesToCacheOnVHD {
...
         $urls = $env:WindowsPrivatePackagesURL.Split(",")
         foreach ($url in $urls) {
             $fileName = [IO.Path]::GetFileName($url.Split("?")[0])
             $dest = [IO.Path]::Combine($dir, $fileName)
 
             Write-Log "Downloading a private package to $dest"
             Download-FileWithAzCopy -URL $URL -Dest $dest
+
+            # Example: v1.29.2-hotfix.2024101-1int.zip
+            $version = $fileName.Split('-')[0].SubString(1)
```
bcho, line 300: nit: do we need to check if the fileName contains `-` before accessing `[0]`? Not sure what's the behavior of accessing a zero length array in powershell.

AbelHu: The filename should follow the format of the example. If `$fileName` does not contain `-`, `$fileName.Split('-')[0].SubString(1)` will return the full file name without the first character. We can check the build logs to confirm whether it is set correctly.

bcho (approved):
LGTM

vhdbuilder/packer/configure-windows-vhd.ps1
serbrech: Have we explored using pester to unit test the powershell in agentbaker?

https://github.com/pester/Pester

This is quite dense in logic, and the consequencers of a bug are quite dramatic on the end user experience

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -182,7 +182,18 @@ function Get-KubePackage {
         [Parameter(Mandatory = $true)][string]
         $KubeBinariesSASURL
     )
-    Logs-To-Event -TaskName "AKS.WindowsCSE.DownloadKubletBinaries" -TaskMessage "Start to download kubelet binaries and unzip. KubeBinariesPackageSASURL: $global:KubeBinariesPackageSASURL"
+    $mappingFile = [Io.path]::Combine($global:CacheDir, "private-packages\mapping.json")
```
serbrech, line 185: > We can check the build logs to confirm whether it is set correctly.

Have we explored using pester to unit test the powershell in agentbaker?
https://github.com/pester/Pester

This is quite dense in logic, and the consequences of a bug are quite dramatic on the end user experience

serbrech: looks like we do use pester in other parts?
https://github.com/Azure/AgentBaker/blob/master/parts/windows/windowscsehelper.tests.ps1

AbelHu: Yes. Will add the ut.

AbelHu: Added ut.

serbrech (approved):

staging/cse/windows/kubeletfunc.tests.ps1
```diff
@@ -0,0 +1,65 @@
...
+
+    Context 'mapping file exists' {
+        BeforeEach {
+            Mock Test-Path -MockWith { $true }
+            Mock Get-Content -MockWith {
+                Param(
+                    $Path
+                )
+@'
+                {
```
serbrech, line 31: so much clearer to understand what's happening with a unit test :)
thanks!

vhdbuilder/packer/configure-windows-vhd.ps1
AbelHu: @bcho Added tests too.

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -182,7 +182,18 @@ function Get-KubePackage {
...
         $KubeBinariesSASURL
     )
-    Logs-To-Event -TaskName "AKS.WindowsCSE.DownloadKubletBinaries" -TaskMessage "Start to download kubelet binaries and unzip. KubeBinariesPackageSASURL: $global:KubeBinariesPackageSASURL"
+    $mappingFile = [Io.path]::Combine($global:CacheDir, "private-packages\mapping.json")
+    if (Test-Path $mappingFile) {
+        $urls = @{}
+        (ConvertFrom-Json ((Get-Content $mappingFile -ErrorAction Stop) | Out-String)).psobject.properties | Foreach { $urls[$_.Name] = $_.Value }
+        if ($urls.ContainsKey($global:KubeBinariesVersion)) {
+            Write-Log "Found $global:KubeBinariesVersion in $mappingFile"
+            $KubeBinariesSASURL = $urls[$global:KubeBinariesVersion]
```
liuxintong, line 191: Looks like this script will be executed per node, do we have different urls for different clouds, like mooncake, fairfax and airgap?

staging/cse/windows/kubeletfunc.ps1
liuxintong, line 185: Do we have plan to support the package version mapping mechanism for Linux as well? I understand it might be out of the scope of this PR.

staging/cse/windows/kubeletfunc.ps1
AbelHu: We do not need any urls for different clouds since these packages are cached so it will not be downloaded

staging/cse/windows/kubeletfunc.ps1
AbelHu: Shared this with Linux team to do the same thing.

junjiezhang1997 (approved):

PR #4498, opened 2024-06-11
cleanup: streamline ARM64 VHD build. Reduce discrepancies between builds.

r2k1:
/kind cleanup

Build ARM64 image in a manner similar to other architectures. Reduce discrepancies between builds.

Currently, VHD images for ARM64 are built in the following sequence:

1) Packer builds OS Disk snapshot.
2) OS Disk snapshot is converted to SIG.

During this transition, Azure resource tags are lost.

I think in the past, images didn't push directly to SIG but first to "azure legacy managed images," which didn't support ARM, so a fork (#1547) was created for ARM. Since then, the AMD64 process has been streamlined, but the fork persists.

It should also reduce number of azure service dependencies used in the build pipeline.


djsly (approved):
Awesome cleanup!

chengliangli0918 (approved):

PR #4507, opened 2024-06-12
Xinhl/acr

fseldow:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -957,6 +957,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"ShouldEnableCustomData": func() bool {
 			return !config.DisableCustomData
 		},
+		"GetArtifactCacheContainerRegistryServer": func() string {
```
djsly, line 960: where is this used ?

fseldow: used for NetworkIsolatedCluster. The private acr to pull OCI artifact (mar artifact cached in acr via artifact cache)

fseldow: will it be better to parse this variable from agent baker data model manageCluster?

PR #4512, opened 2024-06-13
test: Simplify e2e test configuration

r2k1:
/kind test

Simplify configuration for e2e.

There is a single instance of configuration, and I don't see a need for a second one (unless we want to write tests for tests). So, I extracted it to a global variable for the ease of use.

It may be not the "clean code", but I think it's much easier to deal with. Few main benefits:
1) Ease of use. I want to use one of the Azure clients when loading a resource ID. Currently, I need to modify the entire chain of functions between the main function and the resourceId function
2) Clarity. There are much less hoops to find where the configuration is coming from. I can go to `config.BuildID` declaration and see straight away that's coming from ENV variable `BUILD_ID`


djsly (approved):

.pipelines/e2e.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
       export BUILD_ID=$(Build.BuildId)
 
       cd e2e
-      go test -timeout ${{ variables.defaultTimeout }}m -v -run Test_All ./
+      go test -parallel 100 -timeout ${{ variables.defaultTimeout }}m -v -run Test_All ./
```
djsly, line 46: could we expose  the parallel count as an external pipeline variable to allow changing outside of the code in case it causes issues in the future

e2e/e2e-local.sh
```diff
@@ -13,4 +13,4 @@ export LOCATION
 export AZURE_TENANT_ID
 
 go version
-go test -timeout $TIMEOUT -v -run Test_All ./
\ No newline at end of file
+go test -parallel 100 -timeout $TIMEOUT -v -run Test_All ./
```
djsly, line 16: can we also allow for the parallel count to be set as an ENV with default 100 is not set.

.pipelines/e2e.yaml
r2k1: added

djsly (approved):

PR #4513, opened 2024-06-13
Fix access level for default Kubelet file discovered with CIS 1.8 benchmark 

djsly:
**What type of PR is this?**

/kind bug
Changes the access level to 600 for the default kubelet file

From Kube-Bench
[FAIL] 4.1.9 If the kubelet config.yaml configuration file is being used validate permissions set to 600 or more restrictive (Automated)

== Remediations node ==
4.1.9 Run the following command (using the config file location identified in the Audit step)
chmod 600 /etc/default/kubelet

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
I think you're missing a `make generate`

Devinwong:
Why do we need to change the access to 600? Is it to tighten the access level?

djsly:
> Why do we need to change the access to 600? Is it to tighten the access level?
From Kube-Bench
[FAIL] 4.1.9 If the kubelet config.yaml configuration file is being used validate permissions set to 600 or more restrictive (Automated)

== Remediations node ==
4.1.9 Run the following command (using the config file location identified in the Audit step)
chmod 600 /etc/default/kubelet

cameronmeissner:
will let abe2e's run

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,7 @@ ensureKubelet() {
     if [ -n "${AZURE_ENVIRONMENT_FILEPATH}" ]; then
         echo "AZURE_ENVIRONMENT_FILEPATH=${AZURE_ENVIRONMENT_FILEPATH}" >> "${KUBELET_DEFAULT_FILE}"
     fi
+    chmod 0600 "${KUBELET_DEFAULT_FILE}"
```
anujmaheshwari1, line 408: i suppose we do this for all files lower down this function but missed it for /etc/default/kubelet

anujmaheshwari1 (approved):

Devinwong (approved):

cameronmeissner:
fixing abe2e's at the moment, then will let it run :)

cameronmeissner (approved):
all green

PR #4518, opened 2024-06-14
build: run e2e tests after VHD build

r2k1:
Run e2e tests in ADO after VHD image build


.pipelines/e2e.yaml
```diff
@@ -3,77 +3,19 @@ trigger: none
...
+      - e2e
+      - parts/linux
+      - pkg/agent
     exclude:
-    - e2e/windows
-    - pkg/agent/datamodel/sig_config*.go # SIG config changes
-    - pkg/agent/datamodel/*.json # SIG version changes
-    - pkg/agent/testdata/AKSWindows* # Windows test data
-    
-variables:
```
djsly, line 20: not needed anymore ?

.pipelines/templates/e2e-template.yaml
```diff
@@ -0,0 +1,51 @@
+jobs:
+  - job: E2E_Tests
+    pool:
+      name: nodesigtest-pool
+    timeoutInMinutes: 90
+    displayName: Run AgentBaker E2E
+    variables:
+      - group: ab-e2e
```
djsly, line 8: which variables are stored in that group ?

.pipelines/templates/e2e-template.yaml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - bash: |
+          az login --identity
+          az account set -s $(AZURE_SUBSCRIPTION_ID)
+        displayName: az login
+      - bash: bash .pipelines/scripts/setup_go.sh
+        displayName: setup go
+      - bash: |
```
djsly, line 16: should we give this step a `displayName` ?

.pipelines/templates/e2e-template.yaml
r2k1: Azure Subscription ID, Resource group etc. Things to create test clusters and download SIG images.

.pipelines/templates/e2e-template.yaml
r2k1: I copied most of the code from another pipeline.

It's confusing, but `displayName` are defined in the bottom of most steps

.pipelines/e2e.yaml
r2k1: it's been moved to e2e-template.yaml, so it can be shared between 2 different pipelines

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -2,991 +2,931 @@ name: $(Date:yyyyMMdd)$(Rev:.r)_$(Build.SourceBranchName)_$(BuildID)
...
-  default: false
-- name: build2204minimalgen2containerd
-  displayName: Build 2204 Minimal Gen2 Containerd
-  type: boolean
-  default: false
-- name: dryrun
-  displayName: Dry run
-  type: boolean
-  default: False
+  - name: build1804containerd
```
cameronmeissner, line 5: do we want all this extra identation?

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1034,107 +972,104 @@ stages:
...
+              echo '##vso[task.setvariable variable=ARCHITECTURE]X86_64'
+              echo '##vso[task.setvariable variable=ENABLE_TRUSTED_LAUNCH]False'
+              echo '##vso[task.setvariable variable=SGX_INSTALL]False'
+            displayName: Setup Build Variables
+          - template: ./templates/.builder-release-template.yaml
+            parameters:
+              artifactName: 2204-minimal-gen2-containerd
+  - stage: Run_E2E_Tests
+    variables:
+      VHD_BUILD_ID: $(Build.BuildId)
```
cameronmeissner, line 1073: let's also pass

`LOCATION: $(PACKER_BUILD_LOCATION)`

.pipelines/.vsts-vhd-builder-release.yaml
r2k1: on `parameters` it's probably a personal preference. Extra indentation is less confusing for me personally.

`stages` has changed. Prior there was multiple stages with 1 job each. Now here is a single stage with multiple jobs in it.

.pipelines/.vsts-vhd-builder-release.yaml
r2k1: added

.pipelines/e2e.yaml
```diff
@@ -3,78 +3,19 @@ trigger: none
...
-      done
-
-      if [ -n "${vmssResourceIds// }" ]; then
-        az resource delete --ids ${vmssResourceIds}
-      fi
-    displayName: ensure vmss deletion
-    condition: always()
- 
- 
+  - template: ./templates/e2e-template.yaml
```
cameronmeissner, line 21: can we add a new boolean parameter as a "feature flag" to determine whether we run the e2e's?

r2k1: Where would you use it?

e2e tests are run after the build. So, a failure there doesn't change much.

djsly: shouldnt the e2e prevent an official release from happening ?   Are we saying that the e2e failures will not present the release of the build artifact ? (for now ?) 

@cameronmeissner  what is your concern ? I'm guessing before the e2e were not executed as part of the build ? I thought this was just ensuring that the e2e would be using the latest images that was just built ? unless you are worried about the fact that e2e now run during PR pipeline ?

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -32,10 +32,9 @@ variables:
   value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
 
 stages:
-  - stage: build_vhd
-    dependsOn: []
+  - stage: build
     jobs:
-    - job: build
+    - job: build-2204-gen2-containerd
```
djsly, line 37: for my own knowledge, what are we building those for ?  from the looks of it , those are not all of the BHD images we built, since it only 3 of them

.pipelines/templates/e2e-template.yaml
```diff
@@ -0,0 +1,52 @@
...
+          if [ -n "${VHD_BUILD_ID}" ]; then
+            export SIG_VERSION_TAG_NAME=buildId
+            export SIG_VERSION_TAG_VALUE=$VHD_BUILD_ID
+            export IGNORE_SCENARIOS_WITH_MISSING_VHD=true
+          else
+            echo "Build.BuildId is not specified. Running default e2e tests."
+          fi
+          
+          cd e2e
+          go test -parallel 100 -timeout 90m -v -run Test_All ./
```
djsly, line 31: could the parallel and timeout variable be added to the ab-e2e group ?

r2k1: Currently it contains only credentials. I think @cameronmeissner wants to delete them in a favour of managed identity

.pipelines/.vsts-vhd-builder.yaml
r2k1: I'm not 100% sure. But I think it's a small setup to run on each PR. Probably with a goal to consume less resources, be faster and be less flaky.

I think it's the main place where we want to run our e2e. Before people merge PRs.

.pipelines/templates/e2e-template.yaml
r2k1: I tried to make it configurable. But it's not trivial and the feedback loop of 1h+ is killing me.

djsly: that's fair.  we can live with them hardcoded and turn them configurable if we end up needing to change them.

.pipelines/.vsts-vhd-builder.yaml
djsly: ok I see, so a subset of the builds during the PR to get a somewhat good coverage while focussing on efficiency.

.pipelines/e2e.yaml
cameronmeissner: I think the main thing is that we don't currently run abe2e's in AME (and might require some extra work to do so), so having a flag we can easily use to turn on/off the e2e's might be useful

r2k1: My current goal is to run e2e on each PR. With a goal to reduce manual testing and to avoid merging broken things.

I think release is a more manual process I don't know much about, so I can't really suggest the e2e automation there.

r2k1: @cameronmeissner Do you have an example of a specific pipeline run where executing the e2e test would cause any issues?

I don't fully understand the issue, so I can't come up with a good solution at the moment. It’s quite easy to disable in the future. I suggest delaying the attempt to solve potential issues until we encounter them.

cameronmeissner (approved):
maybe @anujmaheshwari1 wants to take a look as well?

anujmaheshwari1 (approved):

PR #4519, opened 2024-06-14
fix: dns resolution for abe2e airgrap security rules

cameronmeissner:
**What type of PR is this?**

fixes incorrect security rules on airgap NSGs within agentbaker e2e's. this was caused by a change to the front-door IP address of `mcr.microsoft.com`, manifesting in provisioning failures with exit code 50.

this PR adds logic to dynamically lookup the IP address of each required DNS name so future changes of this nature no longer break anything

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/airgap.go
+++ b/e2e/airgap.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"strings"
 
 	"github.com/Azure/agentbakere2e/suite"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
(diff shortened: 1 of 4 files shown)


e2e/airgap.go
```diff
@@ -5,19 +5,53 @@ import (
...
+	// https://learn.microsoft.com/en-us/azure/aks/outbound-rules-control-egress#azure-global-required-fqdn--application-rules
+	// note that we explicitly exclude packages.microsoft.com
+	requiredDNSNames := []string{
+		"mcr.microsoft.com",
+		"management.azure.com",
+		"acs-mirror.azureedge.net",
+		clusterFQDN,
+	}
+	var rules []*armnetwork.SecurityRule
+	var priority int32 = 100
```
djsly, line 30: could we call this startingPriority or initialPrioirty ?

djsly (approved):

e2e/airgap.go
cameronmeissner: personally I think it makes sense as is since it gets incremented and applied to every rule

PR #4521, opened 2024-06-14
test: output JSON for VHD build benchmarks and publish as pipeline artifact

zachary-bailey:
What type of PR is this?

/test

What this PR does / why we need it:

This PR changes the workflow for benchmarking the VHD build process and publishes a build artifact at the conclusion of the pipeline. Changes include removing the start_watch function, adding the installJq function, and outputting benchmarking data in JSON instead of formatted strings. At the end of each script, a JSON object will be output in the terminal with performance data for each section.

The installJq function was necessary because jq is not installed until the installDeps function is called in install-dependencies.sh, but it is needed when the build process starts in pre-install-dependencies.sh.

The pipeline artifact that will be published is a JSON file that will contain benchmarking data for the build process. In later changes, github actions will be used to evaluate this artifact and check for build time regression.

The formatting of the artifact can be checked by going to the PR gate build and checking the artifacts list.

Which issue(s) this PR fixes:

This will prevent vhd build times from incrementally increasing due to code changes.

Note: When reviewing "Files changed", please collapse /pkg/agent/testdata to see all relevant file changes
**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,80 @@ should_skip_nvidia_drivers() {
...
+  # data will be stored in a JSON file and published as an artifact to be evaluated by github actions
+  set +x
+
+  local is_final_section=$1
+  local title=$2
 
   local current_time=$(date +%s)
   local end_timestamp=$(date +%H:%M:%S)
-  local difference_in_seconds=$((current_time - ${1}))
+  if [ "$is_final_section" = true ]; then
```
Devinwong, line 414: ==, right?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,80 @@ should_skip_nvidia_drivers() {
...
+      echo "jq was installed: $(jq --version)"
+    else
+      apt_get_install 5 1 60 jq
+      echo "jq was installed: $(jq --version)"
+    fi
+  fi
 }
 
-stop_watch () {
+capture_benchmarks () {
```
Devinwong, line 404: This function has 2 purposes, namely, capture benchmarks for a section and for the whole script.
If I were implementing it, I would separate this function into two, e.g., `capture_section_benchmarks` and `capture_script_benchmarks` (or capture_final_section_benchmarks` if it's more appropriate), instead of using a bool to control the logic.
The reasons behind that are
- Single Responsibility Principle: The idea behind the SRP is that every class, module, or function in a program should have one responsibility/purpose in a program (copied from search engine)
- There isn't too much common logic between when $is_final_section == true vs false. So duplicate codes shouldn't be an issue here. 
- Even there were duplicate codes, I would put them into functions so that `capture_section_benchmarks` and `capture_script_benchmarks` can reuse.

Just my 2 cents :)

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,80 @@ should_skip_nvidia_drivers() {
...
-    printf -v start "     Start time: $script_start_timestamp"
-  else
-    printf -v start "     Start time: $start_timestamp"
+  if [[ "$is_final_section" == true ]]; then 
+    
+    # if current section is the final section of the current script, create a JSON object for the script as a whole
+    script_object=$(jq -n --arg script_name "$title" --arg script_start_timestamp "$script_start_timestamp" --arg end_timestamp "$end_timestamp" --arg total_time_elapsed "$total_time_elapsed" '{($script_name): {"overall": {"start_time": $script_start_timestamp, "end_time": $end_timestamp, "total_time_elapsed": $total_time_elapsed}}}')
+
+    #iterate over the benchmarks array in order to retrieve data from previous sections and create section objects for each
+    for ((i=0; i<${#benchmarks[@]}; i+=4)); do
```
Devinwong, line 439: nit: consider using 4 arrays to store the information, namely, section_name, section_start_timestamp, end_timestamp, total_time_elapsed if there is no obvious benefit to use a single array. Just for easier debugging consideration. (Struct is not supported in bash script, otherwise it would be even better)

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,80 @@ should_skip_nvidia_drivers() {
...
-  start_timestamp=$(date +%H:%M:%S)
+installJq () {
+  # jq is not available until downloaded in install-dependencies.sh with the installDeps function
+  # but it is needed earlier to call the capture_benchmarks function in pre-install-dependencies.sh
+  output=$(jq --version)
+  if [ -n "$output" ]; then
+    echo "$output"
+  else
+    if [[ $OS == $MARINER_OS_NAME ]]; then
+      sudo tdnf install -y jq
```
Devinwong, line 395: should we capture the error in case it fails to install?
For example, in install_dependencies.sh it is doing something like this `apt_get_install 30 1 600 grub-efi || exit 1`

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,80 @@ should_skip_nvidia_drivers() {
...
+      echo "jq was installed: $(jq --version)"
+    else
+      apt_get_install 5 1 60 jq
```
Devinwong, line 398: Same as above. Capture error?

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong, line 404: I just added a shell script unit test and merged to master branch. Please git pull master branch and take a look at the section `shell scripts unit tests` in the readme. It will be great if you can add unit tests to your new functions. Examples are listed in the readme too.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,96 @@ should_skip_nvidia_drivers() {
...
+      sudo tdnf install -y jq && echo "jq was installed: $(jq --version)"
+    else
+      apt_get_install 5 1 60 jq && echo "jq was installed: $(jq --version)"
+    fi
+  fi
 }
 
-stop_watch () {
+capture_benchmark () {
+
```
Devinwong, line 403: extra empty line

zachary-bailey: Thanks, fixed! For both functions

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,96 @@ should_skip_nvidia_drivers() {
...
+  set -x
 }
 
-show_benchmarks () {
-  echo -e "\nBenchmarks:\n"
-  for i in "${benchmarks[@]}"; do
-    echo "   $i"
+process_benchmarks () {
+
```
Devinwong: extra line?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,96 @@ should_skip_nvidia_drivers() {
...
 }
 
-stop_watch () {
+capture_benchmark () {
+
+  set +x
+  # add current section title to the benchmarks array as a string
+  benchmarks+=($1)
+  # use nameref variable to hold that section title to reference that specific section array later in the function
+  declare -n current_section="${benchmarks[-1]}"
```
Devinwong, line 415: Just a note FYR: I also used array index [-1] but in the E2E or test build VHD (I forgot which) it complains -1 is invalid, although my local unit test didn't complain about it. Then I found that this is a new syntax only supported in newer version of bash. If you run into index error later, this will be the place.

zachary-bailey: Thank you. What syntax did you end up using? It has worked just fine in the test pipeline so far

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -384,40 +384,96 @@ should_skip_nvidia_drivers() {
...
+  local is_final_section=${2:-false}
 
   local current_time=$(date +%s)
   local end_timestamp=$(date +%H:%M:%S)
-  local difference_in_seconds=$((current_time - ${1}))
+  if [[ "$is_final_section" == true ]]; then
+    local start_timestamp=$script_start_timestamp
```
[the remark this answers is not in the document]
zachary-bailey: Yes it works! local refers to function scope I believe?

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: I needed to get the last element by getting the size of the array `#{array[@]}` and minus 1. Mind that if array size =0, you will still need to handle the potential -1.

djsly:
LGTM, I will let others approve

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -47,7 +47,7 @@ stages:
             echo '##vso[task.setvariable variable=IMG_SKU]22_04-lts-gen2'
             echo '##vso[task.setvariable variable=IMG_VERSION]latest'
             echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
-            echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D16ds_v5'
+            echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_DS5_v2'
```
djsly, line 50: should we open a PR just to have those updated ?

Devinwong (approved):
Looks good to me. Thanks for all the changes!

PR #4528, opened 2024-06-17
calico new images for caching

paulgmiller:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/components.json
+++ b/parts/linux/cloud-init/artifacts/components.json
@@ -179,11 +179,12 @@
       "downloadURL": "mcr.microsoft.com/oss/calico/cni:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v3.24.6"
+        "v3.24.6",
+        "v3.26.4-c06a60"
       ],
       "prefetchOptimizations": [
         {
-          "version": "v3.24.6",
+          "version": "v3.26.4-c06a60",
           "binaries": [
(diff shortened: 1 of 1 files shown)


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -202,18 +203,20 @@
       "downloadURL": "mcr.microsoft.com/oss/calico/node:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v3.24.6"
+        "v3.24.6",
```
djsly, line 206: should `calico/node` have a `prefetchOptimazations` section ?

djsly (approved):

parts/linux/cloud-init/artifacts/components.json
cameronmeissner: we haven't prefetched this thus far, though not sure if we really _should_ be

cameronmeissner:
do we still need this?

parts/linux/cloud-init/artifacts/components.json
paulgmiller: I mean I don't actually want to push calcio so maybe the better thing to do is drop it completely. Also are we going to cache 4 versions when I bring in calcio 3.28?

paulgmiller:
Killing Calico caching.

PR #4531, opened 2024-06-17
feat: add outbound check variable

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -906,6 +906,12 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetOutboundCommand": func() string {
 			return getOutBoundCmd(config, config.CloudSpecConfig)
 		},
+		"BlockOutboundNetwork": func() bool {
+			if config.OutboundType == datamodel.OutboundTypeBlock || config.OutboundType == datamodel.OutboundTypeNone {
```
cameronmeissner, line 910: are we sure we want `OutboundTypeNone` to indicate the same behavior as `OutboundTypeBlock`?

pkg/agent/baker_test.go
```diff
@@ -859,6 +859,12 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 			},
 		),
 
+		Entry("AKSUbuntu2204 with outbound type blocked", "AKSUbuntu2204+OutboundTypeBlocked", "1.25.6", func(config *datamodel.NodeBootstrappingConfiguration) {
```
cameronmeissner, line 862: let's add a negative test case as well

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -128,7 +128,10 @@ fi
 
 setupCNIDirs
 
-logs_to_events "AKS.CSE.installNetworkPlugin" installNetworkPlugin
+# Network plugin should not install if outbound is not allowed.
+if [ "${BLOCK_OUTBOUND_NETWORK}" == "false" ]; then
+    logs_to_events "AKS.CSE.installNetworkPlugin" installNetworkPlugin
```
cameronmeissner, line 133: have we re-done things to make sure we still get a valid CNI plugin even if RP wants a different version than what's cached, and we want to block outbound network?

pkg/agent/baker.go
AlisonB319: Yes, they both mean the same to us - I've double checked this a handful of times since I think its a bit convoluted but it is what I've been told

parts/linux/cloud-init/artifacts/cse_main.sh
AlisonB319: I'm going to take this portion out of the PR for now, since I would like to get the BLOCK_OUTBOUND_NETWORK var in. But I think I made a mistake here, I'll need to double check, but we should just be getting an alternative URL to download CNI

pkg/agent/baker.go
cameronmeissner: can we add comments to the respective definitions of these outbound types to indicate they mean the same thing?

maybe worth following up and seeing if we can eliminate one or the other altogether

pkg/agent/baker_test.go
```diff
@@ -859,6 +859,18 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
...
+			config.OutboundType = datamodel.OutboundTypeBlock
```
cameronmeissner, line 863: add a test case for `OutboundTypeNone` as well?

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -88,6 +88,7 @@ GPU_NEEDS_FABRIC_MANAGER="{{GPUNeedsFabricManager}}"
 NEEDS_DOCKER_LOGIN="{{and IsDockerContainerRuntime HasPrivateAzureRegistryServer}}"
 IPV6_DUAL_STACK_ENABLED="{{IsIPv6DualStackFeatureEnabled}}"
 OUTBOUND_COMMAND="{{GetOutboundCommand}}"
+BLOCK_OUTBOUND_NETWORK="{{BlockOutboundNetwork}}"
```
cameronmeissner, line 91: not critical yet since we aren't actually using it, though we will want to let the karpenter folks know this is a new parameter they'll also need to set on their side

cameronmeissner (approved):

pkg/agent/baker.go
AlisonB319: Here's a reply from Coco in one of the various NIC chats. I'll leave some commends as well but I'll drop the response in the comment for legacy sake 

What’s the difference between outboundtype=none and outboundtype=block? 
`–outbound-type=none`: AKS doesn't set up an outbound for users, but users can set the outbound rules. If there is an Azure Route Server announcing a default route via an NVA for egress to the internet. In this scenario, outboundType==none would still allow internet egress via the NVA 

`–outbound-type=block`: AKS doesn't set up an outbound for users and AKS actively adds an NSG rule to all cluster nodes to prevent egress to the internet. So users cannot set the outbound rules.  With  `–outbound-type=block` and ` ArtifactSource=Cache`, it’s the ‘zero egree cluster’. There are no outbound rules allowed on the cluster.

djsly: the finny part is that the datamodel only contains those two attribute, so it's unclear what other value there could be

PR #4532, opened 2024-06-18
refactor: (new) move containerd and runc from manifest.json to components.json. Merge DownloadFiles to Packages

Devinwong:
**What type of PR is this?**

/kind feature

- Move containerd and runc from manifest.json to components.json->Packages.
- Merge former DownloadFiles to Packages, in components.json
- Components.cue defines the schema and components.json hosts the actual components including ContainerImages and Packages.
- Packages and corresponding versions are only defined in components.json now. User can define single or multiple package versions to download, depending on the use case. User can also different version or download URL according to different OS Distro (Ubuntu/Mariner) and releases (18.04/20.04 etc.). A package's downloadURL can contains unresolved variables for the scripts to resolve later, to provide an additional level of flexibility.
- Add shellspec to do unit tests. If there is better framework, I can migrate the test codes to that too.
- Update the component data model to fit the new schema changes.

AB E2E passed. https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=97984006&view=results

Fixes #

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -42,20 +42,20 @@ installContainerRuntime() {
     if [[ "${NEEDS_CONTAINERD}" == "true" ]]; then
         echo "in installContainerRuntime - KUBERNETES_VERSION = ${KUBERNETES_VERSION}"
         local containerd_version
-        if [[ -f "$COMPONENTS_FILEPATH" ]]; then
+        if [[ ! -f "$COMPONENTS_FILEPATH" ]]; then
```
Devinwong, line 45: This should be the bug for the E2E test failure.
Added 1 unit test to cover this too.

djsly (approved):

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -41,21 +41,20 @@ cleanupContainerdDlFiles() {
...
         echo "in installContainerRuntime - KUBERNETES_VERSION = ${KUBERNETES_VERSION}"
-        local containerd_version
-        if [[ -f "$COMPONENTS_FILEPATH" ]]; then
+        if [[ ! -f "$COMPONENTS_FILEPATH" ]]; then
             echo "WARNING: $COMPONENTS_FILEPATH not found. Skipping validation."
-            return
+            return 0
         fi
-        os=UBUNTU_OS_NAME
+        os=${UBUNTU_OS_NAME}
```
Devinwong, line 48: This is another bug. Newly added unit test cover this too.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -103,18 +103,19 @@ EOF
...
-        echo "currently installed containerd version ${CURRENT_VERSION} is greater than (or equal to) target base version ${CONTAINERD_VERSION}. skipping installStandaloneContainerd."
+    if semverCompare ${CURRENT_VERSION:-"0.0.0"} ${desiredVersion}; then
+        echo "currently installed containerd version ${CURRENT_VERSION} is greater than (or equal to) target base version ${desiredVersion}. skipping installStandaloneContainerd."
     else
-        echo "installing containerd version ${CONTAINERD_VERSION}"
+        echo "installing containerd version ${desiredVersion}"
         removeContainerd
         # TODO: tie runc to r92 once that's possible on Mariner's pkg repo and if we're still using v1.linux shim
-        if ! dnf_install 30 1 600 moby-containerd; then
+        if ! dnf_install 30 1 600 "moby-containerd-${desiredVersion}"; then
```
cameronmeissner, line 118: did you test to make sure specifying an actual version here works via build logs?

Devinwong, line 118: we were not specifying mobd-containerd version for Mariner. Adding it here.

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -411,7 +411,7 @@
         "mariner": {
           "current": {
             "versions": [
-              "1.4.54"
+              "1.1.9-5.cm2"
```
ganeshkumarashok, line 414: What does "cm2" stand for?

Devinwong: I don't know either. But I got this version from a previous pipeline that we've been using.
`azure-arm:  moby-runc x86_64 1.1.9-5.cm2 mariner-official-base 3.4 M`

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
Devinwong: yes I tested it manually. Let me paste the link.

Devinwong: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=96177599&view=logs&j=da6da462-43f6-5106-fd69-5730003ba730&t=274a1bb5-2aca-56be-d627-49f35f75c502&l=4725
```
==> azure-arm: + dnf_install 30 1 600 moby-containerd-1.6.26-5.cm2
==> azure-arm: + retries=30
==> azure-arm: + wait_sleep=1
==> azure-arm: + timeout=600
==> azure-arm: + shift
==> azure-arm: + shift
==> azure-arm: + shift
==> azure-arm: ++ seq 1 30
==> azure-arm: + for i in $(seq 1 $retries)
==> azure-arm: + dnf install -y moby-containerd-1.6.26-5.cm2
```

cameronmeissner:
will need a rebase from master

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -194,12 +194,17 @@ testPackagesInstalled() {
         break
       fi
       eval "downloadURL=${downloadURL}"
-      fileName=$(basename $downloadURL)
-      dest="$downloadLocation/${fileName}"
+      local fileNameWithExt
+      fileNameWithExt=$(basename $downloadURL)
```
djsly, line 198: what are the possible value for `downloadURL` ?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -80,6 +80,75 @@ cleanUpGPUDrivers() {
...
+    # the one already installed on the node considering the source is built by the user for hotfix or test
+    logs_to_events "AKS.CSE.installContainerRuntime.removeMoby" removeMoby
+    logs_to_events "AKS.CSE.installContainerRuntime.removeContainerd" removeContainerd
+    logs_to_events "AKS.CSE.installContainerRuntime.downloadContainerdFromURL" downloadContainerdFromURL "${containerdOverrideDownloadURL}"
+    logs_to_events "AKS.CSE.installContainerRuntime.installDebPackageFromFile" "installDebPackageFromFile ${CONTAINERD_DEB_FILE}" || exit $ERR_CONTAINERD_INSTALL_TIMEOUT
+    echo "Succeeded to install containerd from user input: ${containerdOverrideDownloadURL}"
+    return 0
+}
+
+installContainerdWithAptGet() {
```
Devinwong, line 111: The logic of this function is mostly from the original func `installStandaloneContainerd`

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -35,31 +38,79 @@ cleanupContainerdDlFiles() {
...
-            containerd_version="$(jq -r .containerd.edge "$MANIFEST_FILEPATH")"
-            if [ "${UBUNTU_RELEASE}" == "18.04" ]; then
-                containerd_version="$(jq -r '.containerd.pinned."1804"' "$MANIFEST_FILEPATH")"
-            fi
-        else
-            echo "WARNING: containerd version not found in manifest, defaulting to hardcoded."
-        fi
+# After the centralized packages changes, the containerd versions are only available in the components.json.
+installContainerdWithComponentsJson() {
+    os=${UBUNTU_OS_NAME}
```
cameronmeissner, line 43: could we factor this out to a somewhat generic `installComponent` function? seems a lot of this logic will be common across all components we install via the components file

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -35,31 +38,79 @@ cleanupContainerdDlFiles() {
...
+    if [[ -z "$UBUNTU_RELEASE" ]]; then
+        os=${MARINER_OS_NAME}
+        os_version="current"
+    fi
+    os_version="${UBUNTU_RELEASE}"
+    containerdPackage=$(jq ".Packages" "$COMPONENTS_FILEPATH" | jq ".[] | select(.name == \"containerd\")") || exit $ERR_CONTAINERD_VERSION_INVALID
+    PackageVersions=()
+    returnPackageVersions "${containerdPackage}" "${os}" "${os_version}"
```
cameronmeissner, line 51: somewhat confused... how is `returnPackageVersions` populating `PackgeVersions`?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -35,31 +38,79 @@ cleanupContainerdDlFiles() {
...
-            if [ "${UBUNTU_RELEASE}" == "18.04" ]; then
-                containerd_version="$(jq -r '.containerd.pinned."1804"' "$MANIFEST_FILEPATH")"
-            fi
-        else
-            echo "WARNING: containerd version not found in manifest, defaulting to hardcoded."
-        fi
+# After the centralized packages changes, the containerd versions are only available in the components.json.
+installContainerdWithComponentsJson() {
+    os=${UBUNTU_OS_NAME}
+    if [[ -z "$UBUNTU_RELEASE" ]]; then
```
cameronmeissner, line 44: can't we just check `if [ "${OS}" == "$UBUNTU_OS_NAME" ] `?

cameronmeissner: ah, I guess we don't really have access to that here, since these functions are supposed to be "common", in that they just work for both ubuntu + mariner (the distro-specific ones are in cse_install_ubuntu.sh/cse_install_mariner.sh)

pkg/agent/vhd/cache/types.go
```diff
@@ -1,16 +1,14 @@
 package cache
 
-// CachedOnVHD represents the cached components on the VHD.
+// OnVHD represents the cached components on the VHD.
 type OnVHD struct {
```
cameronmeissner, line 4: @AlisonB319 do we still need this stuff? I don't think we ever ended up using it for NIC did we?

spec/parts/linux/cloud-init/artifacts/test_components.json
cameronmeissner, line 1: im assuming we always want this file to reflect the exact state of the actual components.json? right now we'd have to manually keep them in sync right?

might be worth using a symlink instead so we only have to worry about updating a single file

spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh
```diff
@@ -0,0 +1,169 @@
...
+
+source parts/linux/cloud-init/artifacts/cse_helpers.sh
+
+readPackage() {
+    local packageName=$1
+    packages=$(jq ".Packages" "./spec/parts/linux/cloud-init/artifacts/test_components.json" | jq ".[] | select(.name == \"$packageName\")")
+    echo "$packages"
+}
+
+Describe 'cse_install.sh'
```
cameronmeissner, line 11: assuming these are unit tests of some sort, can we add a readme to explain what this is/how they work?

cameronmeissner: ah nvm , I see you updated the readme at the root

README.md
```diff
@@ -59,6 +59,36 @@ See the [git documentation](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-H
...
+
+You will need to write `xxx_spec.sh` file for the test.
+
+For example, `AgentBaker/spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh` is a test file for `AgentBaker/parts/linux/cloud-init/artifacts/cse_install.sh`
+
+### Running tests locally
+
+To run all tests, in AgentBaker folder, simply run `bash shellspec`. (Assuming you set the $PATH correctly, otherwise you will need to provide the full path for shellspec binary)
+
+#### Useful commands for debugging
```
cameronmeissner, line 84: could we incorporate installation/usage into the Makefile we already have to keep things simple?

we'd probably want to incorporate this into the `generate` target anyways

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -35,31 +38,79 @@ cleanupContainerdDlFiles() {
...
     fi
+    if [ -f "$COMPONENTS_FILEPATH" ] && jq '.Packages[] | select(.name == "containerd")' < $COMPONENTS_FILEPATH > /dev/null; then
+        echo "Package \"containerd\" exists in $COMPONENTS_FILEPATH."
+        # if the containerd package is available in the components.json, use the components.json to install containerd
+        installContainerdWithComponentsJson
+		return
+    fi
+    echo "Package \"containerd\" does not exist in $COMPONENTS_FILEPATH."
+    # if the containerd package is not available in the components.json, use the manifest.json to install containerd
+    installContainerdWithManifestJson
```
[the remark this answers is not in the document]
Devinwong: I didn't have this initially. But without this the ubuntu2204privatekubepkg will fail as it doesn't have containerd definition in components.json. If we are able not to support older image, we can remove this ManifestJson route.

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: I did't find a good way to return an array of strings in Bash scripts. So the func `returnPackageVersions` is actually modifying the global variable `PackageVersions` , defined 1 line above,

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: You are right. We don't know what os it is yet. So we needed to use the command `UBUNTU_RELEASE=$(lsb_release -r -s)` to identify it.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -128,6 +127,10 @@ installStandaloneContainerd() {
 
 }
 
+ensureRunc() {
+  echo "Mariner Runc is included in the Mariner base image. Skipping downloading and installing Runc"
```
AlisonB319, line 131: Do we have to have this function because cse_instsall_ubuntu has its own ensureRunc()?

pkg/agent/vhd/cache/types.go
AlisonB319: I think there is leftover stuff from api.GetCachedVersionsOnVHD that I will create a new PR to remove

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Hmm. There are 2 reasons I didn't do that. But I am open to that.
1. (Minor) This PR has been quite large. Refactoring the download/install logic will introduce another aspect of complexity which will make it even harder to review.
2. (Major) The download/install logic of each package is quite different. I left a comment in the code that
```*)
      echo "Package name: ${name} not supported for download. Please implement the download logic in the script."
      # We can add a common function to download a generic package here.
      # However, installation could be different for different packages.
      ;;
```

For the 5 packages we have now, we are already using various methods to download/install.
1. cri-tools: download using curl and install after artifact streaming. Not sure if it has to install after that.
2. azure-cni: download a tar ball using curl and untar
3. cni-plugins: similar to azure-cni
4. runc: For Ubuntu, it downloads a deb file using curl and install the deb file. For Mariner, it's included in the base image so do nothing.
5. containerd:
- Ubuntu: download/install either through url or apt-get
- Mariner: using dnf to install

We can still have a generic installComponent. But inside it will still check what package it is to use the corresponding method to download/install.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
Devinwong: In my install-depenedencies.sh around ln 232, if the package is 'runc', ensureRunc will be called. Therefore in both cse_install_ubuntu and cse_install_mariner it needs to have the function. Another way to do that is add an if before that line, if it's Ubuntu, call ensureRunc. Otherwise do nothing (maybe leave the same echo there).

README.md
Devinwong: Will do! If time does not allow, I will do it in the next PR. (Just trying to unblock other folks with this PR)

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
AlisonB319: I'll leave it up to you on what you think is more readable since you're more familiar with this block of code :)

spec/parts/linux/cloud-init/artifacts/test_components.json
Devinwong: done!

pkg/agent/vhd/cache/types.go
cameronmeissner: removed in a separate PR

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -485,4 +486,98 @@ process_benchmarks() {
...
+            downloadURL=$(echo "${package}" | jq ".downloadURIs.ubuntu.${release}.downloadURL" -r)
+            echo ${downloadURL}
+            return
+        fi
+        downloadURL=$(echo "${package}" | jq ".downloadURIs.default.${release}.downloadURL" -r)
+        echo ${downloadURL}
+        return    
+    fi
+    if [[ "${os}" == "${MARINER_OS_NAME}" ]]; then
+        #if .downloadURIs.ubuntu exist, then get the downloadURL from there.
```
anujmaheshwari1, line 570: supernit, comment still has downloadURIs.ubuntu

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -485,4 +486,98 @@ process_benchmarks() {
...
+        fi
+        versions=$(echo "${package}" | jq ".downloadURIs.default.${release}.versions[]" -r)
+        for version in ${versions[@]}; do
+            PackageVersions+=("${version}")
+        done
+        return    
+    fi
+}
+
+returnPackageDownloadURL() {
```
anujmaheshwari1, line 552: this function can be really simple without the if conditionals, correct me if I am wrong:

why we need conditionals is downloadURIs.(ubuntu/mariner/default).${release} right? everything else is the same.

how about we try to access downloadURIs.{osName} and if it doesn't exist, go to downURIs.default?

we can get rid of the osName conditionals then and make the function really simple.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -71,9 +122,11 @@ installNetworkPlugin() {
 }
 
 downloadCNI() {
-    mkdir -p $CNI_DOWNLOADS_DIR
-    CNI_TGZ_TMP=${CNI_PLUGINS_URL##*/} # Use bash builtin ## to remove all chars ("*") up to the final "/"
-    retrycmd_get_tarball 120 5 "$CNI_DOWNLOADS_DIR/${CNI_TGZ_TMP}" ${CNI_PLUGINS_URL} || exit $ERR_CNI_DOWNLOAD_TIMEOUT
+    downloadDir=${1:-${CNI_DOWNLOADS_DIR}}
```
anujmaheshwari1, line 125: nit: can we preserve the casing for variables? we have upper case everywhere in cse scripts

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: Good suggestion! Didn't realize the codes were actually duplicated. Updated as you suggested.

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: Modified the comment to fit the new approach

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -485,4 +486,98 @@ process_benchmarks() {
...
+  local package="$1"
+  local os="$2"
+  local osVersion="$3"
+  local release="current"
+  #For UBUNTU, if $osVersion is 18.04 and "r1804" is also defined in components.json, then $release is set to "r1804"
+  #Similarly for 20.04 and 22.04. Otherwise $release is set to .current.
+  #For MARINER, the release is always set to "current" now.
+  #To add a new release, add a new entry in components.json and add the corresponding release in the below if condition.
+  if [[ "${os}" == "${UBUNTU_OS_NAME}" ]]; then
+    if [[ "${osVersion}" == "18.04" ]] && [[ $(echo "${package}" | jq '.downloadURIs.ubuntu."r1804"') != "null" ]]; then
```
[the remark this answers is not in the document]
Devinwong: Good suggestion! Updated as you suggested and add a unit test to cover it.

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Actually I found multiple naming convention in a single .sh file. For example, `images_to_delete`, `NEEDS_CONTAINERD`, `allMCRImages`. That's why I picked one which is
```
ALL_UPPER_CASE for constants, global variables and environment variables
camelCase for local variables.
```
We can definitely revise and unify all the variables. Does this sound make sense?

parts/linux/cloud-init/artifacts/cse_install.sh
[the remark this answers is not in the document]
cameronmeissner: I think we can probably just rebuild the VHD for that particular e2e scenario, that's something we should do anyways

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner: yeah, Anuj and I were talking about conditional syntax being inconsistent and all over the place as well...

maybe would be better to do that sort of thing in a separate PR

parts/linux/cloud-init/artifacts/cse_install.sh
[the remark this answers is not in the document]
cameronmeissner: while we take care of that we can also temporarily disable the scenario

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -485,4 +486,80 @@ process_benchmarks() {
...
+    local osVersionWithoutDot=$(echo "${osVersion}" | sed 's/\.//g')
+    #For UBUNTU, if $osVersion is 18.04 and "r1804" is also defined in components.json, then $release is set to "r1804"
+    #Similarly for 20.04 and 22.04. Otherwise $release is set to .current.
+    #For MARINER, the release is always set to "current" now.
+    if [[ "${os}" != "${UBUNTU_OS_NAME}" ]]; then
+        echo "${release}"
+        return
+    fi
+    if [[ $(echo "${package}" | jq ".downloadURIs.ubuntu.\"r${osVersionWithoutDot}\"") != "null" ]]; then
+        release="\"r${osVersionWithoutDot}\""
```
cameronmeissner, line 506: are we sure this works properly? seems like it would be assigning to the locally-scoped `release` var

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -42,21 +42,14 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cni:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.5.28",
-        "v1.6.0"
```
cameronmeissner, line 46: why remove 1.6.0?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -104,8 +90,7 @@
       "multiArchVersions": [
         "v0.0.13",
         "v0.0.20",
-        "v0.1.4",
-        "v0.2.0"
```
cameronmeissner, line 108: seems like this would also remove 0.2.0, assuming these are typos?

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: This function will return a string by `echo "${release}"` and the caller will capture it. I have a unit test to cover this so it's working as expected. I run also run a complete abe2e to ensure.

parts/linux/cloud-init/artifacts/components.json
Devinwong: Oh didn't see this. I didn't delete them intentionally. Let me add all back.
Thanks for catching this.

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: so one thing we can do here is assign the output to a non-local variable within the function, then the caller will actually have access to it, that's actually how we do it in our release scripts

I'd personally prefer that over the caller having to capture stdout

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -385,32 +364,119 @@
...
+    {
+      "name": "runc",
+      "downloadLocation": "/opt/runc/downloads",
+      "downloadURIs": {
+        "ubuntu": {
+          "r1804": {
+            "versions": [
+              "1.1.12-ubuntu18.04u1"
+            ],
+            "downloadURL": ""
```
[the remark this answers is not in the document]
Devinwong: fixed!

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: Updated.
Though I personally like to echo it so that we don't need so many global variables. But understand that it's the script convention.

cameronmeissner (changes requested):
seems like from the latest abe2e run mariner is trying to install the containerd version for ubuntu, which doesn't exist:

```
++ echo 1.7.15
+ VERSION_B=1.7.15
+ [[ 1.6.26 == \1\.\7\.\1\5 ]]
++ echo 1.6.26 1.7.15
++ sort -V
++ tr ' ' '\n'
+ sorted='1.6.26
1.7.15'
++ IFS=
++ echo '1.6.26
1.7.15'
++ cut '-d
' -f2
+ highestVersion=1.7.15
+ [[ 1.6.26 == 1.7.15 ]]
+ return 1
+ echo 'installing containerd version 1.7.15'
installing containerd version 1.7.15

...

++ seq 1 30
+ for i in $(seq 1 $retries)
+ dnf install -y moby-containerd-1.7.15
environment: line 4: ctr: command not found
Executed "bash -c cleanupImagesRun" 2 times
Fri Jul 12 00:15:24 UTC 2024,abtest7lln000000, endCleanUpKubeProxyImages
Last metadata expiration check: 7:13:34 ago on Thu Jul 11 17:01:47 2024.
No match for argument: moby-containerd-1.7.15
Error: Unable to find a match: moby-containerd-1.7.15
```

Devinwong:
> seems like from the latest abe2e run mariner is trying to install the containerd version for ubuntu, which doesn't exist:
> 
> ```
> ++ echo 1.7.15
> + VERSION_B=1.7.15
> + [[ 1.6.26 == \1\.\7\.\1\5 ]]
> ++ echo 1.6.26 1.7.15
> ++ sort -V
> ++ tr ' ' '\n'
> + sorted='1.6.26
> 1.7.15'
> ++ IFS=
> ++ echo '1.6.26
> 1.7.15'
> ++ cut '-d
> ' -f2
> + highestVersion=1.7.15
> + [[ 1.6.26 == 1.7.15 ]]
> + return 1
> + echo 'installing containerd version 1.7.15'
> installing containerd version 1.7.15
> 
> ...
> 
> ++ seq 1 30
> + for i in $(seq 1 $retries)
> + dnf install -y moby-containerd-1.7.15
> environment: line 4: ctr: command not found
> Executed "bash -c cleanupImagesRun" 2 times
> Fri Jul 12 00:15:24 UTC 2024,abtest7lln000000, endCleanUpKubeProxyImages
> Last metadata expiration check: 7:13:34 ago on Thu Jul 11 17:01:47 2024.
> No match for argument: moby-containerd-1.7.15
> Error: Unable to find a match: moby-containerd-1.7.15
> ```

@cameronmeissner  can you link me to this error? It seems I couldn't find the same error in this e2e logs. https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=97984006&view=logs&j=beeeaba8-323e-5ce9-426a-8cbca9ffe3b0&t=b32dc13d-e9ba-5292-9703-579f43e16486

README.md
Devinwong: done

Makefile
```diff
@@ -207,4 +216,8 @@ coverage:
 unit-tests:
 	$(GO) test `go list ./... | grep -v e2e` -coverprofile coverage_raw.out -covermode count
 
+.PHONY: validate-components-json
+validate-components-json:
```
cameronmeissner, line 220: nit: `validate-components`?

cameronmeissner (approved):

anujmaheshwari1 (approved):

Makefile
Devinwong: updated

parts/linux/cloud-init/artifacts/cse_install.sh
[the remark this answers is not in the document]
Devinwong: I will remove installContainerdWithManifestJson once I build a newer ubuntu2204privatekubepkg VHD.

spec/parts/linux/cloud-init/artifacts/test_components.json
cameronmeissner, line 1: non-blocking, but why do we actually need a separate file at all? can't we just use the original file directly? (which is basically the same as using a symlink)

Devinwong: You are right. Updated again.

PR #4549, opened 2024-06-24
feat: Add new scanning pipeline yaml

AlisonB319:
**What type of PR is this?**
feat: adding empty pipeline yaml so that I can start testing

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-scanning.yaml
+++ b/.pipelines/.vsts-vhd-scanning.yaml
@@ -1,1 +1,1 @@
-name: $(Date:yyyyMMdd)$(Rev:.r)_$(OS_SKU)$(OS_VERSION)_$(HYPERV_GENERATION)_$(FEATURE_FLAGS)_$(Build.SourceBranchName)./
+# empty file WIP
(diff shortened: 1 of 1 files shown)


.pipelines/.vsts-vhd-scanning.yaml
```diff
@@ -0,0 +1 @@
+# empty file WIP
```
cameronmeissner, line 1: what will the pipeline be used for?

AlisonB319: In addition to scanning during vhd builds, infra? (someone?) wants trivy scans for like n-1 images in production. So this pipeline will pretty much do the same process of scanning vhd vms, just for more images. 

so if we have a vhd and after a couple weeks a new CVE is found, we have like a daily scan that will update an infra table that they're making to track cve's.

anujmaheshwari1 (approved):

PR #4556, opened 2024-06-26
cleanup: remove kube-proxy-images.json

jason1028kr:
**What type of PR is this?**
/kind cleanup

remove kube-proxy-image.json

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -381,6 +381,29 @@
       "multiArchVersions": [
         "v0.7.1"
       ]
+    },
+    {
+      "downloadURL": "mcr.microsoft.com/oss/kubernetes/kube-proxy:*",
```
cameronmeissner, line 386: doesn't seem like we're actually using this during VHD builds (install-dependencies.sh), are we not supposed to be caching this anymore?

cameronmeissner: confirmed offline that we still aim to cache this, should probably wait on merging this until after @Devinwong's change merge into master

jason1028kr: Wouldn't this loop https://github.com/Azure/AgentBaker/pull/4556/files#diff-d6aeb8e845af4c32251915a17dc7c887d73cd628262e83118b9dcd65f7728c98L312-R343 cover the added json entry under 'containerImages' key? I couldn't properly check b/c test pipeline was failing for other reasons, but I'll confirm once that's unblocked. I can also just pause until schema's finalized.

cameronmeissner: ah, yes you're right :)

approved, up to you whether you wanna merge now before Devin's changes or wait until the schema is updated

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -382,6 +382,29 @@
...
+      "amd64OnlyVersions": [],
+      "multiArchVersions": [
```
djsly, line 389: that's way too may version. after we review the Schema, we will need to keep 3 images only (one per AKS versions)

jason1028kr: I believe we'll need to keep corresponding kube-proxy version for each unique major.minor.patch k8s version

Devinwong (approved):

PR #4557, opened 2024-06-27
 feat: enable packer builds in multiple regions for linux

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -76,6 +76,7 @@ steps:
         -e AZURE_VM_SIZE=$(AZURE_VM_SIZE) \
         -e AZURE_RESOURCE_GROUP_NAME=${AZURE_RESOURCE_GROUP_NAME} \
         -e AZURE_LOCATION=${AZURE_LOCATION} \
```
djsly, line 78: is this VAR still used ?  what would be a use case for having a differnet AZURE_LOCATION and PACKER_LOCATION

cameronmeissner: I just hacked this up to see if this stuff even works, which it looks like does so far...

will revisit after further testing :)

PR #4572, opened 2024-07-01
fix: pass LOCATION env var to abe2es

cameronmeissner:
**What type of PR is this?**

right now we don't pass the LOCATION pipeline var into the abe2e environment when running the ADO pipeline, preventing us from running the suite in a separate location.

This PR fixes that issue.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/e2e.yaml
+++ b/.pipelines/e2e.yaml
@@ -58,6 +58,7 @@ jobs:
     displayName: Run AgentBaker E2E
     env:
       VHD_BUILD_ID: $(VHD_BUILD_ID)
+      LOCATION: $(LOCATION)
   - publish: $(System.DefaultWorkingDirectory)/e2e/scenario-logs
     artifact: scenario-logs
     condition: always()


.pipelines/e2e.yaml
```diff
@@ -58,6 +58,7 @@ jobs:
     displayName: Run AgentBaker E2E
     env:
       VHD_BUILD_ID: $(VHD_BUILD_ID)
+      LOCATION: $(LOCATION)
```
UtheMan, line 61: What is the default? eastus?

UtheMan (approved):

.pipelines/e2e.yaml
cameronmeissner: I set the default pipeline var value to westus3 since that's what our VHD build is currently set to, though abe2e's have been defaulting to eastus which has been causing problems since we haven't built anything there in a while

lilypan26: I think cameron mentioned he set it to westus3

PR #4581, opened 2024-07-03
chore: use canonical snapshots in official branches

anujmaheshwari1:
**What type of PR is this?**
This PR enables VHD builds from official branches to use the same publisher base image version as when they were created and apt packages from the time they were created.

Snapshots are currently only supported for 22.04 and above, therefore this PR only effectively modifies build logic from official branch for 22.04 SKUs.

Builds from any other branch and/or master are still the same, this change is no-op for them.

Test build imitating an official branch: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=97388020&view=results

publisher_base_image_version.json 
{
  "20_04-lts-cvm": "20.04.202402050",
  "22_04-lts": "22.04.202407010",
  "22_04-lts-gen2": "22.04.202407010",
  "22_04-lts-arm64": "22.04.202407010",
  "20_04-lts": "20.04.202406140",
  "20_04-lts-gen2": "20.04.202406140"
}

vhd_build_timestamp.json
{"build_timestamp": "20240702T231545Z"}

Test build imitating a regular branch: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=97387996&view=results

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/init-variables.sh
```diff
@@ -1,12 +1,38 @@
...
 CREATE_TIME="$(date +%s)"
 STORAGE_ACCOUNT_NAME="aksimages${CREATE_TIME}$RANDOM"
 
+# This variable will only be set if a VHD build is triggered from an official branch
+VHD_BUILD_TIMESTAMP=""
+
+# Check if the file exists, if it does, the build is triggered from an official branch
+if [ -f "${PUBLISHER_BASE_IMAGE_VERSION_JSON}" ]; then
+  # Ensure that the file is not empty, this will never happen since automation generates the file after each build but still have this check in place
+  if [ -s "${PUBLISHER_BASE_IMAGE_VERSION_JSON}" ]; then
```
cameronmeissner, line 17: should we maybe log something if the file is unexpectedly empty?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -566,6 +566,12 @@ done
 rm -f ./azcopy # cleanup immediately after usage will return in two downloads
 stop_watch $capture_time "Download and Process Kubernetes Packages / Extract Binaries" false
 
+# We dont call out to apt after this in our build process, therefore restoring sources.list back to its original state if we used canonical snapshot earlier
+# This will also always happen when cloud-init is applied, therefore, all AKS nodes will always have a fresh sources.list regardless
+if [[ "${OS_VERSION}" == "22.04" && -n "${VHD_BUILD_TIMESTAMP}" ]]; then
```
cameronmeissner, line 571: nit: `[ "${OS_VERSION}" == "22.04" ] && [  -n "${VHD_BUILD_TIMESTAMP}" ]`

vhdbuilder/packer/init-variables.sh
anujmaheshwari1: uh, it wont ever be since automation takes care of it. even if it is, logs here wont help much, we ll just use latest base image, to debug IF its ever empty, we'd need to look at automation where its generated.

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -102,6 +102,12 @@ else
   # so we just hold the kernel image packages for now on CVM.
   # this still allows us base image and package updates on a weekly cadence.
   if [[ "$IMG_SKU" != "20_04-lts-cvm" ]]; then
+    # Canonical snapshot is only implemented for 20.04 LTS, 22.04 LTS and 23.10 and above
+    # For 20.04, the only SKUs we support are FIPS, and it reaches out to ESM to get the packages, ESM does not have canonical snapshot support
+    # Therefore keeping this to 22.04 only for now
+    if [[ -n "${VHD_BUILD_TIMESTAMP}" && "${OS_VERSION}" == "22.04" ]]; then
```
cameronmeissner, line 108: nit: `[ "${OS_VERSION}" == "22.04" ] && [  -n "${VHD_BUILD_TIMESTAMP}" ]`

vhdbuilder/packer/install-dependencies.sh
anujmaheshwari1: eh, we have this format elsewhere in the file so unless absolutely necessary, not changing

eg https://github.com/Azure/AgentBaker/blob/952c143d3dd7febe6dec46c283b84b6c81c32e57/vhdbuilder/packer/pre-install-dependencies.sh#L123

fwiw, our format isn't consistent at all for conditionals, we need a big refactor to align it everywhere

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -102,6 +102,12 @@ else
...
+      sed -i "s#http://azure.archive.ubuntu.com/ubuntu/#https://snapshot.ubuntu.com/ubuntu/${VHD_BUILD_TIMESTAMP}#g" /etc/apt/sources.list
```
cameronmeissner, line 109: are we duplicating this in install-dependencies.sh?

anujmaheshwari1: reverting :)

anujmaheshwari1: comment in install deps

# We dont call out to apt after this in our build process, therefore restoring sources.list back to its original state if we used canonical snapshot earlier
# This will also always happen when cloud-init is applied, therefore, all AKS nodes will always have a fresh sources.list regardless

vhdbuilder/packer/init-variables.sh
```diff
@@ -1,12 +1,38 @@
...
-
 SETTINGS_JSON="${SETTINGS_JSON:-./packer/settings.json}"
+PUBLISHER_BASE_IMAGE_VERSION_JSON="${PUBLISHER_BASE_IMAGE_VERSION_JSON:-./vhdbuilder/publisher_base_image_version.json}"
+VHD_BUILD_TIMESTAMP_JSON="${VHD_BUILD_TIMESTAMP_JSON:-./vhdbuilder/vhd_build_timestamp.json}"
 SUBSCRIPTION_ID="${SUBSCRIPTION_ID:-$(az account show -o json --query="id" | tr -d '"')}"
 CREATE_TIME="$(date +%s)"
 STORAGE_ACCOUNT_NAME="aksimages${CREATE_TIME}$RANDOM"
 
+# This variable will only be set if a VHD build is triggered from an official branch
+VHD_BUILD_TIMESTAMP=""
```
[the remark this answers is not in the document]
anujmaheshwari1: IMG_VERSION comes from the pipeline yaml and its defaulted to "latest" as it should be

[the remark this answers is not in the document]
anujmaheshwari1: https://github.com/Azure/AgentBaker/blob/952c143d3dd7febe6dec46c283b84b6c81c32e57/.pipelines/.vsts-vhd-builder-release.yaml#L364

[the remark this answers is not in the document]
cameronmeissner: oh okay, forgot about that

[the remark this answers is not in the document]
anujmaheshwari1: we just update IMG_VERSION to the one we used to build the image IF building VHD from official branch, in all other cases, its latest and should be latest

vhdbuilder/packer/init-variables.sh
cameronmeissner: fair enough, thought might just be nice to have something to grep for in build logs if we ever wanna confirm why the latest image is being used when we expected something else to be used

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: we should really do that :)

vhdbuilder/packer/pre-install-dependencies.sh
cameronmeissner: why do we need to do this if we don't ever call out to apt again? and wouldn't we want to always take from the snapshot in all parts of the build?

anujmaheshwari1: because when the customers get a node from this VHD, we want them to have a regular sources.list with references to ubuntu.archive, snapshot is just for us in our build to have a consistent VHD, once customer gets a node with this VHD, we don't want to impose a snapshot on them.

this pattern is also followed for security patches

cameronmeissner (approved):

PR #4582, opened 2024-07-03
fix: update default vm size for abe2es

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/vmss.go
+++ b/e2e/vmss.go
@@ -231,7 +231,7 @@ func getBaseVMSSModel(name, sshPublicKey, customData, cseCmd string, opts *scena
 	return armcompute.VirtualMachineScaleSet{
 		Location: to.Ptr(config.Location),
 		SKU: &armcompute.SKU{
-			Name:     to.Ptr("standard_d2s_v4"),
+			Name:     to.Ptr("Standard_DS2_v2"),
 			Capacity: to.Ptr[int64](1),
 		},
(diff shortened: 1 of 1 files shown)


lilypan26 (approved):

e2e/vmss.go
```diff
@@ -231,7 +231,7 @@ func getBaseVMSSModel(name, sshPublicKey, customData, cseCmd string, opts *scena
 	return armcompute.VirtualMachineScaleSet{
 		Location: to.Ptr(config.Location),
 		SKU: &armcompute.SKU{
-			Name:     to.Ptr("standard_d2s_v4"),
+			Name:     to.Ptr("Standard_DS2_v2"),
```
Devinwong, line 234: do we not need to change the line `defaultAgentPoolVMSize     = "standard_d2s_v4"` in cluster.go?

cameronmeissner: that dictates the vm size for the system agentpool  from which we run the jumpbox pods - those are only created when new clusters are provisioned, thus I don't think we need to

cameronmeissner: if it's still an issue we can always change in another follow-up

djsly: I think it would be cleaner and lest confusion to centralized the SKU to that constant.

PR #4586, opened 2024-07-03
perf: enable shallow replication

zachary-bailey:
**What type of PR is this?**

/kind perf

This PR enables the use of shallow replication in all VHD build pipelines in order to reduce feedback time / increase productivity. Enabling shallow replication decreases pipeline duration by 10-13 minutes depending on the run. Averages derived from 3 runs from master and 3 from this branch.

This is accomplished by installing the azure-arm plugin for packer, which enables the use of shallow replication.

Unsatisfactory VHD Build Times.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests
- [ x] tested upgrade from previous version

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -3,7 +3,7 @@ trigger: none
 
 variables:
   - name: CONTAINER_IMAGE
-    value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
+    value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
 
 pool:
   name: $(POOL_NAME)
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -13,7 +13,8 @@ parameters:
     displayName: Use Shallow Replication
     type: boolean
     default: true
-    description: Uncheck this box if you want to run an E2E test in a different Azure region after this build.
+    metadata:
+      description: Uncheck this box if you want to run an E2E test in a different Azure region after this build.
   - name: useOverrides
     displayName: Use component overrides
     type: boolean
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -169,7 +169,7 @@ parameters:
 
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
+  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.39.1'
   
 pool:
   name: $(POOL_NAME)
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -169,7 +169,7 @@ parameters:
 
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.39.1'
+  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
   
 pool:
   name: $(POOL_NAME)
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -169,7 +169,7 @@ parameters:
 
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
+  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
   
 pool:
   name: $(POOL_NAME)
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -167,13 +167,6 @@ parameters:
     type: boolean
     default: False
 
-variables:
-- name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
-  
-pool:
-  name: $(POOL_NAME)
-
 stages:
   - stage: build
     jobs:
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -9,6 +9,10 @@ pool:
   name: $(POOL_NAME)
 
 parameters:
+  - name: useShallowReplication
+    displayName: Use Shallow Replication
+    type: boolean
+    default: true
   - name: useOverrides
     displayName: Use component overrides
     type: boolean
@@ -17,10 +21,6 @@ parameters:
     displayName: Branch in aks-rp to use for overrides
     type: string
     default: master
-  - name: useShallowReplication
-    displayName: Use Shallow Replication
-    type: boolean
-    default: true
   - name: build1804containerd
     displayName: Build 1804 containerd
     type: boolean
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -169,7 +169,7 @@ parameters:
 
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
+  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
   
 pool:
   name: $(POOL_NAME)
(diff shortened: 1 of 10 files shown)


vhdbuilder/packer/linux-packer-plugin.pkr.hcl
```diff
@@ -0,0 +1,8 @@
+packer {
+  required_plugins {
+    azure = {
+      version = ">= 2.1.7"
```
cameronmeissner, line 4: does the current version we use in the dev container not support shallow replication?

vhdbuilder/packer/init-variables.sh
```diff
@@ -82,6 +82,14 @@ if [ -z "${SUBNET_NAME}" ]; then
 	SUBNET_NAME="packer"
 fi
 
+SHALLOW_REPLICATION_TOGGLE="false"
+if [[ "${MODE}" == "linuxVhdMode" ]] && [[ "${ENVIRONMENT,,}" == "test" ]]; then
+  # continue using full replication for prod builds for now
+  SHALLOW_REPLICATION_TOGGLE="true"
```
cameronmeissner, line 88: after reading this once again: https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=vmsource%2Cazure-cli#shallow-replication, I don't personally think a toggle is actually necessary, should be safe to just enable to all our environments

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -83,13 +83,21 @@ echo "Granting access to $disk_resource_id for 1 hour"
...
-echo "Setting azcopy environment variables with pool identity: $AZURE_MSI_RESOURCE_STRING"
-export AZCOPY_AUTO_LOGIN_TYPE="MSI"
-export AZCOPY_MSI_RESOURCE_STRING="$AZURE_MSI_RESOURCE_STRING"
-
 echo "Uploading $disk_resource_id to ${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"
-azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd" --recursive=true || exit $?
+
+if [ "${OS_TYPE,,}" == "linux" ]; then
+  # Use azcopy login for Linux
+  azcopy login --login-type=MSI
```
cameronmeissner, line 90: is this something we have to do? can we just keep using the current approach for both OS types?

packer.mk
```diff
@@ -75,7 +75,7 @@ init-packer:
 	@./vhdbuilder/packer/init-variables.sh
 
 run-packer: az-login
-	@packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer | tee -a packer-output)
+	@packer init ./vhdbuilder/packer/linux-packer-plugin.pkr.hcl && packer version && ($(MAKE) -f packer.mk init-packer | tee packer-output) && ($(MAKE) -f packer.mk build-packer | tee -a packer-output)
```
cameronmeissner, line 78: nit: move this new command to a new line before the existing line?

vhdbuilder/packer/linux-packer-plugin.pkr.hcl
zachary-bailey: The azure-arm packer plugin did not have the shallow replication feature until v2.0.1 released in October 2023. Our packer version with the go dev container was released in August 2023, so it doesn't have the shallow replication functionality. New packer versions don't have the plugins bundled ahead of time, so I had to add this. I made a new file instead of changing the old one because windows is using that plugin.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
zachary-bailey: Removing the go dev container and using this login method was the only way I could get the convert to vhd blob step to work with Ubuntu. I messaged you on teams about my last ditch effort to continue using the container. I added the os check so that windows workflow was unchanged and they wouldnt end up with failures from my changes since their pipeline uses a container as well.

This was an issue a couple months ago too, but changing the go dev container to version 1.38.0 fixed it. Now it doesn't
help for some reason :/

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -3,12 +3,11 @@ trigger: none
 
 variables:
   - name: CONTAINER_IMAGE
-    value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
+    value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
```
djsly, line 6: rebase ?

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -19,12 +19,12 @@ pr:
     - vhdbuilder/packer/*windows*
     - vhdbuilder/packer/**/*windows*
 
-pool:
-  name: $(POOL_NAME)
-
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
+  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.0'
```
djsly, line 24: rebase

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -19,12 +19,12 @@ pr:
     - vhdbuilder/packer/*windows*
     - vhdbuilder/packer/**/*windows*
 
-pool:
-  name: $(POOL_NAME)
-
 variables:
 - name: CONTAINER_IMAGE
-  value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
```
cameronmeissner, line 27: why the downgrade here?

zachary-bailey:
It depends on how we're making image versions. If shallow replication is used to publish an image version to the gallery from a pipeline, that image version can not be replicated to other regions, and additional replicas can not be made in the source region. 

But if shallow replication is used in the pipeline, and we take the resulting image version and convert it to a VHD blob, then create an image version out of that VHD blob, it works fine. The resulting image version's replication property is marked as "Full" and it can be successfully replicated. That's why I changed it back to test env only, because I was not sure if that would be okay in prod, but it didn't seem like it would cause issues in test. 

@cameronmeissner

.pipelines/.vsts-vhd-builder.yaml
zachary-bailey: Ubuntu images will not complete the convert-to-vhd-blob step without downgrading to 1.38.0. I have not been able to figure out why.

cameronmeissner:
> It depends on how we're making image versions. If shallow replication is used to publish an image version to the gallery from a pipeline, that image version can not be replicated to other regions, and additional replicas can not be made in the source region.
> 
> But if shallow replication is used in the pipeline, and we take the resulting image version and convert it to a VHD blob, then create an image version out of that VHD blob, it works fine. The resulting image version's replication property is marked as "Full" and it can be successfully replicated. That's why I changed it back to test env only, because I was not sure if that would be okay in prod, but it didn't seem like it would cause issues in test.
> 
> @cameronmeissner

I see, yeah since we don't currently run agentbaker e2e's in prod this wouldn't have any impact, though we do run them in test

since we do run the e2e's in test, the issue would arise if we built the VHDs (e.g. ran the build pipeline) in one region, then tried to run the e2e's against said VHD build in another region, at least with the current implementation

though as I said above I think in general we should try to avoid doing that anyway, but in certain cases it might be problematic if we have to use a different region for GPU quota, for example

vhdbuilder/packer/init-variables.sh
```diff
@@ -113,6 +113,14 @@ if [ -z "${SUBNET_NAME}" ]; then
 	SUBNET_NAME="packer"
 fi
 
+USE_SHALLOW_REPLICATION="false"
+if [[ "${OS_TYPE,,}" == "linux" ]] && [[ "${ENVIRONMENT,,}" == "test" ]]; then
```
cameronmeissner, line 117: to make sure we can deal with the inability to replicate to other regions, I recommend having a new boolean pipeline parameter in `.vsts-vhd-builder.yaml`/`.vsts-vhd-builder-release.yaml` which would determine whether or not to use shallow replication, which we can default to true

if we ever need to run abe2e's against a particular build in a _different_ region than where the build takes place, we can simply disable shallow replication with this new parameter

zachary-bailey: Sounds good. Will DM.

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -110,6 +113,7 @@ steps:
       echo MODE=$(MODE) && \
       docker run --rm \
       -v ${PWD}:/go/src/github.com/Azure/AgentBaker \
+      -v /usr/local/bin/packer:/usr/local/bin/packer \
```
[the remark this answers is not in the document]
zachary-bailey: I wanted to ensure compatibility between the new packer plugin and overall packer version. Using the packer version currently in the container means we still have bundled plugins, and the plugin for azure was released before the shallow replication feature. Using the agent VMs packer binary keeps it up to date. 

I'm running a pipeline now though and it seems to be working with just installing the new azure plugin and running the older packer version. Will let you know the result when it concludes. I was also tasked with upgrading packer, but I imagine that can be closed out if there is no way to do it properly.

PR #4589, opened 2024-07-04
fix: ensure abe2e SIG image replication

cameronmeissner:
**What type of PR is this?**

adds new logic to abe2es to ensure that the selected VHDs to run the various scenarios are replicated to the particular region in which the suite is being run

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/config/vhd.go
```diff
@@ -114,29 +185,58 @@ func findLatestImageWithTag(imageID, tagName, tagValue string) (VHDResourceID, e
...
-	return imageID{
-		subscriptionID: matches[1],
-		resourceGroup:  matches[2],
-		galleryName:    matches[3],
-		imageName:      matches[4],
-	}, nil
+	return nil
+}
+
+func newGalleryImageVersionsClient(subscriptionID string) (*armcompute.GalleryImageVersionsClient, error) {
```
r2k1, line 232: currently all azure clients are in `config.Azure`.

r2k1 (approved):
Nice. I've been thinking how to add more replication regions to the packer config.
This seems to be a more elegant solution.

e2e/config/vhd.go
cameronmeissner: ahh good catch, will rework

cameronmeissner: fixed

djsly (approved):

PR #4590, opened 2024-07-04
test: simplify cluster management in e2e tests

r2k1:
- Simplify Cluster Selection and Creation: Current procedures for selecting and creating clusters are overly complex. A new approach: a cluster with a fixed name will be created or updated through a retry loop, which is designed to handle common error scenarios, such as when a cluster is in the "DELETION" phase.
- Better cluster naming improves traceability between tests and AKS clusters.
- Introduce 'KubenetAirgap' Cluster: Add a new "KubenetAirgap" cluster type, which was previously implied but not explicitly defined.
- Prevent Cluster 'Leakage': Currently, airgap clusters are created anew with each test run and only removed during routine garbage collection days later.
- Optimize for Single Test Runs: When only a single test is conducted, the system will now only create or update the necessary cluster
- Accelerate Test Execution: By reusing "airgap" cluster, we can eliminate the typical 5-10 minute wait times previously required for clusters to be prepared, speeding up the overall testing process.


e2e/cluster/cluster.go
```diff
@@ -0,0 +1,503 @@
...
+	})
+	return clusterKubenetAirgap, clusterKubenetAirgapError
+}
+
+func ClusterAzureNetwork(ctx context.Context) (*Cluster, error) {
+	clusterAzureNetworkOnce.Do(func() {
+		clusterAzureNetwork, clusterAzureNetworkError = createNewClusterWithClient(ctx, getAzureNetworkClusterModel(testClusterNamePrefix+"azure-network"))
+	})
+	return clusterAzureNetwork, clusterAzureNetworkError
+}
```
r2k1, line 102: New code is mostly in this file.

Some of the existing functions also has been moved here. So, the airgap creation logic can be referenced in scenarios without causing circular dependency issues.

e2e/cluster/cluster.go
```diff
@@ -0,0 +1,503 @@
...
+var (
+	clusterKubenet       *Cluster
+	clusterKubenetAirgap *Cluster
+	clusterAzureNetwork  *Cluster
+
+	clusterKubenetError       error
+	clusterKubenetAirgapError error
+	clusterAzureNetworkError  error
+
+	clusterKubenetOnce       sync.Once
```
cameronmeissner, line 45: could we just use a singleflight group instead of managing a bunch of different `Once`'s?

cameronmeissner: eh, maybe not, singleflight groups don't guarantee exactly-once execution, it only suppresses concurrent duplicate calls

cameronmeissner: actually, are the `Once`'s here necessary at all? don't we create missing test clusters before running any scenarios?

e2e/cluster/cluster.go
```diff
@@ -0,0 +1,503 @@
...
+
+type Kubeclient struct {
+	Dynamic client.Client
+	Typed   kubernetes.Interface
+	Rest    *rest.Config
+}
+
+// Returns true if the cluster is configured with Azure CNI
+func (c *Cluster) IsAzureCNI() (bool, error) {
+	if c.Model.Properties.NetworkProfile != nil {
```
cameronmeissner, line 63: feels kinda clunky to check for `nil`ness in so many different places...

do you think adding a single place where `nil`ness validation takes place, immediately proceeding the relevant azure calls?

e2e/cluster/cluster.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+
+	return &clusterResp.ManagedCluster, nil
+}
+
+// createNewClusterWithRetry is a wrapper around createNewCluster
+// that retries creating a cluster if it fails with a 409 Conflict error
+// clusters are reused, and sometimes a cluster can be in UPDATING or DELETING state
+// simple retry should be sufficient to avoid such conflicts
+func createNewClusterWithRetry(ctx context.Context, cluster *armcontainerservice.ManagedCluster) (*armcontainerservice.ManagedCluster, error) {
```
cameronmeissner, line 148: can we turn this into a poller and add to `pollers.go` instead?

e2e/cluster/cluster.go
```diff
@@ -0,0 +1,503 @@
...
+		} else {
+			// If it's not a 409 error, return immediately
+			return nil, fmt.Errorf("failed to create cluster: %w", err)
+		}
+	}
+
+	return nil, fmt.Errorf("failed to create cluster after %d attempts due to persistent 409 Conflict: %w", maxRetries, lastErr)
+}
+
+func getKubenetClusterModel(name string) *armcontainerservice.ManagedCluster {
```
cameronmeissner, line 181: I'd opt for having a separate file dedicated to holding all these different mutators we use for modifying cluster models

e2e/cluster/cluster.go
```diff
@@ -0,0 +1,503 @@
...
+    metadata:
+      labels:
+        app: *name
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.azure.com/agentpool: nodepool1
+      hostPID: true
+      containers:
+      - image: mcr.microsoft.com/oss/nginx/nginx:1.21.6
```
cameronmeissner, line 492: tangential, though we should probably update this image tag at some point

e2e/suite_test.go
```diff
@@ -65,32 +57,23 @@ func maybeSkipScenario(t *testing.T, s *scenario.Scenario) {
...
-	clusterConfig, err := chooseCluster(ctx, r, e2eScenario, clusterConfigs)
-	if err != nil {
-		t.Fatal(err)
-	}
+func setupAndRunScenario(ctx context.Context, t *testing.T, e2eScenario *scenario.Scenario, clusterConfig *cluster.Cluster) {
+	log.Printf("chose cluster: %q", *clusterConfig.Model.ID)
 
-	clusterName := *clusterConfig.cluster.Name
-	log.Printf("chose cluster: %q", clusterName)
+	clusterParams, err := pollExtractClusterParameters(ctx, clusterConfig.Kube)
```
cameronmeissner, line 63: I think we can do this once per-cluster before any scenarios are executed, probably with new fields on `cluster.Cluster`

e2e/vmss.go
```diff
@@ -24,38 +24,36 @@ const (
...
+		})
+	}
+
+	return vmssModel, nil
 }
 
 func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName string, publicKeyBytes []byte, opts *scenarioRunOpts) (*armcompute.VirtualMachineScaleSet, error) {
-	model, err := getBaseVMSSModel(vmssName, string(publicKeyBytes), customData, cseCmd, opts)
+	log.Printf("creating VMSS %q in resource group %q", vmssName, *opts.clusterConfig.Model.Properties.NodeResourceGroup)
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Minute)
```
cameronmeissner, line 54: I'd refactor the timeout duration to a separate const

e2e/vmss.go
```diff
@@ -82,55 +80,46 @@ func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName str
...
+		vmssName,
+		model,
+		nil,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("begin create VMSS %q: %w", vmssName, err)
 	}
-	return nil, fmt.Errorf("create VMSS %q: %w", vmssName, pollErr)
+	vmssResp, err := operation.PollUntilDone(ctx, &runtime.PollUntilDoneOptions{
+		Frequency: 10 * time.Second,
```
cameronmeissner, line 94: the retry logic here definitely needs a refactor, though we should keep it since we used to normally see transient CRP-related errors occur quite frequently when creating the VMSS

e2e/vmss.go
```diff
@@ -82,55 +80,46 @@ func createVMSSWithPayload(ctx context.Context, customData, cseCmd, vmssName str
...
 // Additional info: https://learn.microsoft.com/en-us/azure/aks/configure-azure-cni
-func addPodIPConfigsForAzureCNI(vmss *armcompute.VirtualMachineScaleSet, vmssName string, opts *scenarioRunOpts) error {
-	maxPodsPerNode, err := opts.clusterConfig.maxPodsPerNode()
+func addPodIPConfigsForAzureCNI(ctx context.Context, vmss *armcompute.VirtualMachineScaleSet, vmssName string, opts *scenarioRunOpts) error {
+	maxPodsPerNode, err := opts.clusterConfig.MaxPodsPerNode()
 	if err != nil {
 		return fmt.Errorf("failed to read agentpool MaxPods value from chosen cluster model: %w", err)
 	}
 
+	vnet, err := getClusterVNet(ctx, *opts.clusterConfig.Model.Properties.NodeResourceGroup)
```
cameronmeissner, line 111: could this also be something we instantiate on the cluster object before running any scenarios?

cameronmeissner: especially since we also use it in `getBaseVMSSModel`

e2e/cluster/cluster.go
r2k1: I don't mind a small number of duplicates. Especially when they located close to each other.
I can avoid id with something similar to this https://github.com/Azure/AgentBaker/blob/master/e2e/config/vhd.go#L83-L100

But it will introduce additional complexity and makes it hard to understand. 3 clusters aren't much to maintain.

e2e/cluster/cluster.go
r2k1: This function called during each VMSS creation. I don't know details, but it seems like VMSS needs to be configured differently based on the flag.

e2e/vmss.go
r2k1: Is there a good reason for it?
I personally prefer when single-use magical numbers are inlined. So, when I read the function, I don't need to jump between different places to get the values.

e2e/cluster/cluster.go
r2k1: Currently it's a specialized method for cluster creation and not the complicated.

Generalizing it adds extra complexity. And the usage of overly generalized things can be more complicated.

Maybe there is a good well-known way to handle it or a library. But I don't see how I can really simplify it.

e2e/vmss.go
r2k1: What are CRP errors?

I tried to think about errors I'd like to retry on. I could only think about network errors. But it applies to any API call.

e2e/vmss.go
```diff
@@ -185,8 +174,8 @@ func getVMPrivateIPAddress(ctx context.Context, subscription, mcResourceGroupNam
 }
 
 // Returns a newly generated RSA public/private key pair with the private key in PEM format.
-func getNewRSAKeyPair(r *mrand.Rand) (privatePEMBytes []byte, publicKeyBytes []byte, e error) {
-	privateKey, err := rsa.GenerateKey(r, 4096)
+func getNewRSAKeyPair() (privatePEMBytes []byte, publicKeyBytes []byte, e error) {
+	privateKey, err := rsa.GenerateKey(crand.Reader, 4096)
```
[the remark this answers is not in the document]
r2k1: I actually don't get why we passed it around. Methods on`*mrand.Rand` isn't thread safe btw.

e2e/cluster/cluster.go
r2k1: > actually, are the Once's here necessary at all? don't we create missing test clusters before running any scenarios?

This logic has changed a bit.

I have another long-term goal. I'd like each test to be independently executable from the others. I want to try organizing tests differently, with better integration of dev tooling.

Currently, there is a large, shared preparation step, and essentially a cluster setup for one test can break other tests that don't require such a cluster.

e2e/suite_test.go
r2k1: it's actually requires moving a lot of code between packages. I have some ideas how to improve it.
But I don't want to add 2k LOC diff to the PR.

Is it ok if I optimize it later?

e2e/vmss.go
r2k1: we also have retries enabled on the VMSS client itself.

Chain retries in two places is going to multiple total attempt amount.

r2k1: Once I chained 3 retries in different parts of a system. Innocent 10 attempts become a 10^3 = 1000 attempts spam.

e2e/cluster/cluster.go
cameronmeissner: don't worry about it for now, I mainly just meant that it would be nice if we could somehow consolidate where we check for `nil` on certain objects

e2e/vmss.go
cameronmeissner: fair enough, since I guess we're only using it here

e2e/vmss.go
cameronmeissner: so what's annoying about this is that the azure client does not retry service errors automatically, we need logic on top of that layer in order to determine when we should retry, for example when CRP flakes

CRP = compute resource provider, so basically any error related to VM creation that's server-side, e.g. `AllocationFailed`

cameronmeissner (approved):

e2e/vmss.go
r2k1:
```
	StatusCodes: []int{
			http.StatusRequestTimeout,      // 408
			http.StatusTooManyRequests,     // 429
			http.StatusInternalServerError, // 500
			http.StatusBadGateway,          // 502
			http.StatusServiceUnavailable,  // 503
			http.StatusGatewayTimeout,      // 504
			http.StatusNotFound,            // 404
		},
```

Maybe an http status code is missing in this list?

r2k1: I think `AllocationFailed` is a bad error to retry on. If we don't have capacity now, it's likely we will get in a minute

PR #4591, opened 2024-07-05
test: allow filtering tests by a tag

r2k1:
With this change it should be easier to temporary disable some of the tests.

For example, in a case of GPU capacity constraint all gpu scenarios can be disabled by setting `TAGS_TO_SKIP="gpu=true"`

I also noticed that I didn't update the README.md. Updated with an assumption #4590 is merged.


cameronmeissner:
many thanks for this, I've been meaning to do this for the longest time

e2e/scenario/scenario.go
```diff
@@ -54,5 +54,6 @@ func AllScenarios() []*Scenario {
 		ubuntu2204gpu("ubuntu2204-gpu-ncv3", "Standard_NC6s_v3"),
 		ubuntu2204gpu("ubuntu2204-gpu-a100", "Standard_NC24ads_A100_v4"),
 		ubuntu2204gpu("ubuntu2204-gpu-a10", "Standard_NV6ads_A10_v5"),
+		ubuntu2204GPUGridDriver(),
```
cameronmeissner, line 57: oh... I guess we weren't ever running this one lol

cameronmeissner (approved):

PR #4592, opened 2024-07-08
test: Skip tests with SKU capacity issues

r2k1:
Sometimes, there are simply no available nodes with GPUs. In such cases, there's nothing we can do.
The tests are failing, and this can last for weeks. Tests are red, and it builds a habit that red is acceptable.

This change explicitly does what we're currently doing implicitly.

One caveat: we shouldn't ignore tests for a long period. With this change, it's easy to miss them.

Note. Quota errors has `ERROR CODE: OperationNotAllowed`. Such errors will continue failing tests.


djsly (approved):

e2e/config/config.go
```diff
@@ -6,24 +6,31 @@ import (
...
-func getEnvWithDefaultIfEmpty(env string, defaultValue string) string {
+func lookupEnvWithDefaultString(env string, defaultValue string) string {
 	val, ok := os.LookupEnv(env)
 	if !ok {
 		return defaultValue
 	}
 	return val
 }
+
+func lookupEnvWithDefaultBool(env string, defaultValue bool) bool {
```
cameronmeissner, line 31: don't have to do it now, but would be nice if we could use generics here instead

cameronmeissner (approved):
thanks :)

e2e/config/config.go
r2k1: I find generics are hard to read. Maybe when we have more than 3 types.

r2k1: I don't think generics are going to work. We perform a string to a type conversion. Which is different for every type.

PR #4594, opened 2024-07-08
feat: use PowerShell commands to set pagefile size instead of wmic

AbelHu:
**What type of PR is this?**
/kind feature

wmic will be removed in WS2025.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


staging/cse/windows/configfunc.ps1
```diff
@@ -63,7 +63,18 @@ function Adjust-PageFileSize()
 {
     Logs-To-Event -TaskName "AKS.WindowsCSE.AdjustPageFileSize" -TaskMessage "Start to adjust pagefile size"
 
-    wmic pagefileset set InitialSize=8096,MaximumSize=8096
+    try {
```
ShiqianTao, line 66: Not a blocker: should the new code be enabled only for WS2015?
```
if (2015) { new }
else { old }
```

ShiqianTao (approved):

staging/cse/windows/configfunc.ps1
AbelHu: These powershell commands work in all Window OS version AKS supports so I think that we should not differentiate them.

PR #4596, opened 2024-07-08
refactor: one trivy report per container image

jason1028kr:
**What type of PR is this?**
/kind refactor

- For container images, instead of producing single table that includes all trivy report, produce single report in json format for each container image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

- checkin build: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=99040419&view=results
- There will be ~100 reports produced for container images.
Screen Shot 2024-07-25 at 12 52 59 PM

**Release note**:

```
none
```


vhdbuilder/packer/vhd-scanning-exe-on-vm.sh
```diff
@@ -45,14 +47,18 @@ else
 fi
 
 # trivy scan must have run before this
-az storage blob upload --file /opt/azure/containers/trivy-report.json \
+az storage blob upload --file /opt/azure/containers/trivy-report-rootfs.json \
     --container-name ${SIG_CONTAINER_NAME} \
-    --name ${TRIVY_REPORT_NAME} \
+    --name ${TRIVY_REPORT_ROOTFS_NAME} \
```
djsly, line 52: can we avoid the extra var and align the setting to be in place like you did for the blob upload below ?

`--name ${IMAGE_REPORT}-${BUILD_ID}-${TIMESTAMP}.json`

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -93,6 +93,8 @@ az vm run-command invoke \
 TIMESTAMP=$(date +%s%3N)
 TRIVY_REPORT_NAME="trivy-report-${BUILD_ID}-${TIMESTAMP}.json"
 TRIVY_TABLE_NAME="trivy-table-${BUILD_ID}-${TIMESTAMP}.txt"
+#"TRIVY_REPORT_NAME=${TRIVY_REPORT_NAME}" \
```
djsly, line 96: remove if not needed

djsly (changes requested):

vhdbuilder/packer/trivy-scan.sh
```diff
@@ -16,26 +16,23 @@ else
...
-for image in $IMAGE_LIST; do
-    ./trivy --scanners vuln image --ignore-unfixed --severity HIGH,CRITICAL -f table $image >> ${TRIVY_REPORT_TABLE_PATH} || true
+for CONTAINER_IMAGE in $IMAGE_LIST; do
+    TRIVY_REPORT_IMAGE_JSON_PATH=${TRIVY_REPORT_DIRNAME}/trivy-report-image-${CONTAINER_IMAGE}.json
+    ./trivy --scanners vuln image -f json --ignore-unfixed --severity HIGH,CRITICAL -o ${TRIVY_REPORT_IMAGE_JSON_PATH} $CONTAINER_IMAGE  || true
+    chmod a+r "${TRIVY_REPORT_IMAGE_JSON_PATH}"
 done
 
-rm ./trivy 
+rm ./trivy
```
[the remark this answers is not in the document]
jason1028kr: Wouldn't this be better for version update flow?

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -103,15 +103,14 @@ az vm run-command invoke \
...
-        "TRIVY_REPORT_NAME=${TRIVY_REPORT_NAME}" \
-        "TRIVY_TABLE_NAME=${TRIVY_TABLE_NAME}" \
+        "BUILD_ID=${BUILD_ID}" \
+        "TIMESTAMP=${TIMESTAMP}" \
         "SIG_CONTAINER_NAME"=${SIG_CONTAINER_NAME} \
         "STORAGE_ACCOUNT_NAME"=${STORAGE_ACCOUNT_NAME} \
         "ENABLE_TRUSTED_LAUNCH"=${ENABLE_TRUSTED_LAUNCH}
 
-
+az storage blob download-batch -d . --pattern "trivy-report-image-*-${BUILD_ID}-${TIMESTAMP}.json" -s ${SIG_CONTAINER_NAME}  --account-name ${STORAGE_ACCOUNT_NAME} --auth-mode login
```
jason1028kr, line 112: need to check batch works as expected

vhdbuilder/packer/trivy-scan.sh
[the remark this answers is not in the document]
jason1028kr: Also since we're doing the trivy scan within the spun-up vm, I am not sure if leaving the binary for another vm to reuse would be a good idea?

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -176,14 +176,14 @@ steps:
       SourceFolder: '$(System.DefaultWorkingDirectory)'
       Contents: 'trivy-report.json'
       TargetFolder: '$(Build.ArtifactStagingDirectory)'
-    displayName: Copy trivy report
+    displayName: Copy trivy rootfs report
   - task: CopyFiles@2
     condition: eq(variables['IS_NOT_1804'], 'true')
     inputs:
       SourceFolder: '$(System.DefaultWorkingDirectory)'
-      Contents: 'trivy-images-table.txt'
```
AlisonB319, line 184: Should we remove the table.txt all together? Wondering if it would be useful in the future if someone is manually looking at a run. Or would they just view their test build scans in kusto? This is assuming that the table.txt would have all the information that is in all of the per-container jsons (correct me if that would not be the case)

jason1028kr: I see, we would still have an artifact for each container image scan. It's just that each appended table entry would be separated into individual json format. If you think having one table.txt file would be necessary, I can keep that logic on top of jsons.

PR #4607, opened 2024-07-09
fix: align staging gallery image definitions with prod for abe2es

cameronmeissner:
**What type of PR is this?**

changes the name of the definitions to which packer delivers the output VHDs within the staging gallery. this is to ensure that we preserve a 1:1 mapping between SKU and image definition within our staging gallerie(s), as having multiple SKUs mapping to the same image definition was causing problems related to 1804gen2containerd + 1804gen2gpucontainerd tests

also adds some extra logic to abe2e image selection logic to ignore images in a non-Succeeded provisioning state

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/config/vhd.go
```diff
@@ -21,17 +21,18 @@ const (
...
-	VHDUbuntu1804Gen2Containerd      = newSIGImageVersionResourceIDFetcher(imageGallery + "1804Gen2")
-	VHDUbuntu2204Gen2Arm64Containerd = newSIGImageVersionResourceIDFetcher(imageGallery + "2204Gen2Arm64")
-	VHDUbuntu2204Gen2Containerd      = newSIGImageVersionResourceIDFetcher(imageGallery + "2204Gen2")
-	VHDAzureLinuxV2Gen2Arm64         = newSIGImageVersionResourceIDFetcher(imageGallery + "AzureLinuxV2Gen2Arm64")
-	VHDAzureLinuxV2Gen2              = newSIGImageVersionResourceIDFetcher(imageGallery + "AzureLinuxV2Gen2")
-	VHDCBLMarinerV2Gen2Arm64         = newSIGImageVersionResourceIDFetcher(imageGallery + "CBLMarinerV2Gen2Arm64")
-	VHDCBLMarinerV2Gen2              = newSIGImageVersionResourceIDFetcher(imageGallery + "CBLMarinerV2Gen2")
-
-	// this is a particular 2204Gen2 image originally built with private packages,
+	VHDUbuntu1804Gen2Containerd      = newSIGImageVersionResourceIDFetcher(imageGallery + "1804gen2containerd")
```
UtheMan, line 24: Why is the addition of containerd important here?

UtheMan: Also could you add some short description of the problem and why its solved in the PR? By itself its a bit confusing imo

cameronmeissner: just aligning towards what we have in prod to make things less confusing/more consistent

cameronmeissner: will update description before publishing

vhdbuilder/packer/init-variables.sh
```diff
@@ -145,39 +152,27 @@ if [[ "${MODE}" == "linuxVhdMode" ]]; then
...
-			SIG_IMAGE_NAME=AzureLinux${SIG_IMAGE_NAME}
-		fi
-
-		if [[ "${ENABLE_TRUSTED_LAUNCH}" == "True" ]]; then
-			SIG_IMAGE_NAME=${SIG_IMAGE_NAME}TL
-		fi
-
-		if [[ "${HYPERV_GENERATION,,}" == "v2" && ("${OS_SKU}" == "CBLMariner" || "${OS_SKU}" == "AzureLinux" || "${OS_SKU}" == "Ubuntu") ]]; then
-			SIG_IMAGE_NAME=${SIG_IMAGE_NAME}Gen2
+		SIG_IMAGE_NAME=$SKU_NAME
```
lilypan26, line 156: SKU_NAME is set here right? https://github.com/Azure/AgentBaker/blob/master/.pipelines/templates/.builder-release-template.yaml#L54C3-L69C30

is init-variables.sh run after this?

e2e/vmss.go
```diff
@@ -223,10 +220,10 @@ func getVmssName(r *mrand.Rand) string {
 	return fmt.Sprintf(vmssNameTemplate, randomLowercaseString(r, 4))
 }
 
-func getBaseVMSSModel(name, sshPublicKey, customData, cseCmd string, opts *scenarioRunOpts) (armcompute.VirtualMachineScaleSet, error) {
+func getBaseVMSSModel(name, sshPublicKey, customData, cseCmd string, opts *scenarioRunOpts) armcompute.VirtualMachineScaleSet {
 	resourceID, err := config.VHDUbuntu1804Gen2Containerd()
 	if err != nil {
-		return armcompute.VirtualMachineScaleSet{}, fmt.Errorf("get resource ID for VHD: %w", err)
+		log.Printf("get resource ID for VHD, will not set default VHD within base VMSS model: %s", err)
```
lilypan26, line 226: why don't we want to return error here anymore?

cameronmeissner: if a scenario wants to use any VHD other than the default, and it so happens that other VHD does in fact exist, we'd still want the scenario to be runnable

right now, if it's the case that this default VHD doesn't exist, then non of the scenarios will end up being run

vhdbuilder/packer/init-variables.sh
cameronmeissner: yep, init-variables.sh is executed as a part of the `run-packer` make target

djsly (approved):

PR #4615, opened 2024-07-10
feat: install oras on the VHD

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -165,6 +165,10 @@ if [[ $OS == $MARINER_OS_NAME ]]; then
     activateNfConntrack
 fi
 
+# installs oras for installing packages for network isolated clusters
```
AlisonB319, line 168: TODO code changes here, to get an idea of what will need to be done for NIC

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -82,6 +82,22 @@ downloadCredentalProvider() {
     retrycmd_get_tarball 120 5 "$CREDENTIAL_PROVIDER_DOWNLOAD_DIR/$CREDENTIAL_PROVIDER_TGZ_TMP" "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
 }
 
+downloadCredentalProviderWithOras() {
```
djsly, line 85: typo ? credential ?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -82,6 +82,22 @@ downloadCredentalProvider() {
...
+    mkdir -p $CREDENTIAL_PROVIDER_DOWNLOAD_DIR
+    CREDENTIAL_PROVIDER_TGZ_TMP=${CREDENTIAL_PROVIDER_DOWNLOAD_URL##*/}
+    oras pull $CREDENTIAL_PROVIDER_DOWNLOAD_URL -o $CREDENTIAL_PROVIDER_TGZ_TMP || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
+}
+
+installOras() {
+    ORAS_VERSION="1.2.0"
```
djsly, line 92: can you add a TODO , we should have this DEP managed by dependant or Renovate

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -82,6 +82,22 @@ downloadCredentalProvider() {
     retrycmd_get_tarball 120 5 "$CREDENTIAL_PROVIDER_DOWNLOAD_DIR/$CREDENTIAL_PROVIDER_TGZ_TMP" "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
 }
 
+downloadCredentalProviderWithOras() {
+    mkdir -p $CREDENTIAL_PROVIDER_DOWNLOAD_DIR
+    CREDENTIAL_PROVIDER_TGZ_TMP=${CREDENTIAL_PROVIDER_DOWNLOAD_URL##*/}
+    oras pull $CREDENTIAL_PROVIDER_DOWNLOAD_URL -o $CREDENTIAL_PROVIDER_TGZ_TMP || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
+}
+
+installOras() {
```
djsly, line 91: why is this during cloud init and not during Packaker build ?

cameronmeissner: note that we source cse_install.sh during builds though: https://github.com/Azure/AgentBaker/blob/04f13a59a305b8709a8d936a94dc8aa1143e77b2/vhdbuilder/packer/install-dependencies.sh#L17C1-L17C42

cameronmeissner: I'd say it would definitely be worth it to install oras during the build process, especially if it's critical to downloading required dependencies at runtime

vhdbuilder/packer/install-dependencies.sh
cameronmeissner: cse_main.sh and install-dependencies.sh both call `installDeps`, thus this shouldn't be necessary

you could also keep it separate from `installDeps` and call `installOras` here like you're doing, as well as within `cse_main.sh`

either way you should also make sure `installDeps` is idempotent, in that it doesn't actually download anything over the wire when oras is already installed

PR #4628, opened 2024-07-11
test: add tests to check content of sources.list

anujmaheshwari1:
**What type of PR is this?**
This PR adds a test to post content VHD build to ensure that in case VHD is built from an official branch, any snapshot references don't exist in the /etc/apt/sources.list

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -342,6 +342,21 @@ testCloudInit() {
   echo "$test:Finish"
 }
 
+testSourcesList() {
+  test="testSourcesList"
+  sourcesListPath="/etc/apt/sources.list"
+  searchWord="snapshot"
+  # This file should always exist on any Ubuntu node, however, still check its existence for sanity
+  if test -f "${sourcesListPath}"; then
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -342,6 +342,23 @@ testCloudInit() {
   echo "$test:Finish"
 }
 
+testSourcesList() {
+  test="testSourcesList"
+  sourcesListPath="/etc/apt/sources.list"
+  searchWord="snapshot"
+  # This file should always exist on any Ubuntu node, however, still check its existence for sanity
+  if test -f "${sourcesListPath}"; then
```
cameronmeissner, line 350: should we err if this file doesn't exist?

cameronmeissner (approved):

vhdbuilder/packer/test/linux-vhd-content-test.sh
djsly: I guess the test wants to ensure that the file doesn't content any snapshot references., so if there file doesnt' exist it should be fine.  

If we believe this file should always be present I would add another test which verify for this explicitly.

PR #4652, opened 2024-07-15
fix: capture_benchmark function fix for component downloads

zachary-bailey:
**What type of PR is this?**

/kind bug

This PR enables the `capture_benchmark` function to handle arguments that contain whitespaces and hyphens. The code change results in these characters being converted into underscores and the function successfully continuing. 

Benchmarks for the install-dependencies.sh script are not currently available because some new variable names contain hyphens which is breaking the function.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


Devinwong (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -413,7 +413,10 @@ check_array_size() {
 
 capture_benchmark() {
   set +x
-  benchmarks+=($1)
```
cameronmeissner, line 416: for future reference, let's avoid using parameters directly within logic, e.g. always assign to a local variable to give the parameter a meaningful name before using elsewhere within the function/script.

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: Will do!

PR #4653, opened 2024-07-15
test: don't trigger e2e test on non e2e file changes

r2k1:
The pipeline doesn't build new VHD. So, triggering it on VHD changes is pointless.

There is another pipeline that run e2e after VHD build.

### DIFF
--- a/.pipelines/e2e.yaml
+++ b/.pipelines/e2e.yaml
@@ -13,12 +13,7 @@ pr:
       - .pipelines/e2e.yaml
       - .pipelines/templates/e2e-template.yaml
       - e2e
-      - parts/linux
-      - pkg/agent
     exclude:
       - e2e/windows
-      - pkg/agent/datamodel/sig_config*.go # SIG config changes
-      - pkg/agent/datamodel/*.json # SIG version changes
-      - pkg/agent/testdata/AKSWindows* # Windows test data
 jobs:
   - template: ./templates/e2e-template.yaml


.pipelines/e2e.yaml
```diff
@@ -13,12 +13,7 @@ pr:
       - .pipelines/e2e.yaml
       - .pipelines/templates/e2e-template.yaml
       - e2e
-      - parts/linux
```
Devinwong, line 16: Sorry. I am not familiar with the syntax here. Could you briefly explain why removing these lines can achieve the purpose `not to trigger e2e test on non e2e file changes`

r2k1: This part defines what files in PR should trigger the pipeline. Currently it can be triggered quite often.

For example, when you change a CSE script. But a new VHD image isn't built in this pipeline. So, the e2e tests are run on master branch VHDs. Which is pointless, as they don't test any changes in the PR submitted.

With the change we will trigger the pipeline only on e2e test file updates.

Devinwong: I see. Thanks.

Devinwong (approved):

.pipelines/e2e.yaml
cameronmeissner, line 16: we will want to trigger on changes to files within these directories, as they contain the core templates used by agentbakersvc to generate CSE/customdata

cameronmeissner (changes requested):
do we already require the VHD check-in pipeline be triggered off the paths you're removing from here?

cameronmeissner:
I also think we should keep the `pkg/agent` trigger here, as changes to that particular directory imply changes to agentbakersvc code/APIs that must be tested, irregardless of which VHD is being used

r2k1:
Seems like I misunderstood some agent baker parts

PR #4660, opened 2024-07-17
Add a script to auto add Kubernetes patch versions

gaopenghigh:
**What type of PR is this?**

/kind feature

From AKS v20240627, we start to support all Kubernetes patch versions if a minor version is supported, that means we need to update VHD cache more frequently than before. This PR adds a script to help automate adding new patch versions. Now we just need to run `make auto-add-k8s-patch-versions`, if everything is good, then `make generate`.

To keep the latest 2 patch versions always cached in VHD, we actually cache 3 patch versions so latest 2 still in cache even if RP side haven't been updated.

This PR also contains the result of the script: `1.30.0` is added back to cache.  

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


ShiqianTao:
Need to rebase master, but there are conflicts

gaopenghigh:
> Need to rebase master, but there are conflicts

Thank you  @ShiqianTao , rebased.

PR #4662, opened 2024-07-18
feat: Add option to configure --seccompDefault property on kubelet

r2k1:
https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration

Allow user to configure extra kubelet parameter.

Test it manualy:

I created a new nodepool using e2e test. And created next DaemonSet:
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: violating-daemonset
  labels:
    app: violating-app
spec:
  selector:
    matchLabels:
      app: violating-app
  template:
    metadata:
      labels:
        app: violating-app
    spec:
      containers:
        - name: seccomp-test
          image: alpine:latest
          command: ["/bin/sh", "-c"]
          args:
            - |
              echo "Attempting to use unshare..."
              unshare -r echo "This should fail"
              echo "Exit code: $?"
#          securityContext:
#            seccompProfile:
#              type: RuntimeDefault
```

I received different log message on seccomp-enabled node.

But I don't think it's something worth adding a new test to e2e suite.

### DIFF
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -710,6 +710,7 @@ type CustomKubeletConfig struct {
 	ContainerLogMaxSizeMB *int32    `json:"containerLogMaxSizeMB,omitempty"`
 	ContainerLogMaxFiles  *int32    `json:"containerLogMaxFiles,omitempty"`
 	PodMaxPids            *int32    `json:"podMaxPids,omitempty"`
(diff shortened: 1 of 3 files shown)


Devinwong (approved):

UtheMan:
Is the aks-rp part of this already done as well?

r2k1:
> Is the aks-rp part of this already done as well?

I'll update it later

PR #4687, opened 2024-07-23
chore: modify AgentBaker windows test

junjiezhang1997:
**What type of PR is this?**
/kind improvement

1. Clarify workflow titles.
2. Support 23H2 in e2e pipeline.
3. Only use 2022 gen1 and 23H2 gen2 as default to reduce cost.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -53,7 +55,8 @@ jobs:
           export WINDOWS_PACKAGE_VERSION=$KUBERNETES_VERSION
         fi
         export K8S_VERSION=${WINDOWS_PACKAGE_VERSION//./}
-        export WINDOWS_E2E_STORAGE_CONTAINER=${{ parameters.storageAccount }}-$K8S_VERSION
```
AbelHu, line 56: Does the old code not work?

junjiezhang1997: “23H2” contains an uppercase letter, but a valid storage account name shouldn't contain uppercase letters. So I use "declare -l" to change it to lowercase letters.

.pipelines/templates/e2e-windows-template.yaml
```diff
@@ -53,7 +55,8 @@ jobs:
...
+        declare -l WINDOWS_E2E_STORAGE_CONTAINER=${{ parameters.storageAccount }}-$K8S_VERSION
```
AbelHu, line 58: Do we cleanup the storage account and containers?

AbelHu: I found some code but it only touches ws2019
    if [ "$WINDOWS_E2E_IMAGE" == "2019-containerd" ]; then
        cleanupOutdatedFiles
    fi

junjiezhang1997: We don't cleanup the storage account. We just cleanup the containers when the related k8s version is not supported any more in https://github.com/Azure/AgentBaker/blob/master/e2e/windows/e2e-helper.sh#L19

AbelHu (approved):
LGTM. Thanks for the PR.

PR #4688, opened 2024-07-23
chore: bump containerd patch version to .20

UtheMan:
**What type of PR is this?**

Bumps containerd version used on AKS nodes to 1.7.20r

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong (approved):
Looks good to me

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -443,7 +443,7 @@
         "ubuntu": {
```
juan-lee, line 443: what about azurelinux?

UtheMan: I believe azure linux would fall under the "default" case, is that right @Devinwong ?

Devinwong: yes, if there is "mariner", it will use the mariner one. Otherwise it will use the default one

juan-lee (approved):

parts/linux/cloud-init/artifacts/components.json
djsly: so can we remove the `ubuntu-->current` is this should be the same as the default ? I"m guessing the property `ubuntu` is mainly for `r1804` ?

UtheMan:
Merging bypassing the failed e2e run that used outdated VHD (old containerd version)

PR #4691, opened 2024-07-23
feat: add oras binary to cache

AlisonB319:
Cache Oras on the VHD

**What type of PR is this?**

**What this PR does / why we need it**: ORAS will be needed for installing packages for network isolated clusters 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

Oras is included in the download section of release-notes

```
kubelet/kubectl downloaded:
-rwxr-xr-x 1 root root    55331649 Dec 12  2023 /usr/local/bin/crictl
-rwxr-xr-x 1 root root   108036096 Apr 17 01:01 /usr/local/bin/kubelet-1.27.13
-rwxr-xr-x 1 root root    48652288 Apr 17 01:01 /usr/local/bin/kubectl-1.27.13
-rwxr-xr-x 1 root root   110014464 Apr 17 01:03 /usr/local/bin/kubelet-1.28.9
-rwxr-xr-x 1 root root    49209344 Apr 17 01:03 /usr/local/bin/kubectl-1.28.9
-rwxr-xr-x 1 root root   111939584 Apr 18 17:03 /usr/local/bin/kubelet-1.29.4
-rwxr-xr-x 1 root root    49819648 Apr 18 17:03 /usr/local/bin/kubectl-1.29.4
-rwxr-xr-x 1 root root   108036096 May 22 15:51 /usr/local/bin/kubelet-1.27.14
-rwxr-xr-x 1 root root    48652288 May 22 15:51 /usr/local/bin/kubectl-1.27.14
-rwxr-xr-x 1 root root   110014464 May 22 15:53 /usr/local/bin/kubelet-1.28.10
-rwxr-xr-x 1 root root    49209344 May 22 15:53 /usr/local/bin/kubectl-1.28.10
-rwxr-xr-x 1 root root   111939584 May 22 15:53 /usr/local/bin/kubelet-1.29.5
-rwxr-xr-x 1 root root    49819648 May 22 15:53 /usr/local/bin/kubectl-1.29.5
-rwxr-xr-x 1 root root   100100024 May 24 08:44 /usr/local/bin/kubelet-1.30.1
-rwxr-xr-x 1 root root    51454104 May 24 08:44 /usr/local/bin/kubectl-1.30.1
-rwxr-xr-x 1 1001 docker   9863303 May 28 07:33 /usr/local/bin/oras
-rwxr-xr-x 1 root root   100124920 Jun 12 16:43 /usr/local/bin/kubelet-1.30.2
-rwxr-xr-x 1 root root    51454104 Jun 12 16:43 /usr/local/bin/kubectl-1.30.2
-rwxr-xr-x 1 root root   108036096 Jun 13 17:20 /usr/local/bin/kubelet-1.27.15
-rwxr-xr-x 1 root root    48652288 Jun 13 17:20 /usr/local/bin/kubectl-1.27.15
-rwxr-xr-x 1 root root   110014464 Jun 13 17:22 /usr/local/bin/kubelet-1.28.11
-rwxr-xr-x 1 root root    49209344 Jun 13 17:22 /usr/local/bin/kubectl-1.28.11
-rwxr-xr-x 1 root root   111943680 Jun 17 17:18 /usr/local/bin/kubelet-1.29.6
-rwxr-xr-x 1 root root    49819648 Jun 17 17:18 /usr/local/bin/kubectl-1.29.6
-rwxr-xr-x 1 root root         705 Jul 23 19:55 /usr/local/bin/ci-syslog-watcher.sh
-rwxr-xr-x 1 root root    36014944 Jul 23 19:58 /usr/local/bin/containerd-shim-spin-v0-3-0-v1
-rwxr-xr-x 1 root root    39369104 Jul 23 19:58 /usr/local/bin/containerd-shim-wws-v0-8-0-v1
-rwxr-xr-x 1 root root    45334640 Jul 23 19:58 /usr/local/bin/containerd-shim-spin-v0-5-1-v1
-rwxr-xr-x 1 root root    47622592 Jul 23 19:58 /usr/local/bin/containerd-shim-slight-v0-3-0-v1
-rwxr-xr-x 1 root root    52232184 Jul 23 19:58 /usr/local/bin/containerd-shim-slight-v0-5-1-v1
-rwxr-xr-x 1 root root    53775024 Jul 23 19:58 /usr/local/bin/containerd-shim-spin-v0-8-0-v1
-rwxr-xr-x 1 root root    60175432 Jul 23 19:58 /usr/local/bin/containerd-shim-slight-v0-8-0-v1
```


AlisonB319:
I will add the make generate test files after review so that it is easier to review

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -147,6 +147,31 @@ installCredentalProvider() {
...
+    CPU_ARCH=$(getCPUArch)
+    
+    # Check if oras is installed and if it is the expected version.
+    if command -v oras &> /dev/null; then
+        INSTALLED_ORAS_VERSION=$(oras version 2>&1 | grep -oP '(?<=Version:\s)\d+\.\d+\.\d+')
+        if [[ "$INSTALLED_ORAS_VERSION" == "$ORAS_VERSION" ]]; then
+            echo "Oras version $ORAS_VERSION is already installed."
+            return
+        else
+            echo "Oras is installed but has the wrong version $INSTALLED_ORAS_VERSION != $ORAS_VERSION. Keeping installed version."
```
Devinwong, line 162: Question: do we need to uninstall the old version so that the new version can work properly?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -147,6 +147,31 @@ installCredentalProvider() {
     rm -rf ${CREDENTIAL_PROVIDER_DOWNLOAD_DIR}
 }
 
+# TODO (alburgess) have oras version managed by dependant or Renovate
+installOras() {
```
Devinwong, line 151: Suggested to add unit test with shellspec. Please see the instructions in the readme.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -147,6 +147,31 @@ installCredentalProvider() {
...
+    ORAS_VERSION="1.2.0"
```
Devinwong, line 152: We should move the download URL and version to components.json (as a package) instead of hard coding it here. There are some examples you can refer to. Let me know if you encounter any issue.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -55,6 +55,8 @@ EOF
 fi
 capture_benchmark "purge_and_reinstall_ubuntu"
 
+installOras
```
Devinwong, line 58: It will be great if we can move the logic to the package for loop (around ln186). You can find other packages installed there.

parts/linux/cloud-init/artifacts/cse_install.sh
AlisonB319: I actually don't think we really need this check at all. Going to remove

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -147,6 +147,23 @@ installCredentalProvider() {
...
+# TODO (alburgess) have oras version managed by dependant or Renovate
+installOras() {
+    ORAS_DOWNLOAD_DIR=${1}
+    ORAS_DOWNLOAD_URL=${2}
+    ORAS_VERSION=${3}
+
+    mkdir -p $ORAS_DOWNLOAD_DIR
+
+    echo "Installing Oras version $ORAS_VERSION..."
+    ORAS_TMP=${ORAS_DOWNLOAD_URL##*/} # Use bash builtin ## to remove all chars ("*") up to the final "/"
```
djsly, line 159: is this really needed. ?  I feel this was for the cotnainerImages with the hacked `:*` at the end

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -147,6 +147,23 @@ installCredentalProvider() {
...
+    retrycmd_get_tarball 120 5 "$ORAS_DOWNLOAD_DIR/${ORAS_TMP}" ${ORAS_DOWNLOAD_URL} || exit $ERR_ORAS_DOWNLOAD_TIMEOUT
+    tar -zxf oras_${ORAS_VERSION}_*.tar.gz -C $ORAS_DOWNLOAD_DIR/${ORAS_TMP}
+    sudo mv $ORAS_DOWNLOAD_URL/${ORAS_TMP}/oras /usr/local/bin/
+
```
djsly, line 163: can we delete the `"$ORAS_DOWNLOAD_DIR"` after to keep the VHD free of download artifacts

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -147,6 +147,23 @@ installCredentalProvider() {
...
+    ORAS_DOWNLOAD_DIR=${1}
+    ORAS_DOWNLOAD_URL=${2}
+    ORAS_VERSION=${3}
+
+    mkdir -p $ORAS_DOWNLOAD_DIR
+
+    echo "Installing Oras version $ORAS_VERSION..."
+    ORAS_TMP=${ORAS_DOWNLOAD_URL##*/} # Use bash builtin ## to remove all chars ("*") up to the final "/"
+    retrycmd_get_tarball 120 5 "$ORAS_DOWNLOAD_DIR/${ORAS_TMP}" ${ORAS_DOWNLOAD_URL} || exit $ERR_ORAS_DOWNLOAD_TIMEOUT
+    tar -zxf oras_${ORAS_VERSION}_*.tar.gz -C $ORAS_DOWNLOAD_DIR/${ORAS_TMP}
```
djsly, line 161: why not extracting directly under `/usr/local/bin/`. 

`sudo tar -zxf oras_${ORAS_VERSION}_*.tar.gz -C /usr/local/bin/`

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -262,6 +263,13 @@ for p in ${packages[*]}; do
   capture_benchmark "download_${name}"
 done
 
+downloadContainerdWasmShims
```
djsly, line 266: what is this change ? missing a rebase ?

AlisonB319: No, in the future WasmShims is one of the packages that will be downloaded via oras, so the install needs to happen after oras is already installed. Moved so that I could verify that this change doesn't hurt anything

parts/linux/cloud-init/artifacts/cse_install.sh
AlisonB319: @Devinwong 

wc: /opt/oras/downloads/oras_1.2.0_linux_amd64.tar.gz: No such file or directory testKubeBinaries:Error: File size of /opt/oras/downloads/oras_1.2.0_linux_amd64.tar.gz from https://github.com/oras-project/oras/releases/download/v1.2.0/oras_1.2.0_linux_amd64.tar.gz is invalid. Expected file size: 4052022 - downlaoded file size:
Deleting resource group vhd-test-1722036494-2130

VHD build tests fail when we delete the source code. Is there a reason to check for the .tar.gz over checking if the package is installed via like package --version output? 

https://github.com/Azure/AgentBaker/blob/1973b6d78eb92a121cd995053ee3778088453bc9/vhdbuilder/packer/test/linux-vhd-content-test.sh#L129

Devinwong: @AlisonB319  There is another check a few lines before. which is:
`# if there is a directory with expected name, we assume it's been downloaded and extracted properly`
So if the binaries have been installed to a desired directory, this test should also pass.
If you package can be validated through some command like --version, feel free to add one.
I didn't redesign this part of validation, just using the existing logics to test the packages.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -118,6 +118,22 @@ installCredentalProvider() {
     rm -rf ${CREDENTIAL_PROVIDER_DOWNLOAD_DIR}
 }
 
+# TODO (alburgess) have oras version managed by dependant or Renovate
+installOras() {
```
cameronmeissner, line 122: should we also add call to this somewhere in cse_main.sh to ensure it gets installed at runtime if the underlying VHD doesn't already have it cached?

AlisonB319: The only case where we need oras installed is for a network isolated cluster. So, if it is not cached, it won't be able to be installed on those clusters - since it is needed to install everything else. 

As a general question for my understanding - @Devinwong added the loop inside install_dependencies.sh for installing the packages in components.json. Are any of these packages installed via cse?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -390,20 +390,16 @@
...
-        "v1.28.12",
         "v1.29.4",
         "v1.29.4-1",
         "v1.29.5-1",
         "v1.29.6",
-        "v1.29.7",
         "v1.30.1-1",
-        "v1.30.2-hotfix.20240613",
-        "v1.30.3"
+        "v1.30.2-hotfix.20240613"
```
AlisonB319, line 402: I didn't change these, I think it got a little merked up in a rebase from master. But seems like everything changed is the current state of components.kson

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Some do. For example, in the for-loop for `containerd` case, it calls `installStandaloneContainerd`. And in cse_main.sh around ln124, it also calls `installContainerRuntime` which will eventually calls `installStandaloneContainerd`.
But some don't. Actually recently Paul removed the codes to install cni from cse. So now it will only use what cached in the VHD.

For your case, if oras is not cached as expected, when it comes to network isolated cluster bootstrapping the node, it still won't be able to install oras, right?

AlisonB319: Right I don't think a NIC will have the ability to install oras

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -510,6 +510,20 @@
           }
         }
       }
+    },
+    {
+      "name": "oras",
```
Devinwong, line 515: I remember you said oras is needed for another component to be downloaded? If that is the case, oras needs to be in the first place of this list, until we support a new key "dependencies".

AlisonB319: oh interesting, I thought it needed to be first in the install_dependencies loop. But it should be at the top of components.json as well?

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -121,6 +121,16 @@ testPackagesInstalled() {
         echo $test "[INFO] Directory ${extractedPackageDir} exists"
         continue
       fi
+
+      # if the downloadLocation is /usr/local/bin verify that the package is installed
+      if [ "$downloadLocation" == "/usr/local/bin" ]; then
+          if command -v "$name" >/dev/null 2>&1; then
+              echo "$name is installed."
+              continue
+          else
```
Devinwong, line 130: If the command doesn't exist, is it supposed to continue the next check or to exit with error?
The current code is just logging, right?

AlisonB319: The else calls error so if it doesn't exist, I think it will fail. But atm this only runs if the download location is "/usr/local/bin"

parts/linux/cloud-init/artifacts/components.json
Devinwong: The loop doesn't matter actually. (Did I say we should modify the loop? lol)
The list in components.json controls which should be installed first. Make sense?

vhdbuilder/packer/test/linux-vhd-content-test.sh
Devinwong: the `err` is defined above as 
```err() {
  echo "$1:Error: $2" >>/dev/stderr
}
```
So it doesn't exit but just writing to stderr. If you want to exit, you will need to call exit explicitly, like ln 41.

Devinwong (approved):

vhdbuilder/packer/install-dependencies.sh
AlisonB319: ^ my line of thinking here is wrong. Oras will get installed during build so when it is actually needed during provisioning, it is already installed. So the order does not actually matter - since it will be installed and ready to use during provisioning

PR #4694, opened 2024-07-23
test: reduce variety on node families

r2k1:
`az vm create` uses `--size=Standard_DS1_v2` by default.

We use `Standard_D16ds_v5` in node image build. 

This should reduce required node family variety.  Hopefully reducing likelihood of capacity issues.


timmy-wright (approved):

vhdbuilder/packer/test/run-test.sh
```diff
@@ -84,8 +84,8 @@ else
   TARGET_COMMAND_STRING=""
   if [[ "${ARCHITECTURE,,}" == "arm64" ]]; then
     TARGET_COMMAND_STRING+="--size Standard_D2pds_v5"
-  elif [[ "${FEATURE_FLAGS,,}" == "kata" ]]; then
-    TARGET_COMMAND_STRING="--size Standard_D4ds_v5"
+  else
+    TARGET_COMMAND_STRING="--size Standard_D2ds_v5"
```
Devinwong, line 88: How do you know Standard_D2ds_v5 has more availability?

Devinwong: btw, scanning VHD also fails a lot. Would you like to change it in this PR as well?

r2k1: I don't know it has more availability. But we use this family in our builds. 

Before the change we need Dxds_V5 for one step of the build and DSx_v2 for the other step.
After the change we will only need Dxds_V5.

Devinwong (approved):

PR #4696, opened 2024-07-24
perf: execute test, scan, and cleanup steps in parallel

zachary-bailey:
**What type of PR is this?**

/kind perf

This PR changes the pipeline workflow to execute the test, scan, and cleanup steps in parallel. This is accomplished by making changes to the packer.mk file to enable parallelism.

**Notes:**

The scanning VM and its associated variables were renamed to reflect their association with the scanning process to reduce ambiguity between that VM and the testing VM, as they were both previously named TEST_VM and will now be provisioned at the same time.

This PR bumps the Trivy version from 0.40.0 to 0.53.0 just to stay current. If anyone would like to stay at 0.40.0 please let me know.

Added conditional checks in cleanup.sh to prevent attempted deletion of ARM64 managed images (cant be created) and attempted deletion of image versions that have delete resource locks enabled.

Unsatisfactory VHD build times.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -113,6 +113,11 @@ steps:
   - bash: |
         OS_DISK_URI="$(cat packer-output | grep "OSDiskUri:" | cut -d " " -f 2)" && \
         MANAGED_SIG_ID="$(cat packer-output | grep "ManagedImageSharedImageGalleryId:" | cut -d " " -f 2)" && \
+        PKR_RG_NAME="$(cat packer-output | grep "ResourceGroupName" | cut -d "'" -f 2 | head -1)" && \
+        captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
```
djsly, line 117: why not in CPAS ? like the rest ?

zachary-bailey: @djsly I am not sure what a CPA is. Do you mean in all capital letters? If so, this variable is this way all throughout the codebase, I have wondered that myself.

packer.mk
```diff
@@ -99,6 +99,9 @@ test-building-vhd: az-login
 scanning-vhd: az-login
 	@./vhdbuilder/packer/vhd-scanning.sh
 
+test-scan-and-cleanup:
+	@$(MAKE) -f packer.mk cleanup test-building-vhd scanning-vhd -j3 --output-sync=target
```
djsly, line 103: the problem with this approach is that is the scan fails and we need to rerun we will have to rerun the all of the test. same with the cleanup, we will need to rerun all of the test and scan/puplish to rerun the cleanup part

vhdbuilder/packer/cleanup.sh
```diff
@@ -28,15 +28,18 @@ fi
...
 if [[ -n "$AZURE_RESOURCE_GROUP_NAME" && -n "$IMAGE_NAME" ]]; then
-  if [[ "$MODE" != "default" ]]; then
+  if [[ "$MODE" != "default" && ${ARCHITECTURE,,} != "arm64" ]]; then
     id=$(az image show -n ${IMAGE_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} | jq .id)
     if [ -n "$id" ]; then
       echo "deleting managed image ${IMAGE_NAME} under resource group ${AZURE_RESOURCE_GROUP_NAME}"
       az image delete -n ${IMAGE_NAME} -g ${AZURE_RESOURCE_GROUP_NAME}
     fi
+  else
+    echo "Not attempting managed image deletion due to ARM64 architecture."
```
djsly, line 38: what should we do with he ARM64 images ?  are they staying present ?

.pipelines/templates/.builder-release-template.yaml
djsly: CAPS

vhdbuilder/packer/cleanup.sh
zachary-bailey: No managed images are created for ARM64 SKUs, only image versions. That was leading to the error when we tried to delete them.

.pipelines/templates/.builder-release-template.yaml
zachary-bailey: Yes I am not sure why it is that way, I was just trying to keep it consistent with the rest of the repo.

packer.mk
zachary-bailey: Yes good point. I will look for a way to only re-run failed targets.

vhdbuilder/packer/cleanup.sh
cameronmeissner: are we sure about this? if so good catch

vhdbuilder/packer/cleanup.sh
```diff
@@ -155,7 +158,10 @@ if [[ "${MODE}" == "linuxVhdMode" && -n "${AZURE_RESOURCE_GROUP_NAME}" && "${DRY
   # we limit deletion to 15 SIG image versions per image definition
   for image_definition in $(az sig image-definition list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${SIG_GALLERY_NAME} | jq '.[] | select(.name | test("Ubuntu*|CBLMariner*|V1*|V2*|1804*|2004*|2204*")).name' | tr -d '\"' || ""); do
     for image_version in $(az sig image-version list -g ${AZURE_RESOURCE_GROUP_NAME} -r ${SIG_GALLERY_NAME} -i ${image_definition} | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | head -n 15 | tr -d '\"' || ""); do
-      old_sig_version_ids="${old_sig_version_ids} /subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/${SIG_GALLERY_NAME}/images/${image_definition}/versions/${image_version}"
+      lock_name=$(az lock list --resource /subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${AZURE_RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/${SIG_GALLERY_NAME}/images/${image_definition}/versions/${image_version} --query "[0].name" --output tsv)
```
cameronmeissner, line 161: I'm fine with this, though seems redundant if the lock will already take care of preventing the deletion

vhdbuilder/packer/trivy-scan.sh
```diff
@@ -3,7 +3,7 @@ set -euxo pipefail
 
 TRIVY_REPORT_JSON_PATH=/opt/azure/containers/trivy-report.json
 TRIVY_REPORT_TABLE_PATH=/opt/azure/containers/trivy-images-table.txt
-TRIVY_VERSION="0.40.0"
+TRIVY_VERSION="0.53.0"
```
cameronmeissner, line 6: why the bump here?

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -111,58 +111,114 @@ steps:
     retryCountOnTaskFailure: 3
     displayName: Building VHD
   - bash: |
+        echo MODE=$(MODE) && \
```
cameronmeissner, line 114: this might be something you can look into removing, `MODE` should be redundant at this point since it's only used to differentiate between windows/linux (it previously differentiated between modes which haven't existed for a while...), which we can already do using `OS_TYPE`

vhdbuilder/packer/cleanup.sh
```diff
@@ -28,15 +28,18 @@ fi
 
 #clean up managed image
 if [[ -n "$AZURE_RESOURCE_GROUP_NAME" && -n "$IMAGE_NAME" ]]; then
-  if [[ "$MODE" != "default" ]]; then
+  if [[ "$MODE" != "default" && ${ARCHITECTURE,,} != "arm64" ]]; then
```
cameronmeissner, line 31: `"$MODE" != "default"` should always resolve to true, thus it should be safe to remove entirely

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -111,58 +111,114 @@ steps:
...
+
+        if [ ${gnu_exit_code} -ne 0 ]; then
+          echo "The following targets failed:"
+          awk -F'\t' '$7 != 0 {print $9}' log.txt
+          exit 1
+        else
+          echo "All make targets succeeded"
+        fi
+    condition: always()
+    displayName: Test, Scan, and Cleanup
```
cameronmeissner, line 221: maybe something else you can think about, and maybe we already decided to head in this direction but I'm not sure off the top of my head - could we combine the testing/scanning steps such that they use the same underlying VM?

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -111,58 +111,114 @@ steps:
...
-        -e ENABLE_CGROUPV2=${ENABLE_CGROUPV2} \
-        -e GIT_BRANCH=$(Build.SourceBranch) \
-        ${CONTAINER_IMAGE} make -f packer.mk test-building-vhd
-    displayName: Run VHD Tests
+        PKR_RG_NAME="$(cat packer-output | grep "ResourceGroupName" | cut -d "'" -f 2 | head -1)" && \
+        captured_sig_version="$(cat vhdbuilder/packer/settings.json | grep "captured_sig_version" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        IMPORTED_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "imported_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        SIG_IMAGE_NAME="$(cat vhdbuilder/packer/settings.json | grep "sig_image_name" | awk -F':' '{print $2}' | awk -F'"' '{print $2}')" && \
+        IMAGE_NAME="${SIG_IMAGE_NAME}-${captured_sig_version}" && \
+        parallel --joblog log.txt --results results/ --retries 1 ::: \
```
cameronmeissner, line 122: this seems hacky... though I guess from reading around a little bit there isn't a clean way to have certain steps within a single job execute in parallel at the ADO level

cameronmeissner: if we decide this is the best way to accomplish this, I'd at least recommend putting all this into its own `.sh` file probably somewhere under `vhdbuilder` and simply reference that within this yaml template

.pipelines/templates/.builder-release-template.yaml
cameronmeissner: we can align towards CAPS here

.pipelines/templates/.builder-release-template.yaml
zachary-bailey: Yeah agreed with putting it within a script. Actively developing this right now, testing the feasibility of gnu parallel for this. 

I was originally using make's implementation for parallelism which was pretty simple and clean, but @djsly suggested we retry only failed targets instead of all targets if one fails which I think makes sense, so I'm looking for ways to accomplish that.

vhdbuilder/packer/cleanup.sh
zachary-bailey: I'm trying to both parallelize the three steps, but also decrease the amount of time each step takes. I'd hoped this would prevent time wasted on failed deletions, but it did not seem to help very much.

vhdbuilder/packer/cleanup.sh
zachary-bailey: yep!

cameronmeissner (commented):
test

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -1,14 +1,29 @@
...
+
+if [[ -z "$SIG_GALLERY_NAME" ]]; then
+    SIG_GALLERY_NAME="PackerSigGalleryEastUS"
+fi
+# Check to ensure the build step succeeded
+SIG_VERSION=$(az sig image-version show -e ${CAPTURED_SIG_VERSION} -i ${SIG_IMAGE_NAME} -r ${SIG_GALLERY_NAME} -g ${AZURE_RESOURCE_GROUP_NAME} --query id --output tsv || true)
+if [ -z "${SIG_VERSION}" ]; then
+    echo -e "Build step did not produce an image version. Exiting $(basename $0) with exit code 0...\n\n\n"
+    exit 0
+elif [ "$OS_VERSION" == "18.04" ]; then
```
cameronmeissner, line 12: nit: redundant elif

cameronmeissner:
if you don't mind this has inspired me to try an alternative approach I'd like to whip up and get your opinion on if you don't mind holding off on merging

vhdbuilder/packer/trivy-scan.sh
zachary-bailey: Just looking to keep current, I can bring it back down if necessary. I originally bumped in order to make use of parallel scanning but that was a dead end.

I reached out to Jason because I know he is working with the trivy scanning a bit and he said he didn't have a preference.

vhdbuilder/packer/vhd-scanning.sh
zachary-bailey: should they be written as separate if statements?

.pipelines/templates/.builder-release-template.yaml
zachary-bailey: I considered that but was concerned about the resource usage on the vm between several az run commands, the scanning executable, test script, and the cleanup. 

Though I suppose we could run cleanup on the agent, and test and scan on the same vm. Would be willing to try but not sure if thats what you are doing now.

cameronmeissner: `Though I suppose we could run cleanup on the agent, and test and scan on the same vm. Would be willing to try but not sure if thats what you are doing now.`

I meant this

basically combine the test/scan steps into a single step, executed on the same underlying VM in parallel (not agent VM)

then run that ^ step in parallel with cleanup

don't worry about it though, that's probably something we can follow-up with in another PR

retry-function.mk
```diff
@@ -0,0 +1,17 @@
+define retrycmd_if_failure
```
cameronmeissner, line 1: yeah, I'd probably prefer having a new bash script parameterized with envrionment variables we call directly from the make file which can run cleanup/scan/test, which would include the retry logic in-line

vhdbuilder/packer/trivy-scan.sh
cameronmeissner: as long as things still work I'm fine with it, just wanted to check

vhdbuilder/packer/vhd-scanning.sh
cameronmeissner: something like this:

```suggestion
if [  -z "${SIG_VERSION}" ]; then
    echo -e "Build step did not produce an image version. Exiting $(basename $0) with exit code 0...\n\n\n"
    exit 0
fi

if [ "$OS_VERSION" == "18.04" ]; then
    # 18.04 VMs don't have access to new enough 'az' versions to be able to run the az commands in vhd-scanning-vm-exe.sh
    echo -e "Skipping scanning for 18.04\n\n\n"
    exit 0
fi
```

imo that's a bit cleaner

zachary-bailey:
> tried doing this: https://github.com/Azure/AgentBaker/pull/4821/files, which unfortunately doesn't quite work as expected

We can do that! I've done similar before, but by the time I found out how to persist the packer variables, we wouldn't have enough agent VMs to support 3 concurrent jobs x number of skus built. That's why I went with the parallelism in make instead :/. Its a bummer agents cant run steps in parallel.

I'll look into putting the test/scan/parallelism into a script though as you mentioned above

vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -0,0 +1,83 @@
...
+
+SCRIPT_ARRAY+=("./vhdbuilder/packer/cleanup.sh") # Always run cleanup
+
+# Check to ensure the build step succeeded
+SIG_VERSION=$(az sig image-version show \
+-e ${CAPTURED_SIG_VERSION} \
+-i ${SIG_IMAGE_NAME} \
+-r ${SIG_GALLERY_NAME} \
+-g ${AZURE_RESOURCE_GROUP_NAME} \
+--query id --output tsv || true)
```
cameronmeissner, line 35: we should generally discourage `|| true` imo, would prefer to see failures as they occur rather than silencing them and relying on fall back logic

vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -0,0 +1,83 @@
...
+  PID=$!
+  SCRIPT_PIDS[$SCRIPT]=${PID}
+done
+wait ${SCRIPT_PIDS[@]}
+
+echo -e "\nChecking exit codes for each script..."
+STEP_FAILED=false
+for SCRIPT in "${!SCRIPT_PIDS[@]}"; do
+  PID=${SCRIPT_PIDS[$SCRIPT]}
+  wait $PID
```
cameronmeissner, line 70: do we need this `wait` considering we wait for all running PIDs to finish on line 64?

vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -0,0 +1,83 @@
...
+  if [ ${EXIT_CODE} -ne 0 ]; then
+    STEP_FAILED=true
+  fi
+  echo -e "${SCRIPT} exited with code ${EXIT_CODE}"
+done
+
+if [[ "${STEP_FAILED}" == true ]]; then
+  echo -e "\nOne or more scripts failed. Exiting with exit code 1.\n"
+  exit 1
+else
```
cameronmeissner, line 81: nit: redundant else

vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -0,0 +1,83 @@
...
+SIG_VERSION=$(az sig image-version show \
+-e ${CAPTURED_SIG_VERSION} \
+-i ${SIG_IMAGE_NAME} \
+-r ${SIG_GALLERY_NAME} \
+-g ${AZURE_RESOURCE_GROUP_NAME} \
+--query id --output tsv || true)
+
+if [ -z "${SIG_VERSION}" ]; then
+  echo -e "\nBuild step did not produce an image version. Running cleanup and then exiting."
+  retrycmd_if_failure 2 3 "${SCRIPT_ARRAY[@]}"
```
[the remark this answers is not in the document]
zachary-bailey: good call will make the change

vhdbuilder/packer/test-scan-and-cleanup.sh
zachary-bailey: Thanks. Artifact from test/scan scripts that have set -e declared. If the build failed, this command caused the test/scan/cleanup to fail if I didn't || true. Will change because should be fine without set -e.

vhdbuilder/packer/test-scan-and-cleanup.sh
zachary-bailey: I am using wait here to get the exit code for each script/retrycmd. I don't have set -e so that they can fail without failing the rest of the scripts, cleanup in particular. The initial wait is just to make sure that all the retrycmds have finished.

zachary-bailey: Looking at it I can probably just take out the first wait. Will test now.

cameronmeissner (approved):

PR #4707, opened 2024-07-25
test: combine ADO pipelines

r2k1:
Combine 3 pipelines into one.

Make it simple, first new images then run e2e tests on new images.
It should reduce duplication and eliminate disconnect e2e tests and images.


.pipelines/e2e.yaml
cameronmeissner, line 1: I'd probably opt for keeping this pipeline around in case we ever need to test/debug something without the need to build new VHDs

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -13,75 +41,75 @@ parameters:
   - name: build1804containerd
     displayName: Build 1804 containerd
     type: boolean
-    default: true
+    default: false
```
cameronmeissner, line 44: we rely on default values here to know which SKUs to build during official releases

r2k1: why is there SKUs we don't release?

.pipelines/e2e.yaml
r2k1: Can you imagine a use-case where this would be useful?

I would run something locally if I need to debug or change something inside tests. (Or just build a new image once a year)

PR #4711, opened 2024-07-25
feat: Update VHD scanning so it doesn't require RBAC administrator role

babbageclunk:
**What type of PR is this?**

It tweaks the VHD scanning build task so that instead of assigning the Storage Blob Data Contributor role to the system-assigned identity on the VM (which requires that the identity running the build has the Role Based Access Control Administrator role), it assigns a pre-existing scanning identity (with the storage contributor role) to the VM. The identity resource ID is passed in as the pipeline variable SCANNING_MSI_RESOURCE_ID.

This means that we can remove the RBAC admin role from the identity running the build.

It also suppresses the extremely noisy output from `ctr image pull` in the build stage. This accounts for 90% of the output from the step - it takes the logs for that stage from ~250k to ~20k. The volume of image pull logs made it very difficult to read the output from the build step (to the point of the tab crashing). Any errors still get reported on stderr. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

Let me know if I should pull the ctr image pull output change out as a separate PR - I left it in here to trim down these runs while testing the branch.

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -640,7 +640,7 @@ configGPUDrivers() {
     if [[ $OS == $UBUNTU_OS_NAME ]]; then
         mkdir -p /opt/{actions,gpu}
         if [[ "${CONTAINER_RUNTIME}" == "containerd" ]]; then
-            ctr image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG
+            ctr image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG > /dev/null
```
cameronmeissner, line 643: are we sure we want to silence stdout here?

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -71,6 +73,8 @@ if [[ "${OS_TYPE}" == "Linux" && "${ENABLE_TRUSTED_LAUNCH}" == "True" ]]; then
     VM_OPTIONS+=" --security-type TrustedLaunch --enable-secure-boot true --enable-vtpm true"
 fi
 
+SCANNING_ID=$(az identity show --ids "${SCANNING_MSI_RESOURCE_ID}" --query id --output tsv)
```
cameronmeissner, line 76: why are we doing this? SCANNING_ID will have the same value as SCANNING_MSI_RESOURCE_ID after this call

further, the call to create the VM should fail with an error saying the identity doesn't exist, if your aim here was to check for that

vhdbuilder/scripts/automate_helpers.sh
```diff
@@ -16,6 +16,19 @@ retrycmd_if_failure() {
     done
     echo Executed \"$@\" $i times;
 }
+retrycmd_if_failure_nostdout() {
```
cameronmeissner, line 19: probably don't need this here if we're not using it in the automation scripts

vhdbuilder/packer/vhd-scanning.sh
babbageclunk: Oh you're right - I was initially looking up the identity by RG and name and getting the principal ID, to assign that. But I I'm just getting the ID directly passed in, so I don't need to do any lookup. I'll update it.

parts/linux/cloud-init/artifacts/cse_config.sh
babbageclunk: The output from ctr image pull is really noisy - without these changes the output of the build step is 250k lines, with them it's only 20k. The volume was making it really hard to read the logs. And they're very much meant for interactive usage - showing each layer's download progress, and I think it's using ^r characters to update them (which doesn't work in this context). Any problems still get reported on stderr.

vhdbuilder/scripts/automate_helpers.sh
babbageclunk: Good point - I'll pull it out from this one.

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: fair enough, and I don't think we ever really did much debugging using the output from these commands anyways

cameronmeissner (approved):

PR #4712, opened 2024-07-25
perf: use premium SSDs

zachary-bailey:
**What type of PR is this?**

/kind perf

This PR instructs packer to use premium SSDs when building our custom images. This greatly reduces VHD build time given the faster IOPS/Mbps of the Premium SSDs over Standard HDD. The average build time reduction across all SKUs is 00:07:10.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


jason1028kr (approved):

jason1028kr:
assuming we can specify the type on our end, without some approval on quotas/costs

zachary-bailey:
> assuming we can specify the type on our end, without some approval on quotas/costs

@jason1028kr Yes we can! We actually had this set before, but it was inadvertently removed in a different PR. Thanks!

PR #4719, opened 2024-07-26
fix: add log rotation for CIS log file

phealy:
**What type of PR is this?**
/kind bug

/etc/rsyslog.d/60-CIS.conf adds logging for local messages to /var/log/localmessages, but nothing rotates this log file so it can fill the disk on long-lived nodes. This PR adds log rotation for this file.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
/var/log/localmessages will now be rotated with the same configuration as /var/log/messages.
```


parts/linux/cloud-init/artifacts/logrotate-d-rsyslog-CIS.conf
```diff
@@ -0,0 +1,16 @@
...
+  rotate 5
+  daily
+  maxsize 300M
+  missingok
+  notifempty
+  compress
+  delaycompress
+  sharedscripts
+  postrotate
+      systemctl kill -s HUP rsyslog.service
```
djsly, line 14: what is this for ?

djsly (commented):
I think you need to run `make generate` ?

djsly (approved):

phealy:
> I think you need to run `make generate` ?

No changes to CSE or cloud-init, so nothing for generate to do (I did run make generate before committing but there were no changes).

UtheMan:
Question - what sort of things are we logging there? Is 5 days of log history enough in case of issues?

UtheMan (approved):

phealy:
Tested both ubuntu and mariner VHDs to confirm functionality.

PR #4725, opened 2024-07-29
test: add containerd and runc version tests

timmy-wright:
**What type of PR is this?**

/kind test

Add tests for containerd versions so we know they don't regress when I write the override code.

**Which issus
  
- [x ] includes documentation
- [x ] adds unit tests
- [ x] tested upgrade from previous version

**Release note**:

```
none
```


e2e/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbakere2e
 
-go 1.22
```
timmy-wright, line 3: Adding the patch version meant I could run go again.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker
 
-go 1.22
```
timmy-wright, line 3: Adding the patch version meant I could run go again.

e2e/validators.go
```diff
@@ -149,7 +149,7 @@ func makeExecutableCommand(steps []string) string {
 	return command
 }
 
-func serviceCanRestartValidator(serviceName string, restartTimeoutInSeconds int) *LiveVMValidator {
```
timmy-wright, line 152: Changed capitalization to match other validators in the file

e2e/e2e-local.sh
```diff
@@ -6,4 +6,4 @@ set -euxo pipefail
 : "${PARALLEL:=100}"
 
 go version
-go test -parallel $PARALLEL -timeout $TIMEOUT ./...
+go test -parallel $PARALLEL -timeout $TIMEOUT -v ./...
```
djsly, line 9: this isn't too verbose preventing us from properly seeing the test results ?

e2e/scenario_test.go
```diff
@@ -24,6 +24,8 @@ func TestAll(t *testing.T) {
 	t.Run("azurelinuxv2-arm64", Scenario_azurelinuxv2ARM64)
 	t.Run("azurelinuxv2-arm64-airgap", Scenario_azurelinuxv2ARM64AirGap)
 	t.Run("azurelinuxv2-azurecni", Scenario_azurelinuxv2_azurecni)
+	t.Run("azurelinuxv2-containerc-version", Scenario_azurelinuxv2HasRightContainerdVersion)
+	t.Run("azurelinuxv2-rund-version", Scenario_azurelinuxv2HasRightRuncVersion)
```
djsly, line 28: runc ?

e2e/scenario_test.go
```diff
@@ -24,6 +24,8 @@ func TestAll(t *testing.T) {
 	t.Run("azurelinuxv2-arm64", Scenario_azurelinuxv2ARM64)
 	t.Run("azurelinuxv2-arm64-airgap", Scenario_azurelinuxv2ARM64AirGap)
 	t.Run("azurelinuxv2-azurecni", Scenario_azurelinuxv2_azurecni)
+	t.Run("azurelinuxv2-containerc-version", Scenario_azurelinuxv2HasRightContainerdVersion)
```
djsly, line 27: container**d**

e2e/scenario_test.go
```diff
@@ -42,15 +46,19 @@ func TestAll(t *testing.T) {
...
 	t.Run("ubuntu2204-airgap", Scenario_ubuntu2204AirGap)
 	t.Run("ubuntu2204-arm64", Scenario_ubuntu2204ARM64)
 	t.Run("ubuntu2204-artifact-streaming", Scenario_ubuntu2204ArtifactStreaming)
 	t.Run("ubuntu2204-chrony-restarts", Scenario_ubuntu2204ChronyRestarts)
-	t.Run("ubuntu2204-containerd-url", Scenario_ubuntu2204ContainerdURL)
-	t.Run("ubuntu2204-containerd-version", Scenario_ubuntu2204ContainerdVersion)
+	t.Run("ubuntu2204-containerd-version", Scenario_ubuntu2204HasRightContainerdVersion)
+	t.Run("ubuntu2204-runc-version", Scenario_ubuntu2204HasRightRuncVersion)
+	t.Run("ubuntu2204-containerd-override", Scenario_ubuntu2204ContainerdURL)
+	t.Run("ubuntu2204-containerd-cur-ver", Scenario_ubuntu2204ContainerDHasCurrentVersion)
```
djsly, line 61: dome test name are suing `D` some `d` for containerd.

e2e/scenario_test.go
```diff
@@ -919,3 +927,83 @@ func Scenario_ubuntu2204Wasm(t *testing.T) {
 		},
 	})
 }
+
+/*
+We could figure out the expected versions by parsing the components.json and manifest.json files, but then we'd need logic
```
djsly, line 932: manifest.json doesn't exists anymore

e2e/scenario_test.go
```diff
@@ -919,3 +927,83 @@ func Scenario_ubuntu2204Wasm(t *testing.T) {
...
+to resolve the right version based on the distro. It seems easier to have the versions of containerd and runc hard coded here because:
+
+* Having additional code for parsing the files and resolving the version introduces risk of bugs in the test code.
+* Re-using the parser from prod introduces risk that there's a bug in the parser that we don't pick up as it impacts both test and prod code.
+* It's not much effort for a dev to change the hard coded values here and in components.json
+*/
+func Scenario_marinerv2HasRightContainerdVersion(t *testing.T) {
+	// This is the version that's in the manifest file - seems to be being ignored right now.
+	//Scenario_genericHasRightContainerdVersion(t, "marinerv2", "aks-cblmariner-v2-gen2", "1.6.26-5.cm2")
+	Scenario_genericHasRightContainerdVersion(t, "ubuntu2204", "aks-ubuntu-containerd-22.04-gen2", "1.7.20-1")
```
djsly, line 942: I though we don't control the versions of containers for AzLinux, we simply installs the latest available in their package repo ?

go.mod
djsly: I think we are using 1.22.2 with aks-rp

djsly (commented):
@Devinwong  didn't we merge the new logic for containerd and runc ?  I'm not sure why some of the test cases are still talking about the version in manifest.json

e2e/scenario_test.go
```diff
@@ -869,7 +877,7 @@ func Scenario_ubuntu2204gpuNoDriver(t *testing.T) {
 
 func Scenario_ubuntu2204ContainerdURL(t *testing.T) {
 	RunScenario(t, &Scenario{
-		Description: "tests that a node using the Ubuntu 2204 VHD with the ContainerdPackageURL override bootstraps with the provided URL and not the maifest contianerd version",
+		Description: "tests that a node using the Ubuntu 2204 VHD with the ContainerdPackageURL override bootstraps with the provided URL and not the manifest containerd version",
```
UtheMan, line 880: i believe at this point we no longer use manifest file and only utilize components.json, right @Devinwong ?

e2e/scenario_test.go
```diff
@@ -919,3 +927,83 @@ func Scenario_ubuntu2204Wasm(t *testing.T) {
 		},
 	})
 }
+
+/*
+We could figure out the expected versions by parsing the components.json and manifest.json files, but then we'd need logic
+to resolve the right version based on the distro. It seems easier to have the versions of containerd and runc hard coded here because:
```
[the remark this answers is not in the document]
UtheMan: Since components.json is supposed to be the "source of truth" i would prefer if we found to avoid hardcoding expected versions in multiple places

e2e/e2e-local.sh
timmy-wright: I spoke with Artur who said he removed it accidently. Will revert - was trying to get "in progress" messages rather than the scripts just block for a couple of mins as the VMs start up.

e2e/scenario_test.go
timmy-wright: Also containerc! Will fix both.

e2e/scenario_test.go
timmy-wright: Will fix.

e2e/scenario_test.go
timmy-wright: I thought so to, but I'm not sure that's happenning at the moment.

e2e/scenario_test.go
timmy-wright: I looked at the code that installs containerc and the place I found it (which might be the wrong place) it uses manifest.json not components.json

e2e/scenario_test.go
timmy-wright: It sure does exist - and the code I found that installs containerd uses it. But there might be multiple places, so I'll have another hunt.

e2e/scenario_test.go
timmy-wright: Aaah, I was looking at the self-contained install scripts not the current scripts.

e2e/scenario_test.go
timmy-wright: Turns out I was looking at the self-contained VHD install code. That references manifest.json. However, the main script uses components.json. Manifest.json and manifest.cue still exist in the repository tho - they haven't been deleted.

e2e/scenario_test.go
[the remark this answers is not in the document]
timmy-wright: I've looked at that code and think I can reuse it for my tests.

However, I think hard coding the expected value in the test is a good thing in this case. That way when we want to upgrade the version, we have to manually change the value in the test - which also ensures that we've made the right change to components.json to do the upgrade. For something as important as the containerd or runc version, having this double check can be really useful.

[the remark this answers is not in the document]
djsly: I bet 100$ that the next person will run their pipeline and hit a test failure because they will have forgotten to bump the version in the test :)

djsly (approved):

PR #4748, opened 2024-07-31
test: an attempt to fix flaky WASM e2e tests

r2k1:
Currently WASM tests are flaky.

I think it takes a significant amount of time before pod is accessible by IP from other nodes.
Additionally, I found current curl config doesn't retry on connection errors.
I spent hours playing and adjusting configuration, but it takes so much time I couldn't come up with something reliable.

We don't perform same check for non-wasm pods.

In the end I just replaced the curl from debug pod with a pod readiness check.

Additionally, I reduced number of nodes we keep on standby by 50%. And reduced polling intervals to speed up tests.


e2e/validation.go
```diff
@@ -39,35 +36,12 @@ func validateWasm(ctx context.Context, t *testing.T, kube *Kubeclient, nodeName,
 
 	execResult, err := pollExecOnPod(ctx, t, kube, defaultNamespace, debugPodName, getWasmCurlCommand(fmt.Sprintf("http://%s/hello", spinPodIP)))
```
r2k1, line 37: I'm actually not sure what we want to test here.

Wouldn't waiting for pod containers to be in ready state be enough like we do for nginx?

e2e/template.go
```diff
@@ -451,12 +451,16 @@ func getNginxPodTemplate(nodeName string) string {
 kind: Pod
 metadata:
   name: %[1]s-nginx
-  namespace: default
 spec:
   containers:
   - name: nginx
     image: mcr.microsoft.com/oss/nginx/nginx:1.21.6
     imagePullPolicy: IfNotPresent
+    readinessProbe:
```
cameronmeissner, line 459: nice

e2e/cluster.go
```diff
@@ -21,7 +21,7 @@ import (
 // this will avoid potential conflicts with tests running on other branches
 // there is no strict rules or a hidden meaning for the version
 // testClusterNamePrefix is also used for versioning cluster configurations
-const testClusterNamePrefix = "abe2e-v20240725-"
+const testClusterNamePrefix = "abe2e-v20240801-"
```
cameronmeissner, line 24: when would you say one should update this? generally when making large behavioral changes?

r2k1: When you make any change to the cluster config.

I reduced system nodepool size. I think it's not allowed by create or update AKS cluster operation.

e2e/pod.go
```diff
@@ -26,38 +28,32 @@ func getDebugPodName(ctx context.Context, kube *Kubeclient) (string, error) {
...
 	nginxPodManifest := getNginxPodTemplate(nodeName)
-	if err := ensurePod(ctx, kube, nginxPodName, nginxPodManifest); err != nil {
+	if err := ensurePod(ctx, t, namespace, kube, nginxPodName, nginxPodManifest); err != nil {
 		return "", fmt.Errorf("failed to ensure test nginx pod %q: %w", nginxPodName, err)
 	}
 	return nginxPodName, nil
 }
 
-func ensureWasmPods(ctx context.Context, kube *Kubeclient, nodeName string) (string, error) {
+func ensureWasmPod(ctx context.Context, t *testing.T, namespace string, kube *Kubeclient, nodeName string) (string, error) {
```
cameronmeissner, line 40: do we really need this `ensureWasmPod` function anymore? I'd probably opt for inlining this stuff directly in the validator implementation

cameronmeissner (approved):

e2e/cluster.go
cameronmeissner: maybe worth adding a comment to clarify?

e2e/pod.go
r2k1: inlined

e2e/cluster.go
r2k1: I tried. Do you have suggestions to improve it?

e2e/pollers.go
```diff
@@ -161,20 +137,27 @@ func waitUntilNodeReady(ctx context.Context, t *testing.T, kube *Kubeclient, vms
...
 	return wait.PollUntilContextCancel(ctx, waitUntilPodRunningPollInterval, true, func(ctx context.Context) (bool, error) {
 		pod, err := kube.Typed.CoreV1().Pods(defaultNamespace).Get(ctx, podName, metav1.GetOptions{})
 		if err != nil {
 			return false, err
 		}
 
-		return pod.Status.Phase == corev1.PodPhase("Running"), nil
-	})
-}
+		if pod.Status.Phase == "Pending" {
```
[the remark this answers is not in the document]
djsly: I read the wait.PollUntilContextCancel documentation, this is exactly what is happening here :).  return `false` without an error is considered not `cancel` !

e2e/cluster.go
r2k1: all clusters should be automatically deleted after 2-4 days

e2e/pollers.go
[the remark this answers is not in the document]
r2k1: I don't really like interface of `PollUntilContextCancel` I need to dig into description every time I use it.

I don't think we need to use exponential backoffs anywhere in the tests. They increase execution time, and we don't worry about any load on the other end.

PR #4760, opened 2024-08-02
test: improve e2e

r2k1:
- Update e2e readme
- Add .env file support for easier local configuration/debugging
- Automatically version clusters based on configuration hash. Don't need to manually upgrade version anymore.
- Optimize containerd/runc version testing. Reduce total number of tests (and required VMs for testing)
- Improve test graceful shutdowns. Previously tests missed shutdown signals if they started after the signal is received.
- Allow filtering test by its name
- Use flat structure for tests. Better IDE integration. Easier to run a single test.


e2e/config/config.go
```diff
@@ -1,50 +1,40 @@
...
-	ResourceGroupName             = "abe2e-" + Location
-	SIGVersionTagName             = lookupEnvWithDefaultString("SIG_VERSION_TAG_NAME", "branch") // ADO tags every SIG image version with `branch` tag. By specifying `branch=refs/heads/master` we load latest image version from the master branch.
-	SIGVersionTagValue            = lookupEnvWithDefaultString("SIG_VERSION_TAG_VALUE", "refs/heads/master")
-	SkipTestsWithSKUCapacityIssue = lookupEnvWithDefaultBool("SKIP_TESTS_WITH_SKU_CAPACITY_ISSUE", false)
-	SubscriptionID                = lookupEnvWithDefaultString("SUBSCRIPTION_ID", "8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8")
-	TagsToRun                     = os.Getenv("TAGS_TO_RUN")
-	TagsToSkip                    = os.Getenv("TAGS_TO_SKIP")
-	TestTimeout                   = lookupEnvWithDefaultDuration("TEST_TIMEOUT", 18*time.Minute)
-	E2ELoggingDir                 = lookupEnvWithDefaultString("LOGGING_DIR", "scenario-logs")
+	Cfg               = mustLoadConfig()
```
cameronmeissner, line 11: nit: naming seems redundant, considering renaming the package to `suite` so we can do `suite.Config` instead of `config.Cfg`?

can also rename the struct to `Configuration` to accomplish that

e2e/scenario_helpers_test.go
```diff
@@ -15,16 +17,52 @@ import (
...
+	ch := make(chan os.Signal, 2)
+	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		// block until signal is received
+		<-ch
+		fmt.Println("Received cancellation signal, gracefully shutting down the test suite. Cancel again to force exit.")
+		cancel()
+
+		// block until second signal is received
```
cameronmeissner, line 36: I'm probably missing something, though wouldn't we leak this goroutine in cases where the user does a graceful shutdown?

e2e/README.md
```diff
@@ -24,78 +21,98 @@ existsnce, sysctl settings, etc.) is as expected.
...
-`KEEP_VMSS` can also be optionally specified to have the test suite retain the bootstrapped VM(s) for further debugging.
-When this option is specified, the private SSH key used to connect to each VM will be included within each scenario's
-log bundle respectively.
 
-**Note that when using `e2e-local.sh`, a timeout value of 90 minutes is applied to the `go test` command.**
+### Debugging
 
-You may also run the test command with custom arguments yourself (assuming you've properly setup the required
-environment variables) from within the `e2e/` directory like so:
+Set `KEEP_VMSS` to retain bootstrapped VMs for debugging. This will include the private SSH key in each scenario's log bundle. When using the flag, please ensure to run only test you need to debug, as the VMs will not be
```
cameronmeissner, line 52:
```suggestion
Set `KEEP_VMSS=true` to retain bootstrapped VMs for debugging. Setting this will also have the VM's private SSH key included in each scenario's log bundle. When using this flag, please ensure to run only test you need to debug, as the VMs will not be
```

e2e/README.md
```diff
@@ -24,78 +21,98 @@ existsnce, sysctl settings, etc.) is as expected.
...
+```json
+{
+  "go.testFlags": ["-parallel=100"],
+  "go.testTimeout": "30m"
+}
+```
+
+## Package Structure
+
+The top-level package of the Golang E2E implementation is named `e2e_test` and is entirely separate from all AgentBaker packages.
```
cameronmeissner, line 96:
```suggestion
The top-level package of the Golang E2E implementation is named `e2e` and is entirely separate from all AgentBaker packages.
```

e2e/README.md
```diff
@@ -24,78 +21,98 @@ existsnce, sysctl settings, etc.) is as expected.
...
+
+The primary testing function is located in [suite_test.go](suite_test.go), which is run by `go test`.
```
cameronmeissner, line 98:
```suggestion
The definitions and entry points for each test scenario, ran by `go test`, are located in [scenario_test.go](scenario_test.go).
```

e2e/README.md
```diff
@@ -24,78 +21,98 @@ existsnce, sysctl settings, etc.) is as expected.
...
 SIG_VERSION_TAG_NAME=buildId SIG_VERSION_TAG_VALUE=123456789 TAGS_TO_RUN="os=ubuntu2204" ./e2e-local.sh
 ```
 
-### Registering New VHD SKUs for E2E Testing
+### Registering New VHD SKUs
 
-When adding a new scenario which uses a VHD that doesn't currently have an associated entry in the base catalog, please
-Build and add delete-lock to the underlying image version. If it's not done, the image version will be deleted after a few days
-by the garbage collector.
+When adding tests for a new VHD image, ensure to add a delete-lock to prevent the garbage collector from deleting the image version.
```
cameronmeissner, line 115:
```suggestion
```

e2e/README.md
```diff
@@ -113,20 +130,6 @@ assertions on the contents of both the stdout and stderr streams that result fro
 validators can be used to assert on numerous types of properties of the live VM, such as the live file system and kernel
 state.
 
-You can find all implemented scenarios in the [scenario](scenario/) pacakge within files prefixed with `scenario_`.
-The `Scenario` struct definition can be found in [scenario/types.go](scenario/types.go).
-
-### Implementation
```
cameronmeissner, line 119: I'd opt for keeping this section around and updating it with correct instructions

e2e/cluster.go
```diff
@@ -133,7 +131,20 @@ func createCluster(ctx context.Context, t *testing.T, cluster *armcontainerservi
...
+	if err != nil {
+		panic(err)
+	}
+	hasher := sha256.New()
+	_, err = hasher.Write(jsonData)
+	if err != nil {
+		panic(err)
+	}
+	hexHash := hex.EncodeToString(hasher.Sum(nil))
+	return hexHash[:5]
```
cameronmeissner, line 147: I'm assuming we're only taking 0-4 here to conform with name length constraints?

e2e/config/config.go
r2k1: I couldn't figure out a better name. But I feel like `config` is a better package name than `suite`.

e2e/scenario_helpers_test.go
r2k1: This function is called only once on a startup.
And after every test shutdown the test application is closed.

e2e/README.md
r2k1: Isn't it's true? I think we still need to have at least a single locked version for each image

e2e/README.md
r2k1: I thought about it. And my current suggestion would be "copy any test and make some tweaks".

Which is probably isn't really helpful and can be wrong at some stage in the future.

e2e/cluster.go
r2k1: I think we can use up to 64 characters for a cluster name. Potentially can throw another 20 or so.
The main goal is to have enough uniqueness. Something similar to pod names.

cameronmeissner (approved):

PR #4764, opened 2024-08-02
[cleanup] drop calico from caching

paulgmiller:
**What type of PR is this?**
/kind cleanup
We're going to have multiple versions of calico and jsut don't have space to cache them all and we will not be defaulting customer to it in forseeable future so removing. (its already behind latest calico verison so this shouldn't make things worse) 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


wedaly (approved):

cameronmeissner (approved):
bye bye calico

PR #4765, opened 2024-08-02
chore: update npm linux to v1.5.34

huntergregory:
**What type of PR is this?**

NPM version update.

We are rolling out a fix for a race condition in NPM (see release notes).

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

https://github.com/Azure/azure-container-networking/releases/tag/v1.5.34


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -138,7 +138,8 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-npm:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.5.29"
+        "v1.5.29",
```
cameronmeissner, line 141: when can we remove 1.5.29? we're actually running into VHD space issues at the moment (e.g. caching too many container images), thus I want to double-check that we need to move to a state where we're caching both these versions

I'm assuming we will need to do it this way since the RP change needs to rollout first before removing 1.5.29

huntergregory: Is it ok to cache both for now, and then remove the old version in two weeks after a successful rollout?

cameronmeissner: yes that's fine, just wanted confirm that was necessary

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/components.json
AlisonB319: @cameronmeissner fyi I'm seeing this PR is failing the 1804 PR pipeline due to lack of space.

huntergregory: d4daa4d removed .29 due to lack of space. It's ok to remove the previous image while rolling out an NPM image release since NPM does not run in too many cores.

PR #4768, opened 2024-08-02
test: add option to run validation checks from pods running on non-host network, add a check testing if wireserver connection is properly blocked

UtheMan:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/scenario_test.go
```diff
@@ -632,6 +633,23 @@ func Scenario_ubuntu2204(t *testing.T) {
...
+		Description: "Tests that a node using the Ubuntu 2204 VHD can be properly bootstrapped",
+		Config: Config{
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDUbuntu2204Gen2Containerd,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-ubuntu-containerd-22.04-gen2"
+				nbc.AgentPoolProfile.Distro = "aks-ubuntu-containerd-22.04-gen2"
+			},
+			LiveVMValidators: []*LiveVMValidator{
+				FileHasContentsValidator("/opt/azure/containers/kubelet.sh", "iptables -I FORWARD -d 168.63.129.16 -p tcp -m multiport --dports 80,32526 -j DROP"),
```
cameronmeissner, line 647: should we just add this as a common validator instead of having an extra scenario for it?

UtheMan: We could and we probably should, but I think if we wanted to do this it would make more sense to figure out "must have" things like this one and group them into a common validator that runs on all tests

UtheMan: Happy to work on it later on too.

cameronmeissner: spoke offline, sounds like we can just add this to list of the common validators and include a real network check from inside the pod network to verify the ports are blocked

e2e/kube.go
```diff
@@ -76,32 +76,29 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
...
-	manifest := getDebugDaemonset()
-	var ds v1.DaemonSet
-
-	if err := yaml.Unmarshal([]byte(manifest), &ds); err != nil {
-		return fmt.Errorf("failed to unmarshal debug daemonset manifest: %w", err)
+func ensureDebugDaemonsets(ctx context.Context, kube *Kubeclient) error {
+	manifests := getDebugDaemonsetManifests()
+	for _, manifest := range manifests {
+		if err := createDebugDeployment(ctx, kube, manifest); err != nil {
+			return err
```
djsly, line 83: should we continue the loop in case only 1 DebugDeployment is failing ?

djsly, line 83: can we add a log to help identify which manifest name is failing ?

e2e/validation.go
```diff
@@ -102,5 +115,16 @@ func commonLiveVMValidators() []*LiveVMValidator {
 				"cloud-config.txt",
 			},
 		),
+		{
+			Description: "check that curl to wireserver fails",
+			Command:     "curl \"http://168.63.129.16/machine/?comp=goalstate\" -H \"x-ms-version: 2015-04-05\" -s --connect-timeout 10",
```
djsly, line 120: the ports that should be blocked are 443 and 32526

e2e/kube.go
UtheMan: We need both to perform all the checks, so i think its better to fail here

UtheMan: added this to the error message

e2e/kube.go
```diff
@@ -76,32 +76,29 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
...
 	}
-
-	return nil
 }
 
-func getDebugDaemonset() string {
-	return `apiVersion: apps/v1
+func getDebugDaemonsetTemplate(deploymentName, targetNodeLabel string, isHostNetwork bool) string {
+	return fmt.Sprintf(`apiVersion: apps/v1
 kind: Deployment
```
cameronmeissner, line 99: lol, can we change the naming if it's not actually a daemonset

cameronmeissner: probably better for it to actually be a daemonset and then just only select for the system pool, that way we get new replicas if we get new system pool nodes

cameronmeissner: ugh, I don't wanna make you do that here...

I guess if we want to switch it over to actual daemonset then we can leave the naming how it currently is

e2e/kube.go
```diff
@@ -76,32 +76,29 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
...
 metadata:
-  name: &name debug
+  name: &name %s
```
cameronmeissner, line 101: nit: specify positional argument IDs to make this a bit more clear

e.g. `%[1]s`, `%[2]t`, etc.

cameronmeissner: (it's 1-indexed)

e2e/kube.go
```diff
@@ -130,7 +127,26 @@ spec:
           privileged: true
           capabilities:
             add: ["SYS_PTRACE", "SYS_RAWIO"]
-`
+`, deploymentName, isHostNetwork, targetNodeLabel)
+}
+
+func createDebugDeployment(ctx context.Context, kube *Kubeclient, manifest string) error {
```
cameronmeissner, line 133: okay we should align on naming, either put back as daemonset if we're gonna change it to a daemonset in reality or renaming everything else to be deployment

e2e/kube.go
```diff
@@ -130,7 +127,26 @@ spec:
...
+	var ds v1.DaemonSet
```
cameronmeissner, line 134: wait...

how are we marshaling into a DaemonSet if the manifest specifies kind as `Deployment`?

e2e/types.go
```diff
@@ -148,6 +148,10 @@ type LiveVMValidator struct {
 	// IsShellBuiltIn is a boolean flag which indicates whether or not the command is a shell built-in
 	// that will fail when executed with sudo - requires separate command to avoid command not found error on node
 	IsShellBuiltIn bool
+
+	// IsNonHostNetwork is a boolean flags which indicates whether or not the validator should run on a pod that is NOT using
+	// host's network interface. For example when testing connectivity from user pods to certain endpoints, we will set it to true
+	IsNonHostNetwork bool
```
cameronmeissner, line 154: nit: `IsPodNetwork`

e2e/validation.go
```diff
@@ -46,9 +51,18 @@ func runLiveVMValidators(ctx context.Context, t *testing.T, vmssName, privateIP,
...
+		isNonHostValidator := validator.IsNonHostNetwork
+
 		t.Logf("running live VM validator on %s: %q", vmssName, desc)
 
-		execResult, err := pollExecOnVM(ctx, t, opts.clusterConfig.Kube, privateIP, podName, sshPrivateKey, command, isShellBuiltIn)
+		var execResult *podExecResult
+		var err error
+		// Non Host Validators - meaning we want to execute checks through a pod which is NOT connected to host's network
+		if isNonHostValidator {
+			execResult, err = execOnPrivilegedPod(ctx, opts.clusterConfig.Kube, "default", nonHostPodName, command)
```
cameronmeissner, line 62: the difference between `execOnPod` and `execOnPrivilegedPod` is that we wrap the command in `nsenter`, I don't think we actually need that for the new validator?

e2e/validation.go
```diff
@@ -102,5 +116,16 @@ func commonLiveVMValidators() []*LiveVMValidator {
 				"cloud-config.txt",
 			},
 		),
+		{
+			Description: "check that curl to wireserver fails",
+			Command:     "curl 'http://168.63.129.16:32526/vmSettings' --connect-timeout 4",
```
cameronmeissner, line 121: do we actually need the single quotes?

e2e/validation.go
cameronmeissner: I think generally we should be using `execOnPod` for pod validators - `execOnPrivilegedPod` should really only be used for ssh'ing into a target VM (e.g. only used by `execOnVM`)

e2e/types.go
cameronmeissner: honestly I don't think it makes sense to have a `LiveVMValidator` which is really a pod validator... I think we should have a new type called `PodValidator` or similar to distinguish between the 2 rather than overloading this one type

cameronmeissner: we can leave as is and maybe add a TODO to separate pod validators from live VM validators

e2e/kube.go
```diff
@@ -76,32 +76,29 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
...
-	})
+	return nil
+}
 
-	if err != nil {
-		return fmt.Errorf("failed to apply debug daemonset: %w", err)
+func getDebugDaemonsetManifests() []string {
+	return []string{
+		getDebugDaemonsetTemplate(hostNetworkDebugPodNamePrefix, "nodepool1", true),
+		getDebugDaemonsetTemplate(nonHostNetworkDebugPodNamePrefix, "nodepool2", false),
```
cameronmeissner, line 93: im confused, with the new validation, don't we want to schedule a new pod (e.g. just a Pod object, not Deployment/Daemonset) with a node selector selecting for the abe2e node we're testing (by unique name/label), that runs outside the host network?

with the current implementation we're basically creating 2 separate deployments (also not sure the difference between nodepool1 and nodepool2, is nodepool2 the label the e2e nodes will have? don't we want to have a _unique_ selector for each abe2e node?)

cameronmeissner: was your thinking that if all abe2e nodes have the `nodepool2` label then this would automatically schedule onto all abe2e nodes whenever they come up (though it would need to be a DS for that to work)?

UtheMan: This "deployment" is actually a daemonset if you look at the ab e2e cluster resources which are deployed. 

And yeah basically the thinking is that this will just create a separate pod for each of the test nodes and run required checks. The naming there was a little bit confusing between daemonset/deployment, i'm guessing what gets deployed in the end is the type we marshall to, hence we end up with a daemonset.

UtheMan: `Nodepool2` is the label that all the e2e test cases nodes have, that's correct. It was the easiest/quickest way for me to get them to deploy only on the test case nodes, so I ran with it.

e2e/kube.go
UtheMan: That's a good question, I modified Artur's code here - I was also initally confused by the fact that we have deployment yamls but end up with daemonsets on the cluster. But I sorta ran with it...

e2e/kube.go
UtheMan: Fair, i can change the naming to daemonsets. Did that initially but then decided not to for some reason

e2e/kube.go
UtheMan: Naming is inconsistent here, will fix

e2e/validation.go
UtheMan: Can modify that too, one of the errors i ran into was that commands with `ExecOnPod` were not running in bash, which the `execOnPrivilegedPod` does for you, will modify.

e2e/validation.go
UtheMan: Not for the URL, will remove. But they are needed for headers in the second check (which i will add shortly too)

e2e/kube.go
cameronmeissner: okay, can you please also change the `Kind` in the manifest to Daemonset so we avoid all this confusion lol

e2e/pod.go
```diff
@@ -28,6 +34,25 @@ func getDebugPodName(ctx context.Context, kube *Kubeclient) (string, error) {
 	return podName, nil
 }
 
+// Returns the name of a pod that's a member of the 'debug' deployment, running on an aks-nodepool node.
+func getNonHostDebugPodName(ctx context.Context, kube *Kubeclient, labelName, vmssName string) (string, error) {
```
cameronmeissner, line 38: can we rename to something like `findDebugPodNameForVMSS` or something?

`getDebugPodName` above behaves much differently (e.g. doesn't try to search for a particular pod

e2e/validation.go
UtheMan: Got two checks for both ports now

e2e/validation.go
```diff
@@ -102,5 +116,28 @@ func commonLiveVMValidators() []*LiveVMValidator {
...
+			Asserter: func(code, stdout, stderr string) error {
+				if code != "28" {
+					return fmt.Errorf("validator command terminated with exit code %q but expected code 28 (CURL timeout)", code)
+				}
+				return nil
+			},
+			IsPodNetwork: true,
+		},
+		{
+			Description: "check that curl to wireserver port 32526 fails",
```
UtheMan, line 132: Added two separate checks, because I couldn't figure out an easy way to run two CURLs and fetch both exit codes using just a one liner. This will likely be modified/refactored once i get around to adding a separate type of pod validator like discussed in other comments here.

e2e/types.go
```diff
@@ -148,6 +148,10 @@ type LiveVMValidator struct {
 	// IsShellBuiltIn is a boolean flag which indicates whether or not the command is a shell built-in
 	// that will fail when executed with sudo - requires separate command to avoid command not found error on node
 	IsShellBuiltIn bool
+
+	// IsPodNetwork is a boolean flags which indicates whether or not the validator should run on a pod that is NOT using
+	// host's network interface. For example when testing connectivity from user pods to certain endpoints, we will set it to true
+	IsPodNetwork bool
```
r2k1, line 154: Maybe not worth touching now. But I wonder if it's good abstraction. Maybe `LiveVMValidator` can be a function type.

Or `RunScenario` may return a kube clients and we can execute some additional checks in a test function.

e2e/kube.go
r2k1: I didn't write it, just move around. Could be just a typo. Try to fix and see what will happen.

Maybe by coincidence all fields we're using present in both types.

e2e/pod.go
```diff
@@ -13,10 +14,15 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+const (
+	hostNetworkDebugPodNamePrefix    = "debug"
+	nonHostNetworkDebugPodNamePrefix = "debugnonhost"
```
r2k1, line 19: is it a `prefix` or `app: label` value?

e2e/pod.go
```diff
@@ -13,10 +14,15 @@ import (
...
+)
+
 // Returns the name of a pod that's a member of the 'debug' daemonset, running on an aks-nodepool node.
-func getDebugPodName(ctx context.Context, kube *Kubeclient) (string, error) {
+func getDebugPodName(ctx context.Context, kube *Kubeclient, labelName string) (string, error) {
```
r2k1, line 23: `labelName` => `appLabel`

r2k1 (approved):

e2e/validation.go
r2k1: Maybe should add a third one?

That curl from the same node succeeds.

e2e/kube.go
UtheMan: Yep fixed it and it seems to be working the same. So naming is consistent now :)

e2e/validation.go
UtheMan: Did you run that CURL while exec'd into the pod from non-host debug DS? If not, then it will work from the node itself, but not from pod network (so those non-host debug pods)

r2k1: It's hard to test "something shouldn't happen".
Environment may change, wireserver can move to another port but our tests still will be happy proving nothing.
I just want to see the wireserver is actually present on the node before testing the access is blocked.

e2e/types.go
UtheMan: Yeah I agree we should probably extract this out and either add another type of "PodNetwork" validator or introduce some sort of extra abstraction over the validator concept in e2e. Happy to look into that a bit later!

e2e/pod.go
UtheMan: good point, its the name of our daemonset. Will adjust the name

e2e/validation.go
UtheMan: I believe we can treat wireserver address/ports as "const", since they are the same all across azure: https://learn.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16.

I agree we could add a check from the node itself, I understand what you meant now, adding third check!

e2e/kube.go
```diff
@@ -76,32 +76,29 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
...
-	})
+	return nil
+}
 
-	if err != nil {
-		return fmt.Errorf("failed to apply debug daemonset: %w", err)
+func getDebugDaemonsetManifests() []string {
+	return []string{
+		getDebugDaemonsetTemplate(hostNetworkDebugAppLabel, "nodepool1", true),
+		getDebugDaemonsetTemplate(podNetworkDebugAppLabel, "nodepool2", false),
```
cameronmeissner, line 93: can we add a comment saying that `nodepool2` is to select all abe2e nodes?

e2e/types.go
```diff
@@ -148,6 +148,10 @@ type LiveVMValidator struct {
 	// IsShellBuiltIn is a boolean flag which indicates whether or not the command is a shell built-in
 	// that will fail when executed with sudo - requires separate command to avoid command not found error on node
 	IsShellBuiltIn bool
+
+	// IsPodNetwork is a boolean flags which indicates whether or not the validator should run on a pod that is NOT using
+	// host's network interface. For example when testing connectivity from user pods to certain endpoints, we will set it to true
```
cameronmeissner, line 153: pls add a TODO here for the refactor

e2e/pod.go
cameronmeissner: pending rename

cameronmeissner (approved):

e2e/pod.go
UtheMan: Coming right up, sorry missed this one

PR #4780, opened 2024-08-06
cleanup: cleanup azurecns

AlisonB319:
**What type of PR is this?**
/kind cleanup
cleaning up older versions of azurecns

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/components.json
+++ b/parts/linux/cloud-init/artifacts/components.json
@@ -71,6 +71,7 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cns:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
+        "v1.4.52",
         "v1.5.32",
         "v1.6.0"
       ],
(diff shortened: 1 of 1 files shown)


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -71,18 +71,10 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cns:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.4.52",
```
rbtr, line 74: 1.4.52 is still used in AKS (for k8s 1.27)

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -71,18 +71,10 @@
...
-        "v1.5.26",
```
rbtr, line 75: 1.5.26 is also still used in AKS (for k8s 1.28, until we finish rolling out 1.5.32)

rbtr (changes requested):

parts/linux/cloud-init/artifacts/components.json
AlisonB319: Sounds good, I'll leave this PR open for when we are ready to remove 1.5.26

rbtr: 1.5.32 has started rolling out to AKS prod so we can follow up here in ~2 weeks

cameronmeissner:
I believe this was addressed in a separate PR

PR #4785, opened 2024-08-07
feat: allow overrides of installed components in case of embargoed CVEs

timmy-wright:
**What type of PR is this?**

/kind feature

This PR adds the ability to specify overrides to AgentBaker components.json so that Microsoft has the capability of building VHDs with fixes for embargoed CVEs without having the details of the CVE in a public GIT repository. This means that when an embargoed VHD is announced, the AKS team can have the VHDs prepared, rolled out globally, and ready for use ahead of our usual release schedule.

A PR to add a README file to the aks-rp repo is here: https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/10610435

PRD links: 
* short wiki: https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/personalplayground/699564/Embargoed-VHDs
* long word: https://microsoft.sharepoint.com/:w:/t/SIGNodeLifecycle/EdcrgPGTq6xOv1-EQoICer8Bo-7jiZh7uFMGrmAd8o4Yxw?e=FCBvRM 

### Screen snapshot of "run pipeline" screen

image

# Screen snapshot of overrides being applied during a build

image

Fixes #

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,38 @@
...
+    condition: eq('${{ parameters.useOverrides }}', true)
   - bash: |
-        m="linuxVhdMode" && \
-        echo "Set build mode to $m" && \
-        echo "##vso[task.setvariable variable=MODE]$m"
+      set -ex
+      # use rsync rather than cp because it worked when I tested this script. And it's output is pretty.
+      # debated using -u flag here to only copy newer files, but decided that if there are overrides, then
+      # we should use them!
+      rsync -av $(Pipeline.Workspace)/aks-rp/agentbakersvc/overrides/ ./
```
cameronmeissner, line 29: let's make a note somewhere that the `agentbakersvc/overrides` is meant to contain files/directories relative to the AgentBaker repo root

cameronmeissner: this is also where we'd need to be careful about replace vs. merge, right now we're just blindly replacing

cameronmeissner: I think blind replace would pretty much make sense for any file other than components.json

though we also need to think about potential edge cases here - the embargoed VHDs built with overrides will still be bootstrapped with the latest (1P/3P available) CSE/customdata returned by latest agentbakersvc, what if we somehow introduce an issue where the overrides contain a version of a file that is also used at provisioning time (such as cse_install.sh, renamined to provision_install.sh during the build) which is incompatible? (e.g. renaming a function that cse_main.sh relies on, etc.)

cameronmeissner: with that in mind ^, I'm wondering if only allowing components.json replace/merge would be prudent

timmy-wright: I think you're right. Have changed to only replace components.json. Decided on replace vs merge because predicting the results of a merge can be hard, and if we need to do the override, whoever does it can just copy the entire existing components.json and edit it.

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,55 @@
...
+      set -euo pipefail
+      # Just copy components.json. If anything else is needed, then someone can modify the pipeline in an emergency.
+      # we check for existence of both src and destination file. As this pipeline will only be run if an override is happening,
+      # then we it's an error to not do an override.
+      COMPONENT_JSON_SRC=$(Pipeline.Workspace)/aks-rp/agentbakersvc/overrides/components.json
+      COMPONENT_JSON_DEST=./parts/linux/cloud-init/artifacts/components.json
+      if [ ! -f "${COMPONENT_JSON_DEST}" ]; then
+        echo destination components.json file does not exist, not copying as it must have been moved in a refactor: ${COMPONENT_JSON_DEST}
+        exit 1
+      else
```
cameronmeissner, line 34: nit: redundant `else`

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,55 @@
...
+    condition: eq('${{ parameters.useOverrides }}', true)
+  - bash: |
+      set -euo pipefail
+      # Just copy components.json. If anything else is needed, then someone can modify the pipeline in an emergency.
+      # we check for existence of both src and destination file. As this pipeline will only be run if an override is happening,
+      # then we it's an error to not do an override.
+      COMPONENT_JSON_SRC=$(Pipeline.Workspace)/aks-rp/agentbakersvc/overrides/components.json
+      COMPONENT_JSON_DEST=./parts/linux/cloud-init/artifacts/components.json
+      if [ ! -f "${COMPONENT_JSON_DEST}" ]; then
+        echo destination components.json file does not exist, not copying as it must have been moved in a refactor: ${COMPONENT_JSON_DEST}
```
cameronmeissner, line 32:
```suggestion
        echo "destination components.json file does not exist, not copying as it must have been moved in a refactor: ${COMPONENT_JSON_DEST}"
```

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,55 @@
...
+    path: aks-rp
+    fetchDepth: 1
+    fetchTags: false
+    lfs: false
+    condition: eq('${{ parameters.useOverrides }}', true)
+  - bash: |
+      set -euo pipefail
+      # Just copy components.json. If anything else is needed, then someone can modify the pipeline in an emergency.
+      # we check for existence of both src and destination file. As this pipeline will only be run if an override is happening,
+      # then we it's an error to not do an override.
```
cameronmeissner, line 28: not quite sure what you meant here

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,55 @@
...
+        exit 1
+      else
+        echo Found file to overwrite: ${COMPONENT_JSON_DEST}
+      fi
+
+      if [ ! -f "${COMPONENT_JSON_SRC}" ]; then
+        echo src components.json file does not exist, not copying: ${COMPONENT_JSON_SRC}
+        exit 1
+      else
+        echo Found source components.json to use for overrides: ${COMPONENT_JSON_SRC}
```
cameronmeissner, line 42: nit: redundant `else`

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,55 @@
...
+      COMPONENT_JSON_DEST=./parts/linux/cloud-init/artifacts/components.json
+      if [ ! -f "${COMPONENT_JSON_DEST}" ]; then
+        echo destination components.json file does not exist, not copying as it must have been moved in a refactor: ${COMPONENT_JSON_DEST}
+        exit 1
+      else
+        echo Found file to overwrite: ${COMPONENT_JSON_DEST}
+      fi
+
+      if [ ! -f "${COMPONENT_JSON_SRC}" ]; then
+        echo src components.json file does not exist, not copying: ${COMPONENT_JSON_SRC}
```
cameronmeissner, line 39:
```suggestion
        echo "src components.json file does not exist, not copying: ${COMPONENT_JSON_SRC}"
```

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -8,8 +8,15 @@ variables:
...
-
 parameters:
+  - name: useOverrides
+    displayName: Use component overrides
+    type: boolean
+    default: false
+  - name: overrideBranch
+    displayName: Branch in aks-rp to use for overrides
+    type: string
+    default: master
```
cameronmeissner, line 19: this is fine for default, though I think it's worth noting that in these situations I don't think we'll be overriding from master, since otherwise we'd have to wait for the overrides to be merged into master before running the corresponding build

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -1,13 +1,55 @@
...
+    type: string
+    default: 1604
+  - name: useOverrides
+    type: boolean
+    default: false
+  - name: overrideBranch
+    type: string
+    default: master
 steps:
+  - checkout: self
```
cameronmeissner, line 12: does this incur any extra work in the non-overrides case?

timmy-wright: Nope. The catch with multiple repositories is once you specify a checkout step, the default checkout step is removed, so this step brings it back. If anything, it might be more performant than before, because we're specifying not to retrieve git history or tags.

.pipelines/.vsts-vhd-builder-release.yaml
timmy-wright: Will add that to the docs when I write then :)

.pipelines/templates/.builder-release-template.yaml
timmy-wright: Removed.

.pipelines/templates/.builder-release-template.yaml
timmy-wright: Removed.

.pipelines/templates/.builder-release-template.yaml
timmy-wright: Github won't let me commit this, not sure why, so have done it myself.

.pipelines/templates/.builder-release-template.yaml
timmy-wright: Have tried to reword text.

cameronmeissner (approved):

PR #4786, opened 2024-08-07
refactor: container image lister as go binary, add support for reporting image size

cameronmeissner:
refactor: container image lister as go binary, add support for reporting image size

**What type of PR is this?**

refactors container image list (bom) generation during VHD builds to use a new go binary rather than an ad-hoc and error-prone bash script, also adds support for reporting the size (in bytes) of each listed container image

example diff (CBLMarinerV2Gen2ARM64): https://www.diffchecker.com/QiHqB0kW/ 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/lister/go.mod
```diff
@@ -0,0 +1,62 @@
+module github.com/Azure/agentbaker/vhdbuilder/lister
+
+go 1.21
```
cameronmeissner, line 3: this makes me sad but the dev container we use doesn't support go 1.22...

packer.mk
```diff
@@ -106,3 +106,13 @@ build-nbcparser-all:
 build-nbcparser-binary:
 	@echo "Building nbcparser binary"
 	@bash -c "pushd nbcparser && CGO_ENABLED=0 GOOS=linux GOARCH=$(ARCH) go build -o bin/nbcparser-$(ARCH) main.go && popd"
+
+build-lister-binary:
+ifeq (${ARCHITECTURE},X86_64)
+	@echo "Building lister binary for amd64"
+	@bash -c "pushd vhdbuilder/lister && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o bin/lister main.go && popd"
```
djsly, line 113: could we use a variable for the GOARCH and only have one `go build` line ?

```
GOARCH=amd64
ifeq (${ARCHITECTURE},ARM64)
    GOARCH=arm64
endif
@echo "Building lister binary for ${GOARCH}"
@bash -c "pushd vhdbuilder/lister && CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} go build -o bin/lister main.go && popd"
```

cameronmeissner: I tried that exact approach and I was having trouble getting it working, I can revisit though

cameronmeissner: refactored

PR #4787, opened 2024-08-07
test: don't update existing cluster

r2k1:
Currently, we re-use long-running clusters.

The cluster name is based on a configuration hash, so any change in the cluster configuration should lead to the creation of a new cluster.

Currently, the CreateOrUpdate AKS cluster command is called on every test run. It seems to trigger an update even if nothing has changed. Concurrent updates are not allowed and lead to test failures (for example, when tests are executed concurrently for different PRs).

It should reduce test flakiness and slightly speed up test execution.


e2e/cluster.go
```diff
@@ -147,8 +148,29 @@ func hash(cluster *armcontainerservice.ManagedCluster) string {
...
+	if errors.As(err, &azErr) && azErr.StatusCode == 404 {
+		return createNewAKSClusterWithRetry(ctx, t, cluster)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cluster %q: %w", *cluster.Name, err)
+	}
+	switch *existingCluster.Properties.ProvisioningState {
+	case "Succeeded":
+		return &existingCluster.ManagedCluster, nil
+	case "Creating", "Updating":
```
cameronmeissner, line 163: should we add "Upgrading" here as well for completeness? granted I don't think we'd ever actually be in that situation unless AKS force-upgrades us

cameronmeissner (approved):

PR #4792, opened 2024-08-08
feat: use new configmap toggle structure

timmy-wright:
**What type of PR is this?**

/kind feature

The vhd override toggle structure is changing so we need to modify the way we lookup version overrides.

This change modifies the way toggles are passed into AgentBaker. Instead of a "GetMap" and "GetString" methods, a GetLinuxSigImageVersion" method is passed in. This means we can delete all the toggle code in AgentBaker and come code in RP that converts from the RP Rollout toggle structure to the AB toggle structure.

The aks-rp PR to provide this method (and the new toggles) is here: 

https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/10612305

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


pkg/agent/bakerapi.go
```diff
@@ -78,7 +78,9 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context, conf
 		e := toggles.NewEntityFromNodeBootstrappingConfiguration(config)
 		imageVersionOverrides := agentBaker.toggles.GetLinuxNodeImageVersion(e)
 		if imageVersion, ok := imageVersionOverrides[string(distro)]; ok {
```
cameronmeissner, line 80: we can probably simplify this

```suggestion
		if imageVersion := imageVersionOverrides[string(distro)]; imageVersion != "" {
		    nodeBootstrapping.SigImageConfig.Version = imageVersion
		}
```

pkg/agent/bakerapi_test.go
```diff
@@ -184,11 +184,19 @@ var _ = Describe("AgentBaker API implementation tests", func() {
 
 		It("should return the correct bootstrapping data when linux node image version override is present", func() {
 			toggles.Maps = map[string]agenttoggles.MapToggle{
-				"linux-node-image-version": func(entity *agenttoggles.Entity) map[string]string {
+				"vhd-types": func(entity *agenttoggles.Entity) map[string]string {
 					return map[string]string{
-						string(datamodel.AKSUbuntu1604): "202402.27.0",
+						"vhd-type": "override",
```
cameronmeissner, line 189: this would be `1P` in real cases right?

timmy-wright: It would have been, but I reverted this change. Didn't want to leak 1p vs 3p terms into this repo.

pkg/agent/bakerapi.go
```diff
@@ -77,8 +77,12 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context, conf
 		// handle node image version toggle/override
 		e := toggles.NewEntityFromNodeBootstrappingConfiguration(config)
 		imageVersionOverrides := agentBaker.toggles.GetLinuxNodeImageVersion(e)
-		if imageVersion, ok := imageVersionOverrides[string(distro)]; ok {
+		if imageVersion := imageVersionOverrides[string(distro)]; imageVersion != "" {
 			nodeBootstrapping.SigImageConfig.Version = imageVersion
+
+			if imageVersion != "" {
```
cameronmeissner, line 83: don't need this nested if anymore

pkg/agent/bakerapi_test.go
```diff
@@ -339,10 +366,18 @@ var _ = Describe("AgentBaker API implementation tests", func() {
...
 			toggles.Maps = map[string]agenttoggles.MapToggle{
-				"linux-node-image-version": func(entity *agenttoggles.Entity) map[string]string {
+				"vhd-types": func(entity *agenttoggles.Entity) map[string]string {
 					return map[string]string{
-						string(datamodel.AKSUbuntu1804): "202402.27.0",
+						"vhd-type": "override",
+					}
+				},
+				"linux-node-image-version": func(entity *agenttoggles.Entity) map[string]string {
+					if entity.Fields["vhd-type"] == "override" {
```
cameronmeissner, line 375: probably don't want references to `vhd-types` in tests if we're not gonna end up using it

timmy-wright: Removed, but then added it back as part of the split of files.

pkg/agent/bakerapi.go
timmy-wright: Removed.

pkg/agent/bakerapi.go
```diff
@@ -100,7 +100,7 @@ func (agentBaker *agentBakerImpl) GetLatestSigImageConfig(sigConfig datamodel.SI
 	if !distro.IsWindowsDistro() {
 		e := toggles.NewEntityFromEnvironmentInfo(envInfo)
 		imageVersionOverrides := agentBaker.toggles.GetLinuxNodeImageVersion(e)
-		if imageVersion, ok := imageVersionOverrides[string(distro)]; ok {
+		if imageVersion, ok := imageVersionOverrides[string(distro)]; ok && imageVersion != "" {
```
cameronmeissner, line 103: we should be able to simplify everywhere to how we're doing it now on line 80

timmy-wright: Done.

pkg/agent/bakerapi_test.go
```diff
@@ -462,65 +411,5 @@ var _ = Describe("AgentBaker API implementation tests", func() {
 				Expect(config.Gallery).To(Equal("aksazurelinux"))
 			}
 		})
-
-		It("should return correct value for all existing distros with linux node image version override", func() {
```
cameronmeissner, line 466: just want to confirm we're not losing any coverage by removing this

timmy-wright: I don't think we are. However, I've added it back - figured out a way to simulate the old behavior. Better safe than sorry.

cameronmeissner (approved):

PR #4796, opened 2024-08-08
feat: support kubelet serving certificate rotation

cameronmeissner:
**What type of PR is this?**

adds to support to AgentBaker for kubelet serving certificate rotation on both Linux and Windows.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
@Bryce-Soghigian for adding new variable to Karpenter contract

pkg/agent/baker.go
```diff
@@ -266,46 +266,57 @@ func normalizeResourceGroupNameForLabel(resourceGroupName string) string {
...
-			profile.KubernetesConfig.ContainerRuntime == "containerd" {
-			for _, flag := range dockerShimFlags {
-				delete(kubeletFlags, flag)
-			}
+	kubeletFlags := config.KubeletConfig
+
+	// If using kubelet config file, disable DynamicKubeletConfig feature gate and remove dynamic-config-dir
+	// we should only allow users to configure from API (20201101 and later)
+	delete(kubeletFlags, "--dynamic-config-dir")
+	delete(kubeletFlags, "--non-masquerade-cidr")
```
djsly, line 278: what's the link with `non-masquerade-cidr` ?

djsly:
unless I"m blind :) this PR only adds logic to properly configure kubelet correct ? with he new kubelet configurations, do we need to configure something else for those new config to work ?

djsly (approved):

cameronmeissner:
> unless I"m blind :) this PR only adds logic to properly configure kubelet correct ? with he new kubelet configurations, do we need to configure something else for those new config to work ?

the only thing we should need to make sure of when `--rotate-server-certificates` is set is that we don't clobber it with `tls-cert-file` and `tls-private-key-file`, which the new code already takes care of

besides that, there's nothing else for us to do on the AgentBaker side - RP will only start settings this flag based on FE validation logic (gated behind toggles), and at the same we will also start deploying the component which will approve the CSRs that kubelet creates at cert bootstrap/rotation time when `--rotate-server-certificates` is set

pkg/agent/baker.go
cameronmeissner: this is old code that needs to remain, though I couldn't tell you the exact reason why we filter out `--non-masquerade-cidr` - I know for a fact that `--dynamic-config-dir` not being filtered out previously has caused outages

AbelHu (approved):
LGTM. Thanks for the PR.

Devinwong (approved):

pkg/agent/utils.go
```diff
@@ -449,6 +488,7 @@ func getAKSKubeletConfiguration(kc map[string]string) *datamodel.AKSKubeletConfi
 		EnforceNodeAllocatable:         strings.Split(kc["--enforce-node-allocatable"], ","),
 		StreamingConnectionIdleTimeout: datamodel.Duration(kc["--streaming-connection-idle-timeout"]),
 		RotateCertificates:             strToBool(kc["--rotate-certificates"]),
+		ServerTLSBootstrap:             strToBool(kc["--rotate-server-certificates"]),
```
anujmaheshwari1, line 491: will the flag always be set? iirc from the validator, this flag is set for 1.19 and above? what about provisioning for clusters < 1.19? i believe there are still some unsupported clusters out there, if they were to do SP reset, they'd get new logic and then this would fail?

IF they are using custom kubelet config. happy to leave this as is if you think we will always have the flag set, otherwise probably a good idea to first check if its present and then set it?

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -119,6 +119,7 @@ PROXY_VARS="{{GetProxyVariables}}"
 ENABLE_TLS_BOOTSTRAPPING="{{EnableTLSBootstrapping}}"
 ENABLE_SECURE_TLS_BOOTSTRAPPING="{{EnableSecureTLSBootstrapping}}"
 CUSTOM_SECURE_TLS_BOOTSTRAP_AAD_SERVER_APP_ID="{{GetCustomSecureTLSBootstrapAADServerAppID}}"
+ENABLE_KUBELET_SERVING_CERTIFICATE_ROTATION="{{EnableKubeletServingCertificateRotation}}"
```
anujmaheshwari1, line 122: @tallaxes @Bryce-Soghigian FYI, additive change to cse_cmd.sh

anujmaheshwari1: oops you already tagged Bryce earlier, I missed that!

pkg/agent/utils.go
cameronmeissner: I thought about this as well, though `strToBool` will return false if the flag isn't there (the implementation is quite... interesting)

and the RP validator now always explicitly sets the flag

parts/linux/cloud-init/artifacts/cse_cmd.sh
Bryce-Soghigian: ty mr anuj

pkg/agent/utils.go
cameronmeissner: followed-up offline, will put in a fix to RP, ensuring we never set this flag on k8s < 1.8

parts/linux/cloud-init/artifacts/cse_cmd.sh
tallaxes: It looks like if we don't include this flag at all, the current functionality is preserved (i.e. it is equivalent to setting it to `false`, right?

PR #4800, opened 2024-08-09
draft: schema discussions

Devinwong:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -8,512 +10,18 @@
         "1.8.22"
       ]
     },
+
+    //new
     {
-      "downloadURL": "mcr.microsoft.com/oss/kubernetes/metrics-server:*",
+      "name": "nvidia-cuda",
```
UtheMan, line 16: What is the difference between "name" and "fullPackageName"?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -8,512 +10,18 @@
...
+      "type": "containerimages //extend to OCI |Tar|tgz when merging with packages,
+      "action": "cacheandfetch|fetchonly|config",  //where cacheandfetch is cached at build time and also to fetch at provisioning time if not cached. fetchonly is only fetching at provisioning time but not caching. config is neither cache nor fetch but just for storing data for other purpose. Open to change the names of the values.
```
Devinwong, line 18: @ganeshkumarashok do you still a config for passing the value to the CSE? If not I can remove this. Though even it's needed, we can still discuss if there are better ways to put in the component.json

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -8,512 +10,18 @@
         "1.8.22"
       ]
     },
+
+    //new
     {
-      "downloadURL": "mcr.microsoft.com/oss/kubernetes/metrics-server:*",
+      "name": "nvidia-cuda",
+      "type": "containerimages //extend to OCI |Tar|tgz when merging with packages,
```
cameronmeissner, line 17: nit: `containerimage` -> `image`

parts/linux/cloud-init/artifacts/components.json
Devinwong: full Image Name = registryURL + fullPackageName
name is mainly for applying logic in the scripts. For example, 
```
if name=="nvidia-cuda", do A; 
if name =="nvidia-grid", do b;
```
Another option is to put fullPackageNames, instead of name in the if conditions above.
But I guess there could be cases that the 2 fullPackageNames happen to be the same but the registryURLs are different.
Then we need a name for the feature owner to define the logic separately.

parts/linux/cloud-init/artifacts/components.json
cameronmeissner, line 18: for action, I'd probably use `cache`/`fetch`/`config` (also, what will config actually be used for?), where
`cache` signifies that we must always have the component `cache`d  by the time provisioning ends (whether that be during build or provisioning time doesn't matter), and `fetch`signifies we only fetch at runtime

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -8,512 +10,18 @@
...
+      "action": "cacheandfetch|fetchonly|config",  //where cacheandfetch is cached at build time and also to fetch at provisioning time if not cached. fetchonly is only fetching at provisioning time but not caching. config is neither cache nor fetch but just for storing data for other purpose. Open to change the names of the values.
+    "registryURL": "mcr.microsoft.com/",
```
cameronmeissner, line 19: could we have different fields for different `type`s? e.g. I think for type `image` it would make more sense to have `registryURL`+ `repo`, and for type `tar` we have `downloadURL`, etc.

parts/linux/cloud-init/artifacts/components.json
Devinwong: runtime != provisioning time? What is the difference?

I thought there are only 2 timing to download/install a package? One is at building in the packer VM and the other is RP passing CSE when (provisioning) bootstrapping a node?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -1,5 +1,7 @@
+// This is not a valid json. Just for schema discussion
 {
   "ContainerImages": [
+    //old
```
UtheMan, line 4: Could you share an example based on current containerd components.json? I'm trying to understand what the new schema would look like for a component which has multiple different versions based on the SKU

parts/linux/cloud-init/artifacts/components.json
Devinwong: Right I also thought about this.
Or keep separating `ContainerImages` and `Packages` into 2 groups at root level if they have many differences. Then `type` could be more "local". For example, each Package has a `type: OCI | tar| tgz` etc. ContainerImage has no `type`

UtheMan: Do you mean we would have different sets of fields in the same json? That could be annoying eventually when we will want to parse the json in the code (e.g. for testing)

parts/linux/cloud-init/artifacts/components.json
UtheMan: Oh i see, i was just a bit confused about the discrepancy between names here

parts/linux/cloud-init/artifacts/components.json
Devinwong: Updated in the code. Please update the latest commit.
Now I start thinking we should probably leave 2 groups `ContainerImages` and `Packages` because
- their schema are quite different
- avoid breaking older images. (We will be just adding new keys but not remove/modify them)

parts/linux/cloud-init/artifacts/components.json
Devinwong: You are good. That's a good call out.

parts/linux/cloud-init/artifacts/components.json
cameronmeissner: runtime == provisioning time, I use the terms interchangeably which is probably something I should stop doing

my suggestion is:

`cache` = build time/provisioning time
`fetch` = provisioning time only

parts/linux/cloud-init/artifacts/components.json
cameronmeissner: fair, though I don't think it would be a huge deal

we could always keep container images and package separate to deal with that

jason1028kr: Is there a clear benefit to merging `ContainerImages` and `Packages` schema?

parts/linux/cloud-init/artifacts/components.json
Devinwong: I see. That makes sense.
One thing comes to me: Recently `cni` removed fetching at provisioning time. Instead it just uses whatever cached at build time. So in this case, it's the feature owner who determines how to react to the `cache`, `fetch` value from components.json. For example, in components.json, `cni` set to `cache`. But in the scripts, it will only cache it but not also fetch it.

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -8,412 +10,26 @@
...
-            ]
-          }
-        }
-      }
-    },
     {
       "name": "containerd",
+      "type": "OCI |Tar|tgz|custom",
       "downloadLocation": "/opt/containerd/downloads",
+      "deps": ["runc", "<another package>"],
```
Devinwong, line 32: Adding a `deps` (optional) here so that we can ensure the dependencies are installed before this package. Now we are relying on the order of the packages. But having a `deps` should be more reliable. Need additional implementation in the codes for sure.

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -432,88 +49,28 @@
...
-              "1.4.54",
-              "1.5.28",
-              "1.5.32"
-            ],
-            "downloadURL": "https://acs-mirror.azureedge.net/azure-cni/v${version}/binaries/azure-vnet-cni-linux-${CPU_ARCH}-v${version}.tgz"
+              "1.7.13"
+            ]
+          },
+          "v3-kata": {
+            "versions": []
```
Devinwong, line 64: Kata team raised a request. They don't want `containerd` to be (re)installed because it was already installed somewhere else. I am thinking adding a "kata" at the os release version level. 
**My main question: Mariner.v3-kata is at the same level as Mariner.v2, Mariner.v3. Does this sound making sense?**
And if the versions array is empty, the for loop in install-dependencies.sh will just ignore it.
Also I will need to use FEATURE_FLAGS to determine if it's a Kata release.

Devinwong: This feature is completed

PR #4801, opened 2024-08-09
cleanup: removing netcat from ubuntu, moving to curl to connectivity checks

djsly:
**What type of PR is this?**
/kind cleanup
/kind failing-test

It was reported that the NC (netcat) command uses tcp which only looks if the port of open and listening, it doesnt verify that SSL is confiured, we still need to configure --insecure since the apiserver's certificate is self-signed

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
need a `make generate`

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -22,7 +22,7 @@ installDeps() {
     aptmarkWALinuxAgent hold
     apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
 
-    pkg_list=(ca-certificates ceph-common cgroup-lite cifs-utils conntrack cracklib-runtime ebtables ethtool git glusterfs-client htop init-system-helpers inotify-tools iotop iproute2 ipset iptables nftables jq libpam-pwquality libpwquality-tools mount nfs-common pigz socat sysfsutils sysstat util-linux xz-utils netcat-openbsd zip rng-tools kmod gcc make dkms initramfs-tools linux-headers-$(uname -r) linux-modules-extra-$(uname -r))
+    pkg_list=(ca-certificates ceph-common cgroup-lite cifs-utils conntrack cracklib-runtime ebtables ethtool git glusterfs-client htop init-system-helpers inotify-tools iotop iproute2 ipset iptables nftables jq libpam-pwquality libpwquality-tools mount nfs-common pigz socat sysfsutils sysstat util-linux xz-utils zip rng-tools kmod gcc make dkms initramfs-tools linux-headers-$(uname -r) linux-modules-extra-$(uname -r))
```
cameronmeissner, line 25: are we sure we don't need netcat for other reasons?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -355,10 +355,11 @@ if ! [[ ${API_SERVER_NAME} =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
             VALIDATION_ERR=$ERR_K8S_API_SERVER_DNS_LOOKUP_FAIL
         fi
     else
-        logs_to_events "AKS.CSE.apiserverNC" "retrycmd_if_failure ${API_SERVER_CONN_RETRIES} 1 10 nc -vz ${API_SERVER_NAME} 443" || time nc -vz ${API_SERVER_NAME} 443 || VALIDATION_ERR=$ERR_K8S_API_SERVER_CONN_FAIL
+        logs_to_events "AKS.CSE.apiserverCurl" "retrycmd_if_failure ${API_SERVER_CONN_RETRIES} 1 10 curl -v --insecure https://${API_SERVER_NAME}:443" || time curl -v --insecure "https://${API_SERVER_NAME}:443" || VALIDATION_ERR=$ERR_K8S_API_SERVER_CONN_FAIL
```
cameronmeissner, line 358: do we actually want the `--insecure` option here?

cameronmeissner: I guess it will still use TLS but just won't verify the API server's cert

jason1028kr: It'll return self-signed cert error without it, I think.

djsly: Correct. Since it's self signed. You need  -k to have a 0 return code

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
djsly: No. But how can we be certain ?

cameronmeissner: @phealy any idea?

cameronmeissner:
I think at this point this looks good pending decision on whether we want to keep caching NC and RP integration testing, let me know if you want some help going through the vendor process to run RP e2es

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -10,6 +14,27 @@ removeContainerd() {
...
+}
+
+installLatestCurlManually() {
+    # Too Risky to manage within the components.json file since curl is required for the bootstrap logic to work
+    # currently this logic is only needed for 18.04, we will be able to remove this logic once we no longer support 18.04
+    
+    # renovate: datasource=github-releases depName=curl/curl
+    version="8.9.0"
+    deb_file="/tmp/curl.deb"
+    removeCurl || exit $ERR_CURL_REMOVE_TIMEOUT
```
cameronmeissner, line 28: this will be called with `installDeps`, which is called at build-time + provisioning time

we should probably add some logic here to check if a sufficient version curl is already installed, otherwise we'd be increasing provisioning latency across the board

djsly: fixed!

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -58,6 +89,11 @@ installDeps() {
             exit $ERR_APT_INSTALL_TIMEOUT
         fi
     done
+
+    if [ "${UBUNTU_RELEASE}" == "18.04" ]; then
+        # curl version on 18.04 is buggy when targetting http/2 TLS endpoint, manually installing a newer version
+        installLatestCurlManuallyIfNotPresent
```
cameronmeissner, line 95: maybe we should wrap this in an event so we can have direct reporting on the latency this step incurs during provisioning? would most likely be helpful in cases where we unexpectedly see this step trying to install a new version of curl at runtime

I see we already wrap installDeps in an event as called from cse_main.sh, though having more granular telemetry here would be helpful I think

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -10,6 +14,33 @@ removeContainerd() {
...
+    version="8.9.0"
+
+    if curl -V | grep $version > /dev/null; then
+        echo "curl is already installed and at the right version, skipping manual installation"
+        return
+    fi
+    
+    deb_file="/tmp/curl.deb"
+    removeCurl || exit $ERR_CURL_REMOVE_TIMEOUT
+    retrycmd_if_failure 10 5 10 wget https://curl.haxx.se/download/curl-${version}.tar.gz -O $deb_file || exit $ERR_CURL_DOWNLOAD_TIMEOUT
```
cameronmeissner: does curl exist within microsoft's linux package repo? (packages.microsoft.com), or on some other microsoft-managed mirror?

if so we should use that rather than `curl.haxx.se` I think

djsly: it might exists but the version wouldnt be good, the problem is that the latest available version for 1804 a a bug, so a more recent version compiled with the deps found in 1804, must be used.  I don't expect to be able to find such a package. but I can look

cameronmeissner:
discussed offline that we can give curl `/etc/kubernetes/certs/ca.crt` so we don't have to insecurely skip cert verification

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -10,6 +14,38 @@ removeContainerd() {
...
+    version="8.9.0"
+
+    if snap list curl > /dev/null; then
+        echo "curl is already installed and at the right version, skipping manual installation"
+        return
+    fi
+    
+    deb_file="/tmp/curl.tgz"
+    removeCurl || exit $ERR_CURL_REMOVE_TIMEOUT
+    retrycmd_if_failure 3 5 5 wget https://curl.haxx.se/download/curl-${version}.tar.gz -O $deb_file || exit $ERR_CURL_DOWNLOAD_TIMEOUT
```
cameronmeissner, line 35: didn't we decide on keeping NC around for 1804 to avoid this?

djsly: refresh, I removed this, once I realized I forgot.    Curl works when properly authenticated, no need for netcat

PR #4802, opened 2024-08-09
draft: feat: support fs.inotify.max_user_instances as custom os config

djsly:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/pkg/proto/nbcontract/v1/customlinuxosconfig.proto
+++ b/pkg/proto/nbcontract/v1/customlinuxosconfig.proto
@@ -46,14 +46,15 @@ message SysctlConfig {
   optional int32 NetIpv4NeighDefaultGcThresh3 = 18;
   optional int32 NetNetfilterNfConntrackMax = 19;
   optional int32 NetNetfilterNfConntrackBuckets = 20;
-  optional int32 FsInotifyMaxUserWatches = 21;
-  optional int32 FsFileMax = 22;
-  optional int32 FsAioMaxNr = 23;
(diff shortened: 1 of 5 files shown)


pkg/proto/nbcontract/v1/customlinuxosconfig.proto
```diff
@@ -46,14 +46,15 @@ message SysctlConfig {
...
   optional int32 NetNetfilterNfConntrackBuckets = 20;
-  optional int32 FsInotifyMaxUserWatches = 21;
-  optional int32 FsFileMax = 22;
-  optional int32 FsAioMaxNr = 23;
-  optional int32 FsNrOpen = 24;
-  optional int32 KernelThreadsMax = 25;
-  optional int32 VMMaxMapCount = 26;
-  optional int32 VMSwappiness = 27;
-  optional int32 VMVfsCachePressure = 28;
+  optional int32 FsInotifyMaxUserInstances = 21;
```
UtheMan, line 49: Is this proto contract in use currently? If yes, then reassigning fields like this will break the backward compatibility. To avoid, this field should be added as 

`optional int32 FsInotifyMaxUserInstances = 29;`

and everything else left as is. However, if the files are not in use - that's safe to do (for now)

UtheMan: more context: https://protobuf.dev/programming-guides/proto3/#assigning

PR #4813, opened 2024-08-13
feat: add the container registry as a registry mirror for mcr.microsoft.com when bootstrap container registry is enabled

bingosummer:
**What type of PR is this?**

/kind feature

If bootstrap container registry is enabled, we need to pull the image from the bootstrap container registry instead of MCR. This PR is to add a containerd config to use the container registry as a registry mirror.

For example, when pulling `mcr.microsoft.com/mcr/hello-world`, the containerd will pull `{{GetBootstrapProfileContainerRegistryServer}}/mcr/hello-world`.

If bootstrap container registry is not enabled, nothing is changed.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
Add the container registry as a registry mirror for mcr.microsoft.com when bootstrap container registry is enabled
```

### DIFF
--- a/parts/linux/cloud-init/nodecustomdata.yml
+++ b/parts/linux/cloud-init/nodecustomdata.yml
@@ -415,3 +415,13 @@ write_files:
   owner: root
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+{{- if (ne GetBootstrapProfileContainerRegistryServer "") }}
+# TODO(binxi): need to update for sovereign cloud.
+- path: /etc/containerd/certs.d/mcr.microsoft.com/hosts.toml
+  permissions: "0644"
+  owner: root
+  content: |
+    [host."https://{{GetBootstrapProfileContainerRegistryServer}}"]
+      capabilities = ["pull", "resolve"]
+{{- end}}
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -964,6 +964,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetPrivateEgressProxyAddress": func() string {
 			return config.ContainerService.Properties.SecurityProfile.GetProxyAddress()
 		},
+		"GetBootstrapProfileContainerRegistryServer": func() string {
+			return config.ContainerService.Properties.SecurityProfile.GetPrivateEgressContainerRegistryServer()
+		},
 		"IsArtifactStreamingEnabled": func() bool {
 			return config.EnableArtifactStreaming
 		},
--- a/pkg/agent/baker_test.go
+++ b/pkg/agent/baker_test.go
@@ -1354,11 +1354,16 @@ oom_score = 0
 			func(config *datamodel.NodeBootstrappingConfiguration) {
 				config.ContainerService.Properties.SecurityProfile = &datamodel.SecurityProfile{
 					PrivateEgress: &datamodel.PrivateEgress{
-						Enabled:      true,
-						ProxyAddress: "https://test-pe-proxy",
+						Enabled:                 true,
+						ProxyAddress:            "https://test-pe-proxy",
+						ContainerRegistryServer: "testserver.azurecr.io",
 					},
 				}
-			}, nil),
+			}, func(o *nodeBootstrappingOutput) {
+				containerdConfigFileContent := o.files["/etc/containerd/certs.d/mcr.microsoft.com/hosts.toml"].value
+				Expect(strings.Contains(containerdConfigFileContent, "[host.\"https://testserver.azurecr.io\"]")).To(BeTrue())
+				Expect(strings.Contains(containerdConfigFileContent, "capabilities = [\"pull\", \"resolve\"]")).To(BeTrue())
+			}),
 		Entry("AKSUbuntu2204 IMDSRestriction with enable restriction and insert to mangle table", "AKSUbuntu2204+IMDSRestrictionOnWithMangleTable", "1.24.2",
 			func(config *datamodel.NodeBootstrappingConfiguration) {
 				config.EnableIMDSRestriction = true
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -2242,4 +2242,11 @@ func (s *SecurityProfile) GetProxyAddress() string {
 	return ""
 }
 
+func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
+	if s != nil && s.PrivateEgress != nil && s.PrivateEgress.Enabled {
+		return s.PrivateEgress.ContainerRegistryServer
+	}
+	return ""
+}
+
 // SecurityProfile end.
--- a/pkg/agent/datamodel/types_test.go
+++ b/pkg/agent/datamodel/types_test.go
@@ -2702,3 +2702,44 @@ func TestSecurityProfileGetProxyAddress(t *testing.T) {
 		})
 	}
 }
+
+func TestSecurityProfileGetPrivateEgressContainerRegistryServer(t *testing.T) {
+	testContainerRegistryServer := "https://testserver.azurecr.io"
+	cases := []struct {
+		name            string
+		securityProfile *SecurityProfile
+		expected        string
+	}{
+		{
+			name:            "SecurityProfile nil",
+			securityProfile: nil,
+			expected:        "",
+		},
+		{
+			name:            "PrivateEgress nil",
+			securityProfile: &SecurityProfile{},
+			expected:        "",
+		},
+		{
+			name:            "PrivateEgress disabled",
+			securityProfile: &SecurityProfile{PrivateEgress: &PrivateEgress{Enabled: false}},
+			expected:        "",
+		},
+		{
+			name:            "PrivateEgress enabled",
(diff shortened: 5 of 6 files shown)


pkg/agent/baker.go
```diff
@@ -964,6 +964,12 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetPrivateEgressProxyAddress": func() string {
 			return config.ContainerService.Properties.SecurityProfile.GetProxyAddress()
 		},
+		"GetBootstrapProfileContainerRegistryServer": func() string {
```
cameronmeissner, line 967: let's add a few new UTs to baker_test.go to ensure that `/etc/containerd/certs.d/mcr.microsoft.com/hosts.toml` is getting generated as we expect

bingosummer: Thanks! Added the UT.

`/etc/containerd/certs.d/mcr.microsoft.com/hosts.toml` is generated as expected
https://github.com/Azure/AgentBaker/pull/4813/files#diff-a6f4d8f0504efa4d2a9814560ab246a790447d5cab922f79ae9f8ed643154e10R316

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -414,4 +414,15 @@ write_files:
...
-    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
\ No newline at end of file
+    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+{{ if EnableBootstrapProfileContainerRegistry -}}
+- path: /etc/containerd/certs.d/mcr.microsoft.com/hosts.toml
+  permissions: "0644"
+  owner: root
+  content: |
+    server = "https://mcr.microsoft.com"
```
fseldow, line 424: will it be possible to add one comment for TODO sovereign cloud? since lx and mc are using differernt mcr url

pkg/agent/baker_test.go
```diff
@@ -1359,6 +1359,14 @@ oom_score = 0
 					},
 				}
 			}, nil),
+		Entry("AKSUbuntu2204 with bootstrap container registry enabled", "AKSUbuntu2204+BootstrapContainerRegistry", "1.30.0", func(config *datamodel.NodeBootstrappingConfiguration) {
+			config.ContainerService.Properties.SecurityProfile = &datamodel.SecurityProfile{
+				PrivateEgress: &datamodel.PrivateEgress{
+					Enabled:                 true,
```
cameronmeissner, line 1365: can we add some assertions on the generated file content as well?

something like this: https://github.com/Azure/AgentBaker/blob/master/pkg/agent/baker_test.go#L286-L295

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -414,4 +414,16 @@ write_files:
...
\ No newline at end of file
+    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+{{ if EnableBootstrapProfileContainerRegistry -}}
+# TODO(binxi): need to update for sovereign cloud.
+- path: /etc/containerd/certs.d/mcr.microsoft.com/hosts.toml
+  permissions: "0644"
+  owner: root
+  content: |
+    server = "https://mcr.microsoft.com"
```
karataliu, line 425: >server [](http://example.com/codeflow?start=4&length=6)

don't need this field actually? it's only for fallback. suggest to remove

[https://github.com/containerd/containerd/blob/main/docs/hosts.md#server-field](https://github.com/containerd/containerd/blob/main/docs/hosts.md#server-field)

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -414,4 +414,16 @@ write_files:
...
+
+    [host."https://{{GetBootstrapProfileContainerRegistryServer}}"]
+      capabilities = ["pull", "resolve"]
+{{- end}}
```
karataliu, line 429: >}} [](http://example.com/codeflow?start=7&length=2)

some whitespace changes are not necessary in testdata

can you tweak {{- --}} to remove unneed whitespace when the feature is off?

karataliu (changes requested):
comment added

parts/linux/cloud-init/nodecustomdata.yml
bingosummer:
```
server specifies the default server for this registry host namespace.

When host(s) are specified, the hosts will be tried first in the order listed. If all host(s) are tried then server will be used as a fallback.

If server is not specified then the image's registry host namespace will automatically be used.
```

Do you mean, if the server and registry host namespace are same, then we don't need the server field?

pkg/agent/baker_test.go
bingosummer: added

parts/linux/cloud-init/nodecustomdata.yml
bingosummer: removed "server" field since the server and registry host namespace are same and if server is not specified then the image's registry host namespace will automatically be used.

parts/linux/cloud-init/nodecustomdata.yml
bingosummer: removed unneeded whitespace when the feature is off

pkg/agent/datamodel/types.go
```diff
@@ -2235,11 +2235,25 @@ type PrivateEgress struct {
 	ProxyAddress            string `json:"proxyAddress"`
 }
 
+func (s *SecurityProfile) EnableBootstrapProfileContainerRegistry() bool {
```
karataliu, line 2238: >func [](http://example.com/codeflow?start=0&length=4)

1. can we safe one function?
just check whether server name equals empty in yml template?
Only write that file if server name not empty, which makes sense.


2. organize the names?
assume PrivateEgress == BootstrapProfile

in (*SecurityProfile), just align to property
GetPrivateEgressContainerRegistryServer


In baker function, align it with real api.
EnableBootstrapProfileArtifactSourceCache

Then in future, we might change privateegress to bootstrapprofile without changing baker function.

pkg/agent/baker_test.go
```diff
@@ -1350,15 +1350,20 @@ oom_score = 0
 				Expect(sysctlContent).To(ContainSubstring("net.ipv4.neigh.default.gc_thresh3=16384"))
 				Expect(sysctlContent).To(ContainSubstring("net.ipv4.ip_local_reserved_ports=65330"))
 			}),
-		Entry("AKSUbuntu2204 with SecurityProfile", "AKSUbuntu2204+SecurityProfile", "1.26.0",
+		Entry("AKSUbuntu2204 with bootstrap container registry enabled", "AKSUbuntu2204+BootstrapContainerRegistry", "1.30.0",
 			func(config *datamodel.NodeBootstrappingConfiguration) {
```
karataliu, line 1354: no need to do dir renaming? can just use securityprofile. this helps make the real diff more clear.

we can do dir renaming in one dedicate PR later.

bingosummer: kept securityProfile.

pkg/agent/datamodel/types.go
bingosummer: > can we safe one function?
just check whether server name equals empty in yml template?
Only write that file if server name not empty, which makes sense.

Fixed.

```
"EnableBootstrapProfileArtifactSourceCache": func() bool {
			return config.ContainerService.Properties.SecurityProfile.EnablePrivateEgress() && config.ContainerService.Properties.SecurityProfile.GetPrivateEgressContainerRegistryServer() != ""
		},
```

bingosummer: > organize the names?

fixed. But in future, we still need to update baker function when changing privateegress to bootstrapprofile.

```
		"GetBootstrapProfileContainerRegistryServer": func() string {
			return config.ContainerService.Properties.SecurityProfile.GetPrivateEgressContainerRegistryServer()
		},
		"EnableBootstrapProfileArtifactSourceCache": func() bool {
			return config.ContainerService.Properties.SecurityProfile.EnablePrivateEgress() && config.ContainerService.Properties.SecurityProfile.GetPrivateEgressContainerRegistryServer() != ""
		},
```

karataliu: no template change then.

karataliu (approved):
LGTM

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -414,4 +414,14 @@ write_files:
   encoding: gzip
   owner: root
   content: !!binary |
-    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
\ No newline at end of file
+    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+{{- if (ne GetBootstrapProfileContainerRegistryServer "") }}
+# TODO(binxi): need to update for sovereign cloud.
```
cameronmeissner, line 420: do we also need to guard against sov/agc case for now?

cameronmeissner (approved):

parts/linux/cloud-init/nodecustomdata.yml
bingosummer: Currently, this feature is not supported in sov/agc. `GetBootstrapProfileContainerRegistryServer` is empty string in sov/agc, so that this file won't be added.

PR #4817, opened 2024-08-13
chore: add logic to handle serialize image pulls for custom kubelet config

anujmaheshwari1:
This PR adds logic to update AKS Kubelet Config datamodel to include SerializeImagePulls, taken from: https://github.com/kubernetes/kubernetes/blob/5fdab96c6cd7a3f9c179725476c6933757aea2ba/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L489C2-L495C66

This will ensure that when custom kubelet config is specified for k8s 1.31 and above, serialize-image-pulls value is also set in the config file along with other parameters.

**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong (approved):

pkg/agent/utils.go
```diff
@@ -465,6 +465,13 @@ func getAKSKubeletConfiguration(kc map[string]string) *datamodel.AKSKubeletConfi
 		ResolverConfig:                 kc["--resolv-conf"],
 		ContainerLogMaxSize:            kc["--container-log-max-size"],
 	}
+
+	// Serialize Image Pulls will only be set for k8s >= 1.31, currently RP doesnt pass this flag
+	// It will starting with k8s 1.31
+	if value, exists := kc["--serialize-image-pulls"]; exists {
+		kubeletConfig.SerializeImagePulls = strToBoolPtr(value)
```
cameronmeissner, line 472: did we also wanna add a UT for this?

pkg/agent/datamodel/types.go
```diff
@@ -2088,6 +2088,13 @@ type AKSKubeletConfiguration struct {
 	Default: []
 	+optional. */
 	AllowedUnsafeSysctls []string `json:"allowedUnsafeSysctls,omitempty"`
+	// serializeImagePulls when enabled, tells the Kubelet to pull images one
```
djsly, line 2091: by default parallel pull is used ?

djsly (approved):

pkg/agent/datamodel/types.go
anujmaheshwari1: by default, serial pull is used.

pkg/agent/utils.go
anujmaheshwari1: with the original change I had, this was already UTed, but since I reverted it, added UTs specifically for this now

cameronmeissner (approved):

PR #4824, opened 2024-08-15
fix: set kubelet --node-ip to address node startup perf regression

wedaly:
**What type of PR is this?**

/kind bug

In k8s >= 1.29 kubelet no longer sets node internalIP when using external cloud provider https://github.com/kubernetes/kubernetes/pull/121028

This regresses node startup performance in Azure CNI Overlay and Podsubnet clusters, which require the node to be assigned an internal IP before configuring pod networking. To improve node startup performance, explicitly set `--node-ip` so kubelet sets the internal IP when it registers the node.
If this fails, skip setting --node-ip, which is safe because cloud-node-manager will assign it later anyway.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
In k8s 1.29+ pass --node-ip to kubelet to address node startup performance regression.
```


self-contained/bootstrap_config.sh
```diff
@@ -386,6 +386,30 @@ ensureDHCPv6() {
...
+            [ -n "$ipv4Addr" ] && nodeIPAddrs+=("$ipv4Addr")
+            ipv6Addr=$(echo $imdsOutput | jq -r '.network.interface[0].ipv6.ipAddress[0].privateIpAddress // ""')
+            [ -n "$ipv6Addr" ] && nodeIPAddrs+=("$ipv6Addr")
+            nodeIPArg=$(IFS=, ; echo "${nodeIPAddrs[*]}") # join, comma-separated
+            if [ -n "$nodeIPArg" ]; then
+                echo "Adding --node-ip=$nodeIPArg to kubelet flags"
+                KUBELET_FLAGS="$KUBELET_FLAGS --node-ip=$nodeIPArg"
+            fi
+        fi
+    fi
```
wedaly, line 411: I copy-pasted this here. Is that the right way to do this?

pkg/agent/testdata/RawUbuntuContainerd/line70.sh
```diff
@@ -388,6 +388,23 @@ ensureDHCPv6() {
 ensureKubelet() {
     KUBELET_DEFAULT_FILE=/etc/default/kubelet
     mkdir -p /etc/default
+
+    if semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.29.0"; then
+        imdsOutput=$(curl -s -H Metadata:true --noproxy "*" --max-time 5 "http://169.254.169.254/metadata/instance?api-version=2021-02-01" 2> /dev/null)
```
wedaly, line 393: any risk that this will diverge from what cloud-node-manager sets?

Before k8s 1.29 kubelet would try to infer the IP address from Linux routes and interfaces, but would set only the IPv4 address, so there was already some difference between kubelet and cloud-node-manager.

djsly: if cloud-manager sets the IP differently after, is this a bad thing ? could the initial `--node-ip` be temporary and only required for quicker bootstrapping ?

cameronmeissner: right, I think that's the behavior this PR is trying to restore for us specifically, I guess upstream decided to stop doing that since hostNetwork pods created by cloud-provider assigned an external IP would keep the original IP used at kubelet start-time

that's at least my understanding from reading the linked issue/PR

cameronmeissner: `but would set only the IPv4 address` I see that we'd now be setting both v4 and v6 addresses, is that potentially an issue?

self-contained/bootstrap_config.sh
cameronmeissner: cc @Devinwong @lilypan26 

I think that's fine for now... I wonder if we should add some automation which would get invoked by `make generate` to copy relevant changes over to `self-contained`

I know we have similar logic for various things in the RP repo

pkg/agent/testdata/RawUbuntuContainerd/line70.sh
wedaly: I think setting both IPv4 and IPv6 is an improvement over what we had before. Previously there was a race where kubelet would set IPv4 only, then cloud-node-manager would set IPv4+IPv6 later. This also matches the behavior on Windows,  see https://github.com/Azure/AgentBaker/pull/4148.

I'd be open to setting just IPv4 if we think that's less risky, but I'd prefer to align with what cloud-node-manager and Windows CSE do.

self-contained/bootstrap_config.sh
wedaly: that makes sense. Do E2E tests cover k8s 1.29+ with self-contained VHD? want to make sure this doesn't regress that configuration.

pkg/agent/testdata/RawUbuntuContainerd/line70.sh
wedaly: updated to set just ipv4 address, since that's the only one we can get from nmagent

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -398,6 +398,30 @@ ensureDHCPv6() {
...
+
+    # In k8s >= 1.29 kubelet no longer sets node internalIP when using external cloud provider
+    # https://github.com/kubernetes/kubernetes/pull/121028
+    # This regresses node startup performance in Azure CNI Overlay and Podsubnet clusters, which require the node to be
+    # assigned an internal IP before configuring pod networking.
+    # To improve node startup performance, explicitly set `--node-ip` to the IP returned from IMDS so kubelet sets
+    # the internal IP when it registers the node.
+    # If this fails, skip setting --node-ip, which is safe because cloud-node-manager will assign it later anyway.
+    if semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.29.0"; then
+        imdsOutput=$(curl -s -H Metadata:true --noproxy "*" --max-time 5 "http://169.254.169.254/metadata/instance?api-version=2021-02-01" 2> /dev/null)
```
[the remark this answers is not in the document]
wedaly: updated to retrieve the primary ip from nmagent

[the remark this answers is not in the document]
wedaly: double-checked with cloud provider team, and confirmed that cloud-node-manager uses IMDS today: https://github.com/kubernetes-sigs/cloud-provider-azure/blob/d7ddb818430f821f43c0d88b21e3fb99129a3c3a/pkg/provider/azure_instances_v1.go#L116-L159

Updated this PR to continue using IMDS out of concern for diverging from cloud-node-manager, and also to assign IPv6 which isn't available from the nmagent API.

Reached out to IMDS team to clarify whether it's safe to rely on this for retrieving the primary IP. Will keep this PR in draft until I hear back.

pkg/agent/testdata/RawUbuntuContainerd/line70.sh
wedaly: changed it back to setting both ipv4 and ipv6, to match what cloud-node-manager does. This also solves the race condition where host network pods are assigned only IPv4 if they schedule before cloud-node-manager updates the IP.

e2e/validators.go
```diff
@@ -271,3 +273,37 @@ func runcVersionValidator(version string) *LiveVMValidator {
 		},
 	}
 }
+
+// K8s 1.29+ should set --node-ip in kubelet flags due to behavior change in
+// https://github.com/kubernetes/kubernetes/pull/121028
+func kubeletNodeIPValidator() *LiveVMValidator {
```
wedaly, line 279: this depends on https://github.com/Azure/AgentBaker/pull/4849

wedaly: rebased to get that change

parts/linux/cloud-init/artifacts/cse_config.sh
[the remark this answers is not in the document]
wedaly: confirmed with IMDS / nmagent team that the primary NIC is always first. I still have an open question to networking team about whether primary IP is always first, but based on internal design documents and what I'm seeing in the code I believe it is.

e2e/scenario_test.go
```diff
@@ -334,6 +338,9 @@ func Test_marinerv2_azurecni(t *testing.T) {
 				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-cblmariner-v2-gen2"
 				nbc.AgentPoolProfile.Distro = "aks-cblmariner-v2-gen2"
 			},
+			LiveVMValidators: []*LiveVMValidator{
+				kubeletNodeIPValidator(),
```
cameronmeissner, line 342: any reason you explicitly added the new validation to these scenarios, rather than as a common validator?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,30 @@ ensureDHCPv6() {
...
     mkdir -p /etc/default
+
+    # In k8s >= 1.29 kubelet no longer sets node internalIP when using external cloud provider
+    # https://github.com/kubernetes/kubernetes/pull/121028
+    # This regresses node startup performance in Azure CNI Overlay and Podsubnet clusters, which require the node to be
+    # assigned an internal IP before configuring pod networking.
+    # To improve node startup performance, explicitly set `--node-ip` to the IP returned from IMDS so kubelet sets
+    # the internal IP when it registers the node.
+    # If this fails, skip setting --node-ip, which is safe because cloud-node-manager will assign it later anyway.
+    if semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.29.0"; then
```
cameronmeissner, line 416: could we extract this to a separate function and wrap it in `logs_to_events`?

the event could be called something like `AKS.CSE.ensureKubelet.setNodeIPFlag`

e2e/scenario_test.go
wedaly: yes, it's because there's one test `Test_ubuntu2204privatekubepkg` that pins the k8s version to 1.25.6, so it won't have the node IP flag set.

parts/linux/cloud-init/artifacts/cse_config.sh
wedaly: updated to move it into a helper function with `logs_to_events`. Can see that it's working in the provisioning logs:
image

cameronmeissner (approved):

PR #4826, opened 2024-08-16
feat: remove generated testdata from PRs

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.gitattributes
```diff
@@ -0,0 +1 @@
+pkg/agent/testdata/**/*  linguist-generated=true
```
AlisonB319, line 1: This seems to make sure that anything inside pkg/agent/testdata is always auto-collapsed. But there is no github supported way of not having the files in a PR. 

There is a file filter option right under "commits" but since we have the "line33.sh" and most of our changes are also bash files, it wont hide everything. Its too bad that there is not a way to filter on a file path

cameronmeissner:
we definitely need to commit changes made to testdata to preserve the integrity of our unit tests, I find it's fairly simple to just collapse the testdata folder when reviewing PRs

PR #4828, opened 2024-08-16
feat: Kata build doesn't want containerd to be reinstalled.

Devinwong:
**What type of PR is this?**
/kind feature

1. Kata build doesn't want containerd to be reinstalled. Add a feature to support it while keeping the logic visible in components.json
 - Note that in package `containerd`, `marinerkata`, the `versions` array is left with a special tag `<SKIP>`. This is to tell the install-dependencies.sh not to install any `containerd` version for Kata SKU.
2. Update the readme.
3. Update the unit tests.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -443,6 +443,11 @@
             ]
           }
         },
+        "marinerkata": {
```
cameronmeissner, line 446: could we just have logic directly in the install scripts to skip containerd installation if kata is enabled? adding a new object here just for an empty array seems unnecessary

I guess we can leave it like this if we think that we'll have divergent versions of containerd needed for kata-enabled SKUs in the future

Devinwong: That's a good call out indeed. 
We want the logic can be completely visible in the centralized component.json. If we put the logic in the scripts to skip containerd installation for Kata build, then someone will need to look into the scripts again to figure out why kata build doesn't install containerd. And you also made a good point that, in the future Kata build could be requesting divergent versions for MarinerKata v2, v3, etc. So I would tend to keep it like this.

cameronmeissner: what do you think about adding a new keyword to the components schema, something like `base` as a valid version string that indicates we should just take it from the base image?

I think something like that might be a bit more readable/understandable than this

Devinwong: Good idea! Let me try <BASE>

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -101,18 +106,20 @@ installContainerdWithManifestJson() {
 }
 
 installContainerRuntime() {
+    local isKata=${1:-false}
```
cameronmeissner, line 109: I figured `IS_KATA` would already be accessible from this file due to how it's `source`d, would rather just use that directly than add a parameter

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -121,7 +121,7 @@ else
     echo "Golden image; skipping dependencies installation"
 fi
 
-logs_to_events "AKS.CSE.installContainerRuntime" installContainerRuntime
+logs_to_events "AKS.CSE.installContainerRuntime" installContainerRuntime "${IS_KATA}"
```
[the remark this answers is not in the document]
Devinwong: It's actually doing 2 things. logs_to_events in the meantime executing installContainerRuntime. `${@}: This executes the remaining arguments as a command.`
So I think we need it.

spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh
```diff
@@ -15,114 +15,120 @@ Describe 'cse_install.sh'
...
 
-        It 'returns downloadURIs.default.current.downloadURL of package cni-plugins for MARINER'
+        It 'returns downloadURIs.default.current.downloadURL of package cni-plugins for MARINER' "false"
             package=$(readPackage "cni-plugins")
             When call returnPackageDownloadURL "$package" "MARINER" "some_mariner_version"
             The variable PACKAGE_DOWNLOAD_URL should equal "https://acs-mirror.azureedge.net/cni-plugins/v\${version}/binaries/cni-plugins-linux-\${CPU_ARCH}-v\${version}.tgz"
         End
 
-        It 'returns downloadURIs.default.current.downloadURL of package azure-cni for MARINER'
+        It 'returns downloadURIs.default.current.downloadURL of package azure-cni for MARINER' "false"
```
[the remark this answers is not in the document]
Devinwong: no this is typo. Let me revise this file

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Oh that sounds good. Now I don't need it to be passed in. Thanks

parts/linux/cloud-init/artifacts/cse_main.sh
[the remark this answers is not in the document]
Devinwong: I can just use the global var `IS_KATA` now. Removing this passed in.

pkg/agent/testdata/AKSUbuntu1604+Containerd/line33.sh
```diff
@@ -115,7 +115,7 @@ else
     echo "Golden image; skipping dependencies installation"
 fi
 
-logs_to_events "AKS.CSE.installContainerRuntime" installContainerRuntime
+logs_to_events "AKS.CSE.installContainerRuntime" installContainerRuntime "${IS_KATA}"
```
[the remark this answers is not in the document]
Devinwong: Will remove this. Please see the cse_main.sh for reason.

spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh
```diff
@@ -15,114 +15,120 @@ Describe 'cse_install.sh'
...
             package=$(readPackage "runc")
-            When call returnPackageDownloadURL "$package" "MARINER" "some_mariner_version"
+            When call returnPackageDownloadURL "$package" "MARINER" "some_mariner_version" "false"
             The variable PACKAGE_DOWNLOAD_URL should equal ''
         End
 
         It 'returns downloadURIs.mariner.current.downloadURL of package containerd for MARINER'
             package=$(readPackage "containerd")
-            When call returnPackageDownloadURL "$package" "MARINER" "some_mariner_version"
+            When call returnPackageDownloadURL "$package" "MARINER" "some_mariner_version" "false"
```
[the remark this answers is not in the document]
Devinwong: make sense. Updated.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -51,13 +52,17 @@ installContainerdWithComponentsJson() {
     
     containerdPackage=$(jq ".Packages" "$COMPONENTS_FILEPATH" | jq ".[] | select(.name == \"containerd\")") || exit $ERR_CONTAINERD_VERSION_INVALID
     PACKAGE_VERSIONS=()
-    returnPackageVersions "${containerdPackage}" "${os}" "${os_version}"
+    returnPackageVersions "${containerdPackage}" "${os}" "${os_version}" "${isKata}"
```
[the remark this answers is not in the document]
Devinwong: make sense. Updated.

parts/linux/cloud-init/artifacts/components.json
Devinwong: I ended up using `<SKIP>` because in this case it's not using the containerd from base image. There is another step installing a custom containerd for kata so SKIP should be more descriptive.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -199,10 +206,19 @@ for p in ${packages[*]}; do
   #getting metadata for each package
   name=$(echo "${p}" | jq .name -r)
   PACKAGE_VERSIONS=()
-  returnPackageVersions ${p} ${OS} ${OS_VERSION}
+  if [[ "${OS}" == "${MARINER_OS_NAME}" && "${isKata}" == "true" ]]; then
+    local os=${MARINER_KATA_OS_NAME}
```
Devinwong, line 210: introduce a local `os` instead of updating the global `OS`

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -339,6 +347,9 @@ installCNI() {
         os_version="current"
     fi
     os_version="${UBUNTU_RELEASE}"
+    if [[ "${os}" == "${MARINER_OS_NAME}" && "${isKata}" == "true" ]]; then
```
cameronmeissner, line 350: should be `"${IS_KATA}"`?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -199,10 +206,19 @@ for p in ${packages[*]}; do
   #getting metadata for each package
   name=$(echo "${p}" | jq .name -r)
   PACKAGE_VERSIONS=()
-  returnPackageVersions ${p} ${OS} ${OS_VERSION}
+  if [[ "${OS}" == "${MARINER_OS_NAME}" && "${isKata}" == "true" ]]; then
```
cameronmeissner, line 209: also should be `"${IS_KATA}"` as well I think

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -199,10 +206,19 @@ for p in ${packages[*]}; do
...
+    os=${MARINER_KATA_OS_NAME}
+  fi
+  returnPackageVersions ${p} ${os} ${OS_VERSION}
   PACKAGE_DOWNLOAD_URL=""
-  returnPackageDownloadURL ${p} ${OS} ${OS_VERSION}
+  returnPackageDownloadURL ${p} ${os} ${OS_VERSION}
   echo "In components.json, processing components.packages \"${name}\" \"${PACKAGE_VERSIONS[@]}\" \"${PACKAGE_DOWNLOAD_URL}\""
+  
+  # if ${PACKAGE_VERSIONS[@]} count is 0, skip to next package
+  if [[ ${#PACKAGE_VERSIONS[@]} -eq 0 ]]; then
```
cameronmeissner, line 218: now that we're using `<SKIP>`, would this case ever happen?

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: ya my bad. If bash script supports checking if a variable is undefined, that will be awesome

vhdbuilder/packer/install-dependencies.sh
Devinwong: I left a fool-proof mechanism here in case someone leave the array empty or undefined, although in the readme we ask people to put `<SKIP>`. What do you think?

cameronmeissner: yep sounds good, just wanted to clarify

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner: the `-u` option used with the `set` command will have the script forcibly produce an error when an unbound variable is accessed, though we must not be using that option here (and probably because other things would break otherwise...)

cameronmeissner (approved):

PR #4842, opened 2024-08-20
fix: exit with cleanup.sh code after build fails

zachary-bailey:
**What type of PR is this?**

/kind bug

If the vhd build fails, only the cleanup step should be executed. After this process finishes, the test-scan-and-cleanup.sh should exit with the exit code of the cleanup script. This ensures that engineers will see that the cleanup succeeded, even if the build fails. It also prevents test and scan from being run on a failed build.

Pipeline is currently executing test and scan scripts on builds that fail.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/vhdbuilder/packer/test-scan-and-cleanup.sh
+++ b/vhdbuilder/packer/test-scan-and-cleanup.sh
@@ -36,7 +36,8 @@ SIG_VERSION=$(az sig image-version show \
 
 if [ -z "${SIG_VERSION}" ]; then
   echo -e "\nBuild step did not produce an image version. Running cleanup and then exiting."
-  retrycmd_if_failure 2 3 "${SCRIPT_ARRAY[@]}"; exit $?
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -36,7 +36,7 @@ SIG_VERSION=$(az sig image-version show \
 
 if [ -z "${SIG_VERSION}" ]; then
   echo -e "\nBuild step did not produce an image version. Running cleanup and then exiting."
-  retrycmd_if_failure 2 3 "${SCRIPT_ARRAY[@]}" || exit $?
+  retrycmd_if_failure 2 3 "${SCRIPT_ARRAY[@]}"; exit $?
```
cameronmeissner, line 39: could we just have `exit $?` on the proceeding line since that's what we do in most other places?

cameronmeissner (approved):

PR #4843, opened 2024-08-20
test: add build performance step

zachary-bailey:
**What type of PR is this?**

/kind test

This PR changes the benchmarking workflow during the VHD build process to set conditions to merge a build time regression prevention program.

VHD build performance data will now be logged in a separate step after the build/test/scan/cleanup scripts are complete. This also removes the pipeline step that publishes the VHD build performance data as an artifact and simplifies the logic within the benchmarking functions in cse_helpers.sh.

Unsatisfactory VHD build times.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -429,77 +429,40 @@ capture_benchmark() {
...
+    '{($section_name): $total_time_elapsed'})
+    script_object=$(jq -n --argjson script_object "$script_object" --argjson section_object "$section_object" --arg script_name "$(basename $0)" \
+    '$script_object | .[$script_name] += $section_object')
   done
-
-  echo "Benchmarks:"
-  echo "$script_object" | jq -C .
  
-  jq ". += [$script_object]" ${VHD_BUILD_PERF_DATA} > tmp.json && mv tmp.json ${VHD_BUILD_PERF_DATA}
+  jq ". += $script_object" ${VHD_BUILD_PERF_DATA} > tmp.json && mv tmp.json ${VHD_BUILD_PERF_DATA}
```
Devinwong, line 464: If tmp.json already exists, will this command cause any error? Or if you can ensure tmp.json will not exist at this point, it's also fine.

zachary-bailey: Good call since thats such a general file name. I can make it more specific and add a check. Thanks!

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -254,12 +271,6 @@ steps:
       artifactName: 'vhd-image-bom-${{ parameters.artifactName }}'
       targetPath: 'image-bom.json'
     displayName: publish container image list
-  - task: PublishPipelineArtifact@1
```
djsly, line 257: how come we don't need this anymore ?

djsly (approved):

.pipelines/templates/.builder-release-template.yaml
zachary-bailey: Originally, I was publishing the artifact so that it could later be consumed by some service that would evaluate the data and determine if it was within spec. I have since moved the approach to utilizing kusto using the https://github.com/Azure/azure-kusto-go sdk, which allows ingestion from the agent vm during the pipeline. The artifact so far hasn't been used at all as it was just there until I finished the next part of the project. @djsly

Devinwong (approved):

cameronmeissner (approved):

PR #4849, opened 2024-08-21
cleanup: Use k8s 1.29 in AgentBaker E2E tests

wedaly:
**What type of PR is this?**
/kind cleanup

AgentBaker E2E tests were using k8s 1.26, which is past EOL. Update to use k8s 1.29 which is the current default in AKS. This required two other changes to node bootstrapping config:

1. Switch from in-tree to external cloud-provider (`--cloud-config=""` and `--cloud-provider=external`). Otherwise kubelet exits with an error that in-tree cloud-provider is no longer supported ("built-in cloud providers are disabled").
2. Add kubernetes.azure.com/cluster node label. Otherwise, cloud-node-manager and other daemonsets won't schedule on the node, so the node gets stuck with "uninitialized" taint

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/template.go
+++ b/e2e/template.go
@@ -28,7 +28,7 @@ func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
 				ProvisioningState: "",
 				OrchestratorProfile: &datamodel.OrchestratorProfile{
 					OrchestratorType:    "Kubernetes",
-					OrchestratorVersion: "1.26.0",
+					OrchestratorVersion: "1.29.6",
 					KubernetesConfig: &datamodel.KubernetesConfig{
 						KubernetesImageBase:               "",
 						MCRKubernetesImageBase:            "",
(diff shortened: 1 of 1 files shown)


e2e/template.go
```diff
@@ -28,7 +28,7 @@ func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
 				ProvisioningState: "",
 				OrchestratorProfile: &datamodel.OrchestratorProfile{
 					OrchestratorType:    "Kubernetes",
-					OrchestratorVersion: "1.26.0",
+					OrchestratorVersion: "1.29.6",
```
cameronmeissner, line 31: actually, instead of hard-coding this value at all, maybe we can infer this value from the abe2e cluster's control plane properties instead?

@r2k1 maybe something you could look at?

cameronmeissner (approved):

e2e/template.go
wedaly: agreed to hard-code for now to unblock testing on 1.29

r2k1: What happened? Why do you need to change it? Did test cluster auto-upgraded?

r2k1: I think, ideally instead of testing old/current version we should test the latest version

wedaly: 1.26 has been EOL since February: https://kubernetes.io/releases/patch-releases/#non-active-branch-history

I need to update it so I can add E2E tests for this change, which is specific to k8s 1.29 or later: https://github.com/Azure/AgentBaker/pull/4824

PR #4853, opened 2024-08-21
fix: modify timeouts on linux automation pipeline jobs

cameronmeissner:
**What type of PR is this?**

modifies timeout values on linux vhd automation pipeline jobs to ensure that all retries have a chance to be exhausted before job failure

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-automation.yaml
+++ b/.pipelines/.vsts-vhd-automation.yaml
@@ -24,7 +24,7 @@ stages:
     condition: eq('${{ parameters.ImageBump }}', true)
     jobs:
     - job: Automation
-      timeoutInMinutes: 30
+      timeoutInMinutes: 60
       steps:
       - template: ./templates/.configure-azdevops-template.yaml
         parameters:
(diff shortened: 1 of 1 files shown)


.pipelines/.vsts-vhd-automation.yaml
```diff
@@ -69,7 +69,7 @@ stages:
     condition: eq('${{ parameters.CreateRelease }}', true)
     jobs:
     - job: Automation
-      timeoutInMinutes: 60
+      timeoutInMinutes: 0
```
UtheMan, line 72: 0?

cameronmeissner: [The timeoutInMinutes allows a limit to be set for the job execution time. When not specified, the default is 60 minutes. When 0 is specified, the maximum limit is used (described above).](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts)

UtheMan: That will means it would run for up to 6 hours, right?

UtheMan (approved):

PR #4857, opened 2024-08-22
E2E Improvements

r2k1:
Initially I just wanted to try a few tricks to speed up E2E execution. Can't say the results are significant. But still an improvement.

On the way I touched other things:
- update dependencies
- use non-deprecated, corporate-approved images 
- address test flakiness
- minor cleanup


e2e/config/vhd.go
```diff
@@ -237,15 +237,15 @@ func replicateToCurrentRegion(ctx context.Context, t *testing.T, definition sigI
...
+	if _, err := resp.PollUntilDone(ctx, DefaultPollUntilDoneOptions); err != nil {
 		return fmt.Errorf("updating image version target regions: %w", err)
 	}
 
 	return nil
 }
 
 func ensureProvisioningState(version *armcompute.GalleryImageVersion) error {
-	if *version.Properties.ProvisioningState != armcompute.GalleryImageVersionPropertiesProvisioningStateSucceeded {
+	if *version.Properties.ProvisioningState != armcompute.GalleryProvisioningStateSucceeded {
```
cameronmeissner, line 248: are these actually the same? I personally don't see the point in changing this since `version` is a gallery image version rather than a gallery

e2e/template.go
```diff
@@ -448,31 +448,39 @@ func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
...
-    image: mcr.microsoft.com/oss/nginx/nginx:1.21.6
+  - name: mariner
+    image: mcr.microsoft.com/cbl-mariner/busybox:2.0
     imagePullPolicy: IfNotPresent
-    readinessProbe:
+    command: ["sh", "-c"]
+    args:
+    - |
+      mkdir -p /www &&
+      echo '<!DOCTYPE html><html><head><title></title></head><body></body></html>' > /www/index.html &&
```
cameronmeissner, line 465: doesn't nginx accomplish this already?

r2k1: nginx from mcr is a deprecated private image with a lot of stuff in it.

I wanted to use a minimal docker image to speed up pod start up. (it didn't improve much)

e2e/config/vhd.go
r2k1: I upgrade armcompute from v1 to v6. I fixed compilation errors here, original constants are gone

e2e/template.go
r2k1: I wanted to use something like alpine but couldn't find anything reasonable in MCR.

Not sure if we should use other container registries

cameronmeissner: fair enough, I think this is fine as long as it works as we expect, just wanted to make sure

cameronmeissner (approved):

PR #4860, opened 2024-08-22
fix: Don't echo the contents of kubelet client cert files in provisioning log

babbageclunk:
**What type of PR is this?**
/kind bug

We were already suppressing the SP secrets from the cluster provisioning log. Include the kubelet client certificates too.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
Suppress kubelet client certificate details in cluster provisioning log
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -184,6 +184,8 @@ configureK8s() {
     chown root:root "${AZURE_JSON_PATH}"
 
     mkdir -p "/etc/kubernetes/certs"
+
+    set +x
```
cameronmeissner, line 188: oh lol, I see what you meant, it was almost already hidden...

cameronmeissner (approved):

r2k1 (approved):

jason1028kr (approved):

e2e/scenario_test.go
```diff
@@ -901,3 +904,33 @@ func Test_ubuntu2204Wasm(t *testing.T) {
...
+		Description: "tests that none of this set of secrets have been echoed into the provision log",
+		Config: Config{
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDUbuntu2204Gen2Containerd,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+				nbc.ContainerService.Properties.CertificateProfile.ClientPrivateKey = clientPrivateKey
+				nbc.ContainerService.Properties.ServicePrincipalProfile.Secret = spSecret
+			},
+			LiveVMValidators: []*LiveVMValidator{
+				FileExcludesContentsValidator(logPath, clientPrivateKey, "client private key"),
```
cameronmeissner, line 930: I'd probably opt for having these added to the set of common live VM validators we execute in _all_ secnarios, rather than having a specific scenario for this purpose

e2e/scenario_test.go
```diff
@@ -901,3 +904,33 @@ func Test_ubuntu2204Wasm(t *testing.T) {
...
 }
+
+func Test_noEchoedSecretsInProvisionLog(t *testing.T) {
+	logPath := "/var/log/azure/cluster-provision.log"
+	secretContent := "**** this secret should not be logged ****"
+	clientPrivateKey := base64.StdEncoding.EncodeToString([]byte("ClientPrivateKey: " + secretContent))
+	spSecret := base64.StdEncoding.EncodeToString([]byte("ServicePrincipalSecret: " + secretContent))
+
+	// Getting the cluster so we can also check for the bootstrap token.
+	cluster, err := ClusterKubenet(newTestCtx(t), t)
```
cameronmeissner, line 915: nit: this var name is a bit misleading, it's really the NodeBootstrappingConfiguration we use to bootstrap nodes against the e2e cluster

cameronmeissner: ah, I see you're copying what other tests are doing, we should probably change the naming a bit...

fine to leave as is

PR #4866, opened 2024-08-26
fix: make IMDS restriction commands as kubelet's pre-start script

norshtein:
**What type of PR is this?**
/kind fix

This make IMDS restriction commands as kubelet's pre-start script. That's because IMDS restriction is implemented by iptables, and iptables rule will be cleaned when the node is restarted, so we need to add those iptables rule as kubelet's pre-start script to make sure it can be executed every time the node is restarted.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -534,6 +551,66 @@ EOF
 # for DNS.
 iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 80 -j DROP
 iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 32526 -j DROP
+EOF
+
+    # As iptables rule will be cleaned every time the node is restarted, we need to ensure the rule is applied every time kubelet is started.
+    primaryNicIP=$(getPrimaryNicIP)
```
cameronmeissner, line 557: I'd opt for making this a new script we bake onto the VHD directly and then invoke via kubelet pre-start with required parameters

cameronmeissner: we can also add to `nodecustomdata.yaml` to ensure we don't create a dependency on a particular VHD version as well

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -534,6 +551,66 @@ EOF
...
+}
+
+disableIMDSRestriction() {
+    echo "Checking whether IMDS restriction rule exists in mangle table..."
+    iptables -t mangle -S | grep -- '-d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker ensureIMDSRestriction for IMDS restriction feature" -j DROP'
+    if [[ \$? -ne 0 ]]; then
+        echo "IMDS restriction rule does not exist in mangle table, no need to delete"
+    else
+        echo "Deleting IMDS restriction rule from mangle table..."
+        iptables -t mangle -D FORWARD ! -s \$primaryNicIP -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker ensureIMDSRestriction for IMDS restriction feature" -j DROP || exit ${ERR_DELETE_IMDS_RESTRICTION_RULE_FROM_MANGLE_TABLE}
```
cameronmeissner, line 596: if we do it this way I'm pretty sure we'll be losing the exit codes as well, since we start kubelet via: https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_helpers.sh#L287-L299

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: > I'd opt for making this a new script we bake onto the VHD directly and then invoke via kubelet pre-start with required parameters

upon further inspection, it looks like the script already knows how to resolve everything at runtime (I guess mainly the NIC IP from IMDS), thus we won't have to worry about passing in any required parameters when invoking it via ExecStartPre

norshtein: I also thought about making the script as custom data, but I encountered a question about this: the script know how to enable or disable IMDS restriction, but it don't know whether is should perform enable or disable operation, which is controlled by `ENABLE_IMDS_RESTRICTION` parameter. This parameter is provided to AgentBaker in `NodeBootstrapConfig` by AKS RP. If we move the script to custom data, how could we dynamically inject `ENABLE_IMDS_RESTRICTION` value?

norshtein: one more question, 

> we can also add to nodecustomdata.yaml to ensure we don't create a dependency on a particular VHD version as well

could you explain more on this? Per my understanding, current implementation will take effect in all Linux distributions after new AgentBakersvc is deployed(I assume all Linux distribution is using kubelet.service).

cameronmeissner: > I also thought about making the script as custom data, but I encountered a question about this: the script know how to enable or disable IMDS restriction, but it don't know whether is should perform enable or disable operation, which is controlled by `ENABLE_IMDS_RESTRICTION` parameter. This parameter is provided to AgentBaker in `NodeBootstrapConfig` by AKS RP. If we move the script to custom data, how could we dynamically inject `ENABLE_IMDS_RESTRICTION` value?

we should be able to easily do this via kubelet service drop-in

cameronmeissner: > one more question,
> 
> > we can also add to nodecustomdata.yaml to ensure we don't create a dependency on a particular VHD version as well
> 
> could you explain more on this? Per my understanding, current implementation will take effect in all Linux distributions after new AgentBakersvc is deployed(I assume all Linux distribution is using kubelet.service).

generally we try to avoid generating scripts at runtime (via CSE/customdata) and bake as much as possible onto the VHD ahead of time - however, there are some cases in the wild where older VHDs that don't have certain files already on-disk are bootstrapped with the _latest_ version of CSE/customdata, which assumes that those certain are files are on disk, which ends up causing breakage

in your case, I'd opt for putting the script onto the VHD via packer, while _also_ adding a new entry to the customdata payload for the script itself to make sure that even if the latest version of CSE/customdata is executed on a VHD that is missing the script, it will be generated on the fly.

norshtein:
Hi @cameronmeissner , I've refactored the code to use both packer and custom data to cover the most part of the script logic, and generated a drop-in file inside the CSE, could you please help have a review? Thanks!

parts/linux/cloud-init/artifacts/cse_config.sh
norshtein: That makes great sense, refactored code as suggested, thanks for the explanation!

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -402,6 +402,66 @@ write_files:
     iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 80 -j DROP
     iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 32526 -j DROP
 
+- path: /opt/azure/containers/ensure_imds_restriction.sh
+  permissions: "0755"
+  owner: root
+  content: |
```
cameronmeissner, line 408: we don't need to duplicate the script here:

1. add new const to: https://github.com/Azure/AgentBaker/blob/0515c7aee97fc172541691a715ce779ca87cb8e5/pkg/agent/const.go#L43
2. use const and add the b64-encoded script to `cloudInitData` used by the go template: https://github.com/Azure/AgentBaker/blob/0515c7aee97fc172541691a715ce779ca87cb8e5/pkg/agent/variables.go#L20

for example: https://github.com/Azure/AgentBaker/blob/0515c7aee97fc172541691a715ce779ca87cb8e5/parts/linux/cloud-init/nodecustomdata.yml#L262

pkg/agent/testdata/AKSUbuntu1604+Containerd/line105.sh
```diff
@@ -10,6 +10,7 @@ RestartSec=2
 EnvironmentFile=/etc/default/kubelet
 SuccessExitStatus=143
 ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
+ExecStartPre=/bin/bash /opt/azure/containers/ensure_imds_restriction.sh
```
cameronmeissner, line 13: it looks like, but just wanted to confirm that `ensure_imds_restriction.sh` is idempotent - specifying it as an ExecStartPre means that _any_ time the kubelet service is started/restarted, the script will be executed

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
```diff
@@ -389,6 +389,23 @@ ensureDHCPv6() {
     retrycmd_if_failure 120 5 25 modprobe ip6_tables || exit $ERR_MODPROBE_FAIL
 }
 
+getPrimaryNicIP() {
+    local sleepTime=1
+    local maxRetries=10
+    local i=0
+    local ip=""
+
+    while [[ $i -lt $maxRetries ]]; do
```
cameronmeissner, line 398: you can probably just reuse this instead: https://github.com/Azure/AgentBaker/blob/0515c7aee97fc172541691a715ce779ca87cb8e5/parts/linux/cloud-init/artifacts/cse_helpers.sh#L160

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
```diff
@@ -518,6 +535,18 @@ EOF
...
 iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 80 -j DROP
 iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 32526 -j DROP
+EOF
+
+    primaryNicIP=$(getPrimaryNicIP)
+    ENSURE_IMDS_RESTRICTION_DROP_IN="/etc/systemd/system/kubelet.service.d/10-ensure-imds-restriction.conf"
+    mkdir -p "$(dirname "${ENSURE_IMDS_RESTRICTION_DROP_IN}")"
+    touch "${ENSURE_IMDS_RESTRICTION_DROP_IN}"
+    chmod 0600 "${ENSURE_IMDS_RESTRICTION_DROP_IN}"
+    tee "${ENSURE_IMDS_RESTRICTION_DROP_IN}" > /dev/null <<EOF
```
cameronmeissner, line 545: nit: if we use `cat` instead of `tee` then we won't have to worry about piping stdout to `/dev/null`

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
```diff
@@ -0,0 +1,55 @@
...
+ensureIMDSRestrictionRule() {
+    if [[ $insertRuleToMangleTable == true ]]; then
+        echo "Before inserting IMDS restriction rule to mangle table, checking whether the rule already exists..."
+        iptables -t mangle -S | grep -- '-d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker ensureIMDSRestriction for IMDS restriction feature" -j DROP'
+        if [[ $? -eq 0 ]]; then
+            echo "IMDS restriction rule already exists in mangle table, returning..."
+            return 0
+        fi
+        echo "Inserting IMDS restriction rule to mangle table..."
+        iptables -t mangle -I FORWARD 1 ! -s "$primaryNicIP" -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker ensureIMDSRestriction for IMDS restriction feature" -j DROP || exit ${ERR_INSERT_IMDS_RESTRICTION_RULE_INTO_MANGLE_TABLE}
```
cameronmeissner, line 18: I don't believe you'll be able to access these exit codes anymore - you'd need to define them as constants within the script itself or simply `exit 1` instead

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# Ensure IMDS restriction rule is correctly set on the node, this script will be registered as kubelet service's ExecStart
+# to ensure desired iptables rule will be ensured after the node is restarted.
+# Below 3 variables are passed from kubelet service's drop-in file.
+primaryNicIP="${PRIMARY_NIC_IP}"
+insertRuleToMangleTable="${INSERT_IMDS_RESTRICTION_RULE_TO_MANGLE_TABLE}"
+enableIMDSRestriction="${ENABLE_IMDS_RESTRICTION}"
+
+ensureIMDSRestrictionRule() {
```
cameronmeissner, line 9: you'd need to duplicate it within the script itself, but I'd probably recommend using `logs_to_events` like we currently do currently when configuring the restriction rule: https://github.com/Azure/AgentBaker/blob/0515c7aee97fc172541691a715ce779ca87cb8e5/parts/linux/cloud-init/artifacts/cse_helpers.sh#L389

this way we'll still get corresponding event telemetry

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
```diff
@@ -0,0 +1,55 @@
...
+    if [[ $insertRuleToMangleTable == true ]]; then
```
cameronmeissner, line 10:
```suggestion
    if [[ "{$insertRuleToMangleTable,,"} == "true" ]]; then
```

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
```diff
@@ -0,0 +1,55 @@
...
+    iptables -t filter -S | grep -- '-d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker ensureIMDSRestriction for IMDS restriction feature" -j DROP'
+    if [[ $? -ne 0 ]]; then
+         echo "IMDS restriction rule does not exist in filter table, no need to delete"
+    else
+        echo "Deleting IMDS restriction rule from filter table..."
+        iptables -t filter -D FORWARD ! -s $primaryNicIP -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -m comment --comment "AKS managed: added by AgentBaker ensureIMDSRestriction for IMDS restriction feature" -j DROP || exit ${ERR_DELETE_IMDS_RESTRICTION_RULE_FROM_FILTER_TABLE}
+    fi
+}
+
+if [[ ${enableIMDSRestriction} == "true" ]]; then
```
cameronmeissner, line 51:
```suggestion
if [[ "${enableIMDSRestriction,,}" == "true" ]]; then
```
so we ignore capital/lower case

cameronmeissner:
@anujmaheshwari1 mind taking a look as well?

pkg/agent/testdata/AKSUbuntu1604+Containerd/line105.sh
norshtein: yes, that's indeed idempotent, before inserting/deleting, it will firstly check whether the rule is already there/already deleted. So executing it multiple times will get the same result.

parts/linux/cloud-init/nodecustomdata.yml
norshtein: thanks for the suggestion! updated.

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
norshtein: that function does not support return value (`echo $ip` in our case), so I implemented similar retry logic.

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
norshtein: there're several other places using `tee` to write the drop in file, should be OK to follow the convention?

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
norshtein: updated the code to directly `exit 1`

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
norshtein: updated the script here to call `logs_to_events`, curious will `logs_to_events` function be available here?

norshtein: performed an E2E test, `logs_to_events` is not available in kubelet's execprestart script, so update the code back.

norshtein:
updated code according to code review feedbacks, please help take a new iteration review, many thanks!

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
cameronmeissner: that's fine for now, though changing that convention is on my refactor list :)

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
cameronmeissner: sorry I should've been more clear, I meant that you would have to duplicate the implementation of `logs_to_events` within the script directly in order to use it

on second thought though, I think we'd mainly want event telemetry around the lookup against IMDS, which is still happening via CSE scripts, so I think leaving as is would be fine

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
cameronmeissner: I think you can just have your `getPrimaryNicIP` curl IMDS, check for errors, and echo the result without the retry logic

then call `getPrimaryNicIP` using `retrycmd_if_failure`

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -546,6 +563,19 @@ EOF
 # for DNS.
 iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 80 -j DROP
 iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 32526 -j DROP
+EOF
+
+    # As iptables rule will be cleaned every time the node is restarted, we need to ensure the rule is applied every time kubelet is started.
+    primaryNicIP=$(getPrimaryNicIP)
```
cameronmeissner, line 569: I'd opt for using `logs_to_events` here instead just for the IMDS lookup - it will be accessible within this context

maybe something like `AKS.CSE.ensureKubelet.getPrimaryNicIP` would be a good event name

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+# Ensure IMDS restriction rule is correctly set on the node, this script will be registered as kubelet service's ExecStart
+# to ensure desired iptables rule will be ensured after the node is restarted.
+# Below 3 variables are passed from kubelet service's drop-in file.
+primaryNicIP="${PRIMARY_NIC_IP}"
```
cameronmeissner, line 5: I'd also recommend handling cases where any of the required environment variables are empty/unset

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
norshtein: looks like it will print some help messages, in this case, we'll need to parse the result, it will make the code more complex, we may not want this?

```
$ result=$(retrycmd_if_failure 20 5 25 echo "aaa")
$ echo $result
aaa Executed "echo aaa" 1 times
```

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
norshtein: updated code.

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
norshtein: got your point, let's follow-up in below conversation.

parts/linux/cloud-init/artifacts/cse_config.sh
norshtein: updated code.

norshtein:
@cameronmeissner updated code according to code review feedback, could you please help new iteration review? Thanks! 😄

pkg/agent/testdata/AKSUbuntu1604+Containerd/line70.sh
cameronmeissner: yeah... I really dislike the pattern of "returning" values via echoing to stdout

regardless, I don't want to endlessly block you on this - we can keep what you had originally

cameronmeissner: you could also theoretically define a new helper called something like `retrycmd_if_failure_no_output` or something, which does the same thing but just drops the extra echo statements

will leave up to you

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "Enable IMDS restriction: $enableIMDSRestriction"
+
+# If any required variable is not set, exit with error
+if [[ -z "$primaryNicIP" || -z "$insertRuleToMangleTable" || -z "$enableIMDSRestriction" ]]; then
+    echo "One or more required variables are not set, exiting..."
+    exit 1
+fi
+
+ensureIMDSRestrictionRule() {
+    if [[ "{$insertRuleToMangleTable,,}" == "true" ]]; then
```
cameronmeissner, line 20: sorry I must've typo'd in my original suggestion, it should be `"${insertRuleToMangleTable,,}"`

cameronmeissner:
would you also be able to add a new E2E scenario which tests the `ENABLE_IMDS_RESTRICTION=true` path?

you can add said scenario here: https://github.com/Azure/AgentBaker/blob/master/e2e/scenario_test.go, which you can have utilize a new live VM validator which would check whether the proper IP table rules have been added: https://github.com/Azure/AgentBaker/blob/master/e2e/validators.go

parts/linux/cloud-init/artifacts/ensure_imds_restriction.sh
norshtein: ah, good catch, I even hasn't noticed it, fixed typo.

norshtein:
@cameronmeissner I've fixed typo and added E2E test case, added E2E test suite passed, hopefully we're good to merge this PR 😄 

```
--- PASS: Test_imdsrestriction_filtertable (236.69s)
--- PASS: Test_imdsrestriction_mangletable (172.09s)
```

norshtein:
warmly ping @cameronmeissner

cameronmeissner:
gonna go ahead and merge for you so we can ship this in this week's release

cameronmeissner:
can you also prefix the name of the new e2e tests with the type VHD being used? e.g. `ubuntu2204_imdsrestriction_filtertable`

norshtein:
Renamed test suite & rebased on master & fixed conflicts.

norshtein:
rebased on master & fixed conflict again. @cameronmeissner could you please help merge the PR? Thanks!

PR #4867, opened 2024-08-27
feat: oras pull kubernetes-node for network isolated cluster

fseldow:
**What type of PR is this?**

/kind feature

Add method to pull binaries from mcr via oras

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


karataliu (approved):
lgtm

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -213,6 +222,24 @@ retrycmd_get_tarball() {
         fi
     done
 }
+retrycmd_get_tarball_from_registry() {
```
cameronmeissner, line 225:
```suggestion
retrycmd_get_tarball_with_oras() {
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -109,6 +109,13 @@ ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT=205 # Timeout waiting for credential pr
 
 ERR_CNI_VERSION_INVALID=206 # reference CNI (not azure cni) needs a valid version in components.json
 
+ERR_ORAS_PULL_K8S_FAIL=207 # Error pulling kube-node artifact via oras from registry
+ERR_ORAS_PULL_FAIL_RESERVE_1=208 # Error pulling artifact with oras from registry
```
cameronmeissner, line 113: what are these `_*` exit codes for?

fseldow: they are reserved for other binaries like azure-acr-credentials-provider, cni, containerd-shim-wsam.
Reserved now to make these exit code number to be consecutive

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -420,9 +421,25 @@ extractKubeBinaries() {
         k8s_tgz_tmp="${k8s_downloads_dir}/${k8s_tgz_tmp_filename}"
         mkdir -p ${k8s_downloads_dir}
 
-        retrycmd_get_tarball 120 5 "${k8s_tgz_tmp}" ${kube_binary_url} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
-        if [[ ! -f ${k8s_tgz_tmp} ]]; then
-            exit "$ERR_K8S_DOWNLOAD_TIMEOUT"
+        registry_regex='^.+\/.+\/.+:.+$'
+        if [[ ${kube_binary_url} =~ $registry_regex ]]; then
```
AlisonB319, line 425: I'd recommend using the  BLOCK_OUTBOUND_NETWORK var over a regex check to know if a cluster is NIC

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -420,9 +421,25 @@ extractKubeBinaries() {
...
+            echo "detect kube_binary_url, ${kube_binary_url}, as registry url, will use oras to pull artifact binary"
+            # download the kube package from registry as oras artifact
+            if [[ $(isARM64) == 1 ]]; then
```
AlisonB319, line 428: You can reference my WIP PR - https://github.com/Azure/AgentBaker/pull/4879/files#diff-59521d60320265e88e46f4702029a64c03d857baee2923ff450f4997f6184fe8

for how I set CPU arch to make it a bit cleaner

parts/linux/cloud-init/artifacts/cse_install.sh
karataliu: Thanks for the comment. Based on API review discussion, bootstrap from cache (acr) is supported even when block_outbound_network=false, so cannot use BLOCK_OUTBOUND_NETWORK here.

fseldow: in fact outside the function, i am using BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER as the condition to use oras

fseldow:
it seems the test failure is not related to the pr?
Other prs also fail with same error

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -462,8 +480,17 @@ installKubeletKubectlAndKubeProxy() {
     # if the custom url is not specified and the required kubectl/kubelet-version via private url is not installed, install using the default url/package
     if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" ]] || [[ ! -f "/usr/local/bin/kubelet-${KUBERNETES_VERSION}" ]]; then
         if [[ "$install_default_if_missing" == true ]]; then
+            if [[ ! -z ${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER} ]]; then
```
djsly, line 483: per ALlison, we want to prefer using the OutboundNone/OutboundBlock .    This will tell us we need to use ORAS, and when there is a BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER define, we will overwrite the stored REGISTRY url with the one provided by the user.

djsly (changes requested):

parts/linux/cloud-init/artifacts/cse_install.sh
AlisonB319: https://github.com/Azure/AgentBaker/pull/4867#discussion_r1735415421

According to this comment that they resolved, they want to go with BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER for the flag on which to use oras. And if that is the case, we should discuss if BLOCK_OUTBOUND_NETWORK should be removed.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -420,9 +426,21 @@ extractKubeBinaries() {
         k8s_tgz_tmp="${k8s_downloads_dir}/${k8s_tgz_tmp_filename}"
         mkdir -p ${k8s_downloads_dir}
 
-        retrycmd_get_tarball 120 5 "${k8s_tgz_tmp}" ${kube_binary_url} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
-        if [[ ! -f ${k8s_tgz_tmp} ]]; then
-            exit "$ERR_K8S_DOWNLOAD_TIMEOUT"
+        registry_regex='^.+\/.+\/.+:.+$'
```
AlisonB319, line 429: Can we add a comment about exactly what this regex is doing for future clarity

AlisonB319:
What testing has been done to confirm the PR is working as intended?

We probably need a new unit test as I don't think any of our existing tests will hit 
`if [[ ${kube_binary_url} =~ $registry_regex ]]; then`

parts/linux/cloud-init/artifacts/cse_install.sh
fseldow: From my side BLOCK_OUTBOUND_NETWORK is still used for skipping outbound connection check, so that we do not need to remove?

fseldow:
> What testing has been done to confirm the PR is working as intended?
> 
> We probably need a new unit test as I don't think any of our existing tests will hit `if [[ ${kube_binary_url} =~ $registry_regex ]]; then`

currently, i vender the cse in AKS rp and build one standalone environment to test network isolated clsuter
Tried to add ut in airgapped related tests

parts/linux/cloud-init/artifacts/cse_install.sh
fseldow: comment added in the previous line for now
`# if the url is a registry url, use oras to pull the artifact instead of curl`

parts/linux/cloud-init/artifacts/cse_install.sh
fseldow: To remove confusion,
From aks api, we split network isolated cluster into two part, one is outbound type, another is bootstrap profile.
The outbound type defines if we block the egress traffic.
The bootstrap profile defines the artifact source cache or direct (if it is from acr cache).

The bootstrap profile artifact source cache is prerequisite of outbound type none/blocked.
However, we still allow customer to individually set bootstrap profile artifact source as Cache when outbound type is not none/blocked. (defined in PRD).
In this case, we allow customer egress traffic, it is not one complete network isolated cluster, but we may need to switch the image/binary source to acr cache rule, which is. the reason that we use BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER instead of BLOCK_OUTBOUND_NETWORK as the condition here.

e2e/scenario_test.go
```diff
@@ -649,6 +653,7 @@ func Test_ubuntu2204AirGap(t *testing.T) {
 				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-ubuntu-containerd-22.04-gen2"
 				nbc.AgentPoolProfile.Distro = "aks-ubuntu-containerd-22.04-gen2"
 
+				// TODO: define below in the cluster config instead of mutate bootstrapConfig
```
AlisonB319, line 656: nit: could you add TODO (name): so that other maintainers in the repo know who is looking into it

fseldow: added thx

AlisonB319 (approved):

fseldow:
as v1.29.6 has already cached kubernetes-node in VHD, thus the pr checkin test in this pr will not catch the related code path.
Thus checkout one seperated branch that delete the cached kubectl and kubelet and below is the test pipeline result:
https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=102315629&view=artifacts&pathAsName=false&type=publishedArtifacts

djsly (approved):

PR #4868, opened 2024-08-27
Switch over to using the new agent pool vnet in prod

babbageclunk:
**What type of PR is this?**
/kind cleanup

This switches packer to linking the build VMs to the new nodesig-pool-vnet-<region> vnet in the new prod agent pool RG. Without this the prod build can't use the new pool - the private endpoint SSH connection fails.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/init-variables.sh
+++ b/vhdbuilder/packer/init-variables.sh
@@ -82,10 +82,6 @@ fi
 if [ -z "${VNET_RG_NAME}" ]; then
 	if [ "$MODE" == "linuxVhdMode" ]; then
 		VNET_RG_NAME="nodesig-${ENVIRONMENT}-${PACKER_BUILD_LOCATION}-agent-pool"
-		if [ "${ENVIRONMENT,,}" == "prod" ]; then
-			# for now preserve original functionality for prod builds
-			VNET_RG_NAME="nodesigprod-agent-pool"
-		fi
 	fi
 	if [ "$MODE" == "windowsVhdMode" ]; then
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/init-variables.sh
```diff
@@ -99,10 +99,6 @@ fi
 if [ -z "${VNET_NAME}" ]; then
 	if [ "$MODE" == "linuxVhdMode" ]; then
 		VNET_NAME="nodesig-pool-vnet-${PACKER_BUILD_LOCATION}"
-		if [ "${ENVIRONMENT,,}" == "prod" ]; then
```
cameronmeissner, line 102: make sure to switch the pool name over in the pipeline vars once this merges

cameronmeissner (approved):

vhdbuilder/packer/init-variables.sh
```diff
@@ -82,10 +82,6 @@ fi
 if [ -z "${VNET_RG_NAME}" ]; then
 	if [ "$MODE" == "linuxVhdMode" ]; then
 		VNET_RG_NAME="nodesig-${ENVIRONMENT}-${PACKER_BUILD_LOCATION}-agent-pool"
-		if [ "${ENVIRONMENT,,}" == "prod" ]; then
```
babbageclunk, line 85: I missed this originally, which is why the previous build run against this branch failed. Rerunning with this change as well.

PR #4877, opened 2024-08-28
cleanup: remove apiserver cert in favor of cluster CA

cameronmeissner:
**What type of PR is this?**

remove usage of the apiserver certificate sent from RP via:
https://github.com/Azure/AgentBaker/blob/2350ccdc147ee245e08129d2074b8c0e2a57990a/pkg/agent/datamodel/types.go#L461-L462

in favor of always using the cluster CA cert instead, since the cert used by the apiserver should always be signed by the cluster CA cert

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
I think this works, the only risk is if there's any customer relying on this particular cert/key being on-disk

also still need to get 100% green e2e's (mainly green baring windows failures, which seem to be flakes)

PR #4881, opened 2024-08-28
test: run content tests on CVM images

zachary-bailey:
**What type of PR is this?**

/kind test

This PR runs content tests on CVM images in the build pipeline.

Content tests were not being run on CVM images in the build pipeline.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -1,6 +1,21 @@
 #!/bin/bash
 
-# TODO: assert required variables are set
+required_env_vars=(
```
cameronmeissner, line 3: thanks for adding this!

cameronmeissner:
just to confirm, these tests would only be runnable assuming we have a compatible VM size right? in that case would we want to consider PACKER_BUILD_REGION when deciding whether to run them?

zachary-bailey:
> just to confirm, these tests would only be runnable assuming we have a compatible VM size right? in that case would we want to consider PACKER_BUILD_REGION when deciding whether to run them?

The tests for this are the same as all the others. The building/testing is all on Standard VMs. 

The parking lot/regional capacity issues the other day was in regard to the other CVM branch I am working on that uses CVM hardware to build CVM images, but none of those changes are included in this. 

I checked to see if the CVM content tests would still work when building / testing on CVM hardware though and they executed fine, so we will still be good whenever we make that change @cameronmeissner

cameronmeissner (approved):

PR #4882, opened 2024-08-29
chore: add Force when the windowsnodereset task stopping services

ShiqianTao:
**What type of PR is this?**
/kind feature

Add the -Force parameter when stopping services in the windowsnodereset task to ensure they stop completely, e.g., If the service is depended on by other running services, it cannot be reliably stopped without using the -Force parameter.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/staging/cse/windows/provisioningscripts/windowsnodereset.ps1
+++ b/staging/cse/windows/provisioningscripts/windowsnodereset.ps1
@@ -82,19 +82,19 @@ Unregister-HNSRemediatorScriptTask
 # Stop services
 #
 Write-Log "Stopping kubeproxy service"
-Stop-Service kubeproxy
+Stop-Service kubeproxy -Force
 
 Write-Log "Stopping kubelet service"
-Stop-Service kubelet
+Stop-Service kubelet -Force
 
 if ($global:CsiProxyEnabled) {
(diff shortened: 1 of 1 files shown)


staging/cse/windows/provisioningscripts/windowsnodereset.ps1
```diff
@@ -82,19 +82,19 @@ Unregister-HNSRemediatorScriptTask
...
+Stop-Service kubeproxy -Force
 
 Write-Log "Stopping kubelet service"
-Stop-Service kubelet
+Stop-Service kubelet -Force
 
 if ($global:CsiProxyEnabled) {
     Write-Log "Stopping csi-proxy service"
-    Stop-Service csi-proxy
+    Stop-Service csi-proxy -Force
```
AbelHu, line 92: Which services have unexpected dependency so we have to add `-force`? I think that we need to fix unexpected dependency instead of adding `-Force`

ShiqianTao: Customers may add their own services, so if their service set dependency on `kubelet` and running, the `Stop-Service kubelet` would not stop successfully.

PR #4888, opened 2024-08-30
Jasonjung/set env vars

jason1028kr:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/trivy-scan.sh
```diff
@@ -30,6 +30,13 @@ SEVERITY=${17}
 MODULE_VERSION=${18}
 UMSI_PRINCIPAL_ID=${19}
 UMSI_CLIENT_ID=${20}
+BUILD_RUN_NUMBER=${21}
+export BUILD_REPOSITORY_NAME=${22}
+export BUILD_SOURCEBRANCH=${23}
+export BUILD_SOURCEVERSION=${24}
+export SYSTEM_COLLECTIONURI=${25}
+export SYSTEM_TEAMPROJECT=${26}
+export BUILD_BUILDID=${27}
```
cameronmeissner, line 39: can we use `BUILD_ID` instead?

jason1028kr: yep!

jason1028kr: Actually, the `vuln-to-kusto` module uses 'BUILD_BUILDID' env variable to take in the build id values. I believe there are already some dependencies on this env var in different pipelines. I've change to 'BUILDID' in `vhd-scanning.sh` though.

jason1028kr: https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp?path=/devinfraservices/vuln-to-kusto/internal/environment/detector_ado_pipeline.go&version=GBmaster&line=16&lineEnd=17&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

djsly (commented):
fix comment from Cameron and LGTM

vhdbuilder/packer/trivy-scan.sh
cameronmeissner: so I guess we'd need to change vuln-to-kusto internally before being able to make the change here

I'd personally opt for making the changes but definitely non-blocking

cameronmeissner (approved):

vhdbuilder/packer/trivy-scan.sh
jason1028kr: I think there's already vuln-to-kusto usecases that depend on the `BUILD_BUILDID` variable, so I would like to keep this one as is, if that's okay. I can make the changes if this naming becomes an issue down the line.

PR #4891, opened 2024-08-31
node-bootstrapper

r2k1:
The current implementation is a simple wrapper around CSE scripts and a custom data generator for Linux (previously known as "self-contained" efforts).

This approach moves the script generation logic from the API layer into the VHD itself, making virtual machines more consistent and allowing for significant simplification.

By having a Go application as the main entry point to provision the VHD, we unlock a lot of new possibilities.

Switching to the new generation should be as simple as replacing `GetNodeBootstrapping` with `GetNodeBootstrappingV2`.

No changes are introduced to the current VHD provisioning.


installer/build.sh
```diff
@@ -0,0 +1,4 @@
+GOOS=linux GOARCH=amd64 go build -o ./dist/installer-linux-amd64
```
smith1511, line 1: Maybe prefix with "node-..." to make the binaries more self-descriptive?

pkg/agent/bakerapi.go
```diff
@@ -85,6 +87,20 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context, conf
 	return nodeBootstrapping, nil
 }
 
+func (agentBaker *agentBakerImpl) GetNodeBootstrappingV2(ctx context.Context, config *datamodel.NodeBootstrappingConfiguration) (*datamodel.NodeBootstrapping, error) {
```
smith1511, line 90: GetNodeBootstrappingForSelfContained()?

Devinwong:
Who will trigger the main.go and providing the nodebootstrapping config? Assuming it's baked in the VHD, right?

r2k1:
The binary is compiled and added to every new linux VHD.

AgentBaker Server is preparing and running this CSE command:
```
fmt.Sprintf(
		`bash -c "(echo '%s' | base64 -d > config.json && mkdir -p /var/log/azure && /opt/azure/node-bootstrapper provision --provision-config=config.json) > /var/log/azure/node-bootstrapper.log 2>&1"`,
		base64.StdEncoding.EncodeToString(configJSON),
	)
```

We may change it the future. My goal was to exactly replicate current functionality with minimum changes.

In the future we may change this behaviour to a daemon watching for a config file.

e2e/validators.go
```diff
@@ -269,6 +269,21 @@ func containerdVersionValidator(version string) *LiveVMValidator {
 	}
 }
 
+func textValidator(path string, expected string) *LiveVMValidator {
```
cameronmeissner, line 272: doesn't `FileHasContentsValidator` accomplish the same thing?

e2e/nodebootstrapping.go
```diff
@@ -52,11 +52,11 @@ func getNodeBootstrappingForValidation(ctx context.Context, nbc *datamodel.NodeB
 	if err != nil {
 		return nil, err
 	}
-	nodeBootstrapping, err := ab.GetNodeBootstrapping(ctx, nbc)
-	if err != nil {
-		return nil, err
+	if config.Config.EnableNodeBootstrapperTest {
+		return ab.GetNodeBootstrappingV2(ctx, nbc)
+	} else {
```
cameronmeissner, line 57: nit: redundant else

e2e/scenario_test.go
```diff
@@ -892,6 +892,7 @@ func Test_ubuntu2204gpuNoDriver(t *testing.T) {
 }
 
 func Test_ubuntu2204privatekubepkg(t *testing.T) {
+	t.Skip("this VHD can't be updated and self-contained logic doesn't work for it")
```
cameronmeissner, line 895: we probably don't want to blindly disable this, would there be a way we could keep it around for non self-contained?

pkg/agent/baker.go
```diff
@@ -1001,6 +979,44 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+		profile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey] != "" {
+		return profile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey]
+	}
+	if profile.KubeletDiskType == datamodel.TempDisk {
+		return datamodel.TempDiskContainerDataDir
+	}
+	return cs.Properties.OrchestratorProfile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey]
+}
+
+func HasDataDir(config *datamodel.NodeBootstrappingConfiguration) bool {
```
cameronmeissner, line 996: could we add some UTs for these while we're at it?

pkg/agent/datamodel/types.go
```diff
@@ -1742,6 +1742,9 @@ type NodeBootstrappingConfiguration struct {
 	// CNI, which will overwrite the `filter` table so that we can only insert to `mangle` table to avoid
 	// our added rule is overwritten by Cilium.
 	InsertIMDSRestrictionRuleToMangleTable bool
+
+	// Version is require for node-bootstrapper application to determine the version of the config file.
```
cameronmeissner, line 1746:
```suggestion
	// Version is required for node-bootstrapper application to determine the version of the config file.
```

node-bootstrapper/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
cameronmeissner, line 1: fine for now, though might be worth considering a Makefile

node-bootstrapper/main.go
```diff
@@ -0,0 +1,368 @@
...
+		os.Exit(1)
+	}
+	slog.Info("node-bootstrapper finished")
+}
+
+func Run(ctx context.Context) error {
+	if len(os.Args) < 2 {
+		return errors.New("missing command argument")
+	}
+	switch os.Args[1] {
```
cameronmeissner, line 44: we could probably use something like `cobra` to accomplish this

cameronmeissner:
will take another closer look tomorrow

e2e/validators.go
r2k1: I don't think I indented to touch this part. Probably needs to be rebased with master.

e2e/scenario_test.go
r2k1: Can be done with adding `if config.EnableNodeBootstrapperTest`

pkg/agent/baker.go
r2k1: I didn't really touch anything there. 

I exported a few private methods. And did a few "right-click => extract method". So, I can share exactly the same logic of file templating with node-bootstrapper.

installer/build.sh
timmy-wright: Changed.

pkg/agent/bakerapi.go
timmy-wright: Changed.

e2e/nodebootstrapping.go
timmy-wright: Removed it.

pkg/agent/datamodel/types.go
timmy-wright: Fixed.

e2e/validators.go
timmy-wright: Have removed this and replaced with FileHasContentsValidator

pkg/agent/baker.go
timmy-wright: Have added tests for extracted methods.

node-bootstrapper/build.sh
timmy-wright: I agree - using the same build system as the rest of AgentBaker is important. Will change this in a later PR. I have got the testing and linting working with the makefile tho.

Bryce-Soghigian:
👀

e2e/scenario_test.go
timmy-wright: Have removed this skip and added a way for tests to specify which bootstrapping method they want.

self-contained/bootstrap_install.sh
```diff
@@ -60,6 +60,9 @@ installContainerRuntime() {
 }
 
 installNetworkPlugin() {
+    if [[ "${NETWORK_PLUGIN}" = "none" ]]; then
+        return
+    fi
```
r2k1, line 65: @timmy-wright did you change it intentionally?

pkg/agent/bakerapi.go
```diff
@@ -85,6 +87,32 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context, conf
...
+	config.Version = "v0"
+	nodeBootstrapping, err := agentBaker.GetNodeBootstrapping(ctx, config)
+	if err != nil {
+		return nil, err
+	}
+	configJSON, err := json.Marshal(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal nbc, error: %w", err)
+	}
+	//nolint: lll
```
djsly, line 107: what is `lll`

node-bootstrapper/custom_data_test.go
```diff
@@ -0,0 +1,71 @@
...
+		actual := getFile(t, nbc, "/etc/kubernetes/certs/ca.crt", 0600)
+		expected := "test-ca-cert"
+		assert.Equal(t, expected, actual)
+	})
+
+	t.Run("exec_start.conf", func(t *testing.T) {
+		nbc := validNBC()
+		actual := getFile(t, nbc, "/etc/systemd/system/docker.service.d/exec_start.conf", 0644)
+		nbc.ContainerService.Properties.OrchestratorProfile.KubernetesConfig.DockerBridgeSubnet = "1.1.1.1"
+		expected := `[Service]
```
djsly, line 22: where is the expected custom data coming from ?  

do we need to update this testdata when we change the content of the files ?

node-bootstrapper/main.go
```diff
@@ -0,0 +1,396 @@
...
+		return fmt.Errorf("content bootstrap kubeconfig: %w", err)
+	}
+	files["/var/lib/kubelet/bootstrap-kubeconfig"] = File{
+		Content: bootstrapKubeconfig,
+		Mode:    ReadOnlyWorld,
+	}
+	return nil
+}
+
+func contentKubeconfig(config *datamodel.NodeBootstrappingConfiguration) string {
```
djsly, line 236: could we renamed all function starting with `content<artifact>` with maybe something like `generate<artifact>Content` 

When reading the function name I wasn't sure what was happening.

node-bootstrapper/main.go
```diff
@@ -0,0 +1,396 @@
...
+	users := `- name: client
+  user:
+    client-certificate: /etc/kubernetes/certs/client.crt
+    client-key: /etc/kubernetes/certs/client.key`
+
+	return fmt.Sprintf(`
```
djsly, line 242: I feel generating inline string with explicit format is asking for trouble. could we look instead at migrating those template string outside of go and load template files instead ?

node-bootstrapper/main.go
```diff
@@ -0,0 +1,396 @@
...
+- context:
+    cluster: localcluster
+    user: client
+  name: localclustercontext
+current-context: localclustercontext
+`, agent.GetKubernetesEndpoint(config.ContainerService), users)
+}
+
+func contentBootstrapKubeconfig(config *datamodel.NodeBootstrappingConfiguration) (string, error) {
+	data := map[string]any{
```
djsly, line 262: for example, the generated string is really hard to grasp from the code. Having an external template file would help quickly identify what the final file content looks like (formatting too) and which parameters are coming from outside.

node-bootstrapper/build.sh
djsly: agree as well.  let align on build process

e2e/validation.go
```diff
@@ -110,7 +110,7 @@ func commonLiveVMValidators(opts *scenarioRunOpts) []*LiveVMValidator {
 				"cluster-provision-cse-output.log",
 				"cloud-init-files.paved",
 				"vhd-install.complete",
-				"cloud-config.txt",
+				//"cloud-config.txt", // file with UserData
```
djsly, line 113: left over TODO ? no sure why this needs to be removed

e2e/scenario_test.go
```diff
@@ -559,6 +559,20 @@ func Test_ubuntu1804ChronyRestarts(t *testing.T) {
 	})
 }
 
+func Test_ubuntu2204SelfContainedInstaller(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "tests that a new ubuntu 2204 node using self contained installer can be properly bootstrapped",
+		Config: Config{
+			NodeBootstrappingType: SelfContained,
```
djsly, line 566: do we want to use that type `selfcontained` ?  I know we moved to calling this `scriptless` movgin forward. 

What are other expected types ?

e2e/nodebootstrapping.go
```diff
@@ -47,16 +47,20 @@ func getNodeBootstrappingForCoverage(nbc *datamodel.NodeBootstrappingConfigurati
...
+func getNodeBootstrappingForValidation(ctx context.Context, nbc *datamodel.NodeBootstrappingConfiguration, bootstrappingType NodeBootstrappingType) (*datamodel.NodeBootstrapping, error) {
 	ab, err := agent.NewAgentBaker()
 	if err != nil {
 		return nil, err
 	}
-	nodeBootstrapping, err := ab.GetNodeBootstrapping(ctx, nbc)
-	if err != nil {
-		return nil, err
+	switch {
+	case bootstrappingType == SelfContained:
```
djsly, line 56: should we renamed this to scriptless ?

e2e/nodebootstrapping.go
```diff
@@ -47,16 +47,20 @@ func getNodeBootstrappingForCoverage(nbc *datamodel.NodeBootstrappingConfigurati
...
+		return ab.GetNodeBootstrappingForSelfContained(ctx, nbc)
+	case bootstrappingType == CustomScripts:
```
djsly, line 58: what is customScript ?

e2e/installer_test.go
```diff
@@ -0,0 +1,91 @@
...
+		Description: "Tests that a node using the Ubuntu 2204 VHD can be properly bootstrapped",
+		Config: Config{
+			//NodeBootstrappingType: SelfContained,
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDUbuntu2204Gen2Containerd,
+			LiveVMValidators: []*LiveVMValidator{
+				containerdVersionValidator("1.7.20-1"),
+				runcVersionValidator("1.1.12-1"),
+			},
+			CSEOverride:       CSENodeBootstrapper(context.TODO(), t, cluster),
```
djsly, line 39: what is context.TODO() ?

e2e/installer_test.go
```diff
@@ -0,0 +1,91 @@
...
+	configJSON, err := json.Marshal(cluster.NodeBootstrappingConfiguration)
+	require.NoError(t, err)
+
+	binary := compileNodeBootstrapper(t)
+	url, err := config.Azure.UploadAndGetLink(ctx, "node-bootstrapper-"+hashFile(t, binary.Name()), binary)
+	require.NoError(t, err)
+	return fmt.Sprintf(`bash -c "(echo '%s' | base64 -d > config.json && curl -L -o ./node-bootstrapper '%s' && chmod +x ./node-bootstrapper && mkdir -p /var/log/azure && ./node-bootstrapper provision --provision-config=config.json) > /var/log/azure/node-bootstrapper.log 2>&1"`, base64.StdEncoding.EncodeToString(configJSON), url)
+}
+
+func compileNodeBootstrapper(t *testing.T) *os.File {
```
djsly, line 55: are we recompiling the binary with each test ? should look at making this more efficient

e2e/config/config.go
```diff
@@ -32,6 +32,9 @@ type Configuration struct {
 	IgnoreScenariosWithMissingVHD bool          `env:"IGNORE_SCENARIOS_WITH_MISSING_VHD"`
 	SkipTestsWithSKUCapacityIssue bool          `env:"SKIP_TESTS_WITH_SKU_CAPACITY_ISSUE"`
 	KeepVMSS                      bool          `env:"KEEP_VMSS"`
+	BlobStorageAccount            string        `env:"BLOB_STORAGE_ACCOUNT" envDefault:"https://abe2e.blob.core.windows.net"`
```
djsly, line 35: is this a new blob storage ? how was this created ? do we have IaC for it? if this is purely to host the binary for the test, I'm wondering if we could avoid that, knowing that the more azure resources we have to more SFI work we have to do.

e2e/config/azure.go
```diff
@@ -126,37 +134,74 @@ func NewAzureClient(subscription string) (*AzureClient, error) {
 
 	cloud.VMSS, err = armcompute.NewVirtualMachineScaleSetsClient(subscription, credential, opts)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create vmss client: %w", err)
```
djsly, line 129: nit: fail/failed are keywords that are normally used when grepping for what went wrong. I"m not sure how practical it is to remove them here, I would personally keep them

.golangci.yaml
```diff
@@ -148,7 +148,7 @@ linters-settings:
       shadow:
         # Whether to be strict about shadowing; can be noisy.
         # Default: false
-        strict: true
+        strict: false
```
djsly, line 151: why was this changed ?

pkg/agent/bakerapi.go
r2k1: The line with bash script is too long, but I don't want to split it across multiple lines.

pkg/agent/baker.go
```diff
@@ -1001,6 +979,44 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
 
+func GetDataDir(config *datamodel.NodeBootstrappingConfiguration) string {
+	cs := config.ContainerService
+	profile := config.AgentPoolProfile
+	if profile != nil && profile.KubernetesConfig != nil &&
+		profile.KubernetesConfig.ContainerRuntimeConfig != nil &&
+		profile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey] != "" {
+		return profile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey]
+	}
+	if profile.KubeletDiskType == datamodel.TempDisk {
```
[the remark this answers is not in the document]
r2k1: I didn't change the code, just extracted it to an exported function I can use from node-bootstrapper.

It will be a panic, but I wanted to minimize any changes outside of node-boostrapper.

e2e/installer_test.go
r2k1: Good point, I'll take a look.

Initially, there's been a single test. Definitely should re-use the binary.

e2e/config/config.go
r2k1: It's a hack to speed e2e feedback loop.

Without it we need to rebuild the VHD before running any tests. It adds about an hour to the loop.
This code put binary on a node and run a similar CSE command to what we're going to use.

I had difficulty finding a way to upload a binary to a VM. It's not available from a public network. So I couldn't SSH into it. And it's not joined k8s cluster yet when we need to execute it.

I think I created it manually (it's been a while ago).

I don't really like and happy to change if you have any other ideas.

e2e/validation.go
r2k1: This file doesn't exist for scriptless approach. Because we don't supply UserData at all. (It mostly contains bash script).

node-bootstrapper/main.go
r2k1: I hate YAML-generation in templates. Especially when IF conditions are involved.
It gives me HELM vibes with hacks around proper indentation. Tabs vs spaces etc.

In this function I intentionally avoided using templates due to its complexity.

node-bootstrapper/custom_data_test.go
r2k1: There is no "source" for the data. Mainly a reflection of what's currently happening.

Yes, it needs to be updated when files are changed.

node-bootstrapper/main.go
r2k1: I want to delay the decision on how to handle it properly.

Currently there only a couple of small files and I think it's easier to do this way without an overhead of templating.

But I agree, it's probably not sustainable if we're going to move all files we generate. I'm just not sure yet how to approach it.

.golangci.yaml
r2k1: The linter is being really noisy for a typical go error handling.
I think shadowing errors is fine, we shouldn't come up with err2 and err3 variables.

e2e/installer_test.go
r2k1: Similar to `context.Background`, a context that never gets cancelled.
I replaced to a one with an expiration time.

e2e/config/azure.go
r2k1: I avoid including `error` or `failed` in the error message. It just produces a lot of duplication and noise when you wrap errors multiple times:

```
ERROR: error create client: error create VMSS client: error create something: error do something
```
vs

```
ERROR: create client: create VMSS client: create something: do something
```

I don't think it's hard to search them, but it improves the message itself when you surface it (sometimes accidental)

r2k1: It also helps if you ever want to apply a fancy formatting to an error. I think it's a common recommendation to omit the keywords.

e2e/scenario_test.go
r2k1:
```
type NodeBootstrappingType string

const (
	CustomScripts NodeBootstrappingType = "CustomScripts"
	Scriptless    NodeBootstrappingType = "Scriptless"
)
```

e2e/installer_test.go
r2k1: Actually, currently there is a single test only that's using it. Other tests are using binary from the VHD.

I'll optimize it once a usage pattern is more established.

timmy-wright (approved):

self-contained/bootstrap_install.sh
timmy-wright: Yeah, this was going to be part of my next PR. We can either remove it from this one or keep it.

timmy-wright (approved):

pkg/agent/baker.go
```diff
@@ -1001,6 +979,44 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+		return true
+	}
+	return cs.Properties.OrchestratorProfile.KubernetesConfig.ContainerRuntimeConfig != nil &&
+		cs.Properties.OrchestratorProfile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey] != ""
+}
+
+func GetKubernetesEndpoint(cs *datamodel.ContainerService) string {
+	if cs.Properties.HostedMasterProfile == nil {
```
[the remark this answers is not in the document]
timmy-wright: This is existing code that was extracted to make it reusable and testable.

pkg/agent/baker.go
```diff
@@ -1001,6 +979,44 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+}
+
+func HasDataDir(config *datamodel.NodeBootstrappingConfiguration) bool {
+	cs := config.ContainerService
+	profile := config.AgentPoolProfile
+	if profile != nil && profile.KubernetesConfig != nil && profile.KubernetesConfig.ContainerRuntimeConfig != nil &&
+		profile.KubernetesConfig.ContainerRuntimeConfig[datamodel.ContainerDataDirKey] != "" {
+		return true
+	}
+	if profile.KubeletDiskType == datamodel.TempDisk {
```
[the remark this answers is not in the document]
timmy-wright: This is existing code that was extracted to make it reusable and testable.

djsly (approved):

e2e/config/azure.go
djsly: makes sense if the logger prints the level, this is indeed redundant.

PR #4892, opened 2024-09-02
Refactor to move GPU driver to config

ganeshkumarashok:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


nbcparser/pkg/parser/helper.go
```diff
@@ -546,7 +546,7 @@ func getGpuImageSha(vmSize string) string {
 }
 
 func getGpuDriverVersion(vmSize string) string {
-	return common.GetGPUDriverVersion(vmSize)
+	return common.GetNewGPUDriverVersion(vmSize)
```
cameronmeissner, line 549: could we just call `common.GetNewGPUDriverVersion(vmSize)` directly instead of defining this `getGPUDriverVersion` helper? seems a bit redundant to me

pkg/agent/common/gpu_drivers.json
```diff
@@ -0,0 +1,4 @@
+{
+    "cuda": "cuda-550.90.07",
```
cameronmeissner, line 2: just food for thought: what if we had this json and a support go module within the aks-gpu repo itself that we'd vendor into agentbaker code? that way when making updates we'd just build a new container image and bump the version within the aks-gpu repo, and then simply bump the go mod dependency within agentbaker

ganeshkumarashok:
I moved it to config in a separate PR

PR #4894, opened 2024-09-03
chore: add comment for NIC for CNI

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -327,16 +327,6 @@ setupCNIDirs() {
 # The version used to be deteremined by RP/toggle but are now just hadcoded in vhd as they rarely change and require a node image upgrade anyways
 # Latest VHD should have the untar, older should have the tgz. And who knows will have neither. 
 installCNI() {
-
-    if [ ! -f "$COMPONENTS_FILEPATH" ] || ! jq '.Packages[] | select(.name == "cni-plugins")' < $COMPONENTS_FILEPATH > /dev/null; then
```
cameronmeissner, line 331: we recently added this to make the components.json refactor backwards-compatible for older VHDs that don't contain the new changes, specifically for CNI

how will we handle this going forward?

AlisonB319: My assumption would be that older VHDs won't be allowed to utilize bring-your-own ACRs. I'll need to confirm this, and if so, I think adding a comment will suffice.

fseldow: we do plan to add validator to avoid old vhds and cse. thx for the comment

fseldow (approved):

cameronmeissner (approved):

PR #4895, opened 2024-09-03
chore: bump cni cns versions from v1.5.32 to v1.5.35 for release/v1.5

kmurudi:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/components.json
+++ b/parts/linux/cloud-init/artifacts/components.json
@@ -42,13 +42,13 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cni:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.5.32",
+        "v1.5.35",
         "v1.6.0",
         "v1.6.3"
       ],
       "prefetchOptimizations": [
         {
-          "version": "v1.5.32",
+          "version": "v1.5.35",
           "binaries": [
             "dropgz"
           ]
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -42,13 +42,13 @@
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-cni:*",
       "amd64OnlyVersions": [],
       "multiArchVersions": [
-        "v1.5.32",
```
rbtr, line 45: we still need 1.5.32, it is what's currently used in AKS prod

rbtr (changes requested):

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -72,13 +72,13 @@
       "amd64OnlyVersions": [],
       "multiArchVersions": [
         "v1.4.52",
-        "v1.5.32",
+        "v1.5.35",
         "v1.6.0",
         "v1.6.3"
```
rbtr, line 77: one of these can probably be removed though

parts/linux/cloud-init/artifacts/components.json
kmurudi: sure. @rbtr can we add 1.6.5 in the same PR? the tag is going to be released today & we were wondering if that could have any issues to add both versions for 1.5 & 1.6 in the same PR

rbtr: you need to release it first, but otherwise no issue here. we just need to keep N and N+1 version for each release train

rbtr (approved):

miguelgoms (approved):

spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh
miguelgoms, line 1: looks like we need to update the case on line 85 on cse_install_spec as well.

djsly (approved):

cameronmeissner (approved):

PR #4899, opened 2024-09-04
feat: use oras for install credential provider

AlisonB319:
**What type of PR is this?** When a registry url is provided, we want to install credential provider through oras if it is already not cached.

The only time oras download of credential provider will be hit logically, is when 
`if [[ $KUBELET_FLAGS == *"image-credential-provider-config"* && $KUBELET_FLAGS == *"image-credential-provider-bin-dir"* ]]; then` inside `cse_config.sh`

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -486,7 +510,7 @@ installKubeletKubectlAndKubeProxy() {
             if [[ ! -z ${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER} ]]; then
                 # network isolated cluster
                 echo "Detect Bootstrap profile artifact is Cache, will use oras to pull artifact binary"
-                registry_url="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/${K8S_REGISTRY_REPO}:v${KUBERNETES_VERSION}-linux-${CPU_ARCH}"
+                registry_url="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/${K8S_REGISTRY_REPO}kubernetes-node:v${KUBERNETES_VERSION}-linux-${CPU_ARCH}"
```
fseldow, line 513: what about
`${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/${K8S_REGISTRY_REPO}/kubernetes-node`
and change K8S_REGISTRY_REPO from `"oss/binaries/kubernetes/"` to `"oss/binaries/kubernetes"` ?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -140,15 +140,40 @@ installNetworkPlugin() {
...
+        CREDENTIAL_PROVIDER_VERSION=$(echo "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" | grep -oP 'v\d+(\.\d+)*' | sed 's/^v//' | head -n 1)
+    else
+        CREDENTIAL_PROVIDER_DOWNLOAD_URL="https://acs-mirror.azureedge.net/cloud-provider-azure/v${CREDENTIAL_PROVIDER_VERSION}/binaries/azure-acr-credential-provider-linux-${CPU_ARCH}-v${CREDENTIAL_PROVIDER_VERSION}.tar.gz"
+    fi
+
+    if [[ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ]]; then
+        # if there is a container registry then oras is needed to download
+        CREDENTIAL_PROVIDER_DOWNLOAD_URL="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/${K8S_REGISTRY_REPO}azure-acr-credential-provider:v${CREDENTIAL_PROVIDER_VERSION}-linux-${CPU_ARCH}"
+        CREDENTIAL_PROVIDER_TGZ_TMP="${CREDENTIAL_PROVIDER_DOWNLOAD_URL##*/}"
+        if isRegistryUrl "${CREDENTIAL_PROVIDER_DOWNLOAD_URL}"; then
```
fseldow, line 161: the download url is already redefine in line 159 and it should already follow registry regex. Maybe no need isRegistryUrl here?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -140,15 +140,40 @@ installNetworkPlugin() {
...
+        # The version in the URL is unknown. An acs-mirror or registry URL could be passed meaning the version must be extracted from the URL. 
+        CREDENTIAL_PROVIDER_VERSION=$(echo "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" | grep -oP 'v\d+(\.\d+)*' | sed 's/^v//' | head -n 1)
+    else
+        CREDENTIAL_PROVIDER_DOWNLOAD_URL="https://acs-mirror.azureedge.net/cloud-provider-azure/v${CREDENTIAL_PROVIDER_VERSION}/binaries/azure-acr-credential-provider-linux-${CPU_ARCH}-v${CREDENTIAL_PROVIDER_VERSION}.tar.gz"
+    fi
+
+    if [[ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ]]; then
+        # if there is a container registry then oras is needed to download
+        CREDENTIAL_PROVIDER_DOWNLOAD_URL="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/${K8S_REGISTRY_REPO}azure-acr-credential-provider:v${CREDENTIAL_PROVIDER_VERSION}-linux-${CPU_ARCH}"
+        CREDENTIAL_PROVIDER_TGZ_TMP="${CREDENTIAL_PROVIDER_DOWNLOAD_URL##*/}"
```
fseldow, line 160: the true credential_provider file name follows the original download url. `azure-acr-credential-provider-linux-amd64-v1.29.6.tar.gz`
In previous line it is override by registry url, it will cause the TGZ_TMP be `azure-acr-credential-provider:v1.29.6-linux-amd64`

So that i suggest not override GLOBAL variable `CREDENTIAL_PROVIDER_DOWNLOAD_URL`, instead define one new local variable

parts/linux/cloud-init/artifacts/cse_install.sh
AlisonB319: good point

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -140,15 +140,35 @@ installNetworkPlugin() {
...
+        # The version in the URL is unknown. An acs-mirror or registry URL could be passed meaning the version must be extracted from the URL. 
+        CREDENTIAL_PROVIDER_VERSION=$(echo "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" | grep -oP 'v\d+(\.\d+)*' | sed 's/^v//' | head -n 1)
+    else
+        CREDENTIAL_PROVIDER_DOWNLOAD_URL="https://acs-mirror.azureedge.net/cloud-provider-azure/v${CREDENTIAL_PROVIDER_VERSION}/binaries/azure-acr-credential-provider-linux-${CPU_ARCH}-v${CREDENTIAL_PROVIDER_VERSION}.tar.gz"
+    fi
+
+    # if there is a container registry then oras is needed to download
+    BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER:=}"
+    if [[ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ]]; then
+        local credential_provider_download_url_for_oras="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/${K8S_REGISTRY_REPO}/azure-acr-credential-provider:v${CREDENTIAL_PROVIDER_VERSION}-linux-${CPU_ARCH}"
```
djsly, line 160: does `BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER` contain the acr cache name ?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -429,19 +449,18 @@ extractKubeBinaries() {
...
         # if the url is a registry url, use oras to pull the artifact instead of curl
-        registry_regex='^.+\/.+\/.+:.+$'
-        if [[ ${kube_binary_url} =~ $registry_regex ]]; then
+        if isRegistryUrl "${kube_binary_url}"; then
             echo "detect kube_binary_url, ${kube_binary_url}, as registry url, will use oras to pull artifact binary"
             # download the kube package from registry as oras artifact
             k8s_tgz_tmp="${k8s_downloads_dir}/kubernetes-node-linux-${CPU_ARCH}.tar.gz"
             retrycmd_get_tarball_from_registry_with_oras 120 5 "${k8s_tgz_tmp}" ${kube_binary_url} || exit $ERR_ORAS_PULL_K8S_FAIL
-            if [[ ! -f ${k8s_tgz_tmp} ]]; then
+             if [[ ! -f "${k8s_tgz_tmp}" ]]; then
```
djsly, line 457: bad indent ?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -429,19 +449,18 @@ extractKubeBinaries() {
...
                 exit "$ERR_ORAS_PULL_K8S_FAIL"
             fi
         else
             # download the kube package from the default URL
             retrycmd_get_tarball 120 5 "${k8s_tgz_tmp}" ${kube_binary_url} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
-            if [[ ! -f ${k8s_tgz_tmp} ]]; then
+             if [[ ! -f "${k8s_tgz_tmp}" ]]; then
```
djsly, line 463: bad indent ?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -386,6 +386,16 @@ isARM64() {
     fi
 }
 
+isRegistryUrl() {
+    local binary_url=$1
+    registry_regex='^.+\/.+\/.+:.+$'
```
djsly, line 391: add omment which explains what this regex is trying to do. with an example of a valid string.

parts/linux/cloud-init/artifacts/cse_install.sh
fseldow: no it does not, it is just the acr url. Currently, we have changed to no cache name for final registry url. The final url should be `xxx.acr.io/oss/binary/xxxx:v1.29.7-linux-amd64`

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -140,15 +140,35 @@ installNetworkPlugin() {
...
-    CREDENTIAL_PROVIDER_TGZ_TMP=${CREDENTIAL_PROVIDER_DOWNLOAD_URL##*/} # Use bash builtin ## to remove all chars ("*") up to the final "/"
-    retrycmd_get_tarball 120 5 "$CREDENTIAL_PROVIDER_DOWNLOAD_DIR/$CREDENTIAL_PROVIDER_TGZ_TMP" "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
+
+    CREDENTIAL_PROVIDER_DOWNLOAD_URL="${CREDENTIAL_PROVIDER_DOWNLOAD_URL:=}"
+    if [[ -n "${CREDENTIAL_PROVIDER_DOWNLOAD_URL}" ]]; then
+        # CREDENTIAL_PROVIDER_DOWNLOAD_URL is set by linuxCredentialProviderURL
+        # The version in the URL is unknown. An acs-mirror or registry URL could be passed meaning the version must be extracted from the URL. 
+        CREDENTIAL_PROVIDER_VERSION=$(echo "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" | grep -oP 'v\d+(\.\d+)*' | sed 's/^v//' | head -n 1)
+    else
+        CREDENTIAL_PROVIDER_DOWNLOAD_URL="https://acs-mirror.azureedge.net/cloud-provider-azure/v${CREDENTIAL_PROVIDER_VERSION}/binaries/azure-acr-credential-provider-linux-${CPU_ARCH}-v${CREDENTIAL_PROVIDER_VERSION}.tar.gz"
```
fseldow, line 154: not sure why we need to redefine DOWNLOAD_URL here. Is it trying to make sure the default value?
If so seems still possible miss the variable assignment of CREDENTIAL_PROVIDER_VERSION

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -386,6 +386,16 @@ isARM64() {
...
+    if [[ ${binary_url} =~ $registry_regex ]]; then
+        return 0 # true
+    else
```
cameronmeissner, line 394: nit: redundant else

parts/linux/cloud-init/artifacts/cse_install.sh
AlisonB319: This is because in the if statement `CREDENTIAL_PROVIDER_DOWNLOAD_URL` is set by `linuxCredentialProviderURL` and I want to make sure I don't override that. But, there is a chance that CREDENTIAL_PROVIDER_DOWNLOAD_URL` is set by `linuxCredentialProviderURL` and a container registry is provided, in that case I want to take the version that is in the URL string, and use it in the string registry URL

Otherwise, I changed how the function is called, `CREDENTIAL_PROVIDER_DOWNLOAD_URL` is no longer passed in so it needs to be defaulted in the else.

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: @fseldow could you give some insights on the regex here since I just moved it into a helper function.

djsly (approved):

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -140,15 +140,35 @@ installNetworkPlugin() {
     rm -rf $CNI_DOWNLOADS_DIR & 
 }
 
-
-downloadCredentalProvider() {
+# downloadCredentialProvider is most-likely called during build time by install-dependencies.sh.
```
cameronmeissner, line 143: I think it's always called in install-dependencies.sh right? maybe we can clarify that here

cameronmeissner (approved):

PR #4907, opened 2024-09-05
fix: build CVM image on CVM hardware

zachary-bailey:
**What type of PR is this?**

/kind fix

This PR directs the build pipeline to build CVM images on dedicated CVM hardware. This allows apt updates for the CVM image during the build process. Original PR from @anujmaheshwari1 enabling this can be found here: [fix: install upgrades for CVM images in pre-install dependencies](https://github.com/Azure/AgentBaker/pull/4728). This PR also adds the changes originally made in that PR.

CVM breaks on kernel image updates due to null-boot package post-install.

**Notes**:

`NEVER MERGE` label is assigned to this PR until necessary changes in ACG are made to support the new workflow it requires. These changes should be simple. 

If merged, these changes would require the availability and quota of Confidential Compute VMs in the build region, which is why this adds a conditional region check in `.pipelines/.vsts-vhd-builder-release.yaml` for CVMs. This will prevent failing pipelines due to not having Confidential Compute cores available.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


vhdbuilder/packer/vhd-image-builder-cvm.json
```diff
@@ -0,0 +1,669 @@
...
+        "image_sku": "{{user `img_sku`}}",
+        "branch": "{{user `branch`}}"
+      },
+      "location": "{{user `location`}}",
+      "vm_size": "{{user `vm_size`}}",
+      "polling_duration_timeout": "1h",
+      "managed_image_storage_account_type": "Premium_LRS",
+      "secure_boot_enabled": true,
+      "vtpm_enabled": true,
+      "security_type": "ConfidentialVM",
```
cameronmeissner, line 65: are these fields the only diff with the existing base template? if so we might be able to reuse the base template and pass these values as variables initialized within init-variables.sh

vhdbuilder/packer/init-variables.sh
```diff
@@ -157,6 +157,8 @@ if [[ "${MODE}" == "linuxVhdMode" ]]; then
 		elif [[ "${IMG_OFFER,,}" == "azure-linux-3" ]]; then
 			# for Azure Linux 3.0, only use AzureLinux prefix
 			SIG_IMAGE_NAME="AzureLinux${SIG_IMAGE_NAME}"
+		elif [[ ${IMG_SKU} == "20_04-lts-cvm" ]]; then
+			SIG_IMAGE_NAME+="TESTDEF"
```
cameronmeissner, line 161: don't we already add CVM to the name of the corresponding image definition?

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1195,7 +1195,7 @@ stages:
               overrideBranch: ${{ parameters.overrideBranch }}
               artifactName: 2404-gen2-containerd
       - job: build2004cvmgen2containerd
-        condition: eq('${{ parameters.build2004cvmgen2containerd }}', true)
+        condition: and(eq(variables.PACKER_BUILD_LOCATION, 'eastus'), eq('${{ parameters.build2004cvmgen2containerd }}', true))
```
cameronmeissner, line 1198: can we add a comment here explaining why we only build CVM in eastus

vhdbuilder/packer/test/run-test.sh
```diff
@@ -99,6 +99,11 @@ else
     TARGET_COMMAND_STRING+="--security-type TrustedLaunch --enable-secure-boot true --enable-vtpm true"
   fi
 
+  if [[ "${OS_TYPE}" == "Linux" && "${IMG_SKU}" == "20_04-lts-cvm" ]]; then
+    TARGET_COMMAND_STRING="--size Standard_EC8ads_v5"
+    TARGET_COMMAND_STRING+=" --security-type ConfidentialVM --enable-secure-boot true --enable-vtpm true --os-disk-security-encryption-type VMGuestStateOnly --specialized"
```
AlisonB319, line 104: is there a reason we son't make this one line since we are just adding strings?

vhdbuilder/packer/init-variables.sh
zachary-bailey: This is one of the changes that will need to be made once the PR is ready for merge. We have to upload the image to a specialized image definition. Was hoping to discuss during standup/parking lot. Adding TESTDEF right now is just how I'm getting it there, given that SIG_IMAGE_NAME, which is set to SKU_NAME, points to the definition that we currently have.

vhdbuilder/packer/vhd-image-builder-cvm.json
zachary-bailey: These fields and the two in the Shared Image Gallery Destination block. Additionally, the CVM template does not have the managed-image name/RG variables, as the CVM can't have a managed-image source.

vhdbuilder/packer/test/run-test.sh
zachary-bailey: Haha good catch thank you!

cameronmeissner:
it sounds like we decided to set the staging gallery CVM definition to specialized right?

zachary-bailey:
> it sounds like we decided to set the staging gallery CVM definition to specialized right?

Yes. Planning on doing this programmatically, I put the code necessary to do so in init-variables.sh. So the first run, it should look for the def and since its not there, create it. Then upload fine. Anuj asked to review that change before I run it the first time in the test pipeline so just waiting on that. @cameronmeissner

PR #4911, opened 2024-09-06
feat: respect nodepool tags for disabling kubelet serving certificate…

cameronmeissner:
… rotation at runtime

**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -291,6 +293,8 @@ try
     icacls.exe "c:\k"
     Get-ProvisioningScripts
     Get-LogCollectionScripts
+
+    Disable-KubeletServingCertificateRotationForTags
```
cameronmeissner, line 297: this is just for testing, @AbelHu we'll need to remove/comment this out until the script package changes roll out right?

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,70 @@ function Get-KubePackage {
...
+function Get-TagValue {
+    Param(
+        [Parameter(Mandatory=$true)][string]
+        $TagName,
+        [Parameter(Mandatory=$true)][string]
+        $DefaultValue
+    )
+
+    $uri = "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
+    $response = Retry-Command -Command "Invoke-RestMethod" -Args @{Uri=$uri; Method="Get"; ContentType="application/json"; Headers=@{"Metadata"="true"}} -Retries 3 -RetryDelaySeconds 5
```
AbelHu, line 244: We may need to add error code if all retires fail

parts/windows/kuberneteswindowssetup.ps1
AbelHu, line 297: This will fail since Disable-KubeletServingCertificateRotationForTags cannot be found in the current windows cse script package.

1. Update the scripts under staging/cse/windows
2. Release a new cse script package
3. Update the cse package version and add this line

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,70 @@ function Get-KubePackage {
     Remove-Item $zipfile
 }
```
AbelHu, line 211: It is better to add ut to cover these new lines

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,70 @@ function Get-KubePackage {
...
+    $tagName = "aks-disable-kubelet-serving-certificate-rotation"
+    $disabled = Get-TagValue -TagName $tagName -DefaultValue "false"
+    if (!($disabled -eq "true")) {
+        Write-Log "Nodepool tag `"$tagName`" is missing or not set to true, nothing to disable"
+        return
+    }
+
+    Write-Log "Kubelet serving certificate rotation is disabled by nodepool tags, will reconfigure kubelet flags and node labels"
+
+    $global:KubeletConfigArgs = $global:KubeletConfigArgs -replace "--rotate-server-certificates=true", "--rotate-server-certificates=false"
```
AbelHu, line 272: Should we set it to false whem current args do not contain `--rotate-server-certificates`?

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,70 @@ function Get-KubePackage {
...
+    $global:KubeletNodeLabels = Remove-KubeletNodeLabel -KubeletNodeLabels $global:KubeletNodeLabels -Label "kubernetes.azure.com/kubelet-serving-ca=cluster"
```
AbelHu, line 273: This line introduces a dependency that it must be called before `Write-KubeClusterConfig`. It is better to add comment in the code to call this function to higlight it.

staging/cse/windows/kubeletfunc.ps1
cameronmeissner: I was thinking we shouldn't set it to anything if the flag isn't already there (which in this case is fine since it defaults to false anyways)

that way we won't accidentally give this flag to a kubelet version that doesn't support it

staging/cse/windows/kubeletfunc.ps1
cameronmeissner: right now we just default to returning false if all retries fail which I personally think is fine, though I can add an error code if you think that would be better

cameronmeissner: I added a try/catch, where we just return the default value if an error occurs - let me know what you think

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -429,6 +432,17 @@ should_skip_nvidia_drivers() {
     echo "$should_skip"
 }
 
+should_disable_kubelet_serving_certificate_rotation() {
```
djsly, line 435: function name format doesn't match the below functionName. should it be 

`shouldDisableKubeletServingCertificateRotation()` instead ? of the other one is wrong ?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -429,6 +432,17 @@ should_skip_nvidia_drivers() {
...
+    set -x
+    body=$(curl -fsSL -H "Metadata: true" --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01")
```
djsly, line 437: don't we have a helper function to query the metadata service ?

djsly (approved):

e2e/scenario_test.go
```diff
@@ -963,3 +963,73 @@ func Test_ubuntu2204Wasm(t *testing.T) {
...
+				nbc.KubeletConfig["--rotate-server-certificates"] = "true"
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				if vmss.Tags == nil {
+					vmss.Tags = map[string]*string{}
+				}
+				vmss.Tags["aks-disable-kubelet-serving-certificate-rotation"] = to.Ptr("true")
+			},
+			LiveVMValidators: []*LiveVMValidator{
+				FileExcludesContentsValidator("/etc/default/kubelet", "\\-\\-rotate-server-certificates=true", "\\-\\-rotate-server-certificates=true"),
```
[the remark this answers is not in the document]
cameronmeissner: the last argument is the "name" of the content used for debug messages, thus I just repeated it for clarity

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: let me double check, though I don't think we do

cameronmeissner: I decided to copy the format used for the skip gpu driver function, I think we should probably align on a style guide eventually...

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: I decided to copy the format used for the skip gpu driver function, I think we should probably align on a style guide eventually...

staging/cse/windows/kubeletfunc.ps1
cameronmeissner: added

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: double checked we don't, might be worth a follow-up PR to add a helper

jason1028kr (approved):

e2e/scenario_test.go
```diff
@@ -984,6 +984,74 @@ func Test_ubuntu2204Wasm(t *testing.T) {
 	})
 }
 
+func Test_Ubuntu2204DisableKubeletServingCertificateRotationWithTags(t *testing.T) {
```
anujmaheshwari1, line 987: do we want to add a test for when the flag isnt set but the nodepool tag is passed and verify the behavior? to account for any funky commands a cx may pass.

anujmaheshwari1 (approved):

staging/cse/windows/kubeletfunc.ps1
cameronmeissner: in the end I went with a new exit code to align with the linux implementation

e2e/scenario_test.go
cameronmeissner: added

parts/windows/windowscsehelper.ps1
```diff
@@ -72,6 +72,7 @@ $global:WINDOWS_CSE_ERROR_INSTALL_CREDENTIAL_PROVIDER = 65 # exit code for insta
 $global:WINDOWS_CSE_ERROR_DOWNLOAD_CREDEDNTIAL_PROVIDER=66 # exit code for downloading credential provider failure
 $global:WINDOWS_CSE_ERROR_CREDENTIAL_PROVIDER_CONFIG=67 # exit code for checking credential provider config failure
 $global:WINDOWS_CSE_ERROR_ADJUST_PAGEFILE_SIZE=68
+$global:WINDOWS_CSE_ERROR_LOOKUP_INSTANCE_DATA_TAG=69 # exit code for looking up nodepool/VM tags via IMDS
```
AbelHu, line 75: You also need to add the error code name in `$global:ErrorCodeNames`

AbelHu: I updated the UT in https://github.com/Azure/AgentBaker/pull/4924

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,67 @@ function Get-KubePackage {
...
+        return $DefaultValue
+    }
+    return $tag.value
+}
+
+# Note: this function modifies global kubelet config args and node labels. Thus, it MUST
+# be called before Write-KubeClusterConfig, and any other function that relies on the values of
+# kubelet config args and node labels.
+function Disable-KubeletServingCertificateRotationForTags {
+    Logs-To-Event -TaskName "AKS.WindowsCSE.DisableKubeletServingCertificateRotationForTags" -TaskMessage "Check whether to disable kubelet serving certificate rotation via nodepool tags."
```
AbelHu, line 251: It is better to add `$global:EnableKubeletServingCertificateRotation` in the log event.

staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,67 @@ function Get-KubePackage {
...
+    Write-Log "Checking whether to disable kubelet serving certificate rotation for nodepool tags"
+
+    if (!($global:EnableKubeletServingCertificateRotation)) {
+        Write-Log "Kubelet serving certificate rotation is already disabled"
+        return
+    }
+
+    $tagName = "aks-disable-kubelet-serving-certificate-rotation"
+    $disabled = Get-TagValue -TagName $tagName -DefaultValue "false"
+    if (!($disabled -eq "true")) {
```
AbelHu, line 262: `$disabled -ne "true"`?

staging/cse/windows/kubeletfunc.tests.ps1
```diff
@@ -62,4 +62,133 @@ Describe 'Get-KubePackage' {
             Assert-MockCalled -CommandName 'DownloadFileOverHttp' -Exactly -Times 1 -ParameterFilter { $Url -eq 'https://xxx.blob.core.windows.net/kubernetes/v1.29.2/windowszip/v1.29.2-1int.zip' }
         }
     }
+}
+
+Describe 'Disable-KubeletServingCertificateRotationForTags' {
+    BeforeEach {
+        Mock Write-Log
```
AbelHu, line 69: Write-Log is defined in parts/windows/windowscsehelper.ps1 and imported in this file so it seems like that we do not need to mock Write-Log

staging/cse/windows/kubeletfunc.tests.ps1
```diff
@@ -62,4 +62,133 @@ Describe 'Get-KubePackage' {
...
+        Disable-KubeletServingCertificateRotationForTags
+        Compare-Object $global:KubeletConfigArgs "--rotate-certificates=true,--rotate-server-certificates=false,--node-ip=10.0.0.1,anonymous-auth=false" | Should -Be $null
+        Compare-Object $global:KubeletNodeLabels "kubernetes.azure.com/agentpool=wp0" | Should -Be $null
+        Assert-MockCalled -CommandName 'Get-TagValue' -Exactly -Times 1 -ParameterFilter { $TagName -eq 'aks-disable-kubelet-serving-certificate-rotation' -and $DefaultValue -eq "false" }
+    }
+}
+
+Describe 'Get-TagValue' {
+    BeforeEach {
+        Mock Write-Log
```
AbelHu, line 113: ditto

parts/windows/windowscsehelper.ps1
cameronmeissner: thanks, missed this

AbelHu (approved):
LGTM. Thanks for the PR. Suggest testing the change with RP.

PR #4913, opened 2024-09-06
feat: onboard renovate to support auto-upgrading components

Devinwong:
**What type of PR is this?**
/kind feature

1. Onboard renovate to monitor most of the components in components.json to see if there are latest versions. (Monitor `ContainerImages` and `Packages`, except some packages hosted in acs-mirror. They will be moved to mcr)
2. It monitors runc, containerd based on the packages.microsoft.com registry list. It monitors mcr.microsoft.com for containerImages.
3. The auto generated PR can be assigned to corresponding owner for every package/containerImage. The feature is enabled. But still need a list of all the owners to add them to the renovate.json
4. Support `latestVersion` and `PreviousLatestVersion`. It also supports only single `latestVersion` without `PreviousLatestVersion` for those don't need.
5. Update components.cue schema to fit the new features. Add tests to cover the changes.
6. Add `VersionsV2` to replace `Versions` in `packages`. Add `MultiArchVersionsV2` to replace `MultiArchVersions`. Meanwhile supporting fallback case (e.g. old CSE is still expecting `Versions` and `MultiArchVersions` .)

An example that auto-generated PR https://github.com/Devinwong/AgentBaker/pull/121/files for a package.
Another example https://github.com/Devinwong/AgentBaker/pull/136/files for a containerImage.
ab build and e2e passed. https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=103011306&view=results

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


.github/renovate.json
```diff
@@ -1,19 +1,46 @@
 {
```
Devinwong, line 1: will have a wiki page to describe how to configure this renovate.json file

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -534,21 +544,61 @@ returnPackageVersions() {
...
+    # Fallback to versions if versionsV2 is null
+    local versions=($(echo "${package}" | jq -r ".downloadURIs.${os}.${release}.versions[]"))
     for version in ${versions[@]}; do
         PACKAGE_VERSIONS+=("${version}")
     done
-    return 0
+}
+
+# returns multiArchVersionsV2 if it exists, otherwise returns multiArchVersions
+returnMultiArchVersionsV2OrMultiArchVersions() {
```
djsly, line 582: is it really returning or population a global variable ? maybe reconsider the functiion name

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -3,49 +3,58 @@
     {
       "downloadURL": "mcr.microsoft.com/oss/kubernetes/autoscaler/addon-resizer:*",
       "amd64OnlyVersions": [],
-      "multiArchVersions": [
-        "1.8.20",
-        "1.8.22"
+      "multiArchVersionsV2": [
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com,name=oss/kubernetes/autoscaler/addon-resizer",
```
djsly, line 8: can we put a space after `, `

`"registry=https://mcr.microsoft.com, name=oss/kubernetes/autoscaler/addon-resizer",`

.github/renovate.json
```diff
@@ -24,10 +51,47 @@
...
+      ],
+      "matchStringsStrategy": "any",
+      "matchStrings": [
+        "\"renovateTag\":\\s*\"name=(?<packageName>[^,]+),os=ubuntu,release=20\\.04\",\\s*\"latestVersion\":\\s*\"(?<currentValue>[^\"]+)\"[^}]*"
+      ],
+      "datasourceTemplate": "custom.deb2004",
+      "versioningTemplate": "deb"      
+    },
+    {
+      "customType": "regex",
```
djsly, line 84: can you add a description field,

.github/renovate.json
```diff
@@ -24,10 +51,47 @@
...
+        "parts/linux/cloud-init/artifacts/components.json"
+      ],
+      "matchStringsStrategy": "any",
+      "matchStrings": [
+        "\"renovateTag\":\\s*\"name=(?<packageName>[^,]+),os=ubuntu,release=18\\.04\",\\s*\"versions\":\\s*\\[\\s*\"(?<currentValue>[^\"]+)\""
+      ],
+      "datasourceTemplate": "custom.deb1804",
+      "versioningTemplate": "deb"
+    },
+    {
```
djsly, line 71: can you add a description field,

.github/renovate.json
```diff
@@ -24,10 +51,47 @@
...
       "matchStrings": [
-        ".*?\"downloadURL\":\\s*\"(?<registryUrl>[^/]+)/(?<packageName>[^:]+)[\\s\\S]*?\"multiArchVersions\": \\[\\s*\"(?<currentValue>[^\\\"]+)\".*"
+        "\"renovateTag\":\\s*\"registry=(?<registryUrl>[^,]+),name=(?<packageName>[^\"]+)\",\\s*\"latestVersion\":\\s*\"(?<currentValue>[^\"]+)\"(?:[^}]*\"previousLatestVersion\":\\s*\"(?<depType>[^\"]+)\")?"
       ],
       "datasourceTemplate": "docker",
-      "registryUrlTemplate": "https://mcr.microsoft.com"
+      "autoReplaceStringTemplate": "\"renovateTag\": \"registry={{{registryUrl}}},name={{{packageName}}}\",\n          \"latestVersion\": \"{{{newValue}}}\"{{#if depType}},\n          \"previousLatestVersion\": \"{{{currentValue}}}\"{{/if}}"
+    },
+    {
+      "customType": "regex",
```
djsly, line 60: can you add a description field,

.github/renovate.json
```diff
@@ -24,10 +51,47 @@
       ],
       "matchStringsStrategy": "any",
```
djsly, line 52: can you add a description field,

.github/renovate.json
```diff
@@ -24,10 +51,47 @@
...
       "matchStrings": [
-        ".*?\"downloadURL\":\\s*\"(?<registryUrl>[^/]+)/(?<packageName>[^:]+)[\\s\\S]*?\"multiArchVersions\": \\[\\s*\"(?<currentValue>[^\\\"]+)\".*"
+        "\"renovateTag\":\\s*\"registry=(?<registryUrl>[^,]+),name=(?<packageName>[^\"]+)\",\\s*\"latestVersion\":\\s*\"(?<currentValue>[^\"]+)\"(?:[^}]*\"previousLatestVersion\":\\s*\"(?<depType>[^\"]+)\")?"
       ],
       "datasourceTemplate": "docker",
```
djsly, line 56: should we avoid using the `template` ? 

What is the delimiter between all of the custom Regex to ensure they don't pickup the same lines ?

.github/renovate.json
```diff
@@ -1,19 +1,46 @@
...
+    }
+  ],
   "packageRules": [
     {
+      "matchDatasources": ["docker", "custom.deb1804", "custom.deb2004", "custom.deb2204"],
+      "matchUpdateTypes": [
+        "minor"
+      ],
+      "automerge": false,
+      "enabled": false
```
djsly, line 25: why not enabled ?

.github/renovate.json
```diff
@@ -1,19 +1,46 @@
...
         "patch",
         "pin",
         "digest"
       ],
-      "automerge": false
+      "automerge": false,
+      "enabled": true
+    },
+    {
+      "matchPackageNames": ["moby-runc", "moby-containerd"],
```
djsly, line 38: could you fill out the full list, with the right skeleton, we can add some owner after if we don't know about them

.github/renovate.json
```diff
@@ -1,19 +1,46 @@
...
+      "assignees": ["devinwong"]
+    },
+    {
+      "matchPackageNames": ["moby-runc", "moby-containerd"],
```
djsly, line 42: can't this me merged with the previous entry, ? the matcher is the same

djsly (changes requested):

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: Makes sense. Updated to `updateMultiArchVersions` and `updateVersions` as they are both updating global variables.

parts/linux/cloud-init/artifacts/components.json
Devinwong: done

.github/renovate.json
Devinwong: done

.github/renovate.json
Devinwong: done

.github/renovate.json
Devinwong: done

.github/renovate.json
Devinwong: done

.github/renovate.json
Devinwong: This is a bit tricky. If we enable it, using azure-cni as an example, we will see the following PRs.
1. containernetworking/azure-cni	minor	v1.5.32 -> v1.6.6
2. containernetworking/azure-cni	patch	v1.6.3 -> v1.6.6
3. containernetworking/azure-cni	patch	v1.5.32 -> v1.5.36

#2 and #3 are what we need because we want to auto-update the patch version.
By enabling the "minor", #1 will also be created. 
As we discussed last week, we don't want Renovate to update minor for us, otherwise the assignee needs to cancel this PR manually every time (It is every time because v1.5.32 will/should never be updated to v1.6.x). Instead we are relying on the component owner to update to a new minor version.

Devinwong: As a result, there will be many noise PRs created. A comparison of without and with `minor` enabled.
image

.github/renovate.json
Devinwong: I didn't create all the placeholder blocks because multiple components could be assigned to 1 owner. So there could be less blocks.
For example, instead of 
```
   {
      "matchPackageNames": ["componentA"],
      "assignees": ["devinwong"],
    },
   {
      "matchPackageNames": ["componentB"],
      "assignees": ["devinwong"],
    },
   {
      "matchPackageNames": ["componentC"],
      "assignees": ["devinwong"],
    },
```

we can just do
```
   {
      "matchPackageNames": ["componentA, componentB, componentC"],
      "assignees": ["devinwong"],
    },
```

There are 50+ components now.
I can come back to add all the owners once we have a list, with their github IDs. Does that sound good?

.github/renovate.json
Devinwong: yes. done.

.github/renovate.json
Devinwong: `should we avoid using the template ?` => do you mean we should extract datasource from the comments/RenovateTag? or You mean something else?

Devinwong: The types of matchStrings we have now are as follows: (converted to human readable format)
1. "renovateTag": "registry=<RegistryURL>, name=<ContainerImage Name>"
2. "renovateTag": "name=moby-containerd, os=<OS>, release=<Release>"
     - for this type, we have os=ubuntu, release = 18.04|20.04|22.04|24.04
We need to always make sure each custom Regex only matches 1 type of renovateTag so that it won't pick up the same lines.

djsly: should we avoid using the template ? => do you mean we should extract datasource from the comments/RenovateTag? or You mean something else?

- correct, unless it will always be docker

djsly: The types of matchStrings we have now are as follows:

-- can we document this somerwhere ? I know it's in the code, but maybe finding a place to verbalize it will help others looking at updating that file.

.github/renovate.json
djsly: thanks for the explaination, make sure you add that to you TSG/ Training material ww

.github/renovate.json
djsly: I see, so the intention would be to broadcast the new feature to all those changing `components.json` and explaining them that we need them to provide us with the github alias and owner list. correct ?

djsly (approved):

.github/renovate.json
Devinwong: ya. writing this section now.

.github/renovate.json
Devinwong: Correct. I will follow up with adding the owners to this config.

.github/renovate.json
Devinwong: yeah the `containerImages` are always "docker". If it moves to somewhere else in the future, we can update the revonate.json and components.json.
And for `packages`, the custom managers are defined differently. So we still don't need to extract dataSource from RenovateTag. I will keep it as is until we see a demand to abstract the custom managers.
I will document all these.

PR #4923, opened 2024-09-11
feat: put credential provider package info into components.json 

AlisonB319:
**What type of PR is this?**
We want the version of credential provider to be managed by components.json

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -143,28 +143,34 @@ installNetworkPlugin() {
...
     if [[ -n "${CREDENTIAL_PROVIDER_DOWNLOAD_URL}" ]]; then
         # CREDENTIAL_PROVIDER_DOWNLOAD_URL is set by linuxCredentialProviderURL
         # The version in the URL is unknown. An acs-mirror or registry URL could be passed meaning the version must be extracted from the URL. 
-        CREDENTIAL_PROVIDER_VERSION=$(echo "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" | grep -oP 'v\d+(\.\d+)*' | sed 's/^v//' | head -n 1)
-    else
-        CREDENTIAL_PROVIDER_DOWNLOAD_URL="https://acs-mirror.azureedge.net/cloud-provider-azure/v${CREDENTIAL_PROVIDER_VERSION}/binaries/azure-acr-credential-provider-linux-${CPU_ARCH}-v${CREDENTIAL_PROVIDER_VERSION}.tar.gz"
+        cred_version_for_oras=$(echo "$CREDENTIAL_PROVIDER_DOWNLOAD_URL" | grep -oP 'v\d+(\.\d+)*' | sed 's/^v//' | head -n 1)
     fi
 
+    if [[ -n "$2" ]]; then
```
cameronmeissner, line 153: nit: can we clearly define some local vars to capture values of positional arguments to make things more readable?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -274,6 +274,13 @@ for p in ${packages[*]}; do
         evaluatedURL=$(evalPackageDownloadURL ${PACKAGE_DOWNLOAD_URL})
         installOras "${downloadDir}" "${evaluatedURL}" "${version}"
         echo "  - oras version ${version}" >> ${VHD_LOGS_FILEPATH}
+      done
+      ;;
+    "azure-acr-credential-provider")
+      for version in ${PACKAGE_VERSIONS[@]}; do
+        evaluatedURL=$(evalPackageDownloadURL ${PACKAGE_DOWNLOAD_URL})
+        downloadCredentialProvider "${downloadDir}" "${evaluatedURL}" "${version}"
+        echo "  - azure-acr-credential-provider version ${version}" >> ${VHD_LOGS_FILEPATH}
```
cameronmeissner, line 283: let's add some UTs to make sure we're resolving the correct version(s)

AlisonB319: linux-vhd-content-tests.sh actually verifies that the packages are installed properly - https://github.com/Azure/AgentBaker/blob/6d6960b123ea691bb50c2254aa94f9ddad549296/vhdbuilder/packer/test/linux-vhd-content-test.sh#L144

(source I spent a lot of time debugging why I was hitting that error message and it was because I was only downloading 1.29 instead of 1.30 because I was overwriting the version lol)

cameronmeissner: hmm okay, I know we have those spec tests now which is what I was referring to

we can come back and add those later if we're already getting e2e coverage though

cameronmeissner (approved):

PR #4928, opened 2024-09-11
chore: add NVMe disk controller type to new gen2 image defs

anujmaheshwari1:
**What type of PR is this?**
This PR enables tagging any new Gen2 Image Definition with diskControllerTypes=SCSI,NVMe to work with v6 SKUs

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/init-variables.sh
+++ b/vhdbuilder/packer/init-variables.sh
@@ -250,6 +250,10 @@ if [[ "$MODE" == "linuxVhdMode" || "$MODE" == "windowsVhdMode" ]]; then
 			TARGET_COMMAND_STRING+="--features SecurityType=ConfidentialVMSupported"
 		elif [[ ${ENABLE_TRUSTED_LAUNCH} == "True" ]]; then
 			TARGET_COMMAND_STRING+="--features SecurityType=TrustedLaunch"
+		elif [[ ${HYPERV_GENERATION} == "V2" ]]; then
+		  # NVMe SKUs are only enabled for Gen2 VM sizes, and Arm64 doesnt need special tagging + CVM sizes are not v6
+		  # TODO(amaheshwari): add a check for TL + Gen2 combo, currently fine since the only TL SKU is updated already
+		  TARGET_COMMAND_STRING+="--features DiskControllerTypes=SCSI,NVMe"
 		fi
 
 		az sig image-definition create \


vhdbuilder/packer/init-variables.sh
```diff
@@ -250,6 +250,10 @@ if [[ "$MODE" == "linuxVhdMode" || "$MODE" == "windowsVhdMode" ]]; then
 			TARGET_COMMAND_STRING+="--features SecurityType=ConfidentialVMSupported"
 		elif [[ ${ENABLE_TRUSTED_LAUNCH} == "True" ]]; then
 			TARGET_COMMAND_STRING+="--features SecurityType=TrustedLaunch"
+		elif [[ ${HYPERV_GENERATION} == "V2" ]]; then
+		  # NVMe SKUs are only enabled for Gen2 VM sizes, and Arm64 doesnt need special tagging + CVM sizes are not v6
+		  # TODO(amaheshwari): add a check for TL + Gen2 combo, currently fine since the only TL SKU is updated already
```
djsly, line 255: @anujmaheshwari1 , any reason the check cannot be added now ?  is there a technical limitation making that condition hard to implement ?

anujmaheshwari1: I dont have an elegant solution today:

when a new image def with TL is added, we want it to be stamped as "--features SecurityType=TrustedLaunch" but we also want it to be stamped as "--features DiskControllerTypes=SCSI,NVMe"
 
the right way to do that is --features "SecurityType=TrustedLaunch DiskControllerType=SCSI,NVMe"
 
however, when we do that as a TARGET_COMMAND_STRING, bash expands it by escaping the quotes even if I properly already escaped them.
 
which causes validation failure for diskControllerType.

I spent a lot of time trying to make it work but couldnt find an elegant way to do it without a whole bunch of conditionals. Since we are not going to add a new TL image definition anytime soon, I added a TODO for now.

cameronmeissner: @anujmaheshwari1 would this just work if we change the `elif`'s to separate if statements?

cameronmeissner (approved):

cameronmeissner:
assuming we don't need this now that your other PR just merged

djsly (approved):

PR #4930, opened 2024-09-12
Timmy/installer

timmy-wright:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


node-bootstrapper/main.go
```diff
@@ -60,24 +63,36 @@ func (s SensitiveString) UnsafeValue() string {
 	return string(s)
 }
 
+func isWindows() bool {
+	return os.PathSeparator == '\\' && os.PathListSeparator == ';'
```
r2k1, line 67: `runtime.GOOS == "windows"`

node-bootstrapper/main.go
```diff
@@ -60,24 +63,36 @@ func (s SensitiveString) UnsafeValue() string {
...
+}
+
+func mainWithDefer(logFile *os.File) error {
```
r2k1, line 70: Maybe something like `RunWithLog` will carry a better meaning

node-bootstrapper/main.go
```diff
@@ -60,24 +63,36 @@ func (s SensitiveString) UnsafeValue() string {
...
 		fmt.Printf("failed to open log file: %s\n", err)
 		os.Exit(1)
 	}
-	defer logFile.Close()
 
-	logger := slog.New(slog.NewJSONHandler(logFile, nil))
-	slog.SetDefault(logger)
+	// mainWithDefer is responsible for closing the log file. That way we know it's closed
+	// and can use os.Exit safely in this method.
+	err = mainWithDefer(logFile)
```
r2k1, line 91: you can't use logger past this point.
I usually try to put creating and closing things in the same function to avoid such issues.

I think I sorted it out differently in the main.

node-bootstrapper/main.go
```diff
@@ -158,9 +189,27 @@ func provisionStart(ctx context.Context, config *datamodel.NodeBootstrappingConf
...
+
+	var cmd *exec.Cmd
+	if config.AgentPoolProfile.IsWindows() {
+		systemDrive := os.Getenv("SYSTEMDRIVE")
+		slog.Info(fmt.Sprintf("systemDrive: %s\n\n", systemDrive))
+		if systemDrive == "" {
+			systemDrive = "C:"
+		}
+		script := string(cse)
+		script = strings.ReplaceAll(script, "%SYSTEMDRIVE%", systemDrive)
```
r2k1, line 201: You can check https://pkg.go.dev/os#ExpandEnv for an inspiration (seems like it's not windows friendly).

Do you want to envsubst a single variable or all of them?

node-bootstrapper/main.go
```diff
@@ -217,8 +279,28 @@ type File struct {
 	Mode    os.FileMode
 }
 
-func customData(config *datamodel.NodeBootstrappingConfiguration) (map[string]File, error) {
-	contentDockerDaemon, err := generateContentDockerDaemonJSON(config)
+func customData(ctx context.Context, config *datamodel.NodeBootstrappingConfiguration) (map[string]File, error) {
+	if config.AgentPoolProfile.IsWindows() {
+		customData, err2 := OldCustomData(ctx, config)
+		if err2 != nil {
+			log.Fatal("error:", err2)
```
r2k1, line 286: Avoid using log.Fatal unless absolutely necessary. It exists application immediately, without calling any `defer` calls.
We using different logger. 

Also, `err2`?

pkg/agent/datamodel/types.go
```diff
@@ -350,6 +350,18 @@ const (
...
+type BootstrappingMethod string
+
+const (
+	UseArcMsiToMakeCSR        BootstrappingMethod = "UseArcMsiToMakeCSR"
+	UseAzureMsiToMakeCSR      BootstrappingMethod = "UseAzureMsiToMakeCSR"
+	UseArcMsiDirectly         BootstrappingMethod = "UseArcMsiDirectly"
+	UseAzureMsiDirectly       BootstrappingMethod = "UseAzureMsiDirectly"
+	UseSecureTLSBootstrapping BootstrappingMethod = "UseSecureTLSBootstrapping"
+	//nolint:gosec // this is a const string to use in switch statements, not hardcoded credentials
+	UseTLSBootstrapToken BootstrappingMethod = "UseTLSBootstrapToken"
```
r2k1, line 362: I would do something like
```go
const (
	BootstrappingMethodArcMSI                 BootstrappingMethod = "ArcMSI"
	BootstrappingMethodArcMSIToMakeCSR        BootstrappingMethod = "ArcMSIToMakeCSR"
	BootstrappingMethodAzureMSI               BootstrappingMethod = "AzureMSI"
	BootstrappingMethodAzureMSIToMakeCSR      BootstrappingMethod = "AzureMSIToMakeCSR"
	BootstrappingMethodSecureTLSBootstrapping BootstrappingMethod = "SecureTLSBootstrapping"
	//nolint:gosec // this is a const string to use in switch statements, not hardcoded credentials
	BootstrappingMethodTLSBootstrapToken BootstrappingMethod = "TLSBootstrapToken"
)
```

node-bootstrapper/use_kubeconfig.go
```diff
@@ -0,0 +1,396 @@
...
+		datamodel.UseAzureMsiToMakeCSR:
+		if err := useAzureTokenSh(config, files); err != nil {
+			return err
+		}
+
+	case datamodel.UseTLSBootstrapToken, datamodel.UseSecureTLSBootstrapping:
+		break
+
+	default:
+		break
```
r2k1, line 32: do we want to make this field mandatory in the new contract?

pkg/agent/datamodel/types.go
r2k1: It's probably not `BootstrappingMethod`, but something like `ClusterAuthMethod`?

node-bootstrapper/main.go
timmy-wright: Yup - your solution is to not use a defer and to just close it manually. Which also works. So I'll not migrate the "runWithLog" code over :)

node-bootstrapper/main.go
timmy-wright: Much cleaner.

node-bootstrapper/main.go
timmy-wright: Just the one needed to be substituted. Will have a look.

node-bootstrapper/use_kubeconfig.go
timmy-wright: We probably should. Yes.

PR #4936, opened 2024-09-12
feat(abe2e): move kubelet node IP validator to common validators

cameronmeissner:
**What type of PR is this?**

makes the kubelet node ip validator a common validator, though still excludes it from scenarios (privatepkg) that utilize older VHDs that aren't at least on k8s 1.29

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/Azure/agentbakere2e/config"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


wedaly (approved):

e2e/validation.go
```diff
@@ -148,6 +149,12 @@ func commonLiveVMValidators(opts *scenarioRunOpts) []*LiveVMValidator {
 		},
 	}
 	validators = append(validators, leakedSecretsValidators(opts)...)
+
+	// kubeletNodeIPValidator cannot be run on older VHDs with kubelet < 1.29
+	if opts.scenario.VHD.Version != config.VHDUbuntu2204Gen2ContainerdPrivateKubePkg.Version {
+		validators = append(validators, kubeletNodeIPValidator())
+	}
```
AlisonB319, line 156: will this add the nodeIPValidator to all scenarios (besides kubelet 1.29)? Do we want that considering it only previously existed on a subset of scenarios?

djsly: The existing test where is was enabled seem somewhat random. It feels like a copy paste issue where some people copied from test with and without over time. 

I might be wrong

djsly (approved):

e2e/validation.go
cameronmeissner: https://github.com/Azure/AgentBaker/pull/4824#discussion_r1733514591

PR #4973, opened 2024-09-18
test: detect regressions with build performance program

zachary-bailey:
**What type of PR is this?**

/kind test

This PR does the following:

-  Ingests local build performance data into kusto for later analysis
-  Queries aggregate performance data for the relevant SKU in kusto
    - Kusto will contain mean and stdev for every section in every script for every SKU
-  Compares queried data against performance data recorded for the local build
-  Detects regressions and prints them to logs, notifying user with a warning
    - Regression is currently defined as any time over 3 Stdev over the mean, but that can change if team would like a different 
    metric
    - This is to prevent the logs from being too noisy and only showing regressions for sections that are significantly out of spec

This PR will prevent VHD Build time regression by notifying software engineers of their code changes effect on VHD build times. Additionally, it can illuminate problem areas in the build when retry logic is causing the build to take longer than expected.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Reviewer Notes**:

Open to any suggestions related to testing and error handling specifically, as I would like this program to be as robust as possible. Currently, I am wrapping all errors and sending them back to main.

Kusto functions are not currently tested because they are just wrappers around the go SDK functions to keep main clean and readable. 

Once code is approved, I will create a new cluster and change a couple variables as I am currently only using a dev/test kusto instance. After that I will merge the code.


go.sum
zachary-bailey, line 1: On the master branch, these mods are still listed but have yellow squiggly marks, citing that they are no longer needed. I ran go mod tidy in the overall directory and now can't get them back in the go sum because they automatically disappear. Are we keeping them around for some reason?

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// After preparing performance data, evaluate it with EvaluatePerformance and associated helper functions
+func (maps *DataMaps) EvaluatePerformance() error {
+	// Iterate over LocalPerformanceDataMap and compare it against identical sections in QueriedPerformanceDataMap
+	for scriptName, scriptData := range maps.LocalPerformanceDataMap {
+		for section, timeElapsed := range scriptData {
+			// The value of QueriedPerformanceDataMap[scriptName][section] is an array with two elements: [avg, 2*stdev]
+			// Adding these together gives us the maximum time allowed for the section
+			maxTimeAllowed, err := SumSlice(maps.QueriedPerformanceDataMap[scriptName][section])
```
jason1028kr, line 138: Should we have distinct error/message for newly added section?

vhdbuilder/packer/buildperformance/evaluate-build-performance.sh
```diff
@@ -56,5 +55,11 @@ for script in "${scripts[@]}"; do
   echo "##[endgroup]"
 done
 
-rm ${SIG_IMAGE_NAME}-build-performance.json
+mv ${SIG_IMAGE_NAME}-build-performance.json vhdbuilder/packer/buildperformance
+pushd vhdbuilder/packer/buildperformance
+  echo -e "\nRunning build performance evaluation program...\n"
+  ./buildPerformance
```
jason1028kr, line 61: Just wanted to confirm, this would fail the pipeline step if the binary panics right?

zachary-bailey: Thanks for pointing this out! I originally chose panic because I read that log.Fatalf would not allow the deferred functions for closing client, ctx, etc to execute. I will make sure the panics are handled gracefully and the step doesn't fail.

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+			if timeElapsed > maxTimeAllowed {
+				if maps.RegressionMap[scriptName] == nil {
+					maps.RegressionMap[scriptName] = map[string]float64{}
+				}
+				// Record the amount of time the section exceeded the maximum allowed time by
+				maps.RegressionMap[scriptName][section] = timeElapsed - maxTimeAllowed
+			}
+		}
+	}
```
zachary-bailey, line 161: @jason1028kr good catch, thanks! Does the code change mitigate your concerns? On line 138 I added an access check on that key, and assuming it is successful, I then use the returned value for the rest of the section. 

The check for -1 returned from SumSlice ensures that enough data was available in Kusto to calculate both Stdev and mean. If either is not true, the CleanData function changes NaN values to -1 before this point so that the JSON can be unmarshalled and then evaluated in this function.

jason1028kr: Got it, thank you for the explanation. Yep, the change looks good!

vhdbuilder/packer/buildperformance/evaluate-build-performance.sh
jason1028kr: I see, I wasn't sure if we intended to fail the step on error.

vhdbuilder/packer/buildperformance/main.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		panic(err)
+	}
+	defer client.Close()
+	fmt.Println("Kusto client created")
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
+
+	if config.SourceBranch == "refs/heads/zb/regression2" {
```
cameronmeissner, line 39: im assuming we'd get rid of this once we're ready for merge?

vhdbuilder/packer/buildperformance/main.go
```diff
@@ -0,0 +1,62 @@
...
+			fmt.Println("Recovered from panic in main:", r)
+			os.Exit(0)
+		}
+	}()
+
+	config, err := service.SetupConfig()
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println("Program config set")
```
cameronmeissner, line 25: maybe would be better to use the `log` package in this case?

vhdbuilder/packer/buildperformance/pkg/service/helper.go
cameronmeissner, line 1: nit: rename to service.go + service_test.go?

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+		"kustoIngestionMapping":     os.Getenv("BUILD_PERFORMANCE_INGESTION_MAPPING"),
+		"sourceBranch":              os.Getenv("GIT_BRANCH"),
+		"sigImageName":              os.Getenv("SIG_IMAGE_NAME"),
+		"localBuildPerformanceFile": os.Getenv("SIG_IMAGE_NAME") + "-build-performance.json",
+	}
+	missingVar := false
+	for name, value := range envVars {
+		if value == "" {
+			fmt.Printf("missing environment variable \"%s\".", name)
+			missingVar = true
```
cameronmeissner, line 27: nit: we can `break` to short-circuit the loop after setting to true here

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+		"kustoClientId":             os.Getenv("BUILD_PERFORMANCE_CLIENT_ID"),
+		"kustoIngestionMapping":     os.Getenv("BUILD_PERFORMANCE_INGESTION_MAPPING"),
+		"sourceBranch":              os.Getenv("GIT_BRANCH"),
+		"sigImageName":              os.Getenv("SIG_IMAGE_NAME"),
+		"localBuildPerformanceFile": os.Getenv("SIG_IMAGE_NAME") + "-build-performance.json",
+	}
+	missingVar := false
+	for name, value := range envVars {
+		if value == "" {
+			fmt.Printf("missing environment variable \"%s\".", name)
```
cameronmeissner, line 26: use `%q` instead of `\"%s\"` to achieve what you want

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+		// QueriedPerformanceDataMap will hold the aggregated performance data queried from Kusto
+		QueriedPerformanceDataMap: map[string]map[string][]float64{},
+		// RegressionMap will hold all identified regressions in the current build
+		RegressionMap: map[string]map[string]float64{},
+	}
+}
+
+// Prepare performance data for evaluation with PreparePerformanceDataForEvaluation and associated helper functions
+func (maps *DataMaps) PreparePerformanceDataForEvaluation(localBuildPerformanceFile string, queriedData *SKU) error {
+	err := maps.DecodeLocalPerformanceData(localBuildPerformanceFile)
```
cameronmeissner, line 59: nit: in most cases where the func is only returning an error, I like doing something like this instead:

```
if err := maps.DecodeLocalPerformanceData(localBuildPerformanceFile); err != nil {
    ...
}
```

the `err` will then only be scoped to the if-block

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	var m map[string]json.RawMessage
+	err = json.NewDecoder(file).Decode(&m)
+	if err != nil {
+		return fmt.Errorf("error decoding local JSON file: %w", err)
+	}
+
+	key := "scripts"
+	raw := m[key]
+
+	holdingMap := map[string]map[string]string{}
```
cameronmeissner, line 86: I'd considering aliasing a few of these nested map types as custom types for readability

cameronmeissner: e.g. `type holder map[string]map[string]string`

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (maps *DataMaps) DecodeLocalPerformanceData(filePath string) error {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return fmt.Errorf("could not open local JSON file: %w", err)
+	}
+	defer file.Close()
+
+	var m map[string]json.RawMessage
```
cameronmeissner, line 77: I'd prefer we properly define a struct and use `Unmarshal` here rather than using `RawMessage` + `Decode`

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	kustoData := []byte(data.SKUPerformanceData)
+	err := json.Unmarshal(kustoData, &maps.QueriedPerformanceDataMap)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling kusto data: %w", err)
+	}
+	return nil
+}
+
+func (sku *SKU) CleanData() string {
+	var cleanedData string = strings.ReplaceAll(sku.SKUPerformanceData, "NaN", "-1")
```
cameronmeissner, line 127:
```suggestion
	return strings.ReplaceAll(sku.SKUPerformanceData, "NaN", "-1")
```

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+func (maps DataMaps) PrintRegressions() error {
+	prefix := ""
+	indent := "  "
+
+	data, err := json.MarshalIndent(maps.RegressionMap, prefix, indent)
+	if err != nil {
+		return fmt.Errorf("error marshalling regression data: %w", err)
+	}
+
+	fmt.Println(string(data))
```
cameronmeissner, line 196: lets add context to this line if we want to keep it

vhdbuilder/packer/buildperformance/pkg/service/helper.go
```diff
@@ -0,0 +1,198 @@
...
+	if slice[0] == -1 || slice[1] == -1 {
+		sum = -1
+		return sum, nil
+	}
+	sum = slice[0] + slice[1]*2
+	return sum, nil
+}
+
+func (maps DataMaps) PrintRegressions() error {
+	prefix := ""
```
cameronmeissner, line 188: nit: redundant defs

vhdbuilder/packer/buildperformance/pkg/service/kusto_functions.go
```diff
@@ -0,0 +1,63 @@
...
+func CreateKustoClient(kustoEndpoint string, kustoClientId string) (*kusto.Client, error) {
+	kcsb := kusto.NewConnectionStringBuilder(kustoEndpoint).WithUserManagedIdentity(kustoClientId)
+	client, err := kusto.New(kcsb)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+func IngestData(client *kusto.Client, ctx context.Context, kustoDatabase string, kustoTable string, buildPerformanceDataFile string, kustoIngestionMap string) error {
```
cameronmeissner, line 23: nit: we usually have `ctx context.Context` as the first positional argument for funcs that take in a context

vhdbuilder/packer/buildperformance/pkg/service/kusto_functions.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	defer iter.Stop()
+
+	data := SKU{}
+	err = iter.DoOnRowOrError(
+		func(row *table.Row, e *kustoErrors.Error) error {
+			if e != nil {
+				return fmt.Errorf("error while iterating over query table: %w", e)
+			}
+			if err := row.ToStruct(&data); err != nil {
```
cameronmeissner, line 53: would this be decoding to `&data` multiple times with different data if the result set has more than 1 row? if so that might be problematic

vhdbuilder/packer/buildperformance/main.go
zachary-bailey: Will do

vhdbuilder/packer/buildperformance/main.go
zachary-bailey: Yes, this is just a placeholder to enable functionality for this branch. Once its ready to merge, it will be changed to master. That way master will be the source of truth for data analysis, and other branches will evaluate their performance relative to it.

vhdbuilder/packer/buildperformance/pkg/service/helper.go
zachary-bailey: Good idea! Will get on that.

vhdbuilder/packer/buildperformance/pkg/service/helper.go
zachary-bailey: Done. Also renamed `kusto_functions.go` to `kusto_service.go` to keep it consistent with the rest of the package.

vhdbuilder/packer/buildperformance/main.go
zachary-bailey: Changed to using the log package throughout most of the program.

vhdbuilder/packer/buildperformance/pkg/service/helper.go
zachary-bailey: Implemented this suggestion throughout the program, though it did make the `IngestData` call in `main.go` a bit long. Its definition is also long. Will stack arguments/parameters if you think its appropriate. 

I also considered just passing in the whole config object to make the function calls/signatures a bit shorter, but I understand we generally aren't supposed to expose fields to a function if the function doesn't need it. I briefly considered making every field a constant to prevent issues, but ended up just leaving it as is for review. Any preference? @cameronmeissner

vhdbuilder/packer/buildperformance/pkg/service/helper.go
zachary-bailey: Done, thanks for the suggestion. Its a lot better now.

vhdbuilder/packer/buildperformance/pkg/service/kusto_functions.go
zachary-bailey: Will DM

vhdbuilder/packer/buildperformance/pkg/service/kusto_functions.go
zachary-bailey: Fixed.

jason1028kr:
@zachary-bailey If you are still working on this, could we move Kusto interface code/pkg outside the build-performance path? I think I'll be using your code for cve work as well. Not a blocker though, I can do this on my end later as well.

PR #4978, opened 2024-09-19
enforcing bind-mount to complete before Kubelet

djsly:
**What type of PR is this?**

Ensure that bind-mount.service exists before Kubelet starts, preventing a race condition affecting double mounting /var/lib/kubelet.

https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files

oneshot: This type indicates that the process will be short-lived and that systemd should wait for the process to exit before continuing on with other units. This is the default Type= and ExecStart= are not set. It is used for one-off tasks.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/bind-mount.service
+++ b/parts/linux/cloud-init/artifacts/bind-mount.service
@@ -3,6 +3,7 @@ Description=Bind mount kubelet data
 [Service]
 Restart=on-failure
 RemainAfterExit=yes
+Type=oneshot
 ExecStart=/bin/bash /opt/azure/containers/bind-mount.sh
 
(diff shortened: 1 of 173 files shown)


cameronmeissner (approved):

cameronmeissner:
need a `make generate`

juan-lee:
How does making it oneshot ensure that bind-mount.service is there before kubelet?

djsly:
> How does making it oneshot ensure that bind-mount.service is there before kubelet?

@juan-lee description was updated

PR #4989, opened 2024-09-23
kata: Use image for AzL3 Kata UVM

ms-mahuber:
While AzL2 uses an initrd file for the Kata UVM, AzL3 will use an image file aligned with Kata CoCo where an image file is being used.

### DIFF
--- a/vhdbuilder/packer/packer_source.sh
+++ b/vhdbuilder/packer/packer_source.sh
@@ -144,8 +144,32 @@ copyPackerFiles() {
 
   if grep -q "kata" <<< "$FEATURE_FLAGS"; then
     # KataCC SPEC file assumes kata config points to the files exactly under this path
-    KATA_CONFIG_DIR=/var/cache/kata-containers/osbuilder-images/kernel-uvm/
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/packer_source.sh
```diff
@@ -144,9 +144,33 @@ copyPackerFiles() {
 
   if grep -q "kata" <<< "$FEATURE_FLAGS"; then
     # KataCC SPEC file assumes kata config points to the files exactly under this path
-    KATA_CONFIG_DIR=/var/cache/kata-containers/osbuilder-images/kernel-uvm/
+    if [[ $OS_VERSION == "2.0" ]]; then
+      KATA_CONFIG_DIR=/var/cache/kata-containers/osbuilder-images/kernel-uvm
+    elif [[ $OS_VERSION == "3.0" ]]; then
+      KATA_CONFIG_DIR=/usr/share/kata-containers
+    else
+      exit 1
```
cameronmeissner, line 152: can we echo an error message here describing the nature of the issue before exiting?

PR #4991, opened 2024-09-24
feat: Disable cgroup-memory-telemetry.service on AzureLinux

miz060:
**What type of PR is this?**

cgroup-memory-telemetry.service has been failing on AzureLinux/Mariner images because the host uses cgroupfs driver instead of systemd's cgroupsv2 driver. Disable this service until support for AzureLinux/Mariner is added.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -493,9 +493,12 @@ capture_benchmark "download_gpu_device_plugin"
 
 mkdir -p /var/log/azure/Microsoft.Azure.Extensions.CustomScript/events
 
-systemctlEnableAndStart cgroup-memory-telemetry.timer || exit 1
-systemctl enable cgroup-memory-telemetry.service || exit 1
-systemctl restart cgroup-memory-telemetry.service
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -485,9 +485,11 @@ capture_benchmark "download_gpu_device_plugin"
 
 mkdir -p /var/log/azure/Microsoft.Azure.Extensions.CustomScript/events
 
-systemctlEnableAndStart cgroup-memory-telemetry.timer || exit 1
-systemctl enable cgroup-memory-telemetry.service || exit 1
-systemctl restart cgroup-memory-telemetry.service
+if [ ! isMarinerOrAzureLinux "$OS" ]; then
```
cameronmeissner, line 488: can we add a comment for why we're no longer doing this on azure linux?

ganeshkumarashok (approved):

vhdbuilder/packer/install-dependencies.sh
miz060: Comment added.

PR #4996, opened 2024-09-26
feat: use private container registry for network isolated e2e tests

AlisonB319:
**What type of PR is this?**

**What this PR does / why we need it**: For NIC e2e tests we want to make sure they are not using acs-mirror anywhere. After the move to oras installs, acs mirror should no longer be required for airgap tests

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/aks_model.go
```diff
@@ -147,7 +162,6 @@ func getRequiredSecurityRules(clusterFQDN string) ([]*armnetwork.SecurityRule, e
 	requiredDNSNames := []string{
 		"mcr.microsoft.com",
```
fseldow, line 163: i think we can try to remove `mcr.microsoft.com` and `management.azure.com`
or maybe next pr?
Btw, i may miss, how the acr is created? is it in another pr?

fseldow: I guess there is no private dns zone and private endpoint in the cluster vnet to connect the acr. Thus we may need to whitelist the acr fqdn as the workaround in this pr?

AlisonB319: There is a private endpoint for the acr. From what I can tell in the Microsoft docs, we should just need to whitelist the AzureContainerRegistry - I’ve tested these nsg settings in a VM. Why are you thinking we need to whitelist the acr fqdn? Just trying to get more specific on the address that’s whitelisted?

fseldow: the pe for the acr is under vnet named `oras-login-testing-vnet`. Thus, in my understanding, only vm under the same vnet can connect to this acr via private endpoint. Otherwise, it goes through public network. i guess the e2e cluster is not under the same vnet, which causes the e2e failure when acr fqdn `aksvhdtestcr.azurecr.io` is not whitelisted

AlisonB319: I’m not sure that it is the case, I’ll follow up. From what I can tell, we are hitting a redundant e2e failure not related to the pe. We would be seeing timeouts related to Oras pull if it didn’t have the correct permissions for the acr

AlisonB319: You're right, so I'm adding some code to create a PE between the airgap e2e resource group and the ACR, that way the allowAllVnet in the nsg should work without a problem :)

e2e/aks_model.go
```diff
@@ -141,13 +146,52 @@ func airGapSecurityGroup(location, clusterFQDN string) (armnetwork.SecurityGroup
...
+		Location: to.Ptr(config.Config.Location),
+		Properties: &armnetwork.PrivateEndpointProperties{
+			Subnet: &armnetwork.Subnet{
+				ID: to.Ptr(subnetId),
+			},
+			PrivateLinkServiceConnections: []*armnetwork.PrivateLinkServiceConnection{
+				{
+					Name: to.Ptr(endpointName),
+					Properties: &armnetwork.PrivateLinkServiceConnectionProperties{
+						PrivateLinkServiceID: to.Ptr("/subscriptions/8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8/resourceGroups/aksvhdtestbuildrg/providers/Microsoft.ContainerRegistry/registries/aksvhdtestcr"),
```
cameronmeissner, line 161: did we make sure to delete-locks the new ACR?

e2e/aks_model.go
```diff
@@ -141,13 +146,52 @@ func airGapSecurityGroup(location, clusterFQDN string) (armnetwork.SecurityGroup
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create private endpoint in BeginCreateOrUpdate: %w", err)
+	}
+
+	peResp, err := poller.PollUntilDone(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create private endpoint in polling: %w", err)
+	}
+	fmt.Printf("Private Endpoint created or updated with ID: %s\n", *peResp.ID)
```
cameronmeissner, line 185: can we use `t.Logf` instead?

e2e/aks_model.go
```diff
@@ -141,13 +146,52 @@ func airGapSecurityGroup(location, clusterFQDN string) (armnetwork.SecurityGroup
...
+	return nil
+}
+
 func getRequiredSecurityRules(clusterFQDN string) ([]*armnetwork.SecurityRule, error) {
 	// https://learn.microsoft.com/en-us/azure/aks/outbound-rules-control-egress#azure-global-required-fqdn--application-rules
 	// note that we explicitly exclude packages.microsoft.com
 	requiredDNSNames := []string{
 		"mcr.microsoft.com",
```
cameronmeissner, line 193: removing MCR + ARM in a follow-up PR then?

e2e/aks_model.go
AlisonB319: yep! I've had to remove it during testing since it affects the deletion of test private end points, but the plan is for there to be a delete lock.

AlisonB319:
All airgap e2e's are passing, e2e's just seem to be a bit unhappy in general atm.

e2e/scenario_test.go
```diff
@@ -1137,9 +1137,8 @@ func Test_ubuntu2204WasmAirGap(t *testing.T) {
 				nbc.OutboundType = datamodel.OutboundTypeBlock
 				nbc.ContainerService.Properties.SecurityProfile = &datamodel.SecurityProfile{
 					PrivateEgress: &datamodel.PrivateEgress{
-						Enabled: true,
-						// TODO(xinhl): create one private acr instead of mcr.microsoft.com
-						ContainerRegistryServer: "mcr.microsoft.com",
+						Enabled:                 true,
+						ContainerRegistryServer: "aksvhdtestcr.azurecr.io/aks",
```
djsly, line 1141: @xinhl / @AlisonB319 I though that there wouldnt be any remote/cache name that we would be receiving ?

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.13.0 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.10.0 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v6 v6.1.0 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/privatedns/armprivatedns v1.3.0 // indirect
```
djsly, line 27: why is this change needed outside of the e2e go.mod ?

e2e/aks_model.go
djsly: wait, how will the lock be removed when we create a new AKS cluster (hash changes?)

go.mod
AlisonB319: thanks for the catch! It must have gotten then when I was trying to fix my vscode / go / goproxy stuff. Cleaned it up :)

e2e/scenario_test.go
AlisonB319: From your question on 7/13 it seems to still be open ended - https://teams.microsoft.com/l/message/19:5c77d666c90e4cfb98b66a226f74d957@thread.v2/1721762320450?context=%7B%22contextType%22%3A%22chat%22%7D

I've pinged Xhine, so he should see this, this evening.

cameronmeissner (approved):

PR #4999, opened 2024-09-26
feat: auto-generate action workflow

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/workflows/auto-generate.yaml
+++ b/.github/workflows/auto-generate.yaml
@@ -1,0 +1,32 @@
+name: auto-generate
+on: pull_request
+
+jobs:
+  auto-generate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.ref }}
+
+    - uses: actions/setup-go@v3
+      with:
+        go-version: '1.22'
+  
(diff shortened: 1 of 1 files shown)


cameronmeissner:
from further reading it seems that we can't utilize the GITHUB_TOKEN secret to make the auto-generated commit to the PR branch - that's due to the fact that commits made using GITHUB_TOKEN will _not_ trigger CI jobs which would otherwise be triggered when the commit is made by a human author, which in our case would mostly likely cause problems since many of our CI workflows are considered required in all cases: https://github.com/orgs/community/discussions/25702#discussioncomment-3248819

to get around that limitation, we'd need to utilize a personal access token instead, which in and of itself brings it own challenges - at this point I'm not sure if we'd be able to use fine-grained access tokens or if we'd be forced to use classic access tokens. in either case, manually managing PATs is a cumbersome process and isn't really blessed by our current security posture.

further, my understanding is that in order for the workflow to be secure with respect to PRs potentially coming in from forks, we'd need to explicitly only allow it to be executed when being executed from _this_ repository. in other words, the repo owner must be `Azure`:  https://github.com/orgs/community/discussions/26409

PR #5004, opened 2024-09-27
Remove Testdata under sourced control

djsly:
convert script from sh to go, update makefile logic, adding testdata to gitignore

**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,7 +216,7 @@ coverage:
 	@scripts/ginkgo.coverage.sh --codecov
 
 .PHONY: unit-tests
-unit-tests: generate
+unit-tests:
 	$(GO) test `go list ./... | grep -v e2e` -coverprofile coverage_raw.out -covermode count
 
 include versioning.mk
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -99,13 +106,6 @@ generate: bootstrap
 	./hack/tools/bin/cue export ./schemas/manifest.cue > ./parts/linux/cloud-init/artifacts/manifest.json
```
cameronmeissner, line 106: should we make a `validate-manifest` as well?

personally I actually think we can remove all references to manifest.json from everywhere _except_ the provisioning scripts themselves, since we no longer actually use it when building VHDs. though we'll need to keep using it as a fallback in the provisioning scripts for cases where newer CSE/customdata is running on older VHDs

.gitignore
```diff
@@ -19,6 +19,10 @@ test/e2e/kubernetes/junit.xml
 _logs/
 *.swp
 
+# test data
+pkg/agent/testdata/
```
cameronmeissner, line 23: still not sure we can just do this, will discuss offline

PR #5013, opened 2024-09-30
feat: use 1ES agent image for build pipeline

zachary-bailey:
**What type of PR is this?**

/feat

This PR removes the use of the go dev container in the build pipeline, instead utilizing the 1ES agent image and its cached binaries. 

The [`go dev container`](https://github.com/deis/docker-go-dev) is deprecated and will no longer have new releases. This has resulted in having outdated binaries for Go, AzCopy, Packer, AzCLI, and others. This PR will ensure that we are using updated binaries and making use of their new features and security fixes.

- For variables that are dynamically set during the build process (grep'd from packer output or settings), this PR sets them at once in a single pipeline step, enabling their use for the entire pipeline.

- Each pipeline step follows a common ordering convention for keywords: 
     - Task type declaration (Task, Bash, CopyFiles, etc)
     - Bash / Multi-Line bash, if applicable
     - Condition, if applicable
     - Display Name
     - Environment variables / inputs, if applicable
     
- Environment variables are grouped by their origin and type for easy identification and organization. If anyone would like them ordered differently, please let me know! I am hoping to make the yaml easy to read and intuitive for everyone. Ideas are welcome for any ways it could be better.

- `ENABLE_FIPS` was set to 'false' for any image template that did not already have it set. If it wasn't set, SKUs that didn't have it would fail because that variable would not be empty, but would instead be set to `ENABLE_FIPS` which causes a bug. Just a Azure Pipelines quirk.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -83,13 +83,18 @@ echo "Granting access to $disk_resource_id for 1 hour"
...
 
-# TBD: Need to investigate why `azcopy-preview login --login-type=MSI` does not work
-echo "Setting azcopy environment variables with pool identity: $AZURE_MSI_RESOURCE_STRING"
-export AZCOPY_AUTO_LOGIN_TYPE="MSI"
-export AZCOPY_MSI_RESOURCE_STRING="$AZURE_MSI_RESOURCE_STRING"
-
 echo "Uploading $disk_resource_id to ${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"
-azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd" --recursive=true || exit $?
+
+if [[ "${OS_TYPE}" == "Linux" ]]; then
```
Devinwong, line 88: [Just for my learning] Not fully understanding the if condition here. If it isn't a Linux, what will it be? windows?

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -2,8 +2,6 @@ name: $(Date:yyyyMMdd)$(Rev:.r)_$(Build.SourceBranchName)_$(BuildID)
 trigger: none
 
 variables:
-  - name: CONTAINER_IMAGE
-    value: 'mcr.microsoft.com/oss/azcu/go-dev:v1.38.1'
```
Devinwong, line 6: Got you that the go-dev container is deprecated so we removed it. Does it mean the build will now be running on the Agent host instead of the go-dev container on the Agent?

zachary-bailey: The build will be running on the packer VM that is provisioned by the code/packer templates. This will be orchestrated by the agent VM. It used to be orchestrated on the agent vm in the container. This PR just removes the container from the process.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
zachary-bailey: Yes its either Linux or windows. The windows pipeline still uses an older container that has to use the old version of azcopy.

Devinwong (approved):
lgtm

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -83,13 +83,18 @@ echo "Granting access to $disk_resource_id for 1 hour"
...
+  azcopy login --login-type=MSI
```
djsly, line 89: Do we know which MSI is attached to?  Are we in the agent at this time ?

zachary-bailey: I believe this references the VMs system assigned identity, as all the docs show AzCopy requires either a Client/Object ID or MSI resource string to use a user-assigned identity. @djsly

cameronmeissner: this is a user-assigned identity, you only need to specify a client/object/resource ID when using a UAMI if there are multiple identities attached to the same VM

cameronmeissner: in this case it's the identity we attach to the pool agents

cameronmeissner, line 89: so this isn't working in the windows case?

zachary-bailey: I don't believe the windows AzCopy is a current enough version. I am running some pipelines right now to double check and see a couple different methods. Back when I was working on the shallow replication PR, this was the only way I could get things to work without the container. For some reason, I can not seem to get auto-login to work with just the agent even though the resource string is still passed in.

zachary-bailey: Thanks for the explanation!

zachary-bailey: So seems that I need to reach out to the windows team to check on their pipeline, I have to set SIG_FOR_PRODUCTION to true in order for the convert script to run. Would like to do speak with them before doing so to make sure I don't inadvertently cause issues @cameronmeissner

zachary-bailey: I did not leave the TODO comment by the way regarding --login-type=MSI, that was Chou Hu

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -83,13 +83,18 @@ echo "Granting access to $disk_resource_id for 1 hour"
 # shellcheck disable=SC2102
 sas=$(az disk grant-access --ids $disk_resource_id --duration-in-seconds 3600 --query [accessSas] -o tsv)
 
-# TBD: Need to investigate why `azcopy-preview login --login-type=MSI` does not work
+echo "Uploading $disk_resource_id to ${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"
+
 echo "Setting azcopy environment variables with pool identity: $AZURE_MSI_RESOURCE_STRING"
+# TBD: Need to investigate why `azcopy-preview login --login-type=MSI` does not work for Windows
```
cameronmeissner, line 89: @AbelHu I think this is most likely due to the fact that you're using `azcopy-preview`, not completely sure though

cameronmeissner (approved):

PR #5014, opened 2024-09-30
chore: Remove internal link. Update Renovate created branch with prefix to a…

Devinwong:
**What type of PR is this?**
/kind chore

- Remove internal link from readme.
- Update Renovate created branch with prefix to avoid branch collision.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -25,12 +25,38 @@
       "enabled": false
     },
     {
-      "matchDatasources": ["docker", "custom.deb1804", "custom.deb2004", "custom.deb2204", "custom.deb2404"],
+      "matchDatasources": ["docker", "custom.deb1804"],
       "matchUpdateTypes": [
-        "patch",
-        "pin",
-        "digest"
+        "patch"
       ],
+      "additionalBranchPrefix": "deb1804-",
+      "automerge": false,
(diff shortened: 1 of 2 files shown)


.github/renovate.json
```diff
@@ -25,12 +25,46 @@
       "enabled": false
     },
     {
-      "matchDatasources": ["docker", "custom.deb1804", "custom.deb2004", "custom.deb2204", "custom.deb2404"],
+      "matchDatasources": ["docker"],
```
Devinwong, line 28: Splitting the original single block into different because of a branch collision issue.
Before this change, updating containerd for Ubuntu 20.04 and 22.04 will use the same branch name. And the latter one will run into branch collision issue. That's why we only saw auto-update PR for 20.04 but not 22.04. Here we add `additionalBranchPrefix` for each ubuntu release so the branch name will become renovate/[custom datasource name]-moby-containerd-1.7.x to avoid collision. Tested in my fork. Now we can see both 20.04 and 22.04 PRs will be created.

Devinwong: image

djsly (approved):

PR #5015, opened 2024-10-01
test: add benchmarking for test, scan, cleanup, and convert to sig steps

zachary-bailey:
**What type of PR is this?**

/kind test

This PR adds benchmarking to additional scripts in preparation for merging the [regression detection program.](https://github.com/Azure/AgentBaker/pull/4973)

Unsatisfactory VHD Build pipeline duration.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**

Another CSE file was created in order to prevent having to source cse_helpers.sh into the test/scan/cleanup steps as it would cause them to fail.


vhdbuilder/packer/build-performance/build-performance-funcs.sh
```diff
@@ -0,0 +1,61 @@
...
+
+script_start_stopwatch=$(date +%s)
+section_start_stopwatch=$(date +%s)
+SCRIPT_NAME=$(basename $0 .sh)
+SCRIPT_NAME="${SCRIPT_NAME//-/_}"
+declare -A benchmarks=()
+declare -a benchmarks_order=()
+VHD_BUILD_PERF_DATA=${BUILD_PERF_DATA_FILE}
+
+check_array_size() {
```
cameronmeissner, line 11: we already define these here: https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_helpers.sh#L500-L548

could we upload this file to the packer VM and source them within install-deps as well? that way we won't have to worry about updating code in multiple places

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -47,6 +49,7 @@ fi
 # Use $SIG_IMAGE_VERSION for the sig resource, and $CAPTURED_SIG_VERSION (a random number) for the disk resource
 sig_resource_id="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/galleries/${SIG_GALLERY_NAME}/images/${SIG_IMAGE_NAME}/versions/${SIG_IMAGE_VERSION}"
 disk_resource_id="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP_NAME}/providers/Microsoft.Compute/disks/${CAPTURED_SIG_VERSION}"
+capture_benchmark "set_variables"
```
cameronmeissner, line 52: could we prefix the event names with the name of the file? e.g. `convert-sig-to-classic-storage-account/set_variables`?

vhdbuilder/packer/build-performance/build-performance-funcs.sh
zachary-bailey: Do you mean remove these functions from `cse_helpers.sh` and source them to the install scripts from this file only? I like that idea and was initially going in that direction. 

The only reason I ended up not wanting to remove them from `cse_helpers.sh` was because I thought they could be useful for node bootstrapping benchmarking. But looking around yesterday, it seems we already have some infrastructure for logging during the bootstrapping process. 

Unfortunately, sourcing `cse_helpers.sh` to these files (convert, test, scan, etc) fails the step because of the variable assignments, so its a bit of a dilemma. How would you like me to proceed? @cameronmeissner

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
zachary-bailey: Would we be doing that for readability in the code or for the database schema? 

Currently each script is a key-value pair in the [JSON file](https://github.com/Azure/AgentBaker/blob/zb/regression2/vhdbuilder/packer/buildperformance/testdata/test-build-performance-data.json), with the key being the script name and the value being a map of string[string] (event name/timestamp). When the data is analyzed, the JSON collapses and each script name is appended to each event name that it contains. So when we're working with the data in Kusto, this would be `convert_sig_to_classic_storage_account_blob_set_variables`. 

Does that mitigate any concerns? I have a some regex expressions in the query for prefixes/suffixes so if possible it'd be nice to keep it this way.

Either way I can make this title a bit more expressive though, I understand it seems ambiguous by itself. @cameronmeissner

vhdbuilder/packer/build-performance/build-performance-funcs.sh
djsly: should we decouple the variable assignment and the helper functions into two files ? this way one can source the helper functions only ?

djsly: I agree though this is some pretty bad code duplication :(

vhdbuilder/packer/cleanup.sh
```diff
@@ -21,13 +23,15 @@ if [[ -n "$PKR_RG_NAME" ]]; then
     echo "Packer resource group already successfully deleted"
   fi
 fi
+capture_benchmark "delete_packers_rg"
```
djsly, line 26: could we look at augmenting the nomenclature to maybe keep the filename in the banchmaarks steps ? 

e.g. `cleanup_sh-delete_packer_rg`

vhdbuilder/packer/cleanup.sh
```diff
@@ -210,5 +217,8 @@ if [ -n "$old_storage_accounts" ]; then
 else
   echo "did not find any old storage accounts eligible for deletion"
 fi
+capture_benchmark "cleanup_storage_accounts"
 
-echo -e "Packer cleanup successfully completed\n\n\n"
\ No newline at end of file
+echo -e "Packer cleanup successfully completed\n\n\n"
+capture_benchmark "${SCRIPT_NAME}_overall" true
```
djsly, line 223: I think we should use `${SCRIPT_NAME}_` everywhere

djsly (changes requested):

vhdbuilder/packer/build-performance/build-performance-funcs.sh
zachary-bailey: Is it possible to just source the functions into cse_helpers from this new file...? That way they are available to scripts that run after the packer vm is deprovisioned, and still available in cse_helpers for node bootstrap benchmarking later on.

@djsly @cameronmeissner

vhdbuilder/packer/cleanup.sh
zachary-bailey: Made the change, running pipeline now

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
zachary-bailey: Added them in

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -482,71 +482,6 @@ isMarinerOrAzureLinux() {
     return 1
 }
 
-installJq() {
```
djsly, line 485: isnt this used anywhere else ?

parts/linux/cloud-init/artifacts/cse_benchmark_functions.sh
```diff
@@ -0,0 +1,75 @@
...
+  declare -n array_name=$1
+  local array_size=${#array_name[@]}
+  if [[ ${array_size} -gt 0 ]]; then
+    last_index=$(( ${#array_name[@]} - 1 ))
+  else
+    return 1
+  fi
+}
+
+installJq() {
```
djsly, line 20: this is only used in `/vhdbuilder/packer/pre-install-dependencies.sh` we should keep it in `cse_helper.sh` I would say

djsly (changes requested):

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: Note for future: installJq function used to be here but was moved to `cse_helpers.sh`, evalPackageDownloadURL moved up after removal

djsly (approved):

PR #5022, opened 2024-10-03
publishing testdata after successful master build

djsly:
**What type of PR is this?**
/kind feature

Prereq, to remove testdata from the build. we will be publishing the testdata on master build pipeline

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -104,3 +104,18 @@ stages:
...
+    jobs:
+      - job: Generate_Publish_Testdata
+        timeoutInMinutes: 10
+        steps:
+        - bash: |
+            make generate
+          displayName: Generate Testdata
+        - task: 1ES.PublishPipelineArtifact@1
+          inputs:
+            targetPath: $(Pipeline.Workspace)/pkg/agent/testdata
```
cameronmeissner, line 118: how would we consume this from unit test workflows? iterate through the most recent runs and pick out the latest one from master?

I was also originally thinking we'd have a new workflow that would only trigger on merge commits to the master branch which would actually upload/update the testdata, as this stands we'd upload testdata as an artifact to every VHD check-in build

djsly: it wont be a workflow, I'm working on a new ado pipeline.

djsly: what are checkins build ?

cameronmeissner: VHD check-in builds, which is defined by `.vsts-vhd-builder.yaml`

djsly: I"m not sure where else we should upload the testdata ? why not on the master build after things gets merged ?

cameronmeissner: I think builds off the master branch are fine to upload from, though with this change we'd be uploading artifacts to every instance of the checkin build

cameronmeissner: That's also why I was wondering how we'd pick out the correct artifacts from the new pipeline

cameronmeissner: Ah okay, I guess V2 of DownloadPipelineArtifact will default to latest from master branch: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/download-pipeline-artifact-v2?view=azure-pipelines

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -104,3 +104,18 @@ stages:
...
       - template: ./templates/e2e-template.yaml
+  - stage: Generate and Publish Testdata
+    condition: succeeded()
+    jobs:
+      - job: Generate_Publish_Testdata
+        timeoutInMinutes: 10
+        steps:
+        - bash: |
+            make generate
+          displayName: Generate Testdata
```
cameronmeissner, line 115: I think this is already the case, but probably worth double checking that the checkin build will be automatically triggered whenever we end up changing template content

cameronmeissner (approved):
will follow-up with a subsequent PR to tune triggers of the new behavior to master branch builds only

.pipelines/.vsts-vhd-builder.yaml
djsly: correct, you can also select the version. but I was thinking of pulling latest from Master all the time

PR #5029, opened 2024-10-04
feat: add systemd unit for scriptless

lilypan26:
**What type of PR is this?**

This PR adds a systemd unit to trigger bootstrapping for scriptless VHDs. The systemd unit waits for cloud-init completion (which ensure customdata files, in this case the boostrapping contract in json format, are written to disk) and then runs the go binary which takes the contract to generate cse_cmd to kickstart the bootstrapping process.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


lilypan26:
Successful e2e run: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=104976927&view=results

Corresponding build: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=104963277&view=results

e2e/opts.go
```diff
@@ -5,7 +5,8 @@ import (
 )
 
 type scenarioRunOpts struct {
-	clusterConfig *Cluster
-	scenario      *Scenario
-	nbc           *datamodel.NodeBootstrappingConfiguration
+	clusterConfig      *Cluster
+	scenario           *Scenario
+	nbc                *datamodel.NodeBootstrappingConfiguration
+	isSelfContainedVHD bool
```
cameronmeissner, line 11: nit: `scriptless bool`

e2e/scenario_test.go
```diff
@@ -636,6 +636,31 @@ func Test_ubuntu2204(t *testing.T) {
 	})
 }
 
+func Test_ubuntu2204Scriptless(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "Tests that a node using the Ubuntu 2204 VHD can be properly bootstrapped",
```
cameronmeissner, line 641: can we update the description to reference that it's using scriptless?

e2e/vmss.go
```diff
@@ -232,6 +249,94 @@ func getVmssName(t *testing.T) string {
...
 
+func getBaseVMSSModelSelfContained(name, sshPublicKey string, opts *scenarioRunOpts) armcompute.VirtualMachineScaleSet {
+	nbc, err := json.Marshal(baseNodeBootstrappingContract(config.Config.Location, opts))
+	if err != nil {
+		return armcompute.VirtualMachineScaleSet{}
+	}
+
+	nbcEntry := `#cloud-config
+
+write_files:
```
cameronmeissner, line 260: can we move this to `templates.go`?

e2e/vmss.go
```diff
@@ -232,6 +249,94 @@ func getVmssName(t *testing.T) string {
 	return name
 }
 
+func getBaseVMSSModelSelfContained(name, sshPublicKey string, opts *scenarioRunOpts) armcompute.VirtualMachineScaleSet {
+	nbc, err := json.Marshal(baseNodeBootstrappingContract(config.Config.Location, opts))
+	if err != nil {
+		return armcompute.VirtualMachineScaleSet{}
```
cameronmeissner, line 255: let's log the error at least

e2e/vmss.go
```diff
@@ -232,6 +249,94 @@ func getVmssName(t *testing.T) string {
...
+		Properties: &armcompute.VirtualMachineScaleSetProperties{
+			Overprovision: to.Ptr(false),
+			UpgradePolicy: &armcompute.UpgradePolicy{
+				Mode: to.Ptr(armcompute.UpgradeModeManual),
+			},
+			VirtualMachineProfile: &armcompute.VirtualMachineScaleSetVMProfile{
+				OSProfile: &armcompute.VirtualMachineScaleSetOSProfile{
+					ComputerNamePrefix: to.Ptr(name),
+					AdminUsername:      to.Ptr("azureuser"),
+					CustomData:         &encodedCustomData,
```
cameronmeissner, line 283: so I guess only customdata?

nbcparser/main.go
```diff
@@ -1,40 +1,36 @@
...
-// to build: go build main.go.
-// to run: ./main testdata/test_nbc.json.
 //
 //nolint:gosec // generated cse_cmd.sh file needs execute permissions for bootstrapping
+const (
+	CSE_CMD = "cse_cmd.sh"
+)
+
 func main() {
 	if len(os.Args) < parser.MinArgs {
```
cameronmeissner, line 20: nit(non-blocking): would prefer we use stdlib `flags` or `cobra` for flags rather than raw argv, can do that in another PR though

nbcparser/main.go
```diff
@@ -1,40 +1,36 @@
...
-		log.Fatal(err)
-		return
-	}
-	if err = os.WriteFile("cse_cmd.sh", []byte(cseCmd), 0655); err != nil {
-		log.Fatal(err)
-		return
+		fmt.Println(err)
+		os.Exit(1)
 	}
+	fmt.Println(cseCmd)
```
cameronmeissner, line 35: do we still need this?

nbcparser/main.go
```diff
@@ -1,40 +1,36 @@
 package main
 
 import (
-	"log"
```
cameronmeissner, line 4: did you want to actually switch over to `fmt`? personally I'd prefer sticking with `log`

nbcparser/pkg/parser/templates/cse_cmd.sh.gtpl
```diff
@@ -148,4 +148,7 @@ IS_KATA="{{.GetIsKata}}"
 ARTIFACT_STREAMING_ENABLED="{{.GetEnableArtifactStreaming}}"
 SYSCTL_CONTENT="{{getSysctlContent .CustomLinuxOsConfig.GetSysctlConfig}}"
 PRIVATE_EGRESS_PROXY_ADDRESS=""
+BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER=""
+ENABLE_IMDS_RESTRICTION="false"
```
cameronmeissner, line 152: non-blocking: we should add some automation/PR-time checks to make sure this is in-sync with what we have in cse_cmd.sh

parts/linux/cloud-init/artifacts/nbcparser.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -x
+# TODO: propagate exit status to bootstrapper
+if [ ! -f /opt/azure/containers/nbc.json ]; then 
+    exit 1
+fi
+/opt/azure/containers/nbcparser /opt/azure/containers/nbc.json > cse_cmd.sh
```
cameronmeissner, line 7: `/opt/azure/containers/nbcparser` will produce logs to stdout/err right? should we pipe that output to a dedicated log file? maybe something like `/var/log/azure/aks/bootstrap.log`

parts/linux/cloud-init/artifacts/nbcparser.target
cameronmeissner, line 1: nit: `bootstrap.target`/`provision.target`?

cameronmeissner: or maybe even `cloud-init-complete.target` at that point to make things ultra clear

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -80,6 +80,10 @@ capture_benchmark "enable_modified_log_rotate_service"
 systemctlEnableAndStart sync-container-logs.service || exit 1
 capture_benchmark "sync_container_logs"
 
+# enable and start nbcparser.service
+systemctl enable nbcparser.target
```
cameronmeissner, line 84: should we just be consistent and use `systemctlEnableAndStart` like we do for other units? or did we decide that we only wanted to enable here?

parts/linux/cloud-init/artifacts/nbcparser.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -x
+# TODO: propagate exit status to bootstrapper
```
cameronmeissner, line 3: I think we will eventually have a boiler-plate CSE script which would run a script to poll the status of this unit and return an exit code based on the corresponding status - that'll be the only way we can get exit code information back to the caller I believe

e2e/vmss.go
lilypan26: yep, will add the cse script to poll bootstrap status in separate pr

nbcparser/main.go
lilypan26: modified to use `flag`

nbcparser/main.go
lilypan26: switched back to log

nbcparser/main.go
lilypan26: Yes, I was printing the generate cse_cmd to stdout and then piping it into a file to be executed in the systemd unit. However, I decided to just trigger cse_cmd from the go binary itself, refer to recent commits.

vhdbuilder/packer/pre-install-dependencies.sh
lilypan26: we just want to enable, the new service will automatically be started once cloud-init is complete

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -960,14 +960,70 @@ testNBCParserBinary () {
...
+    return 1
+  fi
+  echo "$test: nbcparser.sh exists at $script_path"
+
+  # is-enabled returns:
+  # 'enabled' if the service is enabled.
+  # empty string if the service is not installed.
+  # 'not-found' if the unit files are not present. Encountered with Ubuntu 24.04
+  echo "$test: Checking that $service_name is enabled"
+  local is_enabled=
```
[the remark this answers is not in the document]
lilypan26: hmmm doesn't like that: 
image

parts/linux/cloud-init/artifacts/nbcparser.sh
lilypan26: Yep, will add this in the next PR

e2e/scenario_test.go
```diff
@@ -649,6 +649,31 @@ func Test_ubuntu2204(t *testing.T) {
 	})
 }
 
+func Test_ubuntu2204Scriptless(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "Tests that a node using the Ubuntu 2204 Scriptless VHD can be properly bootstrapped",
+		Config: Config{
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDUbuntu2204Gen2Containerd,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
```
r2k1, line 658: Does it do anything? It looks like `*datamodel.NodeBootstrappingConfiguration` isn't used in your changes.

nbcparser/pkg/parser/templates/cse_cmd.sh.gtpl
lilypan26: Yes good point, will add this as a followup task

parts/linux/cloud-init/artifacts/nbcparser.sh
lilypan26: added!

e2e/scenario_test.go
lilypan26: I use the existing nbc to populate the new contract in `baseNodeBootstrappingContract` in `templates.go`. However,  I think you're right that these overrides aren't doing anything.

nbcparser/main.go
```diff
@@ -1,40 +1,55 @@
...
+	test := flag.Bool("test", false, "test mode")
+	flag.Parse()
+	if *configFile == "" {
+		log.Fatal("bootstrap-config is a required argument")
 	}
 
-	// Read in the JSON file
-	inputJSON, err := os.ReadFile(os.Args[1])
+	cseCmd := parseConfig(*configFile)
+	err := bootstrap(cseCmd, *test)
```
Devinwong, line 24: where do we actually call the bootstrap() with test=true?

nbcparser/main.go
```diff
@@ -1,40 +1,55 @@
...
 	if err != nil {
 		log.Fatal(err)
-		return
 	}
+	log.Println("Provisioning completed successfully")
```
Devinwong, line 28: I could be wrong. Is it more accurate to say Provisioning request initiated(received)? We don't know if it's successful or failed until the scripts really complete, right?

e2e/scenario_test.go
```diff
@@ -649,6 +649,31 @@ func Test_ubuntu2204(t *testing.T) {
...
+				nbc.ContainerService.Properties.CertificateProfile.ClientPrivateKey = "client cert private key"
+				nbc.ContainerService.Properties.ServicePrincipalProfile.Secret = "SP secret"
+			},
+			LiveVMValidators: []*LiveVMValidator{
+				mobyComponentVersionValidator("containerd", getExpectedPackageVersions("containerd", "ubuntu", "r2204")[0], "apt"),
+				mobyComponentVersionValidator("runc", getExpectedPackageVersions("runc", "ubuntu", "r2204")[0], "apt"),
+			},
+		},
+		Tags: Tags{
+			Scriptless: true,
```
Devinwong, line 672: Does that mean by setting this tag `Scriptless`, client like aks-rp can determine sending scriptless contract or the old CSE scripts at runtime? Or this tag is only for e2e test?

e2e/opts.go
```diff
@@ -8,4 +8,5 @@ type scenarioRunOpts struct {
 	clusterConfig *Cluster
 	scenario      *Scenario
 	nbc           *datamodel.NodeBootstrappingConfiguration
+	scriptless    bool
```
anujmaheshwari1, line 11: super duper nit: this feels somewhat incomplete to me, `scriptlessVHD` maybe makes more sense to properly highlight what the scenario option is.

taking the cue from clusterConfig instead of just Config. feel free to ignore, just my opinion.

nbcparser/main.go
lilypan26: in `linux-vhd-content-test.sh` we invoke the binary to test that it runs but don't want to actually run cse

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -1,5 +1,7 @@
 CSE_STARTTIME=$(date)
 CSE_STARTTIME_FORMATTED=$(date +"%F %T.%3N")
+EVENTS_LOGGING_DIR=/var/log/azure/Microsoft.Azure.Extensions.CustomScript/events/
```
anujmaheshwari1, line 3: did we ever determine from Chase if this was actually created by cse and if creating this manually had any implications?

parts/linux/cloud-init/artifacts/bootstrap.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -x
```
anujmaheshwari1, line 2: nit: set -euxo pipefail

parts/linux/cloud-init/artifacts/cse_start.sh
lilypan26: no chase said there should be nothing wrong with creating the directory manually, and I confirmed that I am still able to see the cse status in kusto with scriptless vhds.

parts/linux/cloud-init/artifacts/cloud-init-complete.target
```diff
@@ -0,0 +1,5 @@
+[Unit]
+Description=Custom Target for Cloud Init Completion
```
anujmaheshwari1, line 2: we dont really need this, you can do a `cloud-init status --wait` to ensure that cloud-init has finished before running anything else

e2e/scenario_test.go
lilypan26: No, this is just for e2e testing

nbcparser/main.go
Devinwong: Could you point me where it is? 
The reason I was asking this is to understand if we really need this variable for test purpose (in the meantime it will increase chances for the parser to run in test mode unintentionally, causing provisioning failure in production), or we can simply mock a test function in the test codes.
But if it's really necessary, we still can go with it. No preference now.

e2e/opts.go
cameronmeissner: lol `scriptless` was my original suggestion, I think `scriptlessVHD` is fine too

nbcparser/main.go
lilypan26: Right, im just going to remove the print statement altogether since it doesn't tell us much. The cse script to poll the status of the systemd unit will provide information on whether bootstrapping completed successfully.

e2e/opts.go
lilypan26: i think i prefer simplicity of `scriptless` and we apparently we might rebrand again XD so just going to keep as is

parts/linux/cloud-init/artifacts/cloud-init-complete.target
lilypan26: was able to get it working without this target by using systemd dependencies with cloud-init.target and cloud-final.service :)

nbcparser/main.go
Devinwong: That makes sense. Thanks for clarifying.

Devinwong (approved):
LGTM. Thanks for the persistence for this feature. "No pressure" 😆

anujmaheshwari1 (approved):

PR #5030, opened 2024-10-05
chore: remove mcr.microsoft.com from allowed network isolated e2es

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


AlisonB319:
still seeing 

`failed to resolve reference \"mcr.microsoft.com/oss/kubernetes/kube-proxy:v1.29.8\":`
`failed to pull and unpack image \"mcr.microsoft.com/azuredefender/stable/old-file-cleaner:1.0.148`
`failed to pull and unpack image \"mcr.microsoft.com/azuredefender/stable/pod-collector:1.0.110`
`failed to pull and unpack image \"mcr.microsoft.com/azuredefender/stable/low-level-collector:2.0.99`

even though the e2e is now passing. I wonder if these will get cleaned up once I figure out where "mcr.microsoft.com/oss/kubernetes/kube-proxy:v1.29.8 is getting passed into the e2e tests

AlisonB319:
Looking at kubelet output from a non-airgap test, I see 

`image(mcr.microsoft.com/oss/kubernetes/kube-proxy) is not from ACR, return empty authentication`
`image(mcr.microsoft.com/azuredefender/stable/old-file-cleaner) is not from ACR, return empty authentication`
`image(mcr.microsoft.com/azuredefender/stable/pod-collector) is not from ACR, return empty authentication`
`image(mcr.microsoft.com/azuredefender/stable/low-level-collector) is not from ACR, return empty authentication`

so it might actually be okay that these images are not able to be grabbed - still unsure how `mcr.microsoft.com/oss/kubernetes/kube-proxy:v1.29.8` is being passed into the kubelet tests since overwriting `CustomKubeProxyImage` in `template.go` does not do anything

e2e/kube.go
```diff
@@ -76,19 +76,24 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
...
-	nonHostDS := getDebugDaemonsetTemplate(podNetworkDebugAppLabel, "nodepool2", false)
+	nonHostDS := getDebugDaemonsetTemplate(podNetworkDebugAppLabel, "nodepool2", false, isAirgap)
 	if err := createDebugDaemonset(ctx, kube, nonHostDS); err != nil {
 		return err
 	}
 	return nil
 }
 
-func getDebugDaemonsetTemplate(deploymentName, targetNodeLabel string, isHostNetwork bool) string {
+func getDebugDaemonsetTemplate(deploymentName, targetNodeLabel string, isHostNetwork, isAirgap bool) string {
```
cameronmeissner, line 91: I forget, do we create separate e2e clusters just for airgap scenarios?

im just wondeirng how this would play with other scenarios since every scenario ends up using this

AlisonB319: Yep we do create separate clusters for e2e. It should still use the mcr path for non air gap tests

AlisonB319: If they were the same clusters, they would be subjugated to the airgap rules

e2e/aks_model.go
```diff
@@ -150,7 +150,14 @@ func airGapSecurityGroup(location, clusterFQDN string) (armnetwork.SecurityGroup
 func addPrivateEndpointForACR(ctx context.Context, t *testing.T, nodeResourceGroup string, vnet VNet) error {
 	t.Logf("Adding private endpoint for ACR in rg %s\n", nodeResourceGroup)
 
-	peResp, err := createPrivateEndpoint(ctx, t, nodeResourceGroup, vnet)
+	privateEndpointName := "PE-for-ABE2ETests"
+	if exists, err := privateEndpointExists(ctx, t, nodeResourceGroup, privateEndpointName); err != nil {
```
cameronmeissner, line 154:
```suggestion
	if exists, err := privateEndpointExists(ctx, t, nodeResourceGroup, privateEndpointName); err != nil {
	    return err
	}
	if exists {
	    return nil
	}
```

e2e/aks_model.go
```diff
@@ -179,8 +186,19 @@ func addPrivateEndpointForACR(ctx context.Context, t *testing.T, nodeResourceGro
...
 
-func createPrivateEndpoint(ctx context.Context, t *testing.T, nodeResourceGroup string, vnet VNet) (armnetwork.PrivateEndpointsClientCreateOrUpdateResponse, error) {
-	endpointName := "PE-for-ABE2ETests"
+func privateEndpointExists(ctx context.Context, t *testing.T, nodeResourceGroup, privateEndpointName string) (bool, error) {
+	existingPE, err := config.Azure.PrivateEndpointClient.Get(ctx, nodeResourceGroup, privateEndpointName, nil)
+	if err == nil && existingPE.ID != nil {
+		t.Logf("Private Endpoint already exists with ID: %s\n", *existingPE.ID)
+		return true, nil
+	}
+	if err != nil && !strings.Contains(err.Error(), "ResourceNotFound") {
```
cameronmeissner, line 195: right now we're silencing the error in cases where it doesn't match this pattern, we should probably either always return it or at least always log it in any case if it's non-nil

cameronmeissner (approved):

e2e/aks_model.go
AlisonB319: this doesn't make the ide happy with 
exists declared and not used

can go with 
exists, err := privateEndpointExists(ctx, t, nodeResourceGroup, privateEndpointName)
	if err != nil {
		return err
	}
	if exists {
		t.Logf("Private Endpoint already exists, skipping creation")
		return nil
	}
	
if its about readability

PR #5042, opened 2024-10-09
test: publish bcc installation logs

zachary-bailey:
**What type of PR is this?**

/kind test

This PR publishes bcc install logs after the completion of the VHD build in order to gain visibility into the installation process after failed downloads.

Lack of visibility into bcc tools installation process. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -164,6 +164,21 @@ steps:
     condition: always()
     displayName: Check Build Performance
 
+  - task: PublishPipelineArtifact@0
+    condition: always()
+    displayName: Publish BCC Tools Installation Log
+    inputs:
+      artifactName: 'bcc-tools-installation-log-${{ parameters.artifactName }}'
+      targetPath: 'bcc-tools-installation.log'
+
+  - task: CopyFiles@2
+    condition: always()
+    displayName: Copy BCC Tools Installation Log
+    inputs:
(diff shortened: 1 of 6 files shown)


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -164,6 +164,21 @@ steps:
     condition: always()
     displayName: Check Build Performance
 
+  - task: PublishPipelineArtifact@0
```
djsly, line 167: aint those two tasks inversed ? don' ty ou need to copy and publish after ? Maybe I' dont have the full logic :)

djsly (approved):

.pipelines/templates/.builder-release-template.yaml
zachary-bailey: They are not dependent on one another. The copy task puts it in the staging directory to be published with the bulk artifacts folder at the end of the pipeline, so its convenient to find most of the time. 

But that bulk artifacts folder is only published if the build succeeds, so the publish task before it is necessary in order to make sure its published even if the build fails, since thats when we're most interested in it. @djsly

cameronmeissner: this is an artifact of that fact that we actually publish every artifact twice, after we abandoned a previous change since it required a larger refactor, we can clean this up

cameronmeissner (approved):

PR #5046, opened 2024-10-09
Pull aks-gpu image with k8s.io namespace

ganeshkumarashok:
**What type of PR is this?**
/kind cleanup
/kind fix

This pulls the AKS-GPU image in the k8s.io namespace (instead of default), so that the image can be garbage collected by kubelet and eraser, when not needed.
 
Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


UtheMan (approved):

UtheMan:
lgtm

djsly (approved):

djsly (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -744,7 +744,7 @@ configGPUDrivers() {
     if [[ $OS == $UBUNTU_OS_NAME ]]; then
         mkdir -p /opt/{actions,gpu}
         if [[ "${CONTAINER_RUNTIME}" == "containerd" ]]; then
-            ctr image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG
+            ctr -n k8s.io image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG
```
djsly, line 747: now that you pull it under the k8s.io namespace, when you use the image to install the driver, do you need to reference that namespace ?

ganeshkumarashok: Good call - I've updated the run commands as well. Surprisingly, the Agentbaker e2es for GPUs didn't fail before that. I wonder why.

PR #5069, opened 2024-10-11
chore: upgrade blob csi driver version in vhd image

andyzhangx:
**What type of PR is this?**
/kind feature

chore: upgrade blob csi driver version in vhd image

we have a bug in `v1.23.8`, `v1.24.4`, these two versions are actually not used in aks-rp, so this PR replaces those two version directly.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


AbelHu (approved):
LGTM. Thanks for the PR

PR #5075, opened 2024-10-11
fix: exit when bcc install fails

zachary-bailey:
**What type of PR is this?**

/kind bug

This PR will fail the build if BCC tools installation fails in order to address errors quicker.

BCC install failures were only caught during the testing phase of the VHD build. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -552,6 +552,7 @@ if [ $BCC_EXIT_CODE -eq 0 ]; then
 EOF
 else
   echo "Error: installBcc subshell failed with exit code $BCC_EXIT_CODE" >&2
+  exit $BCC_EXIT_CODE
 fi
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -552,6 +552,7 @@ if [ $BCC_EXIT_CODE -eq 0 ]; then
 EOF
 else
   echo "Error: installBcc subshell failed with exit code $BCC_EXIT_CODE" >&2
+  exit $BCC_EXIT_CODE
 fi
 chmod 755 /var/log/bcc_installation.log
```
djsly, line 557: wont we loose the access to this file  ? I though you added that chmod to allow the publishing of the log to ADO

zachary-bailey: good catch! fixed

djsly (approved):

PR #5077, opened 2024-10-11
using discard_unpacked_layers with CRI to reduce the size on disk of the cached images

djsly:
**What type of PR is this?**

/kind cleanup

Reducing the Packed Image layers (content) under /var/lib/containerd folder, this data is not used normally for creating container, it's used for republishing.  CRICTL allows for this storage to be dleeted post unpacking, using the CRI flag  

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_install.sh
+++ b/parts/linux/cloud-init/artifacts/cse_install.sh
@@ -388,7 +388,7 @@ installCrictl() {
     CPU_ARCH=$(getCPUArch)
     currentVersion=$(crictl --version 2>/dev/null | sed 's/crictl version //g')
     if [[ "${currentVersion}" != "" ]]; then
-        echo "version ${currentVersion} of crictl already installed. skipping installCrictl of target version ${KUBERNETES_VERSION%.*}.0"
+        echo "version ${currentVersion} of crictl already installed. skipping installCrictl of target version ${1%.*}.0"
     else
         # this is only called during cse. VHDs should have crictl binaries pre-cached so no need to download.
         # if the vhd does not have crictl pre-baked, return early
@@ -399,7 +399,7 @@ installCrictl() {
             return 1
         fi
         echo "Unpacking crictl into ${CRICTL_BIN_DIR}"
-        tar zxvf "$CRICTL_DOWNLOAD_DIR/${CRICTL_TGZ_TEMP}" -C ${CRICTL_BIN_DIR}
+        tar zxvf "$CRICTL_DOWNLOAD_DIR/${CRICTL_TGZ_TEMP}" -C ${CRICTL_BIN_DIR} || exit $ERR_CRICTL_OPERATION_ERROR
         chown root:root $CRICTL_BIN_DIR/crictl
         chmod 755 $CRICTL_BIN_DIR/crictl
     fi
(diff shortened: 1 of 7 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -265,6 +266,8 @@ while IFS= read -r p; do
           installStandaloneContainerd "${version}"
         fi
         echo "  - containerd version ${version}" >> ${VHD_LOGS_FILEPATH}
+        # k8s will use images in the k8s.io namespaces - create it
+        ctr namespace create k8s.io
```
ganeshkumarashok, line 270: Do we still need the `ctr` command here? Thought it's all being replaced with `crictl`?

cameronmeissner: crictl is just calling ctr under the hood I believe, so I do think we still need this

djsly: I wasn't sure if crictl was default to the namespace K8s.io. so I kept it to double check if they would land in K8s.oi still

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -339,25 +342,40 @@ fi
...
 
-# k8s will use images in the k8s.io namespaces - create it
-ctr namespace create k8s.io
-cliTool="ctr"
+enable_containerd_discard_unpacked_layers() {
+  containerd_config_file="/etc/containerd/config.toml"
+  echo -e "[plugins."io.containerd.grpc.v1.cri".containerd]\ndiscard_unpacked_layers = true" >> ${containerd_config_file}
+}
 
+disable_containerd_discard_unpacked_layers() {
```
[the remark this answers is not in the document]
djsly: I'm reenabling it at the end. I will need to test. This is on my bucket list. to see what will happen. I guess we dont have an e2e for that right ?

vhdbuilder/packer/install-dependencies.sh
djsly: removed

parts/linux/cloud-init/artifacts/cgroup-memory-telemetry.sh
```diff
@@ -80,5 +80,5 @@ EVENT_JSON=$( jq -n \
     --arg EventTid      "0" \
     '{Timestamp: $Timestamp, OperationId: $OperationId, Version: $Version, TaskName: $TaskName, EventLevel: $EventLevel, Message: $Message, EventPid: $EventPid, EventTid: $EventTid}'
 )
-
+mkdir -p ${EVENTS_LOGGING_DIR}
```
cameronmeissner, line 83: we shouldn't need to do this, WALinuxAgent will do this before us before cloud-init/CSE is invoked

e2e/kubelet/generate-kubelet-flags.sh
```diff
@@ -5,7 +5,12 @@ retrycmd_get_tarball() {
     tar_retries=$1; wait_sleep=$2; tarball=$3; url=$4
     echo "${tar_retries} retries"
     for i in $(seq 1 $tar_retries); do
-        tar -tzf $tarball && break || \
+        if [ -f $tarball ]; then
```
cameronmeissner, line 8: is this an unrelated change?

vhdbuilder/lister/pkg/image/list.go
```diff
@@ -69,5 +69,5 @@ func ListImages(sku, version string) (*List, error) {
 }
 
 func isID(imageName string) bool {
-	return strings.Contains(imageName, "sha256")
+	return (strings.Contains(imageName, "sha256") && !strings.HasPrefix(imageName, "sha256:"))
```
cameronmeissner, line 72:
```suggestion
	return strings.Contains(imageName, "sha256") && !strings.HasPrefix(imageName, "sha256:")
```

ganeshkumarashok:
There are merge conflicts because of the other CSE changes - we need to rebase again

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -221,7 +226,12 @@ retrycmd_get_tarball_from_registry_with_oras() {
     tar_folder=$(dirname "$tarball")
     echo "${tar_retries} retries"
     for i in $(seq 1 $tar_retries); do
-        tar -tzf $tarball && break || \
+        if [ -f $tarball ]; then
+            tar -tzf $tarball
+            if [ $? -eq 0 ]; then 
+                break
+            fi
```
ganeshkumarashok, line 233: Why is this change necessary?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -385,21 +385,22 @@ downloadCrictl() {
...
     CPU_ARCH=$(getCPUArch)
     currentVersion=$(crictl --version 2>/dev/null | sed 's/crictl version //g')
     if [[ "${currentVersion}" != "" ]]; then
-        echo "version ${currentVersion} of crictl already installed. skipping installCrictl of target version ${KUBERNETES_VERSION%.*}.0"
+        echo "version ${currentVersion} of crictl already installed. skipping installCrictl of target version ${crictlVersion%.*}.0"
     else
         # this is only called during cse. VHDs should have crictl binaries pre-cached so no need to download.
         # if the vhd does not have crictl pre-baked, return early
-        CRICTL_TGZ_TEMP="crictl-v${CRICTL_VERSION}-linux-${CPU_ARCH}.tar.gz"
+        CRICTL_TGZ_TEMP="crictl-v${crictlVersion}-linux-${CPU_ARCH}.tar.gz"
```
ganeshkumarashok, line 396: is crictlVersion being set properly? I don't see it being passed to installCrictl during CSE but maybe I'm missing that

PR #5081, opened 2024-10-13
feat: use bicep to create the ACR for e2e tests

AlisonB319:
**What type of PR is this?**

The ACR used for our e2e tests cannot be locked since its lock will prevent the deletion of the private endpoint which does need to be deleted during routine garbage collection for our e2e resource group. This PR creates a new ACR in our e2e resource group using bicep. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/azure-resources/azure-container-registry-airgap.json
```diff
@@ -0,0 +1,48 @@
+{
```
AlisonB319, line 1: The json file gets auto-generated during createAzureResourceWithBicep

e2e/aks_model.go
```diff
@@ -428,3 +458,57 @@ func updateSubnet(ctx context.Context, cluster *armcontainerservice.ManagedClust
...
+
+func createAzureResourceWithBicep(ctx context.Context, t *testing.T, params map[string]interface{}, resourceGroup, filename, deploymentName string) error {
+	currentDir, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("failed to get current working directory: %v", err)
+	}
+
+	bicepFileName := fmt.Sprintf("%s.bicep", filename)
+	bicepFilePath := filepath.Join(currentDir, "azure-resources", bicepFileName)
+	cmd := exec.Command("az", "bicep", "build", "--file", bicepFilePath)
```
cameronmeissner, line 470: this is building the bicep file and then generating an ARM template?

e2e/azure-resources/azure-container-registry-airgap.json
cameronmeissner: could we consider just using an ARM template and getting rid of the bicep so we don't have to delegate out to `az bicep build` during every test run?

cameronmeissner: we might also consider deploying all the other airgap-related resources via an ARM/bicep template as well

e2e/aks_model.go
```diff
@@ -428,3 +458,57 @@ func updateSubnet(ctx context.Context, cluster *armcontainerservice.ManagedClust
...
+	if err := cmd.Run(); err != nil {
+		log.Fatalf("failed to run az bicep build: %v", err)
+	}
+
+	jsonFileName := fmt.Sprintf("%s.json", filename)
+	jsonFilePath := filepath.Join(currentDir, "azure-resources", jsonFileName)
+	if _, err := os.Stat(jsonFilePath); os.IsNotExist(err) {
+		log.Fatalf("JSON file was not created: %v", err)
+	}
+	jsonFile, err := os.ReadFile(jsonFilePath)
```
cameronmeissner, line 480: if we end up just using a static ARM template then I'd recommend using a `go:embed` to have access to the template content rather than manually reading the file

e2e/aks_model.go
cameronmeissner: sad: https://github.com/Azure/bicep/issues/12893

e2e/azure-resources/azure-container-registry-airgap.json
AlisonB319: Sylvain suggested using bicep so we should talk with him about ARM vs bicep. And I agree, I think a separate PR / task can be created for uniformly creating all the resources, depending on which direction we want to go in.

AlisonB319: I will say, I do think bicep files are a bit easier to create and read over the arm/json files. But it does come with the trade off of needing to run `az bicep build`

e2e/aks_model.go
r2k1: it doesn't matter for e2e, we don't compile a test binary and ship it.

Embedding is probably just going to add an extra layer.

e2e/aks_model.go
```diff
@@ -428,3 +458,57 @@ func updateSubnet(ctx context.Context, cluster *armcontainerservice.ManagedClust
 	}
 	return nil
 }
+
+func createAzureResourceWithBicep(ctx context.Context, t *testing.T, params map[string]interface{}, resourceGroup, filename, deploymentName string) error {
+	currentDir, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("failed to get current working directory: %v", err)
+	}
```
r2k1, line 466:
```suggestion
	require.NoError(t, err)
```

e2e/aks_model.go
r2k1: also, the file is probably not available yet at the moment you would embed it

e2e/aks_model.go
r2k1: This will probably need extra movement to make it work in ADO/Github

r2k1 (commented):
I wonder if creating an ACR using a single API call will be easier than using  az cli + bicep + extra code. We are already creating and deleting a bunch of Azure resources using SDK/API.

AlisonB319:
Got a green e2e test suite - https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=105621887&view=results

future e2e's will be red on airgap since the private endpoint is pointing to the old acr which this PR does not use. Once this PR is merged, open PRs will need to rebase

e2e/azure-resources/azure-container-registry-airgap.json
djsly: we should indeed only keep the .bicep file, and run `az bicep build` and I didn't expect that resources were all managed in Go, so it does make it look weird that we have go code to generate half (if not all) of the resource and the ACR is in Bicep... We should definitely look at unifying.  I do have a stronger opinion on the Bicep logic since we benefits from others (Azure Verified Module - AVM team) to augment the module with better security and compliance and we will simply need to bump the versions over time, VS having to update the Go package, SDK, find the new parameters, etc (last time I needed to set auto-upgrade on the AKS node pool, I had to do a major go module bump from v1 to v6..)

e2e/template.go
```diff
@@ -452,7 +452,7 @@ func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
 func getHTTPServerTemplate(podName, nodeName string, isAirgap bool) string {
 	image := "mcr.microsoft.com/cbl-mariner/busybox:2.0"
 	if isAirgap {
-		image = "aksvhdtestcr.azurecr.io/aks/cbl-mariner/busybox:2.0"
+		image = "privateacre2e.azurecr.io/aks/cbl-mariner/busybox:2.0"
```
djsly, line 455: could we use the same var in e2e/scenario_test.go ? var PrivateACRName string = "privateacre2e"

e2e/azure-resources/azure-container-registry-airgap.bicep
```diff
@@ -0,0 +1,27 @@
+@minLength(5)
+@maxLength(50)
+
+@description('Provide a globally unique name of your Azure Container Registry')
+param acrName string
+
+resource acrResource 'Microsoft.ContainerRegistry/registries@2023-01-01-preview' = {
```
djsly, line 7: any reasons the AVM wasn't used , instead of the raw resources ?

djsly: https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/container-registry/registry

e2e/template.go
AlisonB319: fun fact, turns out that global vars defined in _test.go files are not accessible since they are created at runtime. So I'm moving it to a different file so that it can be accessible for all :)

PR #5083, opened 2024-10-14
[node-bootstrapper] configure logging

r2k1:
Log to file from the application. Simplify CSE. It will be hard to modify CSE in the future, we want to avoid any flexibility.


e2e/scenario_test.go
```diff
@@ -568,7 +568,7 @@ func Test_ubuntu2204ScriptlessInstaller(t *testing.T) {
 			Cluster:               ClusterKubenet,
 			VHD:                   config.VHDUbuntu2204Gen2Containerd,
 			LiveVMValidators: []*LiveVMValidator{
-				FileHasContentsValidator("/var/log/azure/node-bootstrapper.log", "node-bootstrapper started"),
+				FileHasContentsValidator("/var/log/azure/node-bootstrapper.log", "node-bootstrapper finished"),
```
timmy-wright, line 571: ❤️

timmy-wright (approved):

node-bootstrapper/main.go
```diff
@@ -18,23 +20,64 @@ import (
 	"github.com/Azure/agentbaker/pkg/agent/datamodel"
 )
 
+// Some options are intentionally non-configurable to avoid customization by users
+// it will help us to avoid introducing any breaking changes in the future.
 const (
+	ConfigVersion                  = "v0"
 	DefaultAksAadAppID             = "6dae42f8-4368-4678-94ff-3960e28e3630"
-	ReadOnlyWorld      os.FileMode = 0644
+	LogFile                        = "/var/log/azure/node-bootstrapper.log"
```
cameronmeissner, line 28: im assuming we're planning on linux-only for the time being?

cameronmeissner (approved):

node-bootstrapper/main.go
r2k1: Good point, we will need to come up with something for windows later.

But before that I need to figure out a way to quickly run windows tests, so I can iterate somehow.

PR #5096, opened 2024-10-15
feat: create private acr in e2e tests

AlisonB319:
**What type of PR is this?**

**What this PR does / why we need it**: Creating the private acr for airgap e2e tests using the azure sdk instead of relying on a manually made object.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/go.mod
```diff
@@ -27,6 +27,7 @@ require (
 )
 
 require (
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerregistry/armcontainerregistry v1.3.0-beta.1 // indirect
 	github.com/Azure/go-autorest/autorest v0.11.29 // indirect
```
AlisonB319, line 31: I had to use the 1.3.0-beta.1 over the stable 1.2.0 in order to have access to `AnonymousPullEnabled` in the update params, which is a required field to be able to pull from the ACR (while we don't have oras login)

e2e/template.go
```diff
@@ -452,7 +452,7 @@ func baseTemplate(location string) *datamodel.NodeBootstrappingConfiguration {
 func getHTTPServerTemplate(podName, nodeName string, isAirgap bool) string {
 	image := "mcr.microsoft.com/cbl-mariner/busybox:2.0"
 	if isAirgap {
-		image = "aksvhdtestcr.azurecr.io/aks/cbl-mariner/busybox:2.0"
+		image = fmt.Sprintf("%s.azurecr.io/aks/cbl-mariner/busybox:2.0", PrivateACRName)
```
r2k1, line 455: is there anything that going push the image into ACR?

If the `PrivateACRName` is going to change or the ACR is deleted then tests will break, right?

Should the image be pulled from mcr and pushed to acr if it doesn't exist?

e2e/aks_model.go
```diff
@@ -201,7 +210,101 @@ func privateEndpointExists(ctx context.Context, t *testing.T, nodeResourceGroup,
...
+			Name: to.Ptr(armcontainerregistry.SKUNamePremium),
+		},
+		Properties: &armcontainerregistry.RegistryProperties{
+			AdminUserEnabled:     to.Ptr(false),
+			AnonymousPullEnabled: to.Ptr(true), // required to pull images from the private ACR without authentication
+		},
+	}
+
+	_, err := config.Azure.RegistriesClient.Get(ctx, nodeResourceGroup, privateACRName, nil)
+	if err != nil && strings.Contains(err.Error(), "ResourceNotFound") {
```
r2k1, line 228: `strings.Contains` is generally not a great way to check error.

The error most likely is a custom type with a field containing error code. I usually put a breakpoint and run debugger to quickly check what I can use from it. Then you can use `errors.As` and check error code (or dig into documentation/source code)

e2e/aks_model.go
```diff
@@ -201,7 +210,101 @@ func privateEndpointExists(ctx context.Context, t *testing.T, nodeResourceGroup,
 	return false, nil
 }
 
-func createPrivateEndpoint(ctx context.Context, t *testing.T, nodeResourceGroup, privateEndpointName string, vnet VNet) (armnetwork.PrivateEndpointsClientCreateOrUpdateResponse, error) {
+func createPrivateAzureContainerRegistry(ctx context.Context, t *testing.T, nodeResourceGroup, privateACRName string) error {
```
r2k1, line 213: Ignore for now, but sometimes I'm thinking about moving such functions to a method on `AzureClient` type

e2e/aks_model.go
```diff
@@ -9,11 +9,15 @@ import (
...
 	"github.com/Azure/agentbakere2e/config"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerregistry/armcontainerregistry"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/privatedns/armprivatedns"
 )
 
+// global var used here and in scenario_test.go, kube.go, template.go
+var PrivateACRName string = "privateacre2e"
```
r2k1, line 19: I would move it to config/config.go

r2k1 (approved):

e2e/template.go
AlisonB319: upstream pipelines deal with taking newly published images and pushing them to MCR. Because ACR has a cache rule to MCR, as long as it exists in MCR we will be able to grab it. 

If the ACR is deleted / a name change then yes the tests will break, but the ACR should get re-created during e2e setup.

r2k1: >Because ACR has a cache rule to MCR, as long as it exists in MCR we will be able to grab it.

Is there something pushed the image into ACR?

e2e/config/config.go
```diff
@@ -11,7 +11,7 @@ import (
 var (
 	Config                      = mustLoadConfig()
 	Azure                       = mustNewAzureClient(Config.SubscriptionID)
-	ResourceGroupName           = "abe2e-" + Config.Location
+	ResourceGroupName           = "abe2e-" + Config.Location + "alison-test"
```
AlisonB319, line 14: For testing will revert

e2e/pollers.go
```diff
@@ -58,14 +59,26 @@ func waitUntilNodeReady(ctx context.Context, t *testing.T, kube *Kubeclient, vms
 	return nodeName
 }
 
-func waitUntilPodReady(ctx context.Context, kube *Kubeclient, podName string) error {
+func waitUntilPodReady(ctx context.Context, kube *Kubeclient, podName string, t *testing.T) error {
```
AlisonB319, line 62: I think adding some logs here can help determine if your e2e test setup is going to fail without having to wait the 15 minutes for the timeout. And if everything is working, it doesn't produce that many more log lines (like 4-6 I believe)

e2e/template.go
AlisonB319: Yeah there is the aks upstream team who has pipelines that get triggered when the open source upstream github repos get new versions. When that happens the images get pushed to MCR, during thoes triggered pipelines, after the aks upstream team does some extra validation. But its supposed to happen basically the same time as the open source github publishes a new version.

e2e/aks_model.go
```diff
@@ -201,7 +201,82 @@ func privateEndpointExists(ctx context.Context, t *testing.T, nodeResourceGroup,
...
+		Location: to.Ptr(config.Config.Location),
+		SKU: &armcontainerregistry.SKU{
+			Name: to.Ptr(armcontainerregistry.SKUNamePremium),
+		},
+		Properties: &armcontainerregistry.RegistryProperties{
+			AdminUserEnabled:     to.Ptr(false),
+			AnonymousPullEnabled: to.Ptr(true), // required to pull images from the private ACR without authentication
+		},
+	}
+	pollerResp, err := config.Azure.RegistriesClient.BeginCreate(
```
AlisonB319, line 230: Only using BeginCreate not BeginUpdate since this ACR will only be created if it does not exist.

e2e/config/config.go
```diff
@@ -9,9 +9,11 @@ import (
 )
 
 var (
-	Config                      = mustLoadConfig()
-	Azure                       = mustNewAzureClient(Config.SubscriptionID)
-	ResourceGroupName           = "abe2e-" + Config.Location
+	Config            = mustLoadConfig()
+	Azure             = mustNewAzureClient(Config.SubscriptionID)
+	ResourceGroupName = "abe2e-" + Config.Location + "alison-test"
```
AlisonB319, line 14: Will revert before merge

e2e/go.mod
```diff
@@ -28,6 +28,7 @@ require (
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.10.0 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerregistry/armcontainerregistry v1.3.0-beta.1 // indirect
```
AlisonB319, line 31: Had to use 1.3.0-beta.1 in order to have access to AnonymousPullEnabled in ACR creation

e2e/aks_model.go
```diff
@@ -201,7 +201,82 @@ func privateEndpointExists(ctx context.Context, t *testing.T, nodeResourceGroup,
...
+
+	t.Logf("ACR does not exist, creating...")
+	createParams := armcontainerregistry.Registry{
+		Location: to.Ptr(config.Config.Location),
+		SKU: &armcontainerregistry.SKU{
+			Name: to.Ptr(armcontainerregistry.SKUNamePremium),
+		},
+		Properties: &armcontainerregistry.RegistryProperties{
+			AdminUserEnabled:     to.Ptr(false),
+			AnonymousPullEnabled: to.Ptr(true), // required to pull images from the private ACR without authentication
```
djsly, line 227: this sucks, we will need to have another one later one with AnonymousPullEnabled == false :(

djsly (approved):
Great stuff !

e2e/aks_model.go
AlisonB319: Yep we can always have 2 ACRs where one requires authentication and split up which airgap tests we want to run / double up. But we can cross that bridge when we get there :)

PR #5107, opened 2024-10-16
Reduce number of e2e tests to reduce VM usage

r2k1:
/kind test

This should reduce pressure on VM quota and save some $$$

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -22,10 +22,12 @@ func Test_azurelinuxv2(t *testing.T) {
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
 				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-azurelinux-v2-gen2"
 				nbc.AgentPoolProfile.Distro = "aks-azurelinux-v2-gen2"
+				nbc.AgentPoolProfile.MessageOfTheDay = "Zm9vYmFyDQo=" // base64 for foobar
 			},
 			LiveVMValidators: []*LiveVMValidator{
 				// for now azure linux reports itself as mariner, so expected version for azure linux is the same as that for mariner
 				mobyComponentVersionValidator("containerd", getExpectedPackageVersions("containerd", "mariner", "current")[0], "dnf"),
+				FileHasContentsValidator("/etc/motd", "foobar"),
 			},
 		},
 	})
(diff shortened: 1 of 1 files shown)


e2e/scenario_test.go
```diff
@@ -1201,39 +1126,3 @@ func Test_ubuntu1804imdsrestriction_mangletable(t *testing.T) {
 		},
 	})
 }
-
-func Test_Ubuntu2204MessageOfTheDay(t *testing.T) {
```
cameronmeissner, line 1205: will wanna check with @UtheMan on these, I believe they were recently added

cameronmeissner:
might also consider just adding a new tag to indicate that a given scenario shouldn't be executed by default?

granted scenarios with said tag would probably just end up rotting anyways

e2e/scenario_test.go
UtheMan: That should be fine, added it as separate test since its a separate feature, but if we are fine with combining these cases like this then i'm all for it.

e2e/scenario_test.go
```diff
@@ -212,33 +214,6 @@ func Test_azurelinuxv2gpu(t *testing.T) {
 	})
 }
 
-func Test_azurelinuxv2gpu_azurecni(t *testing.T) {
```
UtheMan, line 215: @ganeshkumarashok do you recall why we had separate tests for azure cni?

UtheMan:
> might also consider just adding a new tag to indicate that a given scenario shouldn't be executed by default?
> 
> granted scenarios with said tag would probably just end up rotting anyways

We could use some sort of "suggestions" for when to run certain scenarios like there are in RP, but many of the tests are tied to CSE execution which could get borked for other things for no reason sometimes... so not sure how safe that would be.

e2e/scenario_test.go
```diff
@@ -22,10 +22,12 @@ func Test_azurelinuxv2(t *testing.T) {
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
 				nbc.ContainerService.Properties.AgentPoolProfiles[0].Distro = "aks-azurelinux-v2-gen2"
 				nbc.AgentPoolProfile.Distro = "aks-azurelinux-v2-gen2"
+				nbc.AgentPoolProfile.MessageOfTheDay = "Zm9vYmFyDQo=" // base64 for foobar
 			},
 			LiveVMValidators: []*LiveVMValidator{
 				// for now azure linux reports itself as mariner, so expected version for azure linux is the same as that for mariner
 				mobyComponentVersionValidator("containerd", getExpectedPackageVersions("containerd", "mariner", "current")[0], "dnf"),
+				FileHasContentsValidator("/etc/motd", "foobar"),
```
djsly, line 30: personally I dont like to test non default settings in the base happy path. We should actually ensure that ALL settings that are not configured are not showing up in the base happy path test.

e2e/scenario_test.go
```diff
@@ -435,32 +410,6 @@ func Test_marinerv2CustomSysctls(t *testing.T) {
...
-				nbc.EnableNvidia = true
-			},
-			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
-				vmss.SKU.Name = to.Ptr("Standard_NC6s_v3")
-			},
-		},
-	})
-}
-
 func Test_marinerv2gpu_azurecni(t *testing.T) {
```
djsly, line 413: maybe we should keep azure_cni + gpu on the latest AzureLinux ? marinerv2 is somewhat deprecated now no ?

e2e/scenario_test.go
djsly: We should have a Happy Path test where nothing gets configured and we ensure that none of those settings are getting set. 

We could look at combining multiple features into another single e2e test to minimize the number of tests

e2e/scenario_test.go
```diff
@@ -600,33 +549,6 @@ func Test_ubuntu1804gpu(t *testing.T) {
 	})
 }
 
-func Test_ubuntu1804gpu_azurecni(t *testing.T) {
```
djsly, line 603: are we loosing the only 1804gpu test we had ?

e2e/scenario_test.go
r2k1: While it looks like we didn't set nothing, it's far from truth: 

https://github.com/Azure/AgentBaker/blob/master/e2e/template.go#L13-L449

e2e/scenario_test.go
r2k1: we had two of them

e2e/scenario_test.go
r2k1: is it deprecated?

PR #5109, opened 2024-10-17
test: add unit test to verify command line flag generation for seccomp

bravebeaver:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


pkg/agent/utils_test.go
```diff
@@ -743,6 +746,39 @@ var _ = Describe("Test GetOrderedKubeletConfigFlagString", func() {
...
+
+		ap := &datamodel.AgentPoolProfile{
+			CustomKubeletConfig: &datamodel.CustomKubeletConfig{
+				SeccompDefault: to.BoolPtr(false),
+			},
+		}
+
+		expectedStr := "--node-labels=topology.kubernetes.io/region=southcentralus --seccomp-default=true "
+		actualStr := GetOrderedKubeletConfigFlagString(k, cs, ap, true)
+		Expect(expectedStr).To(Equal(actualStr), fmt.Sprintf("Expected: %s\n, but got: %s", expectedStr, actualStr))
```
AlisonB319, line 780: make sure to import fmt if you want to use it here

pkg/agent/utils_test.go
```diff
@@ -743,6 +746,39 @@ var _ = Describe("Test GetOrderedKubeletConfigFlagString", func() {
...
+		}
+
+		ap := &datamodel.AgentPoolProfile{
+			CustomKubeletConfig: &datamodel.CustomKubeletConfig{
+				SeccompDefault: to.BoolPtr(false),
+			},
+		}
+
+		expectedStr := "--node-labels=topology.kubernetes.io/region=southcentralus --seccomp-default=true "
+		actualStr := GetOrderedKubeletConfigFlagString(k, cs, ap, true)
```
AlisonB319, line 779: Seems this function got updated to take a datamodel, I'd just do something like 
```
dm := datamodel.NodeBootstrappingConfiguration{
			KubeletConfig:    k,
			ContainerService: cs,
			AgentPoolProfile: ap,
		}
actualStr := GetOrderedKubeletConfigFlagString(&dm)
```

pkg/agent/utils_test.go
cameronmeissner: yeah the extra message here is unecessary

cameronmeissner:
do we already have e2e tests/validation that cover this scenario? if not might be worth adding as well

pkg/agent/utils_test.go
bravebeaver: weirdly when i ran for the first time it did not print out on my terminal. rebase error 😅. updated

pkg/agent/utils_test.go
bravebeaver: done.. rebased and resolved merge conflicts

pkg/agent/utils_test.go
```diff
@@ -743,6 +746,39 @@ var _ = Describe("Test GetOrderedKubeletConfigFlagString", func() {
...
+							Config: map[string]string{
+								"--seccomp-default": "true",
+							},
+						},
+					},
+				},
+			},
+		}
+		k := map[string]string{
+			"--node-labels": "topology.kubernetes.io/region=southcentralus",
```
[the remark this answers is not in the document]
bravebeaver: to check whether the seccompDefault flag is properly passed down / merged

bravebeaver:
> do we already have e2e tests/validation that cover this scenario? if not might be worth adding as well

that's my plan to add e2e down the path, first to understand how the configs are generated along the way and fill in unit test gaps.

r2k1:
> do we already have e2e tests/validation that cover this scenario? if not might be worth adding as well

While e2e can be handy to use. I think we should avoid putting stuff there. It's better to test things on other layers when possible.
e2e are hard to debug. And reliability drops exponentially with increasing number of tests.

r2k1 (approved):

pkg/agent/utils_test.go
```diff
@@ -712,10 +714,11 @@ var _ = Describe("Test GetOrderedKubeletConfigFlagString", func() {
 			AgentPoolProfile:        &datamodel.AgentPoolProfile{},
 		}
 
-		expectStr := "--event-qps=0 --image-gc-high-threshold=85 --node-status-update-frequency=20s --streaming-connection-idle-timeout=4h0m0s "
+		expectStr := "--event-qps=0 --image-gc-high-threshold=85 --node-status-update-frequency=20s --seccomp-default=true --streaming-connection-idle-timeout=4h0m0s "
```
r2k1, line 717: do you need more than 1 test?

timmy-wright (approved):

PR #5115, opened 2024-10-18
refactor: configure kubelet serving certificate rotation label at runtime

cameronmeissner:
**What type of PR is this?**

this PR moves the logic which adds the special `kubernetes.azure.com/kubelet-serving-ca=cluster` label out of agentbaker and directly into provisioning scripts - this is primarily to ensure that we preserve this functionality when using scriptless, and there is no agentbakersvc

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


staging/cse/windows/kubeletfunc.ps1
```diff
@@ -209,6 +209,24 @@ function Get-KubePackage {
...
+        [Parameter(Mandatory=$true)][string]
+        $KubeletNodeLabels,
+        [Parameter(Mandatory=$true)][string]
+        $Label
+    )
+
+    $labelList = $KubeletNodeLabels -split ","
+    foreach ($existingLabel in $labelList) {
+        if ($existingLabel -eq $Label) {
+            return $KubeletNodeLabels
```
AbelHu, line 223: Add logs here and line 226 so we can know when the label is added for troubleshooting?

AbelHu (approved):
LGTM. Thanks for the PR.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -423,21 +423,37 @@ getPrimaryNicIP() {
...
+        KUBELET_NODE_LABELS="${KUBELET_NODE_LABELS},${LABEL_STRING}"
+    else
+        # node labels shouldn't ever be empty, but we guard against it to be safe
+        KUBELET_NODE_LABELS=$LABEL_STRING
+    fi
+}
+
 # removes the specified LABEL_STRING (which should be in the form of 'label=value') from KUBELET_NODE_LABELS
-clearKubeletNodeLabel() {
+removeKubeletNodeLabel() {
```
lilypan26, line 443: nit: move this and `addKubeletNodeLabel` to cse_helpers? Might be useful in future for other cases

lilypan26:
is it safe to remove the equivalent logic from `baker.go` files? Or do we still need it there as well

cameronmeissner:
> is it safe to remove the equivalent logic from `baker.go` files? Or do we still need it there as well

we need to keep it in there for now since we can't release everything at once for windows (they need to create a new script package before we can start calling new functions from within the primary CSE scripts)

I'll follow-up and remove that logic once windows has everything released

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: moved

lilypan26 (approved):

staging/cse/windows/kubeletfunc.ps1
cameronmeissner: added

PR #5117, opened 2024-10-18
chore: fix the golinter 

AlisonB319:
**What type of PR is this?** cleaning up golinter warnings and errors

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/node-bootstrapper/main.go
+++ b/node-bootstrapper/main.go
@@ -60,12 +60,25 @@ func (s SensitiveString) UnsafeValue() string {
 	return string(s)
 }
 
+type Exit struct{ Code int }
+
+func handleExit() {
+	if e := recover(); e != nil {
+		if exit, ok := e.(Exit); ok {
+			os.Exit(exit.Code)
+		}
+		panic(e) // not an Exit, bubble up
+	}
+}
+
 func main() {
+	defer handleExit()
+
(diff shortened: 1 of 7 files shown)


node-bootstrapper/main.go
```diff
@@ -79,9 +79,11 @@ func main() {
 		var exitErr *exec.ExitError
 		_ = logFile.Close()
 		if errors.As(err, &exitErr) {
-			os.Exit(exitErr.ExitCode())
+			slog.Error("node-bootstrapper finished with exit code", "exitCode", exitErr.ExitCode())
```
r2k1, line 82: This changes the logic. The linter is wrong here. Can you just replace with nolint statements?

AlisonB319: Well the logger isn't _wrong_ per say since you manually close the file and exit instead of letting the defer statement get hit. Seems people have thought about this problem before us - https://stackoverflow.com/questions/27629380/how-to-exit-a-go-program-honoring-deferred-calls

AlisonB319: I pretty much copy pasted the solution from stack overflow, we don't have to go with it and I can add nolint statements. But curious what others think of the solution :)

PR #5124, opened 2024-10-20
bug: compress customdata

timmy-wright:
**What type of PR is this?**

/kind bug

This PR fixes a bug where the custom data we put on a node to bootstrap the node can be too large. It consists of two changes:

* gzip the entire custom data before base-64 encoding the file
* increase gzip compression level from default to "best compression"

From one of the e2e tests:

Size of custom data:
* Before gzip: 83252 bytes
* Increasing compression level: 83092 bytes
* With gzip and better compression: 57248 bytes

Oddly, some of the files inside the custom data are already gzipped. When we removed that inner compression, the overall file size did not drop much (81852 bytes). So we're going with the base64 encoded gzip of a file that contains base64 encoded gzip files. It surprised us too.

To test this worked, we:
* Ran e2e tests to check that nodes could be created in the agent baker e2e test environment.
* Created a standalone environment, versioned this branch into the standlone, created a cluster, and checked the cloud-init directory to verify that a gzipped file was put there.

image
image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
Improve compression of custom data that's sent to the VM for bootstrapping.
```


pkg/agent/utils.go
```diff
@@ -244,10 +244,18 @@ func isCommentAtTheEndOfLine(lastHashIndex int, trimmedToCheck string) bool {
...
+		return writer
+	}
+	return gzip.NewWriter(buf)
+}
+
 // getBase64EncodedGzippedCustomScriptFromStr will return a base64-encoded string of the gzip'd source data.
 func getBase64EncodedGzippedCustomScriptFromStr(str string) string {
 	var gzipB bytes.Buffer
-	w := gzip.NewWriter(&gzipB)
+	w := newGzipWriter(&gzipB)
```
r2k1, line 258: Same as:

```
w, _ := gzip.NewWriterLevel(&gzipB, gzip.BestCompression)
```

pkg/agent/baker_test.go
```diff
@@ -350,6 +350,14 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
 		)
 		Expect(err).To(BeNil())
 		customDataBytes, err := base64.StdEncoding.DecodeString(nodeBootstrapping.CustomData)
+		Expect(err).To(BeNil())
+
+		// optionally unzip the bytes
```
r2k1, line 355: does it need to be optional?

timmy-wright: Interesting question. I made it optional so that the test runner would better match what the cloud-init does - where it is optional. However, we want all custom data to always be zipped from now on, so making it mandatory in the test runner gives us unit testing that they are zipped. So I've changed this to make the zipping mandatory.

pkg/agent/baker_test.go
```diff
@@ -2043,7 +2049,12 @@ func verifyCertsEncoding(cert string) error {
 func getDecodedFilesFromCustomdata(data []byte) (map[string]*decodedValue, error) {
 	var customData cloudInit
 
-	if err := yaml.Unmarshal(data, &customData); err != nil {
+	decodedCse, err := getGzipDecodedValue(data)
+	if err != nil {
+		decodedCse = data
```
r2k1, line 2054: same here, it's a test. I don't think we should be flexible in what we allow in tests

r2k1 (approved):

PR #5130, opened 2024-10-21
Az Linux - nvidia container toolkit version update for CVE fix

ganeshkumarashok:
**What type of PR is this?**
/kind bug

This updates Nvidia Container Runtime to 1.16.2-1 for Az Linux, for fixing CVEs, particularly [CVE-2024-0132](https://nvd.nist.gov/vuln/detail/CVE-2024-0132).

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -75,7 +75,7 @@ installNvidiaFabricManager() {
 
 installNvidiaContainerToolkit() {
     MARINER_NVIDIA_CONTAINER_RUNTIME_VERSION="3.13.0"
-    MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION="1.15.0"
+    MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION="1.16.2"
 
     if [[ $OS_VERSION == "2.0" ]]; then
       MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION="1.13.5"
```
ganeshkumarashok, line 81: Why is there a separate nvidia container toolkit version for 2.0?

henryli001: Azure Linux 2.0 is in stable release stage and all package versions are locked down without any major version upgrades being performed to avoid causing regressions. This is why the container toolkit version on Azure Linux 2.0 is older than that on 3.0. 

For this case, we had to make a major version upgrade as well for 2.0 in order to resolve this critical CVE. So at this point, the container toolkit version on both 2.0 and 3.0 has been upgraded to 1.16.2.

henryli001 (commented):
`    if [[ $OS_VERSION == "2.0" ]]; then
      if ! dnf_install 30 1 600 nvidia-container-runtime-${MARINER_NVIDIA_CONTAINER_RUNTIME_VERSION}; then
        exit $ERR_APT_INSTALL_TIMEOUT
      fi
    fi
}`

The above code block can be removed since we deprecated nvidia-container-runtime package in 2.0 within the recent change to upgrade container toolkit to v1.16.2. This is because the upstream has deprecated the container runtime package and no longer produces new version that goes with container toolkit 1.16.2.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
juan-lee: So we should also update the env var for 2.0?

henryli001: I think we can just remove the following code block to specifically set the env var for container toolkit version on 2.0 since the container toolkit version is the same on both 2.0 and 3.0:

`    if [[ $OS_VERSION == "2.0" ]]; then
      MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION="1.13.5"
    fi`

There's another code block down below in the same function that's also targeting 2.0 only. I have highlighted this in another comment of mine that it can also be removed. After cleaning up these 2.0-specific code, this function is a generic implementation that works for both Azure Linux 2.0 and 3.0.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -75,7 +75,7 @@ installNvidiaFabricManager() {
 
 installNvidiaContainerToolkit() {
     MARINER_NVIDIA_CONTAINER_RUNTIME_VERSION="3.13.0"
```
henryli001, line 77: Let's also remove this line since container runtime is no longer used

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
ganeshkumarashok: Removed the 2.0 specific block. Good to know the background about this.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
ganeshkumarashok: Removed it.

dallasd1 (approved):

henryli001 (approved):

lilypan26 (approved):

PR #5133, opened 2024-10-21
Refactor AKS GPU consumption to use separate images per GPU driver type

ganeshkumarashok:
**What type of PR is this?**
/kind refactor 

We are now producing two separate images for CUDA and GRID, in the aks-gpu repo. This refactor, allows us to consume the two aks-gpu images separately. This will eventually help with automation with Renovate bot, as previously having one image with tags to separate `cuda` and `grid` wouldn't allow renovate to differentiate between them.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -353,8 +353,8 @@ capture_benchmark "${SCRIPT_NAME}_artifact_streaming_download"
 
 if [[ $OS == $UBUNTU_OS_NAME && $(isARM64) != 1 ]]; then  # no ARM64 SKU with GPU now
   gpu_action="copy"
-  NVIDIA_DRIVER_IMAGE_SHA="sha-c66998"
-  export NVIDIA_DRIVER_IMAGE_TAG="cuda-550.90.12-${NVIDIA_DRIVER_IMAGE_SHA}"
+  NVIDIA_DRIVER_IMAGE_SHA="20241008175307"
```
Devinwong, line 356: Do we have a plan to auto-update GPU version/tag and SHA or there isn't a good way to achieve that?

ganeshkumarashok: yes, this PR is to simplify automation. Next PR is to move it to components.json/config file to use Renovate

ganeshkumarashok: We are now producing two separate images for CUDA and GRID, in the aks-gpu repo. This refactor, allows us to consume the two aks-gpu images separately. This will eventually help with automation with Renovate bot, as previously having one image with tags to separate cuda and grid wouldn't allow renovate to differentiate between them.

pkg/agent/datamodel/const.go
```diff
@@ -134,15 +134,15 @@ const (
 
 const (
 	Nvidia470CudaDriverVersion = "cuda-470.82.01"
```
jason1028kr, line 136: Wanted to confirm 470 version doesn't need its prefix removed.

jason1028kr (approved):

pkg/agent/datamodel/const.go
ganeshkumarashok: We've stopped publishing the 470 version and it's only used for a much older SKU. So we're keeping it as it is for now.

PR #5137, opened 2024-10-22
Merge nbcparser and node-bootstrapper

r2k1:
Combine two binaries together. Remove logic duplication.
Remove overhead of managing two different applications.


node-bootstrapper/parser/helper.go
```diff
@@ -673,12 +673,12 @@ func getInitAKSCustomCloudFilepath() string {
...
-	return common.GPUNeedsFabricManager(vmSize)
+	return agent.GPUNeedsFabricManager(vmSize)
 }
 
 func getEnableNvidia(config *nbcontractv1.Configuration) bool {
 	if config.GpuConfig != nil && config.GpuConfig.EnableNvidia != nil {
 		return *config.GpuConfig.EnableNvidia
 	}
-	return common.IsNvidiaEnabledSKU(config.GetVmSize())
+	return false
```
r2k1, line 683: This function is gone. Is it ok to do that?

parts/linux/cloud-init/artifacts/bootstrap.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
-set -euxo pipefail
-# TODO: propagate exit status to bootstrapper
-if [ ! -f /opt/azure/containers/nbc.json ]; then 
-    exit 1
-fi
-mkdir -p /var/log/azure/aks
-/opt/azure/containers/nbcparser --bootstrap-config=/opt/azure/containers/nbc.json > /var/log/azure/aks/bootstrap.log 2>&1
```
Devinwong, line 8: Are we still logging nbcparser/node-bootstrapper process somewhere if we remove this logging command?

pkg/proto/nbcontract/v1/helper.go
```diff
@@ -93,9 +91,6 @@ func (nBCB *NBContractBuilder) deepCopy(src, dst interface{}) error {
 // It returns an error if the contract is invalid.
 // This function should be called after applying all configuration and before sending to downstream component.
 func (nBCB *NBContractBuilder) ValidateNBContract() error {
-	if err := nBCB.validateSemVer(); err != nil {
```
Devinwong, line 96: Why remove this? This is to check if the contract version is valid.

parts/linux/cloud-init/artifacts/bootstrap.sh
r2k1: path to log file is hard-coded in node-bootstrapper. (questionable, but I don't want user to be able to change the path yet, we may want to ways to export it)

stdout/stderr currently replicating output of CSE scripts and doesn't match logging.

pkg/proto/nbcontract/v1/helper.go
r2k1: Two points:

1) We should avoid using semantic versioning (semver) for config versioning. I don't anticipate any functional difference between, for example, config versions 1.0.0 and 1.0.1. Requiring users to constantly update versions for minor or patch-level changes adds unnecessary complexity.
2) We only need to track major version changes. Different versions will correspond to different structs, meaning version parsing and validation will take place before unmarshalling the config. Currently, the only supported version is v0 (this allows us to introduce breaking changes before the final release if needed; if no changes are required, v0 will effectively serve as a synonym for v1).

node-bootstrapper/parser/helper.go
lilypan26: Yes, this is fine.

pkg/agent/bakerapi.go
```diff
@@ -87,32 +87,56 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context, conf
...
 	return nodeBootstrapping, nil
 }
 
+func getScriptlessCustomDataTemplate(config any) (string, error) {
+	nbcJSON, err := json.Marshal(config)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal nbc, error: %w", err)
+	}
+	encodedNBCJson := base64.StdEncoding.EncodeToString(nbcJSON)
+	customDataTAML := fmt.Sprintf(`#cloud-config
```
lilypan26, line 130: nit: typo `customDataYAML`

pkg/agent/bakerapi.go
```diff
@@ -87,32 +87,56 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrapping(ctx context.Context, conf
...
+	}
+
+	if osImageConfig, hasImage := osImageConfigMap[distro]; hasImage {
+		nodeBootstrapping.OSImageConfig = &osImageConfig
+	}
+
 	return nodeBootstrapping, nil
 }
 
+func getScriptlessCustomDataTemplate(config any) (string, error) {
```
lilypan26, line 124: nit: rename to `getScriptlessCustomDataContent`?

node-bootstrapper/main.go
```diff
@@ -112,52 +85,25 @@ func Provision(ctx context.Context) error {
 		return errors.New("--provision-config is required")
```
lilypan26, line 85: I believe if `provision-config` flag is not specified, then `provisionConfig` will return a pointer to the default value which in this case is empty string rather than returning `nil`. So we should check against empty string.

lilypan26 (approved):

pkg/proto/nbcontract/v1/helper.go
Devinwong: 1. User doesn't need to update the version manually. If he/she is using the nbcContractBuilder, it will apply the version from the package.
2. Right. Only major version matters in this validation function. Minor version just returns a warning. So why removing the major version check?

r2k1: I don't think we should force users to use nbcContractBuild to build the config.

https://github.com/Azure/AgentBaker/pull/5137/files#diff-54d0a4d20361aa9831f61f6129558f7567bd116eb245ca57ac21ff4de4d40cd5R40-R41

I moved the validation is there. In the future I think we will have logic like this:
```
1) Parse json. But read only `version` property.
2) If `version == 0`, then parse json into ContractV0 struct, if `version==1` then parse json into ContractV1 struct, otherwise fail.
```
Currently we support only single version, so the logic isn't implemented yet.

Devinwong: Agree. We don't need to force user to use the builder. Also agree we probably don't need semver for contract version otherwise we will need to keep updating the version when we add a new field. If we already have a place to check the version, I am fine with it.

Devinwong (approved):

PR #5149, opened 2024-10-24
documentation: create a readme for new AKSNodeConfig

Devinwong:
**What type of PR is this?**
/kind documentation

Create a readme for new AKSNodeConfig.
Some tiny cleanup of the contract.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/proto/nbcontract/v1/kubeletconfig.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.33.0
+// 	protoc-gen-go v1.35.1
```
timmy-wright, line 3: Should there be a matching version change in a go.mod file or somewhere like that?

timmy-wright (approved):

pkg/proto/nbcontract/v1/kubeletconfig.pb.go
Devinwong: The makefile downloads the binary `buf` every time to compile to .go files. So it doesn't have go.mod file.

PR #5153, opened 2024-10-24
GPUContainerImage schema with os, arch and cache info - move GPU container images to config

ganeshkumarashok:
**What type of PR is this?**
/kind feature

This PR moves GPU versions to a config file (components.json), so that Renovate bot can auto-update it. VHD builds will now consume the cuda version from components.json. It also adds a new schema to auto-update.

There are two new requirements:

aks-gpu-cuda container image is only downloaded for particular combo of OS and arch (Ubuntu - amd64),
aks-gpu-grid container image needs to be present in the config but is never downloaded in the VHD. It's only used in CSE, for certain SKUs.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


schemas/components.cue
```diff
@@ -9,13 +9,27 @@ package components
...
+
 #ContainerImage: {
 	downloadURL: string
 	amd64OnlyVersions:     [...string]
 	multiArchVersionsV2:   [...#VersionV2]
 }
 
+#GPUContainerImage: {
+	downloadURL: string
+	multiArchVersionsV2:   [...#VersionV2]
```
ganeshkumarashok, line 26: Any recommendations for alternate names for `multiArchVersionsV2`? Thought of `Version` but `VersionV2` already exists

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -351,28 +351,104 @@ INSTALLED_RUNC_VERSION=$(runc --version | head -n1 | sed 's/runc version //')
...
+    shouldPull=1
+  fi
+
+  if [[ "$shouldPull" == "1" ]]; then
+    # Extract image details
+    downloadURL=$(echo "${gpuImageToBePulled}" | jq -r '.downloadURL')
+    imageName=$(echo "$downloadURL" | sed 's/:.*$//')
+
+    # Get the latestVersion
+    latestVersion=$(echo "${gpuImageToBePulled}" | jq -r '.multiArchVersionsV2[0].latestVersion')
```
Devinwong, line 397: If you only put [0], later on when you add another multiArchVersionsV2[1] block in components.json, it won't capture it.
Suggested to loop the whole `multiArchVersionsV2` array. And if you only have 1 multiArchVersionsV2 entry in components.json, it will just work like using [0].

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -351,28 +351,104 @@ INSTALLED_RUNC_VERSION=$(runc --version | head -n1 | sed 's/runc version //')
...
+  shouldPull=0  # Default to not pull
+
+  if [[ -n "$osSelectors" ]]; then
+    # osSelectors is provided; check if current OS and arch match any entry
+    while IFS= read -r selector; do
+      os=$(echo "$selector" | jq -r '.os')
+      arch=$(echo "$selector" | jq -r '.arch')
+
+      if [[ "$os" == "$CURRENT_OS" ]]; then
+        if [[ "$arch" == "$CPU_ARCH" ]]; then
```
Devinwong, line 379: It would be great to avoid more than 2 level nested `if` for readability. If you can't figure out a way, consider to put it in a function with descriptive name.

vhdbuilder/packer/install-dependencies.sh
Devinwong: 1 example for your reference: 
At around ln414.
```  
  MULTI_ARCH_VERSIONS=()
  updateMultiArchVersions "${imageToBePulled}"
```

ganeshkumarashok:
This is the much simpler alternate PR we considered (without OS, arch): https://github.com/Azure/AgentBaker/pull/5138, and the more complex alternate (adding it to ContainerImages): https://github.com/Azure/AgentBaker/pull/5139

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -351,28 +351,104 @@ INSTALLED_RUNC_VERSION=$(runc --version | head -n1 | sed 's/runc version //')
...
 
-  cat << EOF >> ${VHD_LOGS_FILEPATH}
-  - nvidia-driver=${NVIDIA_DRIVER_IMAGE_TAG}
-EOF
+  # Extract 'osSelectors' if present
+  osSelectors=$(echo "${gpuImageToBePulled}" | jq -r '.osSelectors // empty')
+
+  shouldPull=0  # Default to not pull
+
+  if [[ -n "$osSelectors" ]]; then
```
Devinwong, line 372: Suggested to put this logic into a function and add unit tests to cover most of the if conditions, so that we don't need to rely on abe2e or RP-e2e to capture issues for us.
One way to do that is put the function into cse_helpers.sh. There is a shellspec unit test file cse_helpers.sh which have some examples to author tests.
The root level readme has some instructions too.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -351,28 +351,104 @@ INSTALLED_RUNC_VERSION=$(runc --version | head -n1 | sed 's/runc version //')
...
+
+    # Set gpu_action if pulling the aks-gpu-cuda image
+    if [[ "$imageName" == "mcr.microsoft.com/aks/aks-gpu-cuda" ]]; then
+      gpu_action="copy"
+
+      # Create necessary directories
+      mkdir -p /opt/{actions,gpu}
+
+      # Check for the "fullgpu" feature flag
+      if grep -q "fullgpu" <<< "$FEATURE_FLAGS"; then
```
Devinwong, line 425: Again, avoid more than 2 level nested if. It's hard to keep track which level it is for debugging and readability.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -351,28 +351,104 @@ INSTALLED_RUNC_VERSION=$(runc --version | head -n1 | sed 's/runc version //')
...
+        echo "Installing GPU driver from image: $fullImage"
+        bash -c "$CTR_GPU_INSTALL_CMD $fullImage gpuinstall /entrypoint.sh install"
+        ret=$?
+        if [[ "$ret" != "0" ]]; then
```
Devinwong, line 429: Again, avoid more than 2 level nested if. It's hard to keep track which level it is for debugging and readability.

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Thanks, updated it to reuse that function

schemas/components.cue
Devinwong: `gpuImageVersion`? 
I 'd love to have GPU in capital letters but it's not following the naming convention.

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -558,6 +558,36 @@
...
+      ]
+    },
+    {
+      "downloadURL": "mcr.microsoft.com/aks/aks-gpu-grid:*",
+      "multiArchVersionsV2": [
+      { 
+        "renovateTag": "registry=https://mcr.microsoft.com, name=aks/aks-gpu-grid",
+        "latestVersion": "535.161.08-20241021235607"
+      }
+  ],
```
Devinwong, line 586: indent

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Actually, under no circumstances would we need to pull two CUDA images for the same VHD

Devinwong: If that's the case, maybe create a `gpuImageVersion` and it's just simple struct instead of an array of struct? Then it's much cleaner and explicit enough. Otherwise one day someone will ask, why we were only picking the first element of an array, but we didn't just create a non-array struct? Does that sound making sense to you?

ganeshkumarashok: Planning to change it to so we constrain it to one gpuVersion per GPUContainerImage: 

#GPUContainerImage: {
	downloadURL: string
	gpuIVersion:   #VersionV2
	cached: bool
	osSelectors?:         [...#OSSelector]
}

schemas/components.cue
ganeshkumarashok: thanks, switched to `gpuImageVersion`. Yes, caps would be nice but isn't consistent with the rest of the fields.

ganeshkumarashok:
Had a discussion earlier and I merged this alternate PR instead: https://github.com/Azure/AgentBaker/pull/5138

PR #5157, opened 2024-10-25
update proxy tests

timmy-wright:
**What type of PR is this?**

/kind bug

* curl can use both the upper case and lower case versions of HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
* it prefers the lower case one if both are there.
* We are setting both in /etc/environment (in cse_config.sh), but not both in this file.
* So it makes sense to set both here.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
Set both upper and lower case shell variables HTTP_PROXY, HTTPS_PROXY, and NO_PROXY during bootstrapping.
```


pkg/agent/variables.go
```diff
@@ -204,17 +204,22 @@ func getOutBoundCmd(nbc *datamodel.NodeBootstrappingConfiguration, cloudSpecConf
 
 func getProxyVariables(nbc *datamodel.NodeBootstrappingConfiguration) string {
 	// only use https proxy, if user doesn't specify httpsProxy we autofill it with value from httpProxy.
+	// note that we set both upper case and lower case varients. curl can use both, but has a preference for the
```
babbageclunk, line 207: Super minor typo:
```suggestion
	// note that we set both upper case and lower case variants. curl can use both, but has a preference for the
```

pkg/agent/variables_test.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/stretchr/testify/assert"
+)
+
+func MutateMap[T any](ts []T, f func(T) T) []T {
```
cameronmeissner, line 12: we don't need to export this

pkg/agent/variables_test.go
```diff
@@ -0,0 +1,103 @@
...
+	for i := range ts {
+		ts[i] = f(ts[i])
+	}
+	return ts
+}
+
+func trim(t string) string { return strings.Trim(t, " ") }
```
cameronmeissner, line 19: is the trimming actually doing anything if we're asserting `Contains`?

timmy-wright: Aaah, Contains is done on the array of strings once we split by ";" characters rather than the strings themselves. So the assertion is "this exact string appears in the array". Which makes the trimming important.

pkg/agent/variables_test.go
timmy-wright: I agree. HAve changed capitalisation.

PR #5162, opened 2024-10-25
Use GPU versions in components.json for CSE

ganeshkumarashok:
**What type of PR is this?**
/kind refactor
/kind design

This is to make CSE consume the same GPU driver versions from components.json (for Ubuntu). This will make it consistent with how the VHD now consumes it, and we will have a single source of truth for GPU drivers on Ubuntu.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/agent/datamodel/const.go
```diff
@@ -132,19 +140,71 @@ const (
...
+	return nil
+}
+
+//nolint:gochecknoinits
+func init() {
+	if err := LoadConfig(); err != nil {
+		panic(fmt.Sprintf("Failed to load configuration: %v", err))
+	}
+}
+
```
lilypan26, line 207: maybe move all this to another file `gpu_components.go`? `consts.go` probably not the best place

ganeshkumarashok: fair, moved it

pkg/agent/datamodel/gpu_components_test.go
```diff
@@ -0,0 +1,24 @@
...
+	}
+
+	if AKSGPUCudaVersionSuffix == "" {
+		t.Error("NvidiaCudaDriverVersion is empty")
+	}
+
+	if AKSGPUGridVersionSuffix == "" {
+		t.Error(("AKSGPUGridVersionSuffix is empty"))
+	}
+}
```
lilypan26, line 24: can we add some tests for checking that the driver version and suffix is parsed correctly from components.json?

ganeshkumarashok: Added regex matching for parsed image driver version and suffix to ensure it's parsed properly.

lilypan26 (approved):

PR #5166, opened 2024-10-26
Scriptless E2E

r2k1:
Modify E2E to allow testing using new configuration struct.

When `AKSNodeConfigMutator` defined then "aks-node-controller" will be used for bootstrapping.

/kind test


e2e/opts.go
cameronmeissner, line 1: thanks for getting rid of / improving this -  something I've been meaning to do for the longest time

cameronmeissner (approved):
thanks for the improvements, lgtm assuming e2es are passing

djsly (approved):

lilypan26 (approved):

PR #5173, opened 2024-10-28
feat: add scriptless bootstrap status poller

lilypan26:
**What type of PR is this?**

Add monitor mode to go binary to poll status of systemd unit and retrieve bootstrap exit status. Add cse processing to AB e2e to simulate how bootstrapper will retrieve provisioning completion information.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


node-bootstrapper/const.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+// Some options are intentionally non-configurable to avoid customization by users
+// it will help us to avoid introducing any breaking changes in the future.
+const (
+	LogFile               = "/var/log/azure/node-bootstrapper.log"
```
cameronmeissner, line 6: are we planning on consumers using these consts? if not I'd unexport them

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
 	return err
 }
 
+// usage example:
+// node-bootstrapper monitor
+func (a *App) Monitor(ctx context.Context) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Minute)
```
cameronmeissner, line 110: probably fine for now, but we'll most likely need to make the timeout configurable

RP currently imposes a VM extension provisioning timeout which is dictated by a toggle

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+
+	for {
+		select {
+		case <-timeoutCtx.Done():
+			// If the timeout or cancel occurs, exit with a timeout error
+			return fmt.Errorf("monitoring timed out: %s still active after 15 minutes", BootstrapService)
+		default:
+			_, err := os.Stat(ProvisionJSONFilePath)
+			// Check the active state of the unit
+			_, sysctlerr := a.runSystemctlCommand(ctx, "status", BootstrapService)
```
cameronmeissner, line 121: can we log `sysctlerr` when non-nil?

e2e/scenario_helpers_test.go
```diff
@@ -153,3 +159,73 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
...
+				t.Logf("failed to get instance view for VM %s: %v", *vmInstance.InstanceID, err)
+				continue
+			}
+			for _, extension := range instanceViewResp.Extensions {
+				for _, status := range extension.Statuses {
+					resp, err := parseLinuxCSEMessage(t, *status)
+					if err != nil {
+						return fmt.Errorf("Parse CSE message with error, error code: %s, raw message: %s", err.Code, *status.Message)
+					}
+					if !strings.EqualFold(resp.ExitCode, "0") {
```
cameronmeissner, line 183: we shouldn't need to case-fold when comparing against an integer

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
 	return err
 }
 
+// usage example:
+// node-bootstrapper monitor
```
cameronmeissner, line 108: can we move this to a readme instead?

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Minute)
+	defer cancel()
+
+	for {
+		select {
+		case <-timeoutCtx.Done():
+			// If the timeout or cancel occurs, exit with a timeout error
+			return fmt.Errorf("monitoring timed out: %s still active after 15 minutes", BootstrapService)
+		default:
+			_, err := os.Stat(ProvisionJSONFilePath)
```
cameronmeissner, line 119: can we check the content of err immediately after the `Stat`? I think that will be more readable, and will also allow us to short-circuit the call to systemctl below in the case where the file exists

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+			// Check the active state of the unit
+			_, sysctlerr := a.runSystemctlCommand(ctx, "status", BootstrapService)
+
+			// if service is inactive or failed, error code will be non-zero
+			if sysctlerr == nil || os.IsNotExist(err) {
+				// Unit is still active, sleep for 3 seconds before checking again
+				time.Sleep(3 * time.Second)
```
cameronmeissner, line 126: might also want to make the sleep time configurable with a sane default

e2e/const.go
```diff
@@ -1,6 +1,8 @@
 package e2e
 
 const (
-	buildIDTagKey    = "buildID"
-	defaultNamespace = "default"
+	buildIDTagKey             = "buildID"
+	defaultNamespace          = "default"
+	extensionErrorCodeRegex   = `ProvisioningState/failed/(\d+)`
```
cameronmeissner, line 6: nit: create a new const block for these to keep things logically separate?

e2e/scenario_helpers_test.go
```diff
@@ -122,9 +129,8 @@ func createAndValidateVM(ctx context.Context, t *testing.T, scenario *Scenario)
 	t.Logf("node %s is ready, proceeding with validation commands...", vmssName)
 
 	vmPrivateIP, err := getVMPrivateIPAddress(ctx, *scenario.Runtime.Cluster.Model.Properties.NodeResourceGroup, vmssName)
-	require.NoError(t, err)
```
djsly, line 125: we are ok with the `err` should we print it ?

pkg/agent/bakerapi.go
```diff
@@ -101,7 +101,7 @@ func (agentBaker *agentBakerImpl) GetNodeBootstrappingForScriptless(
 	}
 
 	nodeBootstrapping := &datamodel.NodeBootstrapping{
-		CSE:        "",
+		CSE:        scriptlessBootstrapStatusCSE,
```
anujmaheshwari1, line 104: nit: keep naming consistent for customData as well, scriptlessCustomData.

if not, then remove the same from CSE, your call

node-bootstrapper/const.go
```diff
@@ -0,0 +1,9 @@
...
+	BootstrapService      = "bootstrap.service"
+	ProvisionJSONFilePath = "/var/log/azure/aks/provision.json"
```
anujmaheshwari1, line 8: nit: consistent naming, LogFile / ProvisionJSONFile or LogFilePath / ProvisionJSONFilePath

question: what is provision.json? whats the purpose of the file?

node-bootstrapper/app.go
anujmaheshwari1, line 119: am I right in understanding that once the unit is finished, it will write the status to this file path?

if yes, why do we need to check both existence of the file path and whether unit is active?

keep sleeping/retrying till unit is running, exit if failed and if not, then check file content? this check seems redundant unless i am missing something.

is there a delay between unit finishing vs writing to file?

e2e/scenario_helpers_test.go
```diff
@@ -153,3 +159,73 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func parseLinuxCSEMessage(t *testing.T, status armcompute.InstanceViewStatus) (*datamodel.CSEStatus, *datamodel.CSEStatusParsingError) {
+	if status.Code == nil || status.Message == nil {
+		t.Logf("No valid Status code or Message provided from cse extension")
```
djsly, line 197: string duplicate, maybe create errorMsg := "No valid Status code or Message provided from cse extension"

e2e/scenario_helpers_test.go
```diff
@@ -153,3 +159,73 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
...
+						t.Logf("CSE completed successfully with exit code 0")
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func parseLinuxCSEMessage(t *testing.T, status armcompute.InstanceViewStatus) (*datamodel.CSEStatus, *datamodel.CSEStatusParsingError) {
```
anujmaheshwari1, line 195: suggestion: verify with https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/resourceprovider/server/microsoft.com/common/vmsshelper/csehelper.go&version=GBmaster&line=464&lineEnd=465&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents that we are covering all cases, nonetheless, good job getting this function out 🥇

node-bootstrapper/const.go
lilypan26: created here: https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_start.sh#L45

which contains provisioning completion information

node-bootstrapper/app.go
lilypan26: Yes im not sure why where seems to be a delay in `provision.json` being present and the systemd unit completing but without the additional file check, I ran into the case where the systemd unit finished but the file was not present.

e2e/scenario_helpers_test.go
lilypan26: Yes, mostly copied the code from there and modified it a bit since AB e2es are much simpler, but we mostly care that the cse output is what we expect which this verifies.

node-bootstrapper/app.go
lilypan26: just a random number I chose, we can tweak it if there are better suggestions. Though idk if we should increase it since you mentioned that in most cases provisioning should be complete by the time cse runs.

e2e/scenario_helpers_test.go
lilypan26: There we a duplicate check since we call `require.NoError` again below so I removed this one. But if the `err` is not nil, `require.NoError` wil log it.

lilypan26:
> Good job! 2 questions: Q1. How do we expect the work flow? Is it like
> 
> 1. AKS-RP calls CRP-API to pass AKSNodeConfig
> 2. Once systemd-unit finds the config is ready, execute AKS-Node-Controller with argument `provision`.
> 3. When will the argument `monitor` be called with?
> 
> We probably need to document the flow at some time later.
> 
> Q2. How does AKS-RP poll the status? Is it like sending a CSE with argument `monitor`?

RP will call CRP-API specifying CSE and customData which it will retrieve from `GetNodeBootstrappingForScriptless`. CSE will be a simple one-liner to call `./node-bootstrapper monitor` which I define here: https://github.com/Azure/AgentBaker/pull/5173/files#diff-9b2dd61f71c944e7360bfd7149d37a9f8a5ce4d92edae14371f95c9072a96d35R117. 

So the flow is that customData containing AKSNodeConfig will get written to disk during cloud-init, after which the systemd unit will start provisioning with `./node-bootstrapper provision`. In parallel, cse will run `./node-bootstrapper monitor` to poll status of the systemd unit until provisioning is complete. 

I'll add a follow-up task to add a read-me documenting the flow and usage of the go binary.

e2e/node_bootstrapper_test.go
```diff
@@ -51,7 +51,7 @@ func Test_ubuntu2204NodeBootstrapper(t *testing.T) {
 			LiveVMValidators: []*LiveVMValidator{
 				mobyComponentVersionValidator("containerd", getExpectedPackageVersions("containerd", "ubuntu", "r2204")[0], "apt"),
 				mobyComponentVersionValidator("runc", getExpectedPackageVersions("runc", "ubuntu", "r2204")[0], "apt"),
-				FileHasContentsValidator("/var/log/azure/node-bootstrapper.log", "node-bootstrapper finished successfully"),
+				FileHasContentsValidator("/var/log/azure/node-bootstrapper.log", "node-bootstrapper finished"),
```
r2k1, line 54: did you finally managed to run it?

node-bootstrapper/app.go
```diff
@@ -56,6 +58,8 @@ func (a *App) run(ctx context.Context, args []string) error {
 			return errors.New("--provision-config is required")
 		}
 		return a.Provision(ctx, ProvisionFlags{ProvisionConfig: *provisionConfig})
+	case "monitor":
```
r2k1, line 61: I feel like there could be better names. Like `watch` or `wait` or something else?

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
 	return err
 }
 
+// usage example:
+// node-bootstrapper monitor
+func (a *App) Monitor(ctx context.Context) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Minute)
+	defer cancel()
+
+	for {
```
r2k1, line 113: A more idiomatic is to use `time.Ticker` instead of `time.Sleep`

node-bootstrapper/app.go
r2k1: On average, `3*time.Second / 2` will be added to the VM provisioning time.

Smaller number => faster VM provision. I think it should be no more than a second. Maybe less then that.

I wonder if there is a better way to watch for a file appearance.

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+			// if service is inactive or failed, error code will be non-zero
+			if sysctlerr == nil || os.IsNotExist(err) {
+				// Unit is still active, sleep for 3 seconds before checking again
+				time.Sleep(3 * time.Second)
+				continue
+			}
+
+			provisionJSON, err := a.getProvisionJSON()
+			if err != nil {
+				return fmt.Errorf("error getting provision.json output, %w", err)
```
r2k1, line 132: error parsing

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+			}
+			// print to stdout so it will be returned by custom script extension
+			fmt.Println(provisionJSON)
```
r2k1, line 135: This part is hard to test, maybe return a `string` in the function?

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+			return nil
+		}
+	}
+}
+
+// runSystemctlCommand is a generic function that runs a systemctl command with specified arguments
+func (a *App) runSystemctlCommand(ctx context.Context, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, "systemctl", args...)
+	output, err := cmd.Output()
```
r2k1, line 144: Try to use `a.cmdRunner` somehow.  (Maybe with some modifications)

I added it so I can assert in tests a certain command has been executed

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+// getProvisionJSON returns the contents of provision.json containing bootstrap status info
+func (a *App) getProvisionJSON() (string, error) {
```
r2k1, line 152: This function is no different than `os.ReadFile(ProvisionJSONFilePath)`

e2e/scenario_helpers_test.go
```diff
@@ -153,3 +159,73 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
...
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get VMSS instances: %v", err)
+		}
+
+		for _, vmInstance := range page.Value {
+			instanceViewResp, err := config.Azure.VMSSVM.GetInstanceView(ctx, resourceGroupName, vmssName, *vmInstance.InstanceID, nil)
+			if err != nil {
+				t.Logf("failed to get instance view for VM %s: %v", *vmInstance.InstanceID, err)
+				continue
```
r2k1, line 175: Is there a reason to continue instead of returning the error?

e2e/scenario_helpers_test.go
```diff
@@ -153,3 +159,73 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
...
+			}
+			for _, extension := range instanceViewResp.Extensions {
+				for _, status := range extension.Statuses {
+					resp, err := parseLinuxCSEMessage(t, *status)
+					if err != nil {
+						return fmt.Errorf("Parse CSE message with error, error code: %s, raw message: %s", err.Code, *status.Message)
+					}
+					if !strings.EqualFold(resp.ExitCode, "0") {
+						return fmt.Errorf("vmssCSE %s, output=%s, error=%s", resp.ExitCode, resp.Output, resp.Error)
+					} else {
```
r2k1, line 185: nit: `else` is redundant and just adds some clutter.

e2e/scenario_helpers_test.go
r2k1: It's better to return `*datamodel.CSEStatus, error`

Using structs instead of `error` is causing a lot of issues. There is significant difference nil-interface and a nil-pointer. And at some stage one will be implicitly converted to another.

e2e/scenario_helpers_test.go
r2k1: I usually try to avoid logging if an error is returned. It generally just adds noise.

node-bootstrapper/app.go
```diff
@@ -100,6 +104,60 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
...
+	// Read the file contents
+	data, err := os.ReadFile(ProvisionJSONFilePath)
```
r2k1, line 154: FYI I'm thinking about something like adding

```
type IO interface {
  ExecCMD(cmd *exec.Cmd) error
  ReadFile(path string) (os.File, error)
  WriteFile(path string) (os.File, error)
}
```

So, we can mock all the external dependencies and have some unit tests

e2e/scenario_helpers_test.go
r2k1: For example, it may lead to situations like this:
```go
_, err1 := parseLinuxCSEMessage(...)
assert.Equal(err1, nil) // true
var err2 error
_, err2 := parseLinuxCSEMessage(...)
assert.Equal(err1, nil) // false
```

e2e/node_bootstrapper_test.go
lilypan26: no :(  i just noticed the validator from the other test was failing because the string in the binary changed

e2e/scenario_helpers_test.go
lilypan26: fixed, remove duplicate log message and just returned error

node-bootstrapper/app.go
lilypan26: changed to provision-wait for now

node-bootstrapper/app.go
lilypan26: good point, originally had more in this function. removed.

node-bootstrapper/app.go
lilypan26: removed to use `cmdRunner` instead, i modified it to call `cmd.Output` rather than `cmd.Run` if thats okay with you

e2e/scenario_helpers_test.go
lilypan26: thought it might still be worth to check the other vms if multiple in the scale set but in this case of ab e2es we only have 1 instance anyway. fixed to return error

e2e/scenario_helpers_test.go
lilypan26: removed.

node-bootstrapper/app.go
lilypan26: found a better way using `fsnotify`

node-bootstrapper/app.go
```diff
@@ -90,7 +103,7 @@ func (a *App) provisionStart(ctx context.Context, cse utils.SensitiveString) err
 	// it's also nice to have a single log file for all the important information, so write to both places
 	cmd.Stdout = io.MultiWriter(os.Stdout, &stdoutBuf)
 	cmd.Stderr = io.MultiWriter(os.Stderr, &stderrBuf)
-	err := a.cmdRunner(cmd)
+	_, err := a.cmdRunner(cmd)
```
r2k1, line 106: I don't think it's going to work.

`cmd.Output` overrides `Stdout` and `Stderr` properties.

I didn't use `Output` in this function mainly to allow streaming of shell script stdout/stderr for a potentially faster feedback. (Don't need to wait until cse execution finished before you can see the output)

node-bootstrapper/app.go
r2k1: It's annoying, but you probably will need to re-implement the `Output` function yourself.

node-bootstrapper/app.go
lilypan26: modified to return the file contents as string, will add tests as well

node-bootstrapper/app.go
lilypan26: actually i changed reverted it back and instead did the same thing in preserving stdout and stderr when executing the systemctl command. this will get returned in cse which is what we want.

node-bootstrapper/app.go
lilypan26: reverted it back

node-bootstrapper/app.go
lilypan26: Changed to only waiting for `provision.json` since that is all we need to check bootstrap completion. If timeout occurs, return status of systemd unit.

node-bootstrapper/app.go
lilypan26: removed timeout since RP already has timeout configured on cse and provisioning scripts have timeout baked in as well

e2e/scenario_helpers_test.go
```diff
@@ -107,6 +111,9 @@ func createAndValidateVM(ctx context.Context, t *testing.T, scenario *Scenario)
 	vmssName := getVmssName(t)
 	createVMSS(ctx, t, vmssName, scenario, privateKeyBytes, publicKeyBytes)
 
+	err = getCustomScriptExtensionStatus(ctx, t, *scenario.Runtime.Cluster.Model.Properties.NodeResourceGroup, vmssName)
+	require.NoError(t, err)
```
r2k1, line 115: Maybe other tests would benefit from it as well?
What if it's moved into `creteVMSS` function?

r2k1: It will also validate the behaviour hasn't changed.

lilypan26: yes currently this will get run for all tests, there is no check for scriptless nodes. I think it makes more sense to keep it here since it is a validation step.

node-bootstrapper/app.go
lilypan26: will create a followup task to add a readme

e2e/scenario_helpers_test.go
r2k1: ah, sorry. I thought I'm looking at different file.

node-bootstrapper/app.go
lilypan26: This would be great, though not sure if we will be able to mock the file wait using the new library..

aks-node-controller/app.go
```diff
@@ -95,6 +102,48 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
 	return err
 }
 
+func (a *App) ProvisionWait(ctx context.Context) (string, error) {
```
r2k1, line 105: Would be nice to have a test watching for a file in tmp folder

aks-node-controller/app.go
```diff
@@ -95,6 +102,48 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
...
+				data, err := os.ReadFile(provisionJSONFilePath)
+				if err != nil {
+					return "", err
+				}
+				return string(data), nil
+			}
+
+		case err := <-watcher.Errors:
+			return "", fmt.Errorf("error watching file: %w", err)
+		}
```
r2k1, line 143: do we need a timeout?

r2k1: may be no necessary if ARM already have one on CSE execution

r2k1: actually, cancellation should be controlled through context.
I think we should just add something like:

```
case err := ctx.Done():
  return ctx.Err()
```

r2k1 (approved):

aks-node-controller/app.go
```diff
@@ -95,6 +102,48 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create watcher: %w", err)
+	}
+	defer watcher.Close()
+
+	// Watch the directory containing the provision complete file
+	dir := filepath.Dir(provisionCompleteFilePath)
+	err = os.MkdirAll(dir, 0755) // create the directory if it doesn't exist
+	if err != nil {
+		return "", err
```
r2k1, line 124: I usually try to wrap all errors from external packages.

Otherwise, you may encounter a cryptic error not searchable in the codebase.

aks-node-controller/app.go
lilypan26: Going to add this in a follow-up PR, and maybe implement the IO interface as you suggested for mocking the file operations.

PR #5176, opened 2024-10-29
[node-bootstrapper] set ENV variables without intermediate template step 

r2k1:
Remove the intermediate template step when setting environment variables in the `node-bootstrapper` package.
Type-safety is lost during templating.

This should reduce number of changes required to modify ENV variables and eliminate errors when a function string reference isn't updated after a code change (introduce type-safety).

It also moves us 1 step closer to shell script removal.


Devinwong:
Didn't look into cmd can set env before executing. This way is definitely better. Thanks for doing this.

node-bootstrapper/parser/parser_test.go
```diff
@@ -538,13 +501,13 @@ func getNBCInstance(jsonFilePath string) *nbcontractv1.Configuration {
 	return nBCB.GetNodeBootstrapConfig()
 }
 
-func generateTestDataIfRequested(t *testing.T, folder, cseCmd string) {
+func generateTestDataIfRequested(t *testing.T, folder string, cmd *exec.Cmd) {
 	if os.Getenv("GENERATE_TEST_DATA") == "true" {
```
Devinwong, line 505: Just seeing this so asking, though it's not part of this PR.
Where is this env var set? ADO pipeline?

Devinwong (approved):

node-bootstrapper/parser/parser_test.go
r2k1: idk. It looks weird for me. I would delete it.

PR #5178, opened 2024-10-29
bug: reject requests where the custom labels are too long for k8s

timmy-wright:
… lengths

**What type of PR is this?**

/kind bug

Sometimes custom labels are too long for k8s and this causes the node to fail to join the cluster. Rather than wait for the node to never join, we reject requests for bootstrapping information when the custom labels are too long.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
Truncate custom K8s labels to 63 characters.
```

### DIFF
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -101,6 +101,10 @@ const (
 	Linux   OSType = "Linux"
 )
 
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+const MaxK8sLabelNameLength = 63
+const MaxK8sLabelValueLength = 63
+
 // KubeletDiskType describes options for placement of the primary kubelet partition.
 // docker images, emptyDir volumes, and pod logs.
 type KubeletDiskType string
@@ -1186,6 +1190,13 @@ func (a *AgentPoolProfile) IsAvailabilitySets() bool {
 	return strings.EqualFold(a.AvailabilityProfile, AvailabilitySet)
 }
 
+func truncateString(str string, length int) string {
+	if len(str) > length {
+		return str[:length]
+	}
+	return str
+}
+
 // GetKubernetesLabels returns a k8s API-compliant labels string for nodes in this profile.
 func (a *AgentPoolProfile) GetKubernetesLabels() string {
 	var buf bytes.Buffer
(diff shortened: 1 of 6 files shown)


djsly (commented):
Wouldn't it be best to have the client reject the label instead of silently truncating the label name and/or value ? I have a feeling customer will waste a lot of time trying to understand why their selector isn't matching the labels. To simply realized later on that the last character of their label was removed. I think it might be best to understand which clients allows for this > 63 char string to be pass.  If we want to truncate at the receiving end. I would opt for a clearer name/label indicating that the string was too long and altered to make to compatible. Eg. My super long label - Invalid-lenght-

pkg/agent/datamodel/types.go
```diff
@@ -1198,7 +1209,9 @@ func (a *AgentPoolProfile) GetKubernetesLabels() string {
 	}
 	sort.Strings(keys)
 	for _, key := range keys {
-		buf.WriteString(fmt.Sprintf(",%s=%s", key, a.CustomNodeLabels[key]))
+		labelName := truncateString(key, MaxK8sLabelNameLength)
+		labelValue := truncateString(a.CustomNodeLabels[key], MaxK8sLabelValueLength)
+		buf.WriteString(fmt.Sprintf(",%s=%s", labelName, labelValue))
```
cameronmeissner, line 1214: we also add labels within the provisioning scripts themselves, we should take that into account here as well

cameronmeissner: oh nevermind I misunderstood, this on a per-label basis, not total length of the entire set of labels

cameronmeissner:
> Wouldn't it be best to have the client reject the label instead of silently truncating the label name and/or value ? I have a feeling customer will waste a lot of time trying to understand why their selector isn't matching the labels. To simply realized later on that the last character of their label was removed. I think it might be best to understand which clients allows for this > 63 char string to be pass. If we want to truncate at the receiving end. I would opt for a clearer name/label indicating that the string was too long and altered to make to compatible. Eg. My super long label - Invalid-lenght-

I figured RP would validate custom labels on the frontend and return a 4XX if something was incorrect, though I guess not?

pkg/agent/datamodel/types.go
```diff
@@ -1186,6 +1190,13 @@ func (a *AgentPoolProfile) IsAvailabilitySets() bool {
...
 
+func truncateString(str string, length int) string {
+	if len(str) > length {
+		return str[:length]
+	}
+	return str
+}
+
 // GetKubernetesLabels returns a k8s API-compliant labels string for nodes in this profile.
 func (a *AgentPoolProfile) GetKubernetesLabels() string {
```
djsly, line 1201: this is too risky I find here, we should instead truncate when we create the label since this Label comes from US, it's tied to the customHeader.  Client side should block it. or since we are dynamically creaeting it, I feel we should split the long label into multiple labels.

djsly (changes requested):

pkg/agent/datamodel/types.go
timmy-wright: Have changed the PR to reject long labels rather than truncate them.

cameronmeissner (changes requested):
blocking merge for CCOA

pkg/agent/datamodel/types.go
timmy-wright: Changed to reject requests.

PR #5187, opened 2024-10-30
fix: append existing env variables to provision cmd

lilypan26:
**What type of PR is this?**

Noticed scriptless e2es were failing now that bootstrap environment variables are generated in go binary https://github.com/Azure/AgentBaker/pull/5176, Fixed by appending existing system environment variables before triggering bootstrap cmd.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


node-bootstrapper/parser/parser.go
```diff
@@ -185,6 +186,7 @@ func BuildCSECmd(ctx context.Context, config *nbcontractv1.Configuration) (*exec
 	for k, v := range getCSEEnv(config) {
 		cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
 	}
-	sort.Strings(cmd.Env) // produce deterministic output
+	cmd.Env = append(cmd.Env, os.Environ()...) // append existing environment variables
```
r2k1, line 189: oh, it would take me ages to find this. Good catch. I didn't expect we need env variables from somewhere else.

Btw, I think it should be:

```
sort.Strings(cmd.Env)
cmd.Env = append(os.Environ(), cmd.Env...)
```

It some edge cases we may want to override the ENV variable.
And I think the last one has precedence.

r2k1 (approved):

PR #5191, opened 2024-10-31
Rename node-bootstrapper => aks-node-controller

r2k1:
/kind cleanup

`node-bootstrapper` been an interim name. We reached a consensus to rename it to `aks-node-controller` to better reflect its purpose.


vhdbuilder/packer/packer_source.sh
```diff
@@ -246,13 +246,13 @@ copyPackerFiles() {
...
-  NODE_BOOTSTRAPPER_DEST=/opt/azure/containers/node-bootstrapper
-  cpAndMode $NODE_BOOTSTRAPPER_SRC $NODE_BOOTSTRAPPER_DEST 755
+  AKS_NODE_CONTROLLER_SRC=/home/packer/aks-node-controller
+  AKS_NODE_CONTROLLER_DEST=/opt/azure/containers/aks-node-controller
+  cpAndMode $AKS_NODE_CONTROLLER_SRC $AKS_NODE_CONTROLLER_DEST 755
 
-  NODE_BOOTSTRAPPER_SERVICE_SRC=/home/packer/bootstrap.service
-  NODE_BOOTSTRAPPER_SERVICE_DEST=/etc/systemd/system/bootstrap.service
-  cpAndMode $NODE_BOOTSTRAPPER_SERVICE_SRC $NODE_BOOTSTRAPPER_SERVICE_DEST 0644
+  AKS_NODE_CONTROLLER_SERVICE_SRC=/home/packer/bootstrap.service
```
cameronmeissner, line 253: im okay either way, but should we rename `bootstrap.server` to `aks-node-controller.service` as well?

cameronmeissner (approved):

vhdbuilder/packer/packer_source.sh
r2k1: Makes sense. It's confused me before.

I think it's nice to make the relationship clear. It's unlikely we will need to change insides to another binary without renaming it.

lilypan26 (approved):

PR #4597, opened 2024-07-08
feat: do not download reference CNI plugins

paulgmiller:
**What type of PR is this?**
Feature (kind of more removing a feature)

**What this PR does / why we need it**:
Reference CNI plugins were formally able to be downloaded from a share however this just caused out of sync issues and was not really used since a node image upgrade was still required. This just always uses the version cached on the vhd

Very open to suggestions on which components.json is used, if it should cache arm64 and amd64 and how to not have to sync vesions between compoennts.json and cse.sh. 

**Which issue(s) this PR fixes**:
none

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)

- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #5198, opened 2024-11-02
Pmiller/dontdownlaodazurecni

paulgmiller:
**What type of PR is this?**

Do the same thing for azure cni as for refrence plugins till cni-installer kills it in 1.31
https://github.com/Azure/AgentBaker/pull/4597/files#diff-d6aeb8e845af4c32251915a17dc7c887d73cd628262e83118b9dcd65f7728c98

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_install.sh
+++ b/parts/linux/cloud-init/artifacts/cse_install.sh
@@ -361,19 +361,6 @@ evalPackageDownloadURL() {
     echo ""
 }
 
-downloadAzureCNI() {
-    mkdir -p ${1-$:CNI_DOWNLOADS_DIR}
-    # At VHD build time, the VNET_CNI_PLUGINS_URL is usually not set. 
-    # So, we will get the URL passed from install-depenencies.sh which is actually from components.json
-    # At node provisioning time, if AKS-RP sets the VNET_CNI_PLUGINS_URL, then we will use that.
-    VNET_CNI_PLUGINS_URL=${2:-$VNET_CNI_PLUGINS_URL}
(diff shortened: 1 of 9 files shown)


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -508,18 +495,58 @@ installCNI() {
     chown -R root:root $CNI_BIN_DIR
 }
 
+# this will never be called after 1.31 hopefully as daemonset will install windows.
```
AbelHu, line 498: I think that you may need to update https://github.com/Azure/AgentBaker/blob/7342099911c73b0c8c4ed026e54e22b384f15a8a/parts/windows/kuberneteswindowssetup.ps1#L390 and https://github.com/Azure/AgentBaker/blob/dev/staging/cse/windows/azurecnifunc.ps1#L1 for Windows.
Suggest passing empty value to `$global:VNetCNIPluginsURL` and updating Install-VnetPlugins to handle the empty URL when it does not need to install Windows Azure CNI.

pkg/proto/nbcontract/v1/networkconfig.proto
```diff
@@ -11,10 +11,10 @@ message NetworkConfig {
   NetworkPolicy network_policy = 2;
 
   // URL to the vnet cni plugins tarball.
-  string vnet_cni_plugins_url = 3;
+  reserved 3;
```
tyler-lloyd, line 14: cool people might do `reserved 3,4` or `reserved 3 to 4` https://protobuf.dev/programming-guides/proto3/#reserved-field-numbers

PR #5201, opened 2024-11-04
change from bool to enum

timmy-wright:
**What type of PR is this?**

/kind feature

Changes from a bool for auth bootstrapping method to an enum, which is more extensible in future, and is needed by stretch.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,13 +1,26 @@
 syntax = "proto3";
 package aksnodeconfig.v1;
- 
+
+enum BootstrappingAuthMethod  {
```
r2k1, line 4: It's a good practice to have zero option as `unspecified` in protobuf enums. (We probably should have a linter for protobuf?)

Otherwise, it will always default to the first option, which frequently is unexpected behaviour.

r2k1: It's ok to default to bootstrap token. But it should be explicitly, rather than a side effect of JSON unmarshalling.

r2k1: Also, other enums use different style

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.pb.go
```diff
@@ -20,17 +20,78 @@ const (
 	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
 )
 
+type BootstrappingAuthMethod int32
+
+const (
+	//nolint:gosec // this is a const string to use in switch statements, not hardcoded credentials
+	BootstrappingAuthMethod_UseTLSBootstrapToken      BootstrappingAuthMethod = 0
```
djsly, line 27: Could we use `= iota` instead of listing all values.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.pb.go
```diff
@@ -20,17 +20,78 @@ const (
...
+	BootstrappingAuthMethod_UseSecureTLSBootstrapping BootstrappingAuthMethod = 1
+	BootstrappingAuthMethod_UseArcMsiToMakeCSR        BootstrappingAuthMethod = 2
+	BootstrappingAuthMethod_UseAzureMsiToMakeCSR      BootstrappingAuthMethod = 3
+	BootstrappingAuthMethod_UseArcMsiDirectly         BootstrappingAuthMethod = 4
+	BootstrappingAuthMethod_UseAzureMsiDirectly       BootstrappingAuthMethod = 5
+)
+
+// Enum value maps for BootstrappingAuthMethod.
+var (
+	BootstrappingAuthMethod_name = map[int32]string{
```
djsly, line 37: what is this used for ?

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.pb.go
```diff
@@ -20,17 +20,78 @@ const (
...
+		0: "UseTLSBootstrapToken",
+		1: "UseSecureTLSBootstrapping",
+		2: "UseArcMsiToMakeCSR",
+		3: "UseAzureMsiToMakeCSR",
+		4: "UseArcMsiDirectly",
+		5: "UseAzureMsiDirectly",
+	}
+	BootstrappingAuthMethod_value = map[string]int32{
```
djsly, line 45: what is this used for ?

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
djsly: https://www.sohamkamani.com/golang/enums/#default-enum-values

which explain the usage of `iota` as well as the `default` value `0` used with an undefined value

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,13 +1,26 @@
...
   // Enable secure TLS bootstrapping for the node.
-  optional bool enable_secure_tls_bootstrapping = 1;
+  BootstrappingAuthMethod bootstrapping_auth_method = 1;
 
   // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
-  string tls_bootstrapping_token = 2; 
+  string tls_bootstrapping_token = 2;
+
+  // Only used when secure TLS bootstrapping is enabled or one of the Azure/Arc methods. This is the appserver appid that the node will use to bootstrap.
+  string custom_aad_appserver_appid = 3;
```
cameronmeissner, line 22: while you're at it please rename to `custom_aad_resource`, I originally planned on doing this in one of my previous PRs that has yet to merge

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,13 +1,26 @@
...
 
-  // Only used when secure TLS bootstrapping is enabled. This is the appserver appid that the node will use to bootstrap.
-  string custom_secure_tls_bootstrapping_appserver_appid = 3; 
+  // Only used when one of the Azure/Arc methods is enabled. This is the client MSI id that the node will use to bootstrap.
```
cameronmeissner, line 24:
```suggestion
  // Only used when one of the Azure/Arc methods is enabled. This is the client ID of the MSI the node will use to bootstrap.
```

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,13 +1,26 @@
 syntax = "proto3";
 package aksnodeconfig.v1;
- 
+
+enum BootstrappingAuthMethod  {
+  //nolint:gosec // this is a const string to use in switch statements, not hardcoded credentials
+  UseTLSBootstrapToken = 0;
```
cameronmeissner, line 6: I think we should add descriptions for each of these values

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,13 +1,26 @@
...
+}
+
 message TLSBootstrappingConfig {
   // Enable secure TLS bootstrapping for the node.
-  optional bool enable_secure_tls_bootstrapping = 1;
+  BootstrappingAuthMethod bootstrapping_auth_method = 1;
 
   // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
-  string tls_bootstrapping_token = 2; 
+  string tls_bootstrapping_token = 2;
```
cameronmeissner, line 19: im not quite as familiar with proto best practices, but I think I read that marking each field as optional and then adding explicit comments to indicate required fields is generally considered best-practice

cameronmeissner: note that we eventually plan on removing all references to bootstrap tokens / vanilla TLS bootstrapping once secure TLS bootstrapping has replaced it

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
```diff
@@ -1,13 +1,26 @@
...
- 
+
+enum BootstrappingAuthMethod  {
+  //nolint:gosec // this is a const string to use in switch statements, not hardcoded credentials
+  UseTLSBootstrapToken = 0;
+  UseSecureTLSBootstrapping = 1;
+  UseArcMsiToMakeCSR = 2;
+  UseAzureMsiToMakeCSR = 3;
+  UseArcMsiDirectly = 4;
+  UseAzureMsiDirectly = 5;
```
cameronmeissner, line 11: can we drop the `Use` prefix since it's not adding any parameterization / distinctive value?

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.pb.go
r2k1: you're commenting on generated code we have no control over

btw, I don't `iota` is great for public enums. One day you change order of constants (or remove one of them) and will swap the values in surprising hard-to-track ways.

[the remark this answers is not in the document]
timmy-wright: Autogenerated code.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.pb.go
timmy-wright: protobuf unmarshalling. This is autogenerated toce.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
timmy-wright: Done.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
timmy-wright: Have marked the 3 string fields as optional. Some Stretch clusters might want to use bootstrap tokens, so we prob shouldn't remove it entirely.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
timmy-wright: Good idea.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.pb.go
timmy-wright: Autogenerated code.

pkg/proto/aksnodeconfig/v1/tlsbootstrappingconfig.proto
timmy-wright: Leaving as an int as this is consistent with the other enums in the protobuf specification.

pkg/proto/aksnodeconfig/v1/bootstrappingconfig.proto
```diff
@@ -0,0 +1,43 @@
...
+  // Only required until Secure TLS bootstrapping in place. Would use kubelet identity after that.
+  optional string tls_bootstrapping_token = 3;
+
+  // Only used when secure TLS bootstrapping is enabled or one of the Azure/Arc methods. This is the appserver appid that the node will use to bootstrap.
+  optional string custom_aad_resource = 4;
+
+  // Only used when one of the Azure/Arc methods is enabled. This is the client ID of the MSI that the node will use to bootstrap.
+  optional string custom_aad_client_id = 5;
+
+  enum BootstrappingAuthMethod  {
```
r2k1, line 20: Add an unspecified option. So, we can distinguish between user intentionally setting BOOTSTRAP_TOKEN vs default value.

pkg/proto/aksnodeconfig/v1/bootstrappingconfig.proto
```diff
@@ -0,0 +1,43 @@
...
+    ARC_MSI = 2;
+    // Nodes running inside Azure can use the Azure Arc MSI to authenticate to an API server. This only works when the cluster is
+    // using AAD authentication.
+    AZURE_MSI = 3;
+  }
+
+  enum ClusterJoinMethod  {
+    // The default behaviour is for the node to make a certificate signing request (CSR) and then
+    // use that CSR for ongoing communication.
+    GENERATE_CSR = 0;
```
r2k1, line 39: Same here

r2k1 (approved):

PR #5215, opened 2024-11-04
fix: add retry for initial trivy scan to handle 429's from ghcr on DB update

cameronmeissner:
… update

**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/trivy-scan.sh
+++ b/vhdbuilder/packer/trivy-scan.sh
@@ -129,7 +129,7 @@ chmod a+x ${MODULE_NAME}
 export PATH="$(pwd):$PATH"
 
 # we do a delayed retry here since it's possible we'll get rate-limited by ghcr.io, which hosts the vulnerability DB
-retrycmd_if_failure 10 60 600 ./trivy --scanners vuln rootfs -f json --skip-dirs /var/lib/containerd --ignore-unfixed --severity ${SEVERITY} -o "${TRIVY_REPORT_ROOTFS_JSON_PATH}" /
+retrycmd_if_failure 10 30 600 ./trivy --scanners vuln rootfs -f json --skip-dirs /var/lib/containerd --ignore-unfixed --severity ${SEVERITY} -o "${TRIVY_REPORT_ROOTFS_JSON_PATH}" /
 
 if [[ -f ${TRIVY_REPORT_ROOTFS_JSON_PATH} ]]; then
     ./vuln-to-kusto-vhd scan-report \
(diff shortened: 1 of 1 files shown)


jason1028kr (approved):

ganeshkumarashok:
Can we consider mirroring it from ghcr to mcr? Is there one that already exists? That could help us avoid throttling.

vhdbuilder/packer/trivy-scan.sh
```diff
@@ -114,7 +128,9 @@ chmod a+x ${MODULE_NAME}
 # shellcheck disable=SC2155
 export PATH="$(pwd):$PATH"
 
-./trivy --scanners vuln rootfs -f json --skip-dirs /var/lib/containerd --ignore-unfixed --severity ${SEVERITY} -o "${TRIVY_REPORT_ROOTFS_JSON_PATH}" /
+# we do a delayed retry here since it's possible we'll get rate-limited by ghcr.io, which hosts the vulnerability DB
+retrycmd_if_failure 10 60 600 ./trivy --scanners vuln rootfs -f json --skip-dirs /var/lib/containerd --ignore-unfixed --severity ${SEVERITY} -o "${TRIVY_REPORT_ROOTFS_JSON_PATH}" /
```
ganeshkumarashok, line 132: 6600 seconds/110 minutes is the max timeout time for: retrycmd_if_failure 10 60 600.  That's a long time, more than certain VHD builds. I think we need to 1) reduce the timeout time, 2) (more medium-term) consider preventing a trivy scan throttling issue from blocking the full build.

cameronmeissner:
`consider using MS Defender tooling instead`

would this be an alternative to trivy altogether? or just another data source we can point trivy towards?

vhdbuilder/packer/trivy-scan.sh
cameronmeissner: I highly doubt we'd ever run into trivy stages taking that long (if all 10 retries failed on 429 then it'd be a little over 10 minutes until we see the failure in the worst case with the current config)

though overall I agree this is a bandaid and not a solution

ganeshkumarashok: I've had to fix such errors for container image pull retries - it's quite time consuming to wait till failure (even for a different reason). And the ADO build logs don't show up until it completes, so it's hard to know why it failed.

ganeshkumarashok:
> `consider using MS Defender tooling instead`
> 
> would this be an alternative to trivy altogether? or just another data source we can point trivy towards?

if we go this route, it'll be an alternate to trivy altogether

vhdbuilder/packer/trivy-scan.sh
cameronmeissner: reduced retry wait time to 30s from 60s

PR #5224, opened 2024-11-05
Support Trivy DB fallbacks and use MAR as primary DB location

djsly:
**What type of PR is this?**
/kind bug

Augment the Tricy CLI with a list of DB registries to ensure that we have fallbacks in case one registry throttles us.


cameronmeissner (changes requested):
I think we can close this?

PR #5249, opened 2024-11-12
Windows E2E

r2k1:
/kind test

Implement windows E2E in golang instead of bash.
Re-use Linux infrastructure and e2e testing suite.


e2e/aks_model.go
```diff
@@ -243,7 +243,7 @@ func createPrivateAzureContainerRegistry(ctx context.Context, t *testing.T, reso
 	}
 
 	t.Logf("Private Azure Container Registry created\n")
-	if err := addCacheRuelsToPrivateAzureContainerRegistry(ctx, t, config.ResourceGroupName, config.PrivateACRName); err != nil {
+	if err := addCacheRuelsToPrivateAzureContainerRegistry(ctx, t, config.ResourceGroup, config.PrivateACRName); err != nil {
```
djsly, line 246: I would keep Name, sometimes api requires ResourceGroupID...  Name and ID are the two strings that are often expected and not being clear about it could results in user errore

e2e/config/azure.go
```diff
@@ -71,24 +74,24 @@ func NewAzureClient() (*AzureClient, error) {
...
-				}).DialContext,
-				ForceAttemptHTTP2:     true,
-				MaxIdleConns:          100,
-				IdleConnTimeout:       90 * time.Second,
-				TLSHandshakeTimeout:   10 * time.Second,
-				ExpectContinueTimeout: 1 * time.Second,
-				TLSClientConfig: &tls.Config{
-					MinVersion: tls.VersionTLS12,
-				},
+		//Transport: armbalancer.New(armbalancer.Options{
```
djsly, line 77: any reasons we are keeping those comments ?

e2e/config/azure.go
```diff
@@ -300,22 +334,23 @@ func (a *AzureClient) createBlobStorageAccount(ctx context.Context) error {
...
+	_, err := a.StorageContainers.Create(ctx, ResourceGroup, Config.BlobStorageAccount(), Config.BlobContainer, armstorage.BlobContainer{}, nil)
 	if err != nil {
 		return fmt.Errorf("create blob container: %w", err)
 	}
 	return nil
 }
 
-func (a *AzureClient) assignReaderRoleToBlobStorage(ctx context.Context, principalID *string) error {
-	scope := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Storage/storageAccounts/%s", Config.SubscriptionID, ResourceGroupName, Config.BlobStorageAccount())
+func (a *AzureClient) assignRolesToVMIdentity(ctx context.Context, principalID *string) error {
```
djsly, line 344: I only see a single Role.  are we expecting to set more roles ? I would expect the function to receive a list of roles if we want to make it more generic

e2e/config/azure.go
```diff
@@ -324,10 +359,105 @@ func (a *AzureClient) assignReaderRoleToBlobStorage(ctx context.Context, princip
...
+}
+
+func (a *AzureClient) ensureReplication(ctx context.Context, image *Image, version *armcompute.GalleryImageVersion) error {
+	if replicatedToCurrentRegion(version) {
+		return nil
+	}
+	return a.replicateToCurrentRegion(ctx, image, version)
+}
+
+func (a *AzureClient) replicateToCurrentRegion(ctx context.Context, image *Image, version *armcompute.GalleryImageVersion) error {
```
djsly, line 419: nit: `replicateImageVersionToCurrentRegion`

e2e/config/azure.go
```diff
@@ -324,10 +359,105 @@ func (a *AzureClient) assignReaderRoleToBlobStorage(ctx context.Context, princip
 		if errors.As(err, &respError) && respError.StatusCode == http.StatusConflict {
 			return nil
 		}
-		return fmt.Errorf("assign reader role: %w", err)
+		return fmt.Errorf("assign Storage Blob Data Contributor role: %w", err)
 	}
 	return nil
+}
+
+func (a *AzureClient) LatestSIGImageVersionByTag(ctx context.Context, image *Image, tagName, tagValue string) (VHDResourceID, error) {
```
djsly, line 367: nit: `GetLatestSIGImageVersionByTag`

e2e/config/azure.go
```diff
@@ -324,10 +359,105 @@ func (a *AzureClient) assignReaderRoleToBlobStorage(ctx context.Context, princip
...
+
+func (a *AzureClient) ensureReplication(ctx context.Context, image *Image, version *armcompute.GalleryImageVersion) error {
+	if replicatedToCurrentRegion(version) {
+		return nil
+	}
+	return a.replicateToCurrentRegion(ctx, image, version)
+}
+
+func (a *AzureClient) replicateToCurrentRegion(ctx context.Context, image *Image, version *armcompute.GalleryImageVersion) error {
+	galleryImageVersion, err := armcompute.NewGalleryImageVersionsClient(image.Gallery.SubscriptionID, a.Credential, a.ArmOptions)
```
djsly, line 420: why moving the Client Constructor inside the function that uses it ? while all other clients are at the top ?

djsly: I'm guessing this is to support passing different SubID, compared to the global client which are configured once ?

e2e/config/azure.go
```diff
@@ -345,3 +475,19 @@ func DefaultRetryOpts() policy.RetryOptions {
 		},
 	}
 }
+
+func replicatedToCurrentRegion(version *armcompute.GalleryImageVersion) bool {
```
djsly, line 479: since this member function isn't part of a GalleryImageVersion class, I would renamed it to be more explicit
nit: `isImageVersionReplicatedToCurrentRegion`

e2e/config/config.go
```diff
@@ -54,21 +56,36 @@ func (c *Configuration) BlobStorageAccountURL() string {
 }
 
 func (c *Configuration) GalleryResourceID() string {
-	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s", c.GallerySubscriptionID, c.GalleryResourceGroupName, c.GalleryName)
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s", c.GallerySubscriptionID, c.GalleryResourceGroup, c.GalleryName)
```
djsly, line 59: same, ResourceGroupName and ResourceGroupID are often used in the SDK. I tend to prefer the explicit naming.

e2e/config/vhd.go
```diff
@@ -8,57 +8,81 @@ import (
...
 )
 
+var (
+	linuxGallery = &Gallery{
+		SubscriptionID: Config.GallerySubscriptionID,
+		ResourceGroup:  Config.GalleryResourceGroup,
+		Name:           Config.GalleryName,
+	}
+
+	windowsGallery = &Gallery{
```
djsly, line 24: can we avoid hardcoding ? and provide distinct configuration to allow switching Gallery dynamically,

e2e/config/vhd.go
```diff
@@ -8,57 +8,81 @@ import (
...
 )
 
 const (
 	noSelectionTagName = "abe2e-ignore"
 )
 
+var (
+	linuxGallery = &Gallery{
+		SubscriptionID: Config.GallerySubscriptionID,
+		ResourceGroup:  Config.GalleryResourceGroup,
```
djsly, line 20: nit: `ResourceGroupName`

e2e/config/vhd.go
```diff
@@ -8,57 +8,81 @@ import (
...
+	windowsGallery = &Gallery{
+		SubscriptionID: "4be8920b-2978-43d7-ab14-04d8549c1d05",
+		ResourceGroup:  "AKS-Windows",
+		Name:           "AKSWindows",
+	}
+)
+
+type Gallery struct {
+	SubscriptionID string
+	ResourceGroup  string
```
djsly, line 33: nit: `ResourceGroupName`

e2e/kube.go
```diff
@@ -59,8 +51,28 @@ func getClusterKubeClient(ctx context.Context, resourceGroupName, clusterName st
 	restConfig.GroupVersion = &schema.GroupVersion{
 		Version: "v1",
 	}
+	// it's test cluster avoid unnecessary rate limiting
```
djsly, line 54: were you affected by rate limiting during tests ?  I'm quite surprised we are issuing that many requests within a short time span

e2e/vmss.go
```diff
@@ -29,53 +31,66 @@ import (
...
 
 	if isAzureCNI {
-		err = addPodIPConfigsForAzureCNI(&model, vmssName, cluster)
-		require.NoError(t, err)
+		err = addPodIPConfigsForAzureCNI(&model, s.Runtime.VMSSName, cluster)
+		require.NoError(s.T, err)
 	}
 
-	scenario.PrepareVMSSModel(ctx, t, &model)
+	if s.VHD.Windows() {
```
djsly, line 62: could we move this logic inside the `getBaseVMSSModel` instead ?

e2e/vmss.go
```diff
@@ -90,40 +105,167 @@ func skipTestIfSKUNotAvailableErr(t *testing.T, err error) {
...
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Minute)
+	defer cancel()
+	defer deleteVMSS(ctx, s)
+
+	if s.VHD.Windows() {
+		extractLogsFromWindowsVM(ctx, s)
+		return
+	}
+
+	logFiles, err := extractLogsFromVM(ctx, s)
```
djsly, line 119: same, could we embed the logic for linux vs windows inside the `extractLogsFromVM`

I would prefer to have `extractLogsFromWindowsVM` and `extractLogsFromLinuxVM` OR even better just `extractLogsFromVM` which uses the Scenario `s` to decide internally how to extract the logs. 
```
extractLogsFromVM() {

if s.VHD.windows
  extractLogsFromWindowsVM
else
  extractLogsFromLinuxVM

}

File: e2e/vmss.go [hunk 15]
```diff
@@ -230,22 +372,31 @@ func getNewRSAKeyPair() (privatePEMBytes []byte, publicKeyBytes []byte, e error)
...
 	return name
 }
 
+func generateVMSSNameWindows(t *testing.T) string {
+	// windows has a limit of 9 characters for VMSS name
+	// and doesn't allow "-"
+	return fmt.Sprintf("win%s", randomLowercaseString(4))
+}
+
+func generateVMSSName(s *Scenario) string {
```
djsly, line 393: nice, I would expect this pattern to be similar for most calls.

djsly (commented):
left a few commments

e2e/config/azure.go
r2k1: I'm still not fully understand if it's actually important for us or not. I think it's not, but here is a chance we will need to resurrect it in the future.

Just want to make it visible until it confirmed we don't have any issues by removing it.

e2e/config/azure.go
r2k1: Roles are defined inside the function. I think I went through a few iterations and eventually came back to a single one.

e2e/config/azure.go
r2k1: I don't like `Get/Set`. Feels redundant to me. Technically half of the methods can have this prefix.

e2e/config/azure.go
r2k1: Yes. Unlike others, subscription id has to me dynamic for this function.

e2e/config/azure.go
r2k1: I don't have a strong opinion on whether we should prefix all `bool` function with `Is`.

But I don't think we're doing it currently.

e2e/config/vhd.go
r2k1: Generally, specifically for testing I find it's easier to define configuration inside tests (unless we want to have different contexts for some reason).

It's just easier to track and modify. An alternative typically requires more jumping through the code, then to the YAML, then trying to find the definition somewhere inside ADO UI.

But it looks like someone added the logic for Linux while I've been working on it. So, I guess there is a reason / migration work going on.

e2e/kube.go
r2k1:
```
DefaultQPS: 5
DefaultBurst: 10
```

Yes, I hit it. I think when we probably run more than 5 tests concurrently on the same cluster. And we poll status with a second interval.

r2k1: I'm not concerned with overloading k8s api

parts/windows/windowscsehelper.ps1
```diff
@@ -183,7 +183,8 @@ function DownloadFileOverHttp {
     )
 
     # First check to see if a file with the same name is already cached on the VHD
-    $fileName = [IO.Path]::GetFileName($Url)
+    $cleanUrl = $Url.Split('?')[0]
```
junjiezhang1997, line 186: I'm not sure why we need this change. In the past, we may need to use url with SAS token which contains "?". It seems that we don't use url that contains "?" now so maybe this change will not influence anything.

r2k1: We use private storage account in tests. So, I need to use SAS keys to download the archive in tests.

r2k1: I tried to assign an MSI to the VM. But it seems like more changes are required in the windows CSE to use it. And I'm not confident enough to change it.

e2e/vmss.go
```diff
@@ -338,6 +517,21 @@ func getBaseVMSSModel(name, sshPublicKey, customData, cseCmd string, cluster *Cl
...
+				config.Config.VMIdentityResourceID(): {},
+			},
+		}
+		model.Properties.VirtualMachineProfile.StorageProfile.OSDisk.OSType = to.Ptr(armcompute.OperatingSystemTypesWindows)
+		model.Properties.VirtualMachineProfile.OSProfile.LinuxConfiguration = nil
+		model.Properties.VirtualMachineProfile.ExtensionProfile.Extensions[0].Properties.Publisher = to.Ptr("Microsoft.Compute")
+		model.Properties.VirtualMachineProfile.ExtensionProfile.Extensions[0].Properties.Type = to.Ptr("CustomScriptExtension")
+		model.Properties.VirtualMachineProfile.ExtensionProfile.Extensions[0].Properties.TypeHandlerVersion = to.Ptr("1.10")
+		model.Properties.VirtualMachineProfile.OSProfile.AdminUsername = to.Ptr("azureuser")
+		model.Properties.VirtualMachineProfile.OSProfile.AdminPassword = to.Ptr("pwnedPassword123!")
```
junjiezhang1997, line 533: Do we need to store password as a secret variable and avoid exposing it in public for securtiy?

r2k1: I wouldn't expect the password is enough to connect to the Windows VM. It's not the case for Linux, but I don't know much Windows.

It's convenient to know it for debugging purposes if you ever need to connect it through Bastion.
I can generate a random password.

parts/windows/windowscsehelper.ps1
junjiezhang1997: Current windows e2e are using MSI with permission to upload and download files from a storage account. An example is here: https://github.com/Azure/AgentBaker/blob/master/e2e/windows/upload-cse-logs.ps1#L11

junjiezhang1997: Got it. Thank you for the clarification.

junjiezhang1997 (approved):

PR #5313, opened 2024-11-21
fix: change vhds for ubuntu2204 network isolated tests

AlisonB319:
**What type of PR is this?**

its not sustainable to use a pinned version for ubuntu2204 airgap tests. Removing on 2/3 and having discussions with RP on what to do for the 3rd. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/config/vhd.go
```diff
@@ -72,7 +72,7 @@ var (
 	}
 
 	// without kubelet, kubectl, credential-provider and wasm
-	VHDUbuntu2204Gen2ContainerdAirgapped = &Image{
+	VHDUbuntu2204Gen2ContainerdAirgappedPinnedVersion = &Image{
```
cameronmeissner, line 75: nit: `VHDUbuntu2204Gen2ContainerdAirgappedK8sNotCached` to align with the corresponding scenario name?

cameronmeissner (approved):

e2e/scenario_test.go
```diff
@@ -694,7 +694,7 @@ func Test_Ubuntu2204Gen2ContainerdAirgapped_K8sNotCached(t *testing.T) {
 		},
 		Config: Config{
 			Cluster: ClusterKubenetAirgap,
-			VHD:     config.VHDUbuntu2204Gen2ContainerdAirgapped,
+			VHD:     config.VHDUbuntu2204Gen2ContainerdAirgappedK8sNotCached,
```
r2k1, line 697: Do we really need this test?
I'd like to see the VHD to go away.

r2k1 (approved):

e2e/scenario_test.go
fseldow: from network isolated cluster feature perspective, i may still need such feature. Because we exist customer still use old VHD version (from frontend validator, we only block version < 20240906). So we need new CSE script does not generate new egress traffic for old vhd

PR #5316, opened 2024-11-21
fix airgapped cluster e2e image pull

fseldow:
**What type of PR is this?**

/kind test

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/config/config.go
```diff
@@ -15,7 +15,7 @@ var (
 	Azure             = mustNewAzureClient()
 	ResourceGroupName = "abe2e-" + Config.Location
 	VMIdentityName    = "abe2e-vm-identity"
-	PrivateACRName    = "privateacre2e" + Config.Location
+	PrivateACRName    = "xinhlprivateacre2e" + Config.Location
```
cameronmeissner, line 18: im assuming this is just for testing atm

fseldow: the name is for temp to avoid conflict with old tests

e2e/config/config.go
```diff
@@ -15,7 +15,7 @@ var (
 	Azure             = mustNewAzureClient()
 	ResourceGroupName = "abe2e-" + Config.Location
```
AlisonB319, line 16: We will want to revert these changes when merging the PR. Then once this PR is merged, delete our e2e resource group and re-run an airgap test and the resources will get recreated.

e2e/kube.go
```diff
@@ -93,7 +93,7 @@ func ensureDebugDaemonsets(ctx context.Context, t *testing.T, kube *Kubeclient,
 func getDebugDaemonsetTemplate(t *testing.T, deploymentName, targetNodeLabel string, isHostNetwork, isAirgap bool) string {
 	image := "mcr.microsoft.com/cbl-mariner/base/core:2.0"
 	if isAirgap {
-		image = fmt.Sprintf("%s.azurecr.io/aks/cbl-mariner/base/core:2.0", config.PrivateACRName)
```
AlisonB319, line 96: Why is this able to be reverted? 

If so, can you remove the isAirgap variable from the function since it isn't being used.

e2e/cluster.go
```diff
@@ -71,7 +71,7 @@ func ClusterKubenet(ctx context.Context, t *testing.T) (*Cluster, error) {
 
 func ClusterKubenetAirgap(ctx context.Context, t *testing.T) (*Cluster, error) {
 	clusterKubenetAirgapOnce.Do(func() {
-		cluster, err := prepareCluster(ctx, t, getKubenetClusterModel("abe2e-kubenet-airgap"), true)
+		cluster, err := prepareCluster(ctx, t, getKubenetClusterModel("abe2e-airgap"), true)
```
AlisonB319, line 74: Are you sure we want every e2e test running with the network plugin? I just want to make sure we are running the e2e's as close as possible to what a customer would be running without assuming they have anything installed

Since I'm assuming Test_Ubuntu2204Gen2ContainerdAirgapped_K8sNotCached needs the plugin to pass, maybe just that test should use the AzureNetwork cluster?

fseldow: in fact we do not support kubenet cluster for network isolated cluster. So that it should be azure network cluster. However, my latest test fails with node not ready. Still restricting the test range. Maybe in this pr we may remain kubenet and unblock the test first

e2e/config/config.go
```diff
@@ -15,7 +15,7 @@ var (
...
 	VMIdentityName    = "abe2e-vm-identity"
-	PrivateACRName    = "privateacre2e" + Config.Location
+	PrivateACRName    = "abe2eprivateacr" + Config.Location
```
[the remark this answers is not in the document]
fseldow: mark it as temp, since i need to ensure the flow can work with different name

e2e/kube.go
fseldow: it is to rever the prefix `aks` in this pr

AlisonB319: I'm not sure I quite understand why removing the aks prefix means we don't need to be able to pull cbl-mariner using our ACR? This might be why you are seeing node not ready?

e2e/cluster.go
AlisonB319: I just ran `Test_Ubuntu2204Gen2ContainerdAirgapped_K8sNotCached` on local (with a test rg and acr) and it was able to pass. Are you still seeing node not ready?

e2e/kube.go
AlisonB319: Oh my bad, when I left this comment I was using a vscode extension that showed a different diff of the entire isAirgap statement removed.

e2e/cluster.go
AlisonB319: For what its worth, the kubelet logs show 
```
Nov 26 09:14:11 abe2e-2024-11-26-bj4f-ubuntu2204gen2containerdairgappedk8000000 kubelet[3491]: I1126 09:14:11.829439    3491 csi_plugin.go:99] kubernetes.io/csi: Trying to validate a new CSI Driver with name: disk.csi.azure.com endpoint: /var/lib/kubelet/plugins/disk.csi.azure.com/csi.sock versions: 1.0.0
Nov 26 09:14:11 abe2e-2024-11-26-bj4f-ubuntu2204gen2containerdairgappedk8000000 kubelet[3491]: I1126 09:14:11.829463    3491 csi_plugin.go:112] kubernetes.io/csi: Register new plugin with name: disk.csi.azure.com at endpoint: /var/lib/kubelet/plugins/disk.csi.azure.com/csi.sock
Nov 26 09:15:32 abe2e-2024-11-26-bj4f-ubuntu2204gen2containerdairgappedk8000000 kubelet[3491]: E1126 09:15:32.347559    3491 kubelet_node_status.go:456] "Node not becoming ready in time after startup"
Nov 26 09:15:32 abe2e-2024-11-26-bj4f-ubuntu2204gen2containerdairgappedk8000000 kubelet[3491]: E1126 09:15:32.839128    3491 kubelet.go:2892] "Container runtime network not ready" networkReady="NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized"
```
for why node is not ready when using getAzureNetworkClusterModel

AlisonB319 (approved):

PR #5321, opened 2024-11-21
fix: resolve Azure Linux nvidia-container-toolkit posttrans error

henryli001:
**What type of PR is this?**

- This PR fixes an issue with Azure Linux AKS GPU where an error takes place within the posttrans phase when installing the nvidia-container-toolkit RPM package. that causes /etc/nvidia-container-runtime/config.toml to not be generated in the OS. The reason of the failure is because nvidia-ctk, the binary used to generate the config file is missing. This is because nvidia-ctk is provided by the nvidia-container-toolkit-base RPM package and this RPM is installed after nvidia-container-toolkit. This PR thus changes the RPM installation sequence to satisfy the package runtime requirements.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
lgtm I think, is this currently causing wide impact in prod? if so then it'd probably worth to cherry-pick this to master so it can roll out in next week's release

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -82,7 +82,7 @@ installNvidiaContainerToolkit() {
       MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION="1.16.2"
     fi
 
-    for nvidia_package in nvidia-container-toolkit-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION} nvidia-container-toolkit-base-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION} libnvidia-container-tools-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION} libnvidia-container1-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION}; do
+    for nvidia_package in libnvidia-container1-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION} libnvidia-container-tools-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION} nvidia-container-toolkit-base-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION} nvidia-container-toolkit-${MARINER_NVIDIA_CONTAINER_TOOLKIT_VERSION}; do
```
dallasd1, line 85: Can we add a comment to show this ordering is necessary

henryli001: Comment added.

henryli001:
> lgtm I think, is this currently causing wide impact in prod? if so then it'd probably worth to cherry-pick this to master so it can roll out in next week's release

This has not been causing impact in prod yet since no customer bug requests have been noticed, but there's a possibility that customer workloads might be impacted by this issue.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
ganeshkumarashok: Were you able to find official Nvidia docs that mentioned about this ordering dependency?

henryli001: I did not find any official doc that directly explains the dependency relationship between these packages. But a good source to refer to would be the official spec file for nvidia container toolkit: https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/packaging/rpm/SPECS/nvidia-container-toolkit.spec where the runtime requirements explain the dependency.

ganeshkumarashok (approved):

ganeshkumarashok:
lgtm!

PR #5325, opened 2024-11-22
feat: dedicated garbage collection pipeline

cameronmeissner:
**What type of PR is this?**

adds an initial implementation of a new garbage collection pipeline specifically aiming to ensure all ephemeral resources created within the new node SIG subscription are removed within a 24-hour window.

right now the garbage collection logic only applies to resource groups with a name matching the following regex: `"vhd-test*|vhd-scanning*|pkr-Resource-Group*"`, thus for now we're only targetting ephemeral resources created by VHD builds.

as a follow-up, we should probably migrate all back-fill deletion logic we run as a part of the VHD build to this new script/pipeline, and only have the VHD build jobs cleanup resources it itself created

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


anujmaheshwari1 (commented):
food for thought: do we need a separate pipeline for this? I understand the benefits here, which is to have scheduled runs every 24h, which ensures our resources are cleaned up.

however, if the rg-cleanup.sh becomes a part of the "Test, Scan and Cleanup" stage in our VHD build, which we are moving to running daily, it gives us the same benefit without a new pipeline.

another thing: I see that we can specify subscription_id as pipeline vars, which is good, but if we just do this as part of the daily builds, and then eventually weekly builds in AME, we are able to effectively GC that sub too, which is an added step here to change the sub_ID in the pipeline var?

i'd like to keep us to as few pipelines as possible, if we can achieve the same thing with the new system. what do you think @cameronmeissner @djsly ?

cameronmeissner:
> food for thought: do we need a separate pipeline for this? I understand the benefits here, which is to have scheduled runs every 24h, which ensures our resources are cleaned up.
> 
> however, if the rg-cleanup.sh becomes a part of the "Test, Scan and Cleanup" stage in our VHD build, which we are moving to running daily, it gives us the same benefit without a new pipeline.
> 
> another thing: I see that we can specify subscription_id as pipeline vars, which is good, but if we just do this as part of the daily builds, and then eventually weekly builds in AME, we are able to effectively GC that sub too, which is an added step here to change the sub_ID in the pipeline var?
> 
> i'd like to keep us to as few pipelines as possible, if we can achieve the same thing with the new system. what do you think @cameronmeissner @djsly ?

fair points, the main reason I wanted a separate pipeline was to prevent potential conflicts (each individual build job would run the same garbage collection routine), though I guess we already do this so it's probably not an issue

also note that we won't be running daily builds, or hopefully any builds at all within AME once we're done with the migration to TME

.pipelines/.vsts-garabge-collection.yaml
```diff
@@ -0,0 +1,20 @@
...
+  default: false
+
+jobs:
+- job: gc
+  displayName: Garbage Collection
+  steps:
+    - bash: |
+        chmod +x ./vhdbuilder/rg-cleanup.sh
+        ./vhdbuilder/rg-cleanup.sh
+      env:
```
djsly, line 17: we need to have a TAG that the GC ignores.  when debugging Build issue or e2e issues it might be best to leave some clusters/vmss online. 

we could look at force deleting after 7 days

cameronmeissner: added

r2k1: We had it for e2e and I didn't like. People often use for debugging and often don't clean after themselves.
So, the resources are wasted for a long time.

I like the simplicity of deleting everything overnight. No complicated logic, no opportunity to leave long-leaving resources, completely predictable.

A day should be enough to debug things. Next day things can be re-created again.

r2k1:
I like an idea of deleting everything at predictable time. (Preferable outside of everyone working hours)

.pipelines/.vsts-garabge-collection.yaml
r2k1: We probably need to have one for some edge-cases and long-living resources. But I'd like to avoid using such tags unless it's absolutely necessary. I don't want to have an option to extend the lifetime. I'd rather everything to either live forever or a single day.

cameronmeissner: I think the current strategy is okay since "by default", things will be removed within 24 hours of creation

you'd need to explicitly add the `gc.skip=true` to the resource to have it stay around longer, and even then it will still be forcibly removed after 7 days

cameronmeissner: we can also iterate on this if we find that other behavior works better in the future

vhdbuilder/packer/test/run-test.sh
```diff
@@ -32,7 +32,7 @@ else
     exit 1
   fi
 fi
-az group create --name $TEST_VM_RESOURCE_GROUP_NAME --location ${AZURE_LOCATION} --tags "source=AgentBaker" "branch=${GIT_BRANCH}"
+az group create --name $TEST_VM_RESOURCE_GROUP_NAME --location ${AZURE_LOCATION} --tags "source=AgentBaker,now=$(date +%s)" "branch=${GIT_BRANCH}"
```
r2k1, line 35: `now == created_at`?

.pipelines/.vsts-garabge-collection.yaml
```diff
@@ -0,0 +1,24 @@
+pool:
```
r2k1, line 1: I don't see triggers. Do you plan to run it manually for some time?

vhdbuilder/scripts/gc.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    if [ "${DRY_RUN,,}" == "true" ]; then
+        echo "DRY_RUN: az group delete -g $group --yes --no-wait"
+        return 0
+    fi
+
+    az group delete -g $group --yes --no-wait || return $?
+}
+
+main "$@"
```
r2k1, line 61: A random thought. Should we in the future write small golang programs?
(Of a have a single one with multiple packages to run)

vhdbuilder/scripts/gc.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+set -euxo pipefail
+
+[ -z "${SUBSCRIPTION_ID:-}" ] && echo "SUBSCRIPTION_ID must be set" && exit 1
+
+SKIP_TAG_NAME="${SKIP_TAG_NAME:-gc.skip}"
+SKIP_TAG_VALUE="${SKIP_TAG_VALUE:-true}"
```
r2k1, line 7: I would probably just embed it in the script.
It will be really hard to find out If someone going to change it in a pipeline.

r2k1 (approved):

vhdbuilder/packer/test/run-test.sh
r2k1: RGs can be also created by e2e. It's also can be done locally.
Do we need to update e2e tests to include the tag?

vhdbuilder/scripts/gc.sh
r2k1: Another random note.

I think you can fetch resource creation time using golang SDK without relying on tags. (But probably not the resource group)

.pipelines/.vsts-garabge-collection.yaml
cameronmeissner: we'll add a schedule to the pipeline definition itself

vhdbuilder/scripts/gc.sh
cameronmeissner: yeah... I actually began to write a go program for this but I opted to do this instead to save time (mainly since I'm primarily supposed to be working on other things)

I think we can revisit this once we start migrating other garbage collection logic out of the build pipeline

vhdbuilder/packer/test/run-test.sh
cameronmeissner: right now we only target RGs with names matching the regex `test*|vhd-scanning*|pkr-Resource-Group*`, and specifically within the node SIG subscription - e2e RGs/resources are still created in the ACS test subscription which is already onboarded to devinfra's garbage collector

all that to say we shouldn't need to change anything on the e2e side

PR #5326, opened 2024-11-22
chore: more descriptive pod logs

AlisonB319:
**What type of PR is this?**

example output 

```
 pollers.go:29: waiting for node abe2e-2024-11-22-xo8q-ubuntu2204gen2containerdairgappedk8 to be ready
    pollers.go:57: node abe2e-2024-11-22-xo8q-ubuntu2204gen2containerdairgappedk8000000 is ready
    pollers.go:76: -- pod metadata --
    pollers.go:77:    Name: test-pod-abe2e-2024-11-22-xo8q-ubuntu2204gen2containerdairgappedk8000000
                      Namespace: default
                      Node: 
                      Status: Pending
                      Start Time: <nil>
    pollers.go:78: -- container(s) info --
    pollers.go:80:    Container: mariner
                      Image: alisonprivateacre2ewestus3.azurecr.io/aks/cbl-mariner/busybox:2.0
                      Ports: [{ 0 80 TCP }]
    pollers.go:82: -- pod events --
    pollers.go:85:   Reason: FailedScheduling, Message: 0/2 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 1 node(s) had untolerated taint {node.cloudprovider.kubernetes.io/uninitialized: true}. preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling., Count: 0, Last Timestamp: 0001-01-01 00:00:00 +0000 UTC
    pollers.go:87: time before timeout: 27m19.024983333s
```

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/pollers.go
```diff
@@ -73,7 +73,19 @@ func waitUntilPodReady(ctx context.Context, kube *Kubeclient, podName string, t
 			remaining := time.Until(deadline)
 			if currentLogTime.Sub(lastLogTime) > logInterval {
 				// this logs every 5 minutes to reduce spam, iterations of poller are continuning as normal.
-				t.Logf("pod %s status: %s time before timeout: %v", podName, pod.Status.Phase, remaining)
+				t.Logf("-- pod metadata --\n")
+				t.Logf("   Name: %s\n              Namespace: %s\n              Node: %s\n              Status: %s\n              Start Time: %s\n", pod.Name, pod.Namespace, pod.Spec.NodeName, pod.Status.Phase, pod.Status.StartTime)
```
cameronmeissner, line 77: this data is nice, though are we also streaming pod container logs at any point? if not then might also be worth including in a log file alongside kubelet/cluster provisioning logs

cameronmeissner: yeah I think streaming pod logs on failures would be a helpful next step

cameronmeissner (approved):

e2e/pollers.go
AlisonB319: From what I could tell we are not streaming and pod logs. I tried to get something to work but it kept sig faulting lol. Might be worth creating a backlog item for generating a file based on streaming pod container logs?

r2k1: Logs are likely to be empty for majority of such failures.

e2e/pollers.go
```diff
@@ -73,7 +73,19 @@ func waitUntilPodReady(ctx context.Context, kube *Kubeclient, podName string, t
 			remaining := time.Until(deadline)
 			if currentLogTime.Sub(lastLogTime) > logInterval {
 				// this logs every 5 minutes to reduce spam, iterations of poller are continuning as normal.
-				t.Logf("pod %s status: %s time before timeout: %v", podName, pod.Status.Phase, remaining)
+				t.Logf("-- pod metadata --\n")
```
r2k1, line 76: I would just dump the `pod.Status` object

e2e/pollers.go
r2k1: FYI. Snippet from another place I used to extract debug info:

```
result.Pods = append(result.Pods, podDebugInfo{
			Name:                  pod.Name,
			Status:                pod.Status,
			ContainerLogs:         t.podLogs(ctx, pod, maxContainers, false),
			PreviousContainerLogs: t.podLogs(ctx, pod, maxContainers, true),
		})
	}
	return result
}

func (t *AgentTunnelKube) podLogs(ctx context.Context, pod v1.Pod, maxContainers int, previous bool) map[string]string {
	containers := make(map[string]string)
	for j, container := range pod.Spec.Containers {
		if j > maxContainers {
			break
		}
		containerLog, err := t.clientset.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &v1.PodLogOptions{
			// without specifying TailLines, log head instead of tail is returned
			// important information is nearly always in the tail
			TailLines:  pointer.Int64(15),
			LimitBytes: pointer.Int64(40000), // safety limit, don't trust user input
			Container:  container.Name,
			Previous:   previous,
		}).DoRaw(ctx)
		if err != nil { //+gocover:ignore:block can't test it
			logger.Error().Str("namespace", t.AgentNamespace).Str("pod_name", pod.Name).
				Str("container_name", container.Name).Err(err).Msg("failed to get container logs")
			continue
		}
		containers[container.Name] = string(containerLog)
	}
	return containers
}
```

r2k1 (approved):
can you extract it to a separate function, something like `extractDebugInfo`, `logPodStatus` (or anything else)

e2e/pollers.go
AlisonB319: imo dumping pod.Status isn't the most readable 

```
Status: {Pending [{PodScheduled False 0001-01-01 00:00:00 +0000 UTC 2024-11-22 14:37:56 -0800 PST Unschedulable 0/8 nodes are available: 2 node(s) didn't match Pod's node affinity/selector, 6 node(s) had untolerated taint {node.cloudprovider.kubernetes.io/uninitialized: true}. preemption: 0/8 nodes are available: 8 Preemption is not helpful for scheduling.}]     []  [] <nil> [] [] BestEffort []  []}
```

instead I've added condition.Reason and condition.Message which was the only other attributes that I think have relevant information
```
    pollers.go:120: -- pod metadata --
    pollers.go:121:    Name: test-pod-abe2e-2024-11-22-tmv6-ubuntu2204gen2containerdairgappedk8000000
                       Namespace: default
                       Node: 
                       Status: Pending
                       Start Time: <nil>
    pollers.go:123:     Reason: Unschedulable
    pollers.go:124:     Message: 0/2 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 1 node(s) had untolerated taint {node.cloudprovider.kubernetes.io/uninitialized: true}. preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling.
```

PR #5334, opened 2024-11-28
Make E2E validation more flexible.

r2k1:
Previously, E2E scenarios were limited to executing shell scripts on the VM.

I updated the tests to support more complex validations, such as running a GPU-enabled pod and verifying its functionality.

This change made the tests more flexible and simplified the overall logic.

### DIFF
--- a/e2e/scenario_helpers_test.go
+++ b/e2e/scenario_helpers_test.go
@@ -147,19 +147,19 @@ func createAndValidateVM(ctx context.Context, s *Scenario) {
 
 	s.T.Logf("node %s is ready, proceeding with validation commands...", s.Runtime.VMSSName)
 
-	vmPrivateIP, err := getVMPrivateIPAddress(ctx, s)
-
 	if s.Config.Validator != nil {
 		s.Config.Validator(ctx, s)
 	}
 
 	require.NoError(s.T, err, "get vm private IP %v", s.Runtime.VMSSName)
-	err = runLiveVMValidators(ctx, s.T, s.Runtime.VMSSName, vmPrivateIP, string(s.Runtime.SSHKeyPrivate), s)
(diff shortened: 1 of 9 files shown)


e2e/types.go
```diff
@@ -142,9 +142,7 @@ type Config struct {
 	// VMConfigMutator is a function which mutates the base VMSS model according to the scenario's requirements
 	VMConfigMutator func(*armcompute.VirtualMachineScaleSet)
 
-	// LiveVMValidators is a slice of LiveVMValidator objects for performing any live VM validation
-	// specific to the scenario that isn't covered in the set of common validators run with all scenarios
-	LiveVMValidators []*LiveVMValidator
+	Validator func(ctx context.Context, s *Scenario)
```
r2k1, line 145: This is the core difference

e2e/scenario_helpers_test.go
```diff
@@ -121,42 +121,46 @@ func maybeSkipScenario(ctx context.Context, t *testing.T, s *Scenario) {
 			t.Fatalf("could not find image for %q: %s", t.Name(), err)
 		}
 	}
-	t.Logf("running scenario %q with vhd: %q, tags %+v", t.Name(), vhd, s.Tags)
+	t.Logf("running scenario vhd: %q, tags %+v", vhd, s.Tags)
```
djsly, line 124: the scenario name wasn't useful ?

e2e/types.go
djsly: could you explain the main reason for this change ?

e2e/scenario_helpers_test.go
r2k1: each log message is associated with a test. It's just a duplication of the available information.

djsly (approved):

PR #5342, opened 2024-12-02
cherry-pick: VHD automation fixes and enhancements

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/scripts/automate_version_bump.sh
+++ b/vhdbuilder/scripts/automate_version_bump.sh
@@ -8,6 +8,7 @@ set -x
 
 NEW_IMAGE_VERSION="${IMAGE_VERSION:-""}"
 VHD_BUILD_ID="${VHD_BUILD_ID:-""}"
+PR_TARGET_BRANCH="${PR_TARGET_BRANCH:-master}"
 
 # This function takes the build ID and reads the queue time.
 # It then sanitizes the queue time in the format that the Canonical snapshot endpoint expects, which is 20230727T000000Z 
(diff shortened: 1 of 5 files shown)


vhdbuilder/scripts/automate_version_bump.sh
```diff
@@ -57,11 +58,11 @@ update_image_version() {
 }
 
 create_image_bump_pr() {
-    create_branch $BRANCH_NAME
+    create_branch $BRANCH_NAME $PR_TARGET_BRANCH
     update_image_version
 
     set +x
-    create_pull_request $NEW_IMAGE_VERSION $GITHUB_PAT $BRANCH_NAME $PR_TITLE
+    create_pull_request $NEW_IMAGE_VERSION $GITHUB_PAT $BRANCH_NAME $PR_TARGET_BRANCH $PR_TITLE
```
timmy-wright, line 65: nit: put the variables inside quotes just in case they have a space in the values. That way bash won't split the strings and pass the wrong args to create_pull_request.

I know that most of those values can't have spaces, so it's a very minor nit.

timmy-wright (approved):

vhdbuilder/scripts/automate_version_bump.sh
cameronmeissner: yep that's fair, I think we should more strictly lint our bash scripts in general (until we're fully in go :))

I'd opt for taking care of this in a separate PR we merge into dev then cherry-pick to master however

PR #5355, opened 2024-12-03
fix: stop dnf from overwriting motd on azure linux nodes if motd is enabled on the node

UtheMan:


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -247,6 +247,9 @@ else
 fi
 
 if [[ "${MESSAGE_OF_THE_DAY}" != "" ]]; then
+    if isMarinerOrAzureLinux "$OS"; then
+      sed -i "s/emit_via = motd/emit_via = stdio/g" /etc/dnf/automatic.conf
```
anujmaheshwari1, line 251: hmm not a big fan on in file updates but I guess its fine, I see that we already check the file status in e2e, so its ok

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -323,6 +323,19 @@ testAuditDNotPresent() {
   echo "$test:Finish"
 }
 
+testMessageOfTheDayComponentsAzureLinux() {
+  os_sku=$1
+  if [[ "$os_sku" == "CBLMariner" || "$os_sku" == "AzureLinux" ]]; then
```
anujmaheshwari1, line 328: nit: can we use `isMarinerOrAzureLinux` this function instead?

parts/linux/cloud-init/artifacts/cse_main.sh
UtheMan: Context: https://github.com/microsoft/azurelinux/commit/56ffe4180550364cce739a07645539ebbd774eb8#diff-b95e1980d5c078a15548240d96a5ba7204b49926ac9303b0f8ca2dfbd4221a3eR55

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -323,6 +323,19 @@ testAuditDNotPresent() {
...
+    # this file controls how /etc/motd file is updated by dnf on azure linux nodes
+    # in CSE, if MotD is enabled, we modify that file to stop dnf from overwriting the user-specified message
+      if [[ ! -f /etc/default/automatic.conf ]]; then
```
anujmaheshwari1, line 331: is this file ever not present on azlinux/mariner?

UtheMan: That I don't know. But since we are relying on it to handle motd cases, I figured that we should have a test ensuring it is there and in case it disappears - investigate. Since those config files are coming directly from Azure Linux.

anujmaheshwari1 (approved):

vhdbuilder/packer/test/linux-vhd-content-test.sh
UtheMan: Will leave as is, that helper is in a completely different place, would need to source 600+ lines long script or refactor more.

vhdbuilder/packer/test/linux-vhd-content-test.sh
UtheMan: Discussed offline, seems like this test is incorrect - those files are populated later on, so we will instead rely on cse checks and e2e tests

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -247,6 +247,9 @@ else
 fi
 
 if [[ "${MESSAGE_OF_THE_DAY}" != "" ]]; then
+    if isMarinerOrAzureLinux "$OS" && [ -f /etc/dnf/automatic.conf ]; then
+      sed -i "s/emit_via = motd/emit_via = stdio/g" /etc/dnf/automatic.conf
```
djsly, line 251: where does it go when `stdio` is used ?

UtheMan: Context: https://dnf.readthedocs.io/en/latest/automatic.html 
just to standard output, i believe this particualr part of dnf is triggered by a systemd unit so that would go into journald

PR #5358, opened 2024-12-03
cleanup: combine Linux/Windows test E2E pipelines

r2k1:
/kind test

Windows shell-based E2E are migrated to the golang.

This change updates windows pipelines to use new tests. It also makes them similar to Linux.


timmy-wright (approved):

.pipelines/e2e.yaml
```diff
@@ -16,12 +16,14 @@ pr:
       - .pipelines/templates/e2e-template.yaml
       - e2e
       - parts/linux
+      - parts/windows
       - pkg/agent
+      - staging/cse/windows/
```
anujmaheshwari1, line 21: I am overall fine with this change, but food for thought: now this pipeline which is combined, will be triggered for linux and windows changes too, but for a linux only change, what if it gets blocked because of a non relevant windows test failure given that its a combined pipeline?

r2k1: I set this for linux pipelines
```
TAGS_TO_SKIP: "os=windows"
```

And this for windows:
```
TAGS_TO_RUN: "os=windows"
```

r2k1: The main goal is to switch from bash to golang

djsly: Il not understanding the ENV. it would be great if we could not have to learn a new inhouse languages to operate the pipeline logic :).  So the skip env are to be changed or they have good value for ever ?

r2k1: Do you have a better idea?

timmy-wright (approved):

PR #5362, opened 2024-12-04
test: e2e fixes and improvements

r2k1:
/kind test

Various E2E fixes. It took me many trial and error attempts before I managed to reliability of GPU tests.

- use ephemeral OS disks. It seems to be faster and produce less IO errors
- add cluster creation timeout for faster feedback on errors
- use long-poll using watcher instead of short-poll. For faster and more reliable feedback on short-lived pods
- Reduce noise in logs
- Use oldest supported kubelet version (instead of unsupported)
- fix GPU tests. In the end I couldn't make them stable.  I can't waste more time on it, so just disabled the bad validator for now.


e2e/node_config.go
```diff
@@ -156,8 +155,8 @@ func baseTemplateLinux(location string) *datamodel.NodeBootstrappingConfiguratio
 						UserAssignedID:                    "",
 						UserAssignedClientID:              "",
 						CustomHyperkubeImage:              "",
-						CustomKubeProxyImage:              "mcr.microsoft.com/oss/kubernetes/kube-proxy:v1.26.0.1",
-						CustomKubeBinaryURL:               "https://acs-mirror.azureedge.net/kubernetes/v1.26.0/binaries/kubernetes-node-linux-amd64.tar.gz",
+						CustomKubeProxyImage:              "mcr.microsoft.com/oss/kubernetes/kube-proxy:v1.27.16", // Oldest supported version, end of life is Jul 2025 https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#kubernetes-versions
```
djsly, line 158: why do we need to use the latest supported version ?

e2e/node_config.go
```diff
@@ -483,8 +482,8 @@ func baseTemplateWindows(location string) *datamodel.NodeBootstrappingConfigurat
 				},
 				AgentPoolProfiles: []*datamodel.AgentPoolProfile{
 					{
-						Name: "winnp",
-						//VMSize:              windowsE2EVmSize,
```
djsly, line 487: remove comment ?

e2e/scenario_test.go
```diff
@@ -190,7 +190,6 @@ func Test_AzureLinuxV2_GPU(t *testing.T) {
 				vmss.SKU.Name = to.Ptr("Standard_NC6s_v3")
 			},
 			Validator: func(ctx context.Context, s *Scenario) {
-				ValidatePodUsingNVidiaGPU(ctx, s)
```
djsly, line 193: why are we removing all GPU tests from all SKUs expect for 1804 ? when 1804 will be EOL in 1 month.

e2e/vmss.go
```diff
@@ -115,45 +116,56 @@ func extractLogsFromVMLinux(ctx context.Context, s *Scenario) {
...
-				logFiles[file+".stdout.txt"] = out
+				logFiles[file] = out
 			}
 
 		}
 		if execResult.stderr != nil {
-			out := execResult.stderr.String()
+			out := removeWarningMessages(execResult.stderr.String())
 			if out != "" {
-				logFiles[file+".stderr.txt"] = out
```
djsly, line 149: why are we now putting the stdout and stderr in teh same log file ?   Should we add a delimited at the beginning and end of each log type ?

```
=================
           STDOUT
=================
```

e2e/vmss.go
```diff
@@ -459,8 +471,12 @@ func getBaseVMSSModel(s *Scenario, customData, cseCmd string) armcompute.Virtual
 				StorageProfile: &armcompute.VirtualMachineScaleSetStorageProfile{
 					OSDisk: &armcompute.VirtualMachineScaleSetOSDisk{
 						CreateOption: to.Ptr(armcompute.DiskCreateOptionTypesFromImage),
-						DiskSizeGB:   to.Ptr(int32(512)),
+						DiskSizeGB:   to.Ptr(int32(50)),
 						OSType:       to.Ptr(armcompute.OperatingSystemTypesLinux),
+						Caching:      to.Ptr(armcompute.CachingTypesReadOnly),
+						DiffDiskSettings: &armcompute.DiffDiskSettings{
+							Option: to.Ptr(armcompute.DiffDiskOptionsLocal),
```
djsly, line 478: what is this option for exactly ? I can't seem to find a clear description for it.

e2e/vmss.go
```diff
@@ -115,45 +116,56 @@ func extractLogsFromVMLinux(ctx context.Context, s *Scenario) {
...
+				s.T.Logf("%q: %s", sourceCmd, out)
 			}
 		}
 	}
 	err = dumpFileMapToDir(s.T, logFiles)
 	require.NoError(s.T, err)
 }
 
+// removeWarningMessages removes noise from logs
+func removeWarningMessages(input string) string {
```
djsly, line 157: looks like more then WArnings are being removed, shoudl the function name be removed ?

e2e/node_config.go
r2k1: I think we should test something we support and actually run in production, rather than something has been deprecated a while ago.

e2e/vmss.go
r2k1: It's not the same file.

Stdout is saved into a file. Stderr (if any) is going to a log stdout. (Maybe combining stdout a stderr saving in a single file is better)

99% of the time the stderr from execution contains:

```
Warning: Permanently added '10.224.0.6' (ED25519) to the list of known hosts.

Authorized uses only. All activity may be monitored and reported.
```

It's just a noise to sift through that I wanted to get rid of.

e2e/vmss.go
r2k1: it's ephemeral disks

e2e/vmss.go
r2k1: What do you mean?

e2e/node_config.go
cameronmeissner: i think it would worthwhile to resolve this version dynamically at runtime by inspecting the e2e cluster's properties and extracting the exacting the control plane's version

e2e/scenario_test.go
r2k1: The validator I added is fragile, so I'm reverting the change in removing it from everywhere.

Just missed a comment for 1804. Removed.

e2e/node_config.go
r2k1: added

e2e/vmss.go
r2k1: changed to write combined output to the same file. (So, I don't need to see noise in test logs, and don't need to download 2 files from CI)

e2e/vmss.go
r2k1: deleted the function

e2e/node_config.go
```diff
@@ -156,8 +154,8 @@ func baseTemplateLinux(location string) *datamodel.NodeBootstrappingConfiguratio
 						UserAssignedID:                    "",
 						UserAssignedClientID:              "",
 						CustomHyperkubeImage:              "",
-						CustomKubeProxyImage:              "mcr.microsoft.com/oss/kubernetes/kube-proxy:v1.26.0.1",
-						CustomKubeBinaryURL:               "https://acs-mirror.azureedge.net/kubernetes/v1.26.0/binaries/kubernetes-node-linux-amd64.tar.gz",
+						CustomKubeProxyImage:              fmt.Sprintf("mcr.microsoft.com/oss/kubernetes/kube-proxy:v%s", k8sVersion),
+						CustomKubeBinaryURL:               fmt.Sprintf("https://acs-mirror.azureedge.net/kubernetes/v%s/binaries/kubernetes-node-linux-amd64.tar.gz", k8sVersion),
```
cameronmeissner, line 158: can we also update OrchestratorVersion as well? (line 139) I think there's a number of other fields we can (and should) resolve dynamically when it comes to component versioning, though maybe we can tackle the rest of later in subsequent PR

e2e/node_config.go
djsly: I meant to write the oldest :). 

I was looking to have Artur come up with Cameron suggestion   Wednesday are all set

djsly (approved):

PR #5367, opened 2024-12-04
feat: ensure user provided motd is displayed on ubuntu nodes when dynamic motd generation is used

UtheMan:
If motd is passed to an ubuntu node, entry in `/etc/update-motd.d` will be created that ensures the contents of `/etc/motd` will be displayed in the automatically generated motd. Context: https://manpages.ubuntu.com/manpages/oracular/en/man8/pam_motd.8.html and https://manpages.ubuntu.com/manpages/oracular/en/man5/update-motd.5.html


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -249,6 +249,12 @@ fi
 if [[ "${MESSAGE_OF_THE_DAY}" != "" ]]; then
     if isMarinerOrAzureLinux "$OS" && [ -f /etc/dnf/automatic.conf ]; then
       sed -i "s/emit_via = motd/emit_via = stdio/g" /etc/dnf/automatic.conf
+    elif [[ $OS == "$UBUNTU_OS_NAME" ]] && [[ -d "/etc/update-motd.d" ]]; then
+          aksCustomMotdUpdatePath=/etc/update-motd.d/99-aks-custom-motd
+          touch "${aksCustomMotdUpdatePath}"
+          chmod 0755 "${aksCustomMotdUpdatePath}"
+          echo '#!/bin/bash' >> "${aksCustomMotdUpdatePath}"
```
cameronmeissner, line 256: can we just use `cat` or `tee` to write the script out in a single call? feel like multiple echo statements might also be more error prone

cameronmeissner: also, can we name it accordingly if it's actually a bash script? e.g. aksCustomMotdUpdateScript

I might be missing something though if it's being named this way, also should we be using `.sh` as the file extension? looks like right now we don't use any extension

e2e/scenario_test.go
```diff
@@ -1131,6 +1131,7 @@ func Test_Ubuntu2204_MessageOfTheDay(t *testing.T) {
 			},
 			Validator: func(ctx context.Context, s *Scenario) {
 				ValidateFileHasContent(ctx, s, "/etc/motd", "foobar")
+				ValidateFileHasContent(ctx, s, "/etc/update-motd.d/99-aks-custom-motd", "/etc/motd")
```
cameronmeissner, line 1134: maybe better to assert the whole string`cat /etc/motd/` is there?

parts/linux/cloud-init/artifacts/cse_main.sh
UtheMan: Naming needs to follow the convention described here: https://manpages.ubuntu.com/manpages/oracular/en/man5/update-motd.5.html

Will pass the whole thing in one call.

cameronmeissner: oh lol I just looked at the PR description, I guess in the docs they don't use `.sh`...

UtheMan: Other names that might be added by Ubuntu can for example be `10-help-text` or `50-landscape-sysinfo`

cameronmeissner (approved):

PR #5384, opened 2024-12-06
feat: only configure custom ca certs if particular node had it passed in security profile

UtheMan:


pkg/agent/datamodel/types.go
```diff
@@ -1773,7 +1773,8 @@ type HTTPProxyConfig struct {
 }
 
 type CustomCATrustConfig struct {
-	CustomCATrustCerts []string `json:"customCATrustCerts,omitempty"`
+	CustomCATrustCerts  []string `json:"customCATrustCerts,omitempty"`
+	EnableCustomCATrust bool     `json:"enableCustomCATrust,omitempty"`
```
cameronmeissner, line 1777: i dont have context so im shooting in the dark, though could we just make this a method that returns `len(c.CustomCATrustCerts) > 0`?

cameronmeissner: whether we can do that depends on whether we'd have cases where this is disabled while also having at least one custom CA

UtheMan: @cameronmeissner how it currently works is that when at least one cert is passed in that array, then they will be installed on all nodes that are in MC. This changes it, so that those certs are only installed on nodes that have the feature explicitly enabled.

PR #5423, opened 2024-12-13
feat: use token for attaching ua

cameronmeissner:
**What type of PR is this?**

allows us to attach UA when building ubuntu SKUs that require ESM via usage of a token, rather than needing to have Canonical individually white-list our build subscriptions

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/init-variables.sh
```diff
@@ -506,6 +506,22 @@ if [ "$MODE" == "windowsVhdMode" ] || [ "${ENVIRONMENT,,}" == "prod" ]; then
 	PACKER_BUILD_LOCATION=$AZURE_LOCATION
 fi
 
+set +x
```
UtheMan, line 509: `set +x` okay to be left enabled there?

UtheMan (approved):

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -85,9 +85,11 @@ if isMarinerOrAzureLinux "$OS"; then
     installFIPS
   fi
 else
-  # Enable ESM for 18.04 and FIPS only
+  # Enable ESM only for 18.04, 20.04, and FIPS
   if [[ "${UBUNTU_RELEASE}" == "18.04" ]] || [[ "${UBUNTU_RELEASE}" == "20.04" ]] || [[ "${ENABLE_FIPS,,}" == "true" ]]; then
+    set +x
     autoAttachUA
```
djsly, line 91: we should renamed, since auto-attach if the name of the feature with the subID in an allow-list,  auto will automatically fetch the Token based on the SubID Lookup.

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -85,9 +85,11 @@ if isMarinerOrAzureLinux "$OS"; then
     installFIPS
   fi
 else
-  # Enable ESM for 18.04 and FIPS only
+  # Enable ESM only for 18.04, 20.04, and FIPS
   if [[ "${UBUNTU_RELEASE}" == "18.04" ]] || [[ "${UBUNTU_RELEASE}" == "20.04" ]] || [[ "${ENABLE_FIPS,,}" == "true" ]]; then
```
djsly, line 89: We discussed and we should enable PRO for all Ubuntu releases,it allows picking up security patches from canonical instead of relying on the community repo patches. 

Canonical said it was fine, also, we have a Ticket to the that as well created by @cameronmeissner  :D

djsly (changes requested):

vhdbuilder/packer/pre-install-dependencies.sh
cameronmeissner: we previously did this but decided to revert since we can't publish pro-enabled images to community image galleries per contract with Canonical, unless you're saying this is what they've now agreed to?

at the moment 1804/2004/fips SKUs are pro-enabled, and NOT published to community image galleries

vhdbuilder/packer/init-variables.sh
cameronmeissner: yep, since below all we really do is create/echo out the packer settings which gets written to stdout anyways

vhdbuilder/packer/pre-install-dependencies.sh
cameronmeissner: fixed

vhdbuilder/packer/pre-install-dependencies.sh
djsly: aint we disabling pro at the end ?

Also, there was this ticket, should we just close it ? or at least explain that we don't need since Pro would be enabled.  ?

I'm a bit confused there, what is the issue in the end, it's the fact that we would be publishing non pro image with packages coming from a pro repo ?

djsly: lets discuss this offline, we don't need it for now

djsly (approved):

PR #5426, opened 2024-12-13
fix: codeql reports that go toolchain want 1.N.P format

djsly:
/kind cleanup

Update go reference to use patch version reported by codeql.

### DIFF
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker/hack/tools
 
-go 1.21
+go 1.21.13
 
 require (
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
--- a/vhdbuilder/lister/go.mod
+++ b/vhdbuilder/lister/go.mod
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker/vhdbuilder/lister
 
-go 1.21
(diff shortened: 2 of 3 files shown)


vhdbuilder/lister/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker/vhdbuilder/lister
 
-go 1.21
+go 1.21.13
```
cameronmeissner, line 3: we can actually bump this up to 1.22 at least, we were previously pinned on 1.21 since the go-dev container only had 1.21 installed

vhdbuilder/prefetch/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker/vhdbuilder/prefetch
 
-go 1.21
+go 1.21.13
```
cameronmeissner, line 3: same with this

hack/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker/hack/tools
 
-go 1.21
+go 1.21.13
```
cameronmeissner, line 3: I don't own this mod, though I'd also assume it would be safe to bump

cameronmeissner (approved):

PR #5428, opened 2024-12-15
Remove go workspaces

r2k1:
- Sync dependency versions using `go work sync`
- Go workspaces seem to be difficult to deal with it. Replace it with `replace` directory we used before that. Make dependencies more predictable and simplify tooling.
- Update dependabot to include transitive dependencies. And explicitly enable all directories ()


aks-node-controller/go.mod
```diff
@@ -20,6 +20,8 @@ require (
...
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	golang.org/x/sys v0.23.0 // indirect
+	golang.org/x/sys v0.25.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-)
\ No newline at end of file
+)
+
+replace github.com/Azure/agentbaker => ../
```
r2k1, line 27: We probably want to get rid of this line somehow in the future.

I don't like RP to consume a library with replace directive. I don't exactly understand what libraries will be downloaded and I don't like it.

cameronmeissner: maybe we could vendor it in, similarly to how k/k vendors client-go code? https://github.com/kubernetes/kubernetes/tree/master/vendor

main downside with that would obviously be a lot of duplicated file content, I'm also not sure how easy it is to update vendors on the fly for dev purposes

cameronmeissner (approved):

PR #5436, opened 2024-12-16
bug: set mode

timmy-wright:
**What type of PR is this?**

/kind chore

sets the MODE variable in the "Getting Shared Access Signature URI" step in the pipeline. This variable is now used as part of the az authentication code, but is not being set in this one step.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -246,6 +246,7 @@ steps:
       -e OS_NAME="Windows" \
       -e SKU_NAME=${SKU_NAME} \
       -e OFFER_NAME="Windows" \
+      -e MODE=$(MODE) \
       -e IMAGE_VERSION=${AKS_WINDOWS_IMAGE_VERSION} \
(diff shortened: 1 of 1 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -246,6 +246,7 @@ steps:
       -e OS_NAME="Windows" \
       -e SKU_NAME=${SKU_NAME} \
       -e OFFER_NAME="Windows" \
+      -e MODE=$(MODE) \
```
AbelHu, line 249: It seems like that https://github.com/Azure/AgentBaker/commit/16c95f214e694a719c989d637338e043a39b5eb1#diff-058b750b8d1a15973e955b1bb546b9c742e9185a4bf4f286c7ff827d071494caR70 blocked windows vhd build. cc @cameronmeissner

AbelHu (approved):
LGTM. Thanks for the PR.

PR #5444, opened 2024-12-17
fix: install missing udev rules for v6 vm sku

andyzhangx:
**What type of PR is this?**
/kind bug

fix: install missing udev rules for v6 vm sku, source udev rules: https://github.com/Azure/azure-vm-utils/blob/main/udev/80-azure-disk.rules

this udev rules are only for detecting "Microsoft NVMe Direct Disk", and waagent team are asking for adding this rule in Ubuntu image, that requires long time, so we set this udev rules in aks image first

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
fix: install missing udev rules for v6 vm sku
```


timmy-wright (approved):
Chatted with Andy - this should be a rule in the Ubuntu image, but that will take ages to happen, so we're adding it here to get the benefit of it earlier.

PR #5445, opened 2024-12-17
chore: Update .NET patch version for Windows

timmy-wright:
**What type of PR is this?**

/kind chore

* Update .NET patch version for WS2019
* Prints out error messages from windows test script if the test script fails.
* Quote variables in VHD test script

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/test/run-test.sh
+++ b/vhdbuilder/packer/test/run-test.sh
@@ -198,7 +198,7 @@ else
   #   Invalid string: control characters from U+0000 through U+001F must be escaped
   errMsg=$(echo -E $ret | jq '.value[]  | select(.code == "ComponentStatus/StdErr/succeeded") | .message')
   # a successful errMsg should be '""' after parsed by `jq`
-  if [ $errMsg != \"\" ]; then
+  if [ "$errMsg" != \"\" ]; then
         echo "Tests failed. Test output is: "
         echo "$ret"
     exit 1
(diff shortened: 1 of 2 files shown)


r2k1 (approved):

vhdbuilder/packer/test/run-test.sh
```diff
@@ -40,15 +40,15 @@ else
 fi
 
 # Create the testing resource group
-az group create --name $TEST_VM_RESOURCE_GROUP_NAME --location ${AZURE_LOCATION} --tags "source=AgentBaker" "now=$(date +%s)" "branch=${GIT_BRANCH}"
+az group create --name "$TEST_VM_RESOURCE_GROUP_NAME" --location "${AZURE_LOCATION}" --tags "source=AgentBaker" "now=$(date +%s)" "branch=${GIT_BRANCH}"
```
babbageclunk, line 43: May as well put braces on the variable use as well? (Same for other cases further down.)

```suggestion
az group create --name "${TEST_VM_RESOURCE_GROUP_NAME}" --location "${AZURE_LOCATION}" --tags "source=AgentBaker" "now=$(date +%s)" "branch=${GIT_BRANCH}"
```

babbageclunk (approved):

vhdbuilder/packer/test/run-test.sh
timmy-wright: Good idea, and I agree. However, I need to get this change into master and don't want to restart the build. Will do this if the build fails and I have to restart it tho :)

PR #5447, opened 2024-12-17
chore: add dedicated pipeline variable for E2E location

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -1268,6 +1268,5 @@ stages:
     condition: ne(variables.SKIP_E2E_TESTS, 'true')
     variables:
       VHD_BUILD_ID: $(Build.BuildId)
-      E2E_LOCATION: $(E2E_LOCATION)
     jobs:
       - template: ./templates/e2e-template.yaml
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -1268,6 +1268,6 @@ stages:
     condition: ne(variables.SKIP_E2E_TESTS, 'true')
     variables:
       VHD_BUILD_ID: $(Build.BuildId)
-      LOCATION: $(E2E_LOCATION)
+      E2E_LOCATION: $(E2E_LOCATION)
     jobs:
       - template: ./templates/e2e-template.yaml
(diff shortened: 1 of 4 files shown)


.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1268,6 +1268,6 @@ stages:
     condition: ne(variables.SKIP_E2E_TESTS, 'true')
     variables:
       VHD_BUILD_ID: $(Build.BuildId)
-      LOCATION: $(PACKER_BUILD_LOCATION)
+      LOCATION: $(E2E_LOCATION)
```
r2k1, line 1271: I don't really like we rename and override the variable multiple times before it reaches the test.
It's quite tricky to track.

Why can't we use `LOCATION` somewhere else?
Can we rename LOCATION => E2E_LOCATION inside tests?

.pipelines/templates/e2e-template.yaml
```diff
@@ -25,8 +25,9 @@ jobs:
           export PATH="/usr/local/go/bin:$PATH"
           go version
           
-          echo "VHD_BUILD_ID=$VHD_BUILD_ID"
+          export LOCATION=$LOCATION
```
r2k1, line 28: does this line do anything?

.pipelines/.vsts-vhd-builder-release.yaml
r2k1: Do we actually need to set it?

We have a default value in code. And it can be overridden with a pipeline variable if necessary

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1268,6 +1268,6 @@ stages:
     condition: ne(variables.SKIP_E2E_TESTS, 'true')
     variables:
       VHD_BUILD_ID: $(Build.BuildId)
-      LOCATION: $(PACKER_BUILD_LOCATION)
+      E2E_LOCATION: $(E2E_LOCATION)
```
r2k1, line 1271: you probably can drop this line now?

cameronmeissner: fair point

r2k1 (approved):

.pipelines/templates/e2e-template.yaml
cameronmeissner: not needed

PR #5099, opened 2024-10-16
fix: Ensure nvidia modprobe is run before kubelet

benjamin-brady:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Removes the  `kubelet restart` from nvidia-modprobe service and ensures nvidia-modprobe is run before kubelet.
**Which issue(s) this PR fixes**:
Incident 487929399, 552859186

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  <!-- Common commit types:
        build: Build 🏭
        chore: Maintenance 🔧
        ci: Continuous Integration 💜
        docs: Documentation 📘
        feat: Features 🌈
        fix: Bug Fixes 🐞
        perf: Performance Improvements 🚀
        refactor: Code Refactoring 💎
        revert: Revert Change ◀️
        style: Code Style 🎶
        security: Security Fix 🛡️
        test: Testing 💚 
- [ ] includes documentation
- [x] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #5450, opened 2024-12-18
fix: Kubelet restarts initiated by nvidia-modprobe service

benjamin-brady:
**What type of PR is this?**

/kind bug

PR #5099 attempted to fix an issue with kubelet restarting after nvidia-modprobe.service starts. The original bugfix ensures kubelet depends on the nvidia-modprobe service so that it doesn't need to be restarted after installation. @tomasgiden reported the issue is still occurring on the latest image versions. It turns out there was still a `restart kubelet` command in the `nodecustomdata.yml` version of the nvidia-modprobe.service which I missed so I've amended it here.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
Fix: kubelet restarting when nvidia-modprobe service is installed.
```


timmy-wright (approved):

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -292,14 +292,14 @@ write_files:
...
+    Before=kubelet.service
     [Service]
     Type=oneshot
     RemainAfterExit=true
     ExecStartPre=/bin/sh -c "dkms autoinstall --verbose"
     ExecStart=/bin/sh -c "nvidia-modprobe -u -c0"
-    ExecStartPost=/bin/sh -c "sleep 10 && systemctl restart kubelet"
     [Install]
-    WantedBy=multi-user.target
+    WantedBy=multi-user.target kubelet.service
```
r2k1, line 302: we want to move away from saving things through custom data, I think you will need to update it somewhere else

r2k1: parts/linux/cloud-init/artifacts/nvidia-modprobe.service

^ I think this should be source of truth. And this file will override the content on VM during provision time.

r2k1: Ah nevermind, it's already done before. I should read the PR description first.

benjamin-brady: I tend to agree it's chaotic having two sources for this service. I'd like to remove the custom data but following the Chesterton's fence principle I don't like to remove things without first knowing why they are there in the first place. Do you think it's safe to remove the nvidia-modprobe service from custom data? I'm not sure if there is a reason it's duplicated in the first place.

e2e/validators.go
```diff
@@ -259,6 +259,13 @@ func ValidateKubeletHasNotStopped(ctx context.Context, s *Scenario) {
 	assert.Contains(s.T, execResult.stdout.String(), "Started Kubelet")
 }
 
+func ValidateServicesDoNotRestartKubelet(ctx context.Context, s *Scenario) {
```
r2k1, line 262: I would name it `ValidateKubeletHasNotBeenRestarted` but it's not critical

r2k1 (approved):

PR #5453, opened 2024-12-18
chore: windows VHD build pipeline tweaks

timmy-wright:
**What type of PR is this?**

/kind chore

* Split the Win VHD build and e2e tests into seperate stages
* Additional checks if a container pull fails due to out of disk space

image

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -162,12 +175,53 @@ stages:
...
               artifactName: 23H2-gen2
 
-  - stage: e2e
-    condition: ne(variables.SKIP_E2E_TESTS, 'true')
+  - stage: e2e_win_2019_gen1
+    dependsOn: [ build_win_2019_gen1 ]
+    condition: and(succeeded(), ne(variables.SKIP_E2E_TESTS, 'true'))
+    variables:
+      VHD_BUILD_ID: $(Build.BuildId)
+      LOCATION: $(PACKER_BUILD_LOCATION)
```
r2k1, line 183: I think you can move 2 variables to upper level to avoid duplication

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -162,12 +175,53 @@ stages:
...
+      TAGS_TO_RUN: "os=windows,imageName=windows-2019-containerd"
```
r2k1, line 184: don't need to specify os=windows if you specify image

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -162,12 +175,53 @@ stages:
...
+    condition: and(succeeded(), ne(variables.SKIP_E2E_TESTS, 'true'))
+    variables:
+      VHD_BUILD_ID: $(Build.BuildId)
+      LOCATION: $(PACKER_BUILD_LOCATION)
+      TAGS_TO_RUN: "os=windows,imageName=windows-2022-containerd-gen2"
+    jobs:
+      - template: ./templates/e2e-template.yaml
+
+  - stage: e2e_win_23H2_gen1
+    dependsOn: [ build_win_23H2_gen1 ]
```
r2k1, line 209: Maybe it will look cleaner if it will be part of the build stage?
This way you won't need to specify dependsOn

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -162,12 +175,53 @@ stages:
...
+    variables:
+      VHD_BUILD_ID: $(Build.BuildId)
+      LOCATION: $(PACKER_BUILD_LOCATION)
+      TAGS_TO_RUN: "os=windows,imageName=windows-2019-containerd"
+    jobs:
+      - template: ./templates/e2e-template.yaml
+
+  - stage: e2e_win_2022_gen1
+    dependsOn: [ build_win_2022_gen1 ]
+    condition: and(succeeded(), ne(variables.SKIP_E2E_TESTS, 'true'))
```
r2k1, line 190: maybe the skip condition can be moved to the template to avoid duplication?

r2k1 (commented):
I think we want to share logic with Linux builds.

But I worry about amount of YAML boilerplate we need to maintain. (Something extra to worry about when adding or dropping support for a VHD)

r2k1:
maybe e2e can be part of `.builder-release-template-windows.yaml` template to avoid duplication?

.pipelines/.vsts-vhd-builder-release-windows.yaml
timmy-wright: I tried that and it didn't work - probably due to the way the template is written.

.pipelines/.vsts-vhd-builder-release-windows.yaml
timmy-wright: This is true, but I still like having it there for readability :)

.pipelines/.vsts-vhd-builder-release-windows.yaml
timmy-wright: Am trying this ... hopefully it works!

.pipelines/.vsts-vhd-builder-release-windows.yaml
timmy-wright: Have removed os tag.

.pipelines/.vsts-vhd-builder-release-windows.yaml
timmy-wright: Have moved to the template.

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -55,121 +55,69 @@ parameters:
   type: boolean
   default: False
 
+variables:
+  VHD_BUILD_ID: $(Build.BuildId)
+  LOCATION: $(PACKER_BUILD_LOCATION)
```
r2k1, line 60: I think now it can be moved closer to the place where it used, in the template

r2k1 (approved):

PR #5456, opened 2024-12-19
chore: make linux e2e tests run for the VHD they are built for.

timmy-wright:
**What type of PR is this?**

/chore

This PR changes the Linux VHD build pipeline so that the e2e tests run for each VHD, and only the e2e tests for that VHD are run. It removes duplication between the PR VHDs we build (and those e2e tests) and makes the template a bit easier to read.

Note that at this time, the e2e test filtering is not working. I'll work on that when I have some headspace.

Screen snapshot of new pipeline structure. Note that this is taken from the PR checkin pipeline, so we are disabling most VHD builds and tests - only building and testing 3 VHDs. This is easily configurable.

image

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


r2k1:
One thing is concerning me. 

We're running a bunch of steps to prepare a test environment. Create an RG, create a cluster etc. Now we're going to attempt to perform it concurrently and 20x times more often. (There is also can be concurrent pipeline execution)

It may produce some subtle and flaky concurrency-related bugs.

It also increases amount of ARM-API calls and makes it more likely to get the subscription throttled.

PR #5482, opened 2024-12-26
feat: Add support for setting a different CNI config on cilium clusters

santhoshmprabhu:
**What type of PR is this?**
/kind feature

We are currently adding support for Windows nodepools on  nodesubnet clusters that use Cilium as the CNI plugin. In this setting, windows nodes are expected to use stateless Azure CNI, while Linux nodes use Cilium. In non-Cilium Nodesubnet clusters, the Windows nodes use stateful Azure CNI. Stateless Azure CNI requires a different conflist than stateful Azure CNI. In this PR, we pass information about whether the cluster is using Cilium or not, so that the conflist can be updated to suit stateful or stateless CNI.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


tamilmani1989 (approved):

staging/cse/windows/azurecnifunc.tests.ps1
```diff
@@ -55,6 +56,25 @@ Describe 'Set-AzureCNIConfig' {
...
+            Set-AzureCNIConfig -AzureCNIConfDir $azureCNIConfDir `
+                -KubeDnsSearchPath $kubeDnsSearchPath `
+                -KubeClusterCIDR $kubeClusterCIDR `
+                -KubeServiceCIDR $kubeServiceCIDR `
+                -VNetCIDR $vNetCIDR `
+                -IsDualStackEnabled $isDualStackEnabled
+
+            $actualConfigJson = Read-Format-Json $azureCNIConfigFile
+            $expectedConfigJson = Read-Format-Json ([Io.path]::Combine($azureCNIConfDir, "AzureCNI.Expect.CiliumNodeSubnet.conflist"))
+            $diffence = Compare-Object $actualConfigJson $expectedConfigJson
```
timmy-wright, line 73: nit: difference needs an "r" :)

staging/cse/windows/azurecnifunc.tests.ps1
```diff
@@ -55,6 +56,25 @@ Describe 'Set-AzureCNIConfig' {
         }
     }
 
+    Context 'Cilium (ebpf dataplane) is enabled' {
```
timmy-wright, line 59: It would be good to see a negative test as well - when ebpf is not enabled.

timmy-wright (approved):

staging/cse/windows/azurecnifunc.tests.ps1
santhoshmprabhu: That's the default, so the test would be exactly the same as the next one in the file (`WinDSR is enabled by default`). Skip it?

santhoshmprabhu: Updated the test case description to highlight that ebpf dataplane is disabled in the default case. Checked with Tim, resolving.

pkg/agent/datamodel/types.go
```diff
@@ -676,6 +676,7 @@ type KubernetesConfig struct {
 	MaximumLoadBalancerRuleCount      int               `json:"maximumLoadBalancerRuleCount,omitempty"`
 	PrivateAzureRegistryServer        string            `json:"privateAzureRegistryServer,omitempty"`
 	NetworkPluginMode                 string            `json:"networkPluginMode,omitempty"`
+	EbpfDataplane                     int               `json:"ebpfDataplane,omitempty"`
```
cameronmeissner, line 679: should this be a bool instead?

maybe an enum of you can have other values than `1`?

cameronmeissner (changes requested):
should this be a bool instead?

maybe an enum of you can have other values than 1?

pkg/agent/datamodel/types.go
cameronmeissner, line 679: should this be a bool instead?

maybe an enum of you can have other values than 1?

santhoshmprabhu: The struct is meant to match the AKS struct here:
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/protos/hcp/types/common/v1/kubernetesconfig.pb.go&version=GBmaster&line=216&lineEnd=217&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

Updated to a matching enum.

cameronmeissner: thanks, please also refrain from sharing ADO links within this repo in the future

pkg/agent/datamodel/types.go
```diff
@@ -622,6 +622,22 @@ type KubernetesAddon struct {
 	Data       string                    `json:"data,omitempty"`
 }
 
+// EbpfDataplane controls the eBPF networking dataplane.
+type EbpfDataplane int32
+
+const (
+	// none means don't install an eBPF dataplane.
+	EbpfDataplane_none EbpfDataplane = 0
```
cameronmeissner, line 630: nit: I'd use golang's iota feature here instead: https://go.dev/wiki/Iota

cameronmeissner: eh, I guess what you have is fine if you're trying to mirror RP

pkg/agent/baker.go
```diff
@@ -623,6 +623,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"IsAzureCNIOverlayFeatureEnabled": func() bool {
 			return cs.Properties.OrchestratorProfile.KubernetesConfig.IsUsingNetworkPluginMode("overlay")
 		},
+		"EbpfDataplane": func() bool {
```
cameronmeissner, line 626: seems like we should change the func name to `EbpfDataplaneCilium` here to reflect what we're actually checking for?

santhoshmprabhu: The only valid value for `EbpfDataplane` is Cilium, so renamed the function to `CiliumDataplaneEnabled`.

pkg/agent/datamodel/types.go
santhoshmprabhu: Leaving this untouched.

cameronmeissner (approved):

PR #5508, opened 2025-01-06
[NI] support non-anonymous access for oras pull

fseldow:
**What type of PR is this?**

/kind feature

[NI] support non-anonymous access for oras pull

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -245,10 +245,11 @@ oras_login_with_identity() {
     local acr_url=$1
     local client_id=$2
     local tenant_id=$3
-
+    set +x
     ACCESS_TOKEN=$(curl "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com%2F&client_id=$client_id" -H Metadata:true -s | jq .access_token -r)
     if [ -z "$ACCESS_TOKEN" ]; then
         echo "Failed to retrieve kubelet token"
+        set -x 
         return 1
     fi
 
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -215,6 +215,54 @@ retrycmd_get_tarball() {
...
+    if [ -z "$ACCESS_TOKEN" ]; then
+        echo "Failed to retrieve kubelet token"
+        return 1
+    fi
+
+    ACR_TOKEN=$(curl -X POST "https://$acr_url/oauth2/exchange" \
+        -H "Content-Type: application/x-www-form-urlencoded" \
+        -d "grant_type=access_token&service=$acr_url&tenant=$tenant_id&access_token=$ACCESS_TOKEN" -s | jq -r '.refresh_token')
+
+    if [ -z "$ACR_TOKEN" ]; then
```
cameronmeissner, line 259: might considering wrapping lines referencing ACR_TOKEN between a:
```
set +x
...
set -x
```
so we don't log it within cluster provisioning logs, I'm assuming the refresh token is fairly short-lived though still better to keep logs clean of any secrets

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -215,6 +215,54 @@ retrycmd_get_tarball() {
...
+    for i in $(seq 1 $access_retries); do
+        if [ $i -eq $access_retries ]; then
+            return $ERR_ORAS_PULL_NETWORK_ACCESS
+        else
+            timeout 60 oras pull $sample_image --registry-config ${ORAS_REGISTRY_CONFIG_FILE} > $ORAS_OUTPUT 2>&1
+            if [[ $? != 0 ]]; then
+                echo "acr access check failed:"
+                error_output=$(cat $ORAS_OUTPUT)
+                if [[ $error_output == *"not found"* ]]; then
+                    echo "Error: Image '$sample_image' not found, please check acr cache is settled correctly"
```
cameronmeissner, line 232:
```suggestion
                    echo "Error: Image '$sample_image' not found, please check acr cache is correctly set"
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -215,6 +215,54 @@ retrycmd_get_tarball() {
...
+retrycmd_acr_access_check() {
+    access_retries=$1; wait_sleep=$2;
+    local acr_url=$3
+    echo "${access_retries} retries for acr access check"
+    sample_image="$acr_url/mcr/hello-world:latest"
+    for i in $(seq 1 $access_retries); do
+        if [ $i -eq $access_retries ]; then
+            return $ERR_ORAS_PULL_NETWORK_ACCESS
+        else
+            timeout 60 oras pull $sample_image --registry-config ${ORAS_REGISTRY_CONFIG_FILE} > $ORAS_OUTPUT 2>&1
```
cameronmeissner, line 227: I'd recommend we just store the output of the pull command in a var rather than a new file, as it stands it also looks like that file would be persisted which we don't want

parts/linux/cloud-init/artifacts/cse_helpers.sh
fseldow: add `set +x` and `set -x` for now

PR #5511, opened 2025-01-06
refactor: manually generate self-signed kubelet serving cert when opting out of certificate rotation

cameronmeissner:
**What type of PR is this?**

in cases where customers want to manually opt-out of kubelet serving cert rotation via nodepool tag, this PR ensures that the self-signed certificate their kubelets end up getting look _exactly_ the same as the self-signed certs they'd get if they hadn't been onboarded to kubelet serving certificate rotation in the first place - this is accomplished by calling the `generateSelfSignedKubeletServingCertificate` function explicitly and passing in `--tls-cert-file`/`--tls-private-key-file` to the kubelet, rather than letting kubelet generate its own self-signed cert at startup time.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
relevant abe2e scenarios are passing, failures are unrelated

lilypan26 (approved):

PR #5524, opened 2025-01-08
feat: create vmss tag to skip binary cleanup during CSE

AlisonB319:
**What type of PR is this?**
We want to have the option of not cleaning up unused binaries on our VHD in order to understand which vulnerabilities on 1P AKS nodes came from the image vs. which were introduced by the customer (adding extensions, packages, etc.)

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -104,7 +104,12 @@ logs_to_events "AKS.CSE.disableSystemdResolved" disableSystemdResolved
 logs_to_events "AKS.CSE.configureAdminUser" configureAdminUser
 
 VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
-if [ -f $VHD_LOGS_FILEPATH ]; then
+export -f should_skip_binary_cleanup
+SKIP_BINARY_CLEANUP=$(retrycmd_if_failure_no_stats 10 1 10 bash -cx should_skip_binary_cleanup)
+if [[ "${SKIP_BINARY_CLEANUP}" == true ]]; then
```
Devinwong, line 109: Is it expected `if [[ "${SKIP_BINARY_CLEANUP}" == true ]]` then it doesn't check `[ -f $VHD_LOGS_FILEPATH ]` ? It seems irrelevant per the variables naming.

AlisonB319: Thats intended because for `[-f $VHD_LOGS_FILEPATH ]` it will call 

`logs_to_events "AKS.CSE.cleanUpContainerImages" cleanUpContainerImages`

which is what I don't want to happen if we are skipping cleanup

Devinwong: Sounds good. Thanks for confirming

Devinwong (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -104,7 +104,12 @@ logs_to_events "AKS.CSE.disableSystemdResolved" disableSystemdResolved
...
+    echo "binaries will not be cleaned up"
+    FULL_INSTALL_REQUIRED=true
```
djsly, line 111: what is this for ?

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: true is a bool in this case ?  the curl results returns a string, so what is the final logic ? any string (value) set for the TAG will be `true` ?

djsly: should `"true"` be used instead ?

pkg/agent/testdata/AKSUbuntu1604+Containerd/line9.sh
```diff
@@ -460,6 +460,17 @@ should_disable_kubelet_serving_certificate_rotation() {
...
 }
 
+should_skip_binary_cleanup() {
+    set -x
+    body=$(curl -fsSL -H "Metadata: true" --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01")
+    ret=$?
+    if [ "$ret" != "0" ]; then
+      return $ret
+    fi
+    should_skip=$(echo "$body" | jq -r '.compute.tagsList[] | select(.name == "SkipBinaryCleanup") | .value')
```
jason1028kr, line 470: How/where does the tag gets applied to the node?

jason1028kr (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
AlisonB319:
```
if [[ $OS == $UBUNTU_OS_NAME ]] && [ "$FULL_INSTALL_REQUIRED" = "true" ]; then
    logs_to_events "AKS.CSE.installDeps" installDeps
else
```
it triggers the installDeps function

pkg/agent/testdata/AKSUbuntu1604+Containerd/line9.sh
AlisonB319: You have to apply it as a vmss tag when you create the cluster
```
az aks create -g $CLUSTER_RG \
  --aks-custom-headers AKSHTTPCustomFeatures=Microsoft.ContainerService/UseCustomizedOSImage,OSImageSubscriptionID=$GALLERY_SUBID,OSImageResourceGroup=$GALLERY_RG,OSImageGallery=$GALLERY_NAME,OSImageName=$OS_IMAGE_NAME,OSImageVersion=$OS_IMAGE_VERSION \
  -n $CLUSTER_NAME \
  --vm-set-type VirtualMachineScaleSets \
  --node-count 1 \
  --node-vm-size $VM_SIZE \
  --tags SkipBinaryCleanup="true" 
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -486,6 +486,17 @@ should_disable_kubelet_serving_certificate_rotation() {
...
 
+should_skip_binary_cleanup() {
+    set -x
+    body=$(curl -fsSL -H "Metadata: true" --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01")
+    ret=$?
+    if [ "$ret" != "0" ]; then
+      return $ret
+    fi
+    should_skip=$(echo "$body" | jq -r '.compute.tagsList[] | select(.name == "SkipBinaryCleanup") | .value')
+    echo "$should_skip"
```
cameronmeissner, line 497: I'd recommend sending this to lower-case with `echo "${should_skip.,,}"`

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner, line 109:
```suggestion
if [ "${SKIP_BINARY_CLEANUP}" == "true" ]; then
```

cameronmeissner, line 109: im not sure this is exactly what we want

cameronmeissner: confirmed offline setting FULL_INSTALL_REQUIRED to true in our VHD is _probably_ fine, and we'll only be doing this internally anyways

cameronmeissner: I might try something like this to make things a bit more readable:


```
export -f should_skip_binary_cleanup
SKIP_BINARY_CLEANUP=$(retrycmd_if_failure_no_stats 10 1 10 bash -cx should_skip_binary_cleanup)

if [ ! -f $VHD_LOGS_FILEPATH ] && [ "${IS_VHD,,}" == "true" ]; then
    echo "Using VHD distro but file $VHD_LOGS_FILEPATH not found"
    exit $ERR_VHD_FILE_NOT_FOUND
fi

FULL_INSTALL_REQUIRED=false
    echo "detected golden image pre-install"
    logs_to_events "AKS.CSE.cleanUpContainerImages" cleanUpContainerImages
elif [ "${SKIP_BINARY_CLEANUP}" == "true" ]; then
    echo "binaries will not be cleaned up"
    FULL_INSTALL_REQUIRED=true
else
    if [[ "${IS_VHD}" = true ]]; then
        echo "Using VHD distro but file $VHD_LOGS_FILEPATH not found"
        exit $ERR_VHD_FILE_NOT_FOUND
    fi
    FULL_INSTALL_REQUIRED=true
fi
```

cameronmeissner: Ideally we'd be able to extract this logic out to a helper function which we could unit test with shellspec, having it directly within cse_main.sh makes things much harder to test unfortunately

e2e/validators.go
```diff
@@ -218,6 +219,30 @@ func ValidateIMDSRestrictionRule(ctx context.Context, s *Scenario, table string)
...
 
+func ValidateMultipleKubeProxyVersionsExist(ctx context.Context, s *Scenario) {
+	execResult := execOnVMForScenario(ctx, s, "ctr --namespace k8s.io images list | grep kube-proxy | awk '{print $1}' | grep -oE '[0-9]+\\.[0-9]+\\.[0-9]+'")
+	if execResult.exitCode != "0" {
+		s.T.Errorf("Failed to list kube-proxy images: %s", execResult.stderr)
+		return
+	}
+
+	versions := bytes.NewBufferString(strings.TrimSpace(execResult.stdout.String()))
+	versionMap := make(map[string]int)
```
cameronmeissner, line 230: supernit: we can accomplish the same thing with a `map[string]struct{}`, since we only care about unique versions (`struct{}` also takes up much less memory)

parts/linux/cloud-init/artifacts/cse_main.sh
AlisonB319: I'll work on extracting, but in the if elif, I think it makes more sense to check for the skip_binary_cleanup first, since `[ -f $VHD_LOGS_FILEPATH ];` just checks if that filepath exists, which it does for the unit tests.

AlisonB319: the jq in 
```
should_skip=$(echo "$body" | jq -r '.compute.tagsList[] | select(.name == "SkipBinaryCleanup") | .value')
    echo "${should_skip,,}"
```
is what causes it to be a bool over a string

AlisonB319: the jq makes it a bool so unless we want to force it to be a string, I think keeping it bool makes sense 
```
should_skip=$(echo "$body" | jq -r '.compute.tagsList[] | select(.name == "SkipBinaryCleanup") | .value')
echo "${should_skip,,}"
```

AlisonB319: if SkipBinaryCleanup=False in the tags, then it will not skip the cleanup, which is as intended, but I doubt anybody would do this as its default behavior without the tag

e2e/scenario_test.go
```diff
@@ -921,6 +921,30 @@ func Test_Ubuntu2204_WASM(t *testing.T) {
...
 }
 
+func Test_AzureLinux_Skip_Binary_Cleanup(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "tests that an AzureLinux node will skip binary cleanup and can be properly bootstrapped",
+		Config: Config{
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDAzureLinuxV2Gen2,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+				if nbc.KubeletConfig == nil {
```
cameronmeissner, line 931: unless we want to actually add something to the kubelet config here, I don't think we need this mutator

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -771,6 +772,32 @@ cleanUpContainerd() {
     rm -Rf $CONTAINERD_DOWNLOADS_DIR
 }
 
+handleContainerImageCleanup() {
```
cameronmeissner, line 775: nit: `getInstallModeAndCleanupContainerImages`?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -103,18 +103,9 @@ logs_to_events "AKS.CSE.disableSystemdResolved" disableSystemdResolved
...
-else
-    if [[ "${IS_VHD}" = true ]]; then
-        echo "Using VHD distro but file $VHD_LOGS_FILEPATH not found"
-        exit $ERR_VHD_FILE_NOT_FOUND
-    fi
-    FULL_INSTALL_REQUIRED=true
-fi
+export -f should_skip_binary_cleanup
+SKIP_BINARY_CLEANUP=$(retrycmd_if_failure_no_stats 10 1 10 bash -cx should_skip_binary_cleanup)
+FULL_INSTALL_REQUIRED=$(handleContainerImageCleanup $SKIP_BINARY_CLEANUP $IS_VHD)
```
cameronmeissner, line 108: I believe `set -e` will be configured at this point, though I think we should explicitly check `$?` for the `exit $ERR_VHD_FILE_NOT_FOUND` case since we're calling the func in a subshell, otherwise if `set -e` isn't set the exit statement will essentially do nothing

we actually do the same thing up above on lines 90-96 when checking `exit should_skip_nvidia_drivers `

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner, line 775: thanks for giving this a shot :)

spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh
```diff
@@ -64,5 +64,12 @@ Describe 'cse_install.sh'
             The output line 2 should equal "Package \"containerd\" does not exist in $COMPONENTS_FILEPATH."
             The output line 3 should equal "mock installContainerdWithManifestJson calling"
         End
+        It 'should skip binary cleanup if SKIP_BINARY_CLEANUP is true'
```
cameronmeissner, line 67: can we put this in a new `Describe` block specifically for testing the new function?

e2e/scenario_test.go
```diff
@@ -921,6 +921,26 @@ func Test_Ubuntu2204_WASM(t *testing.T) {
 	})
 }
 
+func Test_AzureLinux_Skip_Binary_Cleanup(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "tests that an AzureLinux node will skip binary cleanup and can be properly bootstrapped",
+		Config: Config{
+			Cluster:                ClusterKubenet,
+			VHD:                    config.VHDAzureLinuxV2Gen2,
+			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {},
```
cameronmeissner, line 930: I think we can just omit this entirely

non_existent_file.txt
cameronmeissner, line 1: looks like the shellspec cleanup isn't quite working as expected

e2e/scenario_test.go
AlisonB319: Without the BootstrapConfigMutator it throws an error 
```
kube.go:94: waiting for pod app=debug-mariner  in "default" namespace to be ready
    kube.go:142: pod debug-mariner-4zlf9 is ready
    scenario_helpers_test.go:106: exactly one of BootstrapConfigMutator or AKSNodeConfigMutator must be set
--- FAIL: Test_AzureLinux_Skip_Binary_Cleanup (9.55s)
```

djsly (approved):

e2e/scenario_test.go
cameronmeissner: oh, sad, that didn't used to be a tihng

cameronmeissner (approved):

PR #5601, opened 2025-01-17
fix pipeline triggers missing

timmy-wright:
**What type of PR is this?**

/kind chore

Some pipeline triggers were missing for files used in the windows e2e pipelines. Fixed this.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -8,7 +8,9 @@ pr:
   paths:
     include:
     - .pipelines/.vsts-vhd-builder-release-windows.yaml
+    - .pipelines/templates/.build-and-test-windows-vhd-template.yaml
     - .pipelines/templates/.builder-release-template-windows.yaml
+    - .pipelines/templates/e2e-template.yaml
     - packer.mk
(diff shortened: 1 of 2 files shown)


.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -8,7 +8,9 @@ pr:
   paths:
     include:
     - .pipelines/.vsts-vhd-builder-release-windows.yaml
+    - .pipelines/templates/.build-and-test-windows-vhd-template.yaml
```
AbelHu, line 11: 1. Do the tests pass now? It may block all PRs if the answer is no
2. Should the PR be created against dev branch by following the current process?

timmy-wright: 1. This PR will not make the tests pass. Which I'm comfortable with. At least bypassing failing tests will be a conscious decision rather than an accident.
2. dev branch is now gone! So we can make PRs against master again.

AbelHu (approved):
LGTM. Thanks for the PR. It is better to fix the issue before merging to avoid blocking all incoming PRs.

djsly (approved):

PR #5618, opened 2025-01-21
Enable linter for aks-node-controller

r2k1:
**What type of PR is this?**
/kind test

- Enable linter for aks-node-controller. (Currently linter is running only for core AgentBaker code).
- Upgrade golangci-lint version
- Delete deprecated linters

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.github/workflows/golangci-lint.yml
```diff
@@ -18,16 +18,20 @@ jobs:
   golangci:
     name: lint
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        dirs: [".", "aks-node-controller"]
```
djsly, line 23: which Folder were subject to the linter before ?  should we add more folders ?

djsly (approved):

cameronmeissner (approved):

.github/workflows/golangci-lint.yml
r2k1: yep

r2k1: Another option is to find all `go.mod` files and run linters on them.

But there is a few ad-hoc scripts and e2e I'm not sure I want to bother with lint errors.
Maybe explicitly adding important ones is cleaner. Easier to maintain the pipieline.

I don't think we have many other folders we worry about.

PR #5619, opened 2025-01-21
Update windows scripts to use components.json instead of having versions hard coded in a script

timmy-wright:
**What type of PR is this?**

/kind refactor

This is the first in a series of PRs that will migrate the Windows VHD builds to use components.json. This will eventually let us use the same embargo build process for windows as per linux.

This PR just migrates the cached containers and binaries It also refactors the build scripts.

There is one customer visible change. The components.json file and another helpers file is left in c:/k/ I don't see this as a problem because:
* These files are public on our GitHub (in this repo)
* components.json is already left on the Linux VHDs

The reason we needed to leave those there is so we can run VHD tests after snapshotting the VHD. The test scripts used those files and I couldn't find a nice way to copy the files to the VM running the tests.

Note that the c:\k directory is where we put all the AKS componentry. It's similar to how we use /opt/azure/ in the linux world.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong:
This feature is good. So that we can use 1 components.json to define all components' versions. I see that you apply `renovateTag` on the windows related components. Were you able to test if Renovate auto detection works for those components too?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -9,6 +9,18 @@
           "latestVersion": "1.8.22",
           "previousLatestVersion": "1.8.20"
         }
+      ],
+      "windowsVersions": []
```
Devinwong, line 13: Does this need to be here if it's empty? If it's the components.cue will complain, you can consider configure `windowsVersions` as optional too.

timmy-wright: I don't think it has to be there - I prob put it there out of habit. Will remove.

timmy-wright: Done.

timmy-wright:
> This feature is good. So that we can use 1 components.json to define all components' versions. I see that you apply `renovateTag` on the windows related components. Were you able to test if Renovate auto detection works for those components too?

I haven't tested renovate auto detection - have set all windows versions to disable renovate until I do that.

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -870,6 +1286,11 @@
       "name": "nvidia-container-toolkit",
       "downloadLocation": "/usr/local/bin",
       "downloadURIs": {
+        "windows": {
```
Devinwong, line 1289: Can you also make sure the script will not crash if the "windows" block isn't presented? Please ignore if it's already handled.
Not sure about Windows node, but in Linux case, there could be cases like new shell scripts are executing in old VHD image (security patch scenarior) which doesn't always have the latest components.json thus some keys could be missing. We encountered some IcMs because of this.

Devinwong:
With this PR, does it mean we can expect when there is a newer version of a component, Renovate/user just simply updates this components.json file only and the latest will be cached in Windows VHD correctly? If that's the case, it will be super awesome because component owners can just approve/merge the automated PR and everything else about components update in Windows/Linux VHDs will be well taken care of!

parts/linux/cloud-init/artifacts/components.json
timmy-wright: It's already handled, but I have added some powershell unit tests for this case. 

However, Windows doesn't parse components.json at provision time - it's only at build time - so the risk of IcM is less for Windows than linux anyway.

PR #5635, opened 2025-01-22
test: initial powershell test for windows nodes

timmy-wright:
**What type of PR is this?**

/kind chore

This PR adds some e2e tests for Windows VHDs and some infrastructure code to execute arbitrary commands on a windows VHD as part of the e2e tests. It's got some - currently unused - code to check kubelet command line params as well, that's still in progress, but I want it checked in so I don't loose it.

Note that to do this, I had to change the execOnVm method to not add "sudo" to every command executed on the VM. That broke Windows tests because sudo is not a powershell command. Now the sudo is part of each test - which is probably cleaner. It seems that some existing linux tests prob didn't need to run sudo, but I didn't want to check each one just in case it was necessary.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/scenario_win_test.go
```diff
@@ -28,7 +36,10 @@ func Test_Windows2022Containerd(t *testing.T) {
 			VHD:             config.VHDWindows2022Containerd,
 			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {},
 			BootstrapConfigMutator: func(configuration *datamodel.NodeBootstrappingConfiguration) {
-
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				ValidateFileHasContentWindows(ctx, s, "/k/kubeletstart.ps1", "--container-runtime=remote")
```
cameronmeissner, line 41: it would be cool if each validator provided a windows and linux implementation which could be called here, something like: `ValidateFileHasContent.Powershell(...)` and `ValidateFileHasContent.Bash(...)` or similar

e2e/scenario_win_test.go
```diff
@@ -28,7 +36,10 @@ func Test_Windows2022Containerd(t *testing.T) {
...
+				//ValidateProcessHasCliWindows(ctx, s, "kubelet.exe", []string{"--address=0.0.0.0", "--anonymous-auth=false"})
```
cameronmeissner, line 42: can we remove the commented-out bits?

e2e/exec.go
```diff
@@ -100,10 +100,22 @@ func sshString(vmPrivateIP string) string {
 	return fmt.Sprintf(`ssh -i %s -o PasswordAuthentication=no -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ConnectTimeout=5 azureuser@%s`, sshKeyName(vmPrivateIP), vmPrivateIP)
 }
 
+func quoteForBash(command string) string {
+	return fmt.Sprintf("'%s'", strings.ReplaceAll(command, "'", "'\"'\"'"))
```
r2k1, line 104: it feels like voodoo magic. I wonder if there is a function from standard SDK to achieve similar results

e2e/scenario_win_test.go
```diff
@@ -70,8 +87,60 @@ func Test_Windows23H2Gen2(t *testing.T) {
...
 			},
 		},
 	})
 }
+
+func makeExecutablePowershellCommand(steps []string) string {
```
cameronmeissner, line 99: can we move the validation-related code over to validation.go?

r2k1 (approved):

e2e/validators.go
```diff
@@ -30,30 +30,30 @@ func ValidateSysctlConfig(ctx context.Context, s *Scenario, customSysctls map[st
...
 
 func ValidateNvidiaSMIInstalled(ctx context.Context, s *Scenario) {
-	command := "nvidia-smi"
+	command := "sudo nvidia-smi"
 	execOnVMForScenarioValidateExitCode(ctx, s, command, 0, "could not execute nvidia-smi command")
 }
 
 func ValidateNvidiaModProbeInstalled(ctx context.Context, s *Scenario) {
-	command := "nvidia-modprobe"
+	command := "sudo nvidia-modprobe"
```
cameronmeissner, line 51: I think sudo is fine in testing contexts, but why exactly do we need to add it now?

e2e/exec.go
```diff
@@ -100,10 +100,22 @@ func sshString(vmPrivateIP string) string {
...
+			This works in an interesting way:
+			* We create a linux pod on a different node.
+			* On that pod, we add the ssh private key and then run the command on that node using ssh
+
+			This means we can run commands on the node even if they have failed to join to the cluster - as
+		    we're not relying on anything on the node under test, except for the ssh server.
+
+			It does mean we get into quoting complexity as we have to quote to run the command on the pod, and quote again to pass the command through ssh.
+	*/
+	commandToExecute := fmt.Sprintf(`set -x && echo '%s' > %[2]s && chmod 0600 %[2]s && %s %s`, sshPrivateKey, sshKeyName(vmPrivateIP), sshString(vmPrivateIP), quoteForBash(command))
```
cameronmeissner, line 118: can we number each '%s'? right now I find this quite hard to read

e2e/scenario_win_test.go
r2k1: oh, no please. I think 2 functions with a common prefix/suffix is much easier to follow than a chain of function calls like this.

Can be a single wrappper around checking if scenario is for a windows/linux. But I don't think it's necessary going to improve anything.

e2e/exec.go
cameronmeissner: also, do we need the set -x? if we leave it in we'd be exposing the private key

r2k1:
I assume we don't need `sudo` everywhere we use currently. But you can't know it without a long trial and error?

e2e/validation.go
```diff
@@ -34,7 +34,7 @@ func ValidateWASM(ctx context.Context, s *Scenario, nodeName string) {
 }
 
 func ValidateCommonLinux(ctx context.Context, s *Scenario) {
-	execResult := execOnVMForScenarioValidateExitCode(ctx, s, "cat /etc/default/kubelet", 0, "could not read kubelet config")
+	execResult := execOnVMForScenarioValidateExitCode(ctx, s, "sudo cat /etc/default/kubelet", 0, "could not read kubelet config")
```
cameronmeissner, line 37: sorry I'm probably missing something obvious, but why do we need the `sudo` everywhere now? I think it's fine to use in testing contexts, though I'm mainly just wondering why we need it now

r2k1: I think currently it implicitly happens everywhere. I assume windows doesn't like it.
So the change is to call it explicitly everywhere now to allow windows behave differently.

e2e/exec.go
r2k1: Is it possible to use something like this? 
```
ssh user@remote "bash -s" <<EOF
echo "Hello, world!"
echo 'This script contains "quotes" and other characters.'
EOF
```

e2e/scenario_win_test.go
timmy-wright: Yes! Will do this.

e2e/vmss.go
```diff
@@ -129,7 +129,7 @@ func extractLogsFromVMLinux(ctx context.Context, s *Scenario) {
 
 	var logFiles = map[string]string{}
 	for file, sourceCmd := range commandList {
-		execResult, err := execOnVM(ctx, s.Runtime.Cluster.Kube, privateIP, pod.Name, string(s.Runtime.SSHKeyPrivate), sourceCmd)
+		execResult, err := execOnVM(ctx, s.Runtime.Cluster.Kube, privateIP, pod.Name, string(s.Runtime.SSHKeyPrivate), fmt.Sprintf("sudo %s", sourceCmd))
```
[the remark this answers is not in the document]
timmy-wright: Good idea.

e2e/exec.go
timmy-wright: I'm actually working on something like this. Was going to make it a separate PR, but will add it to this PR.

e2e/validators.go
```diff
@@ -99,6 +99,9 @@ func makeExecutableCommand(steps []string) string {
 	quotedCommand := strings.Replace(joinedCommand, "'", "'\"'\"'", -1)
 
 	command := fmt.Sprintf("bash -c '%s'", quotedCommand)
+	if asRoot {
+		command = fmt.Sprintf("sudo %s", command)
+	}
```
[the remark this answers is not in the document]
timmy-wright: I actually noticed a couple of potential double sudos when making this change - so we've made things better already! Will try to remove this one too.

e2e/scenario_helpers_test.go
```diff
@@ -122,6 +122,16 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
...
 	var err error
 	s.Runtime.SSHKeyPrivate, s.Runtime.SSHKeyPublic, err = getNewRSAKeyPair()
+	publicKeyData := datamodel.PublicKey{KeyData: string(s.Runtime.SSHKeyPublic)}
+
+	// check it all.
+	if s.Runtime.NBC != nil && s.Runtime.NBC.ContainerService != nil && s.Runtime.NBC.ContainerService.Properties != nil && s.Runtime.NBC.ContainerService.Properties.LinuxProfile != nil {
+		if s.Runtime.NBC.ContainerService.Properties.LinuxProfile.SSH.PublicKeys == nil {
+			s.Runtime.NBC.ContainerService.Properties.LinuxProfile.SSH.PublicKeys = []datamodel.PublicKey{}
+		}
+		s.Runtime.NBC.ContainerService.Properties.LinuxProfile.SSH.PublicKeys = append(s.Runtime.NBC.ContainerService.Properties.LinuxProfile.SSH.PublicKeys, publicKeyData)
```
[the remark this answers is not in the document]
timmy-wright: Good idea about the comment. Your assumption is correct.

e2e/validation.go
timmy-wright: Yeah, by default we were adding 'sudo" to the start of every command executed on a node as part of a test. Given windows uses powershell, adding "sudo" to the command just won't work. It should be the decision of the test that creates the command anyway.

e2e/scenario_win_test.go
timmy-wright: I'm working on a PR to get those validations working - that I'm going to include in this PR due to some of the other comments. So I'll resolve this comment by getting those working.

e2e/scenario_win_test.go
timmy-wright: I might change the validator to look at the OS running on the node and made decisions based on that. Because different versions of windows might have different powershell versions 😢

timmy-wright: Have combined them into a single wrapper.

e2e/exec.go
```diff
@@ -97,29 +98,80 @@ func sshKeyName(vmPrivateIP string) string {
...
+}
+
+func execBashCommandOnVM(ctx context.Context, s *Scenario, vmPrivateIP, jumpboxPodName, sshPrivateKey, command string) (*podExecResult, error) {
+	image := &config.Image{
+		OS: config.OSUbuntu,
+	}
+	return execScriptOnVm(ctx, s, vmPrivateIP, jumpboxPodName, sshPrivateKey, []string{command}, image)
+}
+
+func execScriptOnVm(ctx context.Context, s *Scenario, vmPrivateIP, jumpboxPodName, sshPrivateKey string, script []string, os *config.Image) (*podExecResult, error) {
```
r2k1, line 115: I feel like passing `script string` is easier and more straightforward

e2e/exec.go
```diff
@@ -97,29 +98,80 @@ func sshKeyName(vmPrivateIP string) string {
...
+		break
+	}
+
+	scriptWithLineBreaks := strings.Join(script, "\n")
+	steps := []string{
+		fmt.Sprintf("echo '%[1]s' > %[2]s", sshPrivateKey, sshKeyName(vmPrivateIP)),
+		"set -x",
+		fmt.Sprintf("echo %[1]s > %[2]s", quoteForBash(scriptWithLineBreaks), scriptFileName),
+		fmt.Sprintf("chmod 0600 %s", sshKeyName(vmPrivateIP)),
+		fmt.Sprintf("chmod 0755 %s", scriptFileName),
```
r2k1, line 150: do you need to chmod after scp or before?

e2e/exec.go
```diff
@@ -97,29 +98,80 @@ func sshKeyName(vmPrivateIP string) string {
...
 
-func execOnVM(ctx context.Context, kube *Kubeclient, vmPrivateIP, jumpboxPodName, sshPrivateKey, command string) (*podExecResult, error) {
-	sshCommand := fmt.Sprintf(`echo '%s' > %[2]s && chmod 0600 %[2]s && %s`, sshPrivateKey, sshKeyName(vmPrivateIP), sshString(vmPrivateIP))
-	sshCommand = sshCommand + " sudo"
-	commandToExecute := fmt.Sprintf("%s %s", sshCommand, command)
+func quoteForBash(command string) string {
+	return fmt.Sprintf("'%s'", strings.ReplaceAll(command, "'", "'\"'\"'"))
+}
+
+func execBashCommandOnVM(ctx context.Context, s *Scenario, vmPrivateIP, jumpboxPodName, sshPrivateKey, command string) (*podExecResult, error) {
```
r2k1, line 108: You need to decide either separate `execBash` from `execPowershell` or combine.

Currently you're mixing approaches.

r2k1 (approved):

e2e/exec.go
timmy-wright: Yeah, but I wanted to pass around an array of lines just in case we need to use different line endings. Who knows with Windows!

e2e/exec.go
timmy-wright: I think scp maintains permissions so that's why I chmod before - otherwise it's a seperate command to change the permissions before executing it. Although we're specifying the interpreter so the chmod is probably unnecessary anyway.

e2e/exec.go
timmy-wright: This particular is only bash and only used in one specific case - there is no execPowershell. The rest is "execScriptOnVm" and we decide on the interpreter at the last minute based on the OS type of the node.

PR #5636, opened 2025-01-22
chore: don't run e2e tests if the VHD isn't being built

timmy-wright:
**What type of PR is this?**

/kind chore

Currently if someone runs the windows VHD build pipelines for 1 VHD, the e2e tests are run for all windows VHDs. This change disables the e2e tests for the VHDs not being built.
 
Sample pipeline: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=113082893&view=results

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
@@ -33,11 +33,11 @@ parameters:
 
 stages:
   - stage: build_${{ parameters.stageName }}
+    condition: eq('${{ parameters.build }}', True)
     dependsOn: [ ]
     jobs:
       - job: build_${{ parameters.stageName }}
         dependsOn: []
-        condition: eq('${{ parameters.build }}', True)
         timeoutInMinutes: 180
         steps:
           - bash: |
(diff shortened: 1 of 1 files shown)


.pipelines/templates/.build-and-test-windows-vhd-template.yaml
```diff
@@ -57,6 +57,7 @@ stages:
 
   - stage: e2e_${{ parameters.stageName }}
     dependsOn: build_${{ parameters.stageName }}
+    condition: eq('${{ parameters.build }}', True)
```
r2k1, line 60: do you need the condition if you already have `dependsOn`?

r2k1 (approved):

.pipelines/templates/.build-and-test-windows-vhd-template.yaml
timmy-wright: It turns out you are right and moving the condition for the build stage is sufficient. The prev placement of the build condition wasn't working. But I quite like having both.

PR #5643, opened 2025-01-22
fix: build cvm on cvm hardware

zachary-bailey:
**What type of PR is this?**

/kind feature

This PR changes the build process to use CVM hardware for CVM image builds. This will allow the use of `apt-get dist-upgrade` in the install scripts, resulting in the ability to install newer kernel versions.

This will fix the CVEs that are currently unmitigated for CVM nodepools.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1174,7 +1174,7 @@ stages:
               echo '##vso[task.setvariable variable=IMG_SKU]20_04-lts-cvm'
               echo '##vso[task.setvariable variable=IMG_VERSION]latest'
               echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
-              echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D16ds_v5'
+              echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_DC16ads_v5'
```
cameronmeissner, line 1177: let's make sure we have quota for this size

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -67,6 +71,21 @@ if [[ ${OS_TYPE} == "Linux" && ${ENABLE_TRUSTED_LAUNCH} == "True" ]]; then
       } \
     } \
   }"
+elif [[ ${OS_TYPE} == "Linux" && ${IMG_SKU} == "20_04-lts-cvm" ]]; then
```
cameronmeissner, line 74:
```suggestion
elif [ "${OS_TYPE}" == "Linux" ]  && [ "${IMG_SKU}" == "20_04-lts-cvm" ]; then
```

vhdbuilder/packer/init-variables.sh
```diff
@@ -71,6 +71,10 @@ if [ "$MODE" == "linuxVhdMode" ] && [ -z "${PACKER_BUILD_LOCATION}" ]; then
 	exit 1
 fi
 
+if [ "$MODE" == "linuxVhdMode" ] && [ -z "${CVM_PACKER_BUILD_LOCATION}" ]; then
+	CVM_PACKER_BUILD_LOCATION=$PACKER_BUILD_LOCATION
```
cameronmeissner, line 75: should we echo something out here to note that CVM_PACKER_BUILD_LOCATION wasn't specified and we're defaulting to PACKER_BUILD_LOCATION?

cameronmeissner: actually, could we just add something like

```
if [ -z "${CVM_PACKER_BUILD_LOCATION}" ]; then
    echo "CVM_PACKER_BUILD_LOCATION is set to ${CVM_PACKER_BUILD_LOCATION}, will use as PACKER_BUILD_LOCATION"
    PACKER_BUILD_LOCATION=$CVM_PACKER_BUILD_LOCATION
fi
```

then we wouldn't need to change how we set VNET_RG_NAME/VNET_NAME

vhdbuilder/packer/init-variables.sh
```diff
@@ -167,6 +180,8 @@ if [[ "${MODE}" == "linuxVhdMode" ]]; then
 		elif [[ "${IMG_OFFER,,}" == "azure-linux-3" ]]; then
 			# for Azure Linux 3.0, only use AzureLinux prefix
 			SIG_IMAGE_NAME="AzureLinux${SIG_IMAGE_NAME}"
+		elif [[ ${IMG_SKU} == "20_04-lts-cvm" ]]; then
+      SIG_IMAGE_NAME+="Specialized"
```
cameronmeissner, line 184: indenting seems weird here

vhdbuilder/packer/vhd-image-builder-cvm.json
```diff
@@ -0,0 +1,691 @@
+{
+  "variables": {
+    "subscription_id": "{{env `AZURE_SUBSCRIPTION_ID`}}",
+    "cvm_packer_build_location": "{{env `CVM_PACKER_BUILD_LOCATION`}}",
```
cameronmeissner, line 4: nit: this can just be `packer_build_location` and come from `PACKER_BUILD_LOCATION` like the other templates

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -45,7 +45,11 @@ SCAN_VM_ADMIN_PASSWORD="ScanVM@$(date +%s)"
 set -x
 
 RESOURCE_GROUP_NAME="$SCAN_RESOURCE_PREFIX-$(date +%s)-$RANDOM"
-az group create --name $RESOURCE_GROUP_NAME --location ${PACKER_BUILD_LOCATION} --tags "source=AgentBaker" "now=$(date +%s)" "branch=${GIT_BRANCH}"
+if [ "${ENVIRONMENT,,}" == "test" ] && [ "${IMG_SKU}" == "20_04-lts-cvm" ]; then
+    az group create --name $RESOURCE_GROUP_NAME --location ${CVM_PACKER_BUILD_LOCATION} --tags --tags "source=AgentBaker" "branch=${GIT_BRANCH}"
```
cameronmeissner, line 49: duplicate `--tags` directives

cameronmeissner: instead we could probably also do something like this again:

```
if [ -z "${CVM_PACKER_BUILD_LOCATION}" ]; then
    echo "CVM: will use ${CVM_PACKER_BUILD_LOCATION} as PACKER_BUILD_LOCATION"
    PACKER_BUILD_LOCATION=$CVM_PACKER_BUILD_LOCATION
fi
```

and leave the `az group create ...` unchanged

.pipelines/.vsts-vhd-builder-release.yaml
zachary-bailey: @cameronmeissner I was actually going to bring this up in standup today. I attempted to check in TME yesterday, but for some reason every time I tried to switch the directory it was unsuccessful.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
zachary-bailey: I thought I had gotten all of them, thanks!

vhdbuilder/packer/init-variables.sh
zachary-bailey: Perfect. Will change.

vhdbuilder/packer/vhd-scanning.sh
zachary-bailey: Will do the same as above.

vhdbuilder/packer/init-variables.sh
```diff
@@ -111,6 +116,7 @@ if [ -z "${VNET_NAME}" ]; then
 	fi
 fi
 
+
```
Devinwong, line 119: nit: extra line

vhdbuilder/packer/init-variables.sh
djsly: Good old TAB vs Space.   Zach you should set your VScode to use TAB over Space :)

djsly (approved):

vhdbuilder/packer/init-variables.sh
Devinwong: I think it's because some indents are 2-space and some are tabs

vhdbuilder/packer/test/run-test.sh
```diff
@@ -75,6 +75,10 @@ if [ "${OS_TYPE}" == "Linux" ] && [ "${ENABLE_TRUSTED_LAUNCH}" == "True" ]; then
   TARGET_COMMAND_STRING+="--security-type TrustedLaunch --enable-secure-boot true --enable-vtpm true"
 fi
 
+if [ "${OS_TYPE}" == "Linux" ] && [ ${IMG_SKU} == "20_04-lts-cvm" ]; then
+    TARGET_COMMAND_STRING="--size Standard_DC8ads_v5 --security-type ConfidentialVM --enable-secure-boot true --enable-vtpm true --os-disk-security-encryption-type VMGuestStateOnly --specialized true"
```
Devinwong, line 79: +=?

vhdbuilder/packer/init-variables.sh
zachary-bailey: Thanks for the call out!

vhdbuilder/packer/init-variables.sh
zachary-bailey: Okay, I think it was initially tab. I changed to space and VSCode acknowledged a change. But you want it to be tab?

zachary-bailey: It is currently tab again.

vhdbuilder/packer/test/run-test.sh
zachary-bailey: If I made it +=, it would still have the VM size from standard images in the string, so I had to make it re write the whole string.

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -63,6 +63,10 @@ if [[ "${OS_TYPE}" == "Linux" && "${ENABLE_TRUSTED_LAUNCH}" == "True" ]]; then
     VM_OPTIONS+=" --security-type TrustedLaunch --enable-secure-boot true --enable-vtpm true"
 fi
 
+if [ "${OS_TYPE}" == "Linux" ] && [ ${IMG_SKU} == "20_04-lts-cvm" ]; then
+    VM_OPTIONS="--size Standard_DC8ads_v5 --security-type ConfidentialVM --enable-secure-boot true --enable-vtpm true --os-disk-security-encryption-type VMGuestStateOnly --specialized true"
```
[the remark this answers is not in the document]
zachary-bailey: We do not have any ARM64 CVMs. Just the one AMD image.

vhdbuilder/packer/test/run-test.sh
Devinwong: > If I made it +=, it would still have the VM size from standard images in the string, so I had to make it re write the whole string.

but will you not miss the FIPS setting if it's `=` ? For example `--security-type TrustedLaunch`

vhdbuilder/packer/vhd-scanning.sh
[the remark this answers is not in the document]
Devinwong: I see. Could you please leave a comment for the reason so other people know this is intended? Thanks

vhdbuilder/packer/test/run-test.sh
zachary-bailey: This condition will only be met if `${IMG_SKU} == "20_04-lts-cvm"`

`"20_04-lts-cvm"` is the IMG_SKU only for CVM, not for 2004 or FIPs in general.

For example the IMG_SKU for `2004fipscontainerd` is `20_04-lts`. They are all unique per SKU. So this can only execute on our CVM image.

vhdbuilder/packer/vhd-scanning.sh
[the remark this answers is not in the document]
zachary-bailey: Done! @Devinwong

vhdbuilder/packer/test/run-test.sh
Devinwong: I see. So you are saying actually `"${ENABLE_TRUSTED_LAUNCH}" == "True"` is equivalent to `${IMG_SKU}==20_04-lts`, right? If that's the case, please leave some comments to explain this. Thanks.

Devinwong (approved):

vhdbuilder/packer/test/run-test.sh
zachary-bailey: `ENABLE_TRUSTED_LAUNCH` is a boolean feature flag assigned to some of our images, some of which are Ubuntu and some of which are Mariner/AZLinux. The feature flag alerts Packer that these Packer VMs need to have the TrustedLaunch feature enabled because they need TPM measurement files to run.

So, the above conditional on line 70 checks to see if the `--security-type TrustedLaunch --enable-secure-boot true --enable-vtpm true` flags need to be appended to the command string to enable the VMs successful provisioning. It is unrelated to CVM, though.

vhdbuilder/packer/init-variables.sh
djsly: elif should be aligned

.pipelines/.vsts-vhd-builder-release.yaml
cameronmeissner: hmm, let me try to check

cameronmeissner: checked quota, should be fine for now (100 cores), though can increase later if needed

cameronmeissner (approved):

PR #5646, opened 2025-01-22
fix: remove encoding from scriptless systctl content

Devinwong:
**What type of PR is this?**

/kind bug

Remove unnecessary encoding from systctl content. Validated in a local e2e scenario test.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/aks-node-controller/parser/helper.go
+++ b/aks-node-controller/parser/helper.go
@@ -411,7 +411,7 @@ func getUlimitContent(u *aksnodeconfigv1.UlimitConfig) string {
 		m["LimitMEMLOCK"] = u.GetMaxLockedMemory()
 	}
 
-	return base64.StdEncoding.EncodeToString([]byte(header + createSortedKeyValuePairs(m, " ")))
+	return header + createSortedKeyValuePairs(m, " ")
 }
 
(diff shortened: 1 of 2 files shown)


aks-node-controller/parser/helper.go
```diff
@@ -411,7 +411,7 @@ func getUlimitContent(u *aksnodeconfigv1.UlimitConfig) string {
 		m["LimitMEMLOCK"] = u.GetMaxLockedMemory()
 	}
 
-	return base64.StdEncoding.EncodeToString([]byte(header + createSortedKeyValuePairs(m, " ")))
```
xmzhao0822, line 414: Nice catch!

r2k1 (approved):

aks-node-controller/parser/helper.go
```diff
@@ -411,7 +411,7 @@ func getUlimitContent(u *aksnodeconfigv1.UlimitConfig) string {
...
+	return header + createSortedKeyValuePairs(m, " ")
```
Devinwong, line 414: encoding is not needed

xmzhao0822 (commented):
Nice catch on the encoding logic, no need here

PR #4879, opened 2024-08-28
<stale> feat: use oras to pull binaries - keeping as notes

AlisonB319:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #5393, opened 2024-12-09
refactor: containerd config registry host file to cse

bingosummer:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Move the containerd config registry host file from custom data to the CSE provisioning scripts to make the scriptless/self-contained effort.

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #5650, opened 2025-01-23
feat: add support for non-anonymous login for Network Isolated Clusters

AlisonB319:
**What type of PR is this?**

This PR adds support for non-anonymous ACR pull. It currently is not available, as there is a front end block on using a non-anonymous ACR. At the moment there is no way to get the kubelet identity, so the login cannot be preformed. However, it was decided that we wanted to have the logic in place. 

This is a combination of two PRs 
https://github.com/Azure/AgentBaker/pull/4879/
https://github.com/Azure/AgentBaker/pull/5508/

TODO:
Going to attach a vmss identity through e2es to test out the functionality of oras login

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,40 @@ getPrimaryNicIP() {
     echo "$ip"
 }
 
+orasLogin() {
```
AlisonB319, line 408: @djsly right now the flow is 
check anonymous pull -> success return (this is the current case in prod atm)

check anonymous pull -> failure -> check non-anonymous 

In the chat, it was stated to do the anonymous check first, it might be better to do anonymous first since that is what is currently happening in production? When the customer turns off anonymous in the ACR, then the first check should fail and the oras login will take place. 

it currently does the anonymous check by pulling a helloworld from the acr 

this flow is mostly XInhe's logic, I thought it made sense, but I can adjust as we see fit.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,40 @@ getPrimaryNicIP() {
...
+    echo "Checking access to ACR with anonymous pull"
+    logs_to_events "AKS.CSE.orasLogin.retrycmd_acr_access_check_anon" retrycmd_acr_access_check 10 1 "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}"
+    ret_check_anonymous=$?
+    if [ $ret_check_anonymous -eq 0 ]; then
+        # this conditional should always be hit since only anonymous pull is supported at the moment
+        echo "bootstarp acr '$BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER' access check passed"
+        return 0
+    elif [ $ret_check_anonymous -ne $ERR_ORAS_PULL_UNAUTHORIZED ]; then
```
cameronmeissner, line 416: nit: redundant elif

AlisonB319: this actually isn't redudant, it checks if it is anything other than unauthorized as we fail so the flow is 

successful pull -> return 0 
anything else other than not-authorized -> return failure 

(else) aka non-athorized
try the oras login

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,40 @@ getPrimaryNicIP() {
...
+        echo "retrycmd_acr_access_check failed with something other than unauthorized"
+        return $ret_check_anonymous
+    fi
+
+    # todo verify that an ACR with anonymous turned off throws an unauthorized error
+    # we cannot support anonymous pull until kubelet identity exists on the vmss. This should not be hit until the feature becomes available
+    echo "Failed to access ACR with anonymous pull, will try use kubelet identity for non-anonymous pull"
+    acr_login_server="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER%/}"
+    logs_to_events "AKS.CSE.orasLogin.oras_login_with_identity" oras_login_with_identity $acr_login_server $USER_ASSIGNED_IDENTITY_ID $TENANT_ID
```
cameronmeissner, line 425: nit: `oras_login_with_kubelet_identity`?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,40 @@ getPrimaryNicIP() {
...
+    ret_login=$?
+    if [[ "$ret_login" != 0 ]]; then
```
cameronmeissner, line 427:
```suggestion
    if [ $ret_login$ -ne 0 ]; then
```

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,40 @@ getPrimaryNicIP() {
...
+        echo "Failed to login to oras with kubelet identity"
+        exit $ret_login
+    fi
+
+    logs_to_events "AKS.CSE.orasLogin.retrycmd_acr_access_check_non_anon" retrycmd_acr_access_check 10 1 "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}"
+    ret_check_after_login=$?
+    if [[ "$ret_check_after_login" != 0 ]]; then
```
cameronmeissner, line 434:
```suggestion
    if [ $ret_check_after_login -ne 0 ]; then
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -287,6 +287,37 @@ retrycmd_curl_file() {
         fi
     done
 }
+retrycmd_acr_access_check() {
+    access_retries=$1; wait_sleep=$2;
```
cameronmeissner, line 291: nit: we can use `local` for these too

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -287,6 +287,37 @@ retrycmd_curl_file() {
...
+    local acr_url=$3
+
+    echo "${access_retries} retries for acr access check"
+    sample_image="$acr_url/mcr/hello-world:latest"
```
cameronmeissner, line 295: this image is used to check that the cache is correct? seems a bit brittle, is there a different way we can check this?

cameronmeissner: if we need to do it this way, I'd opt for also removing this image from disk afterwards as well

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +714,42 @@ removeKubeletFlag() {
     fi
 }
 
+oras_login_with_identity() {
```
cameronmeissner, line 717: we might want to `set +x` at the top of this so we don't log the AAD or ACR tokens

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +714,42 @@ removeKubeletFlag() {
...
+    local acr_url=$1
+    local client_id=$2
+    local tenant_id=$3
+
+    raw_access_token=$(curl "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com%2F&client_id=$client_id" -H Metadata:true -s)
```
cameronmeissner, line 722: probably want to wrap the calls to IMDS + ACR exchange endpoint in retries

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: a quick chatgpt check said that I could try pulling a non-existant image 
oras pull <your-registry-name>.azurecr.io/nonexistent:tag
and if it says like error unknown than your logged in, verses an auth error - I'll look into this when I get the e2e testing working

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +714,42 @@ removeKubeletFlag() {
...
+        return $ERR_ORAS_PULL_UNAUTHORIZED
+    fi
+    ACR_TOKEN=$(echo "$raw_acr_token" | jq -e .refresh_token -r 2>/dev/null)
+    if [ -z "$ACR_TOKEN" ] || [ "$ACR_TOKEN" = "null" ]; then
+        echo "failed to parse acr token"
+        return $ERR_ORAS_PULL_UNAUTHORIZED
+    fi
+
+    oras login $acr_url --identity-token $ACR_TOKEN --registry-config ${ORAS_REGISTRY_CONFIG_FILE}
+    ret=$?
```
cameronmeissner, line 747: nit: don't usually need to capture exit codes like this when the command is running within the same shell as the script itself:

```suggestion
    if ! oras login $acr_url --identity-token $ACR_TOKEN --registry-config ${ORAS_REGISTRY_CONFIG_FILE}; then
        echo "failed to login to acr '$acr_url' with identity token"
        return $ERR_ORAS_PULL_UNAUTHORIZED
fi
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner:
```
cam@workbook ~$ oras repo                                                                                                                                                                                   
Repository operations

Usage:
  oras repo [command]

Aliases:
  repo, repository

Available Commands:
  ls          List the repositories under the registry
  tags        Show tags of the target repository

Flags:
  -h, --help   help for repo

Use "oras repo [command] --help" for more information about a command.
```

would `oras repo ls` work?

parts/linux/cloud-init/artifacts/cse_config.sh
cameronmeissner: I was thinking:

```
if [ $ret_check_anonymous -eq 0 ]; then
        # this conditional should always be hit since only anonymous pull is supported at the moment
        echo "bootstarp acr '$BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER' access check passed"
        return 0
fi
if [ $ret_check_anonymous -ne $ERR_ORAS_PULL_UNAUTHORIZED ]; then
        echo "retrycmd_acr_access_check failed with something other than unauthorized"
        return $ret_check_anonymous
fi
```

cameronmeissner: I'm fine either way

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: sounds good, for some reason set +x makes the shellscript unit tests very unhappy, so I'll need to dig a bit here.

they stall out and then throw a huge error message

parts/linux/cloud-init/artifacts/cse_config.sh
djsly: I'm not sure we should download a binary to test access.  

Also, I think that they are looking to not support anonymous pull at all in the future, so it might be best to try non-anonymous first ?

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: ohhh I actually ran into this myself recently, it's seemingly because shellspec is also invoking `set` within the mocked context

I found that if you invoke the function you're trying test using `run` (as opposed to `call`) in the spec, then the function will be called within a sub-shell rather than the parent shell where shellspec is actually running - that will prevent the breaking and infinte hang you're seeing

the other option is to just have the caller invoke `set` instead of doing it directly within the function itself

cameronmeissner: ah I see you're already using `run` in your other specs, so yeah you can just do that to test `oras_login_with_kubelet_identity` as well

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -405,6 +405,40 @@ getPrimaryNicIP() {
     echo "$ip"
 }
 
+orasLogin() {
+    echo "Checking access to ACR with anonymous pull"
+    logs_to_events "AKS.CSE.orasLogin.retrycmd_acr_access_check_anon" retrycmd_acr_access_check 10 1 "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}"
```
anujmaheshwari1, line 410: nit: keep naming consistent, i.e either consider adding _anon to the actual function call, or remove it from the logs_to_events name

parts/linux/cloud-init/artifacts/cse_helpers.sh
anujmaheshwari1: what happens if the hello world image, for some reason ever, is removed? I agree with Cameron, that this is brittle.

parts/linux/cloud-init/artifacts/cse_config.sh
AlisonB319: Curious on your thoughts here, having a slightly different name allows me to essentially mock both function calls in the unit test 
```
logs_to_events() {
                logs_to_events() {
                case "$1" in
                    "AKS.CSE.orasLogin.retrycmd_acr_access_check_anon")
                        return $ERR_ORAS_PULL_UNAUTHORIZED
                        ;;
                    "AKS.CSE.orasLogin.oras_login_with_kubelet_identity")
                        return 0
                        ;;
                    "AKS.CSE.orasLogin.retrycmd_acr_access_check_non_anon")
                        return 1 
                        ;;
                    *)
                        return -1
                        ;;
                esac
            }
```
also while logging, it might make it more clear which attempt its trying, anonymous vs non-anonymous.

However, I am going to tweak the flow so I'll see if I even end up making the retrycmd_acr_access_check call twice

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +708,45 @@ removeKubeletFlag() {
     fi
 }
 
+oras_login_with_kubelet_identity() {
+    local acr_url=$1
+    local client_id=$2
+    local tenant_id=$3
+
+    if [ -z "$client_id" ] || [ -z "$tenant_id" ]; then
+        echo "client_id or tenant_id are not set. Oras login is not possible, proceeding with anonynous pull"
```
Devinwong, line 717: anonymous 😆

e2e/vmss.go
```diff
@@ -53,6 +53,16 @@ func createVMSS(ctx context.Context, s *Scenario) *armcompute.VirtualMachineScal
 	}
 
 	model := getBaseVMSSModel(s, customData, cse)
+	if attachVMIdentity {
+		// add acr pull identity
+		userAssignedIdentity := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.ManagedIdentity/userAssignedIdentities/%s", config.Config.SubscriptionID, config.ResourceGroupName, config.VMIdentityName)
```
Devinwong, line 58: Does user need to create a proper user assigned identity or it's automatically created?
I was just curious if there is no identity, does user have enough instructions to follow.

e2e/scenario_test.go
```diff
@@ -674,6 +674,40 @@ func Test_Ubuntu2204_AirGap(t *testing.T) {
 	})
 }
 
+// TODO: refactor NonAnonynous tests to use the same cluster as Anonymous airgap
```
Devinwong, line 677: NonAnonymous

spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh
```diff
@@ -168,4 +168,124 @@ Describe 'cse_helpers.sh'
             The variable KUBELET_NODE_LABELS should equal ''
         End
     End
+    Describe 'oras_login_with_kubelet_identity'
```
Devinwong, line 171: pretty cool to see these unit tests and mock!

e2e/kube.go
```diff
@@ -418,6 +427,11 @@ func podHTTPServerLinux(s *Scenario) *corev1.Pod {
 			NodeSelector: map[string]string{
 				"kubernetes.io/hostname": s.Runtime.KubeNodeName,
 			},
+			ImagePullSecrets: []corev1.LocalObjectReference{
```
fseldow, line 430: i think we can add one TODO here.
After Bin's pr https://github.com/Azure/AgentBaker/pull/5393 merged, it should be able to run identity pull.
But anyway this comment is not block to this pr.
@bingosummer

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +708,45 @@ removeKubeletFlag() {
     fi
 }
 
+oras_login_with_kubelet_identity() {
```
fseldow, line 711: should we put `set -x` and `set +x` inside the function to avoid another reference may forget this?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +708,45 @@ removeKubeletFlag() {
...
+    local tenant_id=$3
+
+    if [ -z "$client_id" ] || [ -z "$tenant_id" ]; then
+        echo "client_id or tenant_id are not set. Oras login is not possible, proceeding with anonymous pull"
+        return 
+    fi
+
+    access_url="http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/&client_id=$client_id"
+    raw_access_token=$(retrycmd_get_access_token_for_oras 10 5 $access_url)
+    if [ -z "$raw_access_token" ] || [[ "$raw_access_token" == *"error"* ]]; then
```
fseldow, line 723: not block just curious.
Do we have sample error msg or scenario for `"$raw_access_token" == *"error"*`

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -115,9 +115,9 @@ ERR_CNI_VERSION_INVALID=206 # reference CNI (not azure cni) needs a valid versio
 ERR_ORAS_PULL_K8S_FAIL=207 # Error pulling kube-node artifact via oras from registry
 ERR_ORAS_PULL_FAIL_RESERVE_1=208 # Error pulling artifact with oras from registry
 ERR_ORAS_PULL_CONTAINERD_WASM=209 # Error pulling containerd wasm artifact with oras from registry
-ERR_ORAS_PULL_FAIL_RESERVE_3=210 # Error pulling artifact with oras from registry
-ERR_ORAS_PULL_FAIL_RESERVE_4=211 # Error pulling artifact with oras from registry
-ERR_ORAS_PULL_FAIL_RESERVE_5=212 # Error pulling artifact with oras from registry
+ERR_ORAS_PULL_NETWORK_ACCESS=210 # Error pulling artifact with oras from registry with connection issue
```
fseldow, line 118: we do not identify network access err and incorrect cache rule. maybe we can still set it as `ERR_ORAS_PULL_FAIL_RESERVE_4`?

pkg/agent/testdata/AKSUbuntu1604+Containerd/line9.sh
```diff
@@ -231,6 +231,32 @@ retrycmd_get_tarball_from_registry_with_oras() {
...
+        REFRESH_TOKEN_OUTPUT=$(timeout 60 curl -v -s -X POST -H "Content-Type: application/x-www-form-urlencoded" \
+            -d "grant_type=access_token&service=$acr_url&tenant=$tenant_id&access_token=$ACCESS_TOKEN" \
+            https://$acr_url/oauth2/exchange)
+        if [ -n "$REFRESH_TOKEN_OUTPUT" ]; then 
+            echo "$REFRESH_TOKEN_OUTPUT"
+            return 0
+        fi
+        sleep $wait_sleep
+    done
+    return 1
```
fseldow, line 258: should we view timeout as network access errror, or maybe set one error code as login timeout?

pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line33.sh
```diff
@@ -81,6 +81,12 @@ if [[ ${ID} != "mariner" ]] && [[ ${ID} != "azurelinux" ]]; then
     logs_to_events "AKS.CSE.removeManDbAutoUpdateFlagFile" removeManDbAutoUpdateFlagFile
 fi
 
+if [[ ! -z ${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER} ]]; then
+    set +x 
+    logs_to_events "AKS.CSE.orasLogin.oras_login_with_kubelet_identity" oras_login_with_kubelet_identity "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER%/}" $USER_ASSIGNED_IDENTITY_ID $TENANT_ID || exit $?
```
fseldow, line 86: can we ignore error and just log it for now? since current existed customer cluster may not have acrpull permission. It may bring breaking change

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +708,45 @@ removeKubeletFlag() {
     fi
 }
 
+oras_login_with_kubelet_identity() {
+    local acr_url=$1
+    local client_id=$2
+    local tenant_id=$3
+
+    if [ -z "$client_id" ] || [ -z "$tenant_id" ]; then
```
AlisonB319, line 716: So instead of doing any kind of oras pull to see if non-authenticated works, I instead just return if the client id or tenant id is not set.

@fseldow can you confirm the front end validation that customers, at this time, cannot create a non-anonymous NIC and since the identity cannot be attached the VM yet, this will suffice?

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: I'm open to opinions on how we want to set up the -x +x, for docker login we have 

```
if [ "${NEEDS_DOCKER_LOGIN}" == "true" ]; then
    set +x
    docker login -u $SERVICE_PRINCIPAL_CLIENT_ID -p $SERVICE_PRINCIPAL_CLIENT_SECRET "${AZURE_PRIVATE_REGISTRY_SERVER}"
    set -x
fi
```
in master which was the flow I was following. But I'd love to hear opinions on this

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: `ACCESS_TOKEN_OUTPUT='{"error":"invalid_request","error_description":"Identity not found"}'` this is some example output

e2e/vmss.go
AlisonB319: Sorry I left a comment as a review comment and it never actually posted! But yes the user will need to have a proper user signed identity and give it the correct ACR Pull permissions. There are already exsting microsoft docs around this + the feature in general. 

Right now, however, in CSE we don't actually have access to the VM identity yet. So by mocking this in our e2es, we are able to test out the what the flow looks like once RP has finished the work to let us have access to the identity in CSE :)

pkg/agent/testdata/AKSUbuntu1604+CustomKubeletConfig+CustomLinuxOSConfig/line33.sh
AlisonB319: I thought you said there was front end validation that a customer can't create a non-anonymous ACR? I left another comment in helpers to make sure we're on the same page

parts/linux/cloud-init/artifacts/cse_helpers.sh
fseldow: From my side the clientid and tenant id is not related with if acr enable anonymous access. Because aks cluster will always bind kubelet identity to node and pass client id and tenant id.

Current frontend logic is that if acr anonymous access is disabled, we will block the cluster to enable network isolated feature.
In the future, we will remove this validation or change it to force disable anonymous access

e2e/kube.go
AlisonB319: This PR is already merged and I still have to provide an image secret to the pod so that they have access to the ACR. Are you saying the pod should be able to run identity pull?

e2e/aks_model.go
```diff
@@ -258,9 +266,106 @@ func createPrivateAzureContainerRegistry(ctx context.Context, t *testing.T, clus
...
+	if err := addCacheRulesToPrivateAzureContainerRegistry(ctx, t, config.ResourceGroupName, privateACRName); err != nil {
 		return fmt.Errorf("failed to add cache rules to private acr: %w", err)
 	}
+
+	return nil
+}
+
+func createKubernetesSecret(ctx context.Context, t *testing.T, namespace, kubeconfigPath, secretName, registryName, username, password string) error {
+	t.Logf("Creating Kubernetes secret %s in namespace %s", secretName, namespace)
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
```
cameronmeissner, line 297: we already create a *Kubeclient when preparing the cluster, can we just use that instead?

e2e/aks_model.go
```diff
@@ -258,9 +266,106 @@ func createPrivateAzureContainerRegistry(ctx context.Context, t *testing.T, clus
...
+		},
+	}
+	_, err = clientset.CoreV1().Secrets(namespace).Create(ctx, secret, metav1.CreateOptions{})
+	if err != nil {
+		if !errorsk8s.IsAlreadyExists(err) {
+			t.Logf("failed to create Kubernetes secret: %v", err)
+			return err
+		}
+	}
+	t.Logf("Kubernetes secret created")
```
cameronmeissner, line 337: can we log the name of the secret?

e2e/kube.go
```diff
@@ -285,14 +285,14 @@ func getClusterKubeconfigBytes(ctx context.Context, resourceGroupName, clusterNa
 }
 
 // this is a bit ugly, but we don't want to execute this piece concurrently with other tests
-func (k *Kubeclient) EnsureDebugDaemonsets(ctx context.Context, t *testing.T, isAirgap bool) error {
-	ds := daemonsetDebug(t, hostNetworkDebugAppLabel, "nodepool1", true, isAirgap)
+func (k *Kubeclient) EnsureDebugDaemonsets(ctx context.Context, t *testing.T, isAirgap bool, privateACRName string) error {
```
cameronmeissner, line 288: feels clunky passing `privateACRName` around everywhere, can't we just access it directly from the config package?

e2e/config/config.go
```diff
@@ -13,11 +13,12 @@ import (
...
-	Config            = mustLoadConfig()
-	Azure             = mustNewAzureClient()
-	ResourceGroupName = "abe2e-" + Config.Location
-	VMIdentityName    = "abe2e-vm-identity"
-	PrivateACRName    = "privateacre2e" + Config.Location + "dev" // TODO (alburgess): remove dev once CCOA is over
+	Config                = mustLoadConfig()
+	Azure                 = mustNewAzureClient()
+	ResourceGroupName     = "abe2e-" + Config.Location
+	VMIdentityName        = "abe2e-vm-identity"
+	PrivateACRNameNotAnon = "privateace2enonanonpull" + Config.Location // will have anonymous pull enabled
```
cameronmeissner, line 20: nit: `private-acr-e2e-non-anon-pull`?

e2e/config/config.go
```diff
@@ -13,11 +13,12 @@ import (
...
+	PrivateACRName        = "privateacre2e" + Config.Location           // will not have anonymous pull enabled
```
cameronmeissner, line 21: nit: `private-acr-e2e`?

e2e/kube.go
cameronmeissner: oh, I see 

```
privateACRName := config.PrivateACRName
if isNonAnonymousPull {
privateACRName = config.PrivateACRNameNotAnon
}
```

e2e/scenario_helpers_test.go
```diff
@@ -135,7 +141,13 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
 	}
 
 	require.NoError(s.T, err)
-	createVMSS(ctx, s)
+
+	attachVMIdentity := false
```
cameronmeissner, line 145: can just simplify this to `createVMSS(ctx, s, s.Tags.NonAnonymousACR)`

cameronmeissner: actually, does that mean we can just resolve the bool directly within `createVMSS` and leave off the extra parameter?

e2e/kube.go
cameronmeissner: how about we create a method on the Configuration object which takes in whatever it needs (from a quick glance, maybe just tags?) and resolves the privateACRName - that way we can just call the method directly on config.Config isntead of having to pass this around everywhere

e2e/scenario_test.go
```diff
@@ -717,6 +717,40 @@ func Test_Ubuntu2204_AirGap(t *testing.T) {
 	})
 }
 
+// TODO: refactor NonAnonymous tests to use the same cluster as Anonymous airgap
+// or deprecate anonymous ACR airgap tests once it is unsupported
+func Test_Ubuntu2204_AirGap_NonAnonymousACR(t *testing.T) {
+	ctx := newTestCtx(t)
+	identity, err := config.Azure.UserAssignedIdentities.Get(ctx, config.ResourceGroupName, config.VMIdentityName, nil)
```
cameronmeissner, line 724: could we resolve the identity somewhere directly within `RunScenario` (where testCtx is already created) so we don't have to create duplicate contexts?

cameronmeissner: I personally would lean towards keeping these test definitions as declarative as possible anyways, stuff like this probably shouldn't be here

cameronmeissner: curious about @r2k1's thoughts

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +723,77 @@ removeKubeletFlag() {
...
+
+    set +x 
+    access_url="http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/&client_id=$client_id"
+    raw_access_token=$(retrycmd_get_access_token_for_oras 10 5 $access_url)
+    ret_code=$? 
+    if [ $ret_code -ne 0 ]; then
+        echo "failed to retrieve access token: $ret_code"
+        return $ret_code
+    fi
+    if [ -z "$raw_access_token" ] || [[ "$raw_access_token" == *"error"* ]]; then
```
cameronmeissner, line 755: nit: seems slightly brittle to just check for `*error*`, are you sure this works in all cases?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -271,6 +296,21 @@ retrycmd_get_binary_from_registry_with_oras() {
         fi
     done
 }
+retrycmd_can_oras_discover_acr() {
+    retries=$1; wait_sleep=$2; url=$3
+    repo_name="${url%:*}"  # Removes ":latest" or any other tag
+    for i in $(seq 1 $retries); do
+        output=$(timeout 60 oras discover "$url" --registry-config "$ORAS_REGISTRY_CONFIG_FILE" 2>&1)
```
cameronmeissner, line 303: did you look into seeing if we could use `oras repo ls $acr` instead?

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner, line 755: don't think we need to check if $raw_access_token is empty again since if it's empty ret_code will be non-zero, unless I'm missing something

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -683,4 +723,77 @@ removeKubeletFlag() {
...
+        echo "failed to retrieve refresh token"
+        return $ERR_ORAS_PULL_UNAUTHORIZED
+    fi
+    REFRESH_TOKEN=$(echo "$raw_refresh_token" | jq -r .refresh_token)
+    if [ -z "$REFRESH_TOKEN" ] || [ "$REFRESH_TOKEN" == "null" ]; then
+        echo "failed to parse refresh token"
+        return $ERR_ORAS_PULL_UNAUTHORIZED
+    fi
+    oras login "$acr_url" --identity-token-stdin --registry-config "${ORAS_REGISTRY_CONFIG_FILE}" <<< "$REFRESH_TOKEN"
+    exit_code=$?
```
cameronmeissner, line 781: let's be consistent about ret_code/exit_code, either works for me

spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh
```diff
@@ -168,4 +168,147 @@ Describe 'cse_helpers.sh'
...
+                        exit 0
+                        ;;
+                esac
+                echo -1
+                exit 1
+            fi
+EOF
+
+            chmod +x "$MOCK_BIN_DIR_CURL/curl"
+            export PATH="$MOCK_BIN_DIR_CURL:$PATH"
```
cameronmeissner, line 242: I wanna move shellspec execution to a containerized environment in the future so we can control the shell, once that's the case then this should be fine, though for right now it seems like this might interfere with local environments

spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh
```diff
@@ -168,4 +168,147 @@ Describe 'cse_helpers.sh'
...
+        }
+
+        cleanup_mock_curl() {
+            rm -rf "$MOCK_BIN_DIR_CURL"
+            unset MOCK_BIN_DIR_CURL
```
cameronmeissner, line 247: we'd wanna be sure to remove this from $PATH as well

spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh
cameronmeissner: could we just use a standard mock instead?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -87,6 +87,10 @@ if [[ ${ID} != "mariner" ]] && [[ ${ID} != "azurelinux" ]]; then
     logs_to_events "AKS.CSE.removeManDbAutoUpdateFlagFile" removeManDbAutoUpdateFlagFile
 fi
 
+if [[ ! -z ${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER} ]]; then
```
cameronmeissner, line 90:
```suggestion
if [[ -n ${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER} ]]; then
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: Yeah, because we do actually get a response and it doesn't throw an error code, so we have to check for the result to say error

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: Is there a pro over oras discover? It just gives the sha of the repo if its discoverable or it throws auth error

e2e/kube.go
AlisonB319: The decision on which ACR to use comes from the isNonAnonymousPull tag set by the scenario. Are you saying the config can have a setter thats called after the scenario starts? If not it seems like it needs to be determined dynamically test by test, unless I'm missing something?

e2e/config/config.go
AlisonB319: you can't have - chars in the private acr name :(

e2e/config/config.go
AlisonB319: same as above, only letters and numbers allowed (and they have to be unqiue per subscription too)

e2e/kube.go
fseldow: my bad, bin's pr is not related to non-anonymous access

spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh
AlisonB319: Yeah you're right, doing it this was was left over from when I was calling the curl command directly in the function

e2e/scenario_test.go
AlisonB319: The reason I had to do it there was so that it was passed into nbc, but I can look into mutating the nbc within RunScenario. 
While I agree it is a bit clunky, I think there might be some usefulness to seeing the values set within the Test vs having to know to look for it in the RunScenario. I don't think it overwrites any existing contexts or anything

parts/linux/cloud-init/artifacts/cse_helpers.sh
AlisonB319: Here are the differences between the two 

```
$ oras repo ls privateace2enonanonpullwestus3.azurecr.io
Error response from registry: could not list repositories for "privateace2enonanonpullwestus3.azurecr.io"
GET "https://privateace2enonanonpullwestus3.azurecr.io/v2/_catalog": unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information. CorrelationId: 0b52339c-9eb3-49cd-b848-089cf280e567

$ oras repo ls privateacre2ewestus3.azurecr.io
aks/ip-masq-agent-v2
azuredefender/stable/low-level-collector
azuredefender/stable/old-file-cleaner
azuredefender/stable/pod-collector
azuredefender/stable/security-publisher
cbl-mariner/base/core
cbl-mariner/busybox
mcr/hello-world
oss/binaries/kubernetes/kubernetes-node
oss/kubernetes/azure-cloud-node-manager
oss/kubernetes/kube-proxy
oss/kubernetes/metrics-server
oss/kubernetes-csi/azuredisk-csi
oss/kubernetes-csi/azurefile-csi
oss/kubernetes-csi/csi-node-driver-registrar
oss/kubernetes-csi/livenessprobe

$ oras discover privateacre2ewestus3.azurecr.io/mcr/hello-world:latest
privateacre2ewestus3.azurecr.io/mcr/hello-world@sha256:92c7f9c92844bbbb5d0a101b22f7c2a7949e40f8ea90c8b3bc396879d95e899a

$ oras discover privateace2enonanonpullwestus3.azurecr.io/mcr/hello-world:latest
Error response from registry: HEAD "https://privateace2enonanonpullwestus3.azurecr.io/v2/mcr/hello-world/manifests/latest": unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information. CorrelationId: dac16032-8969-4391-ab16-d021848d66df
```

e2e/kube.go
AlisonB319: Also, eventually anonymous login will be unsupported and with it, I'll be able to clean up stuff like this too

e2e/aks_model.go
AlisonB319: the `clientset, err := kubernetes.NewForConfig(config)` creates a *kubernetes.Clientset which is unused in `* Kubeclient` so I think all of the resources I created here are necessary

e2e/kube.go
cameronmeissner: maybe add a method on config.Configuration, `GetPrivateACRName`, which takes in a `Scenario` object (or just tags...?) and then gives the correct ACR name based on the non-anon tag?

e2e/config/config.go
cameronmeissner: sadness

e2e/scenario_test.go
cameronmeissner: we can leave it as-is, at some point we should probably take a stance on where we want to perform operations like this in the future though

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: I'd personally prefer `oras repo ls` since we wouldn't have to rely on using a dummy image for discovery, and as you said we don't need to specify a registry path

as far as checking the output, something like this might work:

```
output=$(timeout 60 oras repo ls "$url" 2>&1)
if [ $? -eq 124 ]; then # 124 is how you should be able to reliably check if the command timed out or not
    echo "timed out when attempting to oras repo ls $url"
    return $EXIT_CODE_ORAS_REPO_LS_TIMEOUT
fi
if [[ "$output" == *"unauthorized: authentication required"* ]]; then
    echo "acr is not reachable: $output"
    return 1
fi
```

cameronmeissner: I'll leave it up to you since you've played around with this the most

cameronmeissner: adding `--debug` also gives some further info, include HTTP response codes for each of the individual requests it makes - so you could use that and just grep for `200 OK` for the good case - the annoying part is that in the anonymous case you'll see 401 Unauthorized show up even though it eventually succeeds

e.g.

unauthorized:
```
DEBU[0000] Request #0
> Request URL: "https://privateace2enonanonpullwestus3.azurecr.io/v2/_catalog"
> Request method: "GET"
> Request headers:
   "User-Agent": "oras/1.1.0+Homebrew" 
DEBU[0000] Response #0
< Response Status: "401 Unauthorized"
< Response headers:
   "Content-Type": "application/json; charset=utf-8"
   "Www-Authenticate": "Bearer realm=\"https://privateace2enonanonpullwestus3.azurecr.io/oauth2/token\",service=\"privateace2enonanonpullwestus3.azurecr.io\",scope=\"registry:catalog:*\""
   "X-Content-Type-Options": "nosniff"
   "Server": "AzureContainerRegistry"
   "Date": "Fri, 07 Feb 2025 16:50:50 GMT"
   "Connection": "keep-alive"
   "Access-Control-Expose-Headers": "Docker-Content-Digest, WWW-Authenticate, Link, X-Ms-Correlation-Request-Id"
   "Docker-Distribution-Api-Version": "registry/2.0"
   "Strict-Transport-Security": "max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains"
   "Content-Length": "196"
   "X-Ms-Correlation-Request-Id": "107bd5e4-e9b1-4524-9460-4eadcd6499c1" 
DEBU[0000] Request #1
> Request URL: "https://privateace2enonanonpullwestus3.azurecr.io/oauth2/token?scope=registry%3Acatalog%3A%2A&service=privateace2enonanonpullwestus3.azurecr.io"
> Request method: "GET"
> Request headers:
   "User-Agent": "oras/1.1.0+Homebrew" 
DEBU[0000] Response #1
< Response Status: "401 Unauthorized"
< Response headers:
   "X-Ms-Correlation-Request-Id": "5abc239c-a718-4465-9240-8c17b2e79c8c"
   "X-Ms-Ratelimit-Remaining-Calls-Per-Second": "333.316667"
   "Strict-Transport-Security": "max-age=31536000; includeSubDomains"
   "Server": "AzureContainerRegistry"
   "Date": "Fri, 07 Feb 2025 16:50:50 GMT"
   "Content-Type": "text/plain; charset=utf-8"
   "Content-Length": "186"
   "Connection": "keep-alive" 
Error: GET "https://privateace2enonanonpullwestus3.azurecr.io/v2/_catalog": GET "https://privateace2enonanonpullwestus3.azurecr.io/oauth2/token?scope=registry%3Acatalog%3A%2A&service=privateace2enonanonpullwestus3.azurecr.io": response status code 401: unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information. CorrelationId: 5abc239c-a718-4465-9240-8c17b2e79c8c
```

authorized:
```
DEBU[0000] Request #0
> Request URL: "https://privateacre2ewestus3.azurecr.io/v2/_catalog"
> Request method: "GET"
> Request headers:
   "User-Agent": "oras/1.1.0+Homebrew" 
DEBU[0000] Response #0
< Response Status: "401 Unauthorized"
< Response headers:
   "Content-Type": "application/json; charset=utf-8"
   "Content-Length": "196"
   "Www-Authenticate": "Bearer realm=\"https://privateacre2ewestus3.azurecr.io/oauth2/token\",service=\"privateacre2ewestus3.azurecr.io\",scope=\"registry:catalog:*\""
   "Connection": "keep-alive"
   "Strict-Transport-Security": "max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains"
   "X-Content-Type-Options": "nosniff"
   "X-Ms-Correlation-Request-Id": "c388dc26-d027-4f83-9e1a-397bfe703d37"
   "Server": "AzureContainerRegistry"
   "Date": "Fri, 07 Feb 2025 16:50:39 GMT"
   "Access-Control-Expose-Headers": "Docker-Content-Digest, WWW-Authenticate, Link, X-Ms-Correlation-Request-Id"
   "Docker-Distribution-Api-Version": "registry/2.0" 
DEBU[0000] Request #1
> Request URL: "https://privateacre2ewestus3.azurecr.io/oauth2/token?scope=registry%3Acatalog%3A%2A&service=privateacre2ewestus3.azurecr.io"
> Request method: "GET"
> Request headers:
   "User-Agent": "oras/1.1.0+Homebrew" 
DEBU[0000] Response #1
< Response Status: "200 OK"
< Response headers:
   "X-Ms-Correlation-Request-Id": "1cdb38e4-b5e6-4576-99fe-8e1aa0cab169"
   "X-Ms-Ratelimit-Remaining-Calls-Per-Second": "333.316667"
   "Strict-Transport-Security": "max-age=31536000; includeSubDomains"
   "Server": "AzureContainerRegistry"
   "Date": "Fri, 07 Feb 2025 16:50:39 GMT"
   "Content-Type": "application/json; charset=utf-8"
   "Connection": "keep-alive" 
DEBU[0000] Request #2
> Request URL: "https://privateacre2ewestus3.azurecr.io/v2/_catalog"
> Request method: "GET"
> Request headers:
   "Authorization": "*****"
   "User-Agent": "oras/1.1.0+Homebrew" 
DEBU[0000] Response #2
< Response Status: "200 OK"
< Response headers:
   "Access-Control-Expose-Headers": "Docker-Content-Digest, WWW-Authenticate, Link, X-Ms-Correlation-Request-Id"
   "X-Ms-Correlation-Request-Id": "844e8658-2fc3-4e4b-8744-c4619277993b"
   "Server": "AzureContainerRegistry"
   "X-Content-Type-Options": "nosniff"
   "Date": "Fri, 07 Feb 2025 16:50:39 GMT"
   "Content-Length": "564"
   "Connection": "keep-alive"
   "Docker-Distribution-Api-Version": "registry/2.0"
   "Content-Type": "application/json; charset=utf-8"
   "Strict-Transport-Security": "max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains" 
aks/ip-masq-agent-v2
```

cameronmeissner: you can use the exit code of the subshell to indicate success/failure

```
output=$(timeout 60 oras repo ls "$url" --debug 2>&1)
exit_code=$?
if [ $exit_code -eq 124 ]; then
    # timeout when egressing to ACR
    echo "timed out when attempting to oras repo ls $url"
    return $EXIT_CODE_ORAS_REPO_LS_TIMEOUT
fi
if [ $exit_code -eq 0 ]; then
    # 200 response from ACR
    echo "acr is reachable: $output"
    return 0
fi
# at this point, exit code indicates an unsuccessful request of some sort
if [[ "$output" == *"unauthorized: authentication required"* ]]; then
    # 401 response from ACR
    echo "acr is not reachable: $output"
    return $EXIT_CODE_ORAS_UNAUTHORIZED
fi
echo "unexpected response from acr:  $output"
return 1
```

Devinwong (approved):

PR #5666, opened 2025-01-27
add displayname to e2e pipeline

timmy-wright:
**What type of PR is this?**

/kind chore

Currently all the pipeline jobs that run e2e tests have the same name. Which is confusing when bouncing between pipelines. This PR adds more detail to the e2e pipeline name.

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -1270,5 +1270,6 @@ stages:
     variables:
       VHD_BUILD_ID: $(Build.BuildId)
       TAGS_TO_SKIP: "os=windows"
+      TAGS_TO_RUN: ""
     jobs:
       - template: ./templates/e2e-template.yaml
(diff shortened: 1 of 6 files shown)


.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1272,3 +1272,6 @@ stages:
       TAGS_TO_SKIP: "os=windows"
     jobs:
       - template: ./templates/e2e-template.yaml
+        parameters:
+          name: All Linux
```
r2k1, line 1276: not necessary all linux tests, only for VHD build in the pipeline. User picks the VHDs.

r2k1 (approved):
I don't think I had issues with naming. And it adds extra complexity to the pipeline.
But if it solves an annoying issue for you I'm ok with it.

.pipelines/.vsts-vhd-builder-release.yaml
timmy-wright: I think it'll still run all the tests tho - AFAIK there isn't a way to reduce the tests to just the built VHDs.

r2k1: It uses buildID env variable.

And run tests only for VHDs tagged with the build id.

PR #5668, opened 2025-01-28
bug: add env var CONTAINERD_CONFIG_NO_GPU_CONTENT back for scriptless

Devinwong:
**What type of PR is this?**

/kind bug

bug: add env var CONTAINERD_CONFIG_NO_GPU_CONTENT back for scriptless. Otherwise No_GPU e2e test will fail.
Tested pass in a new ab e2e test

Fixes #

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


aks-node-controller/parser/helper.go
```diff
@@ -139,26 +145,33 @@ func getKubenetTemplate() string {
 	return base64.StdEncoding.EncodeToString(kubenetTemplateContent)
 }
 
-func getContainerdConfig(aksnodeconfig *aksnodeconfigv1.Configuration) string {
+// getContainerdConfig returns the base64 encoded containerd config depending on whether the node is with GPU or not.
+func getContainerdConfig(aksnodeconfig *aksnodeconfigv1.Configuration, noGPU bool) string {
```
r2k1, line 149: a minor side note.

I wonde wherher it's better to perform all base64 conversions in the parser.go.
Just where we assign them to ENV variable.

To have see clearly which of the variables are encoded.

Alternativly, it's worth to add base64 suffix to such methods

aks-node-controller/parser/helper.go
```diff
@@ -48,6 +48,12 @@ var (
 	containerdConfigTemplate = template.Must(
 		template.New("containerdconfigforaksnodeconfig").Funcs(getFuncMapForContainerdConfigTemplate()).Parse(containerdConfigTemplateText),
 	)
+	//go:embed  templates/containerd_no_GPU.toml.gtpl
+	containerdConfigNoGPUTemplateText string
+	//nolint:gochecknoglobals
+	containerdConfigNoGPUTemplate = template.Must(
+		template.New("containerdconfigforaksnodeconfig").Funcs(getFuncMapForContainerdConfigTemplate()).Parse(containerdConfigNoGPUTemplateText),
```
r2k1, line 55: It's probably doesn't matter, but `containerdconfigforaksnodeconfig` template name is already used few lines above.

Maybe it's better to pass empty string instead?

aks-node-controller/parser/helper.go
r2k1: Another minor thing. I think it will be cleaner to have 2 methods instead of `noGPU bool`

r2k1 (approved):

aks-node-controller/parser/helper.go
Devinwong: Don't know if the names matter. Updated with 2 different and descriptive names.

aks-node-controller/parser/helper.go
Devinwong: Thanks for the suggestions. Updated.

PR #5669, opened 2025-01-28
chore: adjust apt-get regex looking for errors

UtheMan:


anujmaheshwari1 (approved):
gotta love us a good regex. @UtheMan can you screenshot the package list and have a super short description of why it failed, I think it would be amusing if anyone ever came to look for this.

PR #5671, opened 2025-01-28
Add e2etest ubuntu2204 gpunodriver scriptless

xmzhao0822:
**What type of PR is this?**

/kind test

Add e2etest ubuntu2204 gpunodriver scriptless

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong:
LGTM. Thanks for writing this test to catch a bug in scriptless.

Devinwong (approved):

e2e/scenario_test.go
```diff
@@ -957,6 +957,35 @@ func Test_Ubuntu2204_GPUNoDriver(t *testing.T) {
...
+			AKSNodeConfigMutator: func(config *aksnodeconfigv1.Configuration) {
+				config.VmSize = "Standard_NC6s_v3"
+				config.GpuConfig.ConfigGpuDriver = true
+				config.GpuConfig.GpuDevicePlugin = false
+				config.GpuConfig.EnableNvidia = to.Ptr(true)
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.Tags = map[string]*string{
+					// deliberately case mismatched to agentbaker logic to check case insensitivity
+					"SkipGPUDriverInstall": to.Ptr("true"),
```
r2k1, line 978: do use VMSS tags for provisioning logic?

xmzhao0822:
> LGTM. Thanks for writing this test to catch a bug in scriptless.

No problem :)

e2e/scenario_test.go
Devinwong: Other tests are not using. I think we don't?

r2k1: If it's not required, can you please remove it?

It feels like there is hidden logic behind it otherwise.

Devinwong: oh it's needed because the previous test `Test_Ubuntu2204_GPUNoDriver` also has it. And this tag was checked in cse_main.sh with this logic 
```
export -f should_skip_nvidia_drivers
skip_nvidia_driver_install=$(retrycmd_if_failure_no_stats 10 1 10 bash -cx should_skip_nvidia_drivers)
```
Otherwise the test would fail.

PR #5672, opened 2025-01-28
bug: reduce windows bootstrap command size

timmy-wright:
**What type of PR is this?**

/kind bug

There's a limitation to the windows cmd processor size that we're hitting in calico testing.

https://learn.microsoft.com/en-gb/troubleshoot/windows-client/shell-experience/command-line-string-limitation

Have tested this by running the aks-rp windows calico network test with a custom agentbaker and it passes. Whereas that test is currently failing in master.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
Refactor to move values from csecmd.ps1 to kuberneteswindowssetup.ps1 to avoid command line limitation.
```


pkg/agent/baker.go
```diff
@@ -31,17 +31,15 @@ func InitializeTemplateGenerator() *TemplateGenerator {
 // This function only can be called after the validation of the input NodeBootstrappingConfiguration.
 func (t *TemplateGenerator) getNodeBootstrappingPayload(config *datamodel.NodeBootstrappingConfiguration) string {
 	var customData string
+	// this might seem strange that we're encoding the custom data to a JSON string and then extracting it, but without that serialisation and deserialisation
+	// lots of tests fail.
 	if config.AgentPoolProfile.IsWindows() {
 		customData = getCustomDataFromJSON(t.getWindowsNodeCustomDataJSONObject(config))
+		return base64.StdEncoding.EncodeToString([]byte(customData))
 	} else {
```
cameronmeissner, line 39: nit: we can get rid of the else block now that we return in the windows case

timmy-wright: We can, but that structure of the code implies that windows code paths are less important. Maybe I'll flick the test around and give Linux the early exit instead of Windows :)

timmy-wright: Done.

parts/windows/csecmd.ps1
```diff
@@ -1,21 +1,14 @@
 powershell.exe -ExecutionPolicy Unrestricted -command \"
 $inputFile = '%SYSTEMDRIVE%\AzureData\CustomData.bin';
 $outputFile = '%SYSTEMDRIVE%\AzureData\CustomDataSetupScript.ps1';
-if (!(Test-Path $inputFile)) { throw 'ExitCode: |49|, Output: |WINDOWS_CSE_ERROR_NO_CUSTOM_DATA_BIN|, Error: |C:\AzureData\CustomData.bin does not exist.|' };
+if (!(Test-Path $inputFile)) { throw 'ExitCode: |49|, Output: |WINDOWS_CSE_ERROR_NO_CUSTOM_DATA_BIN|, Error: |C:\AzureData\CustomData.bin  does not exist.|' };
```
cameronmeissner, line 4: nit: looks like an extra space was added between `C:\AzureData\CustomData.bin` and  `does not exist.`, since we're trying to minimize size might be worth cleaning that up

cameronmeissner (approved):

parts/windows/csecmd.ps1
timmy-wright: Even better, I can replace it with "$inputFile" :)

PR #5712, opened 2025-01-30
nit: hide testvm password

jason1028kr:
**What type of PR is this?**
/kind chore

Not sure if it's even necessary. But hide testvm password getting traced

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/test/run-test.sh
```diff
@@ -110,8 +111,7 @@ else
       --public-ip-address "" \
       ${TARGET_COMMAND_STRING}
 fi
-
-echo "VHD test VM username: $TEST_VM_ADMIN_USERNAME, password: $TEST_VM_ADMIN_PASSWORD"
```
cameronmeissner, line 114: I can't remember, isn't there a debug mode where we let the user keep the VM for debugging? if so then we want to upload an ssh key or something to build artifacts

though, I think at the end of the day devs can just reset the password on the VM manually

cameronmeissner (approved):

PR #5730, opened 2025-02-03
refactor config for cri plugin to relocate CRI properties

bravebeaver:
**What type of PR is this?**
/kind feature

CRI plugin in containerd2.0 has few deprecated / renamed features. this is to clean up plugin properties so that no error message is shown when run

```
containerd plugin ls
```

Fixes #31163060

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

actual migration happens [here](https://github.com/containerd/containerd/blob/2633eb581c380276274f81ba78b4fbceedce0734/plugins/cri/images/plugin.go#L173)

however when run `containerd plugin` it is not picking up the new plugin, and showing instead

```
WARN[0000] Ignoring unknown key in TOML for plugin       error="strict mode: fields in the document are missing in the target struct" key=sandbox_image plugin=io.containerd.grpc.v1.cri
``` 

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -1366,15 +1387,236 @@ root = "{{GetDataDir}}"{{- end}}
...
-
-// this pains me, but to make it respect mutability of vmss tags,
-// we cannot use go templates at runtime.
-// CSE needs to be able to generate the full config, with all params,
-// with the tags pulled from wireserver. this is a hack to avoid
-// moving all the go templates to CSE -- we allow two options,
-// duplicate them in CSE base64-encoded, and pick the right one.
-// they're identical except for GPU runtime class.
-const containerdConfigNoGpuTemplateString = `version = 2
+	containerdV2ConfigTemplate ContainerdConfigTemplate = `version = 2
```
Devinwong, line 1390: Not necessarily needs to be in this PR. But in Scriptless we also need the containerd configs. Please refer to `aks-node-controller/parser/templates/containerd_no_GPU.toml.gtpl` and `aks-node-controller/parser/templates/containerd.toml.gtpl`. We simply update the funcMap functions. For example, from `HasDataDir` to `getHasDataDir .KubeletConfig`. As you know there are already some e2e tests in scenario_test.go for 2404. Please add appropriate ones for containerdv2 for scriptless too if there isn't any yet. Again, this will not block this PR.

Devinwong: We can work together for the detailed changes if needed. :)

bravebeaver: i have seen `controller` package and was wondering what should i do there 🙏.  let's improve this

Devinwong (approved):

PR #5741, opened 2025-02-04
add fallback to multi arch versions v2 for windows

timmy-wright:
**What type of PR is this?**

/chore 

Adds a fallback for windows to multi arch v2 so renovate can work for some windows containers.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -70,7 +70,8 @@
           "latestVersion": "1.8.22",
           "previousLatestVersion": "1.8.20"
         }
-      ]
+      ],
+      "windowsVersions": []
```
cameronmeissner, line 74: do we need to always explicitly declare an empty set of windows versions?

timmy-wright: Yes, because otherwise we fallback to the "multiArch" versions and the windows VHDs blow up.

vhdbuilder/packer/windows/components_json_helpers.tests.ps1
```diff
@@ -353,6 +353,12 @@ Describe 'Tests of components.json' {
         $components | Should -Contain "mcr.microsoft.com/oss/kubernetes/pause:3.9"
     }
 
+    it 'has the right version of ciprod' {
+        $components = GetComponentsFromComponentsJson $componentsJson
+
+        $components | Should -Contain "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:win-3.1.24"
```
Devinwong, line 359: Just an FYI. I added a test_components.json for my unit test to avoid coming back to change some values that could be updated in the future. For example, in this case, `ciprod:win-3.1.24`.

Devinwong (approved):

vhdbuilder/packer/windows/components_json_helpers.tests.ps1
timmy-wright: Yup - I contemplated that, but then wanted to have comprehensive tests for windows at first to make sure I had the right versions in the file. Will likely remove these over time. I also use a sample components.json for other tests.

PR #5742, opened 2025-02-04
test(e2e): simplify cluster parameter extraction

cameronmeissner:
**What type of PR is this?**

right now we extract all cluster params except for the bootstrap token using a privileged pod running a node of the e2e cluster's system pool - this is an error prone process and requires at least one node to be running in the system pool at the time of test execution. it turns out we can get the cluster CA certificate data and cluster fqdn directly from the `ListClusterAdminCredentials` API on the AKS ARM client instead

further, this PR removes the API server cert and client key cluster parameters, as they aren't strictly needed with how we bootstrap nodes for e2e tests

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -17,7 +18,11 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v6"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cameronmeissner:
linux e2es failing due to unrelated issues
windows e2e all green

e2e/cluster.go
```diff
@@ -127,15 +138,57 @@ func prepareCluster(ctx context.Context, t *testing.T, cluster *armcontainerserv
...
 		Kube:          kube,
 		SubnetID:      subnetID,
 		Maintenance:   maintenance,
-		ClusterParams: extractClusterParameters(ctx, t, kube),
+		ClusterParams: clusterParams,
 	}, nil
 }
 
+func extractClusterParameters(ctx context.Context, t *testing.T, kube *Kubeclient, cluster *armcontainerservice.ManagedCluster) (*ClusterParams, error) {
+	resp, err := config.Azure.AKS.ListClusterAdminCredentials(ctx, config.ResourceGroupName, *cluster.Name, nil)
```
lilypan26, line 156: nit: idk when it would happen, but should we include a nil check for `cluster.Name`

lilypan26 (approved):

e2e/cluster.go
cameronmeissner: the cluster object comes from a call to AKS.Get/CreateOrUpdate, so it shouldn't ever be the case that we make it this far in test execution and Name is nil without an error having previously occurring

r2k1: the result will be the same whether you check it for nil or not -  a test will fail

r2k1: A kube client is already passed to this function.
Does it mean we already loaded credentials before? Are we doing it second time?

The change makes sense, just wonder if we can optimize it a one step further

r2k1 (approved):

e2e/cluster.go
cameronmeissner: yeah good point, and unfortunately I saw this after I pressed the merge button :D

PR #5749, opened 2025-02-05
chore: update windows cse script version

santhoshmprabhu:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: We updated the CSE scripts  for windows in #5482, and created a new zip file with the changes. This PR updates the version of the zip package to the newest.

We had some issues running e2e tests so enhanced the logging from them to show more status information as the node joins the API server and becomes (or not) ready.

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
@timmy-wright @AbelHu

timmy-wright (approved):
Looks good to me. I tried to write an e2e test to check the functionality but had issues creating a node with the right configuration. Either disable that test and we can work on it next week, or try to fix it as I'm on leave for a couple of days.

santhoshmprabhu:
Had a chat with @timmy-wright - The reason we see the test fail is that nodes aren't getting to the ready state when added to existing clusters. Now, we need specific types of clusters that support cilium nodesubnet, to use EbpfDataplane=true in a windows nodepool, and further require overriding some toggles too. So it makes sense to add an E2E once we have a cluster capable of supporting Cilium Nodesubnet.

PR #5802, opened 2025-02-11
Add pipeline var to skip extension check

timmy-wright:
**What type of PR is this?**

/kind chore

This PR lets us skip a particular check when building test VHDs. For prod (release) VHDs, we don't want any VM extensions on the VHD. But for test, it's OK. Not idea, and we should get policy dispensation for the test subscription, but we don't want that to block us.

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong (approved):
lgtm

r2k1 (approved):

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -88,6 +88,7 @@ steps:
       -e BUILD_NUMBER=$(Build.BuildNumber) \
       -e WINDOWS_SKU=$(WINDOWS_SKU) \
       -e OS_TYPE="Windows" \
+      -e SKIP_EXTENSION_CHECK=$(SKIP_EXTENSION_CHECK) \
```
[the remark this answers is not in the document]
timmy-wright: Bloody good question. Given everything else is explicitly set, I assume we have to set it as well.

vhdbuilder/packer/windows/windows-vhd-builder-sig.json
```diff
@@ -125,7 +125,8 @@
             "elevated_password": "{{.WinRMPassword}}",
             "environment_vars": [
                 "ProvisioningPhase=3",
-                "WindowsSKU={{user `windows_sku`}}"
+                "WindowsSKU={{user `windows_sku`}}",
+                "SKIP_EXTENSION_CHECK={{ user `SKIP_EXTENSION_CHECK` }}"
```
smith1511, line 129: It would be nice to avoid all the stupid variables having to be explicitly being passed through - if we used a well-known env var prefix maybe we could automate it in the pipeline and packager?

smith1511 (approved):
Lookgs good other than my comment/question

vhdbuilder/packer/windows/windows-vhd-builder-sig.json
timmy-wright: It is a good idea!

PR #5803, opened 2025-02-11
fix: allow config "v1"

r2k1:
**What type of PR is this?**

/kind bug

Forcing consumers to specify "v0" config version were a mistake.
While we're in development breaking changes are expected. But we're not going to have different logic maintaining both v0 and v1. So there is no point of forcing consumers to specify v0 and update it to v1 before the release.


timmy-wright (approved):
LGTM.

Devinwong (approved):
lgtm. Yihao is the only consumer for now. So we should be good. Thanks for catching this.

PR #5804, opened 2025-02-11
feat: update kata config for containerd v2

bravebeaver:
**What type of PR is this?**

/kind feat

update containerd config for kata

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -1516,43 +1536,70 @@ root = "{{GetDataDir}}"{{- end}}
   snapshotter = "overlaybd"
   disable_snapshot_annotations = false
 {{- end}}
+{{- if IsKata }}
```
timmy-wright, line 1539: It'd be good to see an e2e test to check that Kata is setup properly for containerd2.

Devinwong (approved):

pkg/agent/baker.go
bravebeaver: agreed! however i could not find existing e2e test for kata and was wondering maybe @miz060  can help?

PR #5813, opened 2025-02-12
chore: more windows renovate changes

timmy-wright:
**What type of PR is this?**

/kind chore

Add more windows containers to renovate.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -246,8 +246,8 @@
       "matchPackageNames": [
         "windows/nanoserver",
         "windows/servercore",
-        "containernetworking/azure-npm",
-        "windows/servercore/iis"
+        "windows/servercore/iis",
+        "oss/kubernetes/windows-gmsa-webhook"
       ],
       "groupName": "windowsbase",
       "assignees": [
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -369,7 +369,7 @@
       ],
       "windowsVersions": [
         {
-          "renovateTag": "<DO_NOT_UPDATE>",
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/azure/secrets-store/provider-azure",
           "latestVersion": "v1.5.3",
           "previousLatestVersion": "v1.6.2"
         }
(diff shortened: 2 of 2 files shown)


parts/linux/cloud-init/artifacts/components.json
```diff
@@ -459,27 +459,27 @@
       ],
       "windowsVersions": [
         {
-          "renovateTag": "<DO_NOT_UPDATE>",
```
mainred, line 462: Hey @timmy-wright , sinc cloud-node-manager windows host process is still rolling out, and monitoring in regions, we need both version with and without windows-hpc for now. we can remove windows-hpc images after windows hpc images is in all regions. This may bring exception to the component.json. Thanks for you understanding.

How should I understand latestVersion and previousLatestVersion?

.github/renovate.json
```diff
@@ -247,7 +247,10 @@
         "windows/nanoserver",
         "windows/servercore",
         "containernetworking/azure-npm",
-        "windows/servercore/iis"
+        "windows/servercore/iis",
+        "oss/kubernetes/windows-gmsa-webhook",
+        "oss/kubernetes-csi/livenessprobe",
+        "oss/kubernetes/pause"
```
djsly, line 253: This one we already manage for Linux. Are we sure it won't run into a conflict ?

parts/linux/cloud-init/artifacts/components.json
```diff
@@ -369,7 +369,7 @@
...
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/azure/secrets-store/provider-azure",
           "latestVersion": "v1.5.3",
```
djsly, line 373: Those versions are inversed.  

Also I think we need to go back to the drawing board. We are duplicating too many entries for windows and Linux now. I don't think we should copy paste all those settings "renovateTag, previousTag, latest tag".  Those should be identical between Linux and windows (use the same properties. Instead of duplicating )

parts/linux/cloud-init/artifacts/components.json
AlisonB319: https://github.com/Azure/AgentBaker/blob/73683611115920e6f556744d81122c8eec19ccd0/.github/README-RENOVATE.md#table-of-contents

Here is the renovate wiki, it should explain the lastestVersion and previousLatest :)

timmy-wright: This PR doesn't change the versions we'll install on the VHD. It enables renovate which will automatically generate PRs when there are new versions available.

parts/linux/cloud-init/artifacts/components.json
timmy-wright: We have the capability to use the same version blob for Windows and Linux, and do this in some places. I plan to go through components.json and update when they are the same. Because windows containers are so much larger than linux containers there are many places where we don't want them to be the same - we want fewer on windows. There are also cases where particular versions (of cni/cns containers in particular) don't work on Windows, so we use an older version.

.github/renovate.json
timmy-wright: Done.

.github/renovate.json
```diff
@@ -247,7 +247,9 @@
         "windows/nanoserver",
         "windows/servercore",
         "containernetworking/azure-npm",
```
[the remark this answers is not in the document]
timmy-wright: Have removed them. Good catch.

Devinwong (approved):

PR #5841, opened 2025-02-13
chore: update ciprod to 3.1.25, exclude it from windows 23H2 SKU

timmy-wright:
**What type of PR is this?**
 
/kind chore

Updates ciprod container to 3.1.25. This made the windows 23H2 VHD become too large and fail the build. After much discussion, we realised that ciprod isn't supported on windows 23H2, so is not needed there.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -383,16 +383,18 @@
       "multiArchVersionsV2": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=azuremonitor/containerinsights/ciprod",
-          "latestVersion": "3.1.24"
+          "latestVersion": "3.1.25"
         }
       ],
       "windowsVersions": [
         {
+          "comment": "ciprod isn't supported on Win23H2 so is excluded as it makes the VHD too large",
           "renovateTag": "registry=https://mcr.microsoft.com, name=azuremonitor/containerinsights/ciprod",
           "latestVersion": "3.1.25",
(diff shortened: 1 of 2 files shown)


parts/common/components.json
```diff
@@ -383,7 +383,21 @@
       "multiArchVersionsV2": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=azuremonitor/containerinsights/ciprod",
-          "latestVersion": "3.1.24"
+          "latestVersion": "3.1.25"
+        }
+      ],
+      "windowsVersions": [
+        {
+          "comment": "ciprod isn't supported on Win23H2 so is excluded as it makes the VHD too large",
```
AbelHu, line 391: all ws2022 images should be able to run in 23H2. Does this mean that container insights will not support WindowsAnnualChannel?

timmy-wright: I was talking to the container insights team and they do not support windows annual.

The main problem with ciprod 3.1.25 is it's almost 1GB larger than 3.1.24 and wouldn't fit on the 23H2 image. So we agreed to remove ciprod from 23H2 so they can upgrade ciprod on the 2 windows versions they do support.

@wanlonghenry for visibility.

r2k1 (approved):

PR #5857, opened 2025-02-16
test: E2E Tweaks

r2k1:
**What type of PR is this?**
/kind test

Minor e2e tweaks and fixes.

### DIFF
--- a/e2e/config/azure.go
+++ b/e2e/config/azure.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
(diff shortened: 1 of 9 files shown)


e2e/config/azure.go
```diff
@@ -491,3 +492,58 @@ func ensureProvisioningState(version *armcompute.GalleryImageVersion) error {
...
+			t.Logf("created VMSS %s in resource group %s", vmssName, resourceGroupName)
+			return vmss, nil
+		}
+
+		// not a retryable error
+		if !retryOn(err) {
+			return nil, err
+		}
+
+		if attempt >= 10 {
```
timmy-wright, line 518: Why 10 retries?

e2e/config/azure.go
```diff
@@ -491,3 +492,58 @@ func ensureProvisioningState(version *armcompute.GalleryImageVersion) error {
 	}
 	return nil
 }
+
+func (a *AzureClient) CreateVMSSWithRetry(ctx context.Context, t *testing.T, resourceGroupName string, vmssName string, parameters armcompute.VirtualMachineScaleSet) (*armcompute.VirtualMachineScaleSet, error) {
+	t.Logf("creating VMSS %s in resource group %s", vmssName, resourceGroupName)
+	delay := 5 * time.Second
```
timmy-wright, line 498: Why 5 seconds? If we hit a quota issue, then 5 seconds likely isn't going to be enough. Actually, why are we retrying at all? It'd be good to have a comment about which types of failures the retry is trying to resolve - that'll help the next developer.

e2e/exec.go
```diff
@@ -172,15 +171,37 @@ func unprivilegedCommandArray() []string {
...
 	result += fmt.Sprintf("az account set --subscription %s\n", config.Config.SubscriptionID)
 	result += fmt.Sprintf("az aks get-credentials --resource-group %s --name %s --overwrite-existing\n", config.ResourceGroupName, *s.Runtime.Cluster.Model.Name)
-	result += fmt.Sprintf(`kubectl exec -it %s -- bash -c "chroot /proc/1/root /bin/bash -c '%s'"`, s.Runtime.DebugHostPod, sshString(s.Runtime.VMPrivateIP))
+	result += fmt.Sprintf(`kubectl exec -it %s -- bash -c "chroot /proc/1/root /bin/bash -c '%s'"`, s.Runtime.Cluster.DebugPod.Name, sshString(s.Runtime.VMPrivateIP))
 	s.T.Log(result)
 	//runtime.Breakpoint() // uncomment to pause the test
 }
+
+func uploadSSHKey(ctx context.Context, kube *Kubeclient, vmPrivateIP, jumpboxPodName string, sshPrivateKey []byte) error {
+	cmd := fmt.Sprintf(`echo '%s' > %[2]s && chmod 0600 %[2]s`, sshPrivateKey, sshKeyName(vmPrivateIP), sshString(vmPrivateIP))
```
timmy-wright, line 198: This probably won't work on Windows.

r2k1: it's extracted from `execScriptOnVm`

e2e/scenario_helpers_test.go
```diff
@@ -39,12 +39,12 @@ func setupSignalHandler() context.Context {
...
 		// block until signal is received
 		<-ch
-		fmt.Println(red("Received cancellation signal, gracefully shutting down the test suite. Cancel again to force exit."))
+		fmt.Println(red("Received cancellation signal, gracefully shutting down the test suite. Cancel again to force exit. (Created Azure resources will not be deleted in this case)"))
 		cancel()
 
 		// block until second signal is received
 		<-ch
-		fmt.Println(red("Received second cancellation signal, forcing exit."))
+		fmt.Println(red("Received second cancellation signal, forcing exit. (You may need to manually delete created Azure resources)"))
```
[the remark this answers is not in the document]
r2k1: hard to do from here, here is no test context

e2e/config/azure.go
r2k1: it's a few lines below. Some ARM-specific issues on new GPUs with low availability.

I suspect it's something to do with the VM being recycled. And I retry tests faster than it's cleaned

e2e/config/azure.go
r2k1: no reason, just something helped me with GPU tests

e2e/exec.go
r2k1: going to revert it, I've written before your changes.

Will add later once I encounter the issue again

timmy-wright (approved):

PR #5862, opened 2025-02-17
chore: fix var substitution in components.json for windows and add CPU_ARCH substutition

timmy-wright:
**What type of PR is this?**
/kind chore

Currently windows components and linux components use different variable substitution mechanisms in components.json, which makes it hard to share urls and download sources. This PR makes Windows consistent with Linux and also sets CPU_ARCH for Windows so it can be used in urls.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -820,7 +820,7 @@
                 "previousLatestVersion": "0.0.50"
               }
             ],
-            "downloadURL": "https://acs-mirror.azureedge.net/aks/windows/cse/aks-windows-cse-scripts-v[version].zip"
+            "downloadURL": "https://acs-mirror.azureedge.net/aks/windows/cse/aks-windows-cse-scripts-v${version}.zip"
           }
         }
       }
@@ -837,7 +837,7 @@
                 "latestVersion": "1.1.5"
               }
             ],
-            "downloadURL": "https://acs-mirror.azureedge.net/ccgakvplugin/v1.1.5/binaries/windows-gmsa-ccgakvplugin-v[version].zip"
+            "downloadURL": "https://acs-mirror.azureedge.net/ccgakvplugin/v1.1.5/binaries/windows-gmsa-ccgakvplugin-v${version}.zip"
           }
         }
       }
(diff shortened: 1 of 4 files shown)


r2k1 (approved):

bravebeaver:
LGTM.. 

just for my own curiosity, i noticed the `${}` syntax which is consistent with bash syntax, which i totally agree. but other than that do you have any other reasons to use it in ps1 syntax?

timmy-wright:
> LGTM..
> 
> just for my own curiosity, i noticed the `${}` syntax which is consistent with bash syntax, which i totally agree. but other than that do you have any other reasons to use it in ps1 syntax?

It's also ps1 syntax - I'm actually using the powershell interperter to do the substitution.

vhdbuilder/packer/windows/components_json_helpers.tests.ps1
```diff
@@ -522,6 +552,36 @@ Describe 'Gets The Versions' {
...
+                latestVersion = "1.8.22"
+            }
+        )
+        $componentsJson.ContainerImages[0].downloadURL = "mcr.microsoft.com/oss/kubernetes/autoscaler/`${varvarvar}/addon-resizer:*"
+
+        $varvarvar="x86"
+        $components = GetComponentsFromComponentsJson $componentsJson
+
+        $components | Should -HaveCount 1
+        $components | Should -Contain "mcr.microsoft.com/oss/kubernetes/autoscaler//addon-resizer:1.8.22"
```
mainred, line 582: is the double backslashes wrongly introduced by the test or the cache url?

parts/common/components.json
```diff
@@ -820,7 +820,7 @@
                 "previousLatestVersion": "0.0.50"
               }
             ],
-            "downloadURL": "https://acs-mirror.azureedge.net/aks/windows/cse/aks-windows-cse-scripts-v[version].zip"
+            "downloadURL": "https://acs-mirror.azureedge.net/aks/windows/cse/aks-windows-cse-scripts-v${version}.zip"
```
mainred, line 823: Make it os agnostic is challenging. Thanks for introducing this change.

PR #5873, opened 2025-02-19
chore: bump cni and cns to v1.6.21 for Windows VHD

QxBytes:
**What type of PR is this?**
/kind feature

Release notes: https://github.com/Azure/azure-container-networking/releases/tag/v1.6.21 
CNS: https://msazure.visualstudio.com/AzureWiki/_wiki/wikis/AzureWiki.wiki/754746/v1.6.21 

Fixes a bug when using vnet scale mode with cilium that would create incorrect iptables rules, causing connectivity to dns and imds to fail from the pod.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

Could someone confirm `testdata/prefetch.sh` does not need to be updated for caching new images?
Noticed https://github.com/Azure/AgentBaker/actions/runs/13402631728/job/37436508667?pr=5873 failing-- should these values always be present, or can the test be modified? This looks new/different from previous version bumps in agentbaker.

**Release note**:

https://github.com/Azure/azure-container-networking/releases/tag/v1.6.21

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -919,7 +919,8 @@
               },
               {
                 "renovateTag": "<DO_NOT_UPDATE>",
-                "latestVersion": "1.6.18"
+                "latestVersion": "1.6.21",
+                "previousLatestVersion": "1.6.18"
               }
             ],
             "downloadURL": "https://acs-mirror.azureedge.net/azure-cni/v${version}/binaries/azure-vnet-cni-windows-amd64-v${version}.zip"
@@ -1191,8 +1192,8 @@
               },
               {
                 "renovateTag": "<DO_NOT_UPDATE>",
-                "latestVersion": "1.6.18",
-                "previousLatestVersion": "1.6.13"
+                "latestVersion": "1.6.21",
+                "previousLatestVersion": "1.6.18"
               }
             ],
             "downloadURL": "https://acs-mirror.azureedge.net/azure-cni/v${version}/binaries/azure-vnet-cni-linux-${CPU_ARCH}-v${version}.tgz"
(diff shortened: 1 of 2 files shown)


rbtr (approved):

parts/common/components.json
```diff
@@ -910,8 +910,8 @@
             "versionsV2": [
               {
                 "renovateTag": "<DO_NOT_UPDATE>",
-                "latestVersion": "1.6.18",
-                "previousLatestVersion": "1.5.38"
+                "latestVersion": "1.6.21",
```
timmy-wright, line 913: It looks like we're going from a 1.6 and a 1.5 to two 1.6.x versions - is this correct or do we want a 1.5.x version on the VHD as well as a 1.6.x version?

rbtr: oooh...something is already wrong, we should have stayed within minor versions.

QxBytes: I'm wondering how the previous version could have been 1.5.38 in the first place-- usually we keep the two latest 1.5.x versions and two latest 1.6.x versions since there is both an active 1.5 version in aks and a 1.6 version in aks at all times (based on what k8 version you create your cluster with)

rbtr: when the VHD gets full, we may drop teh previous versions...this is either a bad merge or someone saw a 1.5 and a 1.6 and thought the 1.5 was the previous version for the 1.6 and collapsed them

timmy-wright (approved):
LG from Windows perspective.

parts/common/components.json
rbtr: > For windows we typically keep one patch version rather than two because we regularly run out of VHD space - the containers can be huge.

I've migrated our latest versions to the Windows HPC base image, so we should contribute less to this 🙂 1.6+ so once AKS 1.27 LTS ends in July we'll be done with the old fat images for CNS, CNI

PR #5875, opened 2025-02-19
Install AMD GPU Kernel drivers if required

r2k1:
**What type of PR is this?**

/kind feature

Add support for AMD GPUs. Install AMD GPU Kernel drivers if user didn't specify `--skip-gpu-driver` install.

AMD Operates differently to NVidia. A kernel driver is required on the host VM. Additionally containers should have ROCm suite installed (which is blowing container size images to 60GB+, but it's on AMD).

Cache required .deb packages on the VM, but doesn't install until required (adds about 30MB overhead).
Once installed it takes about 600-1000MB of additional space.

We don't have a reliable quota for the VM SKUs in the test subscription. And VM allocation isn't reliable either.
So automatic testing is disabled until we manage to sort it out somehow.


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -619,3 +619,40 @@ rm -f ./azcopy # cleanup immediately after usage will return in two downloads
...
+  sudo apt-get update
+
+  # Download to /var/cache/apt/archives/
+  sudo mkdir -p /var/cache/amdgpu-apt/
+  sudo chmod 777 /var/cache/amdgpu-apt/
+  # Download all dependencies of the amdgpu-dkms package
+  # The --reinstall flag is used to ensure that the package is downloaded even if it is already installed
+  # Otherwise installation of some packages like "m4" is skipped because it is already installed
+  # "m4" seems to be deleted at the later stage, making the installation fail
+  sudo apt-get install -o Dir::Cache::Archives="/var/cache/amdgpu-apt" --download-only --reinstall -y m4 amdgpu-dkms autoconf automake autotools-dev amdgpu-dkms-firmware
```
djsly, line 655: autoconf automake autotools-dev ... really ?  can't we avoid having those in the VHD ?    We deffinitly need to remove that after proviosnning the VM

djsly: ah it's only downloeaded to `/var/cache/amdgpu-apt` and removed lateer on

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -206,6 +206,16 @@ EOF
     fi
 fi
 
+if [[ "${AMD_GPU_NODE}" = true ]] && [[ "${skip_gpu_driver_install}" != "true" ]]; then
+    logs_to_events "AKS.CSE.ensureAMDGPUDrivers" ensureAMDGPUDrivers
+else
```
djsly, line 211: should we have those in a reusable function ? I would expect that we run a cleanup logic after we install the GPU drivers

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -844,6 +844,22 @@ ensureGPUDrivers() {
...
+# TODO: this is a temporary ubuntu-only HACK until we get a driver
+ensureAMDGPUDrivers() {
+    echo "Installing AMD GPU drivers"
+
+     # delete amdgpu module from blacklist
+    sudo sed -i '/blacklist amdgpu/d' /etc/modprobe.d/blacklist-radeon-instinct.conf
+
+    pushd /var/cache/amdgpu-apt
+    ls -l
+    sudo dpkg -i *.deb
```
djsly, line 856: why do we need `autoconf automake autotools-dev` those three packages are purely to compile the drivers from scratch. are we doing that ?  

We shouldnt need to keep those... (not even sure why we need them)

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -844,6 +844,22 @@ ensureGPUDrivers() {
...
+    popd
+
+    REBOOTREQUIRED=true
```
djsly, line 859: this is causing us customer headache, we need to ensure the node doesn't get to a ready state before the reboot.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -844,6 +844,22 @@ ensureGPUDrivers() {
     fi
 }
 
+# TODO: this is a temporary ubuntu-only HACK until we get a driver
+ensureAMDGPUDrivers() {
+    echo "Installing AMD GPU drivers"
+
+     # delete amdgpu module from blacklist
+    sudo sed -i '/blacklist amdgpu/d' /etc/modprobe.d/blacklist-radeon-instinct.conf
```
djsly, line 852: can we get a better explanation why we have that ? why understanding was this was required on a VM to enable the access of the physical hardware on the bareVM

e2e/kube.go
```diff
@@ -637,3 +632,52 @@ func nvidiaDevicePluginDaemonSet() *appsv1.DaemonSet {
...
+		},
+		Spec: corev1.PodSpec{
+			PriorityClassName: "system-node-critical",
+			NodeSelector: map[string]string{
+				"kubernetes.io/hostname": s.Runtime.KubeNodeName,
+			},
+			Containers: []corev1.Container{
+				{
+					Name:  "amdgpu-device-plugin-container",
+					Image: "rocm/k8s-device-plugin",
```
djsly, line 650: should pull from MCR, dockerhub pull isn't stable. (assuming this is from docker.io)

e2e/scenario_test.go
```diff
@@ -1664,3 +1664,69 @@ func Test_Ubuntu2404ARM(t *testing.T) {
...
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				ValidateAMDGPU(ctx, s)
+			},
+		},
+	})
+}
+
+func Test_Ubuntu2204Gen2Containerd_AMDGPU_V710(t *testing.T) {
+	// the SKU isn't available in subscriptrion/region we run tests
```
djsly, line 1701: can we add a TODO

e2e/scenario_test.go
```diff
@@ -1664,3 +1664,69 @@ func Test_Ubuntu2404ARM(t *testing.T) {
 		},
 	})
 }
+
+func Test_Ubuntu2404Gen2Containerd_AMDGPU_MI300(t *testing.T) {
+	t.Skip("Provisioning of Standard_ND96isr_MI300X_v5 isn't reliable yet")
```
djsly, line 1669: can we add a TODO

e2e/config/azure.go
```diff
@@ -291,25 +291,27 @@ func (a *AzureClient) UploadAndGetSignedLink(ctx context.Context, blobName strin
...
-	err = a.createBlobStorageContainer(ctx)
-	if err != nil {
-		return "", err
-	}
-
-	if err := a.assignRolesToVMIdentity(ctx, identity.Properties.PrincipalID); err != nil {
-		return "", err
-	}
-	return *identity.Properties.ClientID, nil
+	// HACK: temporary disable to allow running test in different subscription, without enough permissions
```
djsly, line 294: should we uncomment all this ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -619,3 +619,40 @@ rm -f ./azcopy # cleanup immediately after usage will return in two downloads
 echo "install-dependencies step completed successfully"
 capture_benchmark "${SCRIPT_NAME}_overall" true
 process_benchmarks
+
+
+download_amdgpu_drivers() {
```
djsly, line 624: do we have a device-plugin to install as well to ensure the GPU count shows up on the NOde labels ?

e2e/config/azure.go
r2k1: I'll remove it before merging the PR. Once I stop testing it.

Currently it's required only for a single scriptless test that is disabled by default.
Not much harm if it accidentially leak through.

e2e/kube.go
r2k1: I'll need to add it to MCR at some stage. I don't think it's available there yet.

parts/linux/cloud-init/artifacts/cse_config.sh
r2k1: I'd like to know as well why we need it. It took me a lot of time to figure out why drivers didn't work.

I think in the image provided by Canonical it's blacklisted, but nothing explains why it's there.

parts/linux/cloud-init/artifacts/cse_config.sh
r2k1: It's dependency of amdgpu. So I'm not sure what can I do with it.

I can run another set of tests without it (the feedback loop is frutrasting).

parts/linux/cloud-init/artifacts/cse_config.sh
r2k1: I though setting the variable automaticall does it for me. Any guidance on what needs to change?

parts/linux/cloud-init/artifacts/cse_main.sh
r2k1: Here I deleted AMD-related logic for non-AMD users.

Probably can cleanup for everyone.

vhdbuilder/packer/install-dependencies.sh
r2k1: It exists, but probably not in MCR. 

AFAIK we doesn't do it yet for Nvidia. Should be consistent for both of them? 

We probably can addess it separatelly.

e2e/kube.go
djsly: there are SFI to looks for that, we lost the contract preventing us from getting throttled on docker.io so expect throttling. 

We have an ACR for E2E, we should setup a remote/cache for this image and use our own ACR

parts/linux/cloud-init/artifacts/cse_config.sh
djsly: @ganeshkumarashok what are we planning to do for the MIG use case ?

PR #5887, opened 2025-02-20
add pipeline var to skip openssh server installation

timmy-wright:
**What type of PR is this?**

/kind chore

* Pass a pipeline parameter to the script to skip installing open ssh server - this is needed for the 1P project
* Fix the value of SKIP_EXTENSION_CHECK - as per screen snapshot

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


zylxjtu:
Do we have plans to add the options of enable/disable the openssh check in CSE?

cameronmeissner (approved):

timmy-wright:
> Do we have plans to add the options of enable/disable the openssh check in CSE?

@zylxjtu: None that I'm aware of. This pipeline change is just because sometimes we build VHDs based on VHDs that already have openSSH installed. We need open SSH for node debugging, so the CSE check will still be useful. But I'm open to discussions.

PR #5889, opened 2025-02-20
chore: add binaries for 1.32.0 1.32.1

wenhug:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/common/components.json
```diff
@@ -785,6 +785,11 @@
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/kube-proxy",
           "latestVersion": "v1.31.5",
           "previousLatestVersion": "v1.31.4"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/kube-proxy",
```
Devinwong, line 790: while we are caching one more version, is it appropriate to remove the oldest one?

vhdbuilder/packer/windows/components_json_helpers.tests.ps1
```diff
@@ -807,6 +807,8 @@ Describe 'Tests of components.json ' {
         #        $packages["c:\akse-cache\win-k8s\"] | Should -Contain "https://acs-mirror.azureedge.net/kubernetes/v1.31.3/windowszip/v1.31.3-1int.zip"
         $packages["c:\akse-cache\win-k8s\"] | Should -Contain "https://acs-mirror.azureedge.net/kubernetes/v1.31.4/windowszip/v1.31.4-1int.zip"
         $packages["c:\akse-cache\win-k8s\"] | Should -Contain "https://acs-mirror.azureedge.net/kubernetes/v1.31.5/windowszip/v1.31.5-1int.zip"
+        $packages["c:\akse-cache\win-k8s\"] | Should -Contain "https://acs-mirror.azureedge.net/kubernetes/v1.32.0/windowszip/v1.32.0-1int.zip"
```
timmy-wright, line 810: I've got a PR to modify the way were testing these on Windows, so this change shouldn't be necessary once that is merged. There's a risk that if my PR merges before yours, then we might have a test failure in master as a result. I'll monitor that tho.

timmy-wright (approved):

vhdbuilder/packer/windows/components_json_helpers.tests.ps1
wenhug: As you saied ,it cause a test failure, should I remove this change?

timmy-wright: Yup - backing the change out is the easiest fix.

PR #5893, opened 2025-02-20
bug: source helper script before calling functions

bravebeaver:
**What type of PR is this?**

/kindL bug

bug: the function is called before source is sourced

we have seen in the logs that 

```
+ aptmarkWALinuxAgent hold
/opt/azure/containers/provision.sh: line 9: aptmarkWALinuxAgent: command not found
```
**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong (approved):
lgtm.
I thought you were adding a new `source` action and I was worried about duplicate sourcing the same file. But actually you are just moving the logic earlier, so it looks good to me.

PR #5894, opened 2025-02-21
chore: don't use docker container to run make

timmy-wright:
**What type of PR is this?**

/kind chore

the go-dev container image is out of date. This PR migrates the windows VHD build pipeline to use the same structure as the Linux VHD build pipeline. It also extracts the override code from the Linux pipeline and uses it in the Windows pipeline so we can use the same override process for Windows as for Linux.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


bravebeaver:
LGTM

if i read it correctly:

- introduced 2 more variables for the pipeline
- update param in a new scritps
- formatting

bravebeaver (approved):
LGTM

PR #5896, opened 2025-02-21
bug: fix order of spinkube params and url

bravebeaver:
**What type of PR is this?**
/kind bug

it seems that the order of param to call spinkube is wrong.. fix it

clean up use of "local" not in a function, according to linter.
fix misspelt variable name and then realise the params are out of order.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

tested with existing e2e 

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -138,15 +138,15 @@ setupCNIDirs
...
+    versionsWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.versionsV2[].latestVersion' "$COMPONENTS_FILEPATH")
+    downloadLocationWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadLocation' "$COMPONENTS_FILEPATH")
+    downloadURLWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.downloadURL' "$COMPONENTS_FILEPATH")
     logs_to_events "AKS.CSE.installContainerdWasmShims" installContainerdWasmShims "$downloadLocationWasm" "$downloadURLWasm" "$versionsWasm"
 
-    local versionsSpinKube=$(jq -r '.Packages[] | select(.name == spinkube") | .downloadURIs.default.current.versionsV2[].latestVersion' "$COMPONENTS_FILEPATH")
-    local downloadLocationSpinKube=$(jq -r '.Packages[] | select(.name == "spinkube) | .downloadLocation' "$COMPONENTS_FILEPATH")
-    local downloadURLSpinKube=$(jq -r '.Packages[] | select(.name == "spinkube") | .downloadURIs.default.current.downloadURL' "$COMPONENTS_FILEPATH")
```
AlisonB319, line 148: @cameronmeissner had asked that I use local in the PR when I made this function, so we might want to keep them unless something has changed? Otherwise, lgtm thanks for catching this

AlisonB319 (approved):

djsly (approved):

PR #5899, opened 2025-02-21
chore: cache oss/v2/azure/ip-masq-agent-v2

wedaly:
**What type of PR is this?**

/kind cleanup

New version of ip-masq-agent-v2 is built with DALEC to a new repository path.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


robogatikov (approved):
lgtm

parts/common/components.json
```diff
@@ -538,6 +538,17 @@
       ],
```
wedaly, line 538: Keeping the old ip-masq-agent-v2 image for now in case we need to rollback, but we can remove v0.1.14 if we need space

ganeshkumarashok: Can we remove the older version in this PR? And then the one from the older image source in a subsequent one?

ganeshkumarashok: Ideally we remove the older image source in this same PR because caching is an optimization. But I understand your concern too.

wedaly: updated to remove old v0.1.14

djsly (approved):

Devinwong (approved):
lgtm

PR #5901, opened 2025-02-22
test: add taints and tolerations e2e test

Devinwong:
**What type of PR is this?**
/kind test

- Add taints and tolerations e2e test. In the next PR, I will add another one for scriptless.
- Intentionally adding taints test into an existing test to save some testing cost. Otherwise it will need to spin up a new node to test which is expensive.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/scenario_test.go
```diff
@@ -824,18 +824,20 @@ func Test_Ubuntu2204_ArtifactStreaming_Scriptless(t *testing.T) {
 	})
 }
 
-func Test_Ubuntu2204_ChronyRestarts(t *testing.T) {
+func Test_Ubuntu2204_ChronyRestarts_Taints_And_Tolerations(t *testing.T) {
```
Devinwong, line 827: Intentionally adding taints test into an existing test to save some testing cost. Otherwise it will need to spin up a new node to test which is expensive.

timmy-wright: I think we should do this more - ideally have one Test function per VHD and run a lot of scenarios on it.

timmy-wright (approved):

e2e/scenario_test.go
Devinwong: Ya. If the tests are not conflicting with one another.

PR #5902, opened 2025-02-23
perf: update benchmark titles and ingest performance data into kusto

zachary-bailey:
**What type of PR is this?**

/kind performance

This PR updates the benchmark titles through the build process to more closely align with the current build process. It also downloads and executes the build performance program.

The kusto database that the build performance program will ingest into will allow insights into performance and bugs in the VHD build process.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


bravebeaver (approved):
wondering why template variables are named in different cases. but minor.

looks good to me otherwise

PR #5907, opened 2025-02-24
add cloud-node-manager/credential provider v1.32.2 to vhd image

MartinForReal:
**What type of PR is this?**
/kind cleanup

add cloud-node-manager:v1.32.1 to vhd image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -469,6 +469,10 @@
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
           "latestVersion": "v1.31.2",
           "previousLatestVersion": "v1.31.1"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
+          "latestVersion": "v1.32.3"
         }
       ],
       "windowsVersions": [
@@ -511,6 +515,14 @@
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
           "latestVersion": "v1.31.2"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
+          "latestVersion": "v1.32.3"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
+          "latestVersion": "v1.32.3-windows-hpc"
         }
       ]
     },
@@ -868,6 +880,10 @@
         "windows": {
           "default": {
             "versionsV2": [
+              {
+                "renovateTag": "<DO_NOT_UPDATE>",
+                "latestVersion": "1.32.3"
+              },
               {
                 "renovateTag": "<DO_NOT_UPDATE>",
                 "latestVersion": "1.31.2"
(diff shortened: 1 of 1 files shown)


mainred (approved):

parts/common/components.json
```diff
@@ -470,6 +470,10 @@
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
           "latestVersion": "v1.31.2",
           "previousLatestVersion": "v1.31.1"
+        },
```
mainred, line 473: Since renovate can only bump versions and cannot create new release, we don't have to rely on renovate to introduce this cache.

MartinForReal:
https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=116123177&view=logs&jobId=0f8b8c4a-8219-5725-8fee-15683a5710b0&j=0f8b8c4a-8219-5725-8fee-15683a5710b0&t=754f3dbf-ba92-554b-fcb6-442d131fb984 build failed due to qutoa error. Is there any way to resolve this issue?

mainred:
I know @timmy-wright fixed the windows quota issue once, but sure if you can help fix it again and help review the PR.

parts/common/components.json
Devinwong: You are right

Devinwong (approved):

MartinForReal:
@Devinwong PTAL. Thanks!

MartinForReal:
@Devinwong I'm wondering if there is any workaround to fix disk full issue in windows VHD

Devinwong:
But you also need to resolve the conflict first.

parts/common/components.json
```diff
@@ -1274,7 +1290,7 @@
               {
                 "k8sVersion": "1.32",
                 "renovateTag": "OCI_registry=https://mcr.microsoft.com, name=oss/binaries/kubernetes/kubernetes-node",
-                "latestVersion": "v1.32.1",
+                "latestVersion": "v1.32.2",
```
timmy-wright, line 1293: This seems odd that we're updating latestVersion and not updateing previousLatestVersion - usually we'd change previousLatest to the previous latest rather than leave it the same.

vhdbuilder/packer/windows/components-test.json
```diff
@@ -470,6 +470,10 @@
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
           "latestVersion": "v1.31.2",
           "previousLatestVersion": "v1.31.1"
+        },
```
timmy-wright, line 473: This file should not need updating - it's just test data.

parts/common/components.json
```diff
@@ -512,6 +516,14 @@
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
           "latestVersion": "v1.31.2"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/azure-cloud-node-manager",
+          "latestVersion": "v1.32.2"
```
timmy-wright, line 522: I've kicked off the windows VHD pipelines to get better logs (they seemed to fail due to quota, which we should have fixed). However, adding two new container images to the windows VHD is very likely going to run out of disk space. Our VHDs are super limited in size and as the size is a product feature (we tell customers the minimum VHD size they need for windows nodes), changing the max size of our VHDs is a product decision. The normal course of action is to to drop some older versions of cloud-node-manager off the VHD when we add more.

timmy-wright (changes requested):
* revert the changes to components-test.json. It's a test file and we shouldn't change it unless we're adding a new feature to the file.
* ~This PR is likely to hit our VHD size issue, so we might need to remove some older versions of azure-cloud-node-manager~
* one change to the cached version doesn't seem to have updated the previousVersion that we cache on the VHD. This risks a mismatch between what RP wants and what the VHD provides. Which can create IcMs (I had a sev2 yesterday due to this!)

parts/common/components.json
Devinwong: correct. The idea is `lastest` and `latest-1`

parts/common/components.json
timmy-wright: The windows VHD pipeline passed, which means this hasn't caused us to go over the VHD size limit.

PR #5909, opened 2025-02-24
test: add taints e2e test for scriptless.

Devinwong:
**What type of PR is this?**
/kind test

test: add taints e2e test for scriptless.
Fix a related bug. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/scenario_test.go
```diff
@@ -843,18 +843,20 @@ func Test_Ubuntu2204_ChronyRestarts_Taints_And_Tolerations(t *testing.T) {
 	})
 }
 
-func Test_Ubuntu2204_ChronyRestarts_Scriptless(t *testing.T) {
+func Test_Ubuntu2204_ChronyRestarts_Taints_And_Tolerations_Scriptless(t *testing.T) {
```
Devinwong, line 846: Intentionally add the taints test into this existing test.

e2e/validators.go
```diff
@@ -430,7 +430,12 @@ func GetFieldFromJsonObjectOnNode(ctx context.Context, s *Scenario, fileName str
 
 // ValidateTaints checks if the node has the expected taints that are set in the kubelet config with --register-with-taints flag
 func ValidateTaints(ctx context.Context, s *Scenario) {
-	expectedTaints := s.Runtime.NBC.KubeletConfig["--register-with-taints"]
+	var expectedTaints string
+	if s.Runtime.NBC != nil {
+		expectedTaints = s.Runtime.NBC.KubeletConfig["--register-with-taints"]
+	} else {
+		expectedTaints = s.Runtime.AKSNodeConfig.KubeletConfig.KubeletFlags["--register-with-taints"]
```
r2k1, line 437: Can this throw panic?

Additionally, I'd prefer to pass the expected taints explicitly from the scenario validator.

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
```diff
@@ -5,6 +5,7 @@ option go_package = "github.com/Azure/agentbaker/aks-node-controller/pkg/gen/aks
 
 message KubeletConfig {
   // A list of taints to apply to the node.
+  // These are reserved for now. We are using KubeletFlags with flag "--register-with-taints" to register node with taints.
   repeated Taint taints = 1;
```
r2k1, line 9: Is it currently in use?

e2e/validators.go
Devinwong: My previous version was passing the kubeletconfig. But then I found that it's part of the `scenario` so updated with this version. Since this is just a test, I think I can change it to passing the kubeletconfig directly for explicitly.

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
```diff
@@ -38,6 +40,7 @@ message KubeletConfig {
 message Taint {
   string key = 1;
   string effect = 2;
+  string value = 3;
```
r2k1, line 43: I see you add the property, but I don't see any usage of it.
How does it work?

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
Devinwong: It's not. Setting it won't be processed by AKSNodeConfig

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
Devinwong: This is part of `taint`. It's not being used now. My intention was to reserve it for future use. But it's not a must

e2e/validators.go
r2k1: I would change the function signature to something like this:
`func ValidateTaints(ctx context.Context, s *Scenario, expectedTaints string)`

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
r2k1: why do you need to reserve something if we're not using it?

e2e/validators.go
Devinwong: agreed that's even better

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
Devinwong: My original plan was to implement this by extracting taints out from kubelet config. But it seems it will need more time than I have expected. I will remove it from the contract. :)

Devinwong: The linter will complain. Do you know if there is any way to suppress it?
```
proto/aksnodeconfig/v1/kubelet_config.proto:6:1:Previously present field "4" with name "startup_taints" on message "KubeletConfig" was deleted without reserving the name "startup_taints".
proto/aksnodeconfig/v1/kubelet_config.proto:6:1:Previously present field "1" with name "taints" on message "KubeletConfig" was deleted without reserving the number "1".
proto/aksnodeconfig/v1/kubelet_config.proto:6:1:Previously present field "4" with name "startup_taints" on message "KubeletConfig" was deleted without reserving the number "4".
```

Devinwong: I think since it's not enforced in the PR, so I can still push the changes. Once it's in production, we will avoid doing this.

Devinwong: Actually it's enforced.
image

r2k1: probably can use someone with superpower to force-merge the PR?

r2k1: I think it's safe in this case

Devinwong: sounds good. If you approve it, I can try the superpower

r2k1 (approved):

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
r2k1: another note, it seems like CI changes are reacting on breaking changes in .proto

I think it should be possible to change it to react only to breaking change in .json format.

PR #5910, opened 2025-02-25
minor change: Replace the hardcoded VM SKU 

zylxjtu:
**What type of PR is this?**
/kind bug

In the subscription I used for the pipeline run, I do not have the hard coded standard_d2ds_v5 for e2e test, so make it configurableconf

Fixes #
Replace the hardcoded VM SKU with the one configured in the config file.
Make the "nvidia-device-plugin" deployed in linux node only. It is not able to deploy to windows node, although it will not fail the test

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/kube.go
+++ b/e2e/kube.go
@@ -636,6 +636,9 @@ func nvidiaDevicePluginDaemonSet() *appsv1.DaemonSet {
 					},
 				},
 				Spec: corev1.PodSpec{
+					NodeSelector: map[string]string{
+						"kubernetes.io/os": "linux",
+					},
 					Tolerations: []corev1.Toleration{
 						{
 							Key:      "sku",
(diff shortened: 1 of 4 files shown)


e2e/kube.go
```diff
@@ -619,6 +619,9 @@ func nvidiaDevicePluginDaemonSet() *appsv1.DaemonSet {
 					},
 				},
 				Spec: corev1.PodSpec{
+					NodeSelector: map[string]string{
```
rzlink, line 622: Is this change necessary for this PR?

zylxjtu: I think it is better to have. the "nvidia-device-plugin" can not be pulled on windows node, so if you check the pods on the   AKS cluster being tested, you will see a failing pod. it does not fail the e2e test though, but I think we better to fix it.

timmy-wright (commented):
The PR looks good. Given it’s the linux E2E tests failing, we should get someone from the Linux team to comment - I’m not sure if these tests are known to fail in master or not. Your change is super unlikely to have caused any to fail, but better safe than sorry.

zylxjtu:
@cameronmeissner would you please help to take a look at this?

cameronmeissner (approved):
@zylxjtu lgtm, can you try rebasing off master and re-running e2e's to make sure this isn't introducing any issues?

PR #5913, opened 2025-02-25
don't use azcopy-preview

timmy-wright:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
+++ b/vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
@@ -119,7 +119,6 @@ capture_benchmark "${SCRIPT_NAME}_grant_access_to_disk"
 echo "Uploading $disk_resource_id to ${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"
 
 echo "Setting azcopy environment variables with pool identity: $AZURE_MSI_RESOURCE_STRING"
-# TBD: Need to investigate why `azcopy-preview login --login-type=MSI` does not work for Windows
 export AZCOPY_AUTO_LOGIN_TYPE="MSI"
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -119,16 +119,11 @@ capture_benchmark "${SCRIPT_NAME}_grant_access_to_disk"
...
 export AZCOPY_MSI_RESOURCE_STRING="$AZURE_MSI_RESOURCE_STRING"
 
-if [[ "${OS_TYPE}" == "Linux" ]]; then
-  export AZCOPY_CONCURRENCY_VALUE="AUTO"
-  azcopy copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd" --recursive=true || exit $?
-else
-  azcopy-preview copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd" --recursive=true || exit $?
-fi
+export AZCOPY_CONCURRENCY_VALUE="AUTO"
+azcopy copy "${sas}" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd" --recursive=true || exit $?
```
smith1511, line 126: Should we comment that this is provided by the ADO pipeline step?  Which changed?

smith1511 (approved):

zachary-bailey (approved):

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
timmy-wright: Probably. I'm about to do test runs of all our pipelines today as I'm not confident they work after the ADO changes.

PR #5916, opened 2025-02-25
bug: don't need to set agentkey in the cs

timmy-wright:
/kind bug

AgentKey is being set both in the CSE and the run command. It only needs to be set in one place. And the run command is more secure.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


smith1511 (approved):
lgtm

PR #5927, opened 2025-02-26
test: fix the validation of outbound-check-skipped

bingosummer:
**What type of PR is this?**

/kind test

`ls -al /var/run` doesn't show the contents since `/var/run` is a link to `/run`. So we need to run `ls -al /run` to check whether the file `outbound-check-skipped` exists.

```
$ ls -al /var/run
lrwxrwxrwx 1 root root 4 Aug 25  2021 /var/run -> /run
```

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


AbelHu (approved):
LGTM. Thanks for the PR.

PR #5933, opened 2025-02-26
chore: move base image urls to pipeline vars

timmy-wright:
**What type of PR is this?**

 /kind chore

Moves the base image URLs to pipeline params rather than pipeline vars. This will make it easier to add new windows versions in the future.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -59,6 +59,22 @@ parameters:
     displayName: Branch in aks-rp to use for overrides
     type: string
     default: master
+  - name: ws2019BaseImageUrl
+    displayName: WS2019 base image URL
+    type: string
+    default:
+  - name: ws2022BaseImageUrl
+    displayName: WS2022 base image URL
+    type: string
+    default:
+  - name: ws23H2BaseImageUrl
+    displayName: WS23H2 base image URL
+    type: string
+    default:
+  - name: ws2025BaseImageUrl
+    displayName: ws2025 base image URL
+    type: string
+    default:
 
 # Use variable group "ab-windows-ame-tenant" and link it to the pipeline "AKS Windows VHD Build"
 # Use variable group "ab-windows-ame-tenant" and link it to the pipeline "AKS Windows VHD Build - PR check-in gate"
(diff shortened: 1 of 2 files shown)


.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -59,6 +59,22 @@ parameters:
     displayName: Branch in aks-rp to use for overrides
     type: string
     default: master
+  - name: ws2019BaseImageUrl
```
AbelHu, line 62: We keep these variables as secret and do not want to expose the values in any logs.

timmy-wright: I think it's OK to expose them in the devops logs as they are only accessible internally. Also, anyone who has access to the pipelines can view the variables or parameters that were used to run the pipelines - so I'm not sure we gain much by using variables.

.pipelines/templates/.build-and-test-windows-vhds-template.yaml
```diff
@@ -55,7 +71,7 @@ stages:
       buildVmSize: ${{ parameters.buildVmSize }}
       installOpenSshServer: ${{ parameters.installOpenSshServer }}
       skipExtensionCheck: ${{ parameters.skipExtensionCheck }}
-      windowsBaseImageUrl: $(WINDOWS_2019_BASE_IMAGE_URL)
+      windowsBaseImageUrl: ${{ parameters.ws2019BaseImageUrl }}
       windowsNanoImageUrl: $(WINDOWS_2019_NANO_IMAGE_URL)
```
zylxjtu, line 75: do we want to make the windowsNanoImageUrl and windowsCoreImageUrl the same as the windowsbaseimageurl?

timmy-wright: These ones are used far less than the base image url - and I'm not sure they work at the moment anyway - so I'd prefer to keep them as variables until we need them.

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -59,6 +59,22 @@ parameters:
...
+    type: string
+    default:
+  - name: ws23H2BaseImageUrl
+    displayName: WS23H2 base image URL
+    type: string
+    default:
+  - name: ws2025BaseImageUrl
+    displayName: WS2025 base image URL
+    type: string
+    default:
```
cameronmeissner, line 77: nit: `default:` is already the default for string vars :)

PR #5944, opened 2025-02-27
adding parts/linux/cloud-init/artifacts/* to shellcheck

Devinwong:
**What type of PR is this?**
/kind test

- adding parts/linux/cloud-init/artifacts/* to shellcheck
- fixed some shellcheck errors
- suppress some warnings that seems fine to me

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


bravebeaver (approved):

jason1028kr (approved):

cameronmeissner (changes requested):
seems to have an unexpected diff with the master branch

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -72,7 +72,9 @@ if [[ "${SHOULD_CONFIGURE_CUSTOM_CA_TRUST}" == "true" ]]; then
     logs_to_events "AKS.CSE.configureCustomCaCertificate" configureCustomCaCertificate || exit $ERR_UPDATE_CA_CERTS
 fi
 
-registry_domain_name="${MCR_REPOSITORY_BASE%/}"
+registry_domain_name="${MCR_REPOSITORY_BASE:-mcr.microsoft.com}"
```
Devinwong, line 75: add a default value for registry_domain_name in case MCR_REPOSITORY_BASE is not set

Devinwong:
> seems to have an unexpected diff with the master branch

Thanks for catching this.
I rebased again and it looks good to me now.

parts/linux/cloud-init/artifacts/cse_main.sh
timmy-wright: I'm in two minds about this - sometimes it's better to abort when expected data is missing as it makes debugging the problem easier. At least, it'd be nice to log that we're falling back to default.

parts/linux/cloud-init/artifacts/aks-check-network.sh
```diff
@@ -100,7 +100,7 @@ function check_and_curl {
         if [ $response -ge 200 ] && [ $response -lt 400 ]; then
             logs_to_events "AKS.testingTraffic.success" "echo '$(date) - SUCCESS: Successfully tested $url with returned status code $response'"
             break
-        elif [ $response -eq 400 ] && ([ $url == "acs-mirror.azureedge.net" ] || [ $url == "eastus.data.mcr.microsoft.com" ]); then
```
timmy-wright, line 103: So strange that shellcheck doesn't pickup that $url should be in quotes - if there is a space in the url then weird things will happen.

parts/linux/cloud-init/artifacts/aks-log-collector.sh
```diff
@@ -174,7 +174,7 @@ trap "cleanup" EXIT
 # any disk space aside from the ZIP file itself.
 # USAGE: collectToZip FILENAME CMDTORUN
 function collectToZip {
-  command -v "${2}" >/dev/null || { printf "${2} not found, skipping.\n"; return; }
```
timmy-wright, line 177: Does this need to be a printf now that we're not using any formatting capabilities? "echo" might work just as well.

timmy-wright: Oh wait, I read the before/after aroudn the wrong way. But I still wonder - why not just use "echo" isntead of printf?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -303,7 +303,7 @@ installSpinKube(){
 
     for version in "${package_versions[@]}"; do
         containerd_spinkube_url=$(evalPackageDownloadURL ${PACKAGE_DOWNLOAD_URL})
-        downloadSpinKube $download_location $containerd_spinkube_url "v$version" # adding v to version for simplicity
+        downloadSpinKube $download_location $containerd_spinkube_url "$version"
```
timmy-wright, line 306: Where does the $version value come from - I didn't notice us add the "v" back in anywhere, so this might be a regression.

timmy-wright (commented):
Mostly looks good - I had a couple of questions or comments that might not matter.

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: It's from components.json and it actually has a 'v' already in that url. 
"downloadURL": "https://acs-mirror.azureedge.net/spinkube/v${version}/linux/${CPU_ARCH}"

parts/linux/cloud-init/artifacts/cse_main.sh
Devinwong: Agree if it's somewhere MCR_REPOSITORY_BASE is really important, although it still has a default value. https://github.com/Azure/AgentBaker/blob/devinwon/kube_binaries_v_bug/parts/linux/cloud-init/artifacts/cse_config.sh#L344
I believe the owner is Bin Xia.

Here it's just using it to validate DNS health. So whether mcr.microsoft.com or other URL is fine.

parts/linux/cloud-init/artifacts/aks-check-network.sh
Devinwong: I just looked into that. It was because it was a `info` SC2086 not an error and 4 years ago this `info` was set to ignore.
Let me enable it and fix related errors in the next PR otherwise this PR will be very huge.
Thanks for catching this though. It reduces the chance of potential errors.

parts/linux/cloud-init/artifacts/aks-log-collector.sh
Devinwong: I am not sure. Is there a general guideline that we should use echo? There are 3 printf in this .sh in total.
Copilot suggested using printf, although it complained another thing which it didn't really correct in its "corrected codes"
image

parts/linux/cloud-init/artifacts/aks-logrotate.sh
```diff
@@ -3,10 +3,10 @@
 # This will be saved on the target VM within /usr/local/bin/logrotate.sh and invoked by logrotate.service
 
 # Clean non existent log file entries from status file
-cd /var/lib/logrotate
+cd /var/lib/logrotate || exit
```
djsly, line 6: should we echo a message here ?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -239,7 +239,7 @@ installContainerdWasmShims(){
             shims_to_download+=("wws")
         fi
         containerd_wasm_url=$(evalPackageDownloadURL ${PACKAGE_DOWNLOAD_URL})
-        downloadContainerdWasmShims $download_location $containerd_wasm_url "v$version" "${shims_to_download[@]}" # adding v to version for simplicity
+        downloadContainerdWasmShims $download_location $containerd_wasm_url "$version" "${shims_to_download[@]}"
```
djsly, line 242: is this related to shellcheck ?  is $version now storing the `v` ?

djsly (approved):

parts/linux/cloud-init/artifacts/aks-logrotate.sh
Devinwong: Sounds good. Added.

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Let me add the `v` back and do a test now

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1093,7 +1093,6 @@ testWasmRuntimesInstalled() {
   local test="testWasmRuntimesInstalled"
   local wasm_runtimes_path=${1}
   local shim_version=${2}
-  shim_version="v${shim_version}"
```
cameronmeissner, line 1096: we're sure this is correct?

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner, line 75: maybe more (re)usable if we default this directly in `GetMCRRepositoryBase` in the func map in baker.go?

parts/linux/cloud-init/artifacts/aks-logrotate.sh
```diff
@@ -3,10 +3,10 @@
 # This will be saved on the target VM within /usr/local/bin/logrotate.sh and invoked by logrotate.service
 
 # Clean non existent log file entries from status file
-cd /var/lib/logrotate
+cd /var/lib/logrotate || { echo "Failed to change directory to /var/lib/logrotate. Exiting."; exit 1; }
```
cameronmeissner, line 6: this seems silly, could we just `mkdir -p /var/lib/logrotate` and omit this error condition?

vhdbuilder/packer/test/linux-vhd-content-test.sh
Devinwong: Actually no. I just found that this line can't be removed, although it's another bug related to this line. I will add it back for now.

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: I will keep the 'v' for now, though I found there is a bug for that. But I wanted to keep this PR pure for shellcheck.

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: I will keep the 'v' for now, though I found there is a bug for that. But I wanted to keep this PR pure for shellcheck.

parts/linux/cloud-init/artifacts/aks-logrotate.sh
Devinwong: Sounds good. Done

timmy-wright (approved):

parts/linux/cloud-init/artifacts/aks-check-network.sh
timmy-wright: I'm mostly aware about this at the moment because my IDE underlies all unquoted shell variables and warns me about them. Which is handy, but annoying.

cameronmeissner (approved):

PR #5972, opened 2025-03-01
fix kubernetes-binaries name check bug

Devinwong:
**What type of PR is this?**
/kind bug

Fix kubernetes-binaries particular version naming bug. The bug is that even we cached the correct Kubernetes-binaries at build time, it will still download it during provisioning time because cse_install is checking different names. For example, the name it caches is kubectl-v3.1.5 but the name it checks is kubectl-3.1.5. (without v)

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -535,6 +535,8 @@ installAzureCNI() {
 
 extractKubeBinaries() {
     local k8s_version="$1"
+    # Remove leading 'v' if it exists
+    k8s_version="${k8s_version#v}"
```
AlisonB319, line 539: From git blame it seems this has been in cse for around a year ish - is it more likely that this is just a good performace boost rather than what caused a recent regression?

AlisonB319 (approved):

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Ya. This should have been here for long. I couldn't find another root cause of the regression. So my plan was to monitor it with the query in this IcM. https://portal.microsofticm.com/imp/v5/incidents/details/595252640/summary

PR #5974, opened 2025-03-03
bug: use only the last line to assign function return value

bravebeaver:
**What type of PR is this?**

/kind: bug

when a function echo multiple lines, the function result is a multi-line string.

in this case we want the last line.

so that the return value of `FULL_INSTALL_REQUIRED` is correct

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

currently the variable is set as 

```
+ FULL_INSTALL_REQUIRED='binaries will not be cleaned up
true'
```

we want 

```
+ FULL_INSTALL_REQUIRED='true'
```

**Release note**:

```
none
```


e2e/scenario_test.go
```diff
@@ -884,7 +884,7 @@ func Test_Ubuntu2204_ChronyRestarts_Taints_And_Tolerations_Scriptless(t *testing
 
 func Test_AzureLinuxV2_CustomCATrust(t *testing.T) {
 	RunScenario(t, &Scenario{
-		Description: "Tests that a node using the Azure Linux VHD can be properly bootstrapped and custom CA was correctly added",
+		Description: "Tests that a node using the Azure Linux 2204 VHD can be properly bootstrapped and custom CA was correctly added",
```
djsly, line 887: there are no such thing as Azure Linux 2204

e2e/scenario_test.go
```diff
@@ -1240,7 +1240,6 @@ func Test_Ubuntu2204_ContainerdHasCurrentVersion(t *testing.T) {
 			Cluster: ClusterKubenet,
 			VHD:     config.VHDUbuntu2204Gen2Containerd,
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
```
djsly, line 1242: do we need the BootstrapConfigMutator in the end ?

djsly (approved):

djsly:
@bravebeaver what made you work on this ?  @AlisonB319  could this explain why we are not cleaning things up anymore (Skyman issue ?)

PR #5976, opened 2025-03-03
Make a few parameters to be optional in windows e2e tests

zylxjtu:
In this PR, it will:
(1): Make the k8s verison to be configurable
(2): Make the openssh to be optinal in e2e test, this will be helpful for scenerios where openssh will not be able to be installed, but we stil want to test some K8s related scenrios

**What type of PR is this?**
/kind test

(1): Make the k8s verison to be configurable
(2): Make the openssh to be optinal in e2e test, this will be helpful for scenerios where openssh will not be able to be installed, but we stil want to test some K8s related scenrios

All the configurable parameters can be set through env/pipeline

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/config/config.go
+++ b/e2e/config/config.go
@@ -44,11 +44,13 @@ type Configuration struct {
 	GallerySubscriptionIDWindows           string        `env:"GALLERY_SUBSCRIPTION_ID_WINDOWS" envDefault:"4be8920b-2978-43d7-ab14-04d8549c1d05"`
 	IgnoreScenariosWithMissingVHD          bool          `env:"IGNORE_SCENARIOS_WITH_MISSING_VHD"`
 	KeepVMSS                               bool          `env:"KEEP_VMSS"`
+	KubernetsVersion                       string        `env:"KUBERNETS_VERSION" envDefault:"1.30.9"`
 	Location                               string        `env:"E2E_LOCATION" envDefault:"westus3"`
 	SIGVersionTagName                      string        `env:"SIG_VERSION_TAG_NAME" envDefault:"branch"`
 	SIGVersionTagValue                     string        `env:"SIG_VERSION_TAG_VALUE" envDefault:"refs/heads/master"`
 	SkipTestsWithSKUCapacityIssue          bool          `env:"SKIP_TESTS_WITH_SKU_CAPACITY_ISSUE"`
 	SubscriptionID                         string        `env:"SUBSCRIPTION_ID" envDefault:"8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8"`
+	SSHEnabled                             bool          `env:"SSH_ENABLED" envDefault:"true" `
 	TagsToRun                              string        `env:"TAGS_TO_RUN"`
 	TagsToSkip                             string        `env:"TAGS_TO_SKIP"`
 	TestTimeout                            time.Duration `env:"TEST_TIMEOUT" envDefault:"35m"`
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -461,7 +461,7 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 // this been crafted with a lot of trial and pain, some values are not needed, but it takes a lot of time to figure out which ones.
 // and we hope to move on to a different config, so I don't want to invest any more time in this-
 func baseTemplateWindows(t *testing.T, location string) *datamodel.NodeBootstrappingConfiguration {
-	kubernetesVersion := "1.29.9"
+	kubernetesVersion := config.Config.KubernetsVersion
 	config := &datamodel.NodeBootstrappingConfiguration{
 		TenantID:          "tenantID",
 		SubscriptionID:    config.Config.SubscriptionID,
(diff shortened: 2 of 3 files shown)


e2e/scenario_win_test.go
```diff
@@ -18,13 +18,15 @@ func Test_Windows2019Containerd(t *testing.T) {
...
 			BootstrapConfigMutator: func(configuration *datamodel.NodeBootstrappingConfiguration) {},
 			Validator: func(ctx context.Context, s *Scenario) {
-				ValidateWindowsVersionFromWindowsSettings(ctx, s, "2019-containerd")
-				ValidateWindowsProductName(ctx, s, "Windows Server 2019 Datacenter")
-				// TODO: currently the command used to get the display name returns an empty string on WS2019. Need to find a better command.
-				//ValidateWindowsDisplayVersion(ctx, s, "???")
-				ValidateFileHasContent(ctx, s, "/k/kubeletstart.ps1", "--container-runtime=remote")
-				ValidateWindowsProcessHasCliArguments(ctx, s, "kubelet.exe", []string{"--rotate-certificates=true", "--client-ca-file=c:\\k\\ca.crt"})
-				ValidateCiliumIsNotRunningWindows(ctx, s)
+				if config.Config.SSHEnabled {
```
zylxjtu, line 21: without openssh, the following test will not be able to pass, but with the testing AKS cluster available, we can add other test suites through pipeline/yaml file

e2e/node_config.go
```diff
@@ -461,7 +461,7 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 // this been crafted with a lot of trial and pain, some values are not needed, but it takes a lot of time to figure out which ones.
 // and we hope to move on to a different config, so I don't want to invest any more time in this-
 func baseTemplateWindows(t *testing.T, location string) *datamodel.NodeBootstrappingConfiguration {
-	kubernetesVersion := "1.29.9"
+	kubernetesVersion := config.Config.KubernetsVersion
```
msft-linliu, line 464: This change uplifted the default Kubernetes version from 1.29.9 to 1.30.9. I am fine with this change, but want to call out that change in case folks have concerns

timmy-wright: I'll ping the Linux team to check it's OK for them. All good from my side tho.

e2e/scenario_win_test.go
timmy-wright: For the openssh, I see we're bypassing all the checks on the node. I think I have a way around this, but haven't implemented it yet. Basically, at the moment, to run powershell on a node we:

- create a container on a different node
- run bash in that container, 
- scp the powershell script to the node, then 
- ssh to the node and run the script

We could instead

- create a privlidged powershell container on the node under test
- run the powershell script there

I'll add a task to my todo list to implement that functionality as it'd be nice to still run the tests when a ssh server isn't running. But it's not a blocker for this PR.

e2e/node_config.go
cameronmeissner: discussed offline, we'd prefer resolving the k8s version from the e2e cluster's control plane as an overridable default, though we should be careful about writing version-dependent tests

e2e/config/config.go
```diff
@@ -44,11 +44,13 @@ type Configuration struct {
...
 	IgnoreScenariosWithMissingVHD          bool          `env:"IGNORE_SCENARIOS_WITH_MISSING_VHD"`
 	KeepVMSS                               bool          `env:"KEEP_VMSS"`
+	KubernetsVersion                       string        `env:"KUBERNETS_VERSION" envDefault:"1.30.9"`
 	Location                               string        `env:"E2E_LOCATION" envDefault:"westus3"`
 	SIGVersionTagName                      string        `env:"SIG_VERSION_TAG_NAME" envDefault:"branch"`
 	SIGVersionTagValue                     string        `env:"SIG_VERSION_TAG_VALUE" envDefault:"refs/heads/master"`
 	SkipTestsWithSKUCapacityIssue          bool          `env:"SKIP_TESTS_WITH_SKU_CAPACITY_ISSUE"`
 	SubscriptionID                         string        `env:"SUBSCRIPTION_ID" envDefault:"8ecadfc9-d1a3-4ea4-b844-0d9f87e4d7c8"`
+	SSHEnabled                             bool          `env:"SSH_ENABLED" envDefault:"true" `
 	TagsToRun                              string        `env:"TAGS_TO_RUN"`
```
AbelHu, line 54: We have a property `SSHEnabled` in `WindowsProfile` . could we use this one?

e2e/scenario_win_test.go
```diff
@@ -40,12 +42,14 @@ func Test_Windows2022Containerd(t *testing.T) {
 			BootstrapConfigMutator: func(configuration *datamodel.NodeBootstrappingConfiguration) {
 			},
 			Validator: func(ctx context.Context, s *Scenario) {
-				ValidateWindowsVersionFromWindowsSettings(ctx, s, "2022-containerd")
-				ValidateWindowsProductName(ctx, s, "Windows Server 2022 Datacenter")
-				ValidateWindowsDisplayVersion(ctx, s, "21H2")
-				ValidateFileHasContent(ctx, s, "/k/kubeletstart.ps1", "--container-runtime=remote")
-				ValidateWindowsProcessHasCliArguments(ctx, s, "kubelet.exe", []string{"--rotate-certificates=true", "--client-ca-file=c:\\k\\ca.crt"})
-				ValidateCiliumIsNotRunningWindows(ctx, s)
+				if config.Config.SSHEnabled {
```
AbelHu, line 45: With SSH disabled, I think that you can exec into a system container (e.g. csi pod) to run the check. Otherwise, there is no validation.

PR #5978, opened 2025-03-03
fix: extra leading v in ContainerdWasmShims version

Devinwong:
**What type of PR is this?**
/kind bug

fix: extra leading v in ContainerdWasmShims version

By enabling shellcheck on all the cse*.scripts, we are now able to catch this bug.

Before this change, the ab e2e ran like this
```++ jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.versionsV2[].latestVersion' /opt/azure/components.json
+ local 'versionsWasm=v0.3.0
v0.5.1
v0.8.0'
/opt/azure/containers/provision.sh: line 137: local: can only be used in a function
++ jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadLocation' /opt/azure/components.json
**+ local downloadLocationWasm=/usr/local/bin
/opt/azure/containers/provision.sh: line 138: local: can only be used in a function**
++ jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.downloadURL' /opt/azure/components.json
+ local 'downloadURLWasm=mcr.microsoft.com/oss/binaries/deislabs/containerd-wasm-shims:${version}-linux-${CPU_ARCH}'
/opt/azure/containers/provision.sh: line 139: local: can only be used in a function
+ logs_to_events AKS.CSE.installContainerdWasmShims installContainerdWasmShims '' '' ''
```

At both build time and provisioning time, in the script here, https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_main.sh#L146,
there was an invalid `local` before the variable `versionsWasm` so it didn't get the correct versions. As a result, empty string was passed to installContainerdWasmShims. Therefore it didn't do anything. 

By enabling shellcheck we removed the `local`,  and it started to complain this in e2e.
```
Error response from registry: failed to resolve vv0.3.0-linux-amd64: privateacre2ewestus3.azurecr.io/oss/binaries/deislabs/containerd-wasm-shims:vv0.3.0-linux-amd64: not found
```
Note the `vv` which is incorrect. This PR is to fix this issue by removing the extra `v` in the version.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong:
@AlisonB319 I assumed double v in `containerd-wasm-shims:vv0.3.0-linux-amd64` is unexpected, right?

cameronmeissner (approved):
lgtm pending passing tests

Devinwong:
> lgtm pending passing tests

yes I will wait for the tests results to double check.

djsly (approved):

Devinwong:
The AgentBaker E2E PR check failed with error code 215 at Test_Ubuntu2204_GPUA10_Scriptless is expected for now. It's because scriptless need to rebuild a VHD for the scripts changes to apply. I am building the VHDUbuntu2204Gen2Containerd on master branch.
But the PR check-in gate (build build2204gen2containerd) and PR check-in gate (build buildmarinerv2gen2) should pass if this fix is really correct.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1093,7 +1093,6 @@ testWasmRuntimesInstalled() {
   local test="testWasmRuntimesInstalled"
   local wasm_runtimes_path=${1}
   local shim_version=${2}
-  shim_version="v${shim_version}"
```
timmy-wright, line 1096: nit (not blocking): this could be combined with the previous line - no need to assign to a variable and then immediately change it.

timmy-wright (approved):

vhdbuilder/packer/test/linux-vhd-content-test.sh
Devinwong: this line is deleted.

PR #5982, opened 2025-03-04
perf: validate credentials and check apiserver connectivity before starting kubelet

cameronmeissner:
**What type of PR is this?**

we recently discovered that in cases where the kubelet tries to begin bootstrapping its client credential before its bootstrap token has been installed into the running control plane, kubelet will attempt to retry bootstrapping with a very aggressive exponential backoff which ends up causing node registration to be unnecessarily delayed

this PR adds a new script which is responsible for validating that kubelet's credential, whether that be a kubeconfig or bootstrap-kubeconfig, is valid and can be used to successfully authenticate with the apiserver via `kubectl auth whoami`. this script will be run as a further ExecStartPre specified within the kubelet.service unit file.

we already validate that the apiserver is up and reachable before starting kubelet to ensure that we avoid falling into the same trap of kubelet's backoff loop

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/kubelet.service
```diff
@@ -20,6 +20,8 @@ ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
 ExecStartPre=-/sbin/ebtables -t nat --list
 ExecStartPre=-/sbin/iptables -t nat --numeric --list
 
+ExecStartPre=/bin/bash /opt/azure/containers/validate-kubelet-credentials.sh
```
djsly, line 23: are `ExecStartPre` blocking and sequential ? since this is expecting that `/var/lib/kubelet` is ready and we are mount that folder in a `ExecStartPre` right before.

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
```diff
@@ -0,0 +1,50 @@
...
+BOOTSTRAP_KUBECONFIG_PATH="${BOOTSTRAP_KUBECONFIG_PATH:-/var/lib/kubelet/bootstrap-kubeconfig}"
+
+VALIDATE_KUBELET_CREDENTIALS_MAX_RETRIES=${VALIDATE_KUBELET_CREDENTIALS_MAX_RETRIES:-10}
+VALIDATE_KUBELET_CREDENTIALS_RETRY_DELAY_SECONDS=${VALIDATE_KUBELET_CREDENTIALS_RETRY_DELAY_SECONDS:-3}
+VALIDATE_KUBELET_CREDENTIALS_RETRY_TIMEOUT_SECONDS=${VALIDATE_KUBELET_CREDENTIALS_RETRY_TIMEOUT_SECONDS:-5}
+
+function validateKubeconfig {
+    local kubeconfig_path=$1
+
+    if ! retrycmd_if_failure $VALIDATE_KUBELET_CREDENTIALS_MAX_RETRIES \
```
djsly, line 17: is 30sec long enough ?

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
```diff
@@ -0,0 +1,50 @@
...
+        echo "both kubeconfig: $KUBECONFIG_PATH and bootstrap-kubeconfig: $BOOTSTRAP_KUBECONFIG_PATH do not exist, unable to start kubelet"
+        exit 1
+    fi
+
+    if ! which kubectl; then
+        echo "kubectl not found, will skip kubelet credential validation"
+        exit 0
+    fi
+
+    if [ -f "$KUBECONFIG_PATH" ]; then
```
djsly, line 38: do we need this condition since we are checking above that both Kubeconfig file exists

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -409,6 +409,13 @@ write_files:
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "ensureIMDSRestrictionScript"}}
 
+- path: /opt/azure/containers/validate-kubelet-credentials.sh
+  permissions: "0755"
+  encoding: gzip
+  owner: root
+  content: !!binary |
```
djsly, line 416: what does this do ? we pass the script from the CustomData ?

parts/linux/cloud-init/artifacts/kubelet.service
cameronmeissner: yes -

`Additional commands that are executed before or after the command in ExecStart=, respectively. Syntax is the same as for ExecStart=. Multiple command lines are allowed, regardless of the service type (i.e. Type=), and the commands are executed one after the other, serially.`

https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExecStartPre=

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
cameronmeissner: I think it's a sane starting point, in the majority of cases the control plane should already have the bootstrap token installed - though I think it might be better to not fail provisioning by failing kubelet startup if we fail to validate, and rather treat this as a best-effort type approach, that way we don't have to be as worried about this and we can also take a data driven approach as to what the real limit should be after having guest agent event telemetry once this is rolled out widely

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
cameronmeissner: yes, the first condition above checks that _at least_ one of kubeconfig or bootstra-kubeconfig exists, not that both exist

parts/linux/cloud-init/nodecustomdata.yml
cameronmeissner: this adds the file within our customdata payload - until we're confident that newer CSE/customdata is never being executed on older VHDs, this is required as to not take a dependency on VHD version

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
```diff
@@ -0,0 +1,62 @@
...
+            $VALIDATE_KUBELET_CREDENTIALS_RETRY_DELAY_SECONDS \
+            $VALIDATE_KUBELET_CREDENTIALS_RETRY_TIMEOUT_SECONDS \
+            kubectl auth can-i create certificatesigningrequests --kubeconfig $kubeconfig_path >/dev/null
+        code=$?
+    fi
+
+    if [ $code -ne 0 ]; then
+        # for now we simply exit 0 here to prevent provisioning failures in cases where the credential
+        # doesn't become valid until after we've exhausted our retries - kubelet should still eventually be able to register
+        echo "kubelet credential validation failed, will still attempt to start kubelet"
```
djsly, line 34: can we augment the echo with the number of retries and total delays.

PR #5983, opened 2025-03-04
Remove Nvidia device plugin from being download in the VHD

ganeshkumarashok:
**What type of PR is this?**
/kind cleanup 

Since the GPU VHD has been deprecated, we don't need to cache this device plugin on the VHD. Downloading the device-plugin now is unnecessary, takes up the limited VHD space, increases build time and augments the security attack surface area (ex: this is not being auto-updated). 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


UtheMan (approved):

cameronmeissner (approved):
we aren't specifying this on components.json as well are we?

ganeshkumarashok:
> we aren't specifying this on components.json as well are we?

nope

PR #5988, opened 2025-03-05
regression: set windows 2019 disk size

timmy-wright:
**What type of PR is this?**

/kind regression

We should be enforcing a minimum VHD size of 30GB unless a particular VHD overrides this. This default went missing during a refactor. Adding back. It only impacts WS2019 as that VHD doesn't have an override size set.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/init-variables.sh
+++ b/vhdbuilder/packer/init-variables.sh
@@ -352,6 +352,8 @@ if [ "$OS_TYPE" == "Windows" ]; then
   if [ "null" != "${OS_DISK_SIZE}" ]; then
     echo "Setting os_disk_size_gb to the value in windows-settings.json for ${WINDOWS_SKU}: ${OS_DISK_SIZE}"
     os_disk_size_gb=${OS_DISK_SIZE}
+  else
+    os_disk_size_gb="30"
   fi
 
   imported_windows_image_name="${WINDOWS_IMAGE_NAME}-imported-${CREATE_TIME}-${RANDOM}"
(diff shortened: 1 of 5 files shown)


vhdbuilder/packer/init-variables.sh
```diff
@@ -352,6 +352,8 @@ if [ "$OS_TYPE" == "Windows" ]; then
   if [ "null" != "${OS_DISK_SIZE}" ]; then
     echo "Setting os_disk_size_gb to the value in windows-settings.json for ${WINDOWS_SKU}: ${OS_DISK_SIZE}"
     os_disk_size_gb=${OS_DISK_SIZE}
+  else
+    os_disk_size_gb="30"
```
AbelHu, line 356: does this impact any Windows2019 vhd size AKS built in past months?

timmy-wright: It does likely impact the most recent patch tuesday release. I will check the size of that VHD.

AbelHu, line 356: Could we throw an error if OS_DISK_SIZE is not set and set OS_DISK_SIZE for WS2019 in https://github.com/Azure/AgentBaker/blob/45394125976b8acc748602665eb913382cfa1d80/vhdbuilder/packer/windows/windows_settings.json#L23?

timmy-wright: I'll do that as well - but it's good to have a fallback.

timmy-wright: Have set it to 30GB and also changed the schema validator for windows_settings.json to fail if os_disk_size is unset.

smith1511 (approved):

PR #5990, opened 2025-03-05
upgrade windows pause container

timmy-wright:
**What type of PR is this?**

/kind chore

Update the pause container to match what is in rp.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -80,7 +80,7 @@
       "windowsVersions": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/pause",
-          "latestVersion": "3.9"
+          "latestVersion": "3.9-hotfix-20230808"
         }
       ]
     },


parts/common/components.json
```diff
@@ -80,7 +80,7 @@
       "windowsVersions": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/pause",
-          "latestVersion": "3.9"
+          "latestVersion": "3.9-hotfix-20230808"
```
AbelHu, line 83: You may also need to check whether there are wrong values in the recent PRs.

AbelHu (approved):
LGTM. Thanks for the fix.

parts/common/components.json
timmy-wright: Yeah, we need to go through all the image versions and check they match what is in RP.

smith1511, line 83: Does containerd config need to be updated with this?  Or it just grabs the latest 3.9 available?

smith1511 (approved):

parts/common/components.json
timmy-wright: The pause image version is specified in an rp map, that (I assume) is passed to the CSE that creates the containerd config.

PR #5992, opened 2025-03-05
regression: get the WS2019 VHD back to 30GB

timmy-wright:
**What type of PR is this?**

/kind regression

Product limit for WS2019 to have max 30GB VHD size. We had to increase it to 31 GB recently to re-enforce a size (as VHD size validation wasn't working), this PR fixes the size.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


timmy-wright:
> lgtm. Not sure if caching more versions will break the 30GB limit again.

@Devinwong - caching more versions will certainly break the 30GB limit. But at least we'll know!

PR #5998, opened 2025-03-06
chore: improve auto PR comments when windows cached things change

timmy-wright:
**What type of PR is this?**

/kind feature

Uses a different way of updating comments on PRs relating to windows VHDs. This lets us have triggers so we only add comments when necessary.

Also changed the context lines for the diff to 0 - as the context doesn't matter in this case.

When changes:

image

When no changes:

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


bravebeaver:
LGTM. although took me a while to understand why it says changed when the a and b look the same. Do I get it right that it is the file content that has been changed?

bravebeaver (approved):

PR #6004, opened 2025-03-06
disabling verify_DNS_health

djsly:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


zachary-bailey (commented):
For the below line

`registry_domain_name="${MCR_REPOSITORY_BASE%/}"`

Is it possible the karpenter VHDs could have an unbound variable error? Should we comment this out too?

djsly:
> For the below line
> 
> `registry_domain_name="${MCR_REPOSITORY_BASE%/}"`
> 
> Is it possible the karpenter VHDs could have an unbound variable error? Should we comment this out too?
> 
> Just being overly cautious given our predicament

it will just set it to "" and we wont use it after

cameronmeissner:
should also be fine as long as we don't use it, I don't believe we set -u anywhere in provisioning scripts though I could be mistaken

zachary-bailey:
Yeah just looked through, I don't see set -anywhere.

PR #6009, opened 2025-03-07
chore: upgrade CSI driver versions for AKS 1.32

andyzhangx:
**What type of PR is this?**
/kind feature

chore: upgrade CSI driver versions for AKS 1.32

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -544,10 +544,6 @@
       "multiArchVersionsV2": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes-csi/azuredisk-csi",
-          "latestVersion": "v1.29.12"
-        },
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes-csi/azuredisk-csi",
           "latestVersion": "v1.30.9",
           "previousLatestVersion": "v1.30.8"
         },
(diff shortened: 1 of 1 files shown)


parts/common/components.json
```diff
@@ -648,6 +645,10 @@
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes-csi/livenessprobe",
           "latestVersion": "v2.14.0"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes-csi/livenessprobe",
+          "latestVersion": "v2.15.0"
```
timmy-wright, line 651: This means we'll be caching both 2.14 and 2.15 of liveness prob:
* Do we need both of them? They are 164MB each on windows
* If the 2.14.0 is only needed while RP is rolled out to use 2.15 (due to caching) then we should use "previousLatestVersion" for 2.14 (as opposed to a seperate block). This indicates to the next developer that they might be able to remove 2.14 rather than keep it.

parts/common/components.json
```diff
@@ -668,6 +669,10 @@
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes-csi/csi-node-driver-registrar",
           "latestVersion": "v2.12.0"
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes-csi/csi-node-driver-registrar",
+          "latestVersion": "v2.13.0"
```
timmy-wright, line 675: Same question as for liveness probe:

This means we'll be caching both 2.14 and 2.15 of liveness prob:

* Do we need both of them? They are 164MB each on windows
* If the 2.12.0 is only needed while RP is rolled out to use 2.13 (due to caching) then we should use "previousLatestVersion" for 2.12 (as opposed to a separate block). This indicates to the next developer that they might be able to remove 2.12 rather than keep it.

timmy-wright (approved):
LGTM - a couple of questions about liveness probe and csi-node-driver-registrar.

parts/common/components.json
andyzhangx: RP will use 2.15 on March release, and then I will remove liveness and node-driver-registrar old versions

PR #6014, opened 2025-03-07
Add ability to enable CDI in containerd.

abenn135:
**What type of PR is this?**

/kind feature
/kind api-change

This PR adds the ability to enable CDI in containerd. This is needed to support NVIDIA's DRA plugin, which will be used alongside upcoming hardware. 

Fixes # <none in this repo>

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


aks-node-controller/parser/templates/containerd.toml.gtpl
```diff
@@ -2,6 +2,10 @@ version = 2
 oom_score = -999{{if getHasDataDir .KubeletConfig}}
 root = "{{.KubeletConfig.GetContainerDataDir}}"{{- end}}
 [plugins."io.containerd.grpc.v1.cri"]
+  {{- if and ( .GetGpuConfig ) ( .GpuConfig.GetRequiresCdi ) }}
```
cameronmeissner, line 5: @abenn135 aks-node-controller isn't currently being used, for this change to immediately take effect you'd want to edit the containerd config template here: https://github.com/Azure/AgentBaker/blob/master/pkg/agent/baker.go#L1392 which AgentBaker then uses o template this variable: https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/cse_cmd.sh#L157

ganeshkumarashok:
I'd also recommend enabling it in an Agentbaker e2e

pkg/agent/baker.go
```diff
@@ -1411,7 +1418,10 @@ root = "{{GetDataDir}}"{{- end}}
 [plugins."io.containerd.cri.v1.images".registry.headers]
   X-Meta-Source-Client = ["azure/aks"]
 
-[plugins."io.containerd.cri.v1.runtime".containerd]
+{{ if IsCDIEnabled }}[plugins."io.containerd.cri.v1.runtime"]
+  enable_cdi = true
```
nojnhuh, line 1422: CDI is enabled by default in containerd 2.0, so the logic here will need to be inverted such that if CDI is not enabled, then the config needs to disable it. Maybe it's best to explicitly define `enable_cdi` in each case.

https://github.com/containerd/containerd/pull/9621

PR #6016, opened 2025-03-10
test: validate package URL root path for Azure China cloud

mainred:
**What type of PR is this?**
/kind test

This PR validates the packages installed in VHD can be downloadable from Azure China Cloud if their URL can be passed from control plane, and give hints to the next action when their proxy path is not defined the Azure China Cloud proxy server.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -180,9 +180,9 @@ testPackagesInstalled() {
       local extractedPackageDir
       extractedPackageDir="$downloadLocation/${fileNameWithoutExt}"
 
-      # Validate whether package exists in Azure China cloud.
+      # Validate whether package proxy path exists in Azure China cloud.
       if [[ $downloadURL == https://acs-mirror.azureedge.net/* ]]; then
-        testPackageURLInAzureChinaCloud "$downloadURL"
+        testPackageInAzureChinaCloud "$downloadURL"
       fi
 
       # if there is a directory with expected name, we assume it's been downloaded and extracted properly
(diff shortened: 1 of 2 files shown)


mainred:
cc @andyzhangx, you may be asked for help when the test fails.
- add the root path to `rootPathExceptions` if the packages is basically cached in VHD as the only source
- update the root path to `aks` or
- add the proxy_pass

mainred:
cc @timmy-wright @Devinwong

andyzhangx (approved):

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -180,6 +180,11 @@ testPackagesInstalled() {
       local extractedPackageDir
       extractedPackageDir="$downloadLocation/${fileNameWithoutExt}"
 
+      # Validate whether package exists in Azure China cloud.
```
Devinwong, line 183: Don't we have this check already? at line 224?
Is it different?

mainred: L224 and here is different. Here we check the proxy path exists in Azure China cloud, while L224 checks the file size.
Let me merge this two checks

Devinwong: Thanks for making the changes. One more question, if the check fails, meaning the proxy path does not exist in Azure China cloud, what is the next step to fix it?

Devinwong: I was asking this because if somehow it requires changes from the proxy codes, but it will end up blocking the AgentBaker build pipeline which seems doesn't make sense.

mainred: > One more question, if the check fails, meaning the proxy path does not exist in Azure China cloud, what is the next step to fix it?

I add the action item in the error message. The proxy sever is maintained by Andy Zhang, and we can 
- please use root path 'aks' since 'aks' is the default proxy path and has been added to the proxy server
- or contact 'andyzhangx' for help", to add the new proxy pass
- I don't suggest the user to edit the rootPathExceptions, and we want to at least make sure Andy Zhang has the context the new root path.

I can share some context of this issue. We add credential provider, though it's in MAR now, but we use Azure China Cloud mirror URL to specify the binary path, and as the root path is not added in the mirror, downloading the binary failed and then node cse failed. So we want to make sure every mirror root path is under control.

mainred: > I was asking this because if somehow it requires changes from the proxy codes, but it will end up blocking the AgentBaker build pipeline which seems doesn't make sense to me

root path is the same as proxy path, and is closely after the domain name of the 
like in this case, `https://acs-mirror.azureedge.net/cloud-provider-azure/v1.32.2/binaries/azure-acr-credential-provider-windows-amd64-v1.32.2.tar.gz` cloud-provider-azure is the root path and proxy path

updating the proxy, or adding a new proxy pass, is easy, and it's just a configuration change. Also, normally we don't have new root paths, 
We've already added the Azure China cloud file check, and I think the VHD is validating not only the   downloaded binary, but also the source of the binary, if the binaries of different cloud are from different sources, their size will be different.

> Is there already test on the proxy side that can catch this early?

Sadly, we don't have validation of the proxy side so far. because proxy server owner does not even know when the new root path is added by the producer.
migrating the artifacts to MAR is the long-term plan, and we are working on it.

Devinwong: Thanks for the detailed explanations. I will approve it. After merging to master, set's see if this will fail frequently. Otherwise it should be fine.

Devinwong (approved):

PR #6036, opened 2025-03-13
fix: restart nvidia-gridd to get correct License state for GRID drivers

ganeshkumarashok:
**What type of PR is this?**
/kind bug

This PR restarts nvidia-gridd process to get correct License state for GRID drivers. Currently, it gets into an Unlicensed state (can be checked with nvidia-smi). 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -793,6 +793,18 @@ configGPUDrivers() {
         echo "os $OS not supported at this time. skipping configGPUDrivers"
         exit 1
     fi
+    
+    if [[ "$NVIDIA_DRIVER_IMAGE" == *"grid"* ]]; then
+        if [ -x /usr/bin/nvidia-gridd ]; then
+            if nvidia-smi -q | grep -q "License Status.*Unlicensed"; then
+                echo "GRID license status is Unlicensed. Restarting nvidia-gridd..."
+                sudo pkill nvidia-gridd || true
+                sudo /usr/bin/nvidia-gridd &
```
timmy-wright, line 802: I'm surprised nvidia-gridd doesn't run as a system service. Should it be? That way it'll restart automatically.

ganeshkumarashok: Yes, I was thinking about that too. In fact, this code assumes there's a systemd unit: https://github.com/Azure/aks-gpu/blob/0adfc0c125107c12618153d916015326b7ef1247/install.sh#L77, but when i checked logs it doesn't have a systemd unit defined.

ganeshkumarashok: I also think it's worth defining it as a systemd unit in the future. This is a forced restart post kernel module reload. So we might need an explicit force restart even then.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -793,6 +793,18 @@ configGPUDrivers() {
         echo "os $OS not supported at this time. skipping configGPUDrivers"
         exit 1
     fi
+    
+    if [[ "$NVIDIA_DRIVER_IMAGE" == *"grid"* ]]; then
+        if [ -x /usr/bin/nvidia-gridd ]; then
+            if nvidia-smi -q | grep -q "License Status.*Unlicensed"; then
```
Devinwong, line 799: It is highly recommended avoid using nested if more than 2 levels. It will be hard to debug and do unit tests if any.
2 ways that I use to avoid such 3-level if are:
- return early instead of putting the logic into else
- put the additional if into a function
   - e.g. 
   ```
   if A
      if binary exist
        if command output
   ```
   can change to 
   ```
      if A
        execute_command_if_exists()
   ```
This is just an example. The logic may not be exactly correct. You could find your best way too. :)

timmy-wright: I agree - this would be more readable if it was extracted into a function called "restart_nvidia_gridd_if_unlicenses" then the function can have early exits if the if statements fail.

ganeshkumarashok: Thanks Devin and Tim, I've updated the PR this way.

timmy-wright (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
ganeshkumarashok: Is it more readable now?

Devinwong (approved):
Thanks for making the changes. This looks perfect. Approved

PR #6038, opened 2025-03-14
chore: fix issue in windows VHD prod pipeline

timmy-wright:
**What type of PR is this?**

/kind chore

Bug in the windows prod VHD pipeline causing the prod VHDs not to be build properly.

It was basically shell quoting on a command that set a pipeline variable.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -291,6 +291,12 @@ steps:
       SIG_FOR_PRODUCTION: $(SIG_FOR_PRODUCTION)
       OS_TYPE: "Windows"
 
+  - bash: |
+      echo "DRY_RUN: $(variables.DRY_RUN)"
+      echo "DRY_RUN 2: ${DRY_RUN}"
+      echo "SIG_FOR_PRODUCTION: $(variables.SIG_FOR_PRODUCTION)"
+      echo "SIG_FOR_PRODUCTION 2: ${SIG_FOR_PRODUCTION}"
+    displayName: Echo vars used for condition
 
   # Set VHD_NAME and SKU_NAME which will be published.
   # Note: use -a to grep OS_DISK_SAS (packer-output should be read as a binary file in Linux)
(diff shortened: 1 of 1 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -58,8 +58,8 @@ steps:
 
   # We create the variable dry run and the update it, depending on the pipeline and branch type.
   - bash: |
-      export DRY_RUN=${{ parameters.dryrun }}
-      echo '##vso[task.setvariable variable=DRY_RUN]${DRY_RUN}'
+      DRY_RUN=${{ parameters.dryrun }}
+      echo "##vso[task.setvariable variable=DRY_RUN]${DRY_RUN}"
```
timmy-wright, line 62: This is the fix - to let the shell expand the $DRY_RUN variable.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -88,40 +88,47 @@ steps:
   #     Task variable SIG_FOR_PRODUCTION is set to True and passed to the following steps.
   #     Built sig will be deleted because it has been converted to VHD, and thus not needed.
   - bash: |
+      set -e
```
timmy-wright, line 91: I also changed this script to be a strange single liner script to a normal script with a "set -e" to get it to exit if anything goes wrong.

Devinwong (approved):
lgtm

.pipelines/templates/.builder-release-template-windows.yaml
ganeshkumarashok: Is it possible to add an automated check like shell-check that detects such quote issues before a PR is merged?

ganeshkumarashok (approved):

.pipelines/templates/.builder-release-template-windows.yaml
timmy-wright: Maybe. I was about to google to find out if I can write unit tests for ADO pipelines. Will see if we can shell-check scripts embedded in the yaml too. 

I guess we really should extract some of this code into .sh files so it can be shell-checked.

PR #6040, opened 2025-03-17
chore: fix go.sum, fail build if it's not up to date.

timmy-wright:
**What type of PR is this?**

/kind regression

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -10,5 +10,7 @@ jobs:
       with:
         go-version: '1.22'
     - run: |
+        set -ex
+        go build -mod=readonly
         make test
       name: Run unit tests for go code in the repository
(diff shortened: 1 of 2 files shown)


.pipelines/templates/e2e-template.yaml
```diff
@@ -45,6 +45,9 @@ jobs:
           cd e2e
           mkdir -p bin
           GOBIN=`pwd`/bin/ go install github.com/jstemmer/go-junit-report/v2@latest
+          # Yes, we build first. That's because the exit code from "go test" below is eaten by the go-junit-report command. So if there are build problems
+          # then the tests pass. Bah.
+          go build -mod=readonly ./...
           go test -v -parallel 100 -timeout 90m 2>&1 | ./bin/go-junit-report -iocopy -set-exit-code -out "$(Build.SourcesDirectory)/e2e/report.xml"
```
djsly, line 51: wont go test rebuild again ? adding more build time ?

djsly: couldnt we instead separate the `go-junit-report` to consume from a local file instead of the pipe output ?

timmy-wright: I thought about having go-junit-report consume from a local file. The problem is:
* go test exits with a non-0 exit code if tests fail
* go test exits with a non-0 exit code if the build fails

And if tests fail, we want to run the go-junit-report. So we'd need additional logic to determine the cause of the failure.

In any case, the build is super quick. Downloading the dependencies is the slow bit, and that only happens once between the build and test.

cameronmeissner (approved):

PR #6041, opened 2025-03-17
feat: set 755 permissions for custom ca paths

UtheMan:
Adjusts permissions for `/opt/certs` directory used by custom ca feature to `755` 
We use `755` to match permissions used for `usr/local/share/ca-certficates` path, which is the path that is checked by the `update-ca-certificates` tool that updates the trust store. This is to stop anyone from being able to add a file to the path that installs custom CAs in the trust store when feature is used on the node.


bcho (approved):

pkg/agent/testdata/RawUbuntu/line70.sh
```diff
@@ -135,6 +135,7 @@ configureHTTPProxyCA() {
 
 configureCustomCaCertificate() {
     mkdir -p /opt/certs
+    chmod 1755 /opt/certs
```
cameronmeissner, line 138: do we put anything else in this dir?

UtheMan: As far as I know we only use this for custom ca feature. I checked RP/AB and it only appears in paths related to custom ca.

zachary-bailey (commented):
This dir is created in `pre-install-dependencies`, we could also look changing it there

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -52,7 +52,7 @@ systemctlEnableAndStart disk_queue 30 || exit 1
 capture_benchmark "${SCRIPT_NAME}_copy_packer_files_and_enable_logging"
 
 mkdir /opt/certs
-chmod 1666 /opt/certs
+chmod 1755 /opt/certs
```
UtheMan, line 55: @zachary-bailey  I modified this here as well, that should be enough right? Or is there some other place you were thinking of?

zachary-bailey: Yeah thats it!

pkg/agent/testdata/RawUbuntu/line70.sh
cameronmeissner: might be worth leaving a comment clarifying that we don't use this dir for anything else

cameronmeissner (approved):

zachary-bailey (approved):

pkg/agent/testdata/RawUbuntu/line70.sh
UtheMan: Sure, added

PR #6061, opened 2025-03-19
feat: add arm64 and fips to PR gate

zachary-bailey:
**What type of PR is this?**

/kind test

This adds arm64 and FIPs coverage to the PR gate for AgentBaker code changes.

This is accomplished by replacing the azurelinuxV3gen2 VHD with the azurelinuxV3arm64gen2fips VHD. This maintains coverage for azurelinuxV3gen2, while also adding coverage for Arm64 and FIPs.

Prevents changes from unexpectedly failing builds for certain features that don't have PR gate test coverage.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.pipelines/.vsts-vhd-builder.yaml
+++ b/.pipelines/.vsts-vhd-builder.yaml
@@ -87,7 +87,7 @@ stages:
         - template: ./templates/.builder-release-template.yaml
           parameters:
             artifactName: azurelinuxv2-gen2
-    - job: buildAzureLinuxV3gen2fips
+    - job: buildAzureLinuxV3ARM64gen2fips
       timeoutInMinutes: 180
       steps:
         - bash: |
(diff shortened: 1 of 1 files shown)


.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -87,29 +87,29 @@ stages:
         - template: ./templates/.builder-release-template.yaml
           parameters:
             artifactName: azurelinuxv2-gen2
-    - job: buildAzureLinuxV3gen2
+    - job: buildAzureLinuxV3ARM64gen2fips
```
djsly, line 90: why replacing instead of augmenting the scope ?

zachary-bailey: also it is only failing PR gate right now because of the arm64 issue from the other PR

djsly: we are loosing non fips and non arm64 right ?

zachary-bailey: Yes, but its still over the azurelinuxV3gen2 base image. I can add just the base back in as well though

djsly (approved):

PR #6062, opened 2025-03-19
feat: set SDNRemoteArpMacAddress for HNS during AzCNI setup

rbtr:
**What type of PR is this?**
/kind feature 

The SDNRemoteArpMacAddress needs to be set for HNS when AzureCNI is enabled in certain scenarios. We have done this in CNS for a very long time (https://github.com/Azure/azure-container-networking/pull/1306) but some recent changes (https://github.com/Azure/azure-container-networking/pull/2993) have exposed how very fragile HNS is and how dangerous it is to restart it unnecessarily (https://github.com/Azure/azure-container-networking/pull/3498).

This change will move setting the regkey to the CSE scripts so that it is complete and HNS is restarted well before CNS lands, and guarantees it only happens once for the lifetime of the Node.

**Release note**: 

```
set SDNRemoteArpMacAddress for HNS during AzCNI setup
```

### DIFF
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -161,14 +161,16 @@ function Set-AzureCNIConfig
     }
 
     # Set the SDNRemoteArpMacAddress RegKey for HNS when AzureCNI is enabled
+    Write-Log "Setting SDNRemoteArpMacAddress to 12-34-56-78-9a-bc for HNS"
     Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name SDNRemoteArpMacAddress -Value "12-34-56-78-9a-bc"
     # Restart hns service if it is exsting and running, to make the system regkey changes effective.
     $hnsServiceName = 'hns'
     $hnsService = Get-Service -Name $hnsServiceName -ErrorAction SilentlyContinue
     if ($hnsService -and $hnsService.Status -eq 'Running') {
-        Write-Log "hns service is already running. Restart hns."
+        Write-Log "HNS service is already running. Restart HNS."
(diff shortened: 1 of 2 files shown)


staging/cse/windows/azurecnifunc.ps1
```diff
@@ -167,6 +160,16 @@ function Set-AzureCNIConfig
         }
     }
 
+    # Set the SDNRemoteArpMacAddress RegKey for HNS when AzureCNI is enabled
```
tamilmani1989, line 163: will this script be invoked only once for VM lifetime?

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -167,6 +160,16 @@ function Set-AzureCNIConfig
         }
     }
 
+    # Set the SDNRemoteArpMacAddress RegKey for HNS when AzureCNI is enabled
+    Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name SDNRemoteArpMacAddress -Value "12-34-56-78-9a-bc"
```
tamilmani1989, line 164: can we add a log before/after setting reg key?

staging/cse/windows/azurecnifunc.ps1
rbtr: yes, it's the one that writes the conflist on Windows

tamilmani1989 (approved):

Devinwong:
@timmy-wright  to review since it's Windows related

timmy-wright (approved):
This is part of the windows scripts, which are seperate from the VHD and from the Custom Script Extension. So we'll have to build a new version of the windows scripts zip file, deploy it, then update the version of custom scripts used for node provisioning.

PR #6065, opened 2025-03-21
Add support kubelet config file in scritpless

Devinwong:
**What type of PR is this?**
/kind feature

Support kubelet config in scritpless.
Add unit tests
Add e2e tests.
Updated readme.

Fixes #

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


e2e/scenario_test.go
```diff
@@ -15,39 +15,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6"
 )
 
-func Test_AzureLinuxV2(t *testing.T) {
```
Devinwong, line 18: Remove this test as it's covered by `Test_AzureLinuxV2_KubeletCustomConfig` and save some test time.

e2e/scenario_test.go
```diff
@@ -15,39 +15,6 @@ import (
...
-			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
-			},
-			Validator: func(ctx context.Context, s *Scenario) {
-				ValidateInstalledPackageVersion(ctx, s, "moby-containerd", getExpectedPackageVersions("containerd", "mariner", "current")[0])
-			},
-		},
-	})
-}
-
-func Test_AzureLinuxV2_Scriptless(t *testing.T) {
```
Devinwong, line 33: Remove this test as it's covered by `Test_AzureLinuxV2_KubeletCustomConfig_Scpritless` and save some test time.

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
```diff
@@ -34,3 +34,444 @@ enum KubeletDisk {
...
+     ones the Kubelet initially started with, and this may disrupt your node.
+     Default: ""
+     +optional.*/
+  string static_pod_path = 3;
+
+  /* address is the IP address for the Kubelet to serve on (set to 0.0.0.0
+     for all interfaces).
+     Dynamic Kubelet Config (beta): If dynamically updating this field, consider that
+     it may disrupt components that interact with the Kubelet server.
+     Default: "0.0.0.0"
```
Devinwong, line 68: Default values will be added in next PR. Trying not to overwhelm this PR.

Devinwong:
> The latest Buf updates on your PR. Results from workflow [Buf CI / buf (pull_request)](https://github.com/Azure/AgentBaker/actions/runs/13981933068).
> 
> Build	Format	Lint	Breaking	Updated (UTC)
> `✅ passed`	`✅ passed`	`✅ passed`	`❌ failed (3)`	Mar 21, 2025, 12:23 AM

This is an intended breaking change but we are good now since it's not in production yet.

aks-node-controller/app.go
```diff
@@ -86,6 +86,8 @@ func (a *App) run(ctx context.Context, args []string) error {
 
 func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
 	inputJSON, err := os.ReadFile(flags.ProvisionConfig)
+	// log inputJSON
+	slog.Info("devin:inputJSON", "devin:inputJSON", string(inputJSON))
```
djsly, line 90: remove `devin`

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +608,38 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
...
+	if kubeletConfig == nil {
+		return ""
+	}
+	str := kubeletConfig.GetKubeletConfigFileConfig()
+	log.Printf("devin: kubeletConfig.GetKubeletConfigFileConfig() = %s", str)
+	configStringByte, err := protojson.MarshalOptions{
+		Multiline: true,
+		Indent:    "  ",
+	}.Marshal(str)
+	log.Printf("devin: configStringByte = %s", configStringByte)
```
djsly, line 626: more devin

aks-node-controller/parser/parser.go
```diff
@@ -191,7 +193,12 @@ func BuildCSECmd(ctx context.Context, config *aksnodeconfigv1.Configuration) (*e
 	// Convert to one-liner
 	triggerBootstrapScript = strings.ReplaceAll(triggerBootstrapScript, "\n", " ")
 	cmd := exec.CommandContext(ctx, "/bin/bash", "-c", triggerBootstrapScript)
+	//log config
+	slog.Info("devin: config: ", "config", config)
```
djsly, line 197: devin!

aks-node-controller/app.go
Devinwong: ya. Still debugging. Will remove unnecessary logs and headers

aks-node-controller/parser/parser.go
Devinwong: Logs with devin is for my temporary debugging. I caught a bug today so still working on it. Will remove it later.

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +609,68 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
 	return base64.StdEncoding.EncodeToString([]byte(authConfig.GetServicePrincipalSecret()))
 }
 
+func getKubeletFlags(kubeletConfig *aksnodeconfigv1.KubeletConfig) string {
+	return createSortedKeyValuePairs(kubeletConfig.GetKubeletFlags(), " ")
+}
+
+// pass the type of v as an input arguem
```
cameronmeissner, line 616: typo: argument

cameronmeissner: I also think we should disable godot linter...

aks-node-controller/app.go
cameronmeissner: do we need to log the input at all?

aks-node-controller/helpers/utils.go
```diff
@@ -5,17 +5,14 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
cameronmeissner, line 8: changed tab spacing?

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +609,68 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
...
+func marshalToJson(v any) ([]byte, error) {
+	// Originally we can set the Multiline here and it will marshal to a JSON we can use.
+	// However the protojson team intentionally randomly add extra whitespace after the key in the key-value.
+	// E.g., Sometimes it is "key": "value" and sometimes it is "key":  "value".
+	// They did it intentionally to make the output formot not reliable,
+	// because they think it is not a good idea to rely on the JSON output format.
+	// ref: https://github.com/protocolbuffers/protobuf-go/commit/582ab3de426ef0758666e018b422dd20390f7f26
```
cameronmeissner, line 623: this makes me mad :)

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +609,68 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
...
+		var rawMessage json.RawMessage = data
+		jsonByte, err := json.MarshalIndent(rawMessage, "", "  ")
+		if err != nil {
+			log.Printf("error marshalling kubelet config file content: %v", err)
+			return nil, err
+		}
+		jsonByte = append(jsonByte, '\n')
+		return jsonByte, nil
+	}
+	return nil, fmt.Errorf("unsupported type: %T", v)
```
cameronmeissner, line 646: nit: use default case for this?

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +609,68 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
...
+			return nil, err
+		}
+
+		var rawMessage json.RawMessage = data
+		jsonByte, err := json.MarshalIndent(rawMessage, "", "  ")
+		if err != nil {
+			log.Printf("error marshalling kubelet config file content: %v", err)
+			return nil, err
+		}
+		jsonByte = append(jsonByte, '\n')
```
cameronmeissner, line 643: why do we need the new line?

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +609,68 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
...
+	return nil, fmt.Errorf("unsupported type: %T", v)
+}
+
+// getKubeletConfigFileContent converts kubelet flags we set to a file, and return the json content.
+func getKubeletConfigFileContent(kubeletConfig *aksnodeconfigv1.KubeletConfig) string {
+	if kubeletConfig == nil {
+		return ""
+	}
+	kubeletConfigFileConfig := kubeletConfig.GetKubeletConfigFileConfig()
+	log.Printf("devin: kubeletConfig.GetKubeletConfigFileConfig() = %s", kubeletConfigFileConfig)
```
cameronmeissner, line 655: remove debugging code when you have a chance

aks-node-controller/parser/helper.go
```diff
@@ -613,6 +609,68 @@ func getServicePrincipalFileContent(authConfig *aksnodeconfigv1.AuthConfig) stri
...
+	if err != nil {
+		log.Printf("error marshalling kubelet config file content: %v", err)
+		return ""
+	}
+	log.Printf("devin: configStringByte = %s", kubeletConfigFileConfigByte)
+	return string(kubeletConfigFileConfigByte)
+}
+
+func getKubeletConfigFileContentBase64(kubeletConfig *aksnodeconfigv1.KubeletConfig) string {
+	if kubeletConfig == nil {
```
cameronmeissner, line 666: nit: no need for this if we check the same in getKubeletConfigFileContent?

aks-node-controller/parser/helper_test.go
```diff
@@ -17,14 +17,130 @@ limitations under the License.
...
+	" --pod-manifest-path=/etc/kubernetes/manifests" +
+	" --pod-max-pids=-1" +
+	" --protect-kernel-defaults=true" +
+	" --read-only-port=10255" +
+	" --resolv-conf=/etc/resolv.conf" +
+	" --rotate-certificates=true" +
+	" --rotate-server-certificates=true" +
+	" --streaming-connection-idle-timeout=4h0m0s" +
+	" --system-reserved=cpu=2,memory=1Gi" +
+	" --tls-cert-file=/etc/kubernetes/certs/kubeletserver.crt" +
```
cameronmeissner, line 58: hypernit: --tls-private-key-file would also have been passed by the bootstrapper if --tls-cert-file is there, thus I'd probably add it here too to make this more representative of production

cameronmeissner: fyi, the scripts embedded on the VHD will filter these flags out during CSE before starting kubelet when kubelet serving certificate rotation is enabled

cameronmeissner: oh I'm blind, it's already there :D

aks-node-controller/parser/parser.go
```diff
@@ -130,6 +131,7 @@ func getCSEEnv(config *aksnodeconfigv1.Configuration) map[string]string {
 		"PROXY_VARS":                                     getProxyVariables(config.GetHttpProxyConfig()),
 		"ENABLE_SECURE_TLS_BOOTSTRAPPING":                fmt.Sprintf("%v", getEnableSecureTLSBootstrap(config.GetBootstrappingConfig())),
 		"CUSTOM_SECURE_TLS_BOOTSTRAP_AAD_SERVER_APP_ID":  getCustomSecureTLSBootstrapAADServerAppID(config.GetBootstrappingConfig()),
+		"ENABLE_KUBELET_SERVING_CERTIFICATE_ROTATION":    fmt.Sprintf("%v", config.GetKubeletConfig().GetKubeletConfigFileConfig().GetServerTlsBootstrap()),
```
cameronmeissner, line 134: could we just check the output of GetKubeletConfig directly?

e2e/node_config.go
```diff
@@ -23,6 +23,87 @@ import (
...
+			"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
+			"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+			"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305",
+			"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+			"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305",
+			"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+			"TLS_RSA_WITH_AES_256_GCM_SHA384",
+			"TLS_RSA_WITH_AES_128_GCM_SHA256",
+		},
+		RotateCertificates: false,
```
cameronmeissner, line 59: are we override `RotateCertificates` in the node config mutation? we definitely need this to be true for any e2e test to work since we rely on kubelet client TLS bootstrapping for e2e nodes to join the cluster

cameronmeissner:
this is to add support for kubelet config _file_ to scriptless right? can we update the PR description?

aks-node-controller/parser/parser.go
Devinwong: removed

aks-node-controller/parser/helper.go
Devinwong: removed

aks-node-controller/app.go
Devinwong: no we don't. This is just for current debugging. Removed

aks-node-controller/parser/helper.go
Devinwong: updated!

aks-node-controller/helpers/utils.go
Devinwong: The linter did that, not me!
But I see it's all using tabs in this file and other files. So I think we should be good, right?

aks-node-controller/parser/helper.go
Devinwong: I tried to set a default case. But then the protojson.Marshal() will complain because it strictly needs a type that is defined from protobuf so that it can handle the json key naming correctly. If in the future, more protobuf defined types want to use this function, we will need to add it explicitly with 1 more line/case.

aks-node-controller/parser/helper.go
Devinwong: not needed. Removed!

aks-node-controller/parser/helper.go
Devinwong: removed all!

aks-node-controller/parser/helper.go
Devinwong: correct. Removed

aks-node-controller/parser/parser.go
Devinwong: sorry what do you mean?

aks-node-controller/parser/helper.go
Devinwong: Exactly. My unit tests sometimes fail and sometimes pass because of this. So hard to root cause this issue.

e2e/node_config.go
Devinwong: Are you referring to this part of code?
https://github.com/Azure/AgentBaker/blob/devinwon/kubelet_config/parts/linux/cloud-init/artifacts/cse_config.sh#L438
${ENABLE_KUBELET_SERVING_CERTIFICATE_ROTATION} should be true, right?
There are 2 variables look similar to me. One is `--rotate-certificates`, `--rotate-server-certificates`. So I'd like to clarify first.

cameronmeissner: --rotate-certificates allows the kubelet to use the bootstrap token (specified in bootstrap-kubeconfig) to bootstrap its _client_ certificate after startup, --rotate-server-certificates (named serverTLSBootstrap in kubelet config file) allows the kubelet to bootstrap its _serving_ certificate _after_ it has bootstrapped its client certificate - we always want to set --rotate-certificates to true, since currently AKS nodes rely on TLS bootstrapping to be able to join clusters (though this won't be the case once secure TLS bootstrapping is a thing...), though we don't necessarily always want to set --rotate-server-certificates to true, which is why ENABLE_KUBELET_SERVING_CERTIFICATE_ROTATION exists

cameronmeissner: my original comment was just to make sure that in scriptless/aks-node-controller abe2e's that we're setting --rotate-certificates to true, otherwise the nodes would never be able to join the e2e cluster

aks-node-controller/parser/parser.go
cameronmeissner: it just seems odd to me that we need to convert kubelet config to kubelet config file format in order to get the value of --rotate-server-certificates (serverTLSBootstrap), though not the end of the world

e2e/node_config.go
Devinwong: Thanks for clarifying. I did set "--rotate-certificates": "true" to the kubelet flag. But Let me also add it in Kubeletconfig to make sure they are aligned. Thanks for reminding me.

aks-node-controller/parser/parser.go
Devinwong: Ya I was also thinking moving all properties from `kubeletConfigFileConfig` to the previous level `KubeletConfig`.
But then I realized that there are some properties that should be part of `KubeletConfig` but not in `kubeletConfigFileConfig`. For example, `enable_kubelet_config_file`, `kubelet_flags`, etc. `enable_kubelet_config_file` is in neither file but only in our logic. So IMO it will be better to separate them. Does it sound making sound to you too?

```
message KubeletConfig {
  // A map of kubelet flags to their values. The map is sorted by key.
  map<string, string> kubelet_flags = 1;

  // A map of node labels to their values.
  map<string, string> kubelet_node_labels = 2;

  // The type of disk to use for the kubelet.
  KubeletDisk kubelet_disk_type = 3;

  // Whether kubelet config should be enabled.
  bool enable_kubelet_config_file = 4;

  // content of the kubelet config file.
  KubeletConfigFileConfig kubelet_config_file_config = 5;

  // Kubelet client private key
  string kubelet_client_key = 6;

  // The content of the kubelet client certificate file.
  string kubelet_client_cert_content = 7;

  // The path used to mount docker images, emptyDir volumes, and kubelet data.
  string container_data_dir = 8;
}
```

e2e/scenario_helpers_test.go
```diff
@@ -115,13 +116,30 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
...
-		s.Runtime.NBC = nbc
+		// deep copy the nbc so that we can mutate it without affecting the original
+		clonedNbc, err := deepcopy.Anything(nbc)
+		if err != nil {
+			s.T.Fatalf("failed to deep copy node config: %v", err)
+		}
+
+		// Pass the cloned nbc to BootstrapConfigMutator so that it can mutate the properties but not affecting the original one.
+		// Without this, it will cause a race condition when running multiple tests in parallel.
+		s.BootstrapConfigMutator(clonedNbc.(*datamodel.NodeBootstrappingConfiguration))
```
lilypan26, line 127: nit: could we just modify the mutator so that the config is passed by value instead of pass by reference? idk which would be better

Devinwong: This is a good call-out.
I tried to do that. Since it's a big struct so GO complained that following,
`call of s.AKSNodeConfigMutator copies lock value: github.com/Azure/agentbaker/aks-node-controller/pkg/gen/aksnodeconfig/v1.Configuration contains google.golang.org/protobuf/runtime/protoimpl.MessageState contains sync.Mutexcopylocks`
We often see this when we are copying a large struct.
Copilot still suggests us to do this, `Instead of dereferencing the pointer and passing the struct by value, you should pass the pointer itself to avoid copying the sync.Mutex.`

aks-node-controller/parser/parser.go
cameronmeissner: ah I see, we have `KubeletFlags`, on the kubelet config object, could we just do `GetKubeletConfig().GetKubeletFlags()["--rotate-server-certificates"]`?

cameronmeissner (approved):

lilypan26 (approved):

PR #6067, opened 2025-03-21
Add license and GRID driver validation, move nvidia systemd service files

ganeshkumarashok:
**What type of PR is this?**
/kind bug

- Add e2e validation for GRID driver license server state
- Validate status of nvidia-gridd systemd unit
- Update 535 GRID driver with container image that moves nvidia systemd service files to the right path.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner (approved):

e2e/validators.go
```diff
@@ -68,7 +68,21 @@ func ValidateNvidiaModProbeInstalled(ctx context.Context, s *Scenario) {
...
+
+func ValidateNvidiaGRIDLicenseValid(ctx context.Context, s *Scenario) {
+	command := []string{
+		"set -ex",
+		// Capture the license status output, or continue silently if not found
+		"license_status=$(sudo nvidia-smi -q | grep 'License Status' | grep 'Licensed' || true)",
+		// If the output is empty, print an error message and exit with a nonzero code
+		"if [ -z \"$license_status\" ]; then echo 'License status not valid or not found'; exit 1; fi",
+		// Check that nvidia-gridd is active by capturing its is-active output
+		"active_status=$(sudo systemctl is-active nvidia-gridd)",
```
ganeshkumarashok, line 82: Considered adding a similar check for nvidia-topologyd but it's not clear which GPUs and driver versions should have it. GRID license should always be valid though.

e2e/scenario_test.go
```diff
@@ -1039,7 +1039,7 @@ func Test_Ubuntu2204_GPUA100(t *testing.T) {
 }
 
 func Test_Ubuntu2204_GPUA10(t *testing.T) {
-	runScenarioUbuntu2204GPU(t, "Standard_NV6ads_A10_v5")
```
djsly, line 1042: are we still using this Scenario ?

e2e/scenario_test.go
```diff
@@ -1071,6 +1071,35 @@ func runScenarioUbuntu2204GPU(t *testing.T, vmSize string) {
 	})
 }
 
+func runScenarioUbuntuGRID(t *testing.T, vmSize string) {
```
djsly, line 1074: should we keep 2204 in the ScenarioName?  what if we want to test with 2404 to ensure that it's working with the kernel 6.8 ?

e2e/validators.go
djsly: what about persisted ?

e2e/validators.go
```diff
@@ -68,7 +68,21 @@ func ValidateNvidiaModProbeInstalled(ctx context.Context, s *Scenario) {
...
 	}
-	execScriptOnVMForScenarioValidateExitCode(ctx, s, strings.Join(command, "\n"), 0, "cound not execute nvidia-modprobe command")
+	execScriptOnVMForScenarioValidateExitCode(ctx, s, strings.Join(command, "\n"), 0, "could not execute nvidia-modprobe command")
+}
+
+func ValidateNvidiaGRIDLicenseValid(ctx context.Context, s *Scenario) {
+	command := []string{
+		"set -ex",
+		// Capture the license status output, or continue silently if not found
+		"license_status=$(sudo nvidia-smi -q | grep 'License Status' | grep 'Licensed' || true)",
```
djsly, line 78: this will always be valid for the first 20min no ?

djsly (approved):

e2e/scenario_test.go
ganeshkumarashok: yes, for CUDA

e2e/validators.go
ganeshkumarashok: No, it won't necessarily be valid then -- workload will work for the first 20 minutes but after that, if it remains in an unlicensed state, it will get degraded.

e2e/validators.go
ganeshkumarashok: persistenced is not one of the files that got moved, but we could check it too

e2e/scenario_test.go
ganeshkumarashok: We should be checking both 22.04 and 24.04 in that case. How are we switching between the two OS versions?

djsly: this line ? VHD:     config.VHDUbuntu2204Gen2Containerd,

ganeshkumarashok: Thanks, added another test for 24.04

ganeshkumarashok: Removed that test since there were other failures.

PR #6076, opened 2025-03-24
chore: bump crictl-tools

AlisonB319:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -1184,7 +1184,7 @@
             "versionsV2": [
               {
                 "renovateTag": "<DO_NOT_UPDATE>",
-                "latestVersion": "1.31.1"
+                "latestVersion": "1.32.0"
               }
             ],
             "downloadURL": "https://acs-mirror.azureedge.net/cri-tools/v${version}/binaries/crictl-v${version}-linux-${CPU_ARCH}.tar.gz"


cameronmeissner:
lgtm pending upstream build

AlisonB319:
fyi 
usr/local/bin/crictl,filesystem,["CVE-2024-34156"]
is still present in the build

aks-node-controller/Makefile
```diff
@@ -6,7 +6,7 @@ BUF = docker run --volume "$(CURDIR)/../:$(CURDIR)/../" --workdir $(CURDIR) bufb
 proto-generate:
 	@($(BUF) format -w)
 	rm -rf pkg/gen/aksnodeconfig/v1
-	docker build --platform $(shell uname -m) -t protoc-docker - < protoc.Dockerfile
+	docker build --platform linux/amd64 -t protoc-docker - < protoc.Dockerfile
```
Devinwong, line 9: Why do we need this change? I am not sure if they are equivalent because when I run this command on my WSL, I get a different result.
```
root@DESKTOP-1TJD8FB:~/git/AgentBaker/bugbash/proxytest# uname -m
x86_64
```

aks-node-controller/proto/aksnodeconfig/v1/config.proto
```diff
@@ -150,4 +150,7 @@ message Configuration {
 
   // IMDS restriction configuration
   ImdsRestrictionConfig imds_restriction_config = 39;
+
+  // Kubelet Config File Content
+  string kubelet_config_file_content = 40;
```
Devinwong, line 155: We do have this in the KubeletConfig. If the PR gate tests failed, please rebase with master and we shouldn't need to add this.

aks-node-controller/Makefile
Devinwong: oh is it because the Mac Shell doesn't have the uname command?

AlisonB319: When I was running `make proto-generate` it would fail due my mac being arm64. Seems that 
`FROM mcr.microsoft.com/azurelinux/base/core:3.0` from protc.Dockerfile requires linux/amd64? So I think it makes sense to explicitly set that? 

If you run `make proto-generate` does it still work for you for x86_64?

Also I was relying DIYing geting the changes to scriptless (since it seems like there wasn't a way to update the code to run the e2e's manually?) So I might have not have followed the correct flow

aks-node-controller/proto/aksnodeconfig/v1/config.proto
AlisonB319: When was this added? I rebased master last night and the e2e's were failing with 

```
----------------------------------- begin stdout -----------------------------------,
{"time":"2025-03-25T09:46:51.27364042Z","level":"INFO","msg":"aks-node-controller started"}
{"time":"2025-03-25T09:47:05.397629699Z","level":"INFO","msg":"aks-node-controller started"}
{"time":"2025-03-25T09:47:05.428495257Z","level":"ERROR","msg":"aks-node-controller failed","error":"unmarshal provision config: proto: (line 1:2626): unknown field \"kubelet_config_file_content\""}

----------------------------------- end stdout ------------------------------------
```

AlisonB319: re-running now on a different PR with a rebase to cover my bases 🥁

PR #6081, opened 2025-03-25
Pahealy/chrony config hyperv

### LINKED ISSUE #5991: Wrong chrony stratum (doc has updated, AgentBaker has not)
**What happened**:
A lot of files have
```
refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
```

**What you expected to happen**:
```
refclock PHC /dev/ptp_hyperv poll 3 dpoll -2 offset 0 stratum 2
```
like in current version of https://learn.microsoft.com/en-us/azure/virtual-machines/linux/time-sync should be instead present. Or if you don't want ptp_hyperv, ptp0 is okay; what matters is `stratum 2`

**How to reproduce it**:
Use grep on the existing source files.

Wrong stratum leads to user confusion in https://www.mail-archive.com/chrony-users@chrony.tuxfamily.org/msg03770.html.

**Environment**:

- AgentBaker version: HEAD
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

phealy:
**What type of PR is this?**
/kind cleanup

Updates the ubuntu chrony configuration to match the [recommended Azure configuration](https://learn.microsoft.com/en-us/azure/virtual-machines/linux/time-sync#check-for-ptp-clock-source), which uses /dev/ptp_hyperv instead of /dev/ptp0 to avoid possible errors when the PTP source from the mellanox AccelNet NIC is enumerated before the hyperv source.

Also adds a udev rule to put the /dev/ptp_hyperv symlink in place if the symlink isn't already there; this is in place on modern distros (Azure Linux 2/3, Ubuntu 19.10+) and thus shouldn't be needed in most cases, but I wanted to include it for safety's sake so as not to break chrony if it's missing somehow.

Fixes #5991 

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```
Updated chrony configuration to use /dev/ptp_hyperv at stratum 2 per https://learn.microsoft.com/en-us/azure/virtual-machines/linux/time-sync
```

### DIFF
--- a/vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
+++ b/vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
@@ -15,6 +15,7 @@ ERR_STOP_OR_DISABLE_SYSTEMD_TIMESYNCD_TIMEOUT=12 {{/* Timeout waiting for system
 ERR_STOP_OR_DISABLE_NTP_TIMEOUT=13 {{/* Timeout waiting for ntp stop */}}
 ERR_CHRONY_INSTALL_TIMEOUT=14 {{/*Unable to install CHRONY */}}
 ERR_CHRONY_START_TIMEOUT=15 {{/* Unable to start CHRONY */}}
+ERR_UDEV_RULE_RELOAD=16 {{/* Unable to reload udev after ptp_hyperv rule install */}}
 
 
 echo "Sourcing tool_installs_ubuntu.sh"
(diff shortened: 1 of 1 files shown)


vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -198,7 +213,7 @@ maxupdateskew 100.0
 rtcsync
 
 # Settings come from: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync
-refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0
+refclock PHC /dev/ptp_hyperv poll 3 dpoll -2 offset 0 stratum 2
```
djsly, line 216: we dont need to change anything for azurelinux ?

djsly (approved):

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
cameronmeissner: from PR description sounds like this wouldn't ever be needed on AzureLinux since we only build V2/V3, and we've stopped building CBLMarinerV1 sometime ago

PR #6087, opened 2025-03-26
bug: fix get kubelet identity not retrying when imds has temp 400 response

fseldow:
**What type of PR is this?**

/kind bug

We meet issue that imds will sometimes return 400 bad request. It will cause the cse directly exit without any more information. If it is VMSS cluster, the next instance will succeed, however, if it is vms, the cluster will fail 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -253,14 +253,18 @@ retrycmd_get_tarball_from_registry_with_oras() {
...
+        ACCESS_TOKEN_OUTPUT=$(echo "$response" | sed '$d')
+        http_code=$(echo "$response" | tail -n1)
+        if [ -n "$ACCESS_TOKEN_OUTPUT" ] && [ "$http_code" -eq 200 ]; then 
             echo "$ACCESS_TOKEN_OUTPUT"
             return 0
         fi
         sleep $wait_sleep
     done
-    return $ERR_ORAS_PULL_NETWORK_TIMEOUT
+    echo "connect to IMDS to get access token timeout"
```
djsly, line 265: are we sure it's a timeout here ?  could it be another type of failure ? if you get 400 or 404 for all retries, this wound't be considered a timeout, I think this error message could cause confusion.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -253,14 +253,18 @@ retrycmd_get_tarball_from_registry_with_oras() {
...
+    http_code=${http_code:-$ERR_ORAS_IMDS_TIMEOUT}
```
djsly, line 266: you are. mixing HTTP error code with CSE error code here, this could also be confusing,  seeing 210 I would expect this is a valid HTTP_CODE.  I would log something more obvious.  If http_code is set, we log a message saying fail to reach IMDS after after X retires due to error code XYZand we return a CSE ERROR for CURL FAILURE,  if there is no error code, we log a TIMEOUT ERROR and we return a CSE ERROR for TIMEOUT.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -253,14 +253,21 @@ retrycmd_get_tarball_from_registry_with_oras() {
 retrycmd_get_access_token_for_oras() {
     retries=$1; wait_sleep=$2; url=$3
     for i in $(seq 1 $retries); do
-        ACCESS_TOKEN_OUTPUT=$(timeout 60 curl -v -s -H "Metadata:true" --noproxy "*" "$url")
-        if [ -n "$ACCESS_TOKEN_OUTPUT" ]; then 
+        response=$(timeout 60 curl -v -s -H "Metadata:true" --noproxy "*" "$url" -w "\n%{http_code}")
```
cameronmeissner, line 256: if it's a 4xx error, doesn't IMDS return a json object detailing the error details? I know there's an error_description field you can use to get more detailed error info

I'd probably opt for using `jq` to parse out error_description from the response json and check if it's empty or not, if it's non-empty it should indicate an error

cameronmeissner: on second thought, might be worth following the published retry guidance a little more closely: https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance

fseldow: If it is 400 error, it will return json with error details. However, unsure about other 4xx errors. e.g. in acr refresh token part (not IMDS), if it is 401, the boject is empty

About retry time, i extend it for standalone test, will revert it back to guidance value. Sorry for confusion

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -810,16 +817,12 @@ oras_login_with_kubelet_identity() {
 
     set +x 
     access_url="http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/&client_id=$client_id"
-    raw_access_token=$(retrycmd_get_access_token_for_oras 10 5 $access_url)
+    raw_access_token=$(retrycmd_get_access_token_for_oras 5 15 $access_url)
     ret_code=$? 
-    if [ $ret_code -ne 0 ]; then
-        echo "failed to retrieve access token: $ret_code"
+    if [ $ret_code -ne 0 ] || [[ "$raw_access_token" == *"error"* ]]; then
```
UtheMan, line 822: Could you add a comment explaining the expected format of the output here and why we need to check both return code and contents of $raw_access_token?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -253,14 +253,21 @@ retrycmd_get_tarball_from_registry_with_oras() {
...
+        if [ -n "$ACCESS_TOKEN_OUTPUT" ] && [ "$http_code" -eq 200 ]; then 
             echo "$ACCESS_TOKEN_OUTPUT"
             return 0
         fi
         sleep $wait_sleep
     done
-    return $ERR_ORAS_PULL_NETWORK_TIMEOUT
+    if [ -n "$http_code" ]; then
+        echo "failed to retrieve kubelet identity token from IMDS, http code: $http_code, msg: $ACCESS_TOKEN_OUTPUT"
+        return $ERR_ORAS_PULL_UNAUTHORIZED
```
UtheMan, line 267: Would http_code != 200 always mean authorization issue? Not sure whether it would make sense to add more exit codes here, since we also log details about the error in line above

UtheMan (approved):
LGTM, left some questions/nit comments

parts/linux/cloud-init/artifacts/cse_helpers.sh
fseldow: it is not necessary, the second condition is only to keep the old logic to make the pr change as small as possible

parts/linux/cloud-init/artifacts/cse_helpers.sh
fseldow: this part is related to retrieve kubelet identity token to non-anonymous access to acr. Thus whatever the error is, it will result the cse unauthorized to access acr.

Not sure if we need more exit codes here. The sub code could be, as i observed: 
identity not found (400)
url is invalid (400)
imds connection timeout (imds endpoint issue)

I think we can leave one exit code for now to check feedback from customer/support to check if they need that many exit code. WDYT?

parts/linux/cloud-init/artifacts/cse_helpers.sh
fseldow: removed

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: Makes sense, let's leave this as is

PR #6088, opened 2025-03-26
fix: migrate to packages.aks.azure.com for package downloads

zachary-bailey:
**What type of PR is this?**

/kind feat

This PR adds a new package download URL in components.json and adds logic to determine which URL the node should use, based on available network connectivity.

The `resolve_packages_source_url` function is added in order to ensure connectivity to `packages.aks.azure.com` at the beginning of VHD builds and node provisioning. If connectivity cannot be established, the function will fall back to the previous URL. This will allow customers a grace period until they add the new URL into their allowlist. 

Future deprecation of `acs-mirror.azureedge.net`.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


.github/README-RENOVATE.md
```diff
@@ -364,7 +364,7 @@ As of 01/23/2025,
 - All the container images are onboarded to Renovate for auto-update.
 - PMC hosted packages, namely `runc` and `containerd`, are configured as auto-merge patch version.
 - OCI artifacts hosted on MAR(aka MCR) such as `kubernetes-binaries`, `azure-acr-credential-provider` and `containerd-wasm-shims` are onboarded for auto-update.
-- Acs-mirror hosted packages/binaries, namely `cni-plugins`, `azure-cni`, `cri-tools`, etc., are NOT onboarded for auto-update yet. There are plans to move the acs-mirror hosted packages to MCR OCI which will be downloaded by Oras. We will wait for this transition to be completed to understand the details how to manage them.
+- packages.aks.azure.com hosted packages/binaries, namely `cni-plugins`, `azure-cni`, `cri-tools`, etc., are NOT onboarded for auto-update yet. There are plans to move the packages.aks.azure.com hosted packages to MCR OCI which will be downloaded by Oras. We will wait for this transition to be completed to understand the details how to manage them.
```
djsly, line 367: @Devinwong  can you work with Zach here to update this text to be more accurate ?

parts/common/components.json
```diff
@@ -1238,7 +1252,7 @@
               }
             ],
             "downloadURL": "mcr.microsoft.com/oss/binaries/kubernetes/kubernetes-node:${version}-linux-${CPU_ARCH}",
-            "windowsDownloadURL": "https://acs-mirror.azureedge.net/kubernetes/${version}/windowszip/${version}-1int.zip"
+            "windowsDownloadURL": "https://${PACKAGE_DOWNLOAD_BASE_URL}/kubernetes/${version}/windowszip/${version}-1int.zip"
```
djsly, line 1255: shouldnt we ignore windows for now ?

parts/linux/cloud-init/artifacts/aks-check-network.sh
```diff
@@ -100,7 +100,7 @@ function check_and_curl {
         if [ $response -ge 200 ] && [ $response -lt 400 ]; then
             logs_to_events "AKS.testingTraffic.success" "echo '$(date) - SUCCESS: Successfully tested $url with returned status code $response'"
             break
-        elif [ $response -eq 400 ] && { [ "$url" == "acs-mirror.azureedge.net" ] || [ "$url" == "eastus.data.mcr.microsoft.com" ]; }; then
+        elif [ $response -eq 400 ] && { [ "$url" == "https://packages.aks.azure.com" ] || [ "$url" == "eastus.data.mcr.microsoft.com" ]; }; then
```
djsly, line 103: do we need `https://` ?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -694,16 +694,29 @@ updatePackageDownloadURL() {
     updateRelease "${package}" "${os}" "${osVersion}"
     local osLowerCase=$(echo "${os}" | tr '[:upper:]' '[:lower:]')
     
-    #if .downloadURIs.${osLowerCase} exist, then get the downloadURL from there.
+    #if .downloadURIs.${osLowerCase} exist, then get the downloadURL/packagesURL from there.
     #otherwise get the downloadURL from .downloadURIs.default 
     if [[ $(echo "${package}" | jq ".downloadURIs.${osLowerCase}") != "null" ]]; then
-        downloadURL=$(echo "${package}" | jq ".downloadURIs.${osLowerCase}.${RELEASE}.downloadURL" -r)
+        if [[ $(echo "${package}" | jq ".downloadURIs.${osLowerCase}.${RELEASE}.packagesURL") != "null" ]]; then
```
djsly, line 700: why do we need to store both URLs in the components.json ? I'm not a fan of introducing a new `duplicate` property in the Schema :(

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -795,6 +808,35 @@ verify_DNS_health(){
...
 }
 
+resolve_packages_source_url() {
+    echo "Ensuring connectivity to packages.aks.azure.com..."
+
+    local retries=5
+    local wait_sleep=1
+
+    for i in $(seq 1 $retries); do
+      response_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 5 https://packages.aks.azure.com/acs-mirror/healthz)
```
djsly, line 818: do we need --no-proxy ? I know some curl command in the CSE scripts are using that property.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -795,6 +808,35 @@ verify_DNS_health(){
...
+        if [ $i -eq $retries ]; then
+          echo "Executed curl to packages.aks.azure.com $i times. Response code is $response_code"
+          break
+        else
+          sleep $wait_sleep
+        fi
+      fi
+    done
+
+    if [ ${response_code} -eq 200 ]; then
```
djsly, line 831: this if/else statement isn't needed you can add the VAR assignment in the if above with the `echo`. Right before both `break` statements

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -470,7 +474,11 @@ installCNI() {
         echo "WARNING: no cni-plugins components present falling back to hard coded download of 1.4.1. This should error eventually" 
         # could we fail if not Ubuntu2204Gen2ContainerdPrivateKubePkg vhd? Are there others?
         # definitely not handling arm here.
-        retrycmd_get_tarball 120 5 "${CNI_DOWNLOADS_DIR}/refcni.tar.gz" "https://acs-mirror.azureedge.net/cni-plugins/v1.4.1/binaries/cni-plugins-linux-amd64-v1.4.1.tgz" || exit
+        retrycmd_get_tarball 120 5 "${CNI_DOWNLOADS_DIR}/refcni.tar.gz" "https://packages.aks.azure.com/cni-plugins/v1.4.1/binaries/cni-plugins-linux-amd64-v1.4.1.tgz"
```
djsly, line 477: why not aligned the logic to be the same as for `VNET_CNI_PLUGINS_URL`

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -470,7 +474,11 @@ installCNI() {
         echo "WARNING: no cni-plugins components present falling back to hard coded download of 1.4.1. This should error eventually" 
         # could we fail if not Ubuntu2204Gen2ContainerdPrivateKubePkg vhd? Are there others?
         # definitely not handling arm here.
```
djsly, line 476: Can we add a log_to_event with a pattern for tracing the URL used to download the package

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -379,6 +379,10 @@ downloadAzureCNI() {
         echo "VNET_CNI_PLUGINS_URL is not set. Exiting..."
         return
     fi
+
+    if [ $PACKAGE_DOWNLOAD_BASE_URL == "packages.aks.azure.com" ]; then
+        VNET_CNI_PLUGINS_URL="${VNET_CNI_PLUGINS_URL//"acs-mirror.azureedge.net"/$PACKAGE_DOWNLOAD_BASE_URL}"
```
djsly, line 384: Can we add a log_to_event with a pattern for tracing the URL used to download the package

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -633,6 +641,9 @@ installKubeletKubectlAndKubeProxy() {
 
             #TODO: remove the condition check on KUBE_BINARY_URL once RP change is released
             elif (($(echo ${KUBERNETES_VERSION} | cut -d"." -f2) >= 17)) && [ -n "${KUBE_BINARY_URL}" ]; then
+                if [  $PACKAGE_DOWNLOAD_BASE_URL == "packages.aks.azure.com" ]; then
+                    KUBE_BINARY_URL="${KUBE_BINARY_URL//"acs-mirror.azureedge.net"/$PACKAGE_DOWNLOAD_BASE_URL}"
+                fi
                 logs_to_events "AKS.CSE.installKubeletKubectlAndKubeProxy.extractKubeBinaries" extractKubeBinaries ${KUBERNETES_VERSION} ${KUBE_BINARY_URL} false
```
djsly, line 647: can we ensure that the URL gets printed properly ? this could be a clean way to track all usage of URLs

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -154,12 +155,12 @@ logs_to_events "AKS.CSE.installNetworkPlugin" installNetworkPlugin
 if [ "${IS_KRUSTLET}" == "true" ]; then
     versionsWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.versionsV2[].latestVersion' "$COMPONENTS_FILEPATH")
     downloadLocationWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadLocation' "$COMPONENTS_FILEPATH")
-    downloadURLWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.downloadURL' "$COMPONENTS_FILEPATH")
+    downloadURLWasm=$(jq -r '.Packages[] | select(.name == "containerd-wasm-shims") | .downloadURIs.default.current.packagesURL' "$COMPONENTS_FILEPATH")
```
djsly, line 158: still not sure why we are using a new property in the schema ? I dont see a reason why we need to keep using .downloadURL = acs-mirror

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1286,6 +1287,29 @@ testCoreDnsBinaryExtractedAndCached() {
...
+
+  echo "$test: Start"
+
+  resolve_packages_source_url
+  if [ "$PACKAGE_DOWNLOAD_BASE_URL" != "packages.aks.azure.com" ]; then
+    echo "PACKAGE_DOWNLOAD_BASE_URL was not set to packages.aks.azure.com"
+    err "$test: failed to set PACKAGE_DOWNLOAD_BASE_URL to packages.aks.azure.com"
+  fi
+  
+  # Block the IP on local vm to simulate cluster firewall blocking packages.aks.azure.com and retry test to see output
```
djsly, line 1301: this works with FIPs in the end ? :D

djsly (commented):
left some inlines comments

parts/linux/cloud-init/artifacts/aks-check-network.sh
zachary-bailey: Will fix

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: After speaking with @Devinwong, he said there could be compatibility issues if we remove a key from the document.

Is there a case where an older VHD will reference the most up to date components.json? Because if so, they would not have the PACKAGE_DOWNLOAD_BASE_URL defined during CSE

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: I can add it if you would like. It is working fine in current e2es, but if it will handle any edge cases that could come up then I will add

parts/linux/cloud-init/artifacts/cse_install.sh
zachary-bailey: Yes I can add an echo statement here

vhdbuilder/packer/test/linux-vhd-content-test.sh
zachary-bailey: Yes this does work with FIPs!

parts/common/components.json
zachary-bailey: Oops, I did remove the references I made in components.json

This is a URL that is currently in use @djsly

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: Fixed! Thanks

.github/README-RENOVATE.md
Devinwong: @zachary-bailey  could you update with this? =>
For Linux, the three binaries `cni-plugins`, `azure-cni`, `cri-tools` are not onboarded for auto-update yet. The team have been working on it. `cri-tools` package will soon be available, then `cni-plugins` and `azure-cni`.

parts/common/components.json
```diff
@@ -857,7 +857,7 @@
                 "latestVersion": "1.32.3"
               }
             ],
-            "downloadURL": "https://acs-mirror.azureedge.net/cloud-provider-azure/v${version}/binaries/azure-acr-credential-provider-windows-amd64-v${version}.tar.gz"
```
Devinwong, line 860: Are we sure windows binaries are also migrating from acs-mirror to packages.aks.azure.com? We will need to make sure @timmy-wright  is aware of this too.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -802,6 +802,40 @@ verify_DNS_health(){
...
+    for i in $(seq 1 $retries); do
+      response_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 5 --noproxy "*" https://packages.aks.azure.com/acs-mirror/healthz)
+      if [ ${response_code} -eq 200 ]; then
+        PACKAGE_DOWNLOAD_BASE_URL="packages.aks.azure.com"
+        echo "Setting PACKAGE_DOWNLOAD_BASE_URL to $PACKAGE_DOWNLOAD_BASE_URL."
+        break
+      else
+        if [ $i -eq $retries ]; then
+          PACKAGE_DOWNLOAD_BASE_URL="acs-mirror.azureedge.net"
+          echo "Setting PACKAGE_DOWNLOAD_BASE_URL to $PACKAGE_DOWNLOAD_BASE_URL. Please check to ensure cluster firewall has packages.aks.azure.com on its allowlist"
```
Devinwong, line 818: In this case `PACKAGE_DOWNLOAD_BASE_URL` is set to acs-mirror. Why do we ask user to set firewall with packages.aks.azure.com?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -802,6 +802,40 @@ verify_DNS_health(){
...
+          sleep $wait_sleep
+        fi
+      fi
+    done
+}
+
+update_base_url() {
+  initial_url=$1
+  
+  if [ $PACKAGE_DOWNLOAD_BASE_URL == "packages.aks.azure.com" ] && [[ $initial_url == *"acs-mirror.azureedge.net"* ]]; then
```
Devinwong, line 830: [[ ]] for consistency?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -802,6 +802,40 @@ verify_DNS_health(){
     echo "DNS health check passed"
 }
 
+resolve_packages_source_url() {
```
Devinwong, line 805: It will be great if you can write the shellspec unit test for this function. You can find many examples in cse_helpers_spec.sh

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -802,6 +802,40 @@ verify_DNS_health(){
...
+          echo "Setting PACKAGE_DOWNLOAD_BASE_URL to $PACKAGE_DOWNLOAD_BASE_URL. Please check to ensure cluster firewall has packages.aks.azure.com on its allowlist"
+          break
+        else
+          sleep $wait_sleep
+        fi
+      fi
+    done
+}
+
+update_base_url() {
```
Devinwong, line 827: It will be great if you can write the shellspec unit test for this function. You can find many examples in cse_helpers_spec.sh

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -802,6 +802,40 @@ verify_DNS_health(){
...
 }
 
+resolve_packages_source_url() {
+    local retries=5
+    local wait_sleep=1
+
+    for i in $(seq 1 $retries); do
+      response_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 5 --noproxy "*" https://packages.aks.azure.com/acs-mirror/healthz)
+      if [ ${response_code} -eq 200 ]; then
+        PACKAGE_DOWNLOAD_BASE_URL="packages.aks.azure.com"
```
Devinwong, line 812: It seems to me `PACKAGE_DOWNLOAD_BASE_URL` is only a local variable in this function but you are expecting it a global variable. (I could be wrong). 
With unit test, you can ensure that it's not unset out of the scope of this function.

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: There are also steps in the root's readme showing how to run the test

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: this is actually correct, we shoudn't be using [[ ]] unless we explicitly need to do a regex match, otherwise our stuff will not be nearly as bash-exportable as it can/should be

if possible I'd love to add something to shellcheck to ensure that we're using [ ] over [[ ]] _except_ for regex matching

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -156,6 +156,8 @@ downloadCredentialProvider() {
     if [[ -n $cred_provider_url ]]; then
         # CREDENTIAL_PROVIDER_DOWNLOAD_URL is passed in through install-dependencies.sh
         CREDENTIAL_PROVIDER_DOWNLOAD_URL=$cred_provider_url
+        logs_to_events "AKS.CSE.logDownloadURL" "echo $CREDENTIAL_PROVIDER_DOWNLOAD_URL"
```
cameronmeissner, line 159: this seems unnecessary to me if all `logDownloadURL` is doing is calling `echo` once

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner: not sure I understand the point here, logs_to_events should really only be used to gather latency telemetry for individual CSE steps, not tracing variable values

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1296,6 +1297,29 @@ testCorednsBinaryExtractedAndCached() {
...
+  echo "127.0.0.1     packages.aks.azure.com" | sudo tee /etc/hosts > /dev/null
```
cameronmeissner, line 1312: would this make more sense as an abe2e?

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: Found this from copilot.
image

@cameronmeissner do you mean if possible, we should use [ ], unless we really need the feature from [[ ]] (regex is 1 feature)?
Not sure if shellsheck has this check or maybe it's already suppressed. Need to find out the rule to enable it, if there is one.

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: The desired value for `PACKAGE_DOWNLOAD_BASE_URL` is packages.aks.azure.com. If it is set to acs-mirror.azureedge.net, that means that the node was not able to establish connectivity to packages.aks.azure.com. For this reason, we echo out this message so that in debugging the user / dev on call can see the customer needs to allow packages.aks.azure.com in their firewall allowlist. @Devinwong

parts/linux/cloud-init/artifacts/cse_install.sh
zachary-bailey: @djsly had requested this so that for every binary download, we know which URL is being passed to the nodes. after logging what the initial value is, we validate that it is what we want with the `update_base_url` function which makes substitutions based on the value of `PACKAGE_DOWNLOAD_BASE_URL` and the initial value of the URL

vhdbuilder/packer/test/linux-vhd-content-test.sh
zachary-bailey: I was originally planning on that but Sylvain said since connectivity to acs-mirror is already a given, we just need to check that the logic of the function works. This does run in every e2e as is because it will be part of CSE, but the second half of the function will rarely be executed as the nodes in our e2e tests should always be able to establish connectivity to packages.aks.azure.com

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: `PACKAGE_DOWNLOAD_BASE_URL` will be globally scoped due to its lack of the `local` keyword before its declaration

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: done @Devinwong

zachary-bailey: the full functionality of this function is also tested in the build pipeline since we can block IPs in the temp VM. I dont think itd be good to do that on dev's local machines

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: I see. Once user sets the firewall to allow packages.aks.azure.com and do a provisioning again, `PACKAGE_DOWNLOAD_BASE_URL` be set to `packages.aks.azure.com` and then it will work.

parts/common/components.json
Devinwong: This is under the "windows" block. ln 840

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: nit: although the final value is the same. I will prefer moving ln817 before the for-loop like ln808. That's more clear that the initial value is acs-mirror. But since it's not accessible, we echo the message asking user to config the firewall. Ln817 here did cause my initial confusion why we still set it back to acs-mirror if it's not desired.

parts/common/components.json
zachary-bailey: Great catch, thank you

parts/linux/cloud-init/artifacts/cse_helpers.sh
zachary-bailey: So `packages.aks.azure.com` should actually be the default / is the desired URL for CDN. I set that at the beginning then added some comments, does this work now?

pkg/agent/testdata/AKSUbuntu1604+Containerd/line9.sh
```diff
@@ -752,6 +752,40 @@ verify_DNS_health(){
...
+          PACKAGE_DOWNLOAD_BASE_URL="acs-mirror.azureedge.net"
+          echo "Setting PACKAGE_DOWNLOAD_BASE_URL to $PACKAGE_DOWNLOAD_BASE_URL. Please check to ensure cluster firewall has packages.aks.azure.com on its allowlist"
+          break
+        else
+          sleep $wait_sleep
+        fi
+      fi
+    done
+}
+
```
zachary-bailey, line 776: Seems like our comments disappeared, are you okay with how the function is now?

zachary-bailey: @Devinwong

djsly (approved):

Devinwong (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: yeah, as copilot also said, using [[ ]] is not bash-exportable, where as [ ] is POSIX-compliant (and hence should work on any POSIX-based shell)

PR #6094, opened 2025-03-27
test: update e2e config to match evictionHard default value to 100Mi

bravebeaver:
**What type of PR is this?**
/kind test

after k8s 1.29 the evictionHard is set to 100Mi by default

https://learn.microsoft.com/en-us/azure/aks/node-resource-reservations

the PR update the default node config in e2e test to match the default, so that the node created are closer to real life.

**Requirements**:

- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -83,7 +83,7 @@ var baseKubeletConfig = &aksnodeconfigv1.KubeletConfig{
 		PodPidsLimit:                   to.Ptr(int32(-1)),
 		ResolvConf:                     "/run/systemd/resolve/resolv.conf",
 		EvictionHard: map[string]string{
-			"memory.available":  "750Mi",
+			"memory.available":  "100Mi",
 			"nodefs.available":  "10%",
 			"nodefs.inodesFree": "5%",
 		},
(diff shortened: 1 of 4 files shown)


djsly (approved):

e2e/node_config.go
```diff
@@ -83,7 +83,7 @@ var baseKubeletConfig = &aksnodeconfigv1.KubeletConfig{
 		PodPidsLimit:                   to.Ptr(int32(-1)),
 		ResolvConf:                     "/run/systemd/resolve/resolv.conf",
 		EvictionHard: map[string]string{
-			"memory.available":  "750Mi",
+			"memory.available":  "100Mi",
```
smith1511, line 86: Is there a way to pull this from the "official" value used by AB so we know they're always aligned?

smith1511 (approved):

e2e/node_config.go
bravebeaver: not really.. we get the value from rp. 
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=/resourceprovider/sharedlib/agentbaker/nodebootstrappingconfig.go&version=GBmaster&line=275&lineEnd=276&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

ganeshkumarashok:
Looks like the e2e had actually failed in this PR itself: `Test_Ubuntu2204VanillaK8S129EvictionThreshold`

PR #6109, opened 2025-03-28
chore: github action updated to v7 and golangci-lint version v2.0 is …

bravebeaver:
…required https://github.com/golangci/golangci-lint-action
/kind ci

image


.golangci.bck.yaml
```diff
@@ -0,0 +1,313 @@
+# Borrowed from https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322
```
smith1511, line 1: I'd probably just get rid of the old file tbh.

smith1511 (approved):

PR #6110, opened 2025-03-28
Add localdns related files to VHDs

SriHarsha001:
**What type of PR is this?**
/kind feature

This PR adds all the necessary localdns related files to the VHDs.

Fixes # None

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/05-aks-keepconfig.conf
```diff
@@ -0,0 +1,7 @@
+# This network file will modify the default Azure network interface and set
+# networkd to never remove its dhcp configuration, even if the link gets 
+# dropped for some reason. This will help prevent outages since Azure never
+# removes a DHCP lease.
+
+[Network]
+KeepConfiguration=dhcp
```
yewmsft, line 7: q: who uses this config?

parts/linux/cloud-init/artifacts/localdns.service
```diff
@@ -0,0 +1,23 @@
...
+
+[Service]
+Type=notify
+NotifyAccess=all
+WatchdogSec=60
+Restart=on-failure
+KillMode=mixed
+TimeoutStopSec=30
+Slice=localdns.slice
+EnvironmentFile=-/etc/default/localdns.envfile
```
yewmsft, line 19: we are setting localdns.envfile to be optional here. What happens when localdns.envfile does not exist?

SriHarsha001: Not sure what indicates optional here. Is it "-" ?. But I have a check in localdns.sh file to check for localdns.envfile file and env variable inside this file.

SriHarsha001: Ah I see. I did not know "-" indicates optional here. I will update this file and remove "-". This file should not be optional.

SriHarsha001: Removed "-" from the service file.

parts/linux/cloud-init/artifacts/05-aks-keepconfig.conf
SriHarsha001: This file 05-aks-keepconfig.conf is placed in this directory - /etc/systemd/network/10-netplan-eth0.network.d. It is a systemd-networkd drop-in configuration that modifies the behavior of networkd for the default Azure network interface (eth0). The KeepConfiguration=dhcp option ensures that systemd-networkd does not remove the DHCP configuration, even if the network link goes down temporarily.

SriHarsha001: @phealy - How to verify that eth0 is correctly configured to retain dhcp lease ? Any pointers please ?

phealy: Sorry, I think this is a junk file that was in my original test branch and has just been carried over into Sri's work. It doesn't do anything because we're not using systemd-networkd for NIC configuration. This should just be removed.

parts/linux/cloud-init/artifacts/localdns-resolved.conf
```diff
@@ -0,0 +1,10 @@
+# This file modifies systemd-resolved's configuration to make it work better in Azure. 
+# It will fall back to AzureDNS if the DHCP DNS servers (which might be custom) aren't available. 
+# It will ignore multicast and link-local.
+# It won't do any caching (that's handled by localdns).
+
+[Resolve]
+FallbackDNS=168.63.129.16
+MulticastDNS=no
+LLMNR=no
+Cache=no
```
phealy, line 10: Since AKS disables systemd-resolved, this can be removed from the PR; it was from an earlier round of testing and didn't get excluded.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+# Delay coredns shutdown to allow connections to finish.
+: "${LOCALDNS_SHUTDOWN_DELAY:?LOCALDNS_SHUTDOWN_DELAY is not set}"
+# PID file.
+: "${LOCALDNS_PID_FILE:?LOCALDNS_PID_FILE is not set}"
+# Cluster CoreDNS service IP.
+: "${COREDNS_SERVICE_IP:?COREDNS_SERVICE_IP is not set}"
+
+# Check if coredns binary is cached in VHD.
+# --------------------------------------------------------------------------------------------------------------------
+LOCALDNS_SCRIPT_PATH="/opt/azure/containers/localdns"
```
phealy, line 57: Just for clarity's sake this should be moved to the top of the file and then referenced in the value of LOCALDNS_CORE_FILE

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+    s/CoreDNS_Service_IP/${COREDNS_SERVICE_IP}/g;
+    s/Node_Listener_IP/${LOCALDNS_NODE_LISTENER_IP}/g;
+    s/Cluster_Listener_IP/${LOCALDNS_CLUSTER_LISTENER_IP}/g
+" "${LOCALDNS_CORE_FILE}" || { echo "Error: updating corefile failed"; exit 1; }
+
+cat "${LOCALDNS_CORE_FILE}"
+
+# Iptables: build rules.
+# --------------------------------------------------------------------------------------------------------------------
+# These rules skip conntrack for DNS traffic to save conntrack table space.
```
phealy, line 113:
```suggestion
# These rules skip conntrack for DNS traffic to the local DNS service IPs to save conntrack table space. 
```

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+# Takes a percentage value, suffixed with "%". The percentage specifies how much CPU time the unit shall get at maximum, 
+# relative to the total CPU time available on one CPU. Use values > 100% for allotting CPU time on more than one CPU.
+CPU_QUOTA="$(echo "scale=1; ${LOCALDNS_CPU_LIMIT_IN_MILLI_CORES} / 10" | bc)%"
+
+CGROUP_VERSION=$(stat -fc %T /sys/fs/cgroup)
+if [ "${CGROUP_VERSION}" = "cgroup2fs" ] || [ "${CGROUP_VERSION}" = "tmpfs" ]; then
+    sed -i \
+        -e "s/^CPUQuota=[^ ]*/CPUQuota=${CPU_QUOTA}/" \
+        -e "s/^MemoryMax=[^ ]*/MemoryMax=${LOCALDNS_MEMORY_LIMIT_IN_MB}M/" \
+        "${LOCALDNS_SLICE_PATH}" || { echo "Error: updating localdns slice failed"; exit 1; }
```
phealy, line 85: This will need a `systemctl daemon-reload` to take effect, I believe.

phealy: Actually, since we're already running inside that slice in this script, this may not even be adjustable here. This may need to be adjusted in the provisioning scripts to write the file out correctly in the first place.

parts/linux/cloud-init/artifacts/05-aks-keepconfig.conf
SriHarsha001: Thank you for taking a look. I have removed this file.

parts/linux/cloud-init/artifacts/localdns-resolved.conf
SriHarsha001: Thank you for taking a look. I have removed this file.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Sure, moved this variable up.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Sure, updated the comments.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+# This systemd unit runs coredns as a caching with serve-stale functionality for both pod DNS and node DNS queries. 
+# It also upgrades to TCP for better reliability of upstream connections.
+
+LOCALDNS_SCRIPT_PATH="/opt/azure/containers/localdns"
+
+# Verify the required files exists.
+# --------------------------------------------------------------------------------------------------------------------
+# This file contains the environment variables used by localdns systemd unit.
+# This path should match with 'path' defined in parts/linux/cloud-init/nodecustomdata.yml.
+LOCALDNS_ENV_FILE="/etc/default/localdns.envfile"
```
yewmsft, line 15: is there a specific reason we are placing localdns.envfile under "/etc/default"? why not under "/opt/azure/containers/localdns" as well?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+    printf "Error: localdns envfile does not exist at %s.\n" "${LOCALDNS_ENV_FILE}"
+    exit 1
+fi
+
+# This file contains generated Corefile used by localdns systemd unit.
+# This should match with 'path' defined in parts/linux/cloud-init/nodecustomdata.yml.
+LOCALDNS_CORE_FILE="${LOCALDNS_SCRIPT_PATH}/localdns.corefile"
+if [ ! -f "${LOCALDNS_CORE_FILE}" ] || [ ! -s "${LOCALDNS_CORE_FILE}" ]; then
+    printf "Error: localdns corefile either does not exist or is empty at %s.\n" "${LOCALDNS_CORE_FILE}"
+    exit 1
```
yewmsft, line 28: Is using different exit codes here meaningful for debugging?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+if [ ! -f "${LOCALDNS_SLICE_PATH}" ]; then
+    printf "Error: localdns slice file does not exist at %s.\n" "${LOCALDNS_SLICE_PATH}"
+    exit 1
+fi
+
+# Load environment variables from envfile.
+# Each of the below variables must match the field name defined in parts/linux/cloud-init/nodecustomdata.yml.
+# --------------------------------------------------------------------------------------------------------------------
+# This is the IP that localdns service should bind to for node traffic; an APIPA address.
+: "${LOCALDNS_NODE_LISTENER_IP:?LOCALDNS_NODE_LISTENER_IP is not set}"
```
yewmsft, line 43: where are env variables values are set? e.g. LOCALDNS_NODE_LISTENER_IP

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: I am placing it in /etc/default because I remember there was discussion about allowing customer to override these values in future.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: This will be set in nodecustomdata.yml - which I plan to submit in the next PR.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+if [ $? -ne 0 ]; then
+    printf "Error: Failed to execute '%s --plugins'.\n" "${COREDNS_BINARY_PATH}"
+    exit 1
+fi
+
+# Configure CPU and Memory limit.
+# --------------------------------------------------------------------------------------------------------------------
+# Takes a percentage value, suffixed with "%". The percentage specifies how much CPU time the unit shall get at maximum, 
+# relative to the total CPU time available on one CPU. Use values > 100% for allotting CPU time on more than one CPU.
+CPU_QUOTA="$(echo "scale=1; ${LOCALDNS_CPU_LIMIT_IN_MILLI_CORES} / 10" | bc)%"
```
yewmsft, line 78: just curious, what happens when this cpu quota exceeds the node cpu cores

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: +1. This should happen during cloud-init

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,271 @@
...
+UPSTREAM_VNET_DNS_SERVERS=$(awk '/nameserver/ {print $2}' /run/systemd/resolve/resolv.conf | paste -sd' ')
+# Get the upstream VNET DNS servers from /run/systemd/resolve/resolv.conf.
+if [[ -z "${UPSTREAM_VNET_DNS_SERVERS}" ]]; then
+    printf "Error: No Upstream VNET DNS servers found in /run/systemd/resolve/resolv.conf.\n"
+    exit 1
+fi
+
+# Based on customer input, corefile was generated with Vnet_DNS_Server as placeholder in pkg/agent/baker.go.
+# Replace all occurrences of VnetDNS_Server_IP, CoreDNS_Service_IP Node_Listener_IP and Cluster_Listener_IP in localdns corefile.
+sed -i -e "
```
yewmsft, line 102: we should have already filled these data during custome data templating, right? At this point, the corefiles should be ready.

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: And I thought this is already done at cutom data tempating time. No?

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: yes, corefile used by localdns is generated and send to the node using nodecustomdata (we will see this in next PR). At this point, the corefile will have all the values except these - VnetDNS_Server_IP, CoreDNS_Service_IP Node_Listener_IP and Cluster_Listener_IP. Here in localdns.sh file - I find and replace all occurrences of these placeholders with the actual value passed in via .envfile. We will know the upstream VNET DNS server only here on the node. I also need LOCALDNS_CLUSTER_LISTENER_IP in .envfile, so that after localdns unit is up, I need to point kubelet's flag --cluster-dns to LOCALDNS_CLUSTER_LISTENER_IP, which in turn will point all the pods created on the node to LOCALDNS_CLUSTER_LISTENER_IP. I am keeping these variables (including coredns serviceIP) in .envfile at etc/default, so that these can be overridden in the future. It will also enable us to support the scenario where customer might scaledown coredns managed addon and have their own implementation of coredns.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: The above approach works as well. But I think a clean way to do this is to generate slice file in nodecustomdata as you all suggested. My commit on this PR will move all this logic out of localdns.sh file. On the next PR, we will see the logic in nodecustomdata.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: This code logic in out from this file. However, when CPUQuota/MemoryMax limits in slice file exceeds available CPU cores/memory on the node, it can lead to CPU throttling or system OOM, and cause overall system performance degradation. I will have to validate CPULimitInMilliCores/MemoryLimitInMB input in AKS-RP against what is offered for a given VM SKU (and depending on max_concurrent settings for memorylimit) and reject the call if the input is unrealistic.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Any error here will be logged in cluster-provisioning.log for debugging. If anything fails, cse script will fail with exit code 216 currently as defined in cse_helper.sh file (We will see this in next few PRs). localdns.sh file is invoked from main.sh which will capture and surface the failure code of localdns.

SriHarsha001: Please unresolve if you still have any questions or concerns.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Please unresolve if you still have any questions or concerns.

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: Do you want to making more detailed the error exit codes when returning? e.g. corefile does not exist, exit 216, unsupported cgroup version exit 217, fail to cleanup exit 218, and etc.

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: cool, another validator task then

yewmsft: that brings another question, since setting the cpu limit is depend on node specs, which is tied to vm sizes, should we offer default values and allow cx to leave those values out in the api call?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,242 @@
...
+IPTABLES_RULES=()
+for CHAIN in OUTPUT PREROUTING; do
+for IP in ${LOCALDNS_NODE_LISTENER_IP} ${LOCALDNS_CLUSTER_LISTENER_IP}; do
+for PROTO in tcp udp; do
+    IPTABLES_RULES+=("${CHAIN} -p ${PROTO} -d ${IP} --dport 53 -j NOTRACK")
+done; done; done
+
+# Information variables.
+# --------------------------------------------------------------------------------------------------------------------
+DEFAULT_ROUTE_INTERFACE="$(ip -j route get 168.63.129.16 | jq -r '.[0].dev')"
```
yewmsft, line 97: please use a variable for 168.63.129.16

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,242 @@
...
+# Cleanup function will be run on script exit/crash to revert config.
+# --------------------------------------------------------------------------------------------------------------------
+function cleanup {
+    # Disable error handling so that we don't get into a recursive loop.
+    set +e
+
+    # Remove iptables rules to stop forwarding DNS traffic.
+    for RULE in "${IPTABLES_RULES[@]}"; do
+        if eval "${IPTABLES}" -C "${RULE}" 2>/dev/null; then
+            eval "${IPTABLES}" -D "${RULE}"
```
yewmsft, line 111: can this command fail?

yewmsft: or it will just fail silently?

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: yes, default value for LOCALDNS_CPU_LIMIT_IN_MILLI_CORES is 2000 and for MemoryLimitInMB it is 128

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,242 @@
...
+    # Delete the dummy interface if present.
+    if ip link show dev localdns >/dev/null 2>&1; then
+        printf "removing localdns dummy interface.\n"
+        ip link del name localdns
+    fi
+}
+
+# If we're invoked with cleanup, run cleanup.
+if [[ $* == *--cleanup* ]]; then
+    cleanup
```
yewmsft, line 151: How do we know whether cleanup succeeded or not?

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Used a variable for AzureDNSIP.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Ah - this particular snippet was redundant code and will not be triggered. Removed this from localdns.sh file.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,242 @@
...
+COREDNS_COMMAND="${COREDNS_BINARY_PATH} -conf ${LOCALDNS_CORE_FILE} -pidfile ${LOCALDNS_PID_FILE}"
+if [[ ! -z "${SYSTEMD_EXEC_PID:-}" ]]; then
+    # We're running in systemd, so pass the coredns output via systemd-cat.
+    COREDNS_COMMAND="systemd-cat --identifier=localdns-coredns --stderr-priority=3 -- ${COREDNS_COMMAND}"
+fi
+
+printf "starting localdns: %s.\n" "${COREDNS_COMMAND}"
+rm -f ${LOCALDNS_PID_FILE}
+${COREDNS_COMMAND} &
+until [ -f ${LOCALDNS_PID_FILE} ]; do
```
yewmsft, line 181: add a timeout?

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: There will be about 12 different exit codes that I will have to define and use. Please let me know if you want me to do this.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Added the checks in the code.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Added the checks in code.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Added timeout in the code

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: You can use 1 exit code per category, or you can use more granular approach for runtime errors. But anyway, it is up to how you want to process those exit codes. You may add exit code as you see fit.

SriHarsha001: ok, made the code changes.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -164,6 +171,29 @@ CURL_OUTPUT=/tmp/curl_verbose.out
 ORAS_OUTPUT=/tmp/oras_verbose.out
 ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file, not used, but have to define to avoid error "Error: failed to get user home directory: $HOME is not defined" 
 
+# ------------------------------ Used by localdns -----------------------------------
```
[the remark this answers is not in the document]
SriHarsha001: ok, moved the variables back to localdns file.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -862,4 +869,21 @@ oras_login_with_kubelet_identity() {
     echo "successfully logged in to acr '$acr_url' with identity token"
 }
 
+# Localdns corefile is created only when localdns profile has state enabled.
+# This should match with 'path' defined in parts/linux/cloud-init/nodecustomdata.yml.
+LOCALDNS_CORE_FILE="/opt/azure/containers/localdns/localdns.corefile"
+# This function is called in cse_main.sh. 
+# It checks if the localdns corefile exists and is not empty.
+# If the corefile exists and is not empty, it returns 0 - localdns should be enabled.
+# If the corefile does not exist or is empty, it returns 1 - localdns should not be enabled.
```
Devinwong, line 878: I think you mean returns ERR_LOCALDNS_COREFILE_NOTFOUND in the comment

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
...
+for CHAIN in OUTPUT PREROUTING; do
+for IP in ${LOCALDNS_NODE_LISTENER_IP} ${LOCALDNS_CLUSTER_LISTENER_IP}; do
+for PROTO in tcp udp; do
+    IPTABLES_RULES+=("${CHAIN} -p ${PROTO} -d ${IP} --dport 53 -j NOTRACK")
+done; done; done
+
+# Information variables.
+# --------------------------------------------------------------------------------------------------------------------
+DEFAULT_ROUTE_INTERFACE="$(ip -j route get "${AZURE_DNS_IP}" | jq -r '.[0].dev')"
+NETWORK_FILE="$(networkctl --json=short status "${DEFAULT_ROUTE_INTERFACE}" | jq -r '.NetworkFile')"
```
Devinwong, line 103: If `.NetworkFile` doesn't exist, the value of NETWORK_FILE will be empty. And in the next line it will be `NETWORK_DROPIN_DIR=".d"`. Not sure if will cause error in other steps which use this variable. But if you are sure that `.NetworkFile` is definitely there, it should be good.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
...
+                printf "Successfully removed iptables rule: %s.\n" "${RULE}"
+            else
+                printf "Failed to remove iptables rule: %s.\n" "${RULE}"
+                return $ERR_LOCALDNS_FAIL
+            fi
+        fi
+    done
+
+    # Revert the changes made to the DNS configuration if present.
+    if [ -f ${NETWORK_DROPIN_FILE} ]; then
```
Devinwong, line 127: Could you make this 3-level if to be <= 2 level for easier debugging later?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
...
+# Get the upstream VNET DNS servers from /run/systemd/resolve/resolv.conf.
+if [[ -z "${UPSTREAM_VNET_DNS_SERVERS}" ]]; then
+    printf "No Upstream VNET DNS servers found in /run/systemd/resolve/resolv.conf.\n"
+    exit $ERR_LOCALDNS_FAIL
+fi
+
+# Based on customer input, corefile was generated in pkg/agent/baker.go.
+# Replace 168.63.129.16 with VNET DNS ServerIPs only if VNET DNS ServerIPs is not equal to 168.63.129.16.
+if [[ "${UPSTREAM_VNET_DNS_SERVERS}" != "${AZURE_DNS_IP}" ]]; then
+    sed -i -e "s|168.63.129.16|${UPSTREAM_VNET_DNS_SERVERS}|g" "${LOCALDNS_CORE_FILE}" || {
```
Devinwong, line 80: Can't we use the variable ${AZURE_DNS_IP} here?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
...
+IPTABLES_RULES=()
+for CHAIN in OUTPUT PREROUTING; do
+for IP in ${LOCALDNS_NODE_LISTENER_IP} ${LOCALDNS_CLUSTER_LISTENER_IP}; do
+for PROTO in tcp udp; do
+    IPTABLES_RULES+=("${CHAIN} -p ${PROTO} -d ${IP} --dport 53 -j NOTRACK")
+done; done; done
+
+# Information variables.
+# --------------------------------------------------------------------------------------------------------------------
+DEFAULT_ROUTE_INTERFACE="$(ip -j route get "${AZURE_DNS_IP}" | jq -r '.[0].dev')"
```
Devinwong, line 102: Is it guaranteed that the output of `ip -j route command` has an array object so that your jq command always works?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
+#! /bin/bash
```
Devinwong, line 1: (I think I wrote this somewhere but it disappeared)
Apparently this file has all the critical logic. It will be great if you can add some unit tests to cover the critical logic parts. Before that you may need to modify the codes a bit into testable functions so that you can test them. In my own experience, with the unit test it helped me catch many errors like 'empty value', 'file not found', 'file content format unexpected, especially for jq command' or even just syntax error. Therefore, I will highly recommend adding some unit tests to the codes you think it could easily cause errors. One example is `spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh`. You can also find the guidance from readme.

parts/linux/cloud-init/artifacts/cse_helpers.sh
SriHarsha001: Ah yes, I have updated the comment.

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: +1

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
...
+# Check if --plugins command runs successfully.
+builtInPlugins=$("${COREDNS_BINARY_PATH}" --plugins)
+if [ $? -ne 0 ]; then
+    printf "Failed to execute '%s --plugins'.\n" "${COREDNS_BINARY_PATH}"
+    exit $ERR_LOCALDNS_FAIL
+fi
+
+# Replace Vnet_DNS_Server in corefile with VNET DNS Server IPs.
+# --------------------------------------------------------------------------------------------------------------------
+UPSTREAM_VNET_DNS_SERVERS=$(awk '/nameserver/ {print $2}' /run/systemd/resolve/resolv.conf | paste -sd' ')
```
yewmsft, line 70: what if there are multiple nameservers in resolve.conf?

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft, line 80: Consulted with Paul, looks like this is the only way to get the vnet dns servers. We do not do anything about changes made to the vnet. If cx changes the dns server on the vnet, we do not trigger the reboot for them. If they do not reboot, nothing will take effect.

SriHarsha001: Made the change.

SriHarsha001: yes, thank you.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: If there are multiple nameservers, all nameservers will be extracted and joined with a space between the IPs.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,294 @@
...
+                return $ERR_LOCALDNS_FAIL
+            fi
+        else
+            printf "Failed to remove %s.\n" "${NETWORK_DROPIN_FILE}"
+            return $ERR_LOCALDNS_FAIL
+        fi
+    fi
+
+    # Trigger localdns shutdown, if running.
+    if [ -n "${COREDNS_PID:-}" ]; then
```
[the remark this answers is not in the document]
SriHarsha001: I have replaced this line with -     

# COREDNS_PID should be non-empty and a valid PID number.
if [[ -n "${COREDNS_PID}" ]] && [[ "${COREDNS_PID}" =~ ^[0-9]+$ ]]; then

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Azurelinux, mariner and ubuntu uses netplan and systemd-networkd, on which networkfile should exists. But I have added a check here just in case.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: I want to avoid sourcing cse_helper file in localdns file. There are only 4 exit codes, out of these 4, only 1 is repeatedly used in the file and the other 3 are used just ones. I want to exit with code 1 in the entire file (if there are any failures and log that info) and then exit with 216 defined in cse_helper for localdns start failure in cse_config. Please let me know if you still insist I source cse_helper file.

yewmsft: you don't need to soucing cse_helper at all. You can simply return 1, 2, 3 and 4, and at cse_main, return the real exit code.

SriHarsha001: ok.

yewmsft: also, it is all about how you want to handle those errors in RP, e.g. do you want to do something differently when encounter files not found vs fail to start?

yewmsft: another point is, would those different exit codes help you in live site investigations when an CRI come and claiming their localdns can't start?

yewmsft: another option will be, you can just hard code those exit codes in the systemd unit and return outside

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Added a check here.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: I have added few additional checks. This is inside a cleanup func and triggered by a trap, so do not want to exit prematurely.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: I have added cse_helpers_spec file to this PR. I am working on localdns test script currently.

SriHarsha001: ok, I have created localdns_spec.sh file which can be used to test most of the code in localdns.sh file.

```
root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# systemctl status localdns
* localdns.service - Localdns service
     Loaded: loaded (/etc/systemd/system/localdns.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/localdns.service.d
             `-delegate.conf
     Active: active (running) since Fri 2025-04-04 05:05:23 UTC; 19min ago
   Main PID: 3466 (localdns.sh)
      Tasks: 11 (limit: 9196)
     Memory: 62.6M (available: 65.3M)
        CPU: 2.215s
     CGroup: /localdns.slice/localdns.service
             |- 3466 /bin/bash /opt/azure/containers/localdns/localdns.sh
             |- 3528 /opt/azure/containers/localdns/binary/coredns -conf /opt/azure/containers/localdns/localdns.corefile -pidfile /run/localdns.pid
             `-15554 sleep 12

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# RESOLV_CONF="/run/systemd/resolve/resolv.conf"
root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# UPSTREAM_VNET_DNS_SERVERS=$(awk '/nameserver/ {print $2}' "$RESOLV_CONF" | paste -sd' ')
root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# echo $UPSTREAM_VNET_DNS_SERVERS
169.254.10.10

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# NETWORK_FILE="$(networkctl --json=short status "${DEFAULT_ROUTE_INTERFACE}" 2>/dev/null | jq -r '.NetworkFile')"
root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# echo $NETWORK_FILE
/run/systemd/network/10-netplan-eth0.network

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# ip link show localdns
4: localdns: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether d6:81:01:89:12:d2 brd ff:ff:ff:ff:ff:ff

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# iptables-save | grep NOTRACK
-A PREROUTING -d 169.254.10.10/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK
-A PREROUTING -d 169.254.10.10/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK
-A PREROUTING -d 169.254.10.11/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK
-A PREROUTING -d 169.254.10.11/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK
-A OUTPUT -d 169.254.10.10/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK
-A OUTPUT -d 169.254.10.10/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK
-A OUTPUT -d 169.254.10.11/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK
-A OUTPUT -d 169.254.10.11/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# cat /etc/default/kubelet | grep -- "--cluster-dns"
KUBELET_FLAGS=--address=0.0.0.0 --anonymous-auth=false --authentication-token-webhook=true --authorization-mode=Webhook --cgroups-per-qos=true --client-ca-file=/etc/kubernetes/certs/ca.crt --cloud-config= --cloud-provider=external --cluster-dns=169.254.10.11 

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# ip addr show dev localdns
4: localdns: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether d6:81:01:89:12:d2 brd ff:ff:ff:ff:ff:ff
    inet 169.254.10.10/32 scope global localdns
       valid_lft forever preferred_lft forever
    inet 169.254.10.11/32 scope global localdns
       valid_lft forever preferred_lft forever
    inet6 fe80::c413:85ff:fed6:3ffa/64 scope link
       valid_lft forever preferred_lft forever

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# cat /etc/resolv.conf
# This is /run/systemd/resolve/resolv.conf managed by man:systemd-resolved(8).
# Do not edit.
#
# This file might be symlinked as /etc/resolv.conf. If you're looking at
# /etc/resolv.conf and seeing this text, you have followed the symlink.
#
# This is a dynamic resolv.conf file for connecting local clients directly to
# all known uplink DNS servers. This file lists all configured search domains.
#
# Third party programs should typically not access this file directly, but only
# through the symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a
# different way, replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 169.254.10.10
search bdv5efs32ubelfytguaosg4dsd.phxx.internal.cloudapp.net

root@rclv-2025-04-03-azurelinuxv2000000 [ / ]# curl -s http://169.254.10.10:9253/metrics | grep goversion
# HELP coredns_build_info A metric with a constant '1' value labeled by version, revision, and goversion from which CoreDNS was built.
coredns_build_info{goversion="go1.23.3 X:systemcrypto",revision="1f0a41a66597cb8ab4aace8ea5b5bad880bcd23b",version="1.9.4"} 1
```

anujmaheshwari1 (dismissed):

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -0,0 +1,496 @@
...
+# If the health check fails, the script will exit and systemd will restart the service.
+start_localdns_watchdog() {
+    if [[ -n "${NOTIFY_SOCKET:-}" && -n "${WATCHDOG_USEC:-}" ]]; then
+        # Health check at 20% of WATCHDOG_USEC; this means that we should check.
+        # five times in every watchdog interval, and thus need to fail five checks to get restarted.
+        HEALTH_CHECK_INTERVAL=$((${WATCHDOG_USEC:-5000000} * 20 / 100 / 1000000))
+        echo "Starting watchdog loop at ${HEALTH_CHECK_INTERVAL} second intervals."
+        while true; do
+            if [[ "$($CURL_COMMAND)" == "OK" ]]; then
+                if dig +short +timeout=1 +tries=1 -f <(printf "%s" "$HEALTH_CHECK_DNS_REQUEST"); then
```
Devinwong, line 398: I think we can merge this if condition with the previous line? because you don't have an else case here.

Devinwong (dismissed):
Overall looks good! Thanks for the persistence and added many unit tests to cover the codes! I also learnt a lot throughout this process.

Devinwong (approved):

PR #6117, opened 2025-03-28
test: refactor extractKubeBinaries and add unit test

Devinwong:
**What type of PR is this?**
/kind test

test: refactor extractKubeBinaries and add unit test
The logic of extractKubeBinaries is a bit complicated. Adding some unit tests to validate it.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_install.sh
+++ b/parts/linux/cloud-init/artifacts/cse_install.sh
@@ -533,13 +533,33 @@ installAzureCNI() {
     chown -R root:root $CNI_BIN_DIR
 }
 
+# extract the cached or downloaded kube package and remove
+extractKubeBinariesToUsrLocalBin() {
+    local k8s_tgz_tmp=$1
+    local k8s_version=$2
+    local is_private_url=$3
+
+    if [[ ! -f "${k8s_tgz_tmp}" ]]; then
+        exit "$ERR_ORAS_PULL_K8S_FAIL"
+    fi
+
+    tar --transform="s|.*|&-${k8s_version}|" --show-transformed-names -xzvf "${k8s_tgz_tmp}" \
+        --strip-components=3 -C /usr/local/bin kubernetes/node/bin/kubelet kubernetes/node/bin/kubectl || exit $ERR_K8S_INSTALL_ERR
+    if [[ ! -f /usr/local/bin/kubectl-${k8s_version} ]] || [[ ! -f /usr/local/bin/kubelet-${k8s_version} ]]; then
+        exit $ERR_K8S_INSTALL_ERR
+    fi
+    if [[ $is_private_url == false ]]; then
+        rm -f "${k8s_tgz_tmp}"
+    fi
+}
+
 extractKubeBinaries() {
     local k8s_version="$1"
     # Remove leading 'v' if it exists
     k8s_version="${k8s_version#v}"
     local kube_binary_url="$2"
     local is_private_url="$3"
-    local k8s_downloads_dir="$4"
+    local k8s_downloads_dir=${4:-"/opt/kubernetes/downloads"}
 
     local k8s_tgz_tmp_filename=${kube_binary_url##*/}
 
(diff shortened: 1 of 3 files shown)


e2e/scenario_test.go
```diff
@@ -754,6 +754,10 @@ func Test_Ubuntu2204Gen2_ContainerdAirgappedK8sNotCached(t *testing.T) {
 						ContainerRegistryServer: fmt.Sprintf("%s.azurecr.io", config.PrivateACRName),
 					},
 				}
+				// intentionally using private acr url to get kube binaries
+				nbc.AgentPoolProfile.KubernetesConfig.CustomKubeBinaryURL = fmt.Sprintf(
+					"privateacre2ewestus3.azurecr.io/oss/binaries/kubernetes/kubernetes-node:v%s-linux-amd64",
```
cameronmeissner, line 759: don't we set the private ACR base URL dynamically via config? if so we probably shouldn't hard-code it here

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -559,34 +575,28 @@ extractKubeBinaries() {
...
+        
         # if the url is a registry url, use oras to pull the artifact instead of curl
         if isRegistryUrl "${kube_binary_url}"; then
             echo "detect kube_binary_url, ${kube_binary_url}, as registry url, will use oras to pull artifact binary"
             # download the kube package from registry as oras artifact
             k8s_tgz_tmp="${k8s_downloads_dir}/kubernetes-node-linux-${CPU_ARCH}.tar.gz"
             retrycmd_get_tarball_from_registry_with_oras 120 5 "${k8s_tgz_tmp}" ${kube_binary_url} || exit $ERR_ORAS_PULL_K8S_FAIL
-            if [[ ! -f "${k8s_tgz_tmp}" ]]; then
+            # use `test` instead so that we can mock it in unit tests.
+            if test ! -f "${k8s_tgz_tmp}"; then
```
cameronmeissner, line 586: we shouldn't need to do this, `[ ... ]` / `[[ ... ]]` should already automatically delegate to `test`, did you run into errors without this?

e2e/scenario_test.go
Devinwong: Right the default one is set in `baseTemplateLinux` in node_config.go but I didn't want to change the base one, which is a acs-mirror url, because rp is still using acs-mirror. I think there is a plan to deprecate acs-mirror so by then we can update the `baseTemplateLinux`. What do you think?

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: I wasn't able to mock [ ... ] with `test` earlier. But after seeing your comment, I tried again and it suddenly works. Maybe I did something wrong earlier. Thanks for the suggestion. I changed this back.

Devinwong: oh! The unit tests failed again, which means the `test` mock is not successful.
Examples:
  1) cse_install.sh extractKubeBinaries should use retrycmd_get_tarball_from_registry_with_oras to download kube binaries
     When call extractKubeBinaries 1.31.5 mcr.microsoft.com/oss/binaries/kubernetes/kubernetes-node:FakeTag false /opt/kubernetes/downloads

     1.1) Example aborted (exit status: 207)

          # spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh:1[52](https://github.com/Azure/AgentBaker/actions/runs/14203378914/job/39795313036?pr=6117#step:4:53)-159

  2) cse_install.sh extractKubeBinaries should use retrycmd_get_tarball to download kube binaries
     When call extractKubeBinaries 1.31.5 https://acs-mirror.azureedge.net/kubernetes/v1.31.5/binaries/Fakefile false /opt/kubernetes/downloads

     2.1) Example aborted (exit status: 31)

          # spec/parts/linux/cloud-init/artifacts/cse_install_spec.sh:1[60](https://github.com/Azure/AgentBaker/actions/runs/14203378914/job/39795313036?pr=6117#step:4:61)-166

e2e/scenario_test.go
cameronmeissner: that makes sense, though I was talking about this specifically: https://github.com/Azure/AgentBaker/blob/4e6ec2f0758b3ad813a57de9051d77b9a46728b7/e2e/config/config.go#L21

can we just use that here instead? that way we'll theoretically be able to run this scenario in any region

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: I got what you mean. I do this now.
```
nbc.AgentPoolProfile.KubernetesConfig.CustomKubeBinaryURL = fmt.Sprintf(
					"%s.azurecr.io/oss/binaries/kubernetes/kubernetes-node:v%s-linux-amd64",
					config.PrivateACRName,
					nbc.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion)
```

e2e/scenario_test.go
Devinwong: I got what you mean. I do this now.
```
nbc.AgentPoolProfile.KubernetesConfig.CustomKubeBinaryURL = fmt.Sprintf(
					"%s.azurecr.io/oss/binaries/kubernetes/kubernetes-node:v%s-linux-amd64",
					config.PrivateACRName,
					nbc.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion)
```

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: Updated with create temp file and remove it after the test

PR #6135, opened 2025-03-31
chore: predefine write-log

timmy-wright:
**What type of PR is this?**

/kind chore

In a PR run, I noticed the run failed because write-log was not defined. So this PR makes sure it's defined so we can print the log message.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


bravebeaver (approved):
used only in test.ps1.. LGTM

PR #6138, opened 2025-03-31
fix: don't source provision_source.sh in validate-kubelet-credentials.sh

cameronmeissner:
**What type of PR is this?**

fix: don't source provision_source.sh in validate-kubelet-credentials.sh

sourcing provision_source.sh can have unintended consequences and adds latency, thus this PR removes the source line and adds an implementation of logs_to_events directly in validate-kubelet-credentials.sh

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
```diff
@@ -11,6 +8,36 @@ MAX_RETRIES=${VALIDATE_KUBELET_CREDENTIALS_MAX_RETRIES:-30}
 RETRY_DELAY_SECONDS=${VALIDATE_KUBELET_CREDENTIALS_RETRY_DELAY_SECONDS:-2}
 RETRY_TIMEOUT_SECONDS=${VALIDATE_KUBELET_CREDENTIALS_RETRY_TIMEOUT_SECONDS:-5}
 
+# NOTE: this is available if we source /opt/azure/containers/provision_source.sh, though sourcing that file also runs other unrelated commands
+# which can have unintended consequences and increase latency. As a result, we simply duplicate the implementation of logs_to_events here directly.
+# TODO: move logs_to_events out of provision_source.sh to an independent script so other provisioning/runtime scripts can cleanly source it as needed.
```
Devinwong, line 13: Good catch. Is there any concern we don't move it to an independent script in this PR? Is it because there could be potential regression we need to consider?

djsly (dismissed):

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
cameronmeissner: we need to get this in so we can make sure to include it int his week's release before it causes more problems, though I can follow up with making that independent script afterwards

Devinwong (approved):

PR #6150, opened 2025-04-02
feat: Migrated cri-tools to PMC as packages

Devinwong:
**What type of PR is this?**
/kind feature

- Migrated cri-tools to PMC. Now we are able to use package management tools to install the packages instead of previously extracting binaries from acs-mirror.
- Added content test to validate the tool is install correctly after VHD build.
- AB e2e tests passed. https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=120566734&view=results

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


parts/common/components.json
```diff
@@ -1194,23 +1194,56 @@
...
+                "latestVersion": "1.32.0-ubuntu24.04u3"
+              }
+            ]
+          }
+        },
         "default": {
           "current": {
             "versionsV2": [
               {
                 "renovateTag": "<DO_NOT_UPDATE>",
```
djsly, line 1243: we dont support AzLinux yet ?

Devinwong: For Renovate, we still need to add a customer manager to detect the mariner repo. I can work on it after this PR.
FWIW, now in components.json Default.current means everything else, which includes Mariner v2 and AzureLinux v2 and v3.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -19,6 +17,17 @@ installDeps() {
     if [[ $OS_VERSION == "2.0" ]]; then
       systemctl --now mask nftables.service || exit $ERR_SYSTEMCTL_MASK_FAIL
     fi
+
+    # Install the package repo for the specific OS version.
+    # AzureLinux 3.0 uses the azurelinux-repos-cloud-native repo
+    # Other OS, e.g., Mariner 2.0 uses the mariner-repos-cloud-native repo
+    if [[ $OS_VERSION == "3.0" ]]; then
+    echo "Installing azurelinux-repos-cloud-native"
```
cameronmeissner, line 25: nit: fix indentation

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -80,6 +80,13 @@ cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST /opt/gpu
 }
 
+installCriCtlPackage() {
+    version="${1:-}"
+    updateAptWithMicrosoftPkg
```
cameronmeissner, line 85: don't we already `updateAptWithMicrosoftPkg` before this?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -80,6 +80,13 @@ cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST /opt/gpu
 }
 
+installCriCtlPackage() {
+    version="${1:-}"
```
cameronmeissner, line 84: what happens if we default to empty string here?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -180,4 +195,9 @@ downloadContainerdFromURL() {
     echo "downloadContainerdFromURL not implemented for mariner"
 }
 
+# this directive is used to prevent the script under here from being run when shellspec includes it in the unit tests
+${__SOURCED__:+return}
```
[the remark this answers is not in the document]
Devinwong: ya otherwise, `source` or shellspec's `include` which is equivalent to `source`, will execute all the codes but actually we just want to define the functions

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
Devinwong: hmm. Good point. Let me spin up a VM to check.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
Devinwong: Yes. I will remove it. Thanks for catching this

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
Devinwong: exit with error if version is empty because it's required in components.json

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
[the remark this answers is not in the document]
Devinwong: It turns out not that straight forward to just remove extra `updateAptWithMicrosoftPkg`. I tried different attempts but encountered error like `GPG error: https://packages.microsoft.com/ubuntu/24.04/prod noble InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY EB3E94ADBE1229CF` only happening on Ubuntu 24.04. I will resume to the original status and further investigate this in the next PR.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -19,6 +17,17 @@ installDeps() {
     if [[ $OS_VERSION == "2.0" ]]; then
       systemctl --now mask nftables.service || exit $ERR_SYSTEMCTL_MASK_FAIL
     fi
+
+    # Install the package repo for the specific OS version.
+    # AzureLinux 3.0 uses the azurelinux-repos-cloud-native repo
+    # Other OS, e.g., Mariner 2.0 uses the mariner-repos-cloud-native repo
+    if [[ $OS_VERSION == "3.0" ]]; then
+      echo "Installing azurelinux-repos-cloud-native"
```
mainred, line 25: May  I ask how these repos are maintained? Are we planning to migrate all binaries to PMC in the future?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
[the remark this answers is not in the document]
cameronmeissner: which lines is this preventing shellspec from executing, just the echo below? If so I'd say it's not worth including this

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -243,6 +243,26 @@ downloadCNI() {
     retrycmd_get_tarball 120 5 "$downloadDir/${cniTgzTmp}" ${CNI_PLUGINS_URL} || exit $ERR_CNI_DOWNLOAD_TIMEOUT
 }
 
+downloadAndInstallCriTools() {
+  downloadDir=${1}
+  evaluatedURL=${2}
+  version=${3}
+
+  # if downloadDir and evaluatedURL are not empty, download and install crictl by this override, which is the old way to install
+  if [[ "${downloadDir}" != "" && "${evaluatedURL}" != "" ]]; then
```
cameronmeissner, line 252:
```suggestion
  if [ "${downloadDir}" != "" ] && [ "${evaluatedURL}" != "" ]; then
```
we should see if we can add our own custom rules to shellcheck

Devinwong: let me check

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
Devinwong: There are 3 binaries still hosting in acs-mirror. The plan is to move these all to PMC and use Dalec to manage/build them. Paul's team is working on CNI plugin-ins. Another one to be moved is Azure-cni. This is for Linux only though. Not sure about the plan of Windows.

vhdbuilder/packer/install-dependencies.sh
Devinwong: Updated for this one.
I found a way to check the POSIX compliant scripts but it caught a lot of errors as expected. I will try to fix all of them in a new PR today.
The command will be changed to this one. `shellcheck "--shell=sh"  $(printf -- "-e %s " $IGNORED) $filesToCheck`

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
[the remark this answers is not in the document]
Devinwong: Yes just this line. It was in the beginning of this file before the change, not added by me.
You think it's fine to remove this, right?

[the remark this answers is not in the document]
cameronmeissner: yeah I'd prefer we remove the `${__SOURCED__:+return}` then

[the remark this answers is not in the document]
Devinwong: Then I will also remove the echo in cse_install_ubuntu.sh for alignment

cameronmeissner (approved):

PR #6154, opened 2025-04-04
feat: use --preserve-status for timeout commands in cse

UtheMan:
Adds `--preserve-status` flag to timeout calls in CSE script, to retain the actual exit code for the command that timed out.


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -167,7 +167,7 @@ ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file
 retrycmd_if_failure() {
     retries=$1; wait_sleep=$2; timeout=$3; shift && shift && shift
     for i in $(seq 1 $retries); do
-        timeout $timeout "${@}" && break || \
+        timeout --preserve-status $timeout "${@}" && break || \
```
lilypan26, line 170: Is the plan to eventually override each of these timeout commands with custom error codes? Is it safe to merge this before that happens? im wondering if the error codes returned as is will overlap with some existing ones leading to confusion. Feel like it would be best to have it in the same PR but maybe im missing something

cameronmeissner: as Lily mentioned, we should only ever be returned well-defined exit codes via something like:
```
retry_cmd_if_failure <whatever> || exit $WELL_DEFINED_EXIT_CODE_BASED_ON_TASK
```
if we don't already have that everywhere we use this then this change won't be very helpful by itself

cameronmeissner: we also need to make sure that `timeout` isn't causing the main shell to exit with `124`, that way the `|| exit $WELL_DEFINED_EXIT_CODE_BASED_ON_TASK` would actually be executed

if that is happening, we might need to execute the timeout command in a subshell and then explicitly check the exit code

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 # Timeout waiting for a file
 ERR_FILE_WATCH_TIMEOUT=6 
+CSE_PROGRESS_TIMEOUT_CODE=1
```
UtheMan, line 4: From my understanding. there are two things that need to be addressed in this change to how we handle exit codes in CSE:

- Certain parts of CSE timing out (e.g. timeout is hit when we try to install k8s binaries)
- All of CSE times out (we set a 15m timeout) 

Because of the "global" CSE timeout, we eventually run into an issue where entire script is terminated and that's why we see exit code 124. To address that, we need a way to handle the SIGTERM signal that is sent by the timeout command from `cse_start.sh` 

My suggestion is as follows:
- Add `--preserve-status` to timeout call in `cse_start.sh`
- Add a trap for `SIGTERM` in `cse_main.sh` to handle the "global CSE" timeout during CSE execution.
- Add a "global" variable which will hold a special exit code (or a message?) that gets updated as CSE progresses. This message/exit code would be returned when SIGTERM appears and would be visible in the CSE output.

Drawbacks:
- we need to keep updating a global variable holding the "state" of CSE execution to return proper information about why CSE was terminated - as far as I know there is no "nice way" of doing this - here I am updating the variable as script progresses, but maybe others have some better suggestions? 
- We have to trap the signal in `cse_main.sh` - I could change the timeout call in `cse_start.sh` to send either `SIGUSR1` or `SIGUSR2` when timeout hits, but if CSE is terminated with some other signal (for some random reason) then our custom variable will not be returned (not sure yet how big of a problem this could potentially be).
- More bash fun added to the script

UtheMan: Another option could be to track progress in a file and read from it when the `SIGTERM` happens, but not sure if that's any "nicer" than just doing it with a variable - could potentially avoid issues with parts of the script that are triggered in subshells?

UtheMan: I kind of like the file idea to be fair... will experiment with that a bit

zachary-bailey: How is the file idea going? 

Thats what was going through my mind while I was first reading your comment.

PR #6155, opened 2025-04-04
chore: remove duplicated retrycmd_if_failure_silent func in cse

UtheMan:
`retrycmd_if_failure_silent` is only used in one place and has the same implementation as `retrycmd_if_failure_no_stats` that is used across the code base. Switching the one use to the latter and deleting the now unused code.


pkg/agent/testdata/RawUbuntuContainerd/line9.sh
```diff
@@ -162,18 +162,6 @@ retrycmd_if_failure() {
     echo Executed \"$@\" $i times;
 }
 
-retrycmd_if_failure_silent() {
-    retries=$1; wait_sleep=$2; timeout=$3; shift && shift && shift
-    for i in $(seq 1 $retries); do
-        timeout $timeout "${@}" && break || \
```
djsly, line 168: one command uses `"${@}` while the other command uses `${@}`  I think the one enclosed in `""` is safer ?

UtheMan: Added quotes in the function that remains

cameronmeissner (approved):

PR #6164, opened 2025-04-08
Add types and tests for localdns

SriHarsha001:
**What type of PR is this?**
feature

This PR has types and tests for the new feature 'localdns'. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -2318,7 +2318,7 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
 
 // SecurityProfile end.
 
-// ----------------------- Start of chnages related to localdns ------------------------------------------.
+// ----------------------- Start of changes related to localdns ------------------------------------------.
 // localdns related constants.
 const (
 	// LocalDNSEnabled is a string constant for 'Enabled'.
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -2318,7 +2318,7 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
 
 // SecurityProfile end.
 
-// ----------------------- Start of Changed related to localdns ------------------------------------------
+// ----------------------- Start of Changed related to localdns ------------------------------------------.
 // localdns related constants.
 const (
 	// LocalDNSEnabled is a string constant for 'Enabled'.
@@ -2429,4 +2429,4 @@ func (a *AgentPoolProfile) GetLocalDNSCoreFileData() LocalDNSCoreFileData {
 	return LocalDNSCoreFileData{}
 }
 
-// ----------------------- End of Changed related to localdns ------------------------------------------
+// ----------------------- End of Changed related to localdns ------------------------------------------.
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -830,6 +830,7 @@ type AgentPoolProfile struct {
 	CustomKubeletConfig   *CustomKubeletConfig `json:"customKubeletConfig,omitempty"`
 	CustomLinuxOSConfig   *CustomLinuxOSConfig `json:"customLinuxOSConfig,omitempty"`
 	MessageOfTheDay       string               `json:"messageOfTheDay,omitempty"`
+	LocalDNSProfile       *LocalDNSProfile     `json:"localDNSProfile,omitempty"`
 	/* This is a new property and all old agent pools do no have this field. We need to keep the default
 	behavior to reboot Windows node when it is nil. */
 	NotRebootWindowsNode    *bool                    `json:"notRebootWindowsNode,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,116 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
...
+	if a.ShouldEnableLocalDNS() && a.KubernetesConfig != nil && a.KubernetesConfig.DNSServiceIP != "" {
+		return a.KubernetesConfig.DNSServiceIP
+	}
+	return DefaultCoreDNSServiceIP
+}
+
+// GetLocalDNSCoreFileData returns the object that will be used to generate localdns corefile.
+func (a *AgentPoolProfile) GetLocalDNSCoreFileData() LocalDNSCoreFileData {
+	if a.ShouldEnableLocalDNS() {
+		localdnsCoreFileData := LocalDNSCoreFileData{
```
Devinwong, line 2420: The upper/lower case seems not consistent.
(IMO) If the type is `LocalDNSCoreFileData`, then the variable should be `localDNSCoreFileData` for consistency.

Devinwong (dismissed):
approve with a suggestion

pkg/agent/datamodel/types.go
SriHarsha001: Ah good catch. I have made the change.

Devinwong (approved):

pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,116 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
...
+	MaxConcurrent               *int32 `json:"maxConcurrent,omitempty"`
+	CacheDurationInSeconds      *int32 `json:"cacheDurationInSeconds,omitempty"`
+	ServeStaleDurationInSeconds *int32 `json:"serveStaleDurationInSeconds,omitempty"`
+	ServeStale                  string `json:"serveStale,omitempty"`
+}
+
+// ShouldEnableLocalDNS returns true if LocalDNSProfile is not nil and GoalState is 'Enable'.
+func (a *AgentPoolProfile) ShouldEnableLocalDNS() bool {
+	return a != nil && a.LocalDNSProfile != nil &&
+		strings.EqualFold(a.LocalDNSProfile.State, LocalDNSEnabled)
```
yewmsft, line 2374: This should be mode, right? State should only be set after localdns being enabled or disabled

pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,116 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
...
+	DefaultLocalDNSMemoryLimitInMB string = "128M"
+	// DefaultLocalDNSServiceIP sepcifies the default coredns service IP.
+	DefaultCoreDNSServiceIP string = "10.0.0.10"
+	// AzureDNSIP is the default Azure DNS IP used in localdns.
+	AzureDNSIP string = "168.63.129.16"
+)
+
+// LocalDNSProfile represents localdns configuration for agentpool nodes.
+type LocalDNSProfile struct {
+	State                string                        `json:"state,omitempty"`
```
yewmsft, line 2342: I think we should pass in mode here. Preferred, Required, Disabled

pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,116 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
...
+	return AzureDNSIP
+}
+
+// GetLocalDNSCPULimitInPercentage returns CPU limit in percentage unit that will be used in localdns systemd unit.
+func (a *AgentPoolProfile) GetLocalDNSCPULimitInPercentage() string {
+	if a.ShouldEnableLocalDNS() && a.LocalDNSProfile.CPULimitInMilliCores != nil {
+		// Convert milli-cores to percentage and return as formatted string
+		return fmt.Sprintf("%.1f%%", float64(*a.LocalDNSProfile.CPULimitInMilliCores)/10.0)
+	}
+	return DefaultLocalDNSCPULimitInPercentage
```
yewmsft, line 2398: why returning default value when a.ShouldEnableLocalDNS() is false?

pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,116 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
 }
 
 // SecurityProfile end.
+
+// ----------------------- Start of Changed related to localdns ------------------------------------------.
```
cameronmeissner, line 2321: typo: changes

pkg/agent/datamodel/types.go
SriHarsha001: Mode is what we get an an input from user. In AKS RP, we will decide if we can enable localdns on the nodepool and set the State to either Enabled or Disabled and send it to AgentBaker via NBC.

pkg/agent/datamodel/types.go
SriHarsha001: Mode is what we get an an input from user. In AKS RP, we will decide if we can enable localdns on the nodepool and set the State to either Enabled or Disabled and send it to AgentBaker via NBC.

yewmsft: I would use a different name for this. I would be very confusing since there is a state property in RP API called state.

pkg/agent/datamodel/types.go
SriHarsha001: thank you. Fixed the comment.

pkg/agent/datamodel/types.go
yewmsft: consider using Enable bool. Enable == true -> should enable. Enable == false --> should disable.

yewmsft: and using state = enabled / disabled on RP side to denote the result of the enablement.

yewmsft: Becaue, you can have Enable == true, but enabling operation failed, so on RP side, state shouls still be Disabled.

SriHarsha001: Using EnableLocalDNS as a bool and made the changes.

pkg/agent/datamodel/types.go
SriHarsha001: ok, made the changes and added tests.

pkg/agent/datamodel/types.go
SriHarsha001: Using EnableLocalDNS as a bool and made the changes.

pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,112 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
...
+	return AzureDNSIP
+}
+
+// GetLocalDNSCPULimitInPercentage returns CPU limit in percentage unit that will be used in localdns systemd unit.
+func (a *AgentPoolProfile) GetLocalDNSCPULimitInPercentage() string {
+	if a.LocalDNSProfile.CPULimitInMilliCores != nil {
+		// Convert milli-cores to percentage and return as formatted string.
+		return fmt.Sprintf("%.1f%%", float64(*a.LocalDNSProfile.CPULimitInMilliCores)/10.0)
+	}
+	return ""
```
yewmsft, line 2394: return default here？

pkg/agent/datamodel/types.go
```diff
@@ -2316,3 +2317,112 @@ func (s *SecurityProfile) GetPrivateEgressContainerRegistryServer() string {
...
+}
+
+// GetLocalDNSMemoryLimitInMB returns memory limit in MB that will be used in localdns systemd unit.
+func (a *AgentPoolProfile) GetLocalDNSMemoryLimitInMB() string {
+	if a.LocalDNSProfile.MemoryLimitInMB != nil {
+		// Return memory limit as a string with "M" suffix.
+		return fmt.Sprintf("%dM", *a.LocalDNSProfile.MemoryLimitInMB)
+	}
+	return ""
```
yewmsft, line 2403: same

pkg/agent/datamodel/types.go
yewmsft: or, will it fail if you pass in empty to the core file

SriHarsha001: I think I mis-understood the suggestions earlier. Made the change.

pkg/agent/datamodel/types.go
SriHarsha001: I think I mis-understood the suggestions earlier. Made the change.

yewmsft (approved):

PR #6173, opened 2025-04-09
chore: replace ciprod with latest and latest-1 of servercore

timmy-wright:
**What type of PR is this?**

/kind feature

This PR removes the ciprod container and replaces it with the current servercore container and the previous servercore container. This was agreed with the ciprod team as a way to support their ciprod rollouts.

It also works around a issue with the subscription the PR builds now run in.

servercore tags list is here: https://mcr.microsoft.com/v2/windows/servercore/tags/list

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -6,23 +6,28 @@
       "multiArchVersionsV2": [],
       "windowsVersions": [
         {
+          "comment": "should have latest version and natest version -1",
           "renovateTag": "registry=https://mcr.microsoft.com, name=windows/servercore",
-          "latestVersion": "ltsc2019",
+          "latestVersion": "10.0.17763.7136",
+          "previousLatestVersion": "10.0.17763.7009",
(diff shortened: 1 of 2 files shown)


parts/common/components.json
```diff
@@ -6,23 +6,28 @@
       "multiArchVersionsV2": [],
       "windowsVersions": [
         {
+          "comment": "should have latest version and natest version -1",
```
Devinwong, line 9: typo: natest

Devinwong (approved):

parts/common/components.json
timmy-wright: Annoyingly github auto-merged the PR even tho you'd written a comment! I don't like that behaviour. Will fix this in a future PR.

PR #6174, opened 2025-04-09
Add code to generate corefile

SriHarsha001:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/nodecustomdata.yml
+++ b/parts/linux/cloud-init/nodecustomdata.yml
@@ -429,3 +429,27 @@ write_files:
   owner: root
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+{{if ShouldEnableLocalDNS}}
+- path: /opt/azure/containers/localdns/localdns.corefile
+  permissions: "0644"
+  encoding: gzip
+  owner: root
+  content: !!binary |
+    {{GetGeneratedLocalDNSCoreFile}}
+
+- path: /etc/systemd/system/localdns.slice
+  permissions: "0644"
+  owner: root
+  content: |
+    [Unit]
+    Description=localdns Slice
+    DefaultDependencies=no
+    Before=slices.target
+    Requires=system.slice
+    After=system.slice
+
+    [Slice]
+    MemoryMax={{GetLocalDNSMemoryLimitInMB}}
+    CPUQuota={{GetLocalDNSCPULimitInPercentage}}
+{{- end}}
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -1005,6 +1005,22 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"InsertIMDSRestrictionRuleToMangleTable": func() bool {
 			return config.InsertIMDSRestrictionRuleToMangleTable
 		},
+		"ShouldEnableLocalDNS": func() bool {
+			return profile.ShouldEnableLocalDNS()
+		},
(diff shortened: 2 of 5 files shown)


pkg/agent/baker.go
```diff
@@ -1005,6 +1005,22 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"InsertIMDSRestrictionRuleToMangleTable": func() bool {
 			return config.InsertIMDSRestrictionRuleToMangleTable
 		},
+		"ShouldEnableLocalDNS": func() bool {
+			return profile.ShouldEnableLocalDNS()
+		},
+		"GetGeneratedLocalDNSCoreFile": func() string {
+			output, err := GenerateLocalDNSCoreFile(config, profile, localDNSCoreFileTemplateString)
+			if err != nil {
+				panic(err)
```
Devinwong, line 1014: I think we can just return the error like line 970 `GetSysctlContent` above. panic will cause the app to crash which is not desired in this case.

pkg/agent/baker_test.go
```diff
@@ -189,7 +189,575 @@ var _ = Describe("Assert generated customData and cseCmd", func() {
...
+				Expect(err).To(BeNil())
+				Expect(localDNSCoreFileGzippedBase64Decoded).ToNot(BeEmpty())
+
+				// Decompress the gzipped data.
+				localDNSCorefile, err := getGzipDecodedValue([]byte(localDNSCoreFileGzippedBase64Decoded))
+				Expect(err).To(BeNil())
+				Expect(localDNSCorefile).ToNot(BeEmpty())
+
+				expectedlocalDNSCorefile := `
+# ***********************************************************************************
```
Devinwong, line 621: can we try to reuse the previous expectedlocalDNSCorefile so that we don't need to define the same thing twice?
idea: expectedlocalDNSCorefile2= expectedlocalDNSCorefile + additionalConfig

Devinwong:
Please note that all the changes in this PR will also need to implement in Scriptless milestone 1 which is mostly under the folder aks-node-controller. We can wait for the feature to settle before implementing it in Scriptless.

pkg/agent/baker.go
SriHarsha001: ok sure, made the change.

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -428,4 +428,28 @@ write_files:
   encoding: gzip
   owner: root
   content: !!binary |
-    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
\ No newline at end of file
+    {{GetVariableProperty "cloudInitData" "customSearchDomainsScript"}}
+
+{{if ShouldEnableLocalDNS}}
```
yewmsft, line 433: @Devinwong How do we test custom data?

pkg/agent/baker_test.go
SriHarsha001: I have defined a string constant and using it in 2 test cases. For the other 2 tests, the output changes across the tests which needs to be tests for different values based on different inputs. So unable to templatize the expected output string. PTAL.

parts/linux/cloud-init/nodecustomdata.yml
Devinwong: Just checked the codes, it seems we don't have unit test for custom data now. If needed, it can be added in baker_test.go

yewmsft (approved):

Devinwong (approved):

PR #6175, opened 2025-04-09
Add code to update kubelet flag if localdns is enabled

SriHarsha001:
**What type of PR is this?**
feature

This PR has code changes to override '--cluster-dns' flag of kubelet, if LocalDNS is enabled for the agentpool.
If localdns is enabled for the agentpool, then Kubelet's --cluster-dns flag should be pointed to localdns clusterlistenerIP address. This is because, all the pods created by kubelet on the node will be pointing to clusterlistenerIP when localdns is enabled (ie etc/resolv.conf). In the next few PRs, we will see that ensureKubelet will be called after enableLocalDNS in cse_main file. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/pkg/agent/utils.go
+++ b/pkg/agent/utils.go
@@ -362,6 +362,12 @@ func GetOrderedKubeletConfigFlagString(config *datamodel.NodeBootstrappingConfig
 	sort.Strings(keys)
 	var buf bytes.Buffer
 	for _, key := range keys {
+		/* Override --cluster-dns if LocalDNS is enabled for the agentpool.
+		If localdns is enabled for the agentpool,
+		then Kubelet's --cluster-dns flag should be pointed to localdns clusterlistenerIP address. */
+		if key == "--cluster-dns" && profile.ShouldEnableLocalDNS() {
+			k[key] = profile.GetLocalDNSClusterListenerIP()
+		}
 		buf.WriteString(fmt.Sprintf("%s=%s ", key, k[key]))
 	}
 	return buf.String()
--- a/pkg/agent/utils_test.go
+++ b/pkg/agent/utils_test.go
@@ -673,6 +673,95 @@ var _ = Describe("Test GetOrderedKubeletConfigFlagString", func() {
 		actualStr := GetOrderedKubeletConfigFlagString(config)
 		Expect(expectedStr).To(Equal(actualStr))
 	})
+
+	// ------------------------------- Start of tests related to Localdns ---------------------------------------.
+	It("should return localdns clusterlistenerIP value for --cluster-dns when localdns is enabled", func() {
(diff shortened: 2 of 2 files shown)


pkg/agent/utils.go
```diff
@@ -362,6 +362,12 @@ func GetOrderedKubeletConfigFlagString(config *datamodel.NodeBootstrappingConfig
 	sort.Strings(keys)
 	var buf bytes.Buffer
 	for _, key := range keys {
+		/* Override --cluster-dns if LocalDNS is enabled for the agentpool.
+		If localdns is enabled for the agentpool,
+		then Kubelet's --cluster-dns flag should be pointed to localdns clusterlistenerIP address. */
+		if key == "--cluster-dns" && profile.ShouldEnableLocalDNS() {
```
cameronmeissner, line 368: @Devinwong @lilypan26 we should probably take a stance moving forward whether we want to modifying kubelet config at runtime within CSE or via AgentBaker / AKSNodeConfig helpers moving forward

for context, the changes I made for kubelet-serving modify kubelet flags within CSE so we don't have to duplicate the same logic across AgentBaker and AKSNodeConfig helpers, so wondering if we should consider the same here

cameronmeissner, line 368: do we also want to add this to TranslatedKubeletConfigFileFlags as well?

anujmaheshwari1: or if we want them to be done in RP itself like I did for parallel image pulls. 

@SriHarsha001 do we have a local dns validator in RP frontend? we can consider doing this there as well.

SriHarsha001: Could you please share a reference - I will take a look. The reason, I did here is because there are few places where kubelet flags were getting touched, so figured this was the last place before the kubelet data is sent to the node before enabling kubelet.

SriHarsha001: Discussed with Anuj regarding this. I will explore the option to move this logic out of AgentBaker and set the kubelet flag in AKS RP validator. I will not proceed with this PR for now. Thank you all for taking a look.

timmy-wright: It's also worth considering if this change should just apply to Linux or to both Linux and Windows. I'm not sure your desired scope of change.

yewmsft, line 368: I am a little bit confused about this, who is passing "--cluster-dns" config in? is it from RP vmss.go

yewmsft: Set this value in Default(), and validate the value in Validate()

yewmsft: It will be Linux only. And it will be set in localDns validator.

PR #6205, opened 2025-04-11
Add e2e test validations for localdns

SriHarsha001:
**What type of PR is this?**
feature

This PR adds e2e tests validations for localdns. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


e2e/validators.go
```diff
@@ -531,3 +530,23 @@ func ValidateTaints(ctx context.Context, s *Scenario, expectedTaints string) {
...
+
+// ValidateLocalDNSService checks if the localdns service is running and active.
+func ValidateLocalDNSService(ctx context.Context, s *Scenario, serviceName string) {
+	steps := []string{
+		"set -ex",
+		// Verify the localdns service is running and active.
+		fmt.Sprintf("(systemctl -n 5 status %s || true)", serviceName),
+		fmt.Sprintf("systemctl is-active %s", serviceName),
+	}
+	execScriptOnVMForScenarioValidateExitCode(ctx, s, strings.Join(steps, "\n"), 0, "command to restart localdns service failed")
```
cameronmeissner, line 542: nit: this isn't restarting localdns is it?

e2e/scenario_test.go
```diff
@@ -35,6 +38,8 @@ func Test_AzureLinuxV2_AirGap(t *testing.T) {
 			},
 			Validator: func(ctx context.Context, s *Scenario) {
 				ValidateDirectoryContent(ctx, s, "/run", []string{"outbound-check-skipped"})
+				ValidateLocalDNSService(ctx, s, LocalDNSService)
```
cameronmeissner, line 41: shouldn't we only do this for localDNS-enabled scenarios? 

also if we do want to do this for every scenario, would prefer you add the validation steps directly to the command validation logic rather than adding to each individual scenario here

e2e/validators.go
SriHarsha001: yeah, missed out to update the comment.

e2e/scenario_test.go
SriHarsha001: yes, I need to test for every scenario. I was not aware of this common validation logic in e2e/validation.go before. I had made the removed all the checks from scenario_test.go and added the check in ValidateCommonLinux.

Devinwong (approved):
lgtm

PR #6239, opened 2025-04-19
enable shellspec running in a container

### LINKED ISSUE #5322: shellspec unit tests can fail due to unpredictable dev environment
Shellspec unit tests currently use whatever shell happens to be installed on the developer's machine. 

The problem with this is that the scripts shellspec is supposed to be testing are specifically meant to be run with a bash shell. This means that if a developer is using a different shell, such as zsh, shellspec tests will end up failing in any case where the scripts being tested utilize shell syntax that is not exportable from bash.

Solution would be to have shellspec tests run within a container where we can control exactly which shell is available within the environment.

Devinwong:
**What type of PR is this?**
/kind test

Enable shellspec running in a container so that it can support all platforms (WSL, Mac, etc.)
Updated readme to reflect to the changes.

Fixes #5322

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


Makefile
```diff
@@ -77,9 +77,13 @@ validate-go:
 validate-shell:
 	@./.pipelines/scripts/verify_shell.sh
 
-.PHONY: shellspec # TODO: have this run in a container with a stable bash installation
+.PHONY: shellspec
 shellspec: bootstrap
```
Devinwong, line 81: check if I need to keep `bootstrap` here

Devinwong: No it's not needed. Removed

.github/workflows/shellspec.yaml
```diff
@@ -11,4 +11,9 @@ jobs:
         go-version: '1.24'
     - run: |
         make shellspec
-      name: Run shell/bash script unit tests with shellspec
\ No newline at end of file
+      name: Run shell/bash script unit tests with shellspec
+    - run: echo "Shell/Bash script unit tests failed. Please check the logs for details. To run the shellspec unit tests locally, please refer to ./spec/README.md"
```
cameronmeissner, line 15: are these necessary? should be able to get results from the `make shellspec` output right?

.github/workflows/shellspec.yaml
```diff
@@ -11,4 +11,9 @@ jobs:
...
+      if: failure()
+      name: Print failure message
+    - run: echo "Shell/Bash script unit tests completed successfully!"
```
cameronmeissner, line 18: I think this will run even in the failure case

cameronmeissner:
cool to see this working! I'm excited to try this out locally

.github/workflows/shellspec.yaml
Devinwong: Actually it won't. I tested with both cases.

.github/workflows/shellspec.yaml
Devinwong: `make shellspec` will run all the tests but without error details. It should be helpful for people to run a specific case with details, in terms of debugging.
 And I found that many people don't know there is a readme describing the unit test. That's why I put it here too.

.github/workflows/shellspec.yaml
Devinwong: Right it won't be printed.
`This is because GitHub Actions stops executing subsequent steps in a job by default when a step fails, unless you explicitly configure it to continue.`

cameronmeissner: sorry, I was referring to the `Print completion message` case, though I'm assumign you also double-chceked this

Devinwong: right I mean the completion message too 🙂

cameronmeissner: ah I see, what you pasted above probably applies to _all_ steps in the job rather than just the immediately proceeding one

cameronmeissner (approved):

PR #6248, opened 2025-04-23
Immediate exit with error during VHD build time

Devinwong:
**What type of PR is this?**
/kind bug

- Immediate exit with error during VHD build time
  - We were not always exit with error during VHD build time. Therefore, **some container images pull failures were ignored**. 
  - Turn on `set -euo pipefail` in install-dependencies.sh so that it will exit with error during VHD build time.
- Fix some bugs in the build scripts that populated some errors but were ignored.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -268,7 +268,13 @@ downloadContainerdWasmShims() {
     fi
 
     for shim in "${shims_to_download[@]}"; do
-        retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_wasm_filepath/containerd-shim-${shim}-${binary_version}-v1" "$containerd_wasm_url/containerd-shim-${shim}-v1" 2>&1 | tee $CURL_OUTPUT | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT) &
+        output_file="$containerd_wasm_filepath/containerd-shim-${shim}-${binary_version}-v1"
+        download_url="$containerd_wasm_url/containerd-shim-${shim}-v1"
+        retrycmd_if_failure 30 5 60 curl -fSLv -o "$output_file" "$download_url" 2>&1 | tee $CURL_OUTPUT &
```
Devinwong, line 273: We only want the `curl` command to be running in background (with syntax `&`).
Splitting them into multiple steps for easier debugging and can exit with the error code correctly.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -322,7 +328,13 @@ downloadSpinKube(){
         return 
     fi
     
-    retrycmd_if_failure 30 5 60 curl -fSLv -o "$containerd_spinkube_filepath/containerd-shim-spin-v2" "$containerd_spinkube_url/containerd-shim-spin-v2" 2>&1 | tee $CURL_OUTPUT | grep -E "^(curl:.*)|([eE]rr.*)$" && (cat $CURL_OUTPUT && exit $ERR_KRUSTLET_DOWNLOAD_TIMEOUT) &
+    output_file="$containerd_spinkube_filepath/containerd-shim-spin-v2"
+    download_url="$containerd_spinkube_url/containerd-shim-spin-v2"
+    retrycmd_if_failure 30 5 60 curl -fSLv -o "$output_file" "$download_url" 2>&1 | tee $CURL_OUTPUT &
```
Devinwong, line 333: Similarly with above, We only want the `curl` command to be running in background (with syntax `&`).
Splitting them into multiple steps for easier debugging and can exit with the error code correctly.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -651,11 +663,11 @@ pullContainerImage() {
     CONTAINER_IMAGE_URL=$2
     echo "pulling the image ${CONTAINER_IMAGE_URL} using ${CLI_TOOL}"
     if [[ ${CLI_TOOL} == "ctr" ]]; then
-        logs_to_events "AKS.CSE.imagepullctr.${CONTAINER_IMAGE_URL}" "retrycmd_if_failure 2 1 120 ctr --namespace k8s.io image pull $CONTAINER_IMAGE_URL" || (echo "timed out pulling image ${CONTAINER_IMAGE_URL} via ctr" && exit $ERR_CONTAINERD_CTR_IMG_PULL_TIMEOUT)
+        logs_to_events "AKS.CSE.imagepullctr.${CONTAINER_IMAGE_URL}" "retrycmd_if_failure 2 1 120 ctr --namespace k8s.io image pull $CONTAINER_IMAGE_URL" || (echo "timed out pulling image ${CONTAINER_IMAGE_URL} via ctr" && return $ERR_CONTAINERD_CTR_IMG_PULL_TIMEOUT)
```
Devinwong, line 666: Return the error codes and it will be caught from the previous level caller who calls `pullContainerImage` as a background process

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -272,11 +278,18 @@ ensureRunc() {
     # if on a vhd-built image, first check if we've cached the deb file
     if [ -f $VHD_LOGS_FILEPATH ]; then
         RUNC_DEB_PATTERN="moby-runc_*.deb"
-        RUNC_DEB_FILE=$(find ${RUNC_DOWNLOADS_DIR} -type f -iname "${RUNC_DEB_PATTERN}" | sort -V | tail -n1)
-        if [[ -f "${RUNC_DEB_FILE}" ]]; then
+        RUNC_DEB_FILES=$(find "${RUNC_DOWNLOADS_DIR}" -type f -iname "${RUNC_DEB_PATTERN}" 2>/dev/null || true)
```
Devinwong, line 281: Split it into multiple steps for easier debugging.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -228,7 +228,7 @@ retrycmd_get_tarball() {
     tar_retries=$1; wait_sleep=$2; tarball=$3; url=$4
     echo "${tar_retries} retries"
     for i in $(seq 1 $tar_retries); do
-        tar -tzf $tarball && break || \
+        [ -f $tarball ] && tar -tzf $tarball && break || \
```
cameronmeissner, line 231: I'd prefer we check existence before the loop and exit early if it doesn't exist

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -530,7 +530,10 @@ logs_to_events() {
         --arg EventTid    "0" \
         '{Timestamp: $Timestamp, OperationId: $OperationId, Version: $Version, TaskName: $TaskName, EventLevel: $EventLevel, Message: $Message, EventPid: $EventPid, EventTid: $EventTid}'
     )
-    echo ${json_string} > ${EVENTS_LOGGING_DIR}${eventsFileName}.json
+    if [ ! -f "${EVENTS_LOGGING_DIR}${eventsFileName}.json" ]; then
```
cameronmeissner, line 533: I believe we'd only need to check existence of `EVENTS_LOGGING_DIR`, `>` should create the file if the parent directory exists

as a result we could simplify this all down to `mkdir -p ${EVENTS_LOGGING_DIR}` without any conditional

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -154,9 +154,11 @@ installStandaloneContainerd() {
     local desiredVersion="${1:-}"
     #e.g., desiredVersion will look like this 1.6.26-5.cm2
     # azure-built runtimes have a "+azure" suffix in their version strings (i.e 1.4.1+azure). remove that here.
-    CURRENT_VERSION=$(containerd -version | cut -d " " -f 3 | sed 's|v||' | cut -d "+" -f 1)
-    # v1.4.1 is our lowest supported version of containerd
-    
+    # check if containerd command is available before running it
+    if command -v containerd &> /dev/null; then
```
cameronmeissner, line 158: `if which containerd &> /dev/null` also works

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
cameronmeissner, line 281: wouldn't `|| ""` make more sense here?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -272,11 +278,18 @@ ensureRunc() {
...
+        if [ -z "${RUNC_DEB_FILES}" ]; then
```
cameronmeissner, line 282:
```suggestion
        RUNC_DEB_FILE=""
        if [ ! -z "${RUNC_DEB_FILES}" ]; then
            RUNC_DEB_FILE=$(echo "${RUNC_DEB_FILES}" | sort -V | tail -n1)
        fi
```

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -457,7 +460,9 @@ EOF
 
 fi
 
-ls -ltr /opt/gpu/* >> ${VHD_LOGS_FILEPATH}
+if [ -d "/opt/gpu" ] && [ "$(ls -A /opt/gpu)" ]; then
```
cameronmeissner, line 463: why isn't `[ -d "/opt/gpu" ]` enough here?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -514,10 +519,16 @@ while IFS= read -r imageToBePulled; do
     image_pids+=($!)
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
     while [ "$(jobs -p | wc -l)" -ge "$parallel_container_image_pull_limit" ]; do
-      wait -n
+      wait -n || { 
+        ret=$?
+        echo "A background job pullContainerImage failed: ${ret}. Exiting..." >&2
+        kill -9 $(jobs -p) 2>/dev/null || echo "Failed to force kill some background jobs"
```
cameronmeissner, line 525: im confused by this, how do we know which background jobs we're killing? what timeout are we giving to `wait`?

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: oh. What the function doing here is to download the tarball if it doesn't exist. If it does exist, untar it.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
Devinwong: Correct. Both works

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
Devinwong: oh typo here. But I need to do some changes here. Let me follow up.

vhdbuilder/packer/install-dependencies.sh
Devinwong: I tested this. If the folder exists but empty, next line `ls -ltr` will return error. So we need the second check.

vhdbuilder/packer/install-dependencies.sh
Devinwong: What I did here was to kill all background jobs because we are anyway terminating the whole VHD build process. But let me make the changes just killing the `image_pids` to be more precise. 
As for timeout, according to copilot, `The wait -n command in Bash does not have a built-in timeout.` The `wait` will pass if the background job finishes or it will fail if the background job returns error.

parts/linux/cloud-init/artifacts/cse_helpers.sh
Devinwong: updated!

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
Devinwong: updated

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
Devinwong: updated the logic.

parts/linux/cloud-init/artifacts/cse_install.sh
djsly: im guessing the caller will return the same underlying error code ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -266,7 +267,9 @@ downloadAndInstallCriTools() {
 }
 
 echo "VHD will be built with containerd as the container runtime"
-updateAptWithMicrosoftPkg
+if [ "${OS}" = "${UBUNTU_OS_NAME}" ]; then
+  updateAptWithMicrosoftPkg
+fi
 capture_benchmark "${SCRIPT_NAME}_update_apt_with_msft_pkg"
```
djsly, line 273: this shuld be in the if statement maybe ?

vhdbuilder/packer/install-dependencies.sh
djsly: ❯ mkdir test
❯ ls -ltr test
total 0
❯ echo $?
0

wierd, I dont get an error code

❯ ls -ltr test/*
zsh: no matches found: test/*
❯ echo $?
1

the * does it

djsly (approved):

djsly (approved):

PR #6251, opened 2025-04-23
fix: correctly check exit code returned from retry operations

UtheMan:
Ensures that when one of the long running retry operations fails, we properly exit with appropriate exit code during CSE execution, removes unused helper functions


parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
```diff
@@ -6,6 +6,6 @@ source "${CSE_DISTRO_HELPERS_FILEPATH}"
 echo "  dns-search ${CUSTOM_SEARCH_DOMAIN_NAME}" | tee -a /etc/network/interfaces.d/50-cloud-init.cfg
 systemctl_restart 20 5 10 networking
```
UtheMan, line 7: Technically we should/could also exit from here, but not sure if it's a failure we want to propagate? 

In `cse_main.sh` we do:

```
if [ "${HAS_CUSTOM_SEARCH_DOMAIN}" == "true" ]; then
    "${CUSTOM_SEARCH_DOMAIN_FILEPATH}" > /opt/azure/containers/setup-custom-search-domain.log 2>&1 || exit $ERR_CUSTOM_SEARCH_DOMAINS_FAIL
fi
```

But in the current state of the script, I believe we would only report that error if the last line of the script fails somehow?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -161,7 +161,7 @@ installStandaloneContainerd() {
         echo "currently installed containerd version ${CURRENT_VERSION} is greater than (or equal to) target base version ${desiredVersion}. skipping installStandaloneContainerd."
     else
         echo "installing containerd version ${desiredVersion}"
-        removeContainerd
+        removeContainerd || exit $ERR_CONTAINERD_INSTALL_TIMEOUT
```
UtheMan, line 164: Is it okay for us to exit here? Otherwise the error from `retrycmd_if_failure` in `removeContainerd` is ignored

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -469,7 +469,7 @@ installCNI() {
         echo "WARNING: no cni-plugins components present falling back to hard coded download of 1.4.1. This should error eventually" 
         # could we fail if not Ubuntu2204Gen2ContainerdPrivateKubePkg vhd? Are there others?
         # definitely not handling arm here.
-        retrycmd_get_tarball 120 5 "${CNI_DOWNLOADS_DIR}/refcni.tar.gz" "https://${PACKAGE_DOWNLOAD_BASE_URL}/cni-plugins/v1.4.1/binaries/cni-plugins-linux-amd64-v1.4.1.tgz" || exit
+        retrycmd_get_tarball 120 5 "${CNI_DOWNLOADS_DIR}/refcni.tar.gz" "https://${PACKAGE_DOWNLOAD_BASE_URL}/cni-plugins/v1.4.1/binaries/cni-plugins-linux-amd64-v1.4.1.tgz" || exit $ERR_CNI_DOWNLOAD_TIMEOUT
```
UtheMan, line 472: We exit already, but looks like we didn't add an exit code here cc @zachary-bailey ?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
Devinwong: yes I think we should exit with error as you are doing here.

UtheMan: cc @aadhar-agarwal / @anujmaheshwari1

parts/linux/cloud-init/artifacts/cse_install.sh
Devinwong: yes I think exiting with this error is good. (Although technically it might not be really a timeout. But if that's the most matched error code we have now, I am also good with this)

parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
cameronmeissner: is this script called during CSE?

parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
```diff
@@ -6,6 +6,6 @@ source "${CSE_DISTRO_HELPERS_FILEPATH}"
...
 wait_for_apt_locks
-retrycmd_if_failure 10 5 120 apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit
+retrycmd_if_failure 10 5 120 apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit || exit 1
```
cameronmeissner, line 9: we probably want a generic package installation error code rather than `1` here

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -357,21 +357,6 @@ retrycmd_curl_file() {
         fi
     done
 }
-wait_for_file() {
```
cameronmeissner, line 360: just confirming these aren't being used?

parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
cameronmeissner, line 9: can we add a comment clarifying that the `exit 1` will be trapped by `exit $ERR_CUSTOM_SEARCH_DOMAINS_FAIL` within CSE?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -411,7 +411,7 @@ if [ $VALIDATION_ERR -ne 0 ]; then
 fi
 
 # Call enableLocalDNS to enable localdns if localdns profile has EnableLocalDNS set to true.
-logs_to_events "AKS.CSE.enableLocalDNS" enableLocalDNS
+logs_to_events "AKS.CSE.enableLocalDNS" enableLocalDNS || exit $?
```
UtheMan, line 414: cc @SriHarsha001 - this will make CSE exit with the exit code that is returned by `enableLocalDNS` if its `!=0`

SriHarsha001: yeah, sounds good.

parts/linux/cloud-init/artifacts/cse_install.sh
zachary-bailey: Yes, when I was making changes to the scripts I noticed there were some areas where we didn't have exit codes, I believe this was one of them. I left them as is.

UtheMan: Got it, i'll add it here then

zachary-bailey: Not sure that it would always be a timeout either, but its probably better to at least have *some* related exit code

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
djsly: install_timeout error for `remove` command ?

djsly (approved):

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
UtheMan: @djsly removed for now to align with how Ubuntu side handles this, we can adjust if needed later.

cameronmeissner (approved):

Devinwong (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -336,7 +336,7 @@ After=bind-mount.service
 EOF
 fi
 
-logs_to_events "AKS.CSE.ensureSysctl" ensureSysctl
+logs_to_events "AKS.CSE.ensureSysctl" ensureSysctl || exit $ERR_SYSCTL_RELOAD
```
anujmaheshwari1, line 339: qq: does ensureSysctl itself not have an exit code? wondering if we should stick to a pattern of 

ensureSomething || exit $?

as we did with enableLocalDNS

just a thought, not a hard block, but will help keep code uniform.

parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
anujmaheshwari1: I could be wrong, but I think we dont support custom search domain anymore, this is legacy code carried over from aks-engine. it isnt in the ManagedCluster data model as well

parts/linux/cloud-init/artifacts/cse_main.sh
UtheMan: It doesn't at the moment, ir uses `retrycmd_if_failure` and will return 1 if that fails, but no specific code there from what I can see.

parts/linux/cloud-init/artifacts/setup-custom-search-domains.sh
UtheMan: Ah okay, will leave this as it is for now then, we could remove the whole thing later on.

PR #6260, opened 2025-04-24
chore: Migrate acs-mirror to packages.aks.azure.com

smith1511:
**What type of PR is this?**
/kind deprecation

acs-mirror is being deprecated, move to packages.aks.azure.com.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/windows/windowscsehelper.ps1
```diff
@@ -206,6 +213,10 @@ function DownloadFileOverHttp {
         }
         [System.Net.ServicePointManager]::SecurityProtocol = $secureProtocols
 
+        $OriginalUrl = $Url
```
AbelHu, line 216: Should we simplify the logic?
1. Use the original url to download
2. Do nothing if the package can be found in the cache
3. Use the alternative endpoint if it needs to download the package and it fails in downloading the package with all retries.

bravebeaver: iirc  the current logic to 
- check whether the package exists in cache, match by versions
- download only if it is not cached

i guess we can change the function to be lazy invoked, aka only 
- when a download is required 
- when the original url does not work with retries, 
- resolve baseurl then and use it for the rest of the provision flow

parts/windows/windowscsehelper.ps1
```diff
@@ -472,4 +483,65 @@ function Logs-To-Event {
...
+    Param(
+        [Parameter(Mandatory = $true)][string]
+        $InitialUrl
+    )
+
+    # Replace domain based on the current package download FQDN
+    if (($global:PackageDownloadFqdn -eq "packages.aks.azure.com") -and ($InitialUrl -like "https://acs-mirror.azureedge.net/*")) {
+        $InitialUrl = $InitialUrl -replace "acs-mirror.azureedge.net", $global:PackageDownloadFqdn
+    } elseif (($global:PackageDownloadFqdn -eq "acs-mirror.azureedge.net") -and ($InitialUrl -like "https://packages.aks.azure.com/*")) {
+        $InitialUrl = $InitialUrl -replace "packages.aks.azure.com", $global:PackageDownloadFqdn
```
bravebeaver, line 543: it took me a while to reason the necessity of this branch.  this covers the firewall not updated case right? shall we add a comment to it? we might want to log it too

parts/windows/windowscsehelper.ps1
smith1511: Sorry I wrote two comments inline in VS Code (I was very impressed that it popped up my PR comments in the IDE), but they never actually got posted back here!

On the ordering of checks, I think either way there's potential performance implications during the cutover.

If we got with the logic in this PR, checking packages.aks..., first, then we're setup for the cutover and don't need a followup PR.  If we check the legacy endpoint first, then this logic will need to be reverted after Jun 11th.

> i guess we can change the function to be lazy invoked, aka only
>
> - when a download is required
> - when the original url does not work with retries,
> - resolve baseurl then and use it for the rest of the provision flow

I guess I just like the logic outside the download so that we can test it properly, then the download logic doesn't have to change.  If we do the lazy approach I think it makes it a bit harder to test?

parts/windows/windowscsehelper.ps1
smith1511: Correct - yes good idea re. comments.

parts/windows/windowscsehelper.ps1
smith1511: I've updated the logic to simplify at, and lazy load the URL on first download attempt.

e2e/vmss.go
```diff
@@ -178,14 +180,28 @@ $CollectedLogs=(Get-ChildItem . -Filter "*_logs.zip" -File)[0].Name
...
+.\azcopy.exe copy "network_config.txt" "$arg1/network_config.txt"
 `
 
 // extractLogsFromVMWindows runs a script on windows VM to collect logs and upload them to a blob storage
 // it then lists the blobs in the container and prints the content of each blob
 func extractLogsFromVMWindows(ctx context.Context, s *Scenario) {
 	if !s.T.Failed() {
 		s.T.Logf("skipping logs extraction from windows VM, as the test didn't fail")
-		return
+		//return
```
smith1511, line 204: Remove before merge

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -220,6 +220,13 @@ try
         return
     }
 
+    ### TODO (christis) remove this!!!
+    # disable packages to test
+    $hosts = "C:\\Windows\\System32\\drivers\\etc\\hosts"
```
smith1511, line 225: Remove before merge

parts/windows/windowscsehelper.ps1
```diff
@@ -187,9 +193,9 @@ function DownloadFileOverHttp {
     $fileName = [IO.Path]::GetFileName($cleanUrl)
 
     $search = @()
-    if (Test-Path $global:CacheDir) {
-        $search = [IO.Directory]::GetFiles($global:CacheDir, $fileName, [IO.SearchOption]::AllDirectories)
-    }
+    # if ($global:CacheDir -and (Test-Path $global:CacheDir)) {
+    #     $search = [IO.Directory]::GetFiles($global:CacheDir, $fileName, [IO.SearchOption]::AllDirectories)
```
smith1511, line 197: Revert

cameronmeissner (approved):

PR #6262, opened 2025-04-24
fix: remove bcc log from VHD

zachary-bailey:
**What type of PR is this?**

/kind regression

This PR removes the bcc installation log file from linux VHDs. This file is being removed because it is not necessary to keep on the VHDs in production and is also currently over-permissioned. Change requested by the Azure Linux team.

File is currently on Linux VHDs in production and over-permissioned.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


jepio (approved):
This makes sense, I found the same issue fixing CIS findings on Ubuntu.

cameronmeissner (approved):

PR #6295, opened 2025-04-25
applying -ne and -eq to numerical values

Devinwong:
**What type of PR is this?**
/kind cleanup

Applying -ne and -eq to numerical values comparison.
I used some patterns to scan through the .sh files and tried to change all of them. I could have missed some though. Fortunately not many places.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
+++ b/vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
@@ -101,7 +101,7 @@ installBpftrace() {
     local bpftrace_url="https://upstreamartifacts.azureedge.net/$bpftrace_bin/$version"
     local bpftrace_filepath="/usr/local/bin/$bpftrace_bin"
     local tools_filepath="/usr/local/share/$bpftrace_bin"
-    if [ "$(isARM64)" = 1 ]; then
+    if [ "$(isARM64)" -eq 1 ]; then
         # install bpftrace tool using default bpftrace apt package
         # the binary at "$bpftrace_url/$bpftrace_bin" is not for arm64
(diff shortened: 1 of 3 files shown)


Devinwong:
will do make generate after approval.

vhdbuilder/scripts/linux/ubuntu/tool_installs_ubuntu.sh
```diff
@@ -101,7 +101,7 @@ installBpftrace() {
     local bpftrace_url="https://upstreamartifacts.azureedge.net/$bpftrace_bin/$version"
     local bpftrace_filepath="/usr/local/bin/$bpftrace_bin"
     local tools_filepath="/usr/local/share/$bpftrace_bin"
-    if [ "$(isARM64)" = 1 ]; then
```
djsly, line 104: Did we have a bug here?!!  Shouldn't it be == ? lol

Devinwong: The old way was also fine. There are different ways to do a comparison. Some work some don't and it's easily getting confused. For this case specifically, 
- using `=` wasn't the best practice but still worked. Since 1 is numerical value, the best practice is using `-eq`. 
  - `=` is comparing strings and it will automatically convert the values to strings, even it's an integer. So for this case, even it was `=`, they would be treated as strings to compare.
- `==` is not POSIX compliant (means it's specific to BASH only thus less portable). But in terms of logic, it will work.

zachary-bailey (approved):
lgtm!

cameronmeissner (approved):

cameronmeissner (approved):

PR #6301, opened 2025-04-28
feat: return early from retrycmd_if_failure/retrycmd_silent when global CSE timeout is approaching

UtheMan:
Modifies retry logic to return  early from long running retry operations when CSE timeout is approaching


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
...
 ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file, not used, but have to define to avoid error "Error: failed to get user home directory: $HOME is not defined" 
 
+# Checks if the elapsed time since CSEStartTime exceeds a specified maximum duration or the default 13 minutes.
+# Default value is based on the global CSE timeout of 15 minutes.
+# functions that use this should handle exit code 1 from this function to gracefully handle a "global" cse timeout (mean to avoid random 124 exit codes with no context)
+check_cse_timeout() {
+    shouldLog=${1:-true}
+    maxDurationSeconds=780 # 780 seconds = 13 minutes
+
+    if [[ -z "$CSE_STARTTIME_FORMATTED" ]]; then
```
cameronmeissner, line 181: again, new SC rules incoming:

```suggestion
    if [[-z "$CSE_STARTTIME_FORMATTED" ]; then
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
...
+        if [[ $shouldLog == "true" ]]; then
```
cameronmeissner, line 182:
```suggestion
        if [ "$shouldLog" = "true" ]; then
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
...
+            echo "Warning: CSE_STARTTIME_FORMATTED environment variable is not set."
+        fi
+        # Return 0 to avoid CSE errors in case something went wrong when setting the start time in cse_start.sh
+        # TODO mumanski - or should we actually error here?
+        return 0
+    fi
+
+    cse_start_seconds=$(date -d "$CSE_STARTTIME_FORMATTED" +%s)
+    if [[ $? -ne 0 ]]; then
```
cameronmeissner, line 191:
```suggestion
    if [ "$?" -ne 0 ]; then
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
...
+        if [[ $shouldLog == "true" ]]; then
```
cameronmeissner, line 192:
```suggestion
        if [ "$shouldLog" = "true" ]; then
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
 ORAS_OUTPUT=/tmp/oras_verbose.out
 ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file, not used, but have to define to avoid error "Error: failed to get user home directory: $HOME is not defined" 
 
+# Checks if the elapsed time since CSEStartTime exceeds a specified maximum duration or the default 13 minutes.
```
UtheMan, line 174: Was thinking about adding an argument to specify the timeout here, but later removed it - not sure if its something we want - could potentially get messy if people start calling this in other places with different timeouts?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
...
+        fi
+        return 0
+    fi
+
+    # Calculate elapsed time based on seconds since epoch
+    elapsed_seconds=$(($(date +%s) - cse_start_seconds))
+
+    # Check if elapsed time exceeds the maximum duration
+    if [[ $elapsed_seconds -gt $maxDurationSeconds ]]; then
+        if [[ $shouldLog == "true" ]]; then
```
cameronmeissner, line 203: you get the idea :D

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
 ORAS_OUTPUT=/tmp/oras_verbose.out
 ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file, not used, but have to define to avoid error "Error: failed to get user home directory: $HOME is not defined" 
 
+# Checks if the elapsed time since CSEStartTime exceeds a specified maximum duration or the default 13 minutes.
+# Default value is based on the global CSE timeout of 15 minutes.
+# functions that use this should handle exit code 1 from this function to gracefully handle a "global" cse timeout (mean to avoid random 124 exit codes with no context)
+check_cse_timeout() {
+    shouldLog=${1:-true}
+    maxDurationSeconds=780 # 780 seconds = 13 minutes
```
cameronmeissner, line 179: can we leave a comment on why 13 minutes based on what we talked about last week?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,8 +171,47 @@ CURL_OUTPUT=/tmp/curl_verbose.out
 ORAS_OUTPUT=/tmp/oras_verbose.out
 ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file, not used, but have to define to avoid error "Error: failed to get user home directory: $HOME is not defined" 
 
+# Checks if the elapsed time since CSEStartTime exceeds a specified maximum duration or the default 13 minutes.
+# Default value is based on the global CSE timeout of 15 minutes.
+# functions that use this should handle exit code 1 from this function to gracefully handle a "global" cse timeout (mean to avoid random 124 exit codes with no context)
+check_cse_timeout() {
+    shouldLog=${1:-true}
```
cameronmeissner, line 178:
```suggestion
    shouldLog="${1:-true}"
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -191,6 +230,14 @@ _retrycmd_internal() {
             break # Success, exit loop
         fi
 
+        # TODO mumanski - handle this in different places/modify error messages?
+        # Check if CSE timeout is approaching - exit early to avoid 124 exit code from the global timeout
+        check_cse_timeout $shouldLog
```
cameronmeissner, line 235:
```suggestion
        if ! check_cse_timeout $shouldLog; then ...
```

cameronmeissner: actually make sure this works with the `return` instead of `exit`, I'm assuming it does tohugh

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -191,6 +230,14 @@ _retrycmd_internal() {
...
+        if [[ $? -ne 0 ]]; then
+            echo "CSE timeout approaching, exiting early." >&2
+            return 2
```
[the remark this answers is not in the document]
UtheMan: Yeah the idea is that this will return 1 or 2 in case of errors and callers are supposed to handle those codes in whatever way they deem correct. So for example different handlers for code 1 (ran out of retries and ending) and for code 2 (ending because global timeout is approaching)

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: Added something, let me know if that works

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: Changed and it is tested by `returns 2 and times out when retrycmd_internal exceeds the CSE timeout` in `cse_retry_helpers_spec.sh`

Devinwong (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
bravebeaver: can we please have a link to the 15 min global timeout? as a reminder to when the global timeout changes (may be double linked comments ⛓️‍💥)

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -2,6 +2,7 @@
 
 CSE_STARTTIME=$(date)
 CSE_STARTTIME_FORMATTED=$(date +"%F %T.%3N")
+export CSE_STARTTIME_FORMATTED # Export for child processes
```
bravebeaver, line 5: why do we export formatted date and parse it later, rather than just export the datetime? just being curious..

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -171,45 +171,81 @@ CURL_OUTPUT=/tmp/curl_verbose.out
...
+    # Calculate elapsed time based on seconds since epoch
+    elapsedSeconds=$(($(date +%s) - cseStartSeconds))
+    if [ "$elapsedSeconds" -gt "$maxDurationSeconds" ]; then
+        if [ "$shouldLog" = "true" ]; then
+            echo "Error: CSE has been running for $elapsedSeconds seconds, exceeding the limit of $maxDurationSeconds seconds." >&2
+        fi
+        return 1
+    fi
+
+    return 0
```
bravebeaver, line 206: do i read it right that this is a best effort approach? aka if i do not invoke the method correctly it behaves as if it has not been invoked at all.

parts/linux/cloud-init/artifacts/cse_start.sh
UtheMan: Good point, just what I did quickly when first adding this change, changed this to export

`export CSE_STARTTIME_SECONDS=$(date -d "$CSE_STARTTIME_FORMATTED" +%s)` from `cse_start.sh` immediately - reduces amount of things that `check_cse_timeout` has to do.

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: Adding extra details here and a corresponding comment in `cse_start.sh`

UtheMan:
Seeing A100 GPU e2e failures only again.

PR #6318, opened 2025-04-30
refactor: don't block on pre-start script termination when starting kubelet.service

cameronmeissner:
**What type of PR is this?**

to prevent CSE latency from being impacted by kubelet credential validation, we now start kubelet via systemctl with the `--no-block` option. This means that we will no longer block on waiting for all of kubelet's pre-start scripts terminating successfully before considering kubelet as having started.

We still perform a status check after starting kubelet to ensure that startup operation is successfully queued by systemd.

For now we check on the status immediately after starting kubelet, though the new systemctlEnableAndStartNoBlock also takes in a status_check_wait parameter which will have the executing script sleep for a certain amount of time before checking on the status of the unit. We might consider adding a delay to the kubelet start, say 1 or 2 seconds to ensure that the majority of pre-start scripts have terminated successfully, though for now we aren't doing this since it's not quite deterministic how long we'd need to wait to guarantee that we're checking kubelet's status after all pre-start scripts except credential validation have terminated successfully.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -423,35 +428,79 @@ _systemctl_retry_svc_operation() {
...
     test "$(printf '%s\n' "$@" | sort -rV | head -n 1)" = "$1"
 }
 
 systemctlEnableAndStart() {
     service=$1; timeout=$2
-    systemctl_restart 100 5 $2 $1
+    systemctl_restart 100 5 $timeout $service
     RESTART_STATUS=$?
-    systemctl status $1 --no-pager -l > /var/log/azure/$1-status.log
+    systemctl status $service --no-pager -l > /var/log/azure/$service-status.log
```
djsly, line 460: could be safer to use ${service} for the filename

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
```diff
@@ -40,27 +40,16 @@ logs_to_events() {
     fi
 }
 
-validateBootstrapKubeconfig() {
+validate() {
     local kubeconfig_path=$1
+    
+    # extract the bootstrap token directly from bootstrap-kubeconfig since we can't use LoadCredential (not supported on 1804)
```
djsly, line 46: why do we care about 1804 ?

djsly (approved):

parts/linux/cloud-init/artifacts/validate-kubelet-credentials.sh
cameronmeissner: at the moment we're still building 1804 VHDs and bootstrapping 1804 nodes, so we kinda still need to deal with it - once it's completely removed/deprecated we could come back to this

PR #6329, opened 2025-05-01
feat: add early exit and consolidate retry to curl helpers

UtheMan:
Consolidates curl file helpers and adds early exit from CSE logic to them


spec/parts/linux/cloud-init/artifacts/cse_retry_helpers_spec.sh
```diff
@@ -113,6 +114,99 @@ Describe 'long running cse helper functions'
             End
         End
 
+        Describe 'file curl'
```
Devinwong, line 117: The tests are awesome!

Devinwong (approved):
lgtm.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -267,22 +267,40 @@ retrycmd_nslookup() {
...
-    for i in $(seq 1 $tar_retries); do
-        [ -f "$tarball" ] && tar -tzf "$tarball" && break || \
-        if [ "$i" -eq "$tar_retries" ]; then
+
+_retry_file_curl_internal() {
+    retries=$1; waitSleep=$2; timeout=$3; filePath=$4; url=$5; checksToRun=( "${@:6}" )
+    echo "${retries} file curl retries"
+    for i in $(seq 1 $retries); do 
+        # Use eval to execute the checksToRun string as a command
+        ( eval "$checksToRun" ) && break || if [ "$i" -eq "$retries" ]; then
```
juan-lee, line 276: Is it possible for checksToRun not to be set?

cameronmeissner, line 276: nit: redundant elif since we're returning

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -267,22 +267,40 @@ retrycmd_nslookup() {
...
 }
-retrycmd_get_tarball() {
-    tar_retries=$1; wait_sleep=$2; tarball=$3; url=$4
-    echo "${tar_retries} retries"
-    for i in $(seq 1 $tar_retries); do
-        [ -f "$tarball" ] && tar -tzf "$tarball" && break || \
-        if [ "$i" -eq "$tar_retries" ]; then
+
+_retry_file_curl_internal() {
+    retries=$1; waitSleep=$2; timeout=$3; filePath=$4; url=$5; checksToRun=( "${@:6}" )
```
cameronmeissner, line 272: non-blocking: did we decide to move towards camelCase? I'm fine with it I just think we should be consistent and not mix with snake_case if possible

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner, line 276: im not sure if this chaining will work exactly how you want it, I'd probably wrap the `( eval "$checksToRun" ) && break` in another pair of parens

cameronmeissner: will eval just no-op if passed an empty string? if not then we should deefinitely default it, caller could easily mistakenly not set it

UtheMan: Yeah it will be treated as an empty string and eval will return 0, breaking the retry loop. 
Adding an extra test + comment explaining what `checkToRun` are supposed to be

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: We did not, it is kind of all over the place right now, we could maybe agree on one style and refactor? Because i've seen both versions being used interchangeably

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: I think it should work fine, see for example 

```
Describe 'retry_file_curl_internal'
    It "returns 1 if checksToRun fail and retries are exhausted"
```

UtheMan: Fixed (i think)

UtheMan:
Windows e2e failing for now, bypassing

PR #6330, opened 2025-05-02
feat: add SSH service validation and ensure proper configuration on Ubuntu

bravebeaver:
**What type of PR is this?**

/kind regression

after ubuntu 2210 the ssh service is changed to socket based activation. we need it always.

**Requirements**:

- [x] conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [x] adds unit tests

https://discourse.ubuntu.com/t/sshd-now-uses-socket-based-activation-ubuntu-22-10-and-later/30189

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -910,4 +913,28 @@ oras_login_with_kubelet_identity() {
...
+    if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+        # for Ubuntu 22.10+ or newer uses socket activation
+        if semverCompare "$OS_VERSION" "22.10"; then
+            # For Ubuntu 22.10+ with socket-based SSH activation
+            if systemctl is-active --quiet ssh.socket; then
+                # Socket is active, disable it and switch to service-based activation
+                systemctl disable --now ssh.socket
+                # Remove the socket-based configuration file if present
+                if [ -f /etc/systemd/system/ssh.service.d/00-socket.conf ]; then
+                    rm /etc/systemd/system/ssh.service.d/00-socket.conf
```
cameronmeissner, line 927: this is a lot of indentation, would prefer an "exit-fast" approach  instead, something like:

```
if [ "$OS" != "$UBUNTU_OS_NAME" ]; then
  return 0
fi
if ! semverCompare "$OS_VERSION" "22.10"; then
  return 0
fi
if ! systemctl is-active --quiet ssh.socket; then
  return 0
fi

systemctl disable --now ssh.socket

....

systemctlEnableAndStart ssh 30 || return $ERR_SYSTEMCTL_START_FAIL

```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -910,4 +913,28 @@ oras_login_with_kubelet_identity() {
     echo "successfully logged in to acr '$acr_url' with identity token"
 }
 
+ensureSSHService() {
+    # If not Ubuntu, no changes needed
+    if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+        # for Ubuntu 22.10+ or newer uses socket activation
+        if semverCompare "$OS_VERSION" "22.10"; then
```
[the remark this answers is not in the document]
bravebeaver: yes. i think only since 2210 ubuntu changes ssh to socket based activation

e2e/scenario_test.go
```diff
@@ -1795,3 +1795,51 @@ func Test_Ubuntu2404ARM(t *testing.T) {
...
+			Description: "Tests that SSH service is properly configured and running on Ubuntu 24.04",
+			Config: Config{
+				Cluster: ClusterKubenet,
+				VHD:     osConfig.vhd,
+				BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+					// No specific configuration needed - testing default behavior
+				},
+				Validator: func(ctx context.Context, s *Scenario) {
+					// Verify SSH service is active and running
+					ValidateSystemdUnitIsRunning(ctx, s, "ssh")
```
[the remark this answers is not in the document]
bravebeaver: totally!

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -915,4 +918,33 @@ oras_login_with_kubelet_identity() {
     echo "successfully logged in to acr '$acr_url' with identity token"
 }
 
+ensureSSHService() {
```
cameronmeissner, line 921: if you're up for it please add a shellspec test for this

e2e/validation.go
```diff
@@ -95,6 +95,23 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 	}
 }
 
+func ValidateSSHServiceEnabled(ctx context.Context, s *Scenario) {
```
cameronmeissner, line 98: are we calling this somewhere?

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner, line 921: now that I think about it, can we do this at build-time instead? I don't think we need any runtime/cluster specific config to know when to enable/disable SSH right?

(doing it this way will also make it easier to verify that ssh is enabled in actual nodes)

bravebeaver: totally! now moved to tools_install section

e2e/scenario_test.go
```diff
@@ -227,8 +227,7 @@ func Test_AzureLinuxV2_GPUAzureCNI(t *testing.T) {
 			},
 			Validator: func(ctx context.Context, s *Scenario) {
 			},
-		},
-	})
```
djsly, line 231: those changes form Copilot are Wrong I got them aswell and it breaks the e2es after. I'm not sure why he's lost...

e2e/validators.go
```diff
@@ -588,3 +588,12 @@ func ValidateLocalDNSResolution(ctx context.Context, s *Scenario) {
 	assert.Contains(s.T, execResult.stdout.String(), "status: NOERROR")
 	assert.Contains(s.T, execResult.stdout.String(), "SERVER: 169.254.10.10")
 }
+
+// ExecuteCommandWithOutput runs a shell command on the VM under test and returns its output as a string
+func ExecuteCommandWithOutput(ctx context.Context, s *Scenario, cmd string) string {
```
djsly, line 593: didnt we already have a helper function that does this. ?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -436,7 +436,7 @@ sysctl_reload() {
     retrycmd_silent $1 $2 $3 "false" sysctl --system
 }
 version_gte() {
-  test "$(printf '%s\n' "$@" | sort -rV | head -n 1)" == "$1"
+  test "$(printf '%s\n' "$@" | sort -rV | head -n 1)" = "$1"
```
djsly, line 439: I removed this function in my PR since it's not users anywhere

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -127,6 +127,11 @@ fi
 if ! isMarinerOrAzureLinux "$OS"; then
   overrideNetworkConfig || exit 1
   disableNtpAndTimesyncdInstallChrony || exit 1
+
+  # Configure SSH service during VHD build for Ubuntu 22.10+
+  if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+    configureSSHService || echo "SSH Service configuration failed, but continuing VHD build"
```
djsly, line 133: I'm wondering if we could turn this `echo` into a ADO Warning such that it makes the build step "yellow"

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -125,6 +125,8 @@ fi
 
 logs_to_events "AKS.CSE.disableSystemdResolved" disableSystemdResolved
 
+# ensureSSHService has been moved to VHD build phase to improve node provisioning performance
```
djsly, line 128: where is `ensureSSHService` getting removed ?  I only see the new comment getting added.

Is there a configuration that we would want to be configured at bootstraptime for the SSH service ?

e2e/scenario_test.go
bravebeaver: always check...😂

spec/parts/linux/cloud-init/artifacts/cse_helpers_spec.sh
```diff
@@ -336,4 +336,87 @@ Describe 'cse_helpers.sh'
             The output line 1 should equal "KUBE_BINARY_URL is formatted unexpectedly, will use the kubernetes version as binary version: v$KUBERNETES_VERSION"
         End
     End
+
+    Describe 'configureSSHService'
```
Devinwong, line 340: It's great to see the unit tests!

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -130,6 +130,9 @@ if ! isMarinerOrAzureLinux "$OS"; then
 fi
 capture_benchmark "${SCRIPT_NAME}_validate_container_runtime_and_override_ubuntu_net_config"
 
+# Configure SSH service during VHD build for Ubuntu 22.10+
+configureSSHService "$OS" "$OS_VERSION" || echo "##vso[task.logissue type=warning]SSH Service configuration failed, but continuing VHD build"
+
```
Devinwong, line 135: I think we can also add `capture_benchmark` to capture the timing for future use. If it's very short, that's idea. If one day we find it unusually long, that means there must be a bug somewhere related.

Devinwong (approved):

PR #6331, opened 2025-05-02
chore: Update chrony.conf with new NTP settings

phealy:
This PR updates the chrony.conf file on Ubuntu and adds support for writing it out on Azure Linux in order to add support for using the time.windows.com backend sources via NTP. This works around a current problem in Azure that can cause the PHC (clock sync from the hypervisor) to be less accurate than desired; until that is fixed, the NTP sources will be preferred if they are available.

**What type of PR is this?**
/kind feature

See above

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
AKS Linux nodes will now attempt to sync NTP time from time.windows.com backends to increase clock accuracy. If these backends are unavailable the ptp sync from the host operating system will be used.
```


vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -115,6 +115,98 @@ disableTimesyncd() {
     systemctl stop systemd-timesyncd || exit 1
     systemctl disable systemd-timesyncd || exit 1
     systemctl mask systemd-timesyncd || exit 1
+
+    # Update chrony configuration
+    cat > /etc/chrony/chrony.conf <<EOF
```
djsly, line 120: this file is also created in `disableNtpAndTimesyncdInstallChrony` 

shoudl we simply update the lines we want to update instead of doing a full redrop ?

djsly, line 120: is this a big change for AzLinux ? we were relying on the default config from the OS before I guess.

phealy: It's a minor change, the config is 90% what was there before. The old ubuntu config file was copied from the Azure Linux default.

PR #6343, opened 2025-05-05
chore: enable systemd watchdog for kubelet on k8s >=1.32

djsly:
**What type of PR is this?**

/kind feature

This enables the newly supported systemd watchdog in kubelet on k8s 1.32, which will force systemd to restart kubelet if it becomes unresponsive

**Release note**:

```
none
```


pkg/agent/testdata/RawUbuntuContainerd/line70.sh
```diff
@@ -494,6 +494,13 @@ ensureKubelet() {
         logs_to_events "AKS.CSE.ensureKubelet.setKubeletNodeIPFlag" setKubeletNodeIPFlag
     fi
 
+    if semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.32.0"; then
+        tee "/etc/systemd/system/kubelet.service.d/10-watchdog.conf" > /dev/null <<'EOF'
+[Service]
+WatchdogSec=30s
```
cameronmeissner, line 500: I'd say it's worth adding a step in common linux validation of agentbaker e2e's to ensure this drop-in exists on k8s 1.32+

cameronmeissner: [GetAutoUpgradeScheduleByResource](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#WatchdogSec=)

should we set NotifyAccess= as well?

djsly: regarding NotifyAccess, I think the default is valid. it's not set on GCE

https://github.com/kubernetes/kubernetes/pull/127566/files

e2e/aks_model.go
```diff
@@ -19,6 +19,37 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/privatedns/armprivatedns"
 )
 
+// getLatestGAKubernetesVersion returns the latest GA Kubernetes version for the given location.
+func getLatestGAKubernetesVersion(location string) (string, error) {
+	versions, err := config.Azure.AKS.ListKubernetesVersions(context.Background(), location, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to list Kubernetes versions: %w", err)
+	}
+	if versions.Values == nil || len(versions.Values) == 0 {
```
cameronmeissner, line 28: `len` will handle nil slices and return 0

```suggestion
	len(versions.Values) == 0 {
```

e2e/aks_model.go
```diff
@@ -19,6 +19,37 @@ import (
...
+		return "", fmt.Errorf("no Kubernetes versions available")
+	}
+	for _, k8s_version := range versions.Values {
```
cameronmeissner, line 31: nit: please use camelCase to be consistent with the rest of the repo

e2e/aks_model.go
```diff
@@ -19,6 +19,37 @@ import (
...
+		if k8s_version.Version == nil {
+			continue
+		}
+		if k8s_version.IsPreview == nil || !*k8s_version.IsPreview {
```
cameronmeissner, line 35: is there also an `IsLTS` flag we need to check for?

cameronmeissner: I might be missing something, though with the current logic it seems like we aren't guaranteed to get the latest? just the first non-preview version returned by the `ListKubernetesVersions` call

e2e/aks_model.go
```diff
@@ -19,6 +19,37 @@ import (
...
+}
+
+// getLatestKubernetesVersionClusterModel returns a cluster model with the latest GA Kubernetes version.
+func getLatestKubernetesVersionClusterModel(name string) (*armcontainerservice.ManagedCluster, error) {
+	version, err := getLatestGAKubernetesVersion(config.Config.Location)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get latest GA Kubernetes version: %w", err)
+	}
+	model := getBaseClusterModel(name)
+	model.Properties.KubernetesVersion = to.Ptr(version)
```
cameronmeissner, line 49: I believe it gets defaulted AKS RP anyways though I'd consider also setting `model.Properties.AgentPoolProfiles[0].OrchestratorVersion = to.Ptr(version)` as well

e2e/config/vhd.go
```diff
@@ -264,3 +272,17 @@ func (id VHDResourceID) Short() string {
 	}
 	return str
 }
+
+func GetRandomAmd64VHD() *Image {
```
cameronmeissner, line 276: nit: `GetLinuxAMD64VHD()`

pkg/agent/const.go
```diff
@@ -86,12 +86,7 @@ const (
 	ensureNoDupEbtablesService = "linux/cloud-init/artifacts/ensure-no-dup.service"
 
 	// drop ins.
```
cameronmeissner, line 88: nit: can probably drop this comment

e2e/aks_model.go
cameronmeissner: can we also log out the chosen k8s version for easier debugging?

djsly: All future versions are LTS now...

e2e/aks_model.go
djsly: no need, like you said, the AgentPool versionfollows the control-plane version :)

e2e/config/vhd.go
djsly: GetRandomLinuxAMD64VHD, I want to keep Random in the func name

e2e/aks_model.go
```diff
@@ -19,6 +20,68 @@ import (
...
+	t.Logf("Available Kubernetes versions for location %s:", location)
+	for _, k8sVersion := range versions.Values {
+		if k8sVersion == nil {
+			continue
+		}
+		t.Logf("- %s", *k8sVersion.Version)
+
+		// Skip preview versions
+		if k8sVersion.IsPreview != nil && *k8sVersion.IsPreview {
+			t.Log(" - - is in preview, skipping")
```
cameronmeissner, line 44: missing the '%s' here?

e2e/aks_model.go
```diff
@@ -19,6 +20,68 @@ import (
...
+			t.Logf("- - %s", patchVersion)
+			// Initialize latestVersion with first GA version found
+			if latestPatchVersion == "" {
+				latestPatchVersion = patchVersion
+				t.Logf(" - - first latest found, updating to: %s", latestPatchVersion)
+				continue
+			}
+			// Compare versions
+			isGreater := agent.IsKubernetesVersionGe(patchVersion, latestPatchVersion)
+			if isGreater {
```
cameronmeissner, line 60: nit:

```suggestion
			if agent.IsKubernetesVersionGe(patchVersion, latestPatchVersion) {
```

pkg/agent/testdata/RawUbuntuContainerd/line70.sh
```diff
@@ -494,6 +494,14 @@ ensureKubelet() {
         logs_to_events "AKS.CSE.ensureKubelet.setKubeletNodeIPFlag" setKubeletNodeIPFlag
     fi
 
+    if semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.32.0"; then
+        tee "/etc/systemd/system/kubelet.service.d/10-watchdog.conf" > /dev/null <<'EOF'
+[Service]
+WatchdogSec=60s
+NotifyAccess=all
```
cameronmeissner, line 501: I guess we can probably get rid of the NotifyAccess line now that we're back to having systemd fork kubelet directly

e2e/aks_model.go
djsly: no it under the version 
- 1.27.1
- - is in preview, skipping
- 1.27.2
- -  is in preview, skipping

cameronmeissner (approved):

zachary-bailey (approved):

PR #6362, opened 2025-05-09
2025 05B image

zylxjtu:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -8,26 +8,26 @@
         {
           "comment": "should have latest version and natest version -1",
           "renovateTag": "registry=https://mcr.microsoft.com, name=windows/servercore",
-          "latestVersion": "10.0.17763.7136",
-          "previousLatestVersion": "10.0.17763.7009",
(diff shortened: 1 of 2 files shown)


parts/common/components.json
```diff
@@ -8,26 +8,26 @@
         {
           "comment": "should have latest version and natest version -1",
           "renovateTag": "registry=https://mcr.microsoft.com, name=windows/servercore",
-          "latestVersion": "10.0.17763.7136",
-          "previousLatestVersion": "10.0.17763.7009",
+          "latestVersion": "ltsc2019",
```
timmy-wright, line 11: For these, I'd prefer us to specify the actual version rather than a floating tag. This will make it more obvious which we are caching on a VHD release.

Ideally, the pipeline that updates the VHD base version would do this for us. I might update it to do that sometime.

timmy-wright (changes requested):

PR #6363, opened 2025-05-09
fix: return 216 when enable localdns failed

yewmsft:
**What type of PR is this?**
/kind bug

Previously, enable local dns failure will always return 1. Returning 216 instead for better error handling in RP.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -951,8 +951,8 @@ enableLocalDNS() {
     local enable_localdns_result=$?
 
     if [ "$enable_localdns_result" -ne 0 ]; then
```
cameronmeissner, line 953: how about we just simplify this whole thing to:

```suggestion
    if ! systemctlEnableAndStart localdns 30; then
      echo "Enable localdns failed."
      return $ERR_LOCALDNS_FAIL
    fi
```

cameronmeissner: the `return 0` at the end is also redundant

yewmsft: good point

cameronmeissner (approved):

SriHarsha001 (approved):

cameronmeissner (approved):

PR #6365, opened 2025-05-12
fix: ignore unnecessary repo config when using custom repo service

YaoC:
**What type of PR is this?**

/kind bug

For ubuntu, the `Allowed-Origins` for unattended-upgrades is:

```
Unattended-Upgrade::Allowed-Origins {
	"${distro_id}:${distro_codename}"; // e.g. ubuntu:jammy
	"${distro_id}:${distro_codename}-security"; // e.g. ubuntu:jammy-security
};
```

When using custom repo service, the unrelated config in `/etc/apt/sources.list.d` may cause upgrade to fail because the custom repo service doesn't provide those metadata. So we should ignore them.

For azurelinux, we no longer specify specific repos, so that we can flexibly handle new added repos.

Fixes #

This feature is not yet enabled.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -102,6 +102,12 @@ if [ -z "${live_patching_repo_service}" ]; then
     # No live patching repo service annotation, so we need to change to use the ubuntu snapshot repo
     # e.g. replace http://10.224.0.5/ubuntu/ with https://snapshot.ubuntu.com/ubuntu/20250318T000000Z
     sed -i 's/http:\/\/[0-9]\+.[0-9]\+.[0-9]\+.[0-9]\+\/ubuntu\//https:\/\/snapshot.ubuntu.com\/ubuntu\/'"${golden_timestamp}"'/g' ${source_list_path}
+    # unignore repo in /etc/apt/sources.list.d
```
yewmsft, line 105: super nit: "unignore" feels weird

yewmsft: consider "restore" or "reactivate"

yewmsft (approved):

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
YaoC: Haha, copilot helped me for this. 😂 Updated.

cameronmeissner (approved):

Devinwong:
Does this only affect upgrade scenario? May I know if there is a test to cover this scenario?

YaoC:
> Does this only affect upgrade scenario? May I know if there is a test to cover this scenario?

Yes, I think so. Now we just run it manually. Let me consider automating it.

Devinwong (approved):

PR #6371, opened 2025-05-13
Enable backfile cleanup

timmy-wright:
**What type of PR is this?**

/kind chore

enable backfill cleanup - used to cleanup windows test builds

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/e2e-windows.yaml
```diff
@@ -41,6 +41,7 @@ pr:
       - .pipelines/e2e-windows.yaml
       - .pipelines/templates/e2e-template.yaml
       - .pipelines/scripts/e2e_run.sh
+      - .pipelines/scripts/e2e_delete_vmss.sh
```
cameronmeissner, line 44: I don't think we need the VMSS deletion scripts anymore tbh, we have explicit steps in the e2e code now which garbage collects any VMSS instances that were left around every time it's run

cameronmeissner (approved):

.pipelines/e2e-windows.yaml
timmy-wright: That makes sense - might look into this in a separate PR. I just wanted to add the dependency here in case.

smith1511 (approved):

PR #6372, opened 2025-05-13
fix: update SSH service validation to check socket status and capture…

bravebeaver:
perf: add benchmarking for ssh reconfiguration

- fix e2e exit code for checking socket
- add benchmark during provision

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -132,6 +132,7 @@ capture_benchmark "${SCRIPT_NAME}_validate_container_runtime_and_override_ubuntu
 
 # Configure SSH service during VHD build for Ubuntu 22.10+
 configureSSHService "$OS" "$OS_VERSION" || echo "##vso[task.logissue type=warning]SSH Service configuration failed, but continuing VHD build"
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -132,6 +132,7 @@ capture_benchmark "${SCRIPT_NAME}_validate_container_runtime_and_override_ubuntu
 
 # Configure SSH service during VHD build for Ubuntu 22.10+
```
djsly, line 133: should we add the check in this file instead, to call `configureSSHService` only if on the right OS.  I think the majority of the helper functions simply execute the required logic, the Decisions is all done in this file for other logic.

cameronmeissner: I agree with this, I think it's implement the way it is currently since we were originally thinking it would be part of CSE, but now that we're doing it during the build I think adding the OS version check out here makes more sense

Devinwong: Oh right. This makes sense to me too. Didn't realize this in the previous review.

PR #6378, opened 2025-05-14
fix(e2e): fix unbound variables in e2e_run.sh

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/scripts/e2e_run.sh
+++ b/.pipelines/scripts/e2e_run.sh
@@ -32,6 +32,7 @@ echo "##vso[task.setvariable variable=LOGGING_DIR]$LOGGING_DIR"
 # make sure the logging directory exists
 mkdir -p "${DefaultWorkingDirectory}/e2e/${LOGGING_DIR}"
 
+# default any unbound required variables if necessary
 VHD_BUILD_ID="${VHD_BUILD_ID:-}"
 IGNORE_SCENARIOS_WITH_MISSING_VHD="${IGNORE_SCENARIOS_WITH_MISSING_VHD:-}"
 LOGGING_DIR="${LOGGING_DIR:-}"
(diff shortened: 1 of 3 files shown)


djsly (approved):

.pipelines/scripts/e2e_run.sh
```diff
@@ -32,16 +32,25 @@ echo "##vso[task.setvariable variable=LOGGING_DIR]$LOGGING_DIR"
...
-echo "SIG_GALLERY_NAME: ${SIG_GALLERY_NAME:-}"
+# default any unbound required variables if necessary
+VHD_BUILD_ID="${VHD_BUILD_ID:-}"
+IGNORE_SCENARIOS_WITH_MISSING_VHD="${IGNORE_SCENARIOS_WITH_MISSING_VHD:-}"
+LOGGING_DIR="${LOGGING_DIR:-}"
+E2E_SUBSCRIPTION_ID="${E2E_SUBSCRIPTION_ID:-}"
+TAGS_TO_SKIP="${TAGS_TO_SKIP:-}"
+TAGS_TO_RUN="${TAGS_TO_RUN:-}"
+GALLERY_NAME="${GALLERY_NAME:-}"
+SIG_GALLERY_NAME="${SIG_GALLERY_NAME:-}"
```
jepio, line 43: Looks good, a common pattern with the same effect is also:
```
: "${SIG_GALLERY_NAME:=}"
```
(doesn't require spelling things out twice)

jepio:
```
       ${parameter:=word}
              Assign Default Values.  If parameter is unset or null, the expansion of word is assigned  to  parame‐
              ter.   The  value of parameter is then substituted.  Positional parameters and special parameters may
              not be assigned to in this way.
```

cameronmeissner: thanks for pointing that out, will keep that in my back pocket for next time :)

PR #6382, opened 2025-05-15
renovate rules for servercore and nanoserver

timmy-wright:
**What type of PR is this?**

/kind chore

This PR adds renovate rules for servercore and nanoserver.

The tags for these containers are strange. They always start with "10.0" and then have a number indicating the windows container version, and then the patch version. Example:

* 10.0.20348.3328

20348 means "windows 2022 style container". Wheras 26100 means "windows 2025 style container" and "17763" means "windows 2019 style container"

To ensure that the upgrade does not upgrade between windows versions - ie a WS2019 container doesn't get changed to 2025, we use the windows version as the major version, and the 4th part as the patch version. Yeah, it's odd, but it seems to work.

I also made WS2025 cache both WS2022 and WS2025 style containers as it supports both.

Screen snapshot of generated PR on my agentbaker fork (seems to be the only way to test renovate changes):

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -320,7 +320,21 @@
     {
       "matchPackageNames": [
         "windows/nanoserver",
-        "windows/servercore",
+        "windows/servercore"
+      ],
+      "groupName": "windowsbase",
+      "assignees": [
+        "timmy-wright"
+      ],
+      "reviewers": [
+        "timmy-wright"
+      ],
+      "enabled": true,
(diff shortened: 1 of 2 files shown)


smith1511 (approved):

benjamin-brady (approved):

zylxjtu:
if we cache the container base image with @sha tag, but most of the container will have base image @ltscxxx, will AKS have to download the image with the ltscxxx tag and also cache them?

djsly (approved):

PR #6387, opened 2025-05-15
Artifact streaming: use systemd helper functions to enable and start

ganeshkumarashok:
**What type of PR is this?**
/kind bug

- This enables, restarts and logs systemd units with appropriate helper functions that have timeouts. This makes it more resilient.
- Previously acr-nodemon systemd unit was started but not enabled. This meant that it would not run post a reboot. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -377,11 +376,9 @@ ensureArtifactStreaming() {
...
   modprobe target_core_user
   curl -X PUT 'localhost:8578/config?ns=_default&enable_suffix=azurecr.io&stream_format=overlaybd' -O
-  systemctl enable /opt/overlaybd/overlaybd-tcmu.service
-  systemctl enable /opt/overlaybd/snapshotter/overlaybd-snapshotter.service
-  systemctl start overlaybd-tcmu
-  systemctl start overlaybd-snapshotter
-  systemctl start acr-nodemon
+  systemctlEnableAndStart /opt/overlaybd/overlaybd-tcmu.service 30
+  systemctlEnableAndStart /opt/overlaybd/snapshotter/overlaybd-snapshotter.service 30
+  systemctlEnableAndStart acr-nodemon 30
```
ganeshkumarashok, line 381: Previously, I just had acr-nodemon starting (with a systemctl enable). I discovered two issues from the extended artifact streaming e2e I'd added -- acr-nodemon was not resilient to reboots and it occasionally timed out, causing a failure. This will provide it an explicit timeout value.

Devinwong (approved):
lgtm

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -377,11 +376,11 @@ ensureArtifactStreaming() {
...
   modprobe target_core_user
   curl -X PUT 'localhost:8578/config?ns=_default&enable_suffix=azurecr.io&stream_format=overlaybd' -O
-  systemctl enable /opt/overlaybd/overlaybd-tcmu.service
-  systemctl enable /opt/overlaybd/snapshotter/overlaybd-snapshotter.service
-  systemctl start overlaybd-tcmu
-  systemctl start overlaybd-snapshotter
-  systemctl start acr-nodemon
+  systemctl link /opt/overlaybd/overlaybd-tcmu.service
+  systemctl link /opt/overlaybd/snapshotter/overlaybd-snapshotter.service
+  systemctlEnableAndStart /opt/overlaybd/overlaybd-tcmu.service 30
```
cameronmeissner, line 381: are we sure 30 seconds is enough time to start these units? (only asking since this is top-of-mind right now due to the bootstrap token validation stuff)

ganeshkumarashok: It's not certain that 30s will be enough always. These helper functions will retry so it _should_  succeed eventually. 
Or I could make it go back to the blocking calls on systemctl start and systemctl enable. Thoughts on either option?

AB E2E should test the systemd status, so it should catch any regular issues.

Devinwong (approved):

PR #6392, opened 2025-05-16
feat: Add CIS compliance scanning to Ubuntu 22/24-base builds

jepio:
**What type of PR is this?**

/kind feature

This PR introduces scanning for CIS L1 compliance to those Ubuntu 22.04 and Ubuntu 24.04 builds that rely on the `-base` packer template. Scanning is performed during `test-scan-cleanup`. The CIS report is publish to ADO artifacts. This PR will be followed by config changes that increase our CIS compliance.

We can also finally clean up the `asc-baseline` package that is not used any longer.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -30,6 +30,15 @@ steps:
       vstsPackageVersion: '1.1.0-268'
       downloadDirectory: vhdbuilder/packer
 
+  - task: UniversalPackages@0
+    displayName: Download CIS scanner
+    inputs:
+      command: download
+      vstsFeed: AzCoreLinux-Compliance
+      vstsFeedPackage: cisassessor
+      vstsPackageVersion: '0.0.17'
+      downloadDirectory: vhdbuilder/
+
   - task: DownloadPipelineArtifact@2
     condition: or(eq(variables.OS_SKU, 'CBLMariner'), eq(variables.OS_SKU, 'AzureLinux'))
     displayName: 'Download Kata CC UVM artifact'
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -185,6 +185,16 @@ steps:
       Contents: 'trivy-images-table.txt'
       TargetFolder: '$(Build.ArtifactStagingDirectory)'
 
+  - task: CopyFiles@2
+    condition: and(eq(variables.OS_SKU, 'Ubuntu'), in(variables.OS_VERSION, '22.04', '24.04'), eq(variables.FEATURE_FLAGS, 'None'))
+    displayName: Copy CIS Reports
+    inputs:
+      SourceFolder: '$(System.DefaultWorkingDirectory)'
+      Contents: |
+        cis-report.html
+        cis-report.txt
+      TargetFolder: '$(Build.ArtifactStagingDirectory)'
+
   - task: PublishPipelineArtifact@0
     displayName: Publish Release Notes
     inputs:
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -22,15 +22,6 @@ steps:
     displayName: Get Build Mode
 
   - task: UniversalPackages@0
-    displayName: Download Asc Baseline
-    inputs:
-      command: download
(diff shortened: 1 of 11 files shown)


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -176,6 +176,16 @@ steps:
...
 
+  - task: CopyFiles@2
+    condition: and(eq(variables.OS_SKU, 'Ubuntu'), eq(variables.ARCHITECTURE, 'X86_64'), in(variables.OS_VERSION, '22.04', '24.04'), eq(variables.FEATURE_FLAGS, 'None'))
+    displayName: Copy CIS Reports
+    inputs:
+      SourceFolder: '$(System.DefaultWorkingDirectory)'
+      Contents: |
+        cis-report.html
+        cis-report.txt
+      TargetFolder: '$(Build.ArtifactStagingDirectory)'
```
jepio, line 187: Do these artifacts get auto-published to the release notes folder during release? I hope not. I'd rather we keep these reports internal for now - not because there's anything to hide but because I can't guarantee the stability of the format at this time.

cameronmeissner: no, we specifically select the *-release-notes and *image-bom artifacts when building the release notes PRs, these won't be included even if they're now included in build artifacts

pkg/agent/variables.go
```diff
@@ -62,7 +62,6 @@ func getCustomDataVariables(config *datamodel.NodeBootstrappingConfiguration) pa
 	}
 
 	if !cs.Properties.IsVHDDistroForAllNodes() {
```
cameronmeissner, line 64: I don't think we really need this conditional at all anymore...

cameronmeissner: at this point the agent pools included in `Properties` should only include the single one for which agentbaker is serving the bootstrapping request

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -176,6 +176,16 @@ steps:
       Contents: 'trivy-images-table.txt'
       TargetFolder: '$(Build.ArtifactStagingDirectory)'
 
+  - task: CopyFiles@2
+    condition: and(eq(variables.OS_SKU, 'Ubuntu'), eq(variables.ARCHITECTURE, 'X86_64'), in(variables.OS_VERSION, '22.04', '24.04'), eq(variables.FEATURE_FLAGS, 'None'))
```
cameronmeissner, line 180: why the ` eq(variables.FEATURE_FLAGS, 'None')`?

.pipelines/templates/.builder-release-template.yaml
```diff
@@ -22,12 +22,12 @@ steps:
     displayName: Get Build Mode
 
   - task: UniversalPackages@0
-    displayName: Download Asc Baseline
```
cameronmeissner, line 25: this serves as an ASC baseline replacement?

.pipelines/templates/.builder-release-template.yaml
zachary-bailey: Also, why aren't we scanning ubuntu arm64?

.pipelines/templates/.builder-release-template.yaml
jepio: Yes it does, the outdated ASC baseline binaries that we consumed here set settings that overlap and are incompatible with CIS.

pkg/agent/variables.go
jepio: following up offline, this block and all referenced files can potentially be cleaned up, but I'd do that in a separate pr.

.pipelines/templates/.builder-release-template.yaml
jepio: added arm64 scanning, but i'd still like to keep scanning scoped to a small number of images.

jepio:
Currently looking into whether I can move the scanning to the `test-scan-and-cleanup.sh` step.

The scanning has to upload an artifact to the vm and fetch a pipeline artifact, the upload/download was easy to do through packer. In `test-scan-and-cleanup.sh`, i'd need push/pull the files through a storage account.

djsly (approved):

PR #6415, opened 2025-05-22
Add the logic for cotainer base iamge cache validation

zylxjtu:
This is used for windows container base image cache validation through test pipeline running

**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -103,6 +103,17 @@ steps:
       echo "ARCHITECTURE: ${{ parameters.architecture }}"
       echo "##vso[task.setvariable variable=TEST_VM_RESOURCE_GROUP_NAME]$TEST_VM_RESOURCE_GROUP_NAME"
 
+      # For pipelines where the container base images are provided via storage URLs (e.g., the test pipeline),
+      # the base image tag will be 'ltscxxx', which differs from the values specified in parts/common/component.json.
+      # As a result, cache validation behaves differently. To address this, we check if the container base image URL is set,
+      # and use this environment variable to control the cache validation logic in run-test.sh.
+      if [[ -n "${{ parameters.windowsNanoImageUrl }}" || -n "${{ parameters.windowsCoreImageUrl }}" ]]; then
+        export CONTAINTER_BASE_URLS_EXISTING=true
+      else
+        export CONTAINTER_BASE_URLS_EXISTING=false
+      fi
+      echo "CONTAINTER_BASE_URLS_EXISTING is $CONTAINTER_BASE_URLS_EXISTING"
+
       ./vhdbuilder/packer/test/run-test.sh
     displayName: Run VHD cache test
     # I've seen some transient failures which seem timeing related so let's retry for now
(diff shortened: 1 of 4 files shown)


vhdbuilder/packer/test/windows-vhd-content-test.ps1
```diff
@@ -322,6 +323,41 @@ function Test-ImagesPulled
...
+               $targetImagesToPull += "mcr.microsoft.com/windows/servercore:ltsc2022"
+               $targetImagesToPull += "mcr.microsoft.com/windows/nanoserver:ltsc2025"
+               $targetImagesToPull += "mcr.microsoft.com/windows/servercore:ltsc2025"
+            }
+            default {
+               Write-OutputWithTimestamp "No additional images to pull for SKU: $windowsSKU"
+               exit 1
+            }
+        }
+    }
```
msft-linliu, line 359: I thought we decided in the meeting that if the nano server and servercore URLs are not empty, we will skip the cache validation. But I am cool with this too.

vhdbuilder/packer/test/windows-vhd-content-test.ps1
```diff
@@ -322,6 +323,41 @@ function Test-ImagesPulled
     #    only image reference as a workaround
     $pulledImages = (ctr.exe -n k8s.io image ls -q | Select-String -notmatch "sha256:.*" | % { $_.Line })
 
+    # We need to check the image list from URL to make sure the image list is up to date
+    if ($validatecontainerBaseImageFromUrl -eq $true)
```
timmy-wright, line 327: It'd be good to have a comment describing where this variable is set - given the value is passed from the yaml through run-test.sh and then to this script, it'll make the next developer's life easier to have it documented here.

timmy-wright: No, wait. This is the test file. I meant to write that on the prod file. Will write it there too.

vhdbuilder/packer/test/windows-vhd-content-test.ps1
timmy-wright: I guess my concern is that the container image we download during an override (using the urls) might not have the ltscNNNN tag on it, so these tests might fail unnecessilary.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -102,6 +102,12 @@ steps:
       export TEST_VM_RESOURCE_GROUP_NAME="vhd-test-$(date +%s)-$RANDOM"
       echo "ARCHITECTURE: ${{ parameters.architecture }}"
       echo "##vso[task.setvariable variable=TEST_VM_RESOURCE_GROUP_NAME]$TEST_VM_RESOURCE_GROUP_NAME"
+      if [[ -n "${{ parameters.windowsNanoImageUrl }}" || -n "${{ parameters.windowsCoreImageUrl }}" ]]; then
```
timmy-wright, line 105: It'd be good to have a comment here describing what the variable is used for - to help a future developer.

zylxjtu: will add a comment

vhdbuilder/packer/test/windows-vhd-content-test.ps1
zylxjtu: added comments

timmy-wright (approved):

PR #6440, opened 2025-06-02
feat: cache windows CSE packages on VHD

bravebeaver:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/pkg/agent/testdata/AKSWindows2019+CustomCloud+ootcredentialprovider/CustomData
+++ b/pkg/agent/testdata/AKSWindows2019+CustomCloud+ootcredentialprovider/CustomData
@@ -231,7 +231,7 @@ try
  
         if ($search.Count -eq 0) {
             Write-Log "Could not find windows cse package on VHD. Use remote version instead."
-            $WindowsCSEScriptsPackage = "aks-windows-cse-scripts-0.0.52.zip"
+            $WindowsCSEScriptsPackage = "aks-windows-cse-scripts-v0.0.52.zip"
         }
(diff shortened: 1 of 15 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -61,7 +65,29 @@ steps:
     env:
       MODE: "windowsVhdMode"
       SUBSCRIPTION_ID: $(SUBSCRIPTION_ID)
+  - script: |
+      mkdir -p $(windows.cse.working.dir)
```
timmy-wright: We should put this in a seperate .sh file - that way shellcheck will check the scripts for potential problems. Shellcheckignores scripts embedded in the pipeline yaml.

.pipelines/scripts is usually where I put them.

timmy-wright (changes requested):
Given the potential outage we had last week, I'm on "high alert" for any changes that might happen that make our windows-scripts and AgentBaker incompatible. There are two scenarios I think we need to add tests for:

1. Tests that the VHD produced in this run can be bootstrapped by the existing production agent baker.

2. Tests that the CSE produced by AgentBaker in this run can bootstrap the existing production VHD.

My current risk appetite is "we shouldn't merge this PR without those additional tests".

timmy-wright (approved):
LGTU

timmy-wright (approved):
STILL LGTU

image

PR #6447, opened 2025-06-02
For AzureUSGovCloud: use 535 GRID drivers

ganeshkumarashok:
**What type of PR is this?**
/kind bug

We need this since 550 GRID drivers are not compatible with the host drivers in Fairfax cloud.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -814,6 +814,13 @@ configGPUDrivers() {
     if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
         mkdir -p /opt/{actions,gpu}
         if [ "${CONTAINER_RUNTIME}" = "containerd" ]; then
+            # if target cloud is AzureUSGovernmentCloud, and NVIDIA_DRIVER_IMAGE contains GRID then use the specific 535 NVIDIA driver image
```
djsly, line 817: should augment with the why, until when, etc.  it doesnt explain or the next person when we should revert this.

djsly (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
ganeshkumarashok: added additional comments

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -814,6 +814,14 @@ configGPUDrivers() {
...
+            # this is being added because of the host driver in Fairfax is not compatible with the 550 guest drivers now
+            # It can be removed and synced with 550/other clouds, once it is moved to a compatible host driver (check with the HPC team)
+            if [ "$TARGET_CLOUD" = "AzureUSGovernmentCloud" ] && \
```
ganeshkumarashok, line 820: Would this capture just Fairfax appropriately?

ganeshkumarashok: Confirmed offline: yes.

djsly (approved):

ganeshkumarashok:
Lack of testing is a concern for this

parts/linux/cloud-init/artifacts/cse_config.sh
Devinwong, line 820: Not necessarily now if this PR is in a rush. But I would recommend to add a shellspec unit test to cover all the if else conditions. Now it's getting complicated. With the tests, we can ensure (to some extent) that if later something are changed in this function, there will be no regression.

Devinwong (approved):
approved with suggestions

parts/linux/cloud-init/artifacts/cse_config.sh
Devinwong: For 3 layers of if-else, we should create 6-8 unit tests

djsly: right, not sure why I didn't request it ;)  if lack of testing is a concern, it's even more crucial to add the required unit tests no ?

ganeshkumarashok: Good point, will check that out

ganeshkumarashok: Thanks!

PR #6463, opened 2025-06-04
fix: file ownership on aks-secure-tls-bootstrap-client

jepio:
**What type of PR is this?**

/kind bug
/kind test

The aks-secure-tls-bootstrap-client binary is owned by `501:20`. Unpack with `no-same-owner` to correct the ownership to `root:root`, and add a vhd test to catch such issues in the future.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


djsly (approved):

jepio:
> @jepio thanks for the fix, I guess ownership is already correct for other binaries we're extracting with tar? I wonder if it would be a good idea to create a helper to ensure that `--no-same-owner` is used when we need to untar any sort of binary we're caching

Done now.

PR #6468, opened 2025-06-05
feat: Added AzL3.0 kata image.

Bickor:
**What type of PR is this?**
/kind feature

Adds AzureLinux3.0 kata due to deprecation of AzureLinux2.0 kata.

Fixes #

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker_test.go
```diff
@@ -1485,6 +1485,14 @@ testdomain567.com:53 {
 			}
 		}, nil),
 
+		Entry("AzureLinux v3 with kata", "AzureLinuxV3+Kata", "1.28.0", func(config *datamodel.NodeBootstrappingConfiguration) {
```
cameronmeissner, line 1487: you need to also specify a folder name + k8s version here, also make sure that every folder name is unique

cameronmeissner: also make sure you're running `make generate` to ensure testdata is up to date

Bickor: Those things have already been checked. Folder name and k8s version both set and folder name is unique.

`make` was also run. The tests pass locally but just not on github.

Could there be a difference in how the testdata is generated? Or is a standardized image used for this step?

Bickor: Was missing a rebase with testdata changes. Fixed now.

djsly (approved):

PR #6473, opened 2025-06-06
fix: adjust cloud-init check to allow recoverable errors and add logging

UtheMan:
Adds handling for recoverable errors from cloud-init execution and adds logging.
When recoverable errors are encountered, cloud-init status will return code 2, see https://cloudinit.readthedocs.io/en/latest/explanation/return_codes.html


parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -1,9 +1,9 @@
 PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log";
 echo $(date),$(hostname) > ${PROVISION_OUTPUT};
 {{if ShouldEnableCustomData}}
-cloud-init status --wait > /dev/null 2>&1;
-[ "$?" -ne 0 ] && echo 'cloud-init failed' >> ${PROVISION_OUTPUT} && exit 1;
-echo "cloud-init succeeded" >> ${PROVISION_OUTPUT};
+/bin/bash -c "source /opt/azure/containers/provision_source.sh; handleCloudInitStatus \"${PROVISION_OUTPUT}\"; returnStatus=\$?; echo \"Cloud init status check exit code: \$returnStatus\" >> ${PROVISION_OUTPUT}; exit \$returnStatus" >> ${PROVISION_OUTPUT} 2>&1;
```
cameronmeissner: be careful about sourcing entire provisioning scripts like this, anything not defined in a function will be executed via the source cmd

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -1,9 +1,9 @@
 PROVISION_OUTPUT="/var/log/azure/cluster-provision-cse-output.log";
 echo $(date),$(hostname) > ${PROVISION_OUTPUT};
 {{if ShouldEnableCustomData}}
```
UtheMan, line 3: TODO/WIP: add handling for a situation where sourced file/function are missing

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1040,4 +1042,108 @@ extract_tarball() {
...
+    local endTime=$(date +"%F %T.%3N")
+    local task="AKS.CSE.CloudInitStatusCheck"
+    local eventsFileName=$(date +%s%3N)
+    jsonString=$( jq -n \
+        --arg Timestamp   "${startTime}" \
+        --arg OperationId "${endTime}" \
+        --arg Version     "1.23" \
+        --arg TaskName    "${task}" \
+        --arg EventLevel  "Informational" \
+        --arg Message     "${eventMessage}" \
```
UtheMan: TODO/WIP: adjust formatting of this message for easier parsing

parts/linux/cloud-init/artifacts/cse_cmd.sh
UtheMan: This is `cse_helpers.sh` under the hood, so should be safe to source like this, I wanted to avoid having to add an extra file to the VHD, but maybe it's a better/safer way

parts/linux/cloud-init/artifacts/cse_cmd.sh
UtheMan: Changed the approach - no longer sourcing anything in `cse_cmd.sh`

parts/linux/cloud-init/artifacts/cse_cmd.sh
UtheMan: Changed the approach - moved the check to `cse_main.sh`.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -48,6 +48,9 @@ fi
 
 echo $(date),$(hostname), startcustomscript>>/opt/m
 
+# check cloud init status and log details, we already exit in cse_cmd.sh if cloud-init returns exit code 1, here we just log details of errors that cloud-init encountered
+handleCloudInitStatus "/var/log/azure/cluster-provision-cse-output.log";
```
anujmaheshwari1: why do we need this if we're already exiting earlier? I assume these logs will already be present in the cloud-init logs

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1040,4 +1042,60 @@ extract_tarball() {
...
+    # Capture detailed cloud-init status for tracking of different errors
+    cloud-init status --wait > /dev/null 2>&1;
+    local cloudInitExitCode=$?;
+
+    if [ "$cloudInitExitCode" -eq 0 ]; then
+        cloudInitMessage="cloud-init succeeded"
+        echo "${cloudInitMessage}" >> "${provisionOutput}";
+        # we want to avoid logging more information for success case (majority of them)
+        return 0
+    elif [ "$cloudInitExitCode" -eq 1 ]; then
```
anujmaheshwari1: this statement for when the exit code is 1 is actually redundant right? since we would have exited earlier.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1040,4 +1042,60 @@ extract_tarball() {
     esac
 }
 
-#HELPERSEOF
+handleCloudInitStatus() {
```
anujmaheshwari1: this func mostly LGTM except the exit code 1 redundancy I mentioned, but I am not sure I understand its purpose 100% given that everything else is just warning, can you elaborate on what value this is adding?

parts/linux/cloud-init/artifacts/cse_main.sh
UtheMan: `handleCloudInitStatus` logs more details and sends data to kusto

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: Yeah, I left it like this in case something changes in `cse_cmd.sh`, just in case and to make sure `handleCloudInitStatus` covers all the cases. I wasn't exactly sure how to handle this, because adding extra logging to `cse_cmd.sh` is a bit tricky/annoying. And we mostly care about getting details about recoverable errors.

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: replied to the comment added to `cse_main.sh`

parts/linux/cloud-init/artifacts/cse_main.sh
Devinwong: Please pin the kusto query somewhere so that on-call people can refer to it when seeing CloudInit issue. Thanks for working on this PR! :)

UtheMan: Will do, once I write it :')

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: I'm confused, how are we ever calling this function in the case where cloud-init fails? as @anujmaheshwari1 mentioned we would've already exited with code 223 before ever hitting this

cameronmeissner: discussed offline, we'll at least get better logging on recoverable errors

parts/linux/cloud-init/artifacts/cse_main.sh
UtheMan: Forgot to remove this, fix coming soon

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1040,4 +1042,5 @@ extract_tarball() {
     esac
 }
 
-#HELPERSEOF
+
```
anujmaheshwari1: extra line?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -334,6 +334,13 @@ write_files:
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "ensureIMDSRestrictionScript"}}
 
+- path: /opt/azure/containers/cloud-init-status-check.sh
```
anujmaheshwari1, line 337: uhh, lets think about karpenter here.

there is not concept of placing files on the disk through custom data for karpenter. if our bootstrapping relies on this file, we are going to break karpenter with the next release.

are we making sure that this file is also present on disk and the contents are exactly the same? i.e we will never get in a case where the file contents being placed from cloud init will overwrite whats on the disk.

anujmaheshwari1: do we absolutely want to do it this way? i want to make sure we have considered any evident benefits of this way vs old way, since my preference will always be not to add anything in cloud init.

parts/linux/cloud-init/artifacts/cse_helpers.sh
UtheMan: removed

parts/linux/cloud-init/nodecustomdata.yml
UtheMan: My understanding is that since we are adding this file to the VHD as well, karpenter case won't be affected.

-  If they use a new VHD with old custom data -> file will still be there on the VHD. 
- If they get an old VHD but new cse/custom data -> file will be added with cloud-init

Wanted to handle this like that since it seemed like the "cleanest" way to get proper logging in `cse_cmd.sh`

parts/linux/cloud-init/artifacts/cloud-init-status-check.sh
```diff
@@ -0,0 +1,69 @@
...
+        --arg EventTid    "0" \
+        '{Timestamp: $Timestamp, OperationId: $OperationId, Version: $Version, TaskName: $TaskName, EventLevel: $EventLevel, Message: $Message, EventPid: $EventPid, EventTid: $EventTid}'
+    )
+    mkdir -p ${EVENTS_LOGGING_DIR}
+    echo ${jsonString} > ${EVENTS_LOGGING_DIR}${eventsFileName}.json
+
+    if [ "$cloudInitExitCode" -eq 1 ]; then 
+        # if cloud-init exited with code 1, we exit with ERR_CLOUD_INIT_FAILED indicating non-recoverable error in cloud init
+        return $ERR_CLOUD_INIT_FAILED
+    else
```
cameronmeissner: nit: redundant else/return

parts/linux/cloud-init/nodecustomdata.yml
cameronmeissner: yeah we should be good since we're baking on the VHD + utilizing nodecustomdata.yaml, I don't like that we still have to do this but until we're fully on scriptless we'll need to use nodecustomdata.yaml to ensure that running on an older VHD doesn't break us

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cloud-init-status-check.sh
UtheMan: fixed, thx

parts/linux/cloud-init/artifacts/cloud-init-status-check.sh
```diff
@@ -0,0 +1,68 @@
...
+    )
+    mkdir -p ${EVENTS_LOGGING_DIR}
+    echo ${jsonString} > ${EVENTS_LOGGING_DIR}${eventsFileName}.json
+
+    if [ "$cloudInitExitCode" -eq 1 ]; then 
+        # if cloud-init exited with code 1, we exit with ERR_CLOUD_INIT_FAILED indicating non-recoverable error in cloud init
+        return $ERR_CLOUD_INIT_FAILED
+    fi
+    # if cloud-init exited with code 2 (recoverable errors), we return 0 to allow CSE to progress
+    return 0
```
cameronmeissner, line 67: don't even need this either but doesn't hurt

PR #6479, opened 2025-06-09
fix: use current cse pacakge in e2e as defined in componts.json

bravebeaver:
**What type of PR is this?**
/kind test

the current default e2e windows profile is using an outdated version of cse script. use folder path so that the current one is  used in e2e test.

previously this does not matter since e2e always uses the cse on the branch, however this has been fixed by another PR.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -3,10 +3,11 @@ package e2e
 import (
 	"encoding/base64"
 	"fmt"
+	"testing"
+
 	"github.com/Azure/agentbaker/aks-node-controller/helpers"
 	aksnodeconfigv1 "github.com/Azure/agentbaker/aks-node-controller/pkg/gen/aksnodeconfig/v1"
 	"github.com/Masterminds/semver"
-	"testing"
 
 	"github.com/Azure/agentbaker/e2e/config"
 	"github.com/Azure/agentbaker/pkg/agent"
@@ -468,7 +469,7 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 				WindowsProvisioningScriptsPackageURL: "https://packages.aks.azure.com/aks-engine/windows/provisioning/signedscripts-v0.2.2.zip",
 				WindowsPauseImageURL:                 "mcr.microsoft.com/oss/kubernetes/pause:1.4.0",
 				AlwaysPullWindowsPauseImage:          false,
-				CseScriptsPackageURL:                 "https://packages.aks.azure.com/aks/windows/cse/csescripts-v0.0.1.zip",
+				CseScriptsPackageURL:                 "https://packages.aks.azure.com/aks/windows/cse/",
 				CNIARM64PluginsDownloadURL:           "https://acs-mirror.azureedge.net/cni-plugins/v0.8.7/binaries/cni-plugins-linux-arm64-v0.8.7.tgz",
 				VnetCNIARM64LinuxPluginsDownloadURL:  "https://acs-mirror.azureedge.net/azure-cni/v1.4.13/binaries/azure-vnet-cni-linux-arm64-v1.4.14.tgz",
 			},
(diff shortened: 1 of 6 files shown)


e2e/node_config.go
```diff
@@ -468,7 +469,7 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 				WindowsProvisioningScriptsPackageURL: "https://packages.aks.azure.com/aks-engine/windows/provisioning/signedscripts-v0.2.2.zip",
 				WindowsPauseImageURL:                 "mcr.microsoft.com/oss/kubernetes/pause:1.4.0",
 				AlwaysPullWindowsPauseImage:          false,
-				CseScriptsPackageURL:                 "https://packages.aks.azure.com/aks/windows/cse/csescripts-v0.0.1.zip",
+				CseScriptsPackageURL:                 "https://packages.aks.azure.com/aks/windows/cse/",
```
djsly, line 472: the function is called `baseLinuxTemplate` are we controlling this version as well in the component.json ? I'm a bit configused where this is coming from and where it's being created.

bravebeaver: the logic for windows to install cse at provision time is [here](https://github.com/Azure/AgentBaker/blob/c03652307949f9259dfc84aca858afdaf788e692/parts/windows/kuberneteswindowssetup.ps1#L228)

AgentBaker picks cse package in the order of (expected behavior): 
- specific cse package version if RP sets it (as in the current e2e setup)
- [version](https://github.com/Azure/AgentBaker/blob/c03652307949f9259dfc84aca858afdaf788e692/parts/common/components.json#L818) defined in components.json, if RP set folder path only

with the current e2e default setup, all windows e2e are to use the fully specified version of cse, v0.0.1 at all time.

we dont have this problem previously in e2e because we were using cse on the branch (instead of the one used at production time during provision), which was then fixed by this [PR](https://github.com/Azure/AgentBaker/pull/6432). 

we need this PR to e2e to behave in alignment with production. with default setting that RP passes in a folder path, and the provisioning script decides which exact version to use.

djsly: Why are we changing the LinuxBase for a windows version ?

bravebeaver: it is confusing to leave it.. it should not be used at all. i think a cleaner approach is just to remove the line entirely.

e2e/node_config.go
```diff
@@ -762,16 +763,17 @@ DXRqvV7TWO2hndliQq3BW385ZkiephlrmpUVM= r2k1@arturs-mbp.lan`,
...
-				CNIPluginsDownloadURL:                "https://acs-mirror.azureedge.net/cni/cni-plugins-amd64-v0.7.6.tgz",
-				CSIProxyDownloadURL:                  "https://packages.aks.azure.com/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz",
-				CalicoImageBase:                      "calico/",
-				ContainerdDownloadURLBase:            "https://storage.googleapis.com/cri-containerd-release/",
-				CseScriptsPackageURL:                 "https://packages.aks.azure.com/aks/windows/cse/csescripts-v0.0.1.zip",
+				ACIConnectorImageBase:       "microsoft/",
+				AlwaysPullWindowsPauseImage: false,
+				AzureCNIImageBase:           "mcr.microsoft.com/containernetworking/",
+				AzureTelemetryPID:           "",
+				CNIARM64PluginsDownloadURL:  "https://acs-mirror.azureedge.net/cni-plugins/v0.8.7/binaries/cni-plugins-linux-arm64-v0.8.7.tgz",
```
smith1511, line 770: Shoudl we update all refs to acs-mirror?

smith1511 (approved):

e2e/node_config.go
bravebeaver: Will do

e2e/node_config.go
timmy-wright: This is odd - as Sylvain points out, this change is to "baseTemplateLinux" function. It should be in the "baseTemplateWindows" function which is further down the file.

djsly: @bravebeaver  why was this merged ? should we look into this change affecting Linux Base Template first ?

PR #6481, opened 2025-06-09
fix: don't install aznfs on ARM64

cameronmeissner:
**What type of PR is this?**

fixes build failures on 2404/2204 ARM64 caused by this package not seemingly existing on PMC for arm64: 

```
2025-06-08T08:18:31Z: ==> azure-arm: + apt-get install -o Dpkg::Options::=--force-confold --no-install-recommends -y aznfs=0.3.15
2025-06-08T08:18:31Z:     azure-arm: Reading package lists...
2025-06-08T08:18:31Z:     azure-arm: Building dependency tree...
2025-06-08T08:18:31Z:     azure-arm: Reading state information...
2025-06-08T08:18:31Z:     azure-arm: Package aznfs is not available, but is referred to by another package.
2025-06-08T08:18:31Z:     azure-arm: This may mean that the package is missing, has been obsoleted, or
2025-06-08T08:18:31Z:     azure-arm: is only available from another source
2025-06-08T08:18:31Z:     azure-arm:
2025-06-08T08:18:31Z: ==> azure-arm: E: Version '0.3.15' for 'aznfs' was not found
```

also adds 2204ARM64 to the list of PR check-in SKUs so we can catch these sorts of issues in the future before merging

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -51,7 +51,9 @@ installDeps() {
     fi
 
     if [ "${OSVERSION}" = "22.04" ] || [ "${OSVERSION}" = "24.04" ]; then
-        pkg_list+=("aznfs=0.3.15")
+        if [ "$(isARM64)" -eq 0 ]; then
+            pkg_list+=("aznfs=0.3.15")
```
ganeshkumarashok, line 55: Do you know why this package doesn't install for ARM 64?

cameronmeissner: seems like PMC doesn't have an entry for it for arm64?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -51,7 +51,9 @@ installDeps() {
     fi
 
     if [ "${OSVERSION}" = "22.04" ] || [ "${OSVERSION}" = "24.04" ]; then
-        pkg_list+=("aznfs=0.3.15")
```
ganeshkumarashok, line 54: Was there a reason why this was not added to components.json?

djsly: cause os packages are not managed by components.json. This is a bigger tasks if we decide we want to go this route. 

@andyzhangx  normally we don't specify versions for OS packages, is there a reason why we are not using the latest from the Repo ? this will cause us issue with update detection. We should simply use latest avaialble

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
djsly: could it be another version ? we should be ppinning os package versions, we don't do that anywhere, not sure why this is a first.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
andyzhangx: there are two versions for different purposes now, finally the two versions would be combined, let's use the spefic version for now

andyzhangx (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
cameronmeissner: @andyzhangx explained above, it seems this will be a temporary one-off

PR #6494, opened 2025-06-12
Chore: set +x

timmy-wright:
**What type of PR is this?**

/kind chore

Turns off echoing commands in the shell scripts as this was causing setvariable to be called multiple times in the ADO pipeline and the scripts were failing.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/templates/.generate-github-token.yaml
```diff
@@ -1,6 +1,8 @@
 steps:
 - bash: |
     set -euo pipefail
+    # don't echo commands - this will cause the ADO setvariable to fail in strange ways. Also, token security.
+    set +x
```
djsly: isnt this the default ?

e2e/config/azure.go
```diff
@@ -448,7 +448,14 @@ func (a *AzureClient) ensureReplication(ctx context.Context, t *testing.T, image
 		return nil
 	}
 	t.Logf("##vso[task.logissue type=warning;]Replicating to region %s: image version %s", Config.Location, *version.ID)
-	return a.replicateImageVersionToCurrentRegion(ctx, image, version)
+
+	start := time.Now() // Record the start time
+	err := a.replicateImageVersionToCurrentRegion(ctx, image, version)
+	elapsed := time.Since(start) // Calculate the elapsed time
+
+	fmt.Printf("##vso[task.logissue type=warning;]Replication took: %s (%s)\n", elapsed, *version.ID)
```
djsly: why is this a warning ?

e2e/pod.go
```diff
@@ -30,6 +33,10 @@ func ensurePod(ctx context.Context, s *Scenario, pod *corev1.Pod) {
 
 	_, err = kube.WaitUntilPodRunning(ctx, s.T, pod.Namespace, "", "metadata.name="+pod.Name)
 	require.NoErrorf(s.T, err, "failed to wait for pod %q to be in running state", pod.Name)
+
+	timeForReady := time.Since(start)
+	fmt.Printf("##vso[task.logissue type=warning;]Time for pod %q to get ready was %s\n", pod.Name, timeForReady)
```
djsly: same, where is this a warning ?

spec/shellspec.Dockerfile
```diff
@@ -2,5 +2,5 @@
 # Don't forget to log in to the Azure Container Registry before building this image:
 # az acr login --name aksdataplanedev
 FROM aksdataplanedev.azurecr.io/shellspec/shellspec-debian:0.28.1
-RUN apt-get update && apt-get install -y --no-install-recommends gawk jq curl && apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN apt-get -o DPkg::Lock::Timeout=60 update && apt-get install -o DPkg::Lock::Timeout=60 -y --no-install-recommends gawk jq curl && apt-get -o DPkg::Lock::Timeout=60 clean && rm -rf /var/lib/apt/lists/*
```
djsly: please add a config on the client vm instead of updating all single apt-get calls. this makes the code hard to read

https://tecadmin.net/apt-configuration-file-options/
https://manpages.ubuntu.com/manpages/trusty/man5/apt.conf.5.html

djsly (changes requested):

.pipelines/templates/.generate-github-token.yaml
timmy-wright: Yeah, but I wanted to specify it explicitly, so the next developer is forced to read the comment.

e2e/config/azure.go
timmy-wright: I tried to find docs, and the only two options I could see were "warning" and "error". Happy to change if you can find one for info!

spec/shellspec.Dockerfile
timmy-wright: Will do. Good idea!

e2e/config/azure.go
djsly: but what's the intention here ? warning print in orange in the log, error prints in red. 

why do we want this information to pop out ?

.pipelines/templates/.generate-github-token.yaml
timmy-wright: Looks like this file has been removed anyway - I rebased and it's gone!

e2e/config/azure.go
timmy-wright: Intention is to have it extracted from the logs and shown on the pipeline screen - as per the screen snapshot. I'll try logging with an "information" tag and see what that does.

e2e/pod.go
timmy-wright: Have changed to only log as a warning if the duration is over a limit.

e2e/config/azure.go
timmy-wright: Have changed to only log as a warning if the duration is over a limit.

It turns out "information" didn't work - only warning and error are valid types.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -616,9 +616,9 @@ if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
   apt_get_purge 20 30 120 snapd || exit 1
   apt_get_purge 20 30 120 apache2-utils || exit 1
 
-  apt-get -y autoclean || exit 1
-  apt-get -y autoremove --purge || exit 1
-  apt-get -y clean || exit 1
+  apt-get -o DPkg::Lock::Timeout=60 -y autoclean || exit 1
```
djsly: same here can't we drop the apt.conf file on the VM were all those scripts are executed ?

djsly (approved):

vhdbuilder/packer/install-dependencies.sh
timmy-wright: Removed the new apt arguments.

PR #6495, opened 2025-06-12
flake: startup nanoserver pod not iis for windows node testing

timmy-wright:
**What type of PR is this?**

/kind flake

Currently to test that pods can start on windows, we startup an iis container. However, iis is based on an old version of servercore and takes ages to download.

This PR:
* adds validation that there are the right number of servercore and nanoserver for each windows VHD
* checks we can start pods for all those cached versions of servercore and nanoserver

That will enhance our testing of windows containers and reduce test run time as we're no longer starting up an uncached iis container.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/test_helpers_test.go
```diff
@@ -0,0 +1,128 @@
+package e2e
+
+import (
+	"fmt"
+	"github.com/Azure/agentbaker/e2e/config"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestImagesAreFullySpecified(t *testing.T) {
```
r2k1, line 10: Are these tests for tests?

e2e/validation.go
```diff
@@ -12,13 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func ValidatePodRunning(ctx context.Context, s *Scenario) {
-	testPod := func() *corev1.Pod {
-		if s.VHD.OS == config.OSWindows {
-			return podHTTPServerWindows(s)
-		}
-		return podHTTPServerLinux(s)
```
djsly, line 20: I dont see anymore `ValidatePodRunning` getting called with a linux pod.  was this function never used before ?

e2e/scenario_helpers_test.go
```diff
@@ -261,6 +289,62 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
 	return expectedVersions
 }
 
+func Map[T, U any](ts []T, f func(T) U) []U {
```
djsly: should this be in a more generic file ?

e2e/scenario_helpers_test.go
```diff
@@ -261,6 +289,62 @@ func getExpectedPackageVersions(packageName, distro, release string) []string {
...
+	var expectedVersions []string
+	// since we control this json, we assume its going to be properly formatted here
+	jsonBytes, _ := os.ReadFile("../parts/common/components.json")
+
+	containerImages := gjson.GetBytes(jsonBytes, "ContainerImages") //fmt.Sprintf("ContainerImages", containerName))
+
+	for _, containerImage := range containerImages.Array() {
+		imageDownloadUrl := containerImage.Get("downloadURL").String()
+		if strings.EqualFold(imageDownloadUrl, containerName) {
+			packages := containerImage.Get("windowsVersions")
```
djsly: why WindowsVersions ? I don't see any if clauses that checks for Windows here.   is this functuion supposed to be generic and provide support for windows and linux ?

e2e/test_helpers_test.go
djsly, line 10: is this test only going to fun on Windows ?  how is the test seleccted for a specific OS ? should it be named differently ?

e2e/test_helpers_test.go
```diff
@@ -0,0 +1,128 @@
+package e2e
```
djsly, line 1: seems that  this file is windows specific, should have a better name ?

e2e/test_helpers_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestWs2025Gen2ServerCore(t *testing.T) {
+	serverCoreVersions := getContainerImageTags("mcr.microsoft.com/windows/servercore:*", "2025-gen2")
+	t.Logf("found servercore version %v", serverCoreVersions)
+	require.Len(t, serverCoreVersions, 4)
+}
+
+func TestWs2025Gen2Nanoserver(t *testing.T) {
+	serverCoreVersions := getContainerImages("mcr.microsoft.com/windows/nanoserver:*", "2025-gen2")
+	t.Logf("found servercore version %v", serverCoreVersions)
+	require.Len(t, serverCoreVersions, 2)
```
r2k1, line 27: Does it mean every time we add/remove something to components.json this test is going to fail?

Maybe just test it's not empty

r2k1 (commented):
This is quite complex for what it does.

Can be the images be hard-coded inside VHD struct?

e2e/kube.go
```diff
@@ -527,31 +527,17 @@ func podHTTPServerLinux(s *Scenario) *corev1.Pod {
...
-					},
-					ReadinessProbe: &corev1.Probe{
-						PeriodSeconds: 1,
-						ProbeHandler: corev1.ProbeHandler{
-							HTTPGet: &corev1.HTTPGetAction{
-								Path: "/",
-								Port: intstr.FromInt32(80),
-							},
-						},
-					},
```
r2k1, line 556: A minor thing. But I think readiness probe caught an issue before (maybe not).

I prefer to keep it, but I don't mind loosing it

e2e/validation.go
timmy-wright: It's a slightly different code path and is in scenario_helpers_test.go now:

image

e2e/test_helpers_test.go
timmy-wright: Yes - I only needed the code to extract windows containers. Will rename the methods to make that explicit. Someone else can expand the code in future if they need to.

e2e/scenario_helpers_test.go
timmy-wright: Yes - I only needed the code to extract windows containers. Will rename the methods to make that explicit. Someone else can expand the code in future if they need to.

e2e/test_helpers_test.go
timmy-wright: This was a nightmare choosing the name.

We had no tests of functions in a file called "scenario_helpers_test.go", and I really didn't want to create a file called "scenario_helpers_test_test.go" for tests for functions in that file.

What I've done now is:
* extract testable functions from scenario_helpers_test.go into a file "test_helpers.go", and now the file "test_helpers_test.go" is more obviously tests of functions in that file.

e2e/test_helpers_test.go
timmy-wright: They are tests for helper functions that are used in tests.

e2e/test_helpers_test.go
timmy-wright: No, it tests that there are 2 versions of "servercore" container images cached. Not which two. And our requirement is "cache latest and previous latest" so it should not fail very often at all.

e2e/kube.go
timmy-wright: The readiness probe checks the container is responding to HTTP requests. Neither the servercore nor nanoserver containers have an HTTP server, so this readiness probe won't work anymore. 

If you have any ideas for a working readiness probe, that would be great!

e2e/scenario_helpers_test.go
timmy-wright: I'll move it somewhere more generic - there is the toolkit package that seems appropriate.

e2e/test_helpers_test.go
r2k1: Some tests check len is 1, 2 or 4?

e2e/test_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+	})
+}
+
+// TODO: expand this logic to support linux container images as well
+func getWindowsContainerImageTags(containerName string, windowsVersion string) []string {
+	var expectedVersions []string
+	// since we control this json, we assume its going to be properly formatted here
+	jsonBytes, _ := os.ReadFile("../parts/common/components.json")
+
+	containerImages := gjson.GetBytes(jsonBytes, "ContainerImages") //fmt.Sprintf("ContainerImages", containerName))
```
r2k1, line 296: I usually define a test scoped struct in the function for json parsing. Haven't tried the library before, but it doesn't seem like it simplified much.

I wonder if we have helpers for the components.json somewhere? Or should we create them?

r2k1 (approved):
I wish it to be simpler. But I have no suggestions, so LGTM

e2e/toolkit/arrays.go
```diff
@@ -0,0 +1,9 @@
+package toolkit
+
+func Map[T, U any](ts []T, f func(T) U) []U {
+	us := make([]U, len(ts))
+	for i := range ts {
+		us[i] = f(ts[i])
+	}
+	return us
+}
```
r2k1, line 9: You can also try
https://github.com/samber/lo

I usually just write type-specific for loops instead of such constructions. LOC is comparable, but usually easier to understand

e2e/test_helpers.go
timmy-wright: We did have helpers for components.json, but they didn't parse the right part of components.json - so I was extending the helpers.

bravebeaver (approved):
test-helpers sounds more aligned.

LGTM

PR #6496, opened 2025-06-12
fix: prevent PodSandbox image from getting GCed by configuring pinned=true

djsly:
prevent PodSandbox image from getting GCed by configuring pinned=true
```
root@aks-agentpool-11430180-vmss000000:/# crictl image --pinned | grep pause
mcr.microsoft.com/oss/kubernetes/pause                                                3.6                                               7b178dc69474d       301kB               false
```
```
root@aks-agentpool-11430180-vmss000000:/# ctr --namespace k8s.io image label mcr.microsoft.com/oss/kubernetes/pause:3.6 io.cri-containerd.pinned=pinned   
io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned
```
```
root@aks-agentpool-11430180-vmss000000:/# crictl image --pinned | grep pause
mcr.microsoft.com/oss/kubernetes/pause                                                3.6                                               7b178dc69474d       301kB               true
```
**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -65,11 +65,9 @@ ERR_SYSTEMD_DOCKER_STOP_FAIL=116 # Error stopping dockerd
 ERR_CRICTL_DOWNLOAD_TIMEOUT=117 # Timeout waiting for crictl downloads
 ERR_CRICTL_OPERATION_ERROR=118 # Error executing a crictl operation
 ERR_CTR_OPERATION_ERROR=119 # Error executing a ctr containerd cli operation
+ERR_INVALID_CLI_TOOL=120 # Invalid CLI tool specified, should be one of ctr, crictl, docker
 
-# Azure Stack specific errors
-ERR_AZURE_STACK_GET_ARM_TOKEN=120 # Error generating a token to use with Azure Resource Manager
-ERR_AZURE_STACK_GET_NETWORK_CONFIGURATION=121 # Error fetching the network configuration for the node
-ERR_AZURE_STACK_GET_SUBNET_PREFIX=122 # Error fetching the subnet address prefix for a subnet ID
+# 121, 122, 123 are free for use
 
 # Error code 124 is returned when a `timeout` command times out, and --preserve-status is not specified: https://man7.org/linux/man-pages/man1/timeout.1.html
 ERR_VHD_BUILD_ERROR=125 # Reserved for VHD CI exit conditions
--- a/parts/linux/cloud-init/artifacts/cse_install.sh
+++ b/parts/linux/cloud-init/artifacts/cse_install.sh
@@ -789,6 +789,14 @@ retagContainerImage() {
     fi
 }
 
+labelContainerImage() {
+    CONTAINER_IMAGE_URL=$1
+    LABEL_KEY=$2
+    LABEL_VALUE=$3
+    echo "labeling image ${CONTAINER_IMAGE_URL} with ${LABEL_KEY}=${LABEL_VALUE} using ctr"
+    ctr --namespace k8s.io image label $CONTAINER_IMAGE_URL $LABEL_KEY=$LABEL_VALUE
+}
+
 retagMCRImagesForChina() {
     if [ "${CONTAINER_RUNTIME}" = "containerd" ]; then
         # shellcheck disable=SC2016
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -543,26 +543,48 @@ while IFS= read -r imageToBePulled; do
 done <<< "$ContainerImages"
 echo "Waiting for container image pulls to finish. PID: ${image_pids[@]}"
 wait ${image_pids[@]}
+capture_benchmark "${SCRIPT_NAME}_caching_container_images"
 
-watcher=$(jq '.ContainerImages[] | select(.downloadURL | contains("aks-node-ca-watcher"))' $COMPONENTS_FILEPATH)
-watcherBaseImg=$(echo $watcher | jq -r .downloadURL)
-watcherVersion=$(echo $watcher | jq -r .multiArchVersionsV2[0].latestVersion)
-watcherFullImg=${watcherBaseImg//\*/$watcherVersion}
+retagAKSNodeCAWatcher() {
+  # This function retags the aks-node-ca-watcher image to a static tag
+  # The static tag is used to bootstrap custom CA trust when MCR egress may be intercepted by an untrusted TLS MITM firewall.
+  # The image is never pulled, it is only retagged.
 
-# this image will never get pulled, the tag must be the same across different SHAs.
-# it will only ever be upgraded via node image changes.
-# we do this because the image is used to bootstrap custom CA trust when MCR egress
-# may be intercepted by an untrusted TLS MITM firewall.
-watcherStaticImg=${watcherBaseImg//\*/static}
+  watcher=$(jq '.ContainerImages[] | select(.downloadURL | contains("aks-node-ca-watcher"))' $COMPONENTS_FILEPATH)
+  watcherBaseImg=$(echo $watcher | jq -r .downloadURL)
+  watcherVersion=$(echo $watcher | jq -r .multiArchVersionsV2[0].latestVersion)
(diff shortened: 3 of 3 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -65,11 +65,9 @@ ERR_SYSTEMD_DOCKER_STOP_FAIL=116 # Error stopping dockerd
 ERR_CRICTL_DOWNLOAD_TIMEOUT=117 # Timeout waiting for crictl downloads
 ERR_CRICTL_OPERATION_ERROR=118 # Error executing a crictl operation
 ERR_CTR_OPERATION_ERROR=119 # Error executing a ctr containerd cli operation
+ERR_INVALID_CLI_TOOL=120 # Invalid CLI tool specified, should be one of ctr, crictl, docker
 
-# Azure Stack specific errors
-ERR_AZURE_STACK_GET_ARM_TOKEN=120 # Error generating a token to use with Azure Resource Manager
```
cameronmeissner, line 70: these weren't being used anymore?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -789,6 +789,14 @@ retagContainerImage() {
     fi
 }
 
+labelContainerImage() {
+    CONTAINER_IMAGE_URL=$1
+    LABEL_KEY=$2
+    LABEL_VALUE=$3
+    echo "labeling image ${CONTAINER_IMAGE_URL} with ${LABEL_KEY}=${LABEL_VALUE} using ctr"
+    ctr --namespace k8s.io image label $CONTAINER_IMAGE_URL $LABEL_KEY=$LABEL_VALUE
```
cameronmeissner, line 798: I know at this point we pretty much only use `ctr`, though other functions (like pullContainerImage) support multiple CLI's, should we consider doing that here as well?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -543,26 +543,48 @@ while IFS= read -r imageToBePulled; do
 done <<< "$ContainerImages"
 echo "Waiting for container image pulls to finish. PID: ${image_pids[@]}"
 wait ${image_pids[@]}
+capture_benchmark "${SCRIPT_NAME}_caching_container_images"
+
+retagAKSNodeCAWatcher() {
```
cameronmeissner, line 548: I feel we should define these either in a separate file or either at the top/bottom of this file

parts/linux/cloud-init/artifacts/cse_install.sh
cameronmeissner: can we also add a set of shellspec tests?

parts/linux/cloud-init/artifacts/cse_helpers.sh
djsly: my Grep-fu couldnt find them

parts/linux/cloud-init/artifacts/cse_install.sh
djsly: adding label is only supported by `CTR` I tried to add CLITool Argument at first but there is no point since crictl cannot change the label.

djsly: added comment to explain the limitation

vhdbuilder/packer/install-dependencies.sh
djsly: I Agree - and I was thinking the same, but there are MANY throughout the file, I would prefer to have a cleanup as a separate PR

cameronmeissner: fair point!

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -543,26 +543,48 @@ while IFS= read -r imageToBePulled; do
...
+  retagContainerImage "ctr" ${watcherFullImg} ${watcherStaticImg}
+}
+retagAKSNodeCAWatcher
+capture_benchmark "${SCRIPT_NAME}_retag_aks_node_ca_watcher"
 
-watcher=$(jq '.ContainerImages[] | select(.downloadURL | contains("aks-node-ca-watcher"))' $COMPONENTS_FILEPATH)
-watcherBaseImg=$(echo $watcher | jq -r .downloadURL)
-watcherVersion=$(echo $watcher | jq -r .multiArchVersionsV2[0].latestVersion)
-watcherFullImg=${watcherBaseImg//\*/$watcherVersion}
+pinPodSandboxImage() {
```
cameronmeissner, line 570: should we add something to content tests to ensure this is actually happening? or an agentbaker e2e validator?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -543,26 +543,48 @@ while IFS= read -r imageToBePulled; do
...
+  watcherVersion=$(echo $watcher | jq -r .multiArchVersionsV2[0].latestVersion)
+  watcherFullImg=${watcherBaseImg//\*/$watcherVersion}
+
+  # this image will never get pulled, the tag must be the same across different SHAs.
+  # it will only ever be upgraded via node image changes.
+  # we do this because the image is used to bootstrap custom CA trust when MCR egress
+  # may be intercepted by an untrusted TLS MITM firewall.
+  watcherStaticImg=${watcherBaseImg//\*/static}
+
+  # can't use cliTool because crictl doesn't support retagging.
```
Devinwong: typo: criTool?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -543,26 +543,48 @@ while IFS= read -r imageToBePulled; do
...
+}
+retagAKSNodeCAWatcher
+capture_benchmark "${SCRIPT_NAME}_retag_aks_node_ca_watcher"
 
-watcher=$(jq '.ContainerImages[] | select(.downloadURL | contains("aks-node-ca-watcher"))' $COMPONENTS_FILEPATH)
-watcherBaseImg=$(echo $watcher | jq -r .downloadURL)
-watcherVersion=$(echo $watcher | jq -r .multiArchVersionsV2[0].latestVersion)
-watcherFullImg=${watcherBaseImg//\*/$watcherVersion}
+pinPodSandboxImage() {
+  # This function pins the pod sandbox image to avoid he GC from removing it.
```
Devinwong: typo: the?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -543,26 +543,48 @@ while IFS= read -r imageToBePulled; do
...
+  # This is achieved by setting the "io.cri-containerd.pinned" label on the image with a value of "pinned".
 
-# this image will never get pulled, the tag must be the same across different SHAs.
-# it will only ever be upgraded via node image changes.
-# we do this because the image is used to bootstrap custom CA trust when MCR egress
-# may be intercepted by an untrusted TLS MITM firewall.
-watcherStaticImg=${watcherBaseImg//\*/static}
+  podSandbox=$(jq '.ContainerImages[] | select(.downloadURL | contains("pause"))' $COMPONENTS_FILEPATH)
```
Devinwong, line 575: Am I understanding correctly that it only pins the container image name with "pause". Why do we only pin this image?

vhdbuilder/packer/install-dependencies.sh
djsly: done

vhdbuilder/packer/install-dependencies.sh
djsly: its the $cliTool

vhdbuilder/packer/install-dependencies.sh
djsly: done

vhdbuilder/packer/install-dependencies.sh
djsly: fixed

vhdbuilder/packer/install-dependencies.sh
djsly: this image is critical for pod startup and it isn't supported with private ACR since containerd won't be using azure-acr-credential to fetch it.

djsly: added more explaination

PR #6513, opened 2025-06-16
chore: disabling 1804 builds and initial code base clean up with 1804 logic

djsly:
disabling 1804 builds and initial code base clean up with 1804 logic

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:


.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -13,14 +13,6 @@ parameters:
     displayName: Branch in aks-rp to use for overrides
     type: string
     default: master
-  - name: build1804containerd
-    displayName: Build 1804 containerd
-    type: boolean
-    default: true
-  - name: build1804gen2containerd
```
cameronmeissner, line 20: you also need to freeze the SIG image config templates in sig_config.go to the currently-latest version: https://github.com/Azure/AgentBaker/blob/master/pkg/agent/datamodel/sig_config.go#L479

Devinwong (approved):

.pipelines/.vsts-vhd-builder-release.yaml
djsly: done, please review and resolve if good

PR #6514, opened 2025-06-16
fix: refactor azure.json creation to configureAzureJson and add test coverage

cameronmeissner:
**What type of PR is this?**

fix: refactor azure.json creation to configureAzureJson and add unit test coverage

we recently caused an outage for SP users due to accidentally removing code which writes the SP secret value to the azure config file at /etc/kubernetes/azure.json - this PR refactors this logic such that we can easily have unit + e2e test coverage over this code path to prevent similar regressions in the future

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


e2e/kube.go
```diff
@@ -123,7 +123,9 @@ func (k *Kubeclient) WaitUntilPodRunning(ctx context.Context, t *testing.T, name
 			}
 		case event := <-watcher.ResultChan():
 			if event.Type != "ADDED" && event.Type != "MODIFIED" {
-				t.Logf("skipping event %s", event.Type)
+				if event.Type != "" {
```
Devinwong, line 126: nit: should we have the nil check go first? (put it before line 125?)

cameronmeissner: `event` is of type `watch.Watch` which is a struct, so there isn't any risk of a nil ptr here

Devinwong: I didn't make it clear. I mean should we have `event.Type != ""` check go first? before checking "ADDED" and "MODIFIED", though it doesn't change anything in this case.

cameronmeissner: will keep as-is for now, this was mainly to cleanup logs

PR #6524, opened 2025-06-17
feat: limit repos for mariner and use snapshot-based url for ubuntu arm during security patch

YaoC:
**What type of PR is this?**
/kind feature

These scripts are for live patching (node OS upgrade). Some repos are not related to node OS, for example: [cloud-native](https://packages.microsoft.com/cbl-mariner/2.0/prod/cloud-native/). For ubuntu, the [unattended-upgrades](https://github.com/mvo5/unattended-upgrades) has the whitelist `Unattended-Upgrade::Allowed-Origins` that only allows specific origins. So for mariner/azurelinux, let's limit the allowed repo during security patch and we will have another way to patch other components.

And for ubuntu, we also replace the arm64 URL to support snapshot-based patch.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
```diff
@@ -3,8 +3,42 @@
 set -o nounset
 set -e
 
-# source dnf_update
-source /opt/azure/containers/provision_source_distro.sh
+dnf_update() {
```
yewmsft, line 6: How to we test this?

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -94,8 +94,10 @@ fi
 if [ -z "${live_patching_repo_service}" ]; then
     echo "live patching repo service is not set, use ubuntu snapshot repo"
     # upgrade from base image to a timestamp
-    # e.g. replace https://azure.archive.ubuntu.com/ubuntu/ with https://snapshot.ubuntu.com/ubuntu/20230727T000000Z
+    # e.g. replace http://azure.archive.ubuntu.com/ubuntu/ with https://snapshot.ubuntu.com/ubuntu/20230727T000000Z
     sed -i 's/http:\/\/azure.archive.ubuntu.com\/ubuntu\//https:\/\/snapshot.ubuntu.com\/ubuntu\/'"${golden_timestamp}"'/g' ${source_list_path}
+    # e.g. replace http://ports.ubuntu.com/ubuntu-ports with https://snapshot.ubuntu.com/ubuntu/20230727T000000Z
+    sed -i 's/http:\/\/ports.ubuntu.com\/ubuntu-ports/https:\/\/snapshot.ubuntu.com\/ubuntu\/'"${golden_timestamp}"'/g' ${source_list_path}
```
yewmsft, line 100: is there "/" after "ubuntu-ports/"?

YaoC: No, it's something like this:

```
deb http://ports.ubuntu.com/ubuntu-ports jammy-security main restricted
```

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
YaoC: In our e2e we have the live-patching scenario, the node will execute this script and will add an annotation on the node if it succeeded. We checked both the annotation and logs by running `journalctl -u snapshot-update.service --output cat`.

yewmsft (approved):
LGTM

parts/linux/cloud-init/artifacts/mariner/mariner-package-update.sh
```diff
@@ -3,8 +3,42 @@
...
+    retries=10
+    dnf_update_output=/tmp/dnf-update.out
+    versionID=$(grep '^VERSION_ID=' /etc/os-release  | cut -d'=' -f2 | tr -d '"')
+    for i in $(seq 1 $retries); do
+        if [ "${versionID}" = "3.0" ]; then
```
djsly: since the only diff between azlinux3 and 2 is the repo list, could you create a variables for the repo list and avoid all the code duplication ?

```
if [ "${versionID}" = "3.0" ]; then
repo_list = --repo azurelinux-official-base --repo azurelinux-official-ms-non-oss ...
else
repo_list = --repo mariner-official-microsoft  ...
fi

! (dnf update \
                --exclude mshv-linuxloader \
                --exclude kernel-mshv \
                ${repo_list} \
                -y --refresh 2>&1 | tee $dnf_update_output | grep -E "^([WE]:.*)|([eE]rr.*)$") && \
            cat $dnf_update_output && break || \
            cat $dnf_update_output
```

YaoC: Sounds good, I'll do that.

YaoC: Updated!

djsly (approved):

PR #6525, opened 2025-06-17
Auto-fill Windows image URLs using WCCT JSON

sabinamiezah-msft:
**What type of PR is this?**

/kind feature
/kind cleanup

**What this PR does / why we need it:**

This PR introduces an automation feature that streamlines the process of populating Windows image URLs (base, nano, and servercore) in AgentBaker. Instead of manually updating each image URL during every Patch Tuesday, this feature enables automatic population using a JSON payload file. The payload is downloaded from the WCCT Agentbaker Storage account and parsed to extract the latest image URLs.

**Which issue(s) this PR fixes:**
This reduces manual intervention, minimizes the risk of human error, and ensures consistency across image updates. It also includes comments and conditions in the case where a valid URL is not available for use, falling back to the pipeline's initial [manually filled] URLs.

Requirements:

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -59,7 +59,7 @@ parameters:
   - name: windowsContainerImageJsonUrl
     displayName: Windows Container Image Json URL
     type: string
-    default: https://wcctagentbakerstorage.blob.core.windows.net/simship/2025/06B/payload.json
+    default: ""
   - name: dryrun
     displayName: Dry run
     type: boolean
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -59,7 +59,7 @@ parameters:
   - name: windowsContainerImageJsonUrl
     displayName: Windows Container Image Json URL
     type: string
-    default: ""
+    default: https://wcctagentbakerstorage.blob.core.windows.net/simship/2025/06B/payload.json
   - name: dryrun
     displayName: Dry run
     type: boolean
--- a/.pipelines/.vsts-vhd-builder-release-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-release-windows.yaml
@@ -56,6 +56,10 @@ parameters:
     displayName: Build 2025 Gen 2
     type: boolean
     default: True
+  - name: windowsContainerImageJsonUrl
+    displayName: Windows Container Image Json URL
+    type: string
+    default: ""
   - name: dryrun
     displayName: Dry run
     type: boolean
@@ -102,6 +106,7 @@ stages:
       build23H2gen2: ${{ parameters.build23H2gen2 }}
       build2025:  ${{ parameters.build2025 }}
       build2025gen2:  ${{ parameters.build2025gen2 }}
+      windowsContainerImageJsonUrl: ${{ parameters.windowsContainerImageJsonUrl }}
       skipExtensionCheck: ${{ parameters.skipExtensionCheck }}
       installOpenSshServer: ${{ parameters.installOpenSshServer }}
       overrideBranch: ${{ parameters.overrideBranch }}
--- a/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
@@ -52,6 +52,7 @@ parameters:
   - name: windowsContainerImageJsonUrl
     displayName: Windows container image JSON URL Override
     type: string
+    default: ""
   - name: overrideBranch
     type: string
     default: master
--- a/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
@@ -49,6 +49,10 @@ parameters:
   - name: windowsCoreImageUrl
     displayName: Windows core base container image URL Override
     type: string
+  - name: windowsContainerImageJsonUrl
+    displayName: Windows container image JSON URL Override
+    type: string
+    default: ""
   - name: overrideBranch
     type: string
     default: master
@@ -88,6 +92,7 @@ stages:
               windowsBaseImageUrl: ${{ parameters.windowsBaseImageUrl }}
               windowsNanoImageUrl: ${{ parameters.windowsNanoImageUrl }}
               windowsCoreImageUrl: ${{ parameters.windowsCoreImageUrl }}
+              windowsContainerImageJsonUrl: ${{ parameters.windowsContainerImageJsonUrl }}
               overrideBranch: ${{ parameters.overrideBranch }}
               useOverrides: ${{ parameters.useOverrides }}
               csePackageDir: ${{ parameters.csePackageDir }}
--- a/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
@@ -52,7 +52,6 @@ parameters:
   - name: windowsContainerImageJsonUrl
     displayName: Windows container image JSON URL Override
     type: string
-    default: ""
   - name: overrideBranch
     type: string
     default: master
--- a/.pipelines/templates/.build-and-test-windows-vhds-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhds-template.yaml
@@ -23,7 +23,6 @@ parameters:
   - name: windowsContainerImageJsonUrl
     displayName: Windows container image JSON URL Override
     type: string
-    default: ""
   - name: dryrun
     displayName: Dry run
     type: boolean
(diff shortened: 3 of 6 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -104,6 +104,7 @@ steps:
       WINDOWS_BASE_IMAGE_URL: ${{ parameters.windowsBaseImageUrl }}
       WINDOWS_NANO_IMAGE_URL: ${{ parameters.windowsNanoImageUrl }}
       WINDOWS_CORE_IMAGE_URL: ${{ parameters.windowsCoreImageUrl }}
+      WINDOWS_CONTAINERIMAGE_JSON_URL: $(WINDOWS_CONTAINERIMAGE_JSON_URL)
```
zylxjtu: Maybe make it as a parameter as well? same as WINDOWS_CORE_IMAGE_URL etc?

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -464,14 +537,14 @@ if [ "$OS_TYPE" = "Windows" ]; then
 		windows_sigmode_source_image_version="1.0.0"
 	fi
 
-	# Set nanoserver image url if the pipeline variable is set
-	if [ -n "${WINDOWS_NANO_IMAGE_URL}" ]; then
+	# Set nanoserver image url if the pipeline variable is set and the parameter is not already set
+	if [ -n "${WINDOWS_NANO_IMAGE_URL}" ] && [ -z "${windows_nanoserver_image_url}" ]; then
```
zylxjtu, line 549: I personally prefer if specific url is set, it will take precedence that is remove the [-z xxxx] part check

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -400,6 +400,79 @@ if [ "$OS_TYPE" = "Windows" ]; then
 	# default: build VHD images from a marketplace base image
 	IMPORTED_IMAGE_NAME=$imported_windows_image_name
 	IMPORTED_IMAGE_URL="https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/system/$IMPORTED_IMAGE_NAME.vhd"
+ 
+	if [ -n "${WINDOWS_CONTAINERIMAGE_JSON_URL}" ]; then
+		# Download the json artifact from the url
+		filename=$(basename "$WINDOWS_CONTAINERIMAGE_JSON_URL")
+		echo "Downloading $filename from wcct storage account using AzCopy with Managed Identity Auth"
+		export AZCOPY_AUTO_LOGIN_TYPE="MSI"
```
zylxjtu: I'm wondering if we can merge this azcopy env settings in this file into one single place?

vhdbuilder/packer/produce-packer-settings.sh
sabinamiezah-msft: My thought process was that the [-z ] part will check if the nano/core images are set so that we have a _fallback in the case where a valid URL is not set_. Also, I worry that _removing this part would end with the value just being overriden_ after all of the parsing is performed

zylxjtu: in the case of both json url and base image url are provided, I personally prefer the base image url will take precedence as it is more specific for us to specify. But I'm open to all options and we can check with other's opinion on this.

vhdbuilder/packer/produce-packer-settings.sh
sabinamiezah-msft: I hope I'm understanding you right, what do you think about moving the AZCOPY_AUTO_LOGIN_TYPE export to the top of our file (outside of any conditions). I'm noticing that it is used both in the parsing code block and also in the build vhd block, so centralizing it might help reduce redundancy and improve clarity as you're suggesting.

.pipelines/templates/.builder-release-template-windows.yaml
timmy-wright: I'd prefer using a parameter as well it's much cleaner to have these passed around using parameters rather than global variables - global variables are much harder to debug.

I'm wondering if we can decompose the JSON into the other variables in a step in the pipeline that uses this template - then we won't have variables with overlapping semantics in this template. Not sure if that is possible, but is worth investigating.

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -400,6 +400,81 @@ if [ "$OS_TYPE" = "Windows" ]; then
 	# default: build VHD images from a marketplace base image
 	IMPORTED_IMAGE_NAME=$imported_windows_image_name
 	IMPORTED_IMAGE_URL="https://${STORAGE_ACCOUNT_NAME}.blob.core.windows.net/system/$IMPORTED_IMAGE_NAME.vhd"
+ 	export AZCOPY_AUTO_LOGIN_TYPE="MSI" # use Managed Identity for AzCopy authentication
```
zylxjtu, line 403: export AZCOPY_AUTO_LOGIN_TYPE="MSI"
		export AZCOPY_MSI_RESOURCE_STRING="${AZURE_MSI_RESOURCE_STRING}"

and remove the other setting in the same file?

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -400,6 +400,81 @@ if [ "$OS_TYPE" = "Windows" ]; then
...
+			WINDOWS_BASE_IMAGE_URL="${W23H2_GEN2_BASE_IMAGE_URL}"
+			windows_nanoserver_image_url="${W2022_NANO_IMAGE_URL}"
+			windows_servercore_image_url="${W2022_CORE_IMAGE_URL}"
+		else
+			echo "Unsupported WINDOWS_SKU: ${WINDOWS_SKU}"
+		fi	
+	fi
+		
+	# Check if nano and core urls are set
+	if [ -z "${windows_nanoserver_image_url}" ] || [ -z "${windows_servercore_image_url}" ]; then
```
zylxjtu: why it is different for the WINDOWS_BASE_IMG_URL in here? should all there parameters be logic similar?

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -464,14 +538,14 @@ if [ "$OS_TYPE" = "Windows" ]; then
 		windows_sigmode_source_image_version="1.0.0"
 	fi
 
-	# Set nanoserver image url if the pipeline variable is set
-	if [ -n "${WINDOWS_NANO_IMAGE_URL}" ]; then
+	# Set nanoserver image url if the pipeline variable is set and the parameter is not already set
+	if [ -n "${WINDOWS_NANO_IMAGE_URL}" ] && [ -z "${windows_nanoserver_image_url}" ]; then
```
zylxjtu, line 549: how about the process of the WINDOWS_BASE_IMAGE_URL?

vhdbuilder/packer/produce-packer-settings.sh
sabinamiezah-msft: Good point, I'll add that to the check

vhdbuilder/packer/produce-packer-settings.sh
sabinamiezah-msft: Adding logging for this in the initial condition

timmy-wright (approved):
LGTM. Can you run the test VHD pipeline with and without the variable set to check they still run (I'm paranoid about regressions in our pipelines!)

.pipelines/templates/.build-and-test-windows-vhds-template.yaml
```diff
@@ -20,6 +20,9 @@ parameters:
   - name: build2025gen2
     displayName: Build 2025 Gen 2
     type: boolean
+  - name: windowsContainerImageJsonUrl
+    displayName: Windows container image JSON URL Override
+    type: string
```
timmy-wright, line 25: We probably need a default here as this template is used by other pipelines.

.pipelines/templates/.build-and-test-windows-vhd-template.yaml
```diff
@@ -49,6 +49,9 @@ parameters:
   - name: windowsCoreImageUrl
     displayName: Windows core base container image URL Override
     type: string
+  - name: windowsContainerImageJsonUrl
+    displayName: Windows container image JSON URL Override
+    type: string
```
timmy-wright, line 54: We probably need a default here as this template is used by other pipelines.

.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -37,6 +37,10 @@ parameters:
   - name: windowsCoreImageUrl
     displayName: Windows core base container image URL Override
     type: string
+  - name: windowsContainerImageJsonUrl
+    displayName: Windows container image JSON URL Override
+    type: string
+    default: https://wcctagentbakerstorage.blob.core.windows.net/simship/2025/06B/payload.json
```
timmy-wright: The default should be an empty string or else it's possible that all pipelines will use this json.

.pipelines/.vsts-vhd-builder-release-windows.yaml
```diff
@@ -56,6 +56,10 @@ parameters:
     displayName: Build 2025 Gen 2
     type: boolean
     default: True
+  - name: windowsContainerImageJsonUrl
+    displayName: Windows Container Image Json URL
+    type: string
+    default: https://wcctagentbakerstorage.blob.core.windows.net/simship/2025/06B/payload.json
```
timmy-wright: This default should be an empty string or else all runs of this pipeline will use the hard coded value.

timmy-wright (approved):
LGTM

PR #6535, opened 2025-06-19
feat: onboard RPM packages to Renovate

Devinwong:
**What type of PR is this?**
/kind feature

Onboard RPM packages moby-containerd, containerd2 (for Ubuntu 24.04) and Kubernetes-cri-tools to Renovate for auto-upgrade.

Tested in a fork repo. Need to merge to master and double check.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -55,7 +55,8 @@
         "oss/kubernetes/kube-proxy",
         "oss/binaries/kubernetes/kubernetes-node",
         "oss/binaries/kubernetes/azure-acr-credential-provider",
-        "containernetworking/azure-npm"
+        "containernetworking/azure-npm",
+        "rpm"
       ],
       "matchUpdateTypes": [
         "minor"
@@ -85,9 +86,18 @@
       "enabled": true
     },
     {
+      "matchDatasources": [
+        "rpm"
+      ],
+      "versioning": "regex:(?<major>\\d+)(\\.(?<minor>\\d+))?(\\.(?<patch>\\d+))?(-(?<prerelease>.*))?$",
+      "automerge": false,
+      "enabled": true
+    },
+    {
       "matchPackageNames": [
         "moby-runc",
         "moby-containerd",
+        "containerd2",
         "aks/aks-node-ca-watcher"
       ],
       "matchUpdateTypes": [
@@ -100,10 +110,8 @@
         "devinwong",
         "anujmaheshwari1",
         "cameronmeissner",
-        "AlisonB319",
         "lilypan26",
         "djsly",
-        "jason1028kr",
         "UtheMan",
         "zachary-bailey",
         "ganeshkumarashok"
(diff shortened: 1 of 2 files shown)


.github/renovate.json
```diff
@@ -55,7 +55,8 @@
         "oss/kubernetes/kube-proxy",
         "oss/binaries/kubernetes/kubernetes-node",
         "oss/binaries/kubernetes/azure-acr-credential-provider",
-        "containernetworking/azure-npm"
+        "containernetworking/azure-npm",
+        "rpm"
```
djsly: Why don't we have that for Deb?  I don't get why we have this called "RPM" only

.github/renovate.json
```diff
@@ -84,10 +85,19 @@
       "automerge": false,
       "enabled": true
     },
+    {
+      "matchDatasources": [
+        "rpm"
+      ],
+      "versioning": "regex:(?<major>\\d+)(\\.(?<minor>\\d+))?(\\.(?<patch>\\d+))?(-(?<prerelease>.*))?$",
+      "automerge": false,
```
djsly, line 92: Maybe we should auto merge no ? We have it like that for iubuntu

.github/renovate.json
djsly: Should this be Moby-containerD. Etc.

parts/common/components.json
```diff
@@ -1263,8 +1263,8 @@
           "current": {
             "versionsV2": [
               {
-                "renovateTag": "<DO_NOT_UPDATE>",
-                "latestVersion": "1.32.0"
+                "renovateTag": "registry=https://packages.microsoft.com/azurelinux/3.0/prod/cloud-native/x86_64/repodata/, name=kubernetes-cri-tools, os=azurelinux, release=3.0",
```
djsly: Shouldn't this need to be under a azurelinux field ?

Why don't we have a mariner field ?

Devinwong: Here I am assuming that if there is a new version in azLinux3 repo, there is also a corresponding version in Mariner 2.0 although it might not be always true. Usually it's true.
The precedent "default", "current" means it will apply to other cases if not explicitly defined above. If we want to make it accurate, we will need to split it into Mariner.2, AzureLinux.3, and probably still leaving the default.current case here as a default.

.github/renovate.json
Devinwong: Copied something wrong my fork. Thanks for catching this. Removed "rpm".

I have 1 rule to create PR for minor update for "moby-runc", "moby-containerd", "containerd2" without auto-merge. And another rule with auto-merge for patch update.

.github/renovate.json
Devinwong: answered in the previous comment

parts/common/components.json
djsly: I was simply looking at how we do it for containerD and saw it was seperated.  Also know mariner2 is EOL. They will probably stop updating new version in v2 and add more in v3

.github/renovate.json
ganeshkumarashok: Is auto-merge working properly though? In the aks-gpu repo it still asks for PR approvals when auto-merge is set to true?

PR #6540, opened 2025-06-20
deprecate WASI from scriptless

Devinwong:
**What type of PR is this?**
/kind cleanup
/kind deprecation

Deprecate WASI from scriptless.
The check-in test `[Buf CI]` failure is expected because it tries to prevent us from making breaking changes. But in this case we don't have any client in production yet, it's fine to remove the field.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


aks-node-controller/proto/aksnodeconfig/v1/config.proto
```diff
@@ -24,7 +24,6 @@ option go_package = "github.com/Azure/agentbaker/aks-node-controller/pkg/gen/aks
 enum WorkloadRuntime {
   WORKLOAD_RUNTIME_UNSPECIFIED = 0;
   WORKLOAD_RUNTIME_OCI_CONTAINER = 1;
-  WORKLOAD_RUNTIME_WASM_WASI = 2;
```
cameronmeissner, line 27: maybe still worth it to reserve just in case?

cameronmeissner (approved):

aks-node-controller/proto/aksnodeconfig/v1/config.proto
Devinwong: A replacement of this is using spinkube. So it doesn't look like we will need it back in the future. Removing it will only break the Node Provisioner test build, which is only used by Yihao and me now. We can always add it back when needed. But it seems unlikely(to me). Let me know if you think otherwise. :)

PR #6544, opened 2025-06-21
add crictl for marinerkata

Devinwong:
**What type of PR is this?**

/kind bug

Need to add this back because we are breaking down the default.current into all SKUs as they are using different versions of cri-tools.
AB e2e tests passed.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -1269,6 +1269,16 @@
             ]
           }
         },
+        "marinerkata": {
+          "current": {
+            "versionsV2": [
+             {
+                "renovateTag": "registry=https://packages.microsoft.com/cbl-mariner/2.0/prod/base/x86_64/repodata, name=kubernetes-cri-tools, os=mariner, release=2.0",
+                "latestVersion": "1.32.0-3.cm2"
+              }
+            ]
+          }
+        },
         "azurelinux": {
           "v3.0": {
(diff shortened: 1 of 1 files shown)


parts/common/components.json
```diff
@@ -1269,6 +1269,16 @@
             ]
           }
         },
+        "marinerkata": {
+          "current": {
+            "versionsV2": [
+             {
+                "renovateTag": "registry=https://packages.microsoft.com/cbl-mariner/2.0/prod/base/x86_64/repodata, name=kubernetes-cri-tools, os=mariner, release=2.0",
+                "latestVersion": "1.32.0-3.cm2"
```
djsly, line 1277: Since it's the same url used by the nornmal "mariner" property. Could we use the same configuration for both ? Or they need to be distinct?

djsly (approved):

parts/common/components.json
Devinwong: no we are not able to use the same configuration. If it's both AzureLinux v2 and v3, then we can. But this case it's Mariner v2 and MarinerKata, (different OS), meanwhile Azurelinux v3 is using a different url, so we can't.

PR #6555, opened 2025-06-24
feat: enable lsm bpf on Ubuntu 24.04 and Azure Linux 3.0

santhoshmprabhu:
**What type of PR is this?**
/kind feature

This PR configures LSM BPF on Ubuntu 24.04 and Azure Linux 3.0 VMs. We need LSM BPF for our ongoing project within ACN to enable BPF host routing in Cilium CNI.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -668,6 +668,60 @@ else
 fi
 capture_benchmark "${SCRIPT_NAME}_finish_installing_bcc_tools"
 
+# Configure LSM modules to include BPF
+configureLsmWithBpf() {
+  echo "Configuring LSM modules to include BPF..."
+
+  # Read current LSM modules
+  if [ ! -f /sys/kernel/security/lsm ]; then
```
djsly, line 676: for this to be present, does it need to be in the kernel ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -668,6 +668,60 @@ else
...
+    echo "Warning: /sys/kernel/security/lsm not found, skipping LSM configuration"
+    return 0
+  fi
+
+  local current_lsm=$(cat /sys/kernel/security/lsm)
+  echo "Current LSM modules: $current_lsm"
+
+  # Prepend bpf to the LSM list if not already present
+  if [[ "$current_lsm" != *"bpf"* ]]; then
```
djsly: shouldnt this done only when Cilium is used ?

vhdbuilder/packer/install-dependencies.sh
santhoshmprabhu: This file contains the list of current LSM modules. We're checking whether BPF is part of that, and if not, enabling.

vhdbuilder/packer/install-dependencies.sh
santhoshmprabhu: As discussed offline, we can't check for Cilium as part of node image creation. So enabling on 2404 and azl 3.0.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -584,6 +584,47 @@ testLtsKernel() {
 
 }
 
+testLSMBPF() {
```
djsly, line 587: thi needs to be added to the list of tests that are called.

djsly (changes requested):

vhdbuilder/packer/test/linux-vhd-content-test.sh
santhoshmprabhu: Done

PR #6563, opened 2025-06-26
feat: support dynamically booting from NVIDIA/non-NVIDIA Ubuntu kernels

chewi:
**What type of PR is this?**

/kind feature

GB200 instances require a different kernel to other arm64 instances. It would be most efficient to include both kernel types in the same image. This feature uses GRUB's `smbios` command to decide which kernel to boot by default based on the CPU manufacturer.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

I am not familiar with AgentBaker. The feature has been tested manually by myself using the latest Canonical:ubuntu-24_04-lts:specialized-hardware:latest image under QEMU. It has also been manually tested by @abenn135 on real Azure instances. The packer changes are untested.

### DIFF
--- a/vhdbuilder/packer/packer_source.sh
+++ b/vhdbuilder/packer/packer_source.sh
@@ -378,6 +378,16 @@ copyPackerFiles() {
     cpAndMode $PAM_D_COMMON_AUTH_SRC $PAM_D_COMMON_AUTH_DEST 644
     cpAndMode $PAM_D_COMMON_PASSWORD_SRC $PAM_D_COMMON_PASSWORD_DEST 644
     cpAndMode $USU_SH_SRC $USU_SH_DEST 544
+
+    if [[ ${UBUNTU_RELEASE} == "24.04" && ${CPU_ARCH} == "arm64" ]]; then
+      GRUB_AZ_NV_SCRIPT_SRC=/home/packer/10_azure_nvidia
+      GRUB_AZ_NV_SCRIPT_DEST=/etc/grub.d/10_azure_nvidia
+      cpAndMode $GRUB_AZ_NV_SCRIPT_SRC $GRUB_AZ_NV_SCRIPT_DEST 755
+
+      GRUB_AZ_NV_ENV_SRC=/home/packer/51-azure-nvidia.cfg
+      GRUB_AZ_NV_ENV_DEST=/etc/default/grub.d/51-azure-nvidia.cfg
+      cpAndMode $GRUB_AZ_NV_ENV_SRC $GRUB_AZ_NV_ENV_DEST 644
+    fi
   fi
 
   cpAndMode $NOTICE_SRC $NOTICE_DEST 444
--- a/vhdbuilder/packer/pre-install-dependencies.sh
+++ b/vhdbuilder/packer/pre-install-dependencies.sh
@@ -159,9 +159,6 @@ dpkg -l 'linux-*azure*'
 if apt-cache show "linux-image-${CUSTOM_KERNEL_VERSION}" &>/dev/null; then
     echo "Custom kernel ${CUSTOM_KERNEL_VERSION} is available. Proceeding..."
 
-    # Purge old azure kernels
-    DEBIAN_FRONTEND=noninteractive apt-get remove --purge -y $(dpkg-query -W 'linux-*azure*' | awk '$2 != "" { print $1 }' | paste -s)
-
     # Install the new custom kernel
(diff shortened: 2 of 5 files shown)


vhdbuilder/packer/packer_source.sh
```diff
@@ -378,6 +378,14 @@ copyPackerFiles() {
     cpAndMode $PAM_D_COMMON_AUTH_SRC $PAM_D_COMMON_AUTH_DEST 644
     cpAndMode $PAM_D_COMMON_PASSWORD_SRC $PAM_D_COMMON_PASSWORD_DEST 644
     cpAndMode $USU_SH_SRC $USU_SH_DEST 544
+
```
jepio, line 381: guard this with
```
if [[ ${UBUNTU_RELEASE} == "24.04" && "${CPU_ARCH}" == "arm64" ]]; then
```

vhdbuilder/packer/packer_source.sh
```diff
@@ -378,6 +378,14 @@ copyPackerFiles() {
...
+    GRUB_AZ_NV_SCRIPT_SRC=/home/packer/10_azure_nvidia
+    GRUB_AZ_NV_SCRIPT_DEST=/etc/grub.d/10_azure_nvidia
+    cpAndMode $GRUB_AZ_NV_SCRIPT_SRC $GRUB_AZ_NV_SCRIPT_DEST 755
+
+    GRUB_AZ_NV_ENV_SRC=/home/packer/51-azure-nvidia.cfg
```
jepio: Add entries like this to the packer config (arm64-gen2 only) to make the files appear there:
```diff
diff --git a/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json b/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
index bcce341330..37bc47ed70 100644
--- a/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
+++ b/vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
@@ -614,6 +614,11 @@
       "source": "parts/linux/cloud-init/artifacts/localdns-delegate.conf",
       "destination": "/home/packer/localdns-delegate.conf"
     },
+    {
+      "type": "file",
+      "source": "parts/linux/cloud-init/artifacts/grub-d-10_azure_nvidia",
+      "destination": "/home/packer/grub-d-10_azure_nvidia"
+    },
     {
       "type": "shell",
       "inline": [
```

chewi: Is that not also sensitive to the Ubuntu release? Or only 24.04 is used here?

chewi:
The shell linter is being silly. It's disallowing `[[` even though the script has a `/bin/bash` shebang and there are other `[[` instances in the same file.

jepio (approved):
lgtm - we just need to undo the change that switches the single kernel and make sure to install `linux-azure-nvidia` along with the lts kernel (for arm64 only)

chewi:
I believe that's already in this branch. See vhdbuilder/packer/pre-install-dependencies.sh.

djsly:
@anujmaheshwari1 I think your Branch will need a proper clean up to official support both kernel instal

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -159,9 +159,6 @@ dpkg -l 'linux-*azure*'
 if apt-cache show "linux-image-${CUSTOM_KERNEL_VERSION}" &>/dev/null; then
     echo "Custom kernel ${CUSTOM_KERNEL_VERSION} is available. Proceeding..."
 
-    # Purge old azure kernels
-    DEBIAN_FRONTEND=noninteractive apt-get remove --purge -y $(dpkg-query -W 'linux-*azure*' | awk '$2 != "" { print $1 }' | paste -s)
```
abenn135, line 163: Why'd we delete this? This means we'll have 3 kernels, not two: the 6.11 one that comes with the source VHD, the 6.8 LTS one, and the azure-nvidia one.

chewi: Sorry, I misunderstood what this was doing. The existing GB200 image only had the azure-nvidia image, so I assumed this was just removing the azure image. Please correct as necessary.

vhdbuilder/packer/packer_source.sh
jepio: this will happen on all Ubuntu arm64 builds, but copying files to /home/packer is harmless, that directory will be cleaned up after the build.

PR #6564, opened 2025-06-27
bug: select correct ip address for windows

timmy-wright:
**What type of PR is this?**

/kind bug

In some cases, windows CSE was selecting the wrong NIC in a dual NIC machine. This ensures we select the correct NIC.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -458,6 +458,13 @@ function New-ExternalHnsNetwork
         $netIP = Get-NetIPAddress -ifIndex $na.ifIndex -AddressFamily IPv4 -ErrorAction SilentlyContinue -ErrorVariable netIPErr
         if ($netIP)
         {
+            $prefixOrigin = $netIP.PrefixOrigin
+            # Some VMs have multiple NICs
+            if ($prefixOrigin -ne "WellKnown" -and $prefixOrigin -ne "Dhcp") {
+                Write-Log "Skipping adapter $($na.Name) with PrefixOrigin $prefixOrigin"
+                continue
+            }
+
             $managementIP = $netIP.IPAddress
             $adapterName = $na.Name
 
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -459,7 +459,8 @@ function New-ExternalHnsNetwork
         if ($netIP)
         {
             $prefixOrigin = $netIP.PrefixOrigin
-            # Some VMs have multiple NICs
+            # Some VMs have multiple physical NICs where one NIC is Inteliband and should be ignored for kubelet. It's internal and used for
+            # comms with other VMs in the same cluster. Standard_HC44rs is a sample SKU. Note that this code path is not unit tested.
             if ($prefixOrigin -ne "WellKnown" -and $prefixOrigin -ne "Dhcp") {
                 Write-Log "Skipping adapter $($na.Name) with PrefixOrigin $prefixOrigin"
                 continue
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -523,30 +523,19 @@ function New-ExternalHnsNetwork
     }
 
     # we need the default gateway interface to create the external network
-    $defaultGatewayInterface = (Get-NetRoute "0.0.0.0/0").ifIndex
-    if (!$defaultGatewayInterface)
-    {
-        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_NETWORK_ADAPTER_NOT_EXIST -ErrorMessage "Could not find default gateway interface. Please check the network configuration."
-    }
-
-    $na = get-netadapter -ifindex $defaultGatewayInterface
+    $netIP = Get-NetIPAddress -AddressFamily IPv4 -ErrorAction SilentlyContinue -ErrorVariable netIPErr -IpAddress $ipv4Address
     if (!$netIP)
     {
         Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_NETWORK_ADAPTER_NOT_EXIST -ErrorMessage "Failed to find any network adaptor with default gateway"
     }
-    $adapterName = $na.Name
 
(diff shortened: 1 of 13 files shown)


staging/cse/windows/azurecnifunc.ps1
```diff
@@ -458,6 +458,14 @@ function New-ExternalHnsNetwork
         $netIP = Get-NetIPAddress -ifIndex $na.ifIndex -AddressFamily IPv4 -ErrorAction SilentlyContinue -ErrorVariable netIPErr
         if ($netIP)
         {
+            $prefixOrigin = $netIP.PrefixOrigin
+            # Some VMs have multiple physical NICs where one NIC is Inteliband and should be ignored for kubelet. It's internal and used for
+            # comms with other VMs in the same cluster. Standard_HC44rs is a sample SKU. Note that this code path is not unit tested.
+            if ($prefixOrigin -ne "WellKnown" -and $prefixOrigin -ne "Dhcp") {
```
smith1511: The powershell docs here: https://learn.microsoft.com/en-us/powershell/module/nettcpip/get-netipaddress?view=windowsserver2025-ps

Mention 'Dhcp' and 'DHCP' - It looks like 'Dhcp' is the correct one otherwise if we're not 100% sure we could use 'Like' for condition?

smith1511 (approved):

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -16,7 +16,7 @@ function Install-VnetPlugins
 
     # Download Azure VNET CNI plugins.
     # Mirror from https://github.com/Azure/azure-container-networking/releases
-    $zipfile =  [Io.path]::Combine("$AzureCNIDir", "azure-vnet.zip")
+    $zipfile = [Io.path]::Combine("$AzureCNIDir", "azure-vnet.zip")
```
smith1511, line 19: Do we need to backport this to the existing "staging" Windows scripts?  Or just new VHDs that bundle staging?

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -438,133 +497,121 @@ function GenerateAzureStackCNIConfig
...
+
+    $na = get-netadapter -ifindex $defaultGatewayInterface
+    if (!$netIP)
     {
-        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_NOT_FOUND_MANAGEMENT_IP -ErrorMessage "None of the physical network adapters has an IP address"
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_NETWORK_ADAPTER_NOT_EXIST -ErrorMessage "Failed to find any network adaptor with default gateway"
     }
+    $adapterName = $na.Name
+
+    $netIP = Get-NetIPAddress -ifIndex $defaultGatewayInterface -AddressFamily IPv4 -ErrorAction SilentlyContinue -ErrorVariable netIPErr
```
smith1511: General comment: This is obviously a good change, I do worry slightly about the impact though and wonder if it's possible to log and fallback to existing logic some how?  Also I think we have dual stack e2es, but would be good to double check.

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -438,133 +497,121 @@ function GenerateAzureStackCNIConfig
...
-    Write-Log "Creating new HNS network `"ext`""
-    $externalNetwork = "ext"
-    $nas = @(Get-NetAdapter -Physical)
-    $nodeIPs = @()
-
-    if ($nas.Count -eq 0) {
-        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_NETWORK_ADAPTER_NOT_EXIST -ErrorMessage "Failed to find any physical network adapters"
-    }
+    # lookup the ip address defaults from IMDS
+    $MetadataContent = invoke-webrequest -usebasicparsing -uri "http://169.254.169.254/metadata/instance/network/interface?api-version=2021-02-01" -Headers @{ "metadata" = "true" }
```
smith1511: Since you've fixed all the formatting :-)

Invoke-WebRequest -UseBasicParsing -Uri ...

staging/cse/windows/azurecnifunc.ps1
timmy-wright: Have re-written the code so this is no longer relevent.

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -438,133 +497,114 @@ function GenerateAzureStackCNIConfig
...
         $IsDualStackEnabled
     )
     Logs-To-Event -TaskName "AKS.WindowsCSE.NewExternalHnsNetwork" -TaskMessage "Start to create new external hns network"
 
-    Write-Log "Creating new HNS network `"ext`""
-    $externalNetwork = "ext"
-    $nas = @(Get-NetAdapter -Physical)
-    $nodeIPs = @()
+    # lookup the ip address defaults from IMDS
+    $MetadataContent = invoke-webrequest -usebasicparsing -uri "http://169.254.169.254/metadata/instance/network/interface?api-version=2021-02-01" -Headers @{ "metadata" = "true" }
```
timmy-wright: This is where the code change starts (as opposed to the formatting changes)

smith1511 (approved):

staging/cse/windows/azurecnifunc.ps1
timmy-wright: Yup - did this :)

staging/cse/windows/azurecnifunc.ps1
timmy-wright: I think just new VHDs. The fix is for a specific SKU that isn't used much and there's a new windows VHD release in a couple of weeks so I doubt we'd save much time on the release doing some form of hotfix.

staging/cse/windows/azurecnifunc.ps1
timmy-wright: I think there are RP E2E tests for dualstack, but I would like AgentBaker ones as well - have just been working on those and have almost got them working.

Given the linux team gets the node IP from IMDS, I think it's safe for us to do the same. It's cleaner than trying to find the right network adaptor.

PR #6579, opened 2025-07-02
Bug: Add retry logic to IMDS calls for ip addresses

timmy-wright:
**What type of PR is this?**

/kind bug

This PR adds retry and better error handling logic to the IP address detection for windows nodes. It's been tested as part of a hotfix for a Cx.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/staging/cse/windows/azurecnifunc.ps1
+++ b/staging/cse/windows/azurecnifunc.ps1
@@ -494,6 +494,58 @@ function GenerateAzureStackCNIConfig
     Set-ItemProperty -Path $azureCNIConfigFile -Name IsReadOnly -Value $true
 }
 
+
+function GetIpv4AddressFromParsedContent {
+    param (
+        [Parameter(Mandatory = $true)]
+        $ParsedContent
+    )
+
+    if ($ParsedContent[0].ipv4 -and $ParsedContent[0].ipv4.ipAddress -and $ParsedContent[0].ipv4.ipAddress.Count -gt 0) {
+        return $ParsedContent[0].ipv4.ipAddress[0].privateIpAddress
+    } else {
+        return $null
+    }
+}
+
+function GetIpv6AddressFromParsedContent {
+    param (
+        [Parameter(Mandatory = $true)]
+        $ParsedContent
+    )
+
+    if ($ParsedContent[0].ipv6 -and $ParsedContent[0].ipv6.ipAddress -and $ParsedContent[0].ipv6.ipAddress.Count -gt 0) {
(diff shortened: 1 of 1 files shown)


staging/cse/windows/azurecnifunc.ps1
```diff
@@ -494,6 +494,79 @@ function GenerateAzureStackCNIConfig
...
+
+    for ($i = 0; $i -lt $Retries; $i++) {
+        try
+        {
+            $MetadataContent = Invoke-WebRequest -UseBasicParsing -Uri "http://169.254.169.254/metadata/instance/network/interface?api-version=2021-02-01" -Headers @{ "metadata" = "true" } -TimeoutSec 10 -ErrorAction Stop
+            $ParsedContent = $MetadataContent.Content | ConvertFrom-Json
+            $ipv4Address = GetIpv4AddressFromParsedContent -ParsedContent $ParsedContent
+            if (-not $ipv4Address)
+            {
+                Write-Log "Failed to retrieve IPv4 address from metadata. Will retry in $RetryDelaySeconds seconds. Attempt $( $i + 1 ) of $Retries."
```
r2k1, line 546: Is there a reason to retry on parse errors? Do you expect malformed response?

staging/cse/windows/azurecnifunc.ps1
```diff
@@ -494,6 +494,79 @@ function GenerateAzureStackCNIConfig
...
+    else
+    {
+        return $null
+    }
+}
+
+function GetMetadataContent
+{
+    $Retries = 10
+    $RetryDelaySeconds = 15
```
r2k1: I think 15 seconds is too much. It may significantly increase provisioning time. It's on the same node, so I would spam requests multiple times a second

staging/cse/windows/azurecnifunc.ps1
timmy-wright: IMDS sometimes doesn't come up in time so we need to retry. I'm not sure if IMDS gives a 200 response with bad data in these case or a 5xx response, so retry in both cases makes sense.

staging/cse/windows/azurecnifunc.ps1
timmy-wright: Will change this - have gone with 120 retries at 1 per second. That's 2 minutes which should be fine.

bravebeaver (commented):
in general do we need unit test the change?

smith1511 (approved):

PR #6585, opened 2025-07-02
fix sleep before first fetch attempt

jepio:
**What type of PR is this?**

/kind cleanup
/kind bug ?

Looking at packer logs i noticed a bunch of call to `sleep 5`. I count 25 in total during a build, adding up to 2.5 minutes of unnecessary waiting. The sleeps are part of the retry logic when fetching things, and we were accidentally sleeping after the fetch without first checking if the exit condition was met.

Rework the loops to only sleep after the exit condition is false after the first fetch. This mostly affect our packer builds, although there might be CSE flows that fetch binaries, I'm not sure.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -374,29 +378,6 @@ retrycmd_oras_login() {
     return $exit_code
 }
 
-retrycmd_get_binary_from_registry_with_oras() {
```
cameronmeissner, line 377: assuming this is unused

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
djsly: we are using a duplicate: retrycmd_get_tarball_from_registry_with_oras

djsly (approved):

PR #6595, opened 2025-07-03
feat: CSE script enables BPF LSM and reboots nodes

santhoshmprabhu:
**What type of PR is this?**
/kind feature

This PR is a less intrusive alternative to #6555. Rather than enable LSM BPF by default on images, we enable it only when customers enable the Cilium eBPF host routing feature. This requires a node rotation to trigger the CSE script, and a reboot at the end of the CSE script for the change to take effect. This PR is early stage, not ready for review.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -1053,4 +1053,23 @@ extract_tarball() {
     esac
 }
 
+# Check if LSM BPF configuration is needed/possible
+# Returns 0 if BPF LSM should be configured, 1 if it is already configured or can't be.
+shouldConfigureLsmWithBpf() {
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1053,4 +1053,23 @@ extract_tarball() {
     esac
 }
 
+# Check if LSM BPF configuration is needed/possible
+# Returns 0 if BPF LSM should be configured, 1 if it is already configured or can't be.
+shouldConfigureLsmWithBpf() {
```
cameronmeissner, line 1058: I'm assuming we'll also be adding a field to NodeBootstrappingConfiguration to determine whether we should enable lsm with bpf?

PR #6601, opened 2025-07-07
Add option allowing to re-use VM as a base for VHD

r2k1:
- Added `PreProvisionOnly` flag for VM provisioning. If the flag is enabled then kubelet configuration is skipped and preprovision.complete file marker is added on the VM. This VM can be generalized, customized and used as a base for a nodepool in the future.
- Later VM is expected to be created with the same NodeBootstrapConfig. If preprovision.complete flag is present then most of the configuration is skipped (except kubelet and potentially few other piece)
- Significantly reworked E2E to be able to run all existing tests on the new feature. The test takes double the original time. And also double number of running tests. So, currently I enable them with a ENV variable. Not sure yet if we want to enabled them on the main pipelines.


aks-node-controller/proto/aksnodeconfig/v1/config.proto
```diff
@@ -149,4 +149,8 @@ message Configuration {
 
   // IMDS restriction configuration
   ImdsRestrictionConfig imds_restriction_config = 39;
+
+  // Creates a pre-provisioned image for later node spawning.
+  // Skips kubelet and some component configuration for image capture scenarios.
+  bool pre_provision_only = 40;
```
timmy-wright, line 155: Please don't use a bool! Much better to use an enum for phase for extensibility and readability.

e2e/config/azure.go
```diff
@@ -406,7 +437,7 @@ func (a *AzureClient) LatestSIGImageVersionByTag(ctx context.Context, t *testing
 			if tagName != "" {
 				tag, ok := version.Tags[tagName]
 				if !ok || tag == nil || *tag != tagValue {
-					t.Logf("Skipping version %s as it doesn't have tag %s=%s", *version.ID, tagName, tagValue)
+					//t.Logf("Skipping version %s as it doesn't have tag %s=%s", *version.ID, tagName, tagValue)
```
timmy-wright: This creates lots of noise in the logs!

e2e/config/vhd.go
```diff
@@ -104,12 +104,13 @@ var (
...
-		Version: "1.1704411049.2812",
-		Distro:  datamodel.AKSUbuntuContainerd2204Gen2,
-		Gallery: imageGalleryLinux,
+		Name:                     "2204Gen2",
+		OS:                       OSUbuntu,
+		Arch:                     "amd64",
+		Version:                  "1.1704411049.2812",
+		Distro:                   datamodel.AKSUbuntuContainerd2204Gen2,
+		Gallery:                  imageGalleryLinux,
+		UnsupportedKubeletNodeIP: true,
```
timmy-wright, line 114: It'd be good to have a comment as to why we need this flag.

e2e/exec.go
```diff
@@ -77,19 +79,15 @@ func execScriptOnVm(ctx context.Context, s *Scenario, vmPrivateIP, jumpboxPodNam
...
 	steps := []string{
-		fmt.Sprintf("echo '%[1]s' > %[2]s", sshPrivateKey, sshKeyName(vmPrivateIP)),
 		"set -x",
 		fmt.Sprintf("echo %[1]s > %[2]s", quoteForBash(script.script), scriptFileName),
-		fmt.Sprintf("chmod 0600 %s", sshKeyName(vmPrivateIP)),
 		fmt.Sprintf("chmod 0755 %s", scriptFileName),
 		fmt.Sprintf(`scp -i %[1]s -o PasswordAuthentication=no -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o ConnectTimeout=5 %[3]s azureuser@%[2]s:%[4]s`, sshKeyName(vmPrivateIP), vmPrivateIP, scriptFileName, remoteScriptFileName),
-		fmt.Sprintf("%s %s %s", sshString(vmPrivateIP), interpreter, remoteScriptFileName),
 	}
+	steps = append(steps, fmt.Sprintf("%s %s %s", sshString(vmPrivateIP), interpreter, remoteScriptFileName))
```
timmy-wright: Why is this moved below the steps and appended rather than just being added when we create the steps?

e2e/scenario_test.go
```diff
@@ -14,6 +17,21 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6"
 )
 
+func TestMain(m *testing.M) {
```
timmy-wright: Why was this moved into the linux specific test file?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -2,22 +2,23 @@
...
+
 if [ -f /opt/azure/containers/provision.complete ]; then
-      echo "Already ran to success exiting..."
-      exit 0
+    echo "Already ran to success exiting..."
+    exit 0
 fi
 
-for i in $(seq 1 3600); do
+for i in $(seq 1 120); do
```
timmy-wright, line 11: Need to check with the linux team to verify this decrease is OK.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -2,22 +2,23 @@
...
     fi
-    if [ $i -eq 3600 ]; then
+    if [ $i -eq 120 ]; then
         exit $ERR_FILE_WATCH_TIMEOUT
     else
         sleep 1
     fi
 done
-sed -i "/#HELPERSEOF/d" "${CSE_HELPERS_FILEPATH}"
+#sed -i "/#HELPERSEOF/d" "${CSE_HELPERS_FILEPATH}"
```
timmy-wright: Should check with linux team about commenting this out.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -417,7 +432,9 @@ fi
 # Call enableLocalDNS to enable localdns if localdns profile has EnableLocalDNS set to true.
 logs_to_events "AKS.CSE.enableLocalDNS" enableLocalDNS || exit $?
 
-logs_to_events "AKS.CSE.ensureKubelet" ensureKubelet
+if [ "${PRE_PROVISION_ONLY}" != "true" ]; then
```
timmy-wright: Better to use an enum rather than a bool.

This will also re-run the whole pre-provisioning phase when we do the provisioning. I'm not sure what the requirements here are.

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -49,6 +49,15 @@ JSON_STRING=$( jq -n \
 mkdir -p /var/log/azure/aks
 echo $JSON_STRING | tee /var/log/azure/aks/provision.json
 
+# Create stage marker for two-stage workflow
+if [ "${PRE_PROVISION_ONLY}" = "true" ]; then
```
timmy-wright, line 53: enum not bool please!

parts/windows/csecmd.ps1
```diff
@@ -8,7 +8,14 @@ PowerShell
 -AgentKey ''{{ GetParameter "clientPrivateKey" }}''
 -AADClientSecret ''{{ GetParameter "encodedServicePrincipalClientSecret" }}''
 -CSEResultFilePath %SYSTEMDRIVE%\AzureData\provision.complete >> %SYSTEMDRIVE%\AzureData\CustomDataSetupScript.log 2>&1;
-if (!(Test-Path %SYSTEMDRIVE%\AzureData\provision.complete)) { throw 'ExitCode: |50|, Output: |WINDOWS_CSE_ERROR_NO_CSE_RESULT_LOG|, Error: |C:\AzureData\provision.complete is not generated.|'; };
-$result=(Get-Content %SYSTEMDRIVE%\AzureData\provision.complete);
-if ($result -ne '0') { throw $result; };
+
+{{ if GetPreProvisionOnly }}
```
timmy-wright, line 12: enum :)

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -271,6 +272,19 @@ try
         Install-OpenSSH -SSHKeys $SSHKeys
     }
 
+    if (Test-Path C:\AzureData\preprovision.complete) {
```
timmy-wright: Should base this decision on the variable passed in not based on presence of file.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -412,8 +426,16 @@ try
 
     New-ExternalHnsNetwork -IsDualStackEnabled $global:IsDualStackEnabled
 
-    Install-KubernetesServices `
-        -KubeDir $global:KubeDir
+    # Turn off Firewall to enable pods to talk to service endpoints. (Kubelet should eventually do this)
+    netsh advfirewall set allprofiles state off
+
+    # Conditionally install kubelet services based on configuration
+    if (-not $PreProvisionOnly) {
```
timmy-wright: enum :)

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -511,7 +533,13 @@ finally
         Set-Content -Path $CSEResultFilePath -Value "ExitCode: |$global:ExitCode|, Output: |$($global:ErrorCodeNames[$global:ExitCode])|, Error: |$turncatedErrorMessage|"
     }
     else {
-        Set-Content -Path $CSEResultFilePath -Value $global:ExitCode -Force
+        if ($PreProvisionOnly) {
```
timmy-wright: enum

pkg/agent/baker.go
```diff
@@ -1021,6 +1021,7 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"GetLocalDNSMemoryLimitInMB": func() string {
 			return profile.GetLocalDNSMemoryLimitInMB()
 		},
+		"GetPreProvisionOnly": func() bool { return config.PreProvisionOnly },
```
timmy-wright, line 1156: enum

pkg/agent/datamodel/types.go
```diff
@@ -1780,9 +1780,12 @@ type NodeBootstrappingConfiguration struct {
 	// CNI, which will overwrite the `filter` table so that we can only insert to `mangle` table to avoid
 	// our added rule is overwritten by Cilium.
 	InsertIMDSRestrictionRuleToMangleTable bool
-
 	// Version is required for aks-node-controller application to determine the version of the config file.
 	Version string
+
+	// PreProvisionOnly creates a pre-provisioned image for later node spawning.
+	// Skips kubelet and some component configuration for image capture scenarios.
+	PreProvisionOnly bool
```
timmy-wright, line 1805: string or enum!

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -412,8 +426,16 @@ try
 
     New-ExternalHnsNetwork -IsDualStackEnabled $global:IsDualStackEnabled
 
-    Install-KubernetesServices `
-        -KubeDir $global:KubeDir
+    # Turn off Firewall to enable pods to talk to service endpoints. (Kubelet should eventually do this)
+    netsh advfirewall set allprofiles state off
```
timmy-wright, line 417: Was this moved from somewhere else? I guess calling it multiple times doesn't matter.

timmy-wright (changes requested):
change from bool to enum is the main change.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -695,6 +695,14 @@ EOF
     fi
 }
 
+# completeKubeletConfiguration runs only the kubelet-specific configuration
+# This function can be called independently to complete kubelet setup after
+# the initial node provisioning with PreProvisionOnly=true
+completeKubeletConfiguration() {
```
cameronmeissner: looks like this is unused

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -76,6 +77,20 @@ if [ "${ENABLE_SECURE_TLS_BOOTSTRAPPING}" = "true" ]; then
     logs_to_events "AKS.CSE.configureAndStartSecureTLSBootstrapping" configureAndStartSecureTLSBootstrapping
 fi
 
+# Node has been pre-provisioned before. Complete provisioning and skip the rest of the script.
+if [ -f /opt/azure/containers/preprovision.complete ]; then
```
cameronmeissner: im not exactly sure what the SLA is with this new feature, though with your current implementation, any pre-provisioned VHD created from a VM in a particular cluster's node pool can only be used to create further nodes within that same cluster (and technically even the same node pool) - is that the goal?

this is because, from the looks of it, the pre-provisioned VHD will have a number of cluster and AP-specific data stored on disk which is used to provision the underlying node for k8s (cluster CA + apiserver certificate, bootstrap token, azure.json, to name a few)

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -76,6 +77,20 @@ if [ "${ENABLE_SECURE_TLS_BOOTSTRAPPING}" = "true" ]; then
...
+    # Configure kubelet dependencies and kubelet itself
+    # In kubelet-only mode, always ensure kubelet (skip check doesn't apply)
+    logs_to_events "AKS.CSE.ensureKubelet" ensureKubelet
```
cameronmeissner: you probably want `completeKubeletConfiguration` here instead

parts/windows/kuberneteswindowssetup.ps1
cameronmeissner: I think we actually do we want to use the flag file right? otherwise PreProvisionOnly would need to be true when bootstrapping nodes that are using a pre-provisioned VHD that actually want to join the cluster and act as regular nodes

cameronmeissner: same applies for linux provisioning

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: this is probably fine, and won't be needed once we're fully on scriptless (since we won't be relying on cloud-init to write out cse_helpers.sh)

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: it's to remove this after cloud-init writes the file out: https://github.com/Azure/AgentBaker/blob/d40ee30adbc3fcd33519a3aa79003c5a8f2ac2e6/parts/linux/cloud-init/artifacts/cse_helpers.sh#L1036

I don't think we technically need it (also unclear why we have this EOF directive specifically for cse_helpers.sh...) since bash should just interpret it as another comment string

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -49,6 +49,15 @@ JSON_STRING=$( jq -n \
...
+    # Stage 1: Create marker indicating Stage 2 is needed
+    mkdir -p /opt/azure/containers && touch /opt/azure/containers/preprovision.complete
+    echo "Stage 1 complete - kubelet configuration skipped, Stage 2 required" >> /var/log/azure/cluster-provision.log
+    echo "Created preprovision.complete marker file" >> /var/log/azure/cluster-provision.log
+    exit 0
+fi
+# Always create provision.complete to satisfy CSE framework validation
```
cameronmeissner: this comment is misleading, we're only creating provision.complete when PreProvisionOnly = false right?

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: @r2k1 shouldn't functionally change anything, though now that I think about it more couldn't we have AgentBaker return an empty nodecustomdata.yaml (customdata) payload when bootstrapping a node created from a pre-provisioned VHD?

e2e/config/vhd.go
cameronmeissner: yeah... I guess we need it for this VHD since in the scriptless case we'd be using the original provisioning scripts baked onto it, which don't support getting kubelet node IP from IMDS (this image was built early 2024)

r2k1: Some tests checks reference to a VHD objects before performing a validation.
In my tests I built a new VHD object, so the reference check doesn't work and test fails.

In general, I think it's a cleaner way to skip some default validation checks on some edge cases.

parts/linux/cloud-init/artifacts/cse_main.sh
r2k1: This code waits for 60 minutes for `#HELPERSEOF` line to be present in the base script. And deletes this line after.

In my feature I had to run this twice. So it waited for 60 minutes for the comment line. It added maximum delay to feedback loop (until the test times out). Cost me a lot time to figure it out.

parts/windows/kuberneteswindowssetup.ps1
timmy-wright: I'm a fan of having an enum that tells the script what to do, with these three values:
* full provisioning
* "everything but kubelet" 
* kubelet only

I've found booleans make future changes much harder and we end up with lots of booleans and the interactions between them are complex. I think that the node join mode is like this - there are several booleans for several different join modes, but they are mutually exclusive.

e2e/exec.go
r2k1: reverted, I just shuffled things around couple of times

e2e/scenario_test.go
r2k1: reverted

parts/linux/cloud-init/artifacts/cse_config.sh
r2k1: removed

parts/linux/cloud-init/artifacts/cse_main.sh
r2k1: After we finish pre-provisioning we want to execute a user-provided script.
One of the main tasks of such script is to pull and cache user container images.

We want for the VM to be "almost functionable" but to not join the cluster. Then snapshot the VM and create a VHD in user space. Then this VHD will be used to spawn other VMs in the nodepool.

The goal is to have a customized VHD with user-specific software / images. So user can scale up quick. (Some users have customization that may take up to 1 hour to complete).

Currently having nodepool-specific VHD is the goal. Ideally we want more flexibility, but I'm not sure if it's achievable and nodepool-specific is good enough. We can attempt to improve it later.

parts/windows/kuberneteswindowssetup.ps1
r2k1: Yes, I moved it from another script file. It had a TODO comment to move it the main CSE.

I had network issues without moving line. And it's been hard to find it.

e2e/aks_model.go
```diff
@@ -31,42 +31,43 @@ func getLatestGAKubernetesVersion(location string, t *testing.T) (string, error)
 	}
 
 	var latestPatchVersion string
+	msg := fmt.Sprintf("Available Kubernetes versions for location %s:\n", location)
+	defer func() { t.Log(msg) }()
```
Devinwong, line 35: Dose this logging pattern improve logging performance or what is the reason switching to this?

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -49,6 +49,16 @@ JSON_STRING=$( jq -n \
 mkdir -p /var/log/azure/aks
 echo $JSON_STRING | tee /var/log/azure/aks/provision.json
 
+# Create stage marker for two-stage workflow
+if [ "${PRE_PROVISION_ONLY}" = "true" ]; then
+    # Stage 1: Create marker indicating Stage 2 is needed
```
Devinwong, line 54: Please document it somewhere what are done (high level) in stage 1 & 2 and why we need it for pre-provisioning mode. Also under what scenarios we will need pre-provisioning, etc. Thanks!

e2e/aks_model.go
r2k1: It creates less cluter in logs when it's a single log message. It's also easier to read for concurrent test runs. (When test runs concurrently logs can be mixed)

I used `defer` here because we have multiple returns I wanted to be sure the message isn't lost (on error returns for example)

r2k1:
I had chat with @timmy-wright. I'm stil not convienced the enum is the way to go. And he is ready to tolerate due to time pressure.

In the context of NodeBootstrapConfig/AKSNodeConfig adding new value to an enum is a breaking change and should be avoided. So the main benefit is lost. But it's not obvious and people may add a new value without expecting anything to break.

parts/common/components.json
```diff
@@ -247,8 +247,8 @@
         },
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=containernetworking/azure-cns",
-          "latestVersion": "v1.6.29",
-          "previousLatestVersion": "v1.6.27",
+          "latestVersion": "v1.6.27",
```
timmy-wright: Why the downgrade?

vhdbuilder/packer/windows/windows_settings.json
```diff
@@ -39,14 +39,14 @@
       "os_disk_size": "45",
       "base_image_sku": "2025-datacenter-core-smalldisk",
       "windows_image_name": "windows-2025",
-      "base_image_version": "26100.4656.250712",
+      "base_image_version": "26100.4349.250607",
```
timmy-wright: Why the downgrade?

vhdbuilder/release-notes/security-patch/20250714T000000Z/release_notes_delta_20250714T000000Z.txt
timmy-wright, line 1: Why deleted this file?

staging/cse/windows/provisioningscripts/kubeletstart.ps1
```diff
@@ -63,10 +63,6 @@ $KubeletArgListStr = "@($KubeletArgListStr`)"
 # Used in Azure-CNI version of kubeletstart.ps1
 $KubeletCommandLine = "$global:KubeDir\kubelet.exe " + ($KubeletArgList -join " ")
 
-# Turn off Firewall to enable pods to talk to service endpoints. (Kubelet should eventually do this)
-# TODO move this to CSE
-netsh advfirewall set allprofiles state off
```
timmy-wright, line 68: Can we leave this line here? Because the agentbaker changes are released seperately from this file, we might have a point in time where the firewall is not disabled in either place. And disabling it in both places seems better than not disabling it.

timmy-wright (changes requested):
Looks like some files were reverted to earlier versions - likely by accident. And we need to bring the firewall disable command back for windows scripts.

staging/cse/windows/provisioningscripts/kubeletstart.ps1
r2k1: Added back

vhdbuilder/release-notes/security-patch/20250714T000000Z/release_notes_delta_20250714T000000Z.txt
r2k1: I merged with master. And fixed some previously botched merge errors. It should be fine now.

e2e/aks_model.go
timmy-wright: I've noticed the test logging can be interleved with data from different tests - which makes life hard. So this is great to get fixed! Could also be sensible to log the test name with the logging message, but that's for a future PR.

timmy-wright: Would be good to have a comment in the code explaining this approach of using defer to bundle all the logs tho.

e2e/types.go
```diff
@@ -197,3 +197,11 @@ func (s *Scenario) PrepareVMSSModel(ctx context.Context, t *testing.T, vmss *arm
 		vmss.Tags[buildIDTagKey] = &config.Config.BuildID
 	}
 }
+
+func (s *Scenario) IsWindows() bool {
+	return s.VHD.OS == config.OSWindows
```
timmy-wright, line 204: Could also put this method on the VHD class (type *Image) - isOsWindows or something like that. Sadly the OS is just a string otherwise that would be the obvious place to put it.

Nice having it on the Scenario class too tho.

timmy-wright: We do have a method "IsWindowsSIGDistro" on the Distro class, but not sure if that's accessible from here.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -255,17 +255,17 @@ configureK8s() {
         set -x
     fi
 
+    set +x
     if [ "${ENABLE_SECURE_TLS_BOOTSTRAPPING}" = "false" ] && [ -z "${TLS_BOOTSTRAP_TOKEN:-}" ]; then
         # only create the client cert and key if we're not using vanilla/secure TLS bootstrapping
-        set +x
```
timmy-wright, line 260: We need the +x and -x to stop the cert content getting logged.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -502,16 +562,29 @@ finally
...
-        # $JsonString = "ExitCode: |{0}|, Output: |{1}|, Error: |{2}|"
-        # Max length of the full error message returned by Windows CSE is ~256. We use 240 to be safe.
-        $errorMessageLength = "ExitCode: |$global:ExitCode|, Output: |$($global:ErrorCodeNames[$global:ExitCode])|, Error: ||".Length
-        $turncatedErrorMessage = $global:ErrorMessage.Substring(0, [Math]::Min(240 - $errorMessageLength, $global:ErrorMessage.Length))
-        Set-Content -Path $CSEResultFilePath -Value "ExitCode: |$global:ExitCode|, Output: |$($global:ErrorCodeNames[$global:ExitCode])|, Error: |$turncatedErrorMessage|"
-    }
-    else {
-        Set-Content -Path $CSEResultFilePath -Value $global:ExitCode -Force
+    # Create appropriate completion file based on mode
+    if ($PreProvisionOnly) {
```
timmy-wright: Apart for the file name (provision.complete and preprovision.complete), these two branches appear the same. Can we just set the file name in the if branch then have the writing to file as a common block? That'll save some bytes (CSE files are size constrained) and mean they'll stay in sync in future.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -444,6 +451,23 @@ try
...
     }
 
+    Write-Log "Stage1 completed successfully"
+    Logs-To-Event -TaskName "AKS.WindowsCSE.Stage1" -TaskMessage "Stage1 completed successfully"
+}
+
+# ====== STAGE 2: CLUSTER INTEGRATION ======
+# All operations that should only run when connecting to the actual cluster
+function Stage2 {
+    if ($PreProvisionOnly) {
```
timmy-wright: I have a preference to have this logic in the place that invokes the State1 or Stage2 function as then the "stage control flow" is all in one spot.  I was confused when reading the bit that invokes stage1 and stage2 that we seemed to always invoke them and had to read the individual methods to see that control flow.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -487,6 +511,42 @@ try
...
+{
+    Logs-To-Event -TaskName "AKS.WindowsCSE.ExecuteCustomDataSetupScript" -TaskMessage ".\CustomDataSetupScript.ps1 -MasterIP $MasterIP -KubeDnsServiceIp $KubeDnsServiceIp -MasterFQDNPrefix $MasterFQDNPrefix -Location $Location -AADClientId $AADClientId -NetworkAPIVersion $NetworkAPIVersion -TargetEnvironment $TargetEnvironment -CSEResultFilePath $CSEResultFilePath"
+
+    # Exit early if the script has been executed
+    if (Test-Path -Path $CSEResultFilePath -PathType Leaf) {
+        Write-Log "The script has been executed before, will exit without doing anything."
+        return
+    }
+
+    # The provisioning is split into two stages to support VHD image creation workflows:
```
timmy-wright, line 519: It's great having this comment! Can you check to see if we strip comments before sending this CSE to the VM for provisioning? I'm constantly paranoid about CSE size.

parts/linux/cloud-init/artifacts/cse_config.sh
r2k1: Yes I know. I extended the block where the loggin is disabled.

`IF` statement logged the token and broke some tests. Not sure why it didn't fail before

e2e/types.go
r2k1: I'm not sure where is the best place to put it.

It seems like a common check we perform. And we pass `*Scenario` around.
Given the frequence of the usage I put it the easiest to access spot, didn't want to burry it.

parts/windows/kuberneteswindowssetup.ps1
r2k1: I don't have a strong opinion. I think my reasoning was that under normal circumstances we don't skip the logic, so I didn't want to hide it behind IF
And early returns were handy. (Probably not so much after I rewrote it a couple of times.)

parts/windows/kuberneteswindowssetup.ps1
r2k1: We probably don't do it for Windows (but seems like we have something for linux)

parts/windows/kuberneteswindowssetup.ps1
r2k1: updated

parts/windows/kuberneteswindowssetup.ps1
r2k1: done

aks-node-controller/proto/aksnodeconfig/v1/config.proto
djsly: what was the final rational here in the end ?

e2e/aks_model.go
djsly: I guess the only issue I can see if that now we wont get logs until the test either pass or fail. so it will be hard to see if a test is running / hanging unless we kept the Test Start.

e2e/aks_model.go
```diff
@@ -170,6 +171,11 @@ func getBaseClusterModel(clusterName, location string) *armcontainerservice.Mana
 			NetworkProfile: &armcontainerservice.NetworkProfile{
 				NetworkPlugin: to.Ptr(armcontainerservice.NetworkPluginKubenet),
 			},
+			AddonProfiles: map[string]*armcontainerservice.ManagedClusterAddonProfile{
+				"omsagent": {
```
djsly, line 175: why do we need containerInsight ? I have had many issue with those daemonsets in the pass, causing clusters to fail to start. Also the CPU/MEM request  for those pods might affect our small nodes during e2e when it comes to scheduling, something to be aware.

e2e/config/azure.go
djsly: remove commented code

e2e/config/config.go
```diff
@@ -111,7 +114,10 @@ func (c *Configuration) VMIdentityResourceID(location string) string {
 }
 
 func mustLoadConfig() *Configuration {
-	_ = godotenv.Load(".env")
+	err := godotenv.Load(".env")
+	if err != nil {
+		fmt.Printf("Error loading .env file: %s\n", err)
```
djsly, line 119: How bad is it if this has error ? should we stop processing and fail fast ? I'm guessing you added this detection since you had to bang your head trying to understand what was going on ?

e2e/kube.go
```diff
@@ -415,11 +416,16 @@ func daemonsetDebug(t *testing.T, deploymentName, targetNodeLabel, privateACRNam
 					NodeSelector: map[string]string{
 						"kubernetes.azure.com/agentpool": targetNodeLabel,
 					},
-					ImagePullSecrets: []corev1.LocalObjectReference{
-						{
-							Name: secretName,
-						},
-					},
+					ImagePullSecrets: func() []corev1.LocalObjectReference {
```
djsly, line 419: can we explain why we support both cases ?

e2e/test_helpers.go
```diff
@@ -62,10 +61,9 @@ func constructErrorMessage(subscriptionID, location string) string {
 	return fmt.Sprintf("Received second cancellation signal, forcing exit.\nPlease check https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/%s/resourceGroups/%s/overview and delete any resources created by the test suite", subscriptionID, config.ResourceGroupName(location))
 }
 
-func newTestCtx(t *testing.T, location string) context.Context {
+func newTestCtx(t *testing.T) context.Context {
 	if testCtx.Err() != nil {
-		msg := constructErrorMessage(config.Config.SubscriptionID, location)
-		t.Skip("test suite is shutting down: " + msg)
+		t.Skip("test suite is shutting down: ")
```
djsly: are we expecting to write extra text after the `: `

e2e/test_helpers.go
```diff
@@ -114,9 +109,111 @@ func RunScenario(t *testing.T, s *Scenario) {
 
 	s.Location = strings.ToLower(s.Location)
 
-	ctx := newTestCtx(t, s.Location)
+	ctx := newTestCtx(t)
 	ensureLocationInitialized(ctx, t, s.Location)
 
+	if config.Config.TestPreProvision {
```
djsly, line 114: could we run the `Original` in Parallel as well  in the else claue?  why do we need to run `runScenario(t, s)` differently when `TestPreProvision` is `true` ? (which seems to be defaulted to true). I would have expected to `augment` with more test when we want to run the `PreProvision` Suilte

e2e/test_helpers.go
```diff
@@ -146,35 +243,22 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
...
 	if s.AKSNodeConfigMutator != nil {
 		nodeconfig := nbcToAKSNodeConfigV1(nbc)
-
-		// deep copy the node config so that we can mutate it without affecting the original
-		clonedNodeConfig, err := deepcopy.Anything(nodeconfig)
-		if err != nil {
-			s.T.Fatalf("failed to deep copy node config: %v", err)
-		}
-
 		// Pass the cloned clonedNodeConfig to AKSNodeConfigMutator so that it can mutate the properties but not affecting the original one.
```
djsly: Are we still pass a Cloned nodeconfig ?

How come this isn't required anymore ?

e2e/toolkit/strings.go
```diff
@@ -29,11 +29,11 @@ func StrToInt32(s string) int32 {
 	return int32(i)
 }
 
-func LogDuration(duration time.Duration, warningDuration time.Duration, message string) {
+func LogDuration(t *testing.T, duration time.Duration, warningDuration time.Duration, message string) {
 	if duration > warningDuration {
-		fmt.Printf("##vso[task.logissue type=warning;]%s", message)
+		t.Logf("##vso[task.logissue type=warning;] %s", message)
```
djsly, line 34: for my own info, I'm guessing the `vso` logic still works when using logf ? as long as it goes out to `stdout` ?

e2e/types.go
```diff
@@ -151,6 +151,8 @@ type Config struct {
 
 	// Validator is a function where the scenario can perform any extra validation checks
 	Validator func(ctx context.Context, s *Scenario)
+
+	SkipDefaultValidation bool
```
djsly, line 157: could we add a comment explaining what this new bool is for ? why do we want to skipDefaultValiation and what is `Default` validation ?

parts/linux/cloud-init/artifacts/cse_start.sh
djsly: again no explaination as to why this was resolved :)  Tim lost the fight ?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -39,448 +37,501 @@ python3 /opt/azure/containers/provision_redact_cloud_config.py \
...
+                fi
+            fi
+        fi
+    fi
+}
+
+# The provisioning is split into two stages to support VHD image creation workflows:
+#
+# Stage 1: Base image preparation
+#   - Installs and configures all required components (kubelet, containerd, etc.)
```
djsly, line 508: Where / how can we ensure that users are aware that ALL RP controlled versioning / patching won't work with this approach ?  I'm also guessing Embargo support is out the window ? Users are on their own at this point when it comes to CVE/Patching.

e2e/aks_model.go
r2k1: `defer` isn't `t.Cleanup`. It executes in the end of the current function.

r2k1: `t.Log` adds test name

e2e/aks_model.go
r2k1: It's enabled by default. I disabled it because I had issues with extra resource provisioning.

e2e/config/config.go
r2k1: We don't use it in CI, so it always fails there.
It's only for local debugging. My IDE put an incorrect comment prefix and I had issues with spotting it.

e2e/kube.go
r2k1: I didn't change the logic. I just wanted to disable a print from k8s SDK about incorrect format when secreName is empty.

e2e/toolkit/strings.go
r2k1: I think it works (not 100% sure). Can change it again later if it won't.

e2e/aks_model.go
djsly: oh wait, this is disabling it ?! I didnt read that properly I though you were enabling ! good :D

e2e/types.go
r2k1: I thought it's self-explanatory.

We run a set of common validators (like spawning a pod).

I have a hack for preparing a VHD in two-stage scenario where I'm re-using existing e2e logic. I had to ignore the basic validation in the first stage. I'm running the common validators in the second stage.

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1: Added

e2e/aks_model.go
r2k1: @djsly if you have issues with seeing live logs. Ensure you run tests with `-v` or `-json` flag.
Check e2e/README.md for details

parts/common/components.json
r2k1: reverted

e2e/test_helpers.go
r2k1: cleaned msg

e2e/test_helpers.go
r2k1: Yes, it's aleady works this way. The parallel flag for Original set in another function. (And if you attempt to set it twice you will get a panic)

e2e/test_helpers.go
r2k1: Removed the comment. It's from old days when we had a global variable with default.
It's been changed to a function that always create a new object to avoid such issues completely

parts/linux/cloud-init/artifacts/cse_main.sh
r2k1: It's something we will need to deal with. But it can't be solved at this level anyway.

r2k1: VHD still should behaves in the same way. So techincally, there should be no difference for applying patches on current VHD vs cached VHD

e2e/kube.go
r2k1: Prints to stdout breaks reporting on concurrent tests in GoLand. Not sure about VSCode.

e2e/config/azure.go
r2k1: removed

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1: I believe enums break things in NBC. And Tim gave up.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: isn't the logic that applied the updated binaries in stage1 ?

djsly (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -39,448 +37,501 @@ python3 /opt/azure/containers/provision_redact_cloud_config.py \
...
+get_ubuntu_release() {
+    lsb_release -r -s 2>/dev/null || echo ""
+}
+
+# ====== STAGE 1: BASE IMAGE PREPARATION ======
+# This stage prepares the base VHD image with all necessary components and configurations.
+# IMPORTANT: This stage must NOT join the node to the cluster.
+# After completion, this VHD can be used as a base image for creating new node pools.
+# Users may add custom configurations or pull additional container images after this stage.
+function stage1 {
```
cameronmeissner: nit: `stage1` -> `preprovision`?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -39,448 +37,501 @@ python3 /opt/azure/containers/provision_redact_cloud_config.py \
...
-        STS=$?
-    else
-        STS=0
+}
+
+# ====== STAGE 2: CLUSTER INTEGRATION ======
+# This stage performs cluster-specific operations and hardware configurations.
+# After this stage the node should be fully integrated into the cluster.
+# IMPORTANT: This stage should only run when actually joining a node to the cluster. This step should not be run when creating a VHD image
+function stage2 {
```
cameronmeissner: nit: `stage2` -> `provision`?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -39,448 +37,501 @@ python3 /opt/azure/containers/provision_redact_cloud_config.py \
...
+# The provisioning is split into two stages to support VHD image creation workflows:
+#
+# Stage 1: Base image preparation
+#   - Installs and configures all required components (kubelet, containerd, etc.)
+#   - Sets up system configurations that are common across all nodes
+#   - DOES NOT join the node to any cluster
+#   - After this stage, users can add customizations (e.g., pre-pull additional container images)
+#   - The VM can then be captured as a VHD image for use as a node pool base image
+#
+# Stage 2: Cluster integration and hardware setup
```
cameronmeissner: maybe we can look at this as a follow-up optimization, though I think we should think about passing an empty customdata payload (in non-scriptless case) when provisioning a node using a pre-provisioned VHD

cameronmeissner: non-blocking

e2e/vmss.go
```diff
@@ -251,18 +257,14 @@ func extractLogsFromVMWindows(ctx context.Context, s *Scenario) {
...
 	// Poll the result until the operation is completed
 	runCommandResp, err := pollerResp.PollUntilDone(ctx, config.DefaultPollUntilDoneOptions)
-	if err != nil {
-		s.T.Logf("failed to execute run command on VMSS instance: %s", err)
-		return
-	}
-	s.T.Logf("run command executed successfully: %v", runCommandResp)
+	require.NoError(s.T, err, "failed to poll run command on VMSS instance %s", instanceID)
+
+	respJSON, _ := json.MarshalIndent(runCommandResp, "", "  ")
```
jepio, line 267: Linux E2E failing with:
```
./vmss.go:266:17: undefined: json
```

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -39,448 +37,501 @@ python3 /opt/azure/containers/provision_redact_cloud_config.py \
...
+#   - Performs cluster-specific configurations
+#   - Configures hardware-specific components (GPU drivers, MIG partitions, etc.)
+#   - Establishes connection to the API server
+#   - Joins the node to the cluster
+#   - Only runs when actually provisioning a node, not when creating VHD images
+#
+# In typical deployments, both stages run sequentially during node provisioning.
+# For VHD image creation workflows, only stage1 runs initially, and stage2 runs later
+# when nodes are created from that VHD image.
+stage1
```
djsly: lets update stage1/stage2 with better names

Image

timmy-wright (approved):

PR #6616, opened 2025-07-10
feat: Introduce initial Flatcar support

jepio:
**What type of PR is this?**
/kind feature
/kind test

Introduce support for Flatcar in vhdbuilder and AB.

The vhdbuild is based on the Ubuntu packer files, with minimal changes that were needed to get things working. The vhdbuild uses a CustomData file when brining up the image build VM - this provides a nice sample that demonstrates butane/ignition usage. There are some changes (eg. repartitioning) that can only be done through customdata but this also allowed me to make less changes to the the vhdbuild scripts initially.

The AB CustomData/CSE needed to be extended with support for butane/ignition. The butane yaml input needs to be translated into ignition json which is passed as CustomData to the VM. The CustomData can't be gzipped directly, but the same goal is accomplished by having a "wrapper" ignition structure around the actual json - this allows for full input compression (see CustomData vs CustomData.inner in `pkg/agent/testdata/Flatcar/`).

Add an e2e test case for Arm64 and X86_64 images. This was tested using the `GALLERY_SUBSCRIPTION_ID` override that I also included in this PR.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


aks-node-controller/go.mod
```diff
@@ -9,16 +9,25 @@ require (
 	github.com/fsnotify/fsnotify v1.8.0
 	github.com/google/go-cmp v0.7.0
 	github.com/stretchr/testify v1.10.0
-	google.golang.org/protobuf v1.35.2
+	google.golang.org/protobuf v1.36.5
```
cameronmeissner, line 12: we need to configure dependabot to look at go.mod's other than the root...

e2e/node_config.go
```diff
@@ -653,6 +653,10 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 					GalleryName:   "AKSUbuntuEdgeZone",
 					ResourceGroup: "AKS-Ubuntu-EdgeZone",
 				},
+				"AKSFlatcar": {
```
cameronmeissner, line 656: SIGConfig is entirely unused in agentbaker e2e, though thanks for being consistent

e2e/scenario_test.go
```diff
@@ -14,6 +14,52 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6"
 )
 
+func bootDiagnosticsVMConfigMutator(vmss *armcompute.VirtualMachineScaleSet) {
```
cameronmeissner: I think we should have a separate file/package to house common Scenario config mutators like this

cameronmeissner: honestly we probably want boot diagnostics enabled on all our E2E VMSS models, might be worth adding this as a common mutation

packer.mk
```diff
@@ -123,4 +130,12 @@ build-aks-node-controller:
 
 build-lister-binary:
 	@echo "Building lister binary for $(GOARCH)"
-	@bash -c "pushd vhdbuilder/lister && CGO_ENABLED=0 GOOS=linux GOARCH=$(GOARCH) go build -o bin/lister main.go && popd"
\ No newline at end of file
+	@bash -c "pushd vhdbuilder/lister && CGO_ENABLED=0 GOOS=linux GOARCH=$(GOARCH) go build -o bin/lister main.go && popd"
+
+generate-flatcar-customdata: vhdbuilder/packer/flatcar-customdata.json
```
cameronmeissner, line 135: nit: separate targets with a new line

parts/linux/cloud-init/artifacts/flatcar/cse_helpers_flatcar.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+echo "Sourcing cse_helpers_distro.sh for Flatcar"
+
+stub() {
+    echo "${FUNCNAME[1]} stub"
+}
```
cameronmeissner, line 7: nit: missing new line

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -442,7 +442,7 @@ verify_default_route_interface || exit $ERR_LOCALDNS_FAIL
 NETWORK_FILE="$(networkctl --json=short status "${DEFAULT_ROUTE_INTERFACE}" 2>/dev/null | jq -r '.NetworkFile')"
 verify_network_file || exit $ERR_LOCALDNS_FAIL
 
-NETWORK_DROPIN_DIR="${NETWORK_FILE}.d"
+NETWORK_DROPIN_DIR="/run/systemd/network/${NETWORK_FILE##*/}.d"
```
cameronmeissner, line 445: is this a safe change for other OS's other than flatcar?

@SriHarsha001

vhdbuilder/scripts/linux/flatcar/tool_installs_flatcar.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+echo "Sourcing tool_installs_flatcar.sh"
+
+stub() {
+    echo "${FUNCNAME[1]} stub"
+}
```
cameronmeissner, line 7: nit: missing new line

vhdbuilder/packer/flatcar-customdata.json
```diff
@@ -0,0 +1 @@
+{"ignition":{"version":"3.4.0"},"storage":{"files":[{"path":"/etc/extensions/aks-sysext/usr/lib/extension-release.d/extension-release.aks-sysext","contents":{"compression":"","source":"data:,ID%3Dflatcar%0ASYSEXT_LEVEL%3D1.0%0A"},"mode":420},{"overwrite":true,"path":"/etc/flatcar/update.conf","contents":{"compression":"","source":"data:,SERVER%3Ddisabled%0A"},"mode":420},{"path":"/etc/systemd/system/containerd.service.d/50-default-config.conf","contents":{"compression":"","source":"data:,%5BService%5D%0AEnvironment%3DCONTAINERD_CONFIG%3D%2Fetc%2Fcontainerd%2Fconfig.toml%0A"},"mode":420}],"links":[{"path":"/etc/extensions/aks-sysext/usr/local/bin","hard":false,"target":"/opt/bin"}]}}
```
cameronmeissner, line 1: why do we need a yaml and json version of the same payload?

aks-node-controller/go.mod
djsly: it doesn today. I think it only cares if there are CVEs

parts/linux/cloud-init/artifacts/localdns.sh
jepio: from: https://manpages.ubuntu.com/manpages/bionic/en/man5/systemd.network.5.html
>  In addition to [/etc/systemd/network](file://etc/systemd/network), drop-in ".d" directories can be placed in [/lib/systemd/network](file://lib/systemd/network) or
       [/run/systemd/network](file://run/systemd/network) directories. Drop-in files in [/etc](file://etc/) take precedence over those in [/run](file://run/) which in turn
       take precedence over those in /lib. Drop-in files under any of these directories take precedence over the
       main netdev file wherever located. (Of course, since [/run](file://run/) is temporary and [/usr/lib](file://usr/lib) is for vendors, it is
       unlikely drop-ins should be used in either of those places.)

So this will work for all OS's. It was a decision between `/etc` and `/run` and I figured `/run` since the drop-in is regenerated on every unit execution.

e2e/node_config.go
jepio: it looks unused, but if it can't be populated correctly e2e fails with:
```
    vmss.go:52: 
        	Error Trace:	/mnt/vss/_work/1/s/e2e/vmss.go:52
        	            				/mnt/vss/_work/1/s/e2e/test_helpers.go:153
        	            				/mnt/vss/_work/1/s/e2e/test_helpers.go:92
        	            				/mnt/vss/_work/1/s/e2e/scenario_win_test.go:242
        	Error:      	Received unexpected error:
        	            	unexpected error while constructing env-aware sig configuration for AKSFlatcar: sig gallery configuration for AKSFlatcar not found
        	Test:       	Test_Windows23H2Gen2CachingRegression

```

aks-node-controller/go.mod
jepio: What Sylvain said - dependabot does update other `go.mods` but only for CVE's.

parts/linux/cloud-init/artifacts/flatcar/cse_helpers_flatcar.sh
jepio: done

vhdbuilder/scripts/linux/flatcar/tool_installs_flatcar.sh
jepio: done

e2e/scenario_test.go
jepio: I moved the boot diagnostics enablement into the E2E VMSS model definition directly: https://github.com/Azure/AgentBaker/pull/6616/commits/1f6d4d531d2e7f61a0e212381819b90c5f59fad2

next time we add a new mutator i agree that it would make sense to move them into a separate file

vhdbuilder/packer/flatcar-customdata.json
jepio: So the json payload is what needs to be passed to the vm when it is created, but it's not really meant to be edited by humans. Yaml is the human editable format, and it gets transpiled with the `butane` tool. The advantage is that we catch syntax/configuration issues before an instance is booted.

I'm adding both here, because they won't be touched very often.

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1226,6 +1234,56 @@ stages:
...
+        condition: eq('${{ parameters.buildflatcargen2 }}', true)
+        dependsOn: [ ]
+        timeoutInMinutes: 180
+        steps:
+          - bash: |
+              echo '##vso[task.setvariable variable=OS_SKU]Flatcar'
+              echo '##vso[task.setvariable variable=OS_VERSION]flatcar'
+              echo '##vso[task.setvariable variable=IMG_PUBLISHER]kinvolk'
+              echo '##vso[task.setvariable variable=IMG_OFFER]flatcar-container-linux-corevm-amd64'
+              echo '##vso[task.setvariable variable=IMG_SKU]alpha-gen2'
```
djsly, line 1247: is alpha what we want ? shoudl this be reflected in the resulting VHD name ?   or the assumption right now is that AKS will always be using Alpha

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -1226,6 +1234,56 @@ stages:
...
+        condition: eq('${{ parameters.buildflatcargen2arm64 }}', true)
+        dependsOn: [ ]
+        timeoutInMinutes: 180
+        steps:
+          - bash: |
+              echo '##vso[task.setvariable variable=OS_SKU]Flatcar'
+              echo '##vso[task.setvariable variable=OS_VERSION]flatcar'
+              echo '##vso[task.setvariable variable=IMG_PUBLISHER]kinvolk'
+              echo '##vso[task.setvariable variable=IMG_OFFER]flatcar-container-linux-corevm'
+              echo '##vso[task.setvariable variable=IMG_SKU]alpha'
```
djsly, line 1272: any reason why this is not `apha-gen2` ?

e2e/scenario_test.go
```diff
@@ -14,6 +14,40 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v6"
 )
 
+func Test_Flatcar(t *testing.T) {
```
djsly, line 17: This is perfectly fine for the first PR, but I'm curious is we have a test plan in mind for other feature/mode the OS supports.

e2e/vmss.go
```diff
@@ -472,6 +472,11 @@ func getBaseVMSSModel(s *Scenario, customData, cseCmd, location string) armcompu
 				Mode: to.Ptr(armcompute.UpgradeModeAutomatic),
 			},
 			VirtualMachineProfile: &armcompute.VirtualMachineScaleSetVMProfile{
+				DiagnosticsProfile: &armcompute.DiagnosticsProfile{
```
djsly, line 475: nice, this allows us to access to remote console ?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -138,7 +138,7 @@ ERR_CLOUD_INIT_FAILED=223 # Error indicating that cloud-init returned exit code
 # For unit tests, the OS and OS_VERSION will be set in the unit test script.
 # So whether it's if or else actually doesn't matter to our unit test.
 if find /etc -type f,l -name "*-release" -print -quit 2>/dev/null | grep -q '.'; then
-    OS=$(sort -r /etc/*-release | gawk 'match($0, /^(ID_LIKE=(coreos)|ID=(.*))$/, a) { print toupper(a[2] a[3]); exit }')
+    OS=$(sort -r /etc/*-release | gawk 'match($0, /^(ID=(.*))$/, a) { print toupper(a[2]); exit }')
```
djsly, line 141: I dont understand the logic change here, is this a revert of a previous commit ?  or was `coreos` present siince the begininng and it was now causing issues ?

pkg/agent/baker.go
```diff
@@ -47,8 +52,15 @@ func (t *TemplateGenerator) getWindowsNodeBootstrappingPayload(config *datamodel
...
 	// this might seem strange that we're encoding the custom data to a JSON string and then extracting it, but without that serialisation and deserialisation
 	// lots of tests fail.
-	customData := getCustomDataFromJSON(t.getLinuxNodeCustomDataJSONObject(config))
-	return getBase64EncodedGzippedCustomScriptFromStr(customData)
+	var encoded string
+	if config.AgentPoolProfile.IsFlatcar() {
+		customData := getCustomDataFromJSON(t.getFlatcarLinuxNodeCustomDataJSONObject(config))
+		encoded = base64.StdEncoding.EncodeToString([]byte(customData))
+	} else {
+		customData := getCustomDataFromJSON(t.getLinuxNodeCustomDataJSONObject(config))
```
djsly, line 60: I"m guessing AzLinux and Ubuntu returns the same CustomScript.   What are the main differences between those and the Flatcar one ?

djsly: ok I see, you need to mimic the same payload but you are changing it to butane format to Flatcar to properly support it.

pkg/agent/baker.go
```diff
@@ -501,6 +628,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"IsNoneCNI": func() bool {
 			return cs.Properties.OrchestratorProfile.IsNoneCNI()
 		},
+		"IsFlatcar": func() bool {
+			return profile.IsFlatcar()
```
djsly, line 632: should it be in `profile.Distro.` like the other OS SKU boolean functions ?

djsly: hum I guess the Ubuntu ones are not under `.Distro` ... not sure why he Kata and Mariner ones are... :(

pkg/agent/baker.go
```diff
@@ -982,7 +1112,7 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return base64.StdEncoding.EncodeToString(b.Bytes()), nil
 		},
 		"ShouldEnableCustomData": func() bool {
-			return !config.DisableCustomData
+			return !config.DisableCustomData && !profile.IsFlatcar()
```
djsly, line 1115: @Devinwong  / @r2k1   will this prevent Scriptless to work on Flatcar ?

cameronmeissner (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -352,6 +352,8 @@ while IFS= read -r p; do
           installContainerd "${downloadDir}" "${evaluatedURL}" "${version}"
         elif isMarinerOrAzureLinux "$OS"; then
           installStandaloneContainerd "${version}"
+        elif isFlatcar "$OS"; then
+          installStandaloneContainerd "${version}"
```
djsly, line 356: this is the only package missing from the default flatcar market place image ?

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -216,6 +218,10 @@ fi
 
 echo "Using finalized SIG_IMAGE_NAME: ${SIG_IMAGE_NAME}, SIG_GALLERY_NAME: ${SIG_GALLERY_NAME}"
 
+if [ "${SUBSCRIPTION_ID}" != "${GALLERY_SUBSCRIPTION_ID}" ]; then
```
djsly, line 221: thanks for adding this, I feel it will help with other use cases!

djsly (approved):
I didnt any changs to the packer/test/validate_linux_content.sh file, is this compatible ?

.pipelines/.vsts-vhd-builder-release.yaml
jepio: Yes, alpha is the fastest releasing channel, it's not "alpha quality" or anything. I don't think we need to surface this - the plan is to stick to alpha until we're ready to switch to an lts channel at which point we can follow SafeOS naming.

.pipelines/.vsts-vhd-builder-release.yaml
jepio: I admit the naming is inconsistent across offers but the flatcar-container-linux-corevm offer is arm64 gen2 only.
This is correct:
https://www.flatcar.org/docs/latest/installing/cloud/azure/#arm64

e2e/scenario_test.go
jepio: I intend to add test cases as we work through every feature to make sure it works correctly.
First up will be those related to node configuration.

e2e/vmss.go
jepio: Yes, this enables access to serial console too. This has been helpful for troubleshooting boot/join failures in e2e.

parts/linux/cloud-init/artifacts/cse_helpers.sh
jepio: Flatcar has:
ID=flatcar
ID_LIKE=coreos

This line previously captured OS=COREOS. This is legacy from aks-engine.

So this change makes it such that we can match specifically on OS=FLATCAR.

pkg/agent/baker.go
jepio: Yes, it's a different format and encoding but still generated from the same single nodecustomdata.yml template.

pkg/agent/baker.go
jepio: The implementation calls profile.Distro.IsFlatcarDistro(). I was trying to imitate what I found but there are some inconsistencies.

pkg/agent/baker.go
jepio: Haven't tested scriptless yet - probably should.
As far as I can tell this condition only disables a loop waiting for cloudinit to finish at the start of CSE, Flatcar doesn't have cloudinit and there's a customdata is processed to completion in early boot, guaranteed to be done before extensions are executed.

vhdbuilder/packer/install-dependencies.sh
jepio: containerd is actually present in flatcar images, but it's default configuration is in a different location than /etc.

I used this hook to move the configuration to /etc so that the vhd build can properly run containerd for pulling images.
This is smth I intend to revisit when we start making more changes to support flatcar more deeply.

PR #6617, opened 2025-07-10
perf: measure linux TLS bootstrapping latency

cameronmeissner:
**What type of PR is this?**

this PR adds a mechanism to measure how long it takes between when we start kubelet, and when a fresh kubeconfig has been placed on-disk for the kubelet to use to register itself and join the cluster. this new mechanism utilizes a OneShot systemd service which generates guest agent events when a new kubeconfig file is created within kubelet's PKI directory

these measurements technically only tell us how long TLS bootstrapping (whether vanilla or STLS) delays node registration, these measurements are also very close to being how long it takes the node to actually register itself with the apiserver as well, assuming no connectivity issues.

these measurements will be used to more accurately define SLOs/SLIs around STLS bootstrapping latency (we need to either be as fast or faster than what we observe from these measurements in the vanilla TLS bootstrapping case)

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/measure-tls-bootstrapping-latency.sh
```diff
@@ -0,0 +1,78 @@
...
+        --arg EventTid    "0" \
+        '{Timestamp: $Timestamp, OperationId: $OperationId, Version: $Version, TaskName: $TaskName, EventLevel: $EventLevel, Message: $Message, EventPid: $EventPid, EventTid: $EventTid}'
+    )
+    echo ${json_string} > ${EVENTS_LOGGING_DIR}${eventsFileName}.json
+}
+
+waitForTLSBootstrapping() {
+    if ! command -v inotifywait >/dev/null 2>&1; then
+        echo "inotifywait is not available, unable to wait for TLS bootstrapping"
+        exit 0
```
djsly, line 39: maybe explains why we want return 0

parts/linux/cloud-init/artifacts/measure-tls-bootstrapping-latency.sh
```diff
@@ -0,0 +1,78 @@
...
+        fi
+    done
+
+    # check once more if there was a kubeconfig created after finishing the inotifywait loop
+    # to avoid data skewing, we don't emit a guest agent event in this case
+    # this would only happen if we hit a race condition between first checking kubeconfig existence and starting inotifywait
+    if [ -f "$KUBECONFIG_PATH" ]; then
+        echo "kubeconfig now exists at: $KUBECONFIG_PATH"
+    else
+        echo "kubeconfig was not created after ${WATCH_TIMEOUT_SECONDS}s"
```
djsly, line 74: I would create an AKs.Runtime event here to ensure we can have visibility for how many times we hit timeouts. or else those will just be `hidden`

cameronmeissner: added

djsly (approved):

PR #6627, opened 2025-07-14
enable WS2025 e2e tests, improve e2e test speed by replicating in advance

timmy-wright:
**What type of PR is this?**

/kind chore

replicating during the build seems to add no meaningful time, but saves 7 to 10 mins in the e2e tests:

From the "build" step:
* Replicating to westus3 in build: ==> Wait completed after 41 minutes 23 seconds
* Not replicating to westus3 in build ==> ==> Wait completed after 42 minutes 48 seconds

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/test_helpers.go
+++ b/e2e/test_helpers.go
@@ -248,7 +248,7 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
 	require.NoError(s.T, err)
 
 	start := time.Now() // Record the start time
-	createVMSS(ctx, s)
+	vmss := createVMSS(ctx, s)
 
 	err = getCustomScriptExtensionStatus(ctx, s)
 	require.NoError(s.T, err)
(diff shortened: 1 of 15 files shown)


cameronmeissner (approved):

e2e/test_helpers.go
```diff
@@ -266,6 +265,29 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
 		toolkit.LogDuration(ctx, totalElapse, 3*time.Minute, fmt.Sprintf("Node %s took %s to be created and %s to be ready", s.Runtime.VMSSName, toolkit.FormatDuration(creationElapse), toolkit.FormatDuration(readyElapse)))
 	}
 
+	if s.Config.PostProvisioningReboot != nil && *s.Config.PostProvisioningReboot {
+		vmssAboutToRebootTime := time.Now() // Record the start time
+		s.T.Log("Rebooting the VMSS VM to complete post-provisioning...")
+
+		restart, err := config.Azure.VMSSVM.BeginRestart(ctx, *s.Runtime.Cluster.Model.Properties.NodeResourceGroup, s.Runtime.VMSSName, "", nil)
```
r2k1: Why do you reboot node manually. Shouldn't it happen automatically?

Looks like a hack

timmy-wright: There is a requirement for windows Cilium that the node is rebooted after provisioning. I need to chat to Andrew B about this, how it will happen in prod, and when the reboot should happen in agent baker e2e.

e2e/test_helpers.go
```diff
@@ -266,6 +265,29 @@ func prepareAKSNode(ctx context.Context, s *Scenario) {
...
+		require.NoError(s.T, err, "failed to start reboot vmss operation")
+		time.Sleep(60 * time.Second)
```
r2k1: and another hack, time.Sleep usually either leads to slow tests or to flaky tests

timmy-wright: Yeah, the time.Sleep didn't fix the issue. I can probably remove this whole if statement from the PR and try to get it working in a different one.

e2e/test_helpers.go
timmy-wright: Code removed.

e2e/test_helpers.go
timmy-wright: Code removed.

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -25,6 +25,7 @@ SUBSCRIPTION_ID="${SUBSCRIPTION_ID:-$(az account show -o json --query="id" | tr
 GALLERY_SUBSCRIPTION_ID="${GALLERY_SUBSCRIPTION_ID:-${SUBSCRIPTION_ID}}"
 CREATE_TIME="$(date +%s)"
 STORAGE_ACCOUNT_NAME="aksimages${CREATE_TIME}$RANDOM"
+E2E_LOCATION="${E2E_LOCATION:-westus3}"
```
r2k1, line 28: Maybe it's better to have a default and skip replication if the ENV variable isn't set?

Are we going to use it in multiple other pipelines? Maybe we don't need it every time.

PR #6629, opened 2025-07-14
ci: Use debian archive mirror in shellspec dockerfile

jepio:
The previous repo is no longer accessible.

**What type of PR is this?**
/kind failing-test

The shellspec action is failing because the previous debian repo is no longer accessible. Switch to using the debian archive.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/spec/shellspec.Dockerfile
+++ b/spec/shellspec.Dockerfile
@@ -2,7 +2,8 @@
 # Don't forget to log in to the Azure Container Registry before building this image:
 # az acr login --name aksdataplanedev
 FROM aksdataplanedev.azurecr.io/shellspec/shellspec-debian:0.28.1
-RUN apt-get update &&  \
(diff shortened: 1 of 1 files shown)


spec/shellspec.Dockerfile
```diff
@@ -1,8 +1,9 @@
 # Dockerfile
 # Don't forget to log in to the Azure Container Registry before building this image:
 # az acr login --name aksdataplanedev
-FROM aksdataplanedev.azurecr.io/shellspec/shellspec-debian:0.28.1
-RUN apt-get update &&  \
+FROM shellspec/shellspec-debian:0.28.1
```
cameronmeissner: we need to keep the `aksdataplanedev.azurecr.io` to avoid secure supply chain analysis warnings in ADO

jepio: my bad, i was aware of this but must have screwed up my git

jepio: fixed now

PR #6658, opened 2025-07-18
fix test-scan-and-cleanup for flatcar

jepio:
**What type of PR is this?**
/kind failing-test

test-scan-and-cleanup is failing because:
- can't install az-cli on Flatcar and subsequently
- we're unexpectedly running cis-scanning
- some vhd-content-tests were not handling flatcar correctly

I missed this in testing because I was passing `GALLERY_SUBSCRIPTION_ID` in my testing, and `test-scan-and-cleanup` swallows the error and returns success after cleanup.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/test-scan-and-cleanup.sh
+++ b/vhdbuilder/packer/test-scan-and-cleanup.sh
@@ -34,6 +34,12 @@ done
 
 echo "Present working directory: ${PWD}"
 
+# Set GALLERY_SUBSCRIPTION_ID to default to SUBSCRIPTION_ID if not set
+if [ -z "${GALLERY_SUBSCRIPTION_ID}" ]; then
+  GALLERY_SUBSCRIPTION_ID="${SUBSCRIPTION_ID}"
+fi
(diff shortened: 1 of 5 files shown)


vhdbuilder/packer/test-scan-and-cleanup.sh
```diff
@@ -34,6 +34,13 @@ done
 
 echo "Present working directory: ${PWD}"
 
+# Set GALLERY_SUBSCRIPTION_ID to default to SUBSCRIPTION_ID if not set
+if [ -z "${GALLERY_SUBSCRIPTION_ID}" ]; then
+  GALLERY_SUBSCRIPTION_ID="${SUBSCRIPTION_ID}"
+else
+  echo "Using GALLERY_SUBSCRIPTION_ID: ${GALLERY_SUBSCRIPTION_ID}"
```
cameronmeissner: can we always echo this out for clarity?

jepio: done

PR #6676, opened 2025-07-22
Default Rx Buffer change

seguler:
**What type of PR is this?**

/kind feature

Internal benchmarks revealed rx buffer of 2048 for 4 core or more machines outperformed in throughput tests and eliminated buffer related drops in Azure. It was determined that we'd change the default to 2048 for all customers who have more than 4 cores on their machines.

None

**Requirements**:

        perf: Performance Improvements 🚀
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/aks-node-controller/parser/helper.go
+++ b/aks-node-controller/parser/helper.go
@@ -720,7 +720,7 @@ func getEthtoolRxBufferSize() string {
 	return "2048"
 }
 
-// configureEthtoolService creates the environment file and enables the ethtool systemd service
+// configureEthtoolService creates the environment file and enables the ethtool systemd service.
 func configureEthtoolService() error {
 	// Skip configuration in test environments or when not running as root
 	if os.Getenv("SKIP_ETHTOOL_CONFIG") == "true" || os.Geteuid() != 0 {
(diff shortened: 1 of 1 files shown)


cameronmeissner (changes requested):
would need to add corresponding packer config for this change to work

parts/linux/cloud-init/artifacts/ethtool-rx-buffer.sh
```diff
@@ -0,0 +1,54 @@
...
+    exit 0
+fi
+
+# Read configuration from environment file if it exists
+ETHTOOL_RX_BUFFER_SIZE=${ETHTOOL_RX_BUFFER_SIZE:-2048}
+
+echo "Configuring ethtool RX buffer size to ${ETHTOOL_RX_BUFFER_SIZE} on multi-core system (${cpu_cores} cores)"
+
+# Find the primary network interface (exclude loopback)
+primary_interface=$(ip route | grep default | awk '{print $5}' | head -n1)
```
jepio, line 25: is this going to end up finding the hyperv netvsc NIC?

parts/linux/cloud-init/artifacts/ethtool-rx-buffer.service
jepio, line 1: Rather than needing to discover NICs, how about using a udev rule that matches the correct devices and invokes the script?

PR #6688, opened 2025-07-23
Cache refactoring for E2E tests

r2k1:
**What type of PR is this?**

/kind test
/kind cleanup

- Remove `*testing.T` parameter from cached functions to prevent cross-test contamination from assertions (the main reason for logging changes)
- Unified caching logic: Replace duplicated cache handling code with a single cachedFunc implementation
- Better test isolation: Prepare caching system for tests running in different environments/locations

### DIFF
--- a/e2e/config/vhd.go
+++ b/e2e/config/vhd.go
@@ -16,6 +16,11 @@ const (
 )
 
 var (
+	logf = toolkit.Logf
+	log  = toolkit.Log
+)
+
+var (
 	imageGalleryLinux = &Gallery{
 		SubscriptionID:    Config.GallerySubscriptionIDLinux,
 		ResourceGroupName: Config.GalleryResourceGroupNameLinux,
@@ -243,9 +248,6 @@ func (i *Image) String() string {
 	// a starter for a string for debugging.
 	return fmt.Sprintf("%s %s %s %s", i.OS, i.Name, i.Version, i.Arch)
 }
-
-var logf = toolkit.Logf
-var log = toolkit.Log
 
 func GetVHDResourceID(ctx context.Context, i Image, location string) (VHDResourceID, error) {
 	switch {
--- a/e2e/test_helpers.go
+++ b/e2e/test_helpers.go
@@ -27,6 +27,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+var (
+	logf = toolkit.Logf
+	log  = toolkit.Log
+)
+
 // it's important to share context between tests to allow graceful shutdown
 // cancellation signal can be sent before a test starts, without shared context such test will miss the signal
 var testCtx = setupSignalHandler()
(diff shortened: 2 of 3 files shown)


e2e/cache.go
```diff
@@ -130,10 +135,93 @@ func createGalleryImage(ctx context.Context, request CreateGalleryImageRequest)
 	return resp.GalleryImage, nil
 }
 
+var ClusterLatestKubernetesVersion = cachedFunc(clusterLatestKubernetesVersion)
+
+// clusterLatestKubernetesVersion creates a cluster with the latest available Kubernetes version
+func clusterLatestKubernetesVersion(ctx context.Context, location string) (*Cluster, error) {
```
cameronmeissner, line 141: for clarity/consistency, maybe consider adding a `CreateClusterRequest` type with a `Location` field?

e2e/toolkit/log.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"log"
+	"testing"
+)
+
+// testLoggerKey is a private key to prevent *testing.T from being easily accessed
+type testLoggerKey struct{}
+
+func ContextWithLog(ctx context.Context, t *testing.T) context.Context {
```
cameronmeissner: I understand that `*testing.T` is technically a "logger" since it defines a `Logf` method, though I might consider changing this to `ContextWithT` since we're adding/popping `*testing.T`'s on/off the context

e2e/config/vhd.go
```diff
@@ -237,53 +237,37 @@ type Image struct {
...
 	// a starter for a string for debugging.
 	return fmt.Sprintf("%s %s %s %s", i.OS, i.Name, i.Version, i.Arch)
 }
 
-func (i *Image) VHDResourceID(ctx context.Context, t *testing.T, location string) (VHDResourceID, error) {
-	i.vhdMutex.Lock()
-	if i.vhdLocationCache == nil {
-		i.vhdLocationCache = make(map[string]*perLocationVHDCache)
-	}
+var logf = toolkit.Logf
```
cameronmeissner: would prefer to see aliasing towards the top of the file

e2e/test_helpers.go
```diff
@@ -67,9 +67,14 @@ func newTestCtx(t *testing.T) context.Context {
 	}
 	ctx, cancel := context.WithTimeout(testCtx, config.Config.TestTimeout)
 	t.Cleanup(cancel)
+	// T should be used only for logging, not for assertions or any other logic
+	ctx = toolkit.ContextWithLog(ctx, t)
 	return ctx
 }
 
+var logf = toolkit.Logf
```
cameronmeissner: same thing here

e2e/toolkit/log.go
cameronmeissner: if @paulgmiller were here he'd probably advocate that we define a private wrapper interface around `*testing.T` to ensure that we can only ever call `.Helper()` and `.Log()`/`.Logf()`

cameronmeissner: something like `type testingLogger interface { ... } `

cameronmeissner (approved):

e2e/cache.go
r2k1: I thought about this. But it probably not worth artificially complicating it for the concistency sake

e2e/toolkit/log.go
r2k1: I know it's ugly. And I had issues with it in the past.
But in this case it's only used for testing, so the worst case a test will fail.

Initially I just tried to pass `logfunc` instead of `*testing.T`. But it's been quite annoying.

I think it's ok to pass tracing information through context that will never be using for any control flow.

PR #6690, opened 2025-07-24
Fix containerd configuration for gpu nodes without gpu drivers

r2k1:
**What type of PR is this?**

/kind bug

#6678 Introduced a subtle change. GPU-related logic has been moved to the `nodePrep` phase. This includes setting `skip_nvidia_driver_install` flag which is used by `ensureContainerd` and now called in `basePrep` before.

This change broke containerd configuration on gpu nodes without gpu drivers.


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -49,6 +49,7 @@ ERR_GPU_DRIVERS_START_FAIL=84 # nvidia-modprobe could not be started by systemct
 ERR_GPU_DRIVERS_INSTALL_TIMEOUT=85 # Timeout waiting for GPU drivers install
 ERR_GPU_DEVICE_PLUGIN_START_FAIL=86 # nvidia device plugin could not be started by systemctl
 ERR_GPU_INFO_ROM_CORRUPTED=87 # info ROM corrupted error when executing nvidia-smi
+ERR_NVIDIA_DRIVER_INSTALL=88 # Error determining if nvidia driver install should be skipped
```
r2k1: Is it a good error code? I don't know how we pick a number.

But this variables has been referenced in CSE and not set.

timmy-wright: Maybe add to the end of the list in case there was something here previously that was removed? That way the kusto logs are cleaner.

djsly (approved):

PR #6691, opened 2025-07-24
E2E Refactoring

r2k1:
**What type of PR is this?**

/kind test
/kind cleanup

- Simplify resource group caching logic
- Move non-scenario tests to a separate module (components)
- Delete nvidia plugin daemonset and cluster for use-casee without daemonset. Plugin installation should happen in individual tests. Should make e2e slightly cheaper and slight more reliable
- Add retry on SSH errors


e2e/components/components.go
```diff
@@ -0,0 +1,131 @@
...
+func GetNanoserverImagesForVhd(vhd *config.Image) []string {
+	return GetWindowsContainerImages("mcr.microsoft.com/windows/nanoserver:*", getWindowsEnvVarForName(vhd))
+}
+
+func RemoveLeadingV(version string) string {
+	if len(version) > 0 && version[0] == 'v' {
+		return version[1:]
+	}
+	return version
+}
```
r2k1, line 131: This code is just moved

e2e/components/components.go
```diff
@@ -0,0 +1,131 @@
+package components
```
r2k1, line 1: Code in this file is mostly moved

e2e/components/components.go
```diff
@@ -0,0 +1,131 @@
...
+		return sum
+	})
+	return rightVersion
+}
+
+func GetExpectedPackageVersions(packageName, distro, release string) []string {
+	var expectedVersions []string
+	// since we control this json, we assume its going to be properly formatted here
+
+	// Get the project root dynamically
```
r2k1, line 34: That's the only change to make it work in any pkg (relative paths are relative to the pkg where tests are executed)

Devinwong (approved):
lgtm

PR #6695, opened 2025-07-25
Capture more diagnostics from E2E tests

jepio:
**What type of PR is this?**
/kind test

* capture the output of `journalctl` from Flatcar test cases. There is no rsyslog on Flatcar, so /var/log/syslog is empty
* capture serial console content from Linux VMs in general

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/vmss.go
+++ b/e2e/vmss.go
@@ -134,6 +134,9 @@ func extractLogsFromVMLinux(ctx context.Context, s *Scenario) error {
 		"aks-node-controller.log":          "sudo cat /var/log/azure/aks-node-controller.log",
 		"syslog":                           "sudo cat /var/log/" + syslogHandle,
 	}
+	if s.VHD.OS == config.OSFlatcar {
+		commandList["journald"] = "sudo journalctl --boot=0 --no-pager"
+	}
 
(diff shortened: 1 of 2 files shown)


e2e/vmss.go
```diff
@@ -111,7 +112,72 @@ func extractLogsFromVM(ctx context.Context, s *Scenario) {
...
+			}
+
+			if bootDiagResp.SerialConsoleLogBlobURI == nil {
+				continue
+			}
+			if *bootDiagResp.SerialConsoleLogBlobURI == "" {
+				continue
+			}
+			// Save serial console log if available
+			logFile := fmt.Sprintf("serial-console-vm-%s.log", *vmInstance.InstanceID)
```
djsly, line 149: is this file getting uploaded to ADO artifact automatically ?

jepio: yes, the whole directory is.
Eg in the artifacts here: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=131625449&view=artifacts&pathAsName=false&type=publishedArtifacts
one can directly see that the 2404Gen1 kernel is different than 2404Gen2

PR #6374, opened 2025-05-14
feat(stls): add support for secure TLS bootstrapping to linux provisioning

cameronmeissner:
**What type of PR is this?**

**What this PR does / why we need it**:
adds updated support for secure TLS bootstrapping to linux provisioning

the secure-tls-bootstrap-client is invoked in parallel with the rest of CSE/provisioning via a OneShot systemd unit which will be started at the beginning of CSE. The binary will run with a 15s deadline (subject to change) to minimize provisioning latency. Before kubelet is started, we manually check the status of this unit to ensure that bootstrapping suceeded. If it succeeds, the TLS_BOOTSTRAP_TOKEN variable is unset completely and wiped from memory. Otherwise, the bootstrap token will be kept and used as a fall-back to let kubelet perform vanilla TLS bootstrapping.

Once bootstrap tokens are eliminated entirely from AKS, we will also remove bootstrap tokens as a fallback mechanism.

This PR also cleanups up nodecustomdata.yaml by removing the /var/lib/kubelet/bootstrap-kubeconfig, /var/lib/kubelet/kubeconfig, and /etc/default/kubelet write_files from the customdata payload, since CSE already generates these files correctly without the usage of go templates

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version
- [ ] commits are GPG signed and Github marks them as verified

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #6710, opened 2025-07-29
feat(stls): implement secure TLS bootstrapping for windows

cameronmeissner:
**What type of PR is this?**

this PR adds support for secure TLS bootstrapping to Windows node provisioning. Secure TLS bootstrapping, when enabled, is used as an alternative to vanilla TLS bootstrapping to generate kubelet's client certificate. For reference, here's the PR that was previously merged to add support for Linux provisioning: https://github.com/Azure/AgentBaker/pull/6374

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -299,6 +309,8 @@ function BasePrep {
 
     Write-KubeClusterConfig -MasterIP $MasterIP -KubeDnsServiceIp $KubeDnsServiceIp
 
+    Install-SecureTLSBootstrapClient -KubeDir $global:KubeDir -CustomSecureTLSBootstrapClientDownloadUrl $global:CustomSecureTLSBootstrapClientURL
```
timmy-wright, line 312: This function call - if the VHD doesn't have the right version of the custom scripts on it, the function won't exist and the CSE will fail. Same as if the function has the wrong number of parameters. But that's a different problem :)

m365 copilot tells me that there is a way to see if the function exists:

if (Get-Command -Name MyFunction -ErrorAction SilentlyContinue) {
    Write-Host "Function exists"
} else {
    Write-Host "Function does not exist"
}

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -463,6 +475,14 @@ function NodePrep {
...
         Start-InstallCalico -RootDir "c:\" -KubeServiceCIDR $global:KubeServiceCIDR -KubeDnsServiceIp $KubeDnsServiceIp
     }
+    if ($global:TLSBootstrapToken -or $global:EnableSecureTLSBootstrapping) {
+        Write-Log "Removing temporary kube config"
+        $kubeConfigFile = [io.path]::Combine($KubeDir, "config")
+        Remove-Item $kubeConfigFile
+    }
+
+    # Depends on ca.crt and azure.json
+    ConfigureAndStart-SecureTLSBootstrapping -KubeDir $global:KubeDir -APIServerFQDN $MasterIP -CustomSecureTLSBootstrapAADResource $global:CustomSecureTLSBootstrapAADServerAppID
```
timmy-wright, line 485: Same comment here - ConfigureAndStart-SecureTLSBootstrapping might not exist.

timmy-wright (changes requested):
Need protection against version skew between CSE and custom scripts.

PR #6726, opened 2025-07-31
feat: enable CDI in containerd configuration

juan-lee:
Add Container Device Interface (CDI) support to containerd configuration template and implement comprehensive testing for Ubuntu 2204.

Changes:
- Enable CDI in containerdV1ConfigTemplate (baker.go:1427)
- Add Ubuntu 2204 CDI test case with proper validation
- Update artifact streaming test expectations to include CDI
- Generate test data for all affected configurations

The CDI functionality allows containers to access specialized devices and resources through a standardized interface.

Tests: All tests pass with updated configurations

**What type of PR is this?**
/kind feature

enable_cdi will be needed for DRA

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -1424,6 +1424,7 @@ oom_score = -999{{if HasDataDir }}
 root = "{{GetDataDir}}"{{- end}}
 [plugins."io.containerd.grpc.v1.cri"]
   sandbox_image = "{{GetPodInfraContainerSpec}}"
+  enable_cdi = true
```
cameronmeissner, line 1434: did we also want to make the same change for scriptless right now? if so we'll need to modify the corresponding template here: https://github.com/Azure/AgentBaker/blob/master/aks-node-controller/parser/templates/containerd.toml.gtpl

cameronmeissner: also, what about the non-GPU template?

cameronmeissner: I also wonder if it would be worth adding validation for this setting in agentbaker e2e's

juan-lee: @cameronmeissner I've added enable_cdi to the scriptless files. I think we an skip adding a dedicated ab e2e since it should be implicitly covered by other ab e2es.

cameronmeissner: sounds good, I see you added to the default and non-GPU templates for scriptless, should we also be adding to the non-GPU template in AgentBaker as well?

juan-lee: @cameronmeissner the non-GPU naming confused me. I've removed enable_cdi from non-GPU now.

cameronmeissner (approved):

PR #6730, opened 2025-08-01
e2e: Fix race in replicating the same image to two regions

jepio:
**What type of PR is this?**
/kind failing-test
/kind flake

Cleanup some extra ARM clients that were being created while we already have some available.

Attempt to fix the race that happens when an image (in this case Ubuntu2404Gen2) needs to be replicated to two regions. Since the image version update requests carry the whole region list, if they are executed independently based on the initial state, they will each specify replication to only one of the regions, overriding the other attempt. I hope that ARM can serve as a synchronization point so the sequence is:
```
lock
fetch latest region list
start one replication
unlock
lock
fetch latest region list
start second replication
unlock

<wait until both replications are done>
```

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
afaik ARM should be a reliable sync point

e2e/config/azure.go
```diff
@@ -520,19 +516,36 @@ func (a *AzureClient) ensureReplication(ctx context.Context, image *Image, versi
...
-
-	resp, err := galleryImageVersion.BeginCreateOrUpdate(ctx, image.Gallery.ResourceGroupName, image.Gallery.Name, image.Name, *version.Name, *version, nil)
+	resp, err := func() (*runtime.Poller[armcompute.GalleryImageVersionsClientCreateOrUpdateResponse], error) {
+		client, err := armcompute.NewGalleryImageVersionsClient(image.Gallery.SubscriptionID, a.Credential, a.ArmOptions)
+		if err != nil {
+			return nil, fmt.Errorf("create a new images client: %v", err)
+		}
+		a.replicationMutex.Lock()
+		defer a.replicationMutex.Unlock()
+		resp, err := client.Get(ctx, image.Gallery.ResourceGroupName, image.Gallery.Name, image.Name, image.Version, nil)
```
jepio: @cameronmeissner I screwed up this line, it should be *version.Name instead of image.Version.

If you could continue this, it'll require testing with the vhd image build included (so that we are forced to replicate).

e2e/config/azure.go
```diff
@@ -520,19 +516,36 @@ func (a *AzureClient) ensureReplication(ctx context.Context, image *Image, versi
...
+		if err != nil {
+			return nil, fmt.Errorf("getting live image version info: %w", err)
+		}
+		regions := make([]string, 0, len(version.Properties.PublishingProfile.TargetRegions))
+		for _, targetRegion := range resp.Properties.PublishingProfile.TargetRegions {
+			regions = append(regions, *targetRegion.Name)
+		}
+		logf(ctx, "##vso[task.logissue type=warning;]Replicating to region %s, available regions: %s, image version %s", location, strings.Join(regions, ", "), *version.ID)
+		resp.Properties.PublishingProfile.TargetRegions = append(resp.Properties.PublishingProfile.TargetRegions, &armcompute.TargetRegion{
```
r2k1, line 537: I can't spot a place where the value is used

jepio: my bad, friday evening code.
this is now passed to `BeginCreateOrUpdate` for the replication.

e2e/config/azure.go
jepio: fixed this.

jepio:
> I don't thinks this is right. A test for ubuntu shouldn't block test for azure linux. So mutex should be scoped to the image level at least.

You're right in that the first version was a bit simplified. The blocking was for the duration of `GetImageVersion + BeginUpdateImageVersion` which is ~seconds, compared to polling for the completion of the replication which happens without holding the lock and takes ~10 minutes. I've reworked this to now do locking at the `image` granularity.

> 
> Which should already happen there https://github.com/Azure/AgentBaker/blob/master/e2e/cache.go#L207
> 

Thanks. The cache locks initialization of each `Image+Region` pair but we really need to ensure that two region replication requests for the same image are correctly synchronized. This doesn't quite fit the cache paradigm. I did find a way to fit into that design by requesting a mutex scoped to `Image` and passing it down. What do you think?

> Additionally, keep in mind that multiple tests can concurrently.

I'm aware, is there a place where I missed handling this properly?

e2e/cache.go
```diff
@@ -204,6 +204,12 @@ func isNotFoundErr(err error) bool {
 	return false
 }
 
+var CachedReplicationMutex = cachedFunc(prepareReplicationMutex)
```
r2k1, line 207: Sorry for suggestion using this file. I didn't fully understand the issue when suggesting this.

While I understand what's you're doing here. It will be hard to follow for other people.

Maybe something like `ImageReplicationMutex map[Image]sync.Mutex` on `Azure` struct will be easier to understand?

PR #6736, opened 2025-08-01
fix: remove newlines from CA cert to prevent modification during CSE formatting

lilypan26:
**What type of PR is this?**

Fixes bug where we replace newlines with spaces during CSE formatting. This results in CA certs with newlines being in invalid base64 format.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -1040,7 +1040,10 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		},
 		"GetHTTPProxyCA": func() string {
 			if config.HTTPProxyConfig != nil && config.HTTPProxyConfig.TrustedCA != nil {
-				return *config.HTTPProxyConfig.TrustedCA
+				// remove newline so it does not interfere with cse script formatting
+				trustedCA := strings.ReplaceAll(*config.HTTPProxyConfig.TrustedCA, "\n", "")
```
cameronmeissner: nit: maybe move the sanitization to a helper func since we're doing the same thing in multiple spots?

pkg/agent/baker_test.go
```diff
@@ -1648,14 +1677,17 @@ testdomain567.com:53 {
...
 		Entry("AKSUbuntu1804 with custom ca trust", "AKSUbuntu1804+CustomCATrust", "1.18.14", func(config *datamodel.NodeBootstrappingConfiguration) {
 			config.CustomCATrustConfig = &datamodel.CustomCATrustConfig{
-				CustomCATrustCerts: []string{encodedTestCert, encodedTestCert, encodedTestCert},
+				CustomCATrustCerts: []string{encodedTestCert, encodedTestCert, testCertWithNewline},
 			}
 		}, func(o *nodeBootstrappingOutput) {
 			Expect(o.vars["CUSTOM_CA_TRUST_COUNT"]).To(Equal("3"))
 			Expect(o.vars["SHOULD_CONFIGURE_CUSTOM_CA_TRUST"]).To(Equal("true"))
 			Expect(o.vars["CUSTOM_CA_CERT_0"]).To(Equal(encodedTestCert))
 			err := verifyCertsEncoding(o.vars["CUSTOM_CA_CERT_0"])
```
cameronmeissner, line 1686: does CUSTOM_CA_CERT_1 exist? not sure how we handle duplicates

pkg/agent/baker.go
lilypan26: fixed

pkg/agent/baker_test.go
lilypan26: the test expects there are still three so i think we just keep it. I dont think we do any checks against duplicates.

cameronmeissner (approved):

djsly (approved):

PR #6737, opened 2025-08-04
fix: add front end lock to wait_for_apt_locks function

zachary-bailey:
**What type of PR is this?**

/kind bug

This PR adds front end locks to the `wait_for_apt_locks` function. This will prevent multiple processes from trying and failing to get a lock on the front end while conducting package updates and modifications.

The kernel purge and reinstall logic in `pre-install-dependencies` is occasionally failing due to multiple processes each trying to get a lock on the front end.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
@@ -16,7 +16,7 @@ aptmarkWALinuxAgent() {
 }
 
 wait_for_apt_locks() {
-    while fuser /var/lib/dpkg/lock /var/lib/apt/lists/lock /var/cache/apt/archives/lock >/dev/null 2>&1; do
+    while fuser /var/lib/dpkg/lock /var/lib/apt/lists/lock /var/cache/apt/archives/lock /var/lib/dpkg/lock-frontend >/dev/null 2>&1; do
         echo 'Waiting for release of apt locks'
         sleep 3
     done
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -16,7 +16,7 @@ aptmarkWALinuxAgent() {
 }
 
 wait_for_apt_locks() {
-    while fuser /var/lib/dpkg/lock /var/lib/apt/lists/lock /var/cache/apt/archives/lock >/dev/null 2>&1; do
+    while fuser /var/lib/dpkg/lock /var/lib/apt/lists/lock /var/cache/apt/archives/lock /var/lib/dpkg/lock-frontend >/dev/null 2>&1; do
```
zachary-bailey, line 19: Is there any reason not to add this lock? I understand that this function is used throughout CSE, so I am wondering why this lock wasn't included in the first place.

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -140,10 +140,6 @@ if [[ ${UBUNTU_RELEASE//./} -ge 2204 && "${ENABLE_FIPS,,}" != "true" ]] && ! gre
   LTS_HEADERS="linux-headers-azure-lts-${UBUNTU_RELEASE}"
   LTS_MODULES="linux-modules-extra-azure-lts-${UBUNTU_RELEASE}"
 
-  # Acquiring kernel packages to be removed now to prevent lock conflicts during removal and install
-  wait_for_apt_locks
-  OLD_KERNEL_PACKAGES=$(dpkg-query -W 'linux-*azure*' | awk '$2 != "" { print $1 }' | paste -s)
```
djsly, line 145: why removing this moving it inline ?

zachary-bailey: This is how it originally was, just moving things back to the way they were and it results in less lines of code.

PR #6742, opened 2025-08-05
fix artifactstreaming test issues

djsly:
**What type of PR is this?**
/kind test

fix artifactstreaming test issues, moving static config to basePrep and DynamicConfig to nodePrep

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


jepio (approved):
(my review doesn't count but still) lgtm.

Test_Ubuntu2204_ArtifactStreaming_Scriptless is still failing - that's a different root cause (scriptless issue rather than artifactsteraming issue)?

zachary-bailey (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -398,7 +398,11 @@ ensureArtifactStreaming() {
   systemctl link /opt/overlaybd/snapshotter/overlaybd-snapshotter.service
   systemctlEnableAndStart overlaybd-tcmu.service 30
   systemctlEnableAndStart overlaybd-snapshotter.service 30
-  systemctlEnableAndStart acr-nodemon 30
+  
+}
+
+ensureAcrNodeMon() {
+    systemctlEnableAndStart acr-nodemon 30
```
cameronmeissner, line 405: so we just needed a delay between starting overlaybd services and this one specifically?

djsly: acr-nodemon depends on kubelet, not the rest and the rest was not install last and the node cse as exiting, before we had a grace period when the artifact streaming services were installed and we returned from CSE

ganeshkumarashok (approved):

cameronmeissner (approved):
I wonder if we should add a systemd drop-in to explicitly configure the acr-nodemon dependency on kubelet when artifact streaming is enabled (within ensureArtifactStreaming), though I guess for now this works

djsly:
> ystemd drop-in to explicitly configure the acr-nodemon dependency on kubelet when artifact streaming is enabled (within ensureArtifactStreaming), though I guess for now this works

PR #6758, opened 2025-08-06
Combine Windows and Linux build/test infrastructure

r2k1:
Use Linux infrastructure for windows builds. Reduce duplication, maintain only single set of resource. Avoid drift.

Use same VM type we use for Linux. It should improve build performance and reduce number of region capacity errors.
When we use different SKUs for Linux and Windows then either capacity issue will block PRs / Builds.

/kind cleanup


vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -97,42 +97,27 @@ fi
...
-		else
-			VNET_NAME="nodesig-packer-vnet-${PACKER_BUILD_LOCATION}"
-		fi
-	fi
-	if [ "$MODE" = "windowsVhdMode" ]; then
-		VNET_NAME="nodesig-pool-vnet"
-	fi
+  if [ "${ENVIRONMENT,,}" = "prod" ]; then
+    # TODO(cameissner): build out updated pool resources in prod so we don't have to pivot like this
+    VNET_NAME="nodesig-pool-vnet-${PACKER_BUILD_LOCATION}"
```
cameronmeissner, line 117: do these exist for windows?

r2k1: I'm trying to use linux resources

timmy-wright (approved):

PR #6759, opened 2025-08-06
feat: add GB200 VHD skeleton infrastructure

juan-lee:
Add skeleton infrastructure to support GB200 (Grace Blackwell) VHD builds:

- Add pipeline parameter and job for build2404gb200arm64gen2containerd
- Add GB200 VM sizes to FabricManagerGPUSizes for multi-GPU support
- Add GB200 feature flag handling in VHD build scripts
- Add GB200 distro type AKSUbuntuGB200Arm64Containerd2404Gen2
- Use FEATURE_FLAGS=GB200 for hardware-specific customizations

Following pattern established in PR #2318 for Trusted Launch support. Enables future GB200-specific optimizations while maintaining ARM64 compatibility.

**What type of PR is this?**
/kind feature

GB200 support for AKS

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -137,6 +137,10 @@ parameters:
     displayName: Build 2404 ARM64 Gen2 containerd
     type: boolean
     default: true
+  - name: build2404gb200arm64gen2containerd
+    displayName: Build 2404 GB200 ARM64 Gen2 Containerd
+    type: boolean
+    default: false
```
anujmaheshwari1, line 143: do we want this to be true or we'll flip the switch to build weekly later?

pkg/agent/datamodel/types.go
```diff
@@ -142,68 +142,69 @@ type Distro string
...
+	AKSUbuntuEdgeZoneContainerd1804       Distro = "aks-ubuntu-edgezone-containerd-18.04"
+	AKSUbuntuEdgeZoneContainerd1804Gen2   Distro = "aks-ubuntu-edgezone-containerd-18.04-gen2"
+	AKSUbuntuEdgeZoneContainerd2204       Distro = "aks-ubuntu-edgezone-containerd-22.04"
+	AKSUbuntuEdgeZoneContainerd2204Gen2   Distro = "aks-ubuntu-edgezone-containerd-22.04-gen2"
+	AKSUbuntuContainerd2204               Distro = "aks-ubuntu-containerd-22.04"
+	AKSUbuntuContainerd2204Gen2           Distro = "aks-ubuntu-containerd-22.04-gen2"
+	AKSUbuntuContainerd2004CVMGen2        Distro = "aks-ubuntu-containerd-20.04-cvm-gen2"
+	AKSUbuntuArm64Containerd2204Gen2      Distro = "aks-ubuntu-arm64-containerd-22.04-gen2"
+	AKSUbuntuArm64Containerd2404Gen2      Distro = "aks-ubuntu-arm64-containerd-24.04-gen2"
+	AKSUbuntuGB200Arm64Containerd2404Gen2 Distro = "aks-ubuntu-gb200-arm64-containerd-24.04-gen2"
```
keith-ms: I recommend reversing arm64 and gb200 to remain consistent with the other arm64 images.

pkg/agent/datamodel/types.go
```diff
@@ -286,6 +287,7 @@ var AKSDistrosAvailableOnVHD = []Distro{
 	AKSAzureLinuxV3CVMGen2,
 	AKSUbuntuArm64Containerd2204Gen2,
 	AKSUbuntuArm64Containerd2404Gen2,
+	AKSUbuntuGB200Arm64Containerd2404Gen2,
```
keith-ms: The same as above here. Reverse GB200 and Arm64 for consistency.

vhdbuilder/packer/produce-packer-settings.sh
```diff
@@ -298,7 +304,7 @@ if [ "$MODE" = "linuxVhdMode" ] || [ "$MODE" = "windowsVhdMode" ]; then
 		if [[ ${ARCHITECTURE,,} == "arm64" ]] || grep -q "cvm" <<< "$FEATURE_FLAGS" || [[ ${HYPERV_GENERATION} == "V1" ]]; then
 			TARGET_COMMAND_STRING=""
 			if [ "${ARCHITECTURE,,}" = "arm64" ]; then
-				TARGET_COMMAND_STRING+="--architecture Arm64"
+				TARGET_COMMAND_STRING+="--architecture Arm64 --features DiskControllerTypes=SCSI,NVMe"
```
keith-ms, line 307: Will this negatively impact future arm64 VHD builds? I don't see anything that makes this GB200 specific.

.pipelines/.vsts-vhd-builder-release.yaml
juan-lee: let's flip it later

vhdbuilder/packer/test/run-test.sh
```diff
@@ -80,6 +80,13 @@ if [ "${OS_TYPE}" = "Linux" ] && grep -q "cvm" <<< "$FEATURE_FLAGS"; then
     TARGET_COMMAND_STRING="--size Standard_DC8ads_v5 --security-type ConfidentialVM --enable-secure-boot true --enable-vtpm true --os-disk-security-encryption-type VMGuestStateOnly --specialized true"
 fi
 
+# GB200 specific test VM configuration (uses standard ARM64 VM for now)
+if [ "${OS_TYPE}" = "Linux" ] && grep -q "GB200" <<< "$FEATURE_FLAGS"; then
+    echo "GB200: Using ARM64 VM size for testing"
```
keith-ms, line 85: It doesn't look like there's anything here that impacts testing, but perhaps this is just accepting the defaults of the wider context while printing out the message. Adding a note to make sure this is intentional.

vhdbuilder/packer/produce-packer-settings.sh
anujmaheshwari1: all arm64 VHDs are gen2, and we do want them to be tagged with NVMe disk controller type. also this path will only execute if an image definition is not already existing, so this is a one of script that runs when we create a new image definition. so it wont affect any exisiting VHDs but makes sure all future new arm64 image defs are tagged correctly for NVMe

anujmaheshwari1 (approved):

vhdbuilder/packer/test/run-test.sh
juan-lee: @anujmaheshwari1 should I remove this or is it just noise?

vhdbuilder/packer/produce-packer-settings.sh
cameronmeissner: this also only applies to image definitions with our packer gallery (PackerSigGalleryEastUS)

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -84,6 +84,20 @@ elif [ "${OS_TYPE}" = "Linux" ] && grep -q "cvm" <<< "$FEATURE_FLAGS"; then
       } \
     } \
   }"
+elif [ "${OS_TYPE}" = "Linux" ] && grep -q "GB200" <<< "$FEATURE_FLAGS"; then
```
cameronmeissner, line 87: I guess we're not using this feature flag to do anything specific at build time (yet)?

abenn135:
Couple questions:
* Do we want to call the VHD "GB200"? Presumably we'll use this image for GB300 as well, possibly future NVIDIA-built SKUs. Could call it just "NVIDIA", or "SpecializedHardware", the latter is roughly what Canonical uses.
* How will cx choose this VHD? Will we have a new `--os-type` option on the CLI? Or is it custom VHD via headers? Or can RP choose the right VHD based on SKU?

anujmaheshwari1 (approved):

anson627 (approved):

PR #6774, opened 2025-08-08
bugfix: containerd2 version = 3

bravebeaver:
**What type of PR is this?**
bug fix containerd v2 config version string

- the version is wrong 

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/windows/components-test.json
+++ b/vhdbuilder/packer/windows/components-test.json
@@ -890,6 +890,7 @@
   ],
   "OCIArtifacts": [
     {
+      "name": "oci-test-registry",
       "registry": "mcr.microsoft.com/aks/oci-test-registry:*",
       "windowsDownloadLocation": "c:\\akse-cache\\oci-test\\",
       "windowsVersions": [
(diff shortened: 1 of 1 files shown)


pkg/agent/baker.go
```diff
@@ -1525,7 +1525,7 @@ root = "{{GetDataDir}}"{{- end}}
     ConfigPath = "/opt/confidential-containers/share/defaults/kata-containers/configuration-clh-snp.toml"
 {{- end}}
 `
-	containerdV2ConfigTemplate ContainerdConfigTemplate = `version = 2
+	containerdV2ConfigTemplate ContainerdConfigTemplate = `version = 3
```
timmy-wright, line 1529: Why the version update and no other changes?

timmy-wright (changes requested):
Just wondering about the version update from 2 to 3

pkg/agent/baker.go
bravebeaver: this is an oversight from my end. it should be version 3 from teh start.https://github.com/containerd/containerd/blob/main/docs/PLUGINS.md#version-header

bravebeaver: originally we decided to stay in Version 2 for compatibility reason.

it is later discovered that "some plugin".. we use has adopted version 3.

PR #6784, opened 2025-08-11
feat: add Azure Linux OS Guard

hbeberman:
**What type of PR is this?**
/kind feature

This PR adds Azure Linux OS Guard as a build target for AgentBaker. Azure Linux OS Guard is an immutable container host based on Azure Linux 3.0.

To customize Azure Linux OS Guard this PR adds a new "Builder" for [Image Customizer](https://microsoft.github.io/azure-linux-image-tools/imagecustomizer/README.html). 
Image Customizer offline modifies a VHD artifact pulled from ACR/MCR. The OS Guard VHD artifact used in this build has been customized by the Azure Linux team with the expected partition layout and required AKS package installs, so the bulk of the work occurring in AgentBaker is file copying in all of the required scripts and AKS owned service files, running the package and container image caching, and producing all the same build artifacts that Packer builds do.

This PR also adds a new Agent Baker E2E that boots the customized Azure Linux OS Guard image with Trusted Launch enabled.

N/A

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml
```diff
@@ -0,0 +1,307 @@
...
+      permissions: 600
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/localdns.sh
+      destination: /opt/azure/containers/localdns.sh
+      permissions: 755
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/localdns-delegate.conf
+      destination: /etc/systemd/system/localdns.service.d/delegate.conf
+      permissions: 644
+
+  services:
+    enable:
```
cameronmeissner, line 287: you're missing `secure-tls-bootstrap` here, and its corresponding unit file: https://github.com/Azure/AgentBaker/blob/master/parts/linux/cloud-init/artifacts/secure-tls-bootstrap.service

vhdbuilder/packer/imagecustomizer/scripts/build-imagecustomizer-image.sh
```diff
@@ -0,0 +1,82 @@
...
+fi
+
+if [ ! -r "$CONFIG_FILE" ]; then
+    echo "Error: Config file '$CONFIG_FILE' is not readable" >&2
+    exit 1
+fi
+
+IMAGE_PATH="${OUT_DIR}/$CONFIG/$CONFIG.vhd"
+
+BASE_IMAGE_ORAS=$BASE_IMG:$BASE_IMG_VERSION
```
cameronmeissner, line 46: nit: can we note which vars are being set via ADO pipeline yaml, and probably also check to make sure they're set before using them? basically these:

```
              echo '##vso[task.setvariable variable=IMG_CUSTOMIZER_CONTAINER]mcr.microsoft.com/azurelinux/imagecustomizer'
              echo '##vso[task.setvariable variable=IMG_CUSTOMIZER_VERSION]0.17.0'
              echo '##vso[task.setvariable variable=IMG_CUSTOMIZER_CONFIG]azlosguard'
              echo '##vso[task.setvariable variable=BASE_IMG]azlplatform.azurecr.io/azurelinux/3.0/image/osguard'
              echo '##vso[task.setvariable variable=BASE_IMG_VERSION]3.0.20250809.2'
```

vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml
hbeberman: Thanks! Added the service file to the image build but did not set it to enabled by default since [cse_main.sh](https://github.com/Azure/AgentBaker/blob/8e0dda6e7c42939d57fb21a318e1fa6c4c396d90/parts/linux/cloud-init/artifacts/cse_main.sh#L294) will enable and start it if ENABLE_SECURE_TLS_BOOTSTRAPPING=true.

vhdbuilder/packer/imagecustomizer/scripts/build-imagecustomizer-image.sh
hbeberman: Added env var validation and a comment to the top of build-imagecustomizer.sh.

I also moved the hardcoded vars from azlosguard-postinstall.sh into [azlosguard.yml](https://github.com/Azure/AgentBaker/pull/6784/files#diff-229ceb20077d1bd00436ddde8f005c18fcb69658bb363b25b17e9357e3ffd3f9R312). Looking to consume those directly from the pipeline as well but it'll take a feature request into image customizer to implement.

parts/linux/cloud-init/artifacts/cis.sh
```diff
@@ -185,7 +185,7 @@ function maskNfsServer() {
     # Note that on ubuntu systems, it isn't installed but on mariner/azurelinux we need it
     # due to a dependency, but disable it by default.
     if systemctl list-unit-files nfs-server.service >/dev/null; then
-        systemctl --now mask nfs-server || exit $ERR_SYSTEMCTL_MASK_FAIL
```
cameronmeissner, line 188: why remove `--now`?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -254,8 +260,10 @@ EOF
         logs_to_events "AKS.CSE.ensureNoDupOnPromiscuBridge" ensureNoDupOnPromiscuBridge
     fi
 
-    if [ "$OS" = "$UBUNTU_OS_NAME" ] || isMarinerOrAzureLinux "$OS"; then
-        logs_to_events "AKS.CSE.ubuntuSnapshotUpdate" ensureSnapshotUpdate
+    if ! isAzureLinuxOSGuard "$OS" "$OS_VARIANT"; then
```
cameronmeissner, line 263: why not ensure snapshot updates on OS guard?

pkg/agent/datamodel/sig_config.go
```diff
@@ -229,6 +230,7 @@ var AvailableGen2Distros = []Distro{
 	AKSAzureLinuxV3Arm64Gen2,
 	AKSAzureLinuxV3Arm64Gen2FIPS,
 	AKSAzureLinuxV3CVMGen2,
+	AKSAzureLinuxV3OSGuardGen2FIPSTL,
```
cameronmeissner, line 236: do we plan on adding more OSGuard-based distros other than gen2fipsTL in the future?

parts/linux/cloud-init/artifacts/cis.sh
hbeberman: The `--now` attempts to stop the nfs-server service along with masking it, which isn't really necessary in the cis.sh script since it's run right at the end of a packer build and the VM is shutdown shortly after.

When trying to reuse the script in the offline image customizer build the `--now` causes it to fail because systemd is not running.

In both cases the service is correctly masked when the VHD boots in a cluster.

parts/linux/cloud-init/artifacts/cse_main.sh
hbeberman: We dont support snapshot updates on OS Guard because the /usr partition is immutable, blocking package updates on running systems.

pkg/agent/datamodel/sig_config.go
hbeberman: Yes, we'll be adding ARM64 and non-FIPS builds shortly after the x86_64 FIPS version is stood up

vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -222,6 +222,13 @@ isFlatcar() {
         return 0
     fi
 }
+isAzureLinuxOSGuard() {
+    local os="$1"
+
+    if [ "$os" = "AzureLinuxOSGuard" ]; then
+        return 0
+    fi
```
cameronmeissner, line 237: we need a `return 1` at the end here

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
```diff
@@ -0,0 +1,107 @@
...
+# Setup a symlink for lg-redirect-sysext
+mkdir -p /etc/extensions/lg-redirect-sysext/usr/local/
+mkdir -p /opt/bin
+ln -s /opt/bin /etc/extensions/lg-redirect-sysext/usr/local/bin
+# Bind mount /opt/bin to /usr/local/bin during the build since the redirect sysext is not running
+mount --bind /opt/bin /usr/local/bin
+trap "umount /usr/local/bin" EXIT
+
+# Link /opt/azure/containers to /home/packer for postinstall
+ln -s /opt/azure/containers /home/packer
```
cameronmeissner, line 42: I'm probably missing something, though I wouldn't expect `/home/packer` to exist if we're building outside of packer

vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml
```diff
@@ -0,0 +1,316 @@
...
+      destination: /etc/systemd/system/containerd.service
+      permissions: 600
+    # localdns
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/localdns.service
+      destination: /etc/systemd/system/localdns.service
+      permissions: 600
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/localdns.sh
+      destination: /opt/azure/containers/localdns.sh
+      permissions: 755
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/localdns-delegate.conf
```
cameronmeissner, line 282: we also need `vhdbuilder/packer/prefetch.sh` (which is generated on the fly during the build before imagecustomerizer will be invoked) to land in `/opt/azure/containers/prefetch.sh`

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
hbeberman: A lot of the packer-time only scripts reference /home/packer, and rather modify them all its easier and safer to redirect them into our /opt folder during the image customizer builds.

vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml
```diff
@@ -0,0 +1,316 @@
+os:
+  additionalFiles:
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/azlosguard/10-users-ssh-enable.cfg
+      destination: /etc/cloud/cloud.cfg.d/10-users-ssh-enable.cfg
+      permissions: 644
+    - source: /AgentBaker/parts/linux/cloud-init/artifacts/azlosguard/azurelinux-cloud-native.repo
```
cameronmeissner, line 6: this whole config is basically a packer template + `packer_source.sh` replacement right?

vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml
cameronmeissner: ah I found it, nevermind

hbeberman: We're already dropping `vhdbuilder/packer/prefetch.sh` into `/opt/azure/containers/prefetch.sh` on line 35 of this config

vhdbuilder/packer/vhd-scanning.sh
hbeberman: Ack, also adding it to the isFlatcar function above.

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
```diff
@@ -0,0 +1,107 @@
...
+echo -e "=== OS Guard Info === End" >> ${VHD_LOGS_FILEPATH}
+
+# Copy logs and BOM to the output directory
+cp ${VHD_LOGS_FILEPATH} /_imageconfigs/out/release-notes.txt
+cp /var/log/bcc_installation.log /_imageconfigs/out/bcc-tools-installation.log
+chmod 644 /_imageconfigs/out/bcc-tools-installation.log
+cp /opt/azure/containers/image-bom.json /_imageconfigs/out/image-bom.json
+chmod 644 /_imageconfigs/out/image-bom.json
+mv /opt/azure/vhd-build-performance-data.json /_imageconfigs/out/vhd-build-performance-data.json
+chmod 644 /_imageconfigs/out/vhd-build-performance-data.json
```
cameronmeissner, line 107: I'm assuming we don't need to perform the steps we normally would in `cleanup-vhd.sh` since we're not actually booting the base OS?

vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml
hbeberman: Yeah, the bulk of this config is for copying files from AgentBaker into the VHD's filesystem with the right permissions set, along with enabling (and disabling) some services.

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
hbeberman: Correct, we're not booting the system so we don't have any cloud-init, waagent, or user account cruft to clean up.

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
cameronmeissner: ah I was a bit confused by the naming of this file, I see we're running install-dependencies below on L61, I figured we would've done that before invoking this script

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
```diff
@@ -0,0 +1,107 @@
...
+mkdir -p /opt/bin
+ln -s /opt/bin /etc/extensions/lg-redirect-sysext/usr/local/bin
+# Bind mount /opt/bin to /usr/local/bin during the build since the redirect sysext is not running
+mount --bind /opt/bin /usr/local/bin
+trap "umount /usr/local/bin" EXIT
+
+# Link /opt/azure/containers to /home/packer for postinstall
+ln -s /opt/azure/containers /home/packer
+
+### pre-install-dependencies ###
```
cameronmeissner, line 44: so this is a direct replacement for pre-install-dependencies.sh?

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
```diff
@@ -0,0 +1,107 @@
...
+### install-dependencies ###
+# Start containerd to allow container precaching
+containerd &
+CONTAINERD_PID=$!
+echo "Started containerd with PID $CONTAINERD_PID"
+# shellcheck disable=2064
+trap "kill $CONTAINERD_PID" EXIT
+
+# Precache packages and containers from components.json
+/opt/azure/containers/install-dependencies.sh
```
cameronmeissner, line 61: so all the below is a direct replacement for post-install-dependencies.sh... this is making more sense now

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
hbeberman: Yeah, since most of pre-install-dependencies is enabling services, which is taken care of in the customizer yaml, I replicated the few uncovered but necessary bits out into this script.

vhdbuilder/packer/imagecustomizer/scripts/publish-imagecustomizer-image.sh
```diff
@@ -0,0 +1,88 @@
...
+
+capture_benchmark "${SCRIPT_NAME}_prepare_upload_vhd_to_blob"
+
+echo "Uploading ${OUT_DIR}/${CONFIG}.vhd to ${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"
+
+echo "Setting azcopy environment variables with pool identity: $AZURE_MSI_RESOURCE_STRING"
+export AZCOPY_AUTO_LOGIN_TYPE="MSI"
+export AZCOPY_MSI_RESOURCE_STRING="$AZURE_MSI_RESOURCE_STRING"
+export AZCOPY_CONCURRENCY_VALUE="AUTO"
+azcopy copy "${OUT_DIR}/${CONFIG}.vhd" "${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd" --recursive=true || exit $?
```
cameronmeissner, line 45: we should cleanup/remove `"${CLASSIC_BLOB}/${CAPTURED_SIG_VERSION}.vhd"` if we're not planning on generating publishing info (such as during an official release), otherwise just having the gallery image should be sufficient - this is what we do for all over distros

hbeberman: Added a check on GENERATE_PUBLISHING_INFO and we'll now clean up the storage account VHD if we aren't using it for publishing later. Confirmed both cases work in the test pipeline as well.

cameronmeissner (approved):

PR #6802, opened 2025-08-14
fix: oras login before pull kubelet kubectl

fseldow:
**What type of PR is this?**

/kind bug

oras login need to be in front of kubectl and kubelet installation

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -108,6 +108,18 @@ function basePrep {
         logs_to_events "AKS.CSE.removeManDbAutoUpdateFlagFile" removeManDbAutoUpdateFlagFile
     fi
 
+    # oras login must be in front of configureKubeletAndKubectl and ensureKubelet
```
andyliuliming, line 111: is it possible to add some tests instead of just one comment?

e2e/scenario_test.go
```diff
@@ -674,28 +674,39 @@ func Test_Ubuntu2204_AirGap_NonAnonymousACR(t *testing.T) {
 	})
 }
 
-func Test_Ubuntu2204Gen2_ContainerdAirgappedK8sNotCached(t *testing.T) {
+func Test_Ubuntu2204Gen2_ContainerdAirgappedNonAnonymousK8sNotCached(t *testing.T) {
```
cameronmeissner, line 722: should we keep the existing scenario and add an entirely new one to get full coverage?

fseldow: added back the exising scenario

e2e/scenario_test.go
```diff
@@ -698,6 +698,46 @@ func Test_Ubuntu2204Gen2_ContainerdAirgappedK8sNotCached(t *testing.T) {
...
+				}
+				nbc.AgentPoolProfile.LocalDNSProfile = nil
+				// intentionally using private acr url to get kube binaries
+				nbc.AgentPoolProfile.KubernetesConfig.CustomKubeBinaryURL = fmt.Sprintf(
+					"%s.azurecr.io/oss/binaries/kubernetes/kubernetes-node:v%s-linux-amd64",
+					config.PrivateACRNameNotAnon(location),
+					nbc.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion)
+				nbc.K8sComponents.LinuxCredentialProviderURL = fmt.Sprintf(
+					"https://packages.aks.azure.com/cloud-provider-azure/v%s/binaries/azure-acr-credential-provider-linux-amd64-v%s.tar.gz",
+					"1.32.6",
```
ganeshkumarashok: Can we avoid hardcoding k8s versions in e2e tests? These tend to be brittle when k8s versions change or based on how it's cached.

fseldow: it is temp trial, reverted to dynamical version

e2e/config/vhd.go
```diff
@@ -121,10 +121,10 @@ var (
 
 	// without kubelet, kubectl, credential-provider and wasm
 	VHDUbuntu2204Gen2ContainerdAirgappedK8sNotCached = &Image{
-		Name:    "2204Gen2",
+		Name:    "2204gen2containerd",
 		OS:      OSUbuntu,
 		Arch:    "amd64",
-		Version: "1.1725612526.29638",
+		Version: "1.1755204156.32758",
```
cameronmeissner, line 134: make sure this image version is locked

cameronmeissner (approved):

e2e/config/vhd.go
cameronmeissner: just added myself

fseldow: thank you

ganeshkumarashok (approved):

PR #6804, opened 2025-08-14
feat: add azure credential provider pmc install

lilypan26:
**What type of PR is this?**

Add option to install azure credential provider pkg from pmc rather than from URL since it will no longer be published after k8s 1.34

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -717,7 +717,11 @@ EOF
     if [[ $KUBELET_FLAGS == *"image-credential-provider-config"* && $KUBELET_FLAGS == *"image-credential-provider-bin-dir"* ]]; then
         echo "Configure credential provider for both image-credential-provider-config and image-credential-provider-bin-dir flags are specified in KUBELET_FLAGS"
         logs_to_events "AKS.CSE.ensureKubelet.configCredentialProvider" configCredentialProvider
-        logs_to_events "AKS.CSE.ensureKubelet.installCredentialProvider" installCredentialProvider
+        if [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && ! semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.34.0"; then
```
cameronmeissner: sad we don't have shellspec for ensureKubelet, otherwise I'd say let's add a test case for this

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -149,6 +149,14 @@ EOF
     systemctl restart nvidia-persistenced.service || exit 1
 }
 
+installCredentialProviderFromPMC() {
```
cameronmeissner, line 153: where are we calling these?

parts/linux/cloud-init/artifacts/cse_config.sh
lilypan26: yea :( added some shellspec tests for other helpers though

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
lilypan26: called here: https://github.com/Azure/AgentBaker/pull/6804/files#diff-4f9504893023c819ba00993d7db2a80f9883ef64f249e6e61b06d18094f3d425R732

parts/common/components.json
```diff
@@ -1428,6 +1428,53 @@
         }
       }
     },
+    {
+      "name": "azure-acr-credential-provider-pmc",
```
cameronmeissner: seems this now taken care of by #6979?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -726,7 +726,11 @@ EOF
     if [[ $KUBELET_FLAGS == *"image-credential-provider-config"* && $KUBELET_FLAGS == *"image-credential-provider-bin-dir"* ]]; then
         echo "Configure credential provider for both image-credential-provider-config and image-credential-provider-bin-dir flags are specified in KUBELET_FLAGS"
         logs_to_events "AKS.CSE.ensureKubelet.configCredentialProvider" configCredentialProvider
```
cameronmeissner, line 728: probably makes more sense to configure the credential provider after installing it? I'm not sure what this func does off the top of my head though

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -726,7 +726,11 @@ EOF
...
-        logs_to_events "AKS.CSE.ensureKubelet.installCredentialProvider" installCredentialProvider
+        if [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && ! semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.34.0"; then
```
cameronmeissner: had to go digging a little bit, can we call out in comments that `SHOULD_ENFORCE_KUBE_PMC_INSTALL` is coming a tag we're looking up from IMDS?

also, are there plans to remove respect for this tag at some point?

parts/linux/cloud-init/artifacts/cse_config.sh
lilypan26: i looked through and seems like it just generates some config files,  this was the order before so i'm going to keep it that way

parts/linux/cloud-init/artifacts/cse_config.sh
lilypan26: added a comment in cse_main where its defined

parts/linux/cloud-init/artifacts/cse_config.sh
mainred: config and install credential provider are separate tasks. Installing here means only install the binary, while config is to create configuration files

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -726,7 +726,19 @@ EOF
...
+            logs_to_events "AKS.CSE.ensureKubelet.installCredentialProvider" installCredentialProvider
+        else
+            if isMarinerOrAzureLinux "$OS"; then
+                if [ "$OS_VERSION" = "2.0" ]; then # PMC package installation not supported for AzureLinux V2, only V3
```
mainred, line 734: _mariner2/rpm: The RPM build target for Azure Linux 2 (cbl-mariner 2). ⚠️ Deprecated - will be discontinued in November 2025_
OK, we can build package for az linux2? [dalec-build-defs/onboarding.md at main · Azure/dalec-build-defs](https://github.com/Azure/dalec-build-defs/blob/main/onboarding.md)

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -726,7 +726,19 @@ EOF
...
+                    logs_to_events "AKS.CSE.ensureKubelet.installCredentialProvider" installCredentialProvider
+                else
+                    logs_to_events "AKS.CSE.ensureKubelet.installCredentialProviderFromPMC" "installCredentialProviderFromPMC ${KUBERNETES_VERSION}"
```
mainred, line 737: credential provider has its own release cycle, and the version might not match kubernetes version.
Before this, in installCredentialProvider, we allow RP to setup the URL where stores the version.

mainred: In installCredentialProviderFromPMC , we'll get the version from components.json

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -810,6 +810,40 @@ updatePackageDownloadURL() {
     return
 }
 
+# Function to get latestVersion for a given k8sVersion from components.json
+getLatestVersionForK8sVersion() {
```
mainred: nit
```suggestion
getLatestComopnentVersionFromK8sVersion() {
```

djsly (commented):
Could you rebase and update the renovate owner to be the real owner, this shouldbt be Node SIG. 

Also, 1.34 is available, we should look at updating one packge to 1.34 (maybe Ubuntun 24.04 ?)

djsly (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -726,7 +726,19 @@ EOF
     if [[ $KUBELET_FLAGS == *"image-credential-provider-config"* && $KUBELET_FLAGS == *"image-credential-provider-bin-dir"* ]]; then
         echo "Configure credential provider for both image-credential-provider-config and image-credential-provider-bin-dir flags are specified in KUBELET_FLAGS"
         logs_to_events "AKS.CSE.ensureKubelet.configCredentialProvider" configCredentialProvider
-        logs_to_events "AKS.CSE.ensureKubelet.installCredentialProvider" installCredentialProvider
+        if [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && ! semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.34.0"; then
```
fseldow: can we also add one condition of network isolated cluster? currently ni cluster still need to use old path to use oras for 1.34. if BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER is not empty, we still need to implement `installCredentialProvider`
`[ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ]`

.github/renovate.json
```diff
@@ -206,24 +206,14 @@
...
-        "devinwong",
-        "anujmaheshwari1",
-        "cameronmeissner",
-        "lilypan26",
-        "djsly",
-        "UtheMan",
-        "zachary-bailey",
-        "ganeshkumarashok"
+        "mainred",
+        "MartinForReal",
```
mainred: Please remove MartinForReal, who does not maintain credential provider anymore

mainred: Please help all occurrences of MartinForReal in this file.

parts/linux/cloud-init/artifacts/cse_config.sh
lilypan26: added!

parts/linux/cloud-init/artifacts/cse_helpers.sh
lilypan26: renamed to `getLatestPkgVersionFromK8sVersion`

parts/linux/cloud-init/artifacts/cse_config.sh
lilypan26: I think we decided its not worth adding AzureLinux2 support if it will be deprecated Nov 2025, but we can take this as a separate discussion. cc @djsly

djsly: correct, this is k8s 1.34 only so azlinuxv2 isn't supported

djsly (approved):

PR #6805, opened 2025-08-14
fix: Ensure /etc/protocols exists on Flatcar

invidian:
**What type of PR is this?**
/kind bug

Right now on Flatcar /etc/protocols file does not exist, which affects the Azure NPM and causes issues with the network policies.

Until this is resolved upstream, copy the File to /etc to satisfy Azure NPM.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
@invidian we don't allow pull requests from forks, I've added you to https://github.com/orgs/Azure/teams/agentbakerwrite/members which gives you write access to Azure/AgentBaker, please feel free to recreate PR without a fork

PR #6806, opened 2025-08-14
Add dynamic chrony configuration file.

phealy:
**What type of PR is this?**
/kind feature

Older VM SKUs (D/E/F v2/v3) run on hardware that has occasionally less accurate physical clocks. This configuration adds support for using time.windows.com NTP servers (maintained by Microsoft) in lieu of the PHC on those generations. On newer generations both sources will be considered for best accuracy. If NTP isn't available (blocked by networking or for any other reason) the PHC will be used.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
AKS nodes will now use time.windows.com NTP servers, if available, in addition to or instead of the hardware clock.
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -456,6 +456,23 @@ EOF
...
+    # The chrony package uses a different configuration file structure on Ubuntu and Azure Linux
+    if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+        CHRONY_TWC='/etc/chrony/sources.d/chrony-twc.conf'
+    elif isMarinerOrAzureLinux "$OS"; then
+        CHRONY_TWC='/etc/chrony.sources.d/chrony-twc.conf'
+    fi
+    
+    # Patch chrony configuration to prefer NTP servers if running on older SKUs
+    # This will detect D, E, or F series SKUs of generation 2 or 3.
+    IS_DRIFT_SKU="$(curl -s -H "Metadata:true" -m 5 --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01" \
```
cameronmeissner, line 468: we should wrap the IMDS call in retry logic

cameronmeissner: for example: https://github.com/Azure/AgentBaker/blob/7a32efeeffc18e2f6104bfa6e0d5efad7009b389/parts/linux/cloud-init/artifacts/cse_config.sh#L476C5-L476C137

cameronmeissner: is this best-effort or do we also need to fail CSE if the lookup fails?

cameronmeissner: can/should we also combine with #6799?

vhdbuilder/packer/vhd-image-builder-base.json
```diff
@@ -618,6 +618,26 @@
       "source": "parts/linux/cloud-init/artifacts/pam-d-su",
       "destination": "/home/packer/pam-d-su"
     },
+    {
+      "type": "file",
+      "source": "parts/linux/cloud-init/artifacts/chrony.conf.azurelinux",
```
cameronmeissner, line 623: seems like these are missing?

parts/linux/cloud-init/artifacts/cse_main.sh
phealy: Yeah, I meant to create this as a draft PR. I'll add retry logic. I would say this is best effort and we can proceed if it fails (thoughts on which case we fail into?). #6799 is now closed; I realized that the way I needed to handle this I had to do it in CSE rather than ahead of time, because I need to detect SKU on the deployed VM, not the build VM :)

This has two failure cases, we can pick one:
- We default to assuming it is a drift SKU. If detection fails on a modern SKU, the clock is very slightly less accurate if NTP is available (<0.1ms variance in my testing), and no change if NTP isn't available.
- We default to assuming it is not a drift SKU. The clock is more accurate on modern SKUs if NTP is available, but old SKUs experience bigger drift intermittently.

I prefer failing towards the drift case; thoughts?

cameronmeissner: I agree, it sounds like the variance created by the drift case overall is less impactful

PR #6809, opened 2025-08-15
chore: secrets-store-provider-azure v1.7.1

jiashun0011:
**What type of PR is this?**
/kind feature

For CVE fix.

Fixes #

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```
- cve fix to bump secrets-store-provider-azure cache version to v1.7.1
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -344,34 +344,12 @@
       ]
     },
     {
-      "downloadURL": "mcr.microsoft.com/oss/azure/secrets-store/provider-azure:*",
-      "amd64OnlyVersions": [],
-      "multiArchVersionsV2": [
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/azure/secrets-store/provider-azure",
-          "latestVersion": "v1.6.2"
-        }
-      ],
-      "windowsVersions": [
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/azure/secrets-store/provider-azure",
-          "latestVersion": "v1.6.2"
-        }
-      ]
-    },
-    {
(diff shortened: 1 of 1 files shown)


parts/common/components.json
```diff
@@ -343,35 +343,13 @@
...
     {
       "downloadURL": "mcr.microsoft.com/oss/v2/azure/secrets-store/provider-azure:*",
       "amd64OnlyVersions": [],
       "multiArchVersionsV2": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/azure/secrets-store/provider-azure",
-          "latestVersion": "v1.7.0"
-        }
-      ],
-      "windowsVersions": [
```
jiashun0011, line 371: @timmy-wright as you suggested last time, I removed the `windowsVersions` field.

parts/common/components.json
```diff
@@ -343,35 +343,13 @@
         }
       ]
     },
-    {
-      "downloadURL": "mcr.microsoft.com/oss/azure/secrets-store/provider-azure:*",
```
jiashun0011, line 347: @timmy-wright / @Devinwong do we have the practice that we have to cache both lastet and previous versions? if not, I want to remove the previous version and only cache the latest one due to the CVE alerts.

cameronmeissner: it should be fine to only cache the latest version if previousLatest would otherwise bring in CVEs

parts/common/components.json
timmy-wright: Thanks - if it's the same list of versions as for Linux, then there's no point having a seperate section.

timmy-wright (approved):

PR #6810, opened 2025-08-15
fix: Ensure /etc/protocols exists on Flatcar

invidian:
**What type of PR is this?**
/kind bug

Right now on Flatcar /etc/protocols file does not exist, which affects the Azure NPM and causes issues with the network policies.

Until this is resolved upstream, copy the File to /etc to satisfy Azure NPM.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

I am not sure this patch passes the tests, since I don't have experience with agent baker project. Let's find out :)

**Release note**:

```
none
```


jepio (approved):

jepio:
Agentbaker E2E job:
- Flatcar test failed because it runs tests against existing images.
- OSGuard test failed because build failed to produce an image due to quota (resubmitted)

Flatcar image build is not part of PR check-in gate, kicked off a targeted image build + test run here:
https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=133966973&view=results

invidian:
Remaining e2e failure seems unrelated to my changes, since it's only GPU test. Flatcar tests has passed 2/3 times, with one error which seems unrelated to this PR. Can this be merged then?

PR #6823, opened 2025-08-18
CustomCA and HTTPProxyCA support for Flatcar

jepio:
**What type of PR is this?**

/kind feature

Implement Flatcar specific logic when writing out certificates so that they are correctly added to the system certificate store.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


cameronmeissner (approved):
@lilypan26 @UtheMan should probably take a quick glance as well

parts/linux/cloud-init/artifacts/update_certs.sh
```diff
@@ -6,10 +6,18 @@ certDestination="${1:-/usr/local/share/ca-certificates/certs}"
 updateCmd="${2:-update-ca-certificates -f}"
 destPrefix="aks-custom-"
 
+suffix="crt"
+if [ "${certDestination}" = "/etc/ssl/certs" ]; then
```
UtheMan, line 12: could we add a comment here as well, explaining that this is specifically for flatcar and why it's needed? I know you already have one elsewhere, but I feel like extra context here could be helpful too

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -127,18 +127,24 @@ EOF
...
+    suffix="crt"
     if isAzureLinuxOSGuard "$OS" "$OS_VARIANT"; then
         cert_dest="/etc/pki/ca-trust/source/anchors"
         update_cmd="update-ca-trust"
     elif isMarinerOrAzureLinux "$OS"; then
         cert_dest="/usr/share/pki/ca-trust-source/anchors"
         update_cmd="update-ca-trust"
+    elif isFlatcar "$OS"; then
+        cert_dest="/etc/ssl/certs"
+        update_cmd="update-ca-certificates"
```
UtheMan, line 139: Just trying to understand -  the flatcar version of `update-ca-certificates` only works on files present int `/etc/ssl/certs`? 

E.g. for Ubuntu, calling `update-ca-certificates` takes files from `/usr/local/share/ca-certificates` and then creates required files in `/etc/ssl/certs`. For flatcar we want to go straight to `/etc/ssl/certs`?

parts/linux/cloud-init/artifacts/update_certs.sh
jepio: Make sense, I've added a comment here as well.

parts/linux/cloud-init/artifacts/cse_config.sh
jepio: Yes I think that's the only option we have - `/usr/local/share/ca-certificates` is part of the immutable volume in Flatcar, so Flatcar's `update-ca-certificates` only processes `/etc/ssl/certs`.

UtheMan (approved):

PR #6824, opened 2025-08-18
feat: Add support for using OCI Artifacts as dependency components

abeltrano:
**What type of PR is this?**

/kind feature

* Add support for using [OCI artifacts](https://github.com/opencontainers/artifacts) as dependency components.
* Add `oras` as hack tool.
* Ignore generated prefetch script in source control.

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

* Support has only been added to Windows VHD images; support for Linux planned for future.
* This work is a dependency for adding Cilium-based, L3 Windows Container Networking (WCN) stack package.

**Release note**:

```
Add support for OCI artifacts as components for Windows VHDs
```

### DIFF
--- a/vhdbuilder/packer/windows/configure-windows-vhd.ps1
+++ b/vhdbuilder/packer/windows/configure-windows-vhd.ps1
@@ -157,6 +157,59 @@ function Download-FileWithAzCopy
     popd
 }
 
+function Pull-OCIArtifact
+{
+    param (
+        $Name,
+        $Dest
+    )
+
+    if (!(Test-Path -Path "$global:aksTempDir\oras.exe"))
+    {
+        if (!(Test-Path -Path "$global:aksTempDir"))
+        {
+            Write-Log "Creating temp dir $global:aksTempDir for oras"
+            New-Item -ItemType Directory $global:aksTempDir -Force
+        }
+
+        $orasVersion = '1.2.3'
+        $orasZip = "oras_${orasVersion}_windows_amd64.zip"
+        $orasUrl = "https://github.com/oras-project/oras/releases/download/v${orasVersion}/${orasZip}"
+
+        Write-Log "Downloading oras v${orasVersion}"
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/windows/components-test.json
```diff
@@ -887,5 +887,68 @@
...
       }
     }
+  ],
+  "OCIArtifacts": [
+    {
+      "registry": "mcr.microsoft.com/aks/oci-test-registry:*",
+      "windowsDownloadLocation": "c:\\akse-cache\\oci-test\\",
+      "windowsVersions": [
+        {
+          "renovateTag": "<DO_NOT_UPDATE>",
```
cameronmeissner, line 897: renovate should automatically be able to pick up updates within MCR afaik, why the `DO_NOT_UPDATE`?

cc @Devinwong

cameronmeissner: do you plan on caching OCI artifacts on Linux as well?

abeltrano: > renovate should automatically be able to pick up updates within MCR afaik, why the DO_NOT_UPDATE?

I wasn't certain how/whether renovate would handle artifacts vs container images, and given there are no known use cases for auto-updating OCI artifacts, I left this out of the test cases.

> do you plan on caching OCI artifacts on Linux as well?

Yes. Briefly discussed with @timmy-wright and he thought it could be done later. Though, I'll quickly check how much work it'd be to enable support for Linux VHDs, too.

abeltrano: Doesn't seem challenging, but there are some competing priorities, so I've created #6833 to track adding support for Linux VHDs.

Please re-open if you disagree.

timmy-wright: Yeah - the schema seems open to add Linux artifacts. But the linux caching scheme is ... complex with the linux distro and version override ability so implementing it in this PR seems overkill given we don't need it yet.

PR #6839, opened 2025-08-21
chore: Make name optional in OCIArtifact schema

jepio:
**What type of PR is this?**
/kind failing-test

'make validate-components' (part of 'make generate') currently fails with:

    OCIArtifacts.0.name: incomplete value string
    OCIArtifacts.1.name: incomplete value string
    OCIArtifacts.2.name: incomplete value string
    OCIArtifacts.3.name: incomplete value string
    OCIArtifacts.4.name: incomplete value string

We could add a name to the entries, but the only one that comes to mind is repeating the image name in the registry entry. So make name optional.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/schemas/components.cue
+++ b/schemas/components.cue
@@ -95,7 +95,7 @@ package components
 }
 
 #OCIArtifact: {
-	name: string
+	name?: string
 	registry: string
 	windowsDownloadLocation?: string
 	windowsVersions: [...#WindowsVersion]


jepio:
@abeltrano

abeltrano:
> @abeltrano

Thanks for the fix @jepio. This was intended to be optional, and I had a commit marking it as such, but it got lost in a shuffle when I re-organized the PR to avoid including the new component I will be adding.

PR #6840, opened 2025-08-21
Flatcar "scriptless" support

jepio:
**What type of PR is this?**
/kind feature

Add Flatcar support for "scriptless"/aks-node-controller. Not introducing anything new into the contract because the only change in control-flow that we need is to skip waiting for `cloud-init` to complete. This code is guarded by `DisableCustomData` , so we'll make sure to set that to true for Flatcar.

The other change that would be necessary is to generate customdata in ignition format, but we can also rely on the limited `coreos-cloudinit` fallback implementation that Flatcar ships with. It doesn't support enough for the non-scriptless mode but in scriptless mode the customdata is so simple that we can use this approach.

To make `coreos-cloudinit` work we need to tweak `aks-node-controller.service` dependencies slightly.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -131,7 +131,7 @@ func nbcToAKSNodeConfigV1(nbc *datamodel.NodeBootstrappingConfiguration) *aksnod
 
 	config := &aksnodeconfigv1.Configuration{
 		Version:            "v1",
-		DisableCustomData:  false,
+		DisableCustomData:  nbc.AgentPoolProfile.IsFlatcar(),
 		LinuxAdminUsername: "azureuser",
 		VmSize:             config.Config.DefaultVMSKU,
 		ClusterConfig: &aksnodeconfigv1.ClusterConfig{
--- a/parts/linux/cloud-init/artifacts/aks-node-controller.service
+++ b/parts/linux/cloud-init/artifacts/aks-node-controller.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Parse contract and run csecmd
 After=cloud-init.target
+After=oem-cloudinit.service enable-oem-cloudinit.service
 Wants=cloud-init.target
 
 [Service]
@@ -10,3 +11,4 @@ RemainAfterExit=No
 
 [Install]
 WantedBy=cloud-init.target
+WantedBy=oem-cloudinit.service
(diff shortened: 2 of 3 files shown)


e2e/node_config.go
```diff
@@ -131,7 +131,7 @@ func nbcToAKSNodeConfigV1(nbc *datamodel.NodeBootstrappingConfiguration) *aksnod
 
 	config := &aksnodeconfigv1.Configuration{
 		Version:            "v1",
-		DisableCustomData:  false,
+		DisableCustomData:  nbc.AgentPoolProfile.IsFlatcar(),
```
Devinwong, line 134: Will there be a case that it is not `Flatcar` but still `disableCustomData=false`? I am not very familiar with the logic of this variable though. Asking other folks to take a look too.

cameronmeissner: I'd recommend defining a helper func somewhere, which at the moment can just return `nbc.AgentPoolProfile.IsFlatcar()` which we can easily update later if things change, though I do believe this is the only current case where we'd disable customdata (cloud-init) entirely

jepio: The helper would only be used by e2e though. The thing is that this value is supplied from RP - right now RP never explicitly sets it so it defaults to false.

I intend to wire up the same logic on the RP side (`nbc.DisableCustomData=AgentPoolProfile.IsFlatcar()`. Any concerns?

Devinwong: Then it's fine to me. Thanks for confirming.

Devinwong (approved):
you may need to rebase with latest master to pass all the check-in tests

parts/linux/cloud-init/artifacts/aks-node-controller.service
```diff
@@ -9,4 +11,5 @@ ExecStart=/opt/azure/containers/aks-node-controller provision --provision-config
 RemainAfterExit=No
 
 [Install]
-WantedBy=cloud-init.target
\ No newline at end of file
+WantedBy=cloud-init.target
+WantedBy=multi-user.target
```
jepio: Need to fix this: this causes an ordering cycle on ubuntu :/

jepio: found a way: avoid using multi-user.target and switch to oem-cloudinit.service

PR #6841, opened 2025-08-21
Increase OS disk size from 30GB to 60GB

anson627:
**What type of PR is this?**
/kind chores

unblock gb200 build

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder-release.yaml
+++ b/.pipelines/.vsts-vhd-builder-release.yaml
@@ -16,95 +16,95 @@ parameters:
   - name: buildMarinerV2gen1
     displayName: Build MarinerV2 Gen1
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV2gen1
     displayName: Build AzureLinuxV2 Gen1
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV3gen1
     displayName: Build AzureLinuxV3 Gen1
     type: boolean
-    default: false
+    default: true
   - name: buildMarinerV2gen2
     displayName: Build MarinerV2 Gen2
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV2gen2
     displayName: Build AzureLinuxV2 Gen2
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV3gen2
     displayName: Build AzureLinuxV3 Gen2
     type: boolean
-    default: false
+    default: true
   - name: buildMarinerV2gen1fips
     displayName: Build MarinerV2 Gen1 FIPS
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV2gen1fips
     displayName: Build AzureLinuxV2 Gen1 FIPS
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV3gen1fips
     displayName: Build AzureLinuxV3 Gen1 FIPS
     type: boolean
-    default: false
+    default: true
   - name: buildMarinerV2gen2fips
     displayName: Build MarinerV2 Gen2 FIPS
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV2gen2fips
     displayName: Build AzureLinuxV2 Gen2 FIPS
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV3gen2fips
     displayName: Build AzureLinuxV3 Gen2 FIPS
     type: boolean
-    default: false
+    default: true
   - name: buildMarinerV2gen2kata
     displayName: Build MarinerV2 Gen2 Kata
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV2gen2kata
     displayName: Build AzureLinuxV2 Gen2 Kata
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV3gen2kata
     displayName: Build AzureLinuxV3 Gen2 Kata
     type: boolean
-    default: false
+    default: true
   - name: buildMarinerV2ARM64
     displayName: Build MarinerV2 Gen2 - ARM64
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV2ARM64
     displayName: Build AzureLinuxV2 Gen2 - ARM64
     type: boolean
-    default: false
+    default: true
   - name: buildAzureLinuxV3ARM64
     displayName: Build AzureLinuxV3 Gen2 - ARM64
     type: boolean
-    default: false
+    default: true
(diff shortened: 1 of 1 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -818,6 +818,20 @@ addKubeletNodeLabel() {
...
+    local label_value="${LABEL_STRING#*=}"
+
+    if [ "${#label_value}" -gt 63 ]; then
+        echo "Warning: Label value for '${label_key}' exceeds 63 characters (${#label_value} chars). Truncating to last 63 characters."
+        # Take only the last 63 characters of the value
+        label_value="${label_value: -63}"
+        LABEL_STRING="${label_key}=${label_value}"
+        echo "Truncated label: ${LABEL_STRING}"
+    fi
+
```
keith-ms: This is a reasonable addition. However, the interaction between the provisioning scripts and RP appears to overwrite any changes made by the CSE scripts.

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.pub
keith-ms, line 1: For future reference, this file came from the documentation on this page: https://network.nvidia.com/support/mlnx-ofed-public-repository/

 https://www.mellanox.com/downloads/ofed/RPM-GPG-KEY-Mellanox

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -504,6 +504,25 @@ EOF
...
 fi
 
+if grep -q "GB200" <<< "$FEATURE_FLAGS"; then
+  # The GB200 feature flag should only be set for arm64 and Ubuntu 24.04, but validate
+  if [ ${UBUNTU_RELEASE} = "24.04" ] && [ ${CPU_ARCH} = "arm64" ]; then
+    # The open series driver is required for the GB200 platform. Dmesg output
+    # will appear directing the reader away from the proprietary driver. The GPUs
+    # are also not visible in nvidia-smi output with the proprietary drivers
+    apt install -y \
+      libc6-dev \
```
keith-ms: Why are you installing the glibc headers?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -504,6 +504,25 @@ EOF
...
+      mlnx-ofed-kernel-dkms \
+      mlnx-ofed-kernel-utils \
+      mlnx-ofed-basic \
+      rdma-core \
+      ibverbs-utils \
+      ibverbs-providers
+
+    systemctl restart openibd
```
keith-ms: Should this service be enabled by default?

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
```diff
@@ -634,6 +634,16 @@
       "type": "file",
       "source": "parts/linux/cloud-init/artifacts/51-azure-nvidia.cfg",
       "destination": "/home/packer/51-azure-nvidia.cfg"
+    },
+        {
```
keith-ms: Please fix the spacing issue.

keith-ms (commented):
This is going to cause a merge conflict with the other GB200 work.

parts/linux/cloud-init/artifacts/cse_helpers.sh
anson627: make sense, let me drop this change

vhdbuilder/packer/install-dependencies.sh
anson627: it's dependency of rdma-core, but it should only need for building from source, let me drop and verify

vhdbuilder/packer/install-dependencies.sh
anson627: openibd is a Mellanox/NVIDIA service script included with MLNX_OFED, the restart ensures the right drivers are active.

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
anson627: Done

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -520,6 +519,7 @@ if grep -q "GB200" <<< "$FEATURE_FLAGS"; then
       ibverbs-providers
 
     systemctl restart openibd
+    ofed_info -s
```
keith-ms: I recommend checking for an expected value, not just running the command. Barring some check in the wider file, this command could fail and the build would still succeed.

anson627: adding check here will require hard coded driver version: e.g. 24.10-1.1.4.0, which is set in the list file, that mean future version bump will need change 2 places, I can look into how to make driver version a variable in the next iteration

.pipelines/.vsts-vhd-builder-release.yaml
```diff
@@ -16,95 +16,95 @@ parameters:
   - name: buildMarinerV2gen1
     displayName: Build MarinerV2 Gen1
     type: boolean
-    default: true
```
juan-lee: Are these changed on purpose?

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -15,7 +15,7 @@ VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
 PERFORMANCE_DATA_FILE=/opt/azure/vhd-build-performance-data.json
 
 # Hardcode the desired size of the OS disk so we don't accidently rely on extra disk space
-MAX_BLOCK_COUNT=30298176 # 30 GB
+MAX_BLOCK_COUNT=60596352 # 60 GB
```
djsly, line 18: this needs to be handled in a separate PR. there were a lot of discussion around this change in the past, which will cause some SKU to stop working if I recall properly.

djsly (changes requested):

.pipelines/.vsts-vhd-builder-release.yaml
anson627: this is just to disable "default" images to save build time

vhdbuilder/packer/post-install-dependencies.sh
anson627: this is needed since adding drivers requires more disk space, I can split into another PR

.pipelines/.vsts-vhd-builder-release.yaml
juan-lee: I remembered this was for testing, so it's probably good to switch these all to false.

PR #6852, opened 2025-08-25
update github workflow to include windows test components.json

bravebeaver:
**What type of PR is this?**
test: update github workflow actions so that it behaves the same as make file target

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/workflows/validate-components.yml
+++ b/.github/workflows/validate-components.yml
@@ -31,4 +31,6 @@ jobs:
           export PATH="$PATH:$GOPATH/bin"
           cue vet -c ./schemas/windows_settings.cue ./vhdbuilder/packer/windows/windows_settings.json
           cue eval ./schemas/windows_settings.cue ./vhdbuilder/packer/windows/windows_settings.json
+          cue vet -c ./schemas/components.cue ./vhdbuilder/packer/windows/components-test.json
+          cue eval ./schemas/components.cue ./vhdbuilder/packer/windows/components-test.json
 


.github/workflows/validate-components.yml
```diff
@@ -31,4 +31,6 @@ jobs:
           export PATH="$PATH:$GOPATH/bin"
           cue vet -c ./schemas/windows_settings.cue ./vhdbuilder/packer/windows/windows_settings.json
           cue eval ./schemas/windows_settings.cue ./vhdbuilder/packer/windows/windows_settings.json
+          cue vet -c ./schemas/components.cue ./vhdbuilder/packer/windows/components-test.json
+          cue eval ./schemas/components.cue ./vhdbuilder/packer/windows/components-test.json
```
timmy-wright, line 35: We don't need the eval, just the vet I think. Not sure tho.

timmy-wright (approved):

.github/workflows/validate-components.yml
bravebeaver: copilot mentioned: 


cue vet: Validates that the data conforms to the schema (returns success/failure)
cue eval: Evaluates and outputs the merged/unified configuration

just occured to me that the windows pipeline has components.json override step. should we pull the override step in here?

timmy-wright: No - the override is quite a different thing. What's where we get the override from aks-rp.

PR #6855, opened 2025-08-25
delete cluster and retrying when MC resource group is gone

bravebeaver:
**What type of PR is this?**

make e2e a bit smoother..

when garbage collection deleted MC resource group, the cluster is useless and e2e fails.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -238,63 +238,63 @@ func getExistingCluster(ctx context.Context, location, clusterName string) (*arm
 	resourceGroupName := config.ResourceGroupName(location)
 	existingCluster, err := config.Azure.AKS.Get(ctx, resourceGroupName, clusterName, nil)
 	var azErr *azcore.ResponseError
-	if errors.As(err, &azErr) && azErr.StatusCode == 404 {
-		return nil, nil
+	if errors.As(err, &azErr) {
+		if azErr.StatusCode == 404 {
+			return nil, nil
+		}
 	}
 	if err != nil {
-		return nil, fmt.Errorf("failed to get cluster %q: %w", clusterName, err)
+		return nil, fmt.Errorf("failed to get cluster %s: %s", clusterName, err)
 	}
 
 	switch *existingCluster.Properties.ProvisioningState {
 	case "Succeeded":
 		nodeRGExists, err := isExistingResourceGroup(ctx, *existingCluster.Properties.NodeResourceGroup)
+
 		if err != nil {
-			return nil, fmt.Errorf("checking node resource group existence of cluster %s: %w", clusterName, err)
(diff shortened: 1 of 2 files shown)


e2e/cluster.go
```diff
@@ -231,6 +231,19 @@ func getOrCreateCluster(ctx context.Context, cluster *armcontainerservice.Manage
 		return nil, fmt.Errorf("failed to get cluster %q: %w", *cluster.Name, err)
 	}
 	logf(ctx, "cluster %s already exists in rg %s", *cluster.Name, resourceGroupName)
+
+	// ensure MC_rg as well
+	nodeRGExists, err := isExistingResourceGroup(ctx, *existingCluster.Properties.NodeResourceGroup)
+	if err != nil || !nodeRGExists {
+		logf(ctx, "cluster %s does not have a corresponding node resource group", *cluster.Name)
```
timmy-wright: wariningf? Or have "##vso[task.logissue type=warning]" at the start of the log line so that ADO surfaces the warning rather than it being hidden.

As per https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logissue-log-an-error-or-warning

timmy-wright (approved):

e2e/cluster.go
bravebeaver: em.. i think if it can auto-heal, log might be sufficient...

scanning through the code i feel like warningf or even errorf can be placed at other places, such as create cluster. 

i am going to add warning on delete cluster line, if the cluster cant be deleted and manual intervention is required, linking is probably more useful.

e2e/cluster.go
```diff
@@ -231,6 +231,20 @@ func getOrCreateCluster(ctx context.Context, cluster *armcontainerservice.Manage
 		return nil, fmt.Errorf("failed to get cluster %q: %w", *cluster.Name, err)
 	}
 	logf(ctx, "cluster %s already exists in rg %s", *cluster.Name, resourceGroupName)
+
+	// ensure MC_rg as well
+	nodeRGExists, err := isExistingResourceGroup(ctx, *existingCluster.Properties.NodeResourceGroup)
```
jepio: line 250 has the exact same check. Can you see if it's enough to add the deleteCluster call there?

bravebeaver: indeed.. on that note.. i was wondering. shall we move the MC rg check a bit higher, outside of the cluster provisionign check. so the method `getOrCreateCluster` is to make proper decision depend on the cluster state..

timmy-wright (approved):

PR #6861, opened 2025-08-25
Freezing kata mariner vhd for kata-cc.

Bickor:
**What type of PR is this?**
/kind deprecation

Freezing the Kata AzL2 VHD since the build pipeline is no longer working since AzL2 is has begun deprecation and is being actively removed from ACRs. Additionally, can't migrate to AzL3 because components required to run Kata-CC have been getting removed in Kata-CC, and we need to support the feature at least until ~March 2026. This PR freezes the VHDs in a previous already working build.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/pkg/agent/datamodel/sig_config.go
+++ b/pkg/agent/datamodel/sig_config.go
@@ -439,6 +439,10 @@ const (
 	// DO NOT MODIFY: used for freezing MarinerV2KataGen2TL.
 	FrozenCBLMarinerV2KataGen2TLSIGImageVersion = "2022.12.15"
 
+	// DO NOT MODIFY: Used for freezing Kata VHD for kata-cc runtime, until the image is deprecated.
+	FrozenCBLMarinerV2KataGen2SIGImageVersion string = "202509.04.0"
+	FrozenAzureLinuxV2KataGen2SIGImageVersion string = "202509.04.0"
+
 	// We do not use AKS Windows image versions in AgentBaker. These fake values are only used for unit tests.
 	Windows2019SIGImageVersion string = "17763.2019.221114"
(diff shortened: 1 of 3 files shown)


pkg/agent/datamodel/sig_config.go
```diff
@@ -439,6 +439,10 @@ const (
 	// DO NOT MODIFY: used for freezing MarinerV2KataGen2TL.
 	FrozenCBLMarinerV2KataGen2TLSIGImageVersion = "2022.12.15"
 
+	// DO NOT MODIFY: Used for freezing Kata VHD for kata-cc runtime, until the image is deprecated.
+	FrozenCBLMarinerV2KataGen2SIGImageVersion string = "202508.06.0"
+	FrozenAzureLinuxV2KataGen2SIGImageVersion string = "202508.06.0"
```
Bickor: I think 202508.06 should be the right date since I'm picking the release from this [PR](https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/pullrequest/13230704?_a=files) but please correct me if a different date should be used.

cameronmeissner: @Bickor I'd recommend using `202507.21.0` instead to ensure you're not inheriting any bugs introduced in a few of the latest node images

cameronmeissner: do we also need to freeze kataTL?

Bickor: Frozen to version 202509.04.0 as per teams thread.

TL doesn't need to be frozen because kata doesn't currently support TL and is never built or used.

PR #6871, opened 2025-08-26
Add Doca/Mofed driver to GB200 image

anson627:
**What type of PR is this?**
/kind feature

Add Doca/Mofed driver to GB200 image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -504,6 +504,25 @@ EOF
...
+    apt install -y \
+      mlnx-ofed-kernel-dkms \
+      mlnx-ofed-kernel-utils \
+      mlnx-ofed-basic \
+      rdma-core \
+      ibverbs-utils \
+      ibverbs-providers
+
+    systemctl restart openibd
+    ofed_info -s
```
keith-ms: You should match the output on a regex for what's expected, or at least check the exit code.

keith-ms (approved):
One non-blocking suggestion, otherwise looks good to me.

vhdbuilder/packer/install-dependencies.sh
anson627: will add regex check in the next iteration

junjiezhang1997 (approved):
LGTM

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
```diff
@@ -0,0 +1,9 @@
+#
+# Mellanox Technologies Ltd. public repository configuration file.
+# For more information, refer to http://linux.mellanox.com
+#
+# For future reference:
+# https://network.nvidia.com/support/mlnx-ofed-public-repository/
+# https://www.mellanox.com/downloads/ofed/RPM-GPG-KEY-Mellanox
+# [mlnx_ofed_24.10-1.1.4.0_base]
```
djsly, line 8: how often will this version change ? 1.1.4.0 ?

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
```diff
@@ -0,0 +1,9 @@
...
+deb [signed-by=/etc/apt/keyrings/mellanox_mlnx_ofed.pub] http://linux.mellanox.com/public/repo/mlnx_ofed/24.10-1.1.4.0/ubuntu24.04/arm64 ./
```
djsly, line 9: We don't expose / request user to whitelist this URL `linux.mellanox.com` before we start making this public we should ensure we augment this public list of outbound URL.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -504,6 +504,25 @@ EOF
 
 fi
 
+if grep -q "GB200" <<< "$FEATURE_FLAGS"; then
```
djsly, line 507: is there a reason why this logic isn't baked in the new GB200 VHD only ?  only this "FEATURE_FLAGS" is used for that ?

vhdbuilder/packer/install-dependencies.sh
djsly: is the idea that this gets stored to the cluster-provisionning logs ?  we shoudl redirect the output the file is we want to store the output.

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
anson627: until MAI request new feature that requires another driver upgrade

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
anson627: my understanding is access linux.mellanox.com is only required when packer install the driver into the VHD, and no need to access this URL at bootstrap time or runtime

vhdbuilder/packer/install-dependencies.sh
anson627: the intention here is just to help debug driver installation issue during VHD builds, I can add check or redirect later if needed

vhdbuilder/packer/install-dependencies.sh
anson627: GB200 is unique enough to ensure this logic is baked in GB200 VHD only, as this is the only job has GB200 FEATURE_FLAGS: https://github.com/Azure/AgentBaker/blob/110f8a73eb9c3e4949af880bd391598e3c0c7c4c/.pipelines/.vsts-vhd-builder-release.yaml#L1338C67-L1338C72

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
djsly: this should probably move to components.json and have automatic driver version discovery in the future.  add TODO to ensure we move the fixed version to a global location, keepings hardcoded version string in random files isn't quite efficient

vhdbuilder/packer/install-dependencies.sh
djsly: the string FEATURE_FLAG makes it non obvious that this is specific to VHD building.  we should look at renaming this to VHD_xxxxx

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -504,6 +504,25 @@ EOF
...
+  # The GB200 feature flag should only be set for arm64 and Ubuntu 24.04, but validate
+  if [ ${UBUNTU_RELEASE} = "24.04" ] && [ ${CPU_ARCH} = "arm64" ]; then
+    # The open series driver is required for the GB200 platform. Dmesg output
+    # will appear directing the reader away from the proprietary driver. The GPUs
+    # are also not visible in nvidia-smi output with the proprietary drivers
+    apt install -y \
```
djsly, line 513: we don't want to fix any versions here ? the repo only contains single versions ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -504,6 +504,25 @@ EOF
...
+      mlnx-ofed-kernel-dkms \
+      mlnx-ofed-kernel-utils \
+      mlnx-ofed-basic \
+      rdma-core \
+      ibverbs-utils \
+      ibverbs-providers
```
djsly, line 519: we need content-type test to ensure that the content installed during build is well present,

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
djsly, line 9: is it safe to keep the mellanox repo present post build ? if we simply want to have the driver installed, I fear this repo /key might not work for ever so I would prefer that we remove them from the resulting VHD once the right packages are installed.

djsly (changes requested):

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
anson627: ack

vhdbuilder/packer/install-dependencies.sh
anson627: we want the latest patch version based on what are in mlnx_ofed repo

vhdbuilder/packer/install-dependencies.sh
anson627: ack

vhdbuilder/packer/install-dependencies.sh
djsly: could we export the package versions/driver versions inside the vhd release please.  Customer always complains when they get a new VHD and we bump the driver versions and they couldn't deduct it from our release notes.

parts/linux/cloud-init/artifacts/ubuntu/mellanox_mlnx_ofed.list
djsly: we should remove the repo in the cleanup section I would say.

PR #6872, opened 2025-08-26
fix: truncate long status in cloud init status check to retain json structure if message is too long

UtheMan:
In some cases the error returned by cloud-init status is too long to be properly formatted in logs, so we have to truncate it


parts/linux/cloud-init/artifacts/cloud-init-status-check.sh
```diff
@@ -31,6 +31,17 @@ handleCloudInitStatus() {
...
     
+    # Truncate cloudInitLongStatus if needed to keep Message field under 3072 characters
+    local maxMessageLength=3072
+    local messageOverhead=130  # JSON structure + cloudInitMessage
+    local maxLongStatusLength=$((maxMessageLength - messageOverhead))
+    
+    if [ ${#cloudInitLongStatus} -gt $maxLongStatusLength ]; then
+        local truncationSuffix="...TRUNCATED"
+        local allowedLength=$((maxLongStatusLength - ${#truncationSuffix}))
+        cloudInitLongStatus="${cloudInitLongStatus:0:$allowedLength}${truncationSuffix}"
```
djsly: shouldnt you substrct the lenght of `truncationSuffix` ?

UtheMan: Changed the approach - instead of truncating i will just send the actual errors to events directory. Truncating can result in malformed jsons anyway and I couldn't think of an easy way to handle it nicely (and we don't really need the full output there anyway)

PR #6879, opened 2025-08-27
fix(cse): remove non-host arguments for `oras repo ls`

qweeah:
**What type of PR is this?**
/kind bug

This PR removes an unnecessary argument from the `oras repo ls` command in the CSE script. 

At the moment, `$test_image` has not been assigned a value and remains an empty string. Once unintentionally set, it will be treated as a namespace (see [oras examples](https://oras.land/docs/commands/oras_repo_ls#examples)) and will list repositories only within that specified namespace.

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


qweeah:
Closing since the PR need to be created from a branch in this repository, not from a forked repository.

PR #6885, opened 2025-08-28
fix(cse): remove non-host arguments for `oras repo ls`

qweeah:
**What type of PR is this?**
/kind bug

This PR removes an unset argument from the `oras repo ls` command in the CSE script. 

To verify login success, only host argument is needed. At the moment, `$test_image` has not been assigned a value and remains an empty string. Once unintentionally set, the extra `$test_image` will be treated as a namespace (see [oras examples](https://oras.land/docs/commands/oras_repo_ls#examples)) and will list repositories only within that specified namespace, leading to false negative.

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


djsly (approved):

qweeah:
I noticed that [Agentbaker Windows E2E](https://github.com/Azure/AgentBaker/pull/6885/checks?check_run_id=49070727450) is failing because of some network file taken by another process. It should not be caused by change in this PR.

Anyone had any idea on where to start to trouble shoot such issue?

djsly:
@qweeah you must sign your commits. 

Merging is blocked
Commits must have verified signatures.

djsly (approved):

PR #6894, opened 2025-08-29
Fix LocalDNS Upstream assignment in corefile for localdns

SriHarsha001:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -129,29 +132,44 @@ replace_azurednsip_in_corefile() {
...
     fi
-
     echo "Found upstream VNET DNS servers: ${UPSTREAM_VNET_DNS_SERVERS}"
 
     # Based on customer input, corefile was generated in pkg/agent/baker.go.
     # Replace 168.63.129.16 with VNET DNS ServerIPs only if VNET DNS ServerIPs is not equal to 168.63.129.16
     # and also not equal to the localdns node listener IP to avoid creating a circular dependency.
     # Corefile will have 168.63.129.16 when user input has VnetDNS value for forwarddestination.
     # Note - For root domain under VnetDNSOverrides, all DNS traffic should be forwarded to VnetDNS.
+    cp "${LOCALDNS_CORE_FILE}" "${UPDATED_LOCALDNS_CORE_FILE}" || {
```
yewmsft, line 147: as Patrick suggested, you can drop this copy and do sed **without -i** directly and sending the result to ${UPDATED_LOCALDNS_CORE_FILE}

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -129,29 +132,44 @@ replace_azurednsip_in_corefile() {
...
             echo "Replacing AzureDNSIP in corefile failed."
             return 1 
         }
-        echo "Successfully replaced Azure DNS IP with upstream VNET DNS servers in corefile"
+        echo "Successfully updated ${UPDATED_LOCALDNS_CORE_FILE}"
     else
         echo "Skipping DNS IP replacement. Upstream VNET DNS servers (${UPSTREAM_VNET_DNS_SERVERS}) match either Azure DNS IP (${AZURE_DNS_IP}) or localdns node listener IP (${LOCALDNS_NODE_LISTENER_IP})"
     fi
+
+    chmod 0644 "${UPDATED_LOCALDNS_CORE_FILE}" || {
```
yewmsft: move this after line 172

spec/parts/linux/cloud-init/artifacts/localdns_spec.sh
```diff
@@ -39,6 +41,8 @@ EOF
 cat <<EOF > "$RESOLV_CONF"
 nameserver 10.0.0.1
 nameserver 10.0.0.2
+nameserver 10.0.0.3
+nameserver 10.0.0.4
```
yewmsft, line 45: Just a FYI, this test case does not cover the case Patrick mentioned previously. 
Patrick meant (even though it will only happens in pod resolv.conf, it will not happen here) the 4th+ nameserver will be commented out in the pod resolv.conf.

But you don't need to change this, because that case will never happen on node.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: I am executing sed only when the if condition is met (ie when it is required to replace AzureIP and avoid doing sed always on all the nodes). So for cases where I dont need to sed, updated core file will not be created and will not be available if I dont do a prior copy. And I need to use this updated corefile later to start localdns.

spec/parts/linux/cloud-init/artifacts/localdns_spec.sh
SriHarsha001: I understand. I don't think there is anything wrong here in adding additional nameservers in the test file.

yewmsft: It didn't test the /^nameserver/ change though. But as long as the comment does not happen, it is fine

Devinwong (approved):

PR #6895, opened 2025-08-29
feat: add blacklist nouveau drivers, udev char rules, some comments

abenn135:
**What type of PR is this?**

/kind feature

Brings GB200 VHD installed driver state into parity with what normal GPU nodes get at join time re nvidia configuration.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -498,6 +498,16 @@ fi
 if grep -q "GB200" <<< "$FEATURE_FLAGS"; then
   # The GB200 feature flag should only be set for arm64 and Ubuntu 24.04, but validate
   if [ ${UBUNTU_RELEASE} = "24.04" ] && [ ${CPU_ARCH} = "arm64" ]; then
+    # Need to replicate all functionality from github.com/azure/aks-gpu/install.sh
```
djsly: this is making me cry :(  could you explain the reason behind ?

djsly: in the comments ?   I'm really looking to the day we move the logic back into AgentBaker..

djsly (approved):

vhdbuilder/packer/install-dependencies.sh
abenn135: Done.

PR #6899, opened 2025-08-31
feat: Add Windows Cilium Networking (WCN) feature

abeltrano:
/kind feature

Adds the Windows Cilium Networking (WCN) platform feature, which is to be used in various AKS subscriptions.

Fixes #

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Release note**:

```
Add Windows Cilium Networking (WCN) feature
```


pkg/agent/baker_test.go
```diff
@@ -2686,6 +2686,27 @@ var _ = Describe("Assert generated customData and cseCmd for Windows", func() {
 				config.KubeletConfig["--image-credential-provider-config"] = "c:\\var\\lib\\kubelet\\credential-provider-config.yaml"
 				config.KubeletConfig["--image-credential-provider-bin-dir"] = "c:\\var\\lib\\kubelet\\credential-provider"
 			}),
+		Entry("AKSWindows23H2Gen2 with NextGenNetworking", "AKSWindows23H2Gen2+NextGenNetworking", "1.29.0",
```
abeltrano, line 2689: All other Windows cases use `AKSWindows2019` and didn't seem to set the OS/SKU at all, however, since this feature is currently only available on 23H2, I used that in the entry names.

Please let me know if this should conform to the other entries with the `AKSWindows2019` naming.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -263,6 +270,7 @@ if (-not (Test-Path "C:\AzureData\windows\azurecnifunc.ps1")) {
 . c:\AzureData\windows\kubernetesfunc.ps1
 . c:\AzureData\windows\nvidiagpudriverfunc.ps1
 . c:\AzureData\windows\securetlsbootstrapfunc.ps1
+. c:\AzureData\windows\windowsciliumnetworkingfunc.ps1
```
timmy-wright: There is a chance of mismatched versions of this CSE and the windows scripts that means your ps1 won't exist on older VHDs. Probalby need code to protect against that case.

parts/windows/kuberneteswindowssetup.ps1
```diff
@@ -431,6 +439,10 @@ function BasePrep {
     # Turn off Firewall to enable pods to talk to service endpoints. (Kubelet should eventually do this)
     netsh advfirewall set allprofiles state off
 
+    if ($global:EnableWindowsCilium) {
+        Enable-WindowsCiliumNetworking
```
timmy-wright: Again, there's a chance that this function (enable-windowsciliumnetworking) won't exist. Somewhere else in this file there's a check that a function exists before calling it. You might need to copy that.

pkg/agent/baker_test.go
timmy-wright: I think they all use Win2019 by accident rather than by intention. And Win2019 is EOL so we shouldn't be using it anyway.

abeltrano: Thanks, will resolve then.

parts/windows/kuberneteswindowssetup.ps1
abeltrano: I saw that for something else and wasn't sure it was necessary, because it's done for some and not others. Will add it.

vhdbuilder/packer/windows/configure-windows-vhd.ps1
```diff
@@ -1006,6 +1037,7 @@ try
             Get-ToolsToVHD
             Get-PrivatePackagesToCacheOnVHD
             Log-ReofferUpdate
+            Install-WindowsCilium
```
timmy-wright: Can you put this before the Log-ReofferUpdate? We do that just before each reboot as it tells us ... stuff.

abeltrano: Sure.

parts/windows/kuberneteswindowssetup.ps1
abeltrano: Now checking that the script exists before dot-sourcing it.

parts/windows/kuberneteswindowssetup.ps1
abeltrano: Now validating the command is available before invoking it.

vhdbuilder/packer/windows/configure-windows-vhd.ps1
abeltrano: Done.

timmy-wright (approved):

timmy-wright (approved):

timmy-wright:
Force merged:
* Go unit tests are failing in master
* windows-e2e tests won't work with this type of change
* WS2019 VHD test failed for an unrelated reason

PR #6900, opened 2025-08-31
chore: fix auto windows comment when cpu-arch is in components.json

timmy-wright:
**What type of PR is this?**

/kind chore

Fixes auto-windows comment on PR to contain cpu arch.

image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/scripts/windows/generate_cached_stuff_list.ps1
+++ b/vhdbuilder/scripts/windows/generate_cached_stuff_list.ps1
@@ -49,6 +49,8 @@ if (![string]::IsNullOrEmpty($componentsJsonFileParam))
 
 . "$HelpersFile"
 
+$CPU_ARCH="cpu-arch"
+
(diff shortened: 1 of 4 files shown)


vhdbuilder/scripts/windows/generate_cached_stuff_list.ps1
```diff
@@ -49,6 +49,8 @@ if (![string]::IsNullOrEmpty($componentsJsonFileParam))
 
 . "$HelpersFile"
 
+$CPU_ARCH="cpu-arch"
```
r2k1, line 52: That's a stange variable

r2k1 (approved):

vhdbuilder/scripts/windows/generate_cached_stuff_list.ps1
timmy-wright: variable or value? The value is just so something sensible shows up on the github comment - instead of an empty string.

PR #6903, opened 2025-09-01
Flatcar custom cloud support

jepio:
**What type of PR is this?**
/kind feature

Add `init-custom-cloud-xxx.sh` scripts for Flatcar for national and sovereign clouds.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/init-aks-custom-cloud-flatcar.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+set -x
+mkdir -p /root/AzureCACertificates
+# http://168.63.129.16 is a constant for the host's wireserver endpoint
+certs=$(curl "http://168.63.129.16/machine?comp=acmspackage&type=cacertificates&ext=json")
```
cameronmeissner: this seems like a lot of code duplication, could we possibly detect within the original scripts that we're running on a flatcar distro and pivot accordingly?

jepio: sure - i could do that

cameronmeissner (approved):
I think this looks good, though unfortunately we don't have a great way to test this - at some point we should look at re-writing these scripts to actually be unit-testable and actually add some tests using shellspec

PR #6908, opened 2025-09-02
fix: eval http proxy vars in baseprep

lilypan26:
**What type of PR is this?**

Fixes issue where outbound curl during kubelet/kubectl install is not going through proxy due to missing eval. Change in ordering was introduced here: https://github.com/Azure/AgentBaker/pull/6601. Prior, eval happens before kubelet/kubectl install. Also ensures `resolve_packages_source_url` uses proxy if configured.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -62,6 +62,10 @@ function basePrep {
         sudo systemctl restart systemd-timesyncd
     fi
 
+    if [ -n "${PROXY_VARS}" ]; then
```
djsly, line 68: could you add some comments explaining why we do this :D

djsly (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
lilypan26: added

djsly (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -908,7 +908,7 @@ resolve_packages_source_url() {
     PACKAGE_DOWNLOAD_BASE_URL="packages.aks.azure.com"
     for i in $(seq 1 $retries); do
       # Confirm that we can establish connectivity to packages.aks.azure.com before node provisioning starts
-      response_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 5 --noproxy "*" https://packages.aks.azure.com/acs-mirror/healthz)
+      response_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 5 https://packages.aks.azure.com/acs-mirror/healthz)
```
UtheMan, line 911: This is called from `basePrep` right and HTTP Proxy setup is being done later on, will it work as expected in those environments?

lilypan26: this happens after the eval which is all that's really needed here to make sure the curl goes through the proxy if its configured

UtheMan (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
r2k1: There is node restart may happen between basePrep and nodePrep in the future (with losing all currently set ENV variables).
(Currently no tests are enabled for such use case).

Do we need to eval it in nodePrep phase as well?

PR #6921, opened 2025-09-03
Refactor mirror proxy version to use components.json

ganeshkumarashok:
- Add acr-mirror component to components.json with version 0.2.13
- Update installAndConfigureArtifactStreaming function to read version from components.json
- Support both Ubuntu (deb) and AzureLinux/Mariner (rpm) packages
- Add fallback to hardcoded version if components.json lookup fails
- Centralizes version management for easier maintenance and updates

**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -434,7 +434,26 @@ installAndConfigureArtifactStreaming() {
   # arguments: package name, package extension
   PACKAGE_NAME=$1
   PACKAGE_EXTENSION=$2
-  MIRROR_PROXY_VERSION='0.2.12'
+  
+  # Extract acr-mirror version from components.json
+  MIRROR_PROXY_VERSION=""
+  acr_mirror_package=$(jq '.Packages[] | select(.name == "acr-mirror")' $COMPONENTS_FILEPATH)
+  if [ -n "$acr_mirror_package" ]; then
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -434,7 +434,26 @@ installAndConfigureArtifactStreaming() {
...
+    if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+      MIRROR_PROXY_VERSION=$(echo "$acr_mirror_package" | jq -r '.downloadURIs.ubuntu.default.versionsV2[0].latestVersion')
+    elif isMarinerOrAzureLinux "$OS"; then
+      MIRROR_PROXY_VERSION=$(echo "$acr_mirror_package" | jq -r '.downloadURIs.azurelinux.current.versionsV2[0].latestVersion')
+    fi
+  fi
+  
+  # Error if version not found in components.json
+  if [ -z "$MIRROR_PROXY_VERSION" ] || [ "$MIRROR_PROXY_VERSION" = "null" ]; then
+    echo "Error: Could not find acr-mirror version in components.json for OS: $OS"
```
cameronmeissner, line 451: seems to be a mismatch with the PR description, I guess we're not falling back to a hard-coded version if components.json lookup fails? (I think that's fine, just wanted to double check - unless I'm missing something)

PR #6925, opened 2025-09-04
fix: remove WinOverlay feature gate from Kubeproxy configuration

gaopenghigh:
**What type of PR is this?**
/kind bug

### Symptoms
- Windows pods on v1.34 could not resolve in-cluster DNS names or reach ClusterIP services.
- kube-proxy (Windows) terminated on startup on affected nodes.
- Linux pods and nodes were unaffected.

### Executive summary
On Kubernetes v1.34, the Windows kube-proxy feature gate `WinOverlay` graduated to GA and was locked to its default (true). If kube-proxy is configured with `WinOverlay=false`, kube-proxy exits at startup because overrides are ignored once the gate is locked. The behavior was introduced between v1.34.0-beta.0 and the release candidate/GA builds.

### Upstream change that introduced the behavior
- PR: [kubernetes/kubernetes#133042 — “WinOverlay feature gate to GA”](https://github.com/kubernetes/kubernetes/pull/133042)
- Merged: 2025-07-22 (merge commit 7bf8066a58a24720a30656ac8f08b8c68cdd0a8d)
- Code change: adds a v1.34 entry for WinOverlay with `PreRelease: GA` and `LockToDefault: true` in `pkg/features/kube_features.go`.
  - Diff reference: https://github.com/kubernetes/kubernetes/pull/133042/files#diff-71e3b98f9a6bbf5b8421e26a7ba0c079f397cd8d49abacdad943c66a4f44f03d
  - Effective meaning: once GA + `LockToDefault: true`, user-supplied `--feature-gates=WinOverlay=false` (or equivalent config) is not permitted and causes kube-proxy to terminate.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/pkg/agent/datamodel/types.go
+++ b/pkg/agent/datamodel/types.go
@@ -1185,9 +1185,11 @@ func (p *Properties) GetKubeProxyFeatureGatesWindowsArguments() string {
 		featureGates["IPv6DualStack"] = true
 	}
 	if p.FeatureFlags.IsFeatureEnabled(EnableWinDSR) {
-		// WinOverlay must be set to false.
 		featureGates["WinDSR"] = true
-		featureGates["WinOverlay"] = false
+		// WinOverlay feature gate is GA and locked to true in Kubernetes 1.34.0 and later
+		if p.OrchestratorProfile == nil || !IsKubernetesVersionGe(p.OrchestratorProfile.OrchestratorVersion, "1.34.0") {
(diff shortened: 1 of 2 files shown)


timmy-wright:
Happy for this to be merged - the failing tests are known ones that I need to find time to fix. But commits aren't signed - once that's done I can overide policies and merge. Or someone else can.

PR #6929, opened 2025-09-04
test: Make test ServicePrincipal secret slightly more unique

jepio:
**What type of PR is this?**
/kind flake
/kind failing-test

The previous value of `msi` encodes to bXnp and we had at least one flakey test run where that string was found inside the value of /etc/kubernetes/certs/ca.crt logged to /var/log:

   validation.go:169: expected file /var/log/azure/cluster-provision.log to not have contents "bXNp", but it does

Switch to `**msi**` which encodes to something longer and is much less likely to occur elsewhere.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -436,7 +436,7 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 				},
 				ServicePrincipalProfile: &datamodel.ServicePrincipalProfile{
 					ClientID: "msi",
-					Secret:   "msi",
+					Secret:   "**msi**",
 				},
 				CertificateProfile:  &datamodel.CertificateProfile{},
 				HostedMasterProfile: &datamodel.HostedMasterProfile{},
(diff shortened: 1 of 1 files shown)


r2k1 (approved):

Devinwong (approved):

jepio:
> I wonder if we should focus on fixing the validation logic instead - the value of `msi` is what will actually be there in production when a managed identity is being used instead of a service principal

Are you saying that when we enable a managed identity on a nodepool, RP will use the dummy string "msi" in that field?

cameronmeissner:
> > I wonder if we should focus on fixing the validation logic instead - the value of `msi` is what will actually be there in production when a managed identity is being used instead of a service principal
> 
> Are you saying that when we enable a managed identity on a nodepool, RP will use the dummy string "msi" in that field?

yes that's what I'm saying

PR #6930, opened 2025-09-04
fix dir permissions in /opt

jepio:
**What type of PR is this?**
/kind cleanup

There are some directories in /opt that currently end up non-root owned in our node images:
- /opt/certs
- /opt/scripts
- /opt/azure/containers

This ownership is not needed right now - clean this up and also remove the sticky bit from /opt/certs which is also a historical leftover.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/list-images.sh
+++ b/vhdbuilder/packer/list-images.sh
@@ -29,3 +29,4 @@ pushd /home/packer
 popd
 
 chmod a+r $IMAGE_BOM_PATH
+chmod a+rx /opt/azure/ /opt/azure/containers/


cameronmeissner (approved):
let's make sure all e2e's are green before merge

jepio:
> I missed something in Azure Linux 2, need to review tomorrow.

Fixed the permissions that were blocking packer from copying out `/opt/azure/containers/image-bom.json`.

invidian (approved):

PR #6935, opened 2025-09-04
fix: add arm64 endpoint

zachary-bailey:
**What type of PR is this?**

/kind bug

This PR adds an endpoint for the arm64 sources list for snapshotting with VHD build timestamps.

Hotfix

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/vhdbuilder/packer/pre-install-dependencies.sh
+++ b/vhdbuilder/packer/pre-install-dependencies.sh
@@ -107,7 +107,10 @@ else
   fi
 
   if [ -n "${VHD_BUILD_TIMESTAMP}" ] && [ ${UBUNTU_RELEASE//./} -eq 2204 ]; then
-    if [ "${CPU_ARCH}" = "amd64" ]; then
+    echo "Using packages from desired snapshot..."
+    if [ "${CPU_ARCH}" = "arm64" ]; then
+      sed -i -E "s|http://[^ ]+/ubuntu-ports/?|http://snapshot.ubuntu.com/ubuntu/${VHD_BUILD_TIMESTAMP} |" /etc/apt/sources.list
+    else
       sed -i "s#http://azure.archive.ubuntu.com/ubuntu/#https://snapshot.ubuntu.com/ubuntu/${VHD_BUILD_TIMESTAMP}#g" /etc/apt/sources.list
     fi
   fi


vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -106,8 +106,12 @@ else
     set -x
   fi
 
-  if [ -n "${VHD_BUILD_TIMESTAMP}" ] && [ "${OS_VERSION}" = "22.04" ]; then
-    sed -i "s#http://azure.archive.ubuntu.com/ubuntu/#https://snapshot.ubuntu.com/ubuntu/${VHD_BUILD_TIMESTAMP}#g" /etc/apt/sources.list
+  if [ -n "${VHD_BUILD_TIMESTAMP}" ] && [ ${UBUNTU_RELEASE//./} -le 2204 ]; then
```
djsly: I thought we were doing 22.04 only , why `-le` ?

zachary-bailey: @djsly after speaking with Ye we decided to include 2004

djsly: but why ? do we know if this kernel is affected ?

djsly: canonical confirmed only 5.15 is affected and it should be affecting x86 only

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -106,8 +106,12 @@ else
...
+    if "${CPU_ARCH}" == "arm64"; then
```
anujmaheshwari1: if by itself expects a command to run, you need to wrap this in [[ ]]

if [[ "${CPU_ARCH}" == "arm64" ]]; then
    sed -i "s#http://ports.ubuntu.com/ubuntu-ports/#https://snapshot.ubuntu.com/ubuntu/${VHD_BUILD_TIMESTAMP}#g" /etc/apt/sources.list
else
    sed -i "s#http://azure.archive.ubuntu.com/ubuntu/#https://snapshot.ubuntu.com/ubuntu/${VHD_BUILD_TIMESTAMP}#g" /etc/apt/sources.list
fi

vhdbuilder/packer/pre-install-dependencies.sh
zachary-bailey: fixed @djsly

anujmaheshwari1 (approved):

PR #6938, opened 2025-09-04
install pinned kernel 5.15.0-1092 for Ubuntu 22.04

Devinwong:
**What type of PR is this?**
/kind bug

Canonical confirmed that there is regression in the latest kernel pkgs for Ubuntu 22.04. Therefore we install the pinned kernel version 5.15.0-1092 before they fix the issues.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/pre-install-dependencies.sh
+++ b/vhdbuilder/packer/pre-install-dependencies.sh
@@ -134,29 +134,69 @@ capture_benchmark "${SCRIPT_NAME}_enable_cgroupv2_for_azurelinux"
 
 # shellcheck disable=SC3010
 if [[ ${UBUNTU_RELEASE//./} -ge 2204 && "${ENABLE_FIPS,,}" != "true" ]] && ! grep -q "cvm" <<< "$FEATURE_FLAGS"; then
-  LTS_KERNEL="linux-image-azure-lts-${UBUNTU_RELEASE}"
-  LTS_TOOLS="linux-tools-azure-lts-${UBUNTU_RELEASE}"
-  LTS_CLOUD_TOOLS="linux-cloud-tools-azure-lts-${UBUNTU_RELEASE}"
-  LTS_HEADERS="linux-headers-azure-lts-${UBUNTU_RELEASE}"
-  LTS_MODULES="linux-modules-extra-azure-lts-${UBUNTU_RELEASE}"
-
-  echo "Logging the currently running kernel: $(uname -r)"
-  echo "Before purging kernel, here is a list of kernels/headers installed:"; dpkg -l 'linux-*azure*'
-
-  if apt-cache show "$LTS_KERNEL" &>/dev/null; then
-      echo "LTS kernel is available for ${UBUNTU_RELEASE}, proceeding with purging current kernel and installing LTS kernel..."
-
+  
+  install_pinned_kernel() {
+    TARGET_KERNEL_VER="5.15.0-1092"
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -134,16 +134,47 @@ capture_benchmark "${SCRIPT_NAME}_enable_cgroupv2_for_azurelinux"
 
 # shellcheck disable=SC3010
 if [[ ${UBUNTU_RELEASE//./} -ge 2204 && "${ENABLE_FIPS,,}" != "true" ]] && ! grep -q "cvm" <<< "$FEATURE_FLAGS"; then
-  LTS_KERNEL="linux-image-azure-lts-${UBUNTU_RELEASE}"
-  LTS_TOOLS="linux-tools-azure-lts-${UBUNTU_RELEASE}"
-  LTS_CLOUD_TOOLS="linux-cloud-tools-azure-lts-${UBUNTU_RELEASE}"
-  LTS_HEADERS="linux-headers-azure-lts-${UBUNTU_RELEASE}"
-  LTS_MODULES="linux-modules-extra-azure-lts-${UBUNTU_RELEASE}"
```
r2k1, line 141: What about something like this?
```
KERNEL="linux-image-azure-lts-${UBUNTU_RELEASE}"
TOOLS="linux-tools-azure-lts-${UBUNTU_RELEASE}"
CLOUD_TOOLS="linux-cloud-tools-azure-lts-${UBUNTU_RELEASE}"
HEADERS="linux-headers-azure-lts-${UBUNTU_RELEASE}"
MODULES="linux-modules-extra-azure-lts-${UBUNTU_RELEASE}"
if [[ ${UBUNTU_RELEASE//./} -eq 2204 ]]; # AND ARCH IS X64
  KERNEL="linux-image-5.15.0-1092-azure"
  TOOLS="changeme"
  CLOUD_TOOLS="changeme"
  HEADERS="changeme"
  MODULES="changeme"
fi
```
Is there anything else different?

Devinwong: The echo messages are slightly different because the old ones mention `LTS`. Other than that, only the variables  you listed are different.
What I thought was once the kernel is fixed from upstream, we can simply remove the new function `install_pinned_kernel` so I intentionally didn't change anything in the original flow. But either works.

r2k1: https://github.com/Azure/AgentBaker/pull/6939

Something like this seems more straightforward to me

vhdbuilder/packer/pre-install-dependencies.sh
```diff
@@ -134,16 +134,47 @@ capture_benchmark "${SCRIPT_NAME}_enable_cgroupv2_for_azurelinux"
...
+  
+  install_pinned_kernel() {
```
cameronmeissner, line 138: can we define these funcs somewhere other than within the if-block starting on L136? maybe at the bottom/top of the file?

vhdbuilder/packer/pre-install-dependencies.sh
cameronmeissner: I also prefer this approach if it works as expected

Devinwong:
@r2k1 , we can go with your PR #6939

PR #6946, opened 2025-09-05
Simplify E2E reporting

r2k1:
**What type of PR is this?**

/kind test

Using the -v flag streams log output as tests are executed. However, since our tests run concurrently, the logs from different tests are interleaved, making them difficult to read. Without the -v flag, log messages are grouped by test (and only logs from failed tests are shown). The downside is that logs are not streamed in real time—you’ll need to wait until each test completes before seeing its output.


.pipelines/scripts/e2e_run.sh
```diff
@@ -78,4 +78,4 @@ GOBIN=`pwd`/bin/ go install github.com/jstemmer/go-junit-report/v2@latest
 go build -mod=readonly ./...
 
 # Run the tests! Yey!
-go test -v -parallel 100 -timeout 90m 2>&1 | ./bin/go-junit-report -iocopy -set-exit-code -out "${BUILD_SRC_DIR}/e2e/report.xml"
+go test -parallel 100 -timeout 90m 2>&1 | ./bin/go-junit-report -iocopy -set-exit-code -out "${BUILD_SRC_DIR}/e2e/report.xml"
```
jepio: This would be a nice change, but it breaks the "Tests" tab in ADO - we're missing the test output from failed tests and the passing test cases are needed in the report to track test history across runs (I find this valuable). `go-junit-report` fails to parse the test output without the `=== NAME  Test_<name>` headers, that leads to the missing output from the failing test cases (currently in the "stack trace" box).

Can you implement a custom filter that does the de-multiplexing in the output:
```
go test -v  ...  | ./bin/go-test-demux | ./bin/go-junit-report ...
```

jepio (dismissed):
we should simplify E2E reporting while keeping the junit report complete (failing test output + passing test cases).

PR #6954, opened 2025-09-05
Add localdns protos for scriptless

SriHarsha001:
**What type of PR is this?**
/kind feature

This PR adds localdns protos in scripless framework. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


aks-node-controller/proto/aksnodeconfig/v1/cluster_config.proto
```diff
@@ -51,6 +51,9 @@ message ClusterNetworkConfig {
 
   // Route table name
   string route_table = 5;
+
+  // CoreDNS serviceIP of the cluster
+  string core_d_n_s_service_i_p = 6;
```
Devinwong: We briefly discussed this earlier.
According to protobuf doc https://protobuf.dev/programming-guides/style/, `In all cases, treat abbreviations as though they are single words: use GetDnsRequest rather than GetDNSRequest, dns_request rather than d_n_s_request`.
So I am wondering if it's absolutely necessary to use `d_n_s` in your case?

SriHarsha001: Thank you for sharing the link. This field is ok. I have made the change. I can make the changes without having to change anything else. But for other fields in localdns proto, a lot of code changes including AKS RP will have to be done which is what I am trying to avoid.

PR #6955, opened 2025-09-06
Use dnf instead of tdnf to download and install kubelet/kubectl

aadhar-agarwal:
**What type of PR is this?**
/kind cleanup

- This PR uses dnf to download and install kubelet/kubectl instead of tdnf

- Generally, use dnf to install/download packages instead of tdnf

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -165,9 +165,9 @@ installRPMPackageFromFile() {
 
     rpmFile=$(find "${downloadDir}" -maxdepth 1 -name "${packagePrefix}" -print -quit 2>/dev/null) || rpmFile=""
     if [ -z "${rpmFile}" ]; then
-        tdnf_update || exit $ERR_APT_DIST_UPGRADE_TIMEOUT
+        dnf_update || exit $ERR_APT_DIST_UPGRADE_TIMEOUT
```
lilypan26: Was discussing with Henry, and this line should be `dnf_makecache` instead of `dnf_update`. But I see you're already calling `dnf_makecache` in `dnf_download` so I think we can just get rid of this line altogether.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -193,7 +192,8 @@ downloadPkgFromVersion() {
     packageVersion="${2:-}"
     downloadDir="${3:-"/opt/${packageName}/downloads"}"
     mkdir -p ${downloadDir}
-    tdnf_download 30 1 600 ${downloadDir} ${packageName}=${packageVersion} || exit $ERR_APT_INSTALL_TIMEOUT
+    dnf_download 30 1 600 ${downloadDir} ${packageName}-${packageVersion}  || exit $ERR_APT_INSTALL_TIMEOUT
+    dnf_install 30 1 600 ${packageName}-${packageVersion} || exit $ERR_APT_INSTALL_TIMEOUT
```
hbeberman: Why are we're switching from just calling download to calling both download and install?

aadhar-agarwal: We should just be downloading. Updated!

PR #6961, opened 2025-09-08
test: Fix GetMaintainedLinuxSIGImageConfigMap test after KataCC freeze

jepio:
**What type of PR is this?**
/kind failing-test

Now that KataCC images are frozen, GetMaintainedLinuxSIGImageConfigMap() returns a different output than expected. Fix the unit test.

The test passed before merging because on that day the frozen version was equal to the latest published version of other images. The day after this wasn't the case any more.

Fixes #6861

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
replaced by #6962

PR #6967, opened 2025-09-08
parser, helper and template changes for localdns scriptless

SriHarsha001:
**What type of PR is this?**
/kind feature

This PR adds parser, helper changes for localdns scriptless implementation. It also adds localdns template file. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


aks-node-controller/parser/templates/localdns.toml.gtpl
Devinwong, line 1: If you haven't, could you also check from a VM (probably from e2e) that the final localdns file is 100% as expected?
For examples, all variables are converted to expected values, all if else conditions are correct.

PR #6975, opened 2025-09-09
Add AMDAMA_NODE support

r2k1:


aks-node-controller/parser/helper.go
```diff
@@ -525,6 +525,15 @@ func getIsSgxEnabledSKU(vmSize string) bool {
...
 }
 
+// IsAmdAmaEnabledSKU determines if an VM SKU has AMD AMA GPU HW support.
+func getIsAmdAmaEnabledSKU(vmSize string) bool {
+	switch vmSize {
+	case helpers.VMSizeStandardNM16adsMA35D:
+		return true
+	}
+	return false
+}
```
r2k1, line 535: With this is approach you can only introduce a new VM SKU after releasing a new VHD and re-image.

If you introduce a new variable to AKSNodeConfig then you may to enable the feature without re-imaging things. Though RP release can be slower than VHD release.

I don't know what is better. Just think about it.

We probably should also considered on which layer we want to make the decision based on VM SKU instead of scattering based on the use-case

aks-node-controller/parser/parser.go
```diff
@@ -80,6 +80,7 @@ func getCSEEnv(config *aksnodeconfigv1.Configuration) map[string]string {
 		"API_SERVER_NAME":                                    config.GetApiServerConfig().GetApiServerName(),
 		"IS_VHD":                                             fmt.Sprintf("%v", getIsVHD(config.IsVhd)),
 		"GPU_NODE":                                           fmt.Sprintf("%v", getEnableNvidia(config)),
+		"AMDAMA_NODE":									      fmt.Sprintf("%v", getIsAmdAmaEnabledSKU(config.GetVmSize())),
```
r2k1, line 83: Can you use `agent.IsAmdAmaEnabledSKU` instead?

pkg/agent/datamodel/helper.go
```diff
@@ -44,6 +44,15 @@ func IsSgxEnabledSKU(vmSize string) bool {
...
 }
 
+// IsAmdAmaEnabledSKU determines if an VM SKU has AMD AMA GPU HW support.
+func IsAmdAmaEnabledSKU(vmSize string) bool {
+	switch vmSize {
+	case "Standard_NM16ads_MA35D":
+		return true
+	}
+	return false
+}
```
r2k1, line 54: I don't think you need it? Unless you want to use such method in RP?

aks-node-controller/parser/helper.go
mipresmsft: This is the approach recommended by Sylvain and Yi.

aks-node-controller/parser/parser.go
mipresmsft: Adding support to be consistent with the approach from other SKUs.

pkg/agent/datamodel/helper.go
mipresmsft: Adding support to be consistent with the approach from other SKUs.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -948,6 +948,32 @@ ensureGPUDrivers() {
...
+        sudo tdnf install -y $AMD_AMA_DRIVER_PACKAGE
+        # Install core package
+        sudo tdnf install -y libzip
+        sudo tdnf install -y azurelinux-repos-extended
+        sudo RPM_FRONTEND=noninteractive tdnf install -y https://download.microsoft.com/download/16b04fa7-883e-4a94-88c2-801881a47b28/amd-ama-core_1.3.0-2503242033-amd64.rpm
+        # Install AKS device plugin
+        sudo tdnf install -y amdama-device-plugin.x86_64
+        # Configure huge pages
+        sudo sh -c "echo 'vm.nr_hugepages=4096' >> /etc/sysctl.conf"
+        sudo sh -c "echo 4096 >> /proc/sys/vm/nr_hugepages"
```
r2k1, line 970: `sh -c` feels redundant

aks-node-controller/helpers/const.go
```diff
@@ -11,6 +11,7 @@ const (
 	LoadBalancerStandard = "Standard"
 	VMSizeStandardDc2s   = "Standard_DC2s"
 	VMSizeStandardDc4s   = "Standard_DC4s"
+	VMSizeStandardNM16adsMA35D = "Standard_NM16ads_MA35D"
```
r2k1, line 14: Minor thing. `Standard_` constants doesn't feel like have purpose other than satisfy some random linters.

I would prefer something like `AMDAMAEnabledVMSKUs = ["Standard_NM16ads_MA35D"]` (or any other suitable name)

mipresmsft: Just following the convention...

parts/linux/cloud-init/artifacts/cse_config.sh
mipresmsft: I seem to recall it doesn't work without "sh -c".

PR #6976, opened 2025-09-09
chore: logging & tagging for e2e

timmy-wright:
**What type of PR is this?**

/kind chore

* adds an owner tag to VMSS scale sets when they are run locally - makes life easier when using KEEP_VMSS
* adds a link to the portal for VMSS scale sets

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/types.go
+++ b/e2e/types.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"os/user"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


e2e/types.go
```diff
@@ -187,20 +188,29 @@ func (s *Scenario) PrepareVMSSModel(ctx context.Context, t *testing.T, vmss *arm
...
 
 	if config.Config.BuildID != "" {
-		if vmss.Tags == nil {
-			vmss.Tags = map[string]*string{}
-		}
 		vmss.Tags[buildIDTagKey] = &config.Config.BuildID
 	}
+
+	if config.Config.IsLocalBuild() {
+		currentUser, err := user.Current()
```
r2k1: I think you can get azure user from SDK. it will be more useful.
And probably don't need to check for local build in this case.

timmy-wright: image

Managed to fix.

timmy-wright:
Force merged because failing tests are intermittend and unrelated.

PR #6977, opened 2025-09-09
localdns scriptless cseconfig and e2e updates

SriHarsha001:
**What type of PR is this?**
/kind feature

This PR adds code in cse_config, cse_main and e2e for localdns scriptless testing. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


Devinwong (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1047,4 +1047,38 @@ enableLocalDNS() {
...
 
+# localdns corefile used by localdns systemd unit. 
+LOCALDNS_COREFILE="/opt/azure/containers/localdns/localdns.corefile"
+# localdns slice file used by localdns systemd unit.
+LOCALDNS_SLICEFILE="/etc/systemd/system/localdns.slice"
+# This function is called from cse_main.sh.
+# It creates the localdns corefile and slicefile, then enables and starts localdns.
+# In this function, generated base64 encoded localdns corefile is decoded and written to the corefile path.
+# This function also creates the localdns slice file with memory and cpu limits, that will be used by localdns systemd unit.
+shouldEnableLocalDns() {
```
cameronmeissner, line 1058: nit: can we drop the `should`? -> `enableLocalDNS()`

e2e/node_config.go
```diff
@@ -184,6 +184,75 @@ func nbcToAKSNodeConfigV1(nbc *datamodel.NodeBootstrappingConfiguration) *aksnod
 		HttpProxyConfig: &aksnodeconfigv1.HttpProxyConfig{
 			NoProxyEntries: *nbc.HTTPProxyConfig.NoProxy,
 		},
+		LocalDnsProfile: &aksnodeconfigv1.LocalDnsProfile{
+			EnableLocalDns:       true,
```
cameronmeissner, line 188: I feel we should have a way of switching local DNS on/off, otherwise we'll only be getting test coverage with it enabled within agentbaker e2e's - can we gate this config behind a common set of NBC/NodeConfig mutators we can reference with test scenario definitions instead?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -300,6 +300,11 @@ EOF
     # Call enableLocalDNS to enable localdns if localdns profile has EnableLocalDNS set to true.
     logs_to_events "AKS.CSE.enableLocalDNS" enableLocalDNS || exit $?
 
+    # This is to enable localdns using scriptless.
+    if [ "${SHOULD_ENABLE_LOCALDNS}" = "true" ]; then
+        logs_to_events "AKS.CSE.shouldEnableLocalDns" shouldEnableLocalDns || exit $ERR_LOCALDNS_FAIL
```
cameronmeissner, line 305: nit: `AKS.CSE.enableLocalDNS`

also, this would be how we enable local DNS for both the legacy + scriptless approaches right? or am I missing something?

SriHarsha001: To enable localdns with legacy way, the code will use enableLocalDNS function which is defined in line 301. For scriptless, the code will use shouldEnableLocalDns.

parts/linux/cloud-init/artifacts/cse_config.sh
SriHarsha001: I have another function enableLocalDNS which will be invoked in legacy way (non scriptless)

e2e/node_config.go
SriHarsha001: Ok I will add a disable case in the next PR.

e2e/validation.go
```diff
@@ -114,8 +115,8 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 		ValidateKubeletNodeIP(ctx, s)
 	}
 
-	// TODO: Add support for AKSNodeConfig
-	if s.Runtime.NBC != nil && s.Runtime.NBC.AgentPoolProfile.LocalDNSProfile != nil {
+	// localdns is not supported on 1804, VHDAzureLinux3OSGuard, privatekube and VHDUbuntu2204Gen2ContainerdAirgappedK8sNotCached.
+	if s.VHD != config.VHDUbuntu1804Gen2Containerd && s.VHD != config.VHDUbuntu2204Gen2ContainerdPrivateKubePkg && s.VHD != config.VHDUbuntu2204Gen2ContainerdAirgappedK8sNotCached && s.VHD != config.VHDAzureLinux3OSGuard {
```
r2k1, line 119: I think you can add another property on `Image` struct instead. Like `UnsupportedLocalDNS`

PR #6995, opened 2025-09-10
Install managed GPU device plugin for Ubuntu 22.04, 24.04 and AzLinux 3.0, cleanup and renovate config

ganeshkumarashok:
**What type of PR is this?**

/kind feature

- Download and install k8s-device-plugin package for Ubuntu 24.04, Ubuntu 22.04 and AzLinux 3.0
- Removes the downloaded file, either post installation (for GPU nodes) or always (for non-GPU nodes)
- Adds E2E tests for each of the OSes and checks that the device plugin binary is installed, runs as a systemd unit, the GPU resources are advertised, and that the GPU workload is schedulable
- Added VHD content test
- MIG stub code remains till we decide how MIG config validation is done (via VMSS tags or RP API field). 
- Added renovate config

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -458,6 +458,19 @@ if [ "$OS" = "$MARINER_OS_NAME" ]  && [ "$OS_VERSION" = "2.0" ] && [ "$(isARM64)
...
 fi
 
+# Install GPU device plugin for Ubuntu 24.04 only and disable the systemd unit
+if [ "$OS" = "$UBUNTU_OS_NAME" ] && [ "${UBUNTU_RELEASE}" = "24.04" ]; then
+  echo "Installing k8s-device-plugin for Ubuntu 24.04..."
+  apt_get_install 30 1 600 k8s-device-plugin || exit $ERR_APT
+  
+  # Disable the nvidia-device-plugin systemd unit to prevent auto-start
+  echo "Disabling nvidia-device-plugin systemd unit..."
+  systemctl disable nvidia-device-plugin.service || echo "Warning: Failed to disable nvidia-device-plugin.service"
```
djsly: why a warning ? shouldnt we block the build and investigate of this starts to fail one day ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -458,6 +458,19 @@ if [ "$OS" = "$MARINER_OS_NAME" ]  && [ "$OS_VERSION" = "2.0" ] && [ "$(isARM64)
...
+  systemctl mask nvidia-device-plugin.service || echo "Warning: Failed to mask nvidia-device-plugin.service"
+  
+  echo "  - k8s-device-plugin installed and systemd unit disabled" >> ${VHD_LOGS_FILEPATH}
```
djsly: why some echo are going to `VHD_LOGS_FILEPATH` while others are not ? what about the Warinig/Error ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -458,6 +458,19 @@ if [ "$OS" = "$MARINER_OS_NAME" ]  && [ "$OS_VERSION" = "2.0" ] && [ "$(isARM64)
   installAndConfigureArtifactStreaming acr-mirror-mariner rpm
 fi
 
+# Install GPU device plugin for Ubuntu 24.04 only and disable the systemd unit
+if [ "$OS" = "$UBUNTU_OS_NAME" ] && [ "${UBUNTU_RELEASE}" = "24.04" ]; then
+  echo "Installing k8s-device-plugin for Ubuntu 24.04..."
+  apt_get_install 30 1 600 k8s-device-plugin || exit $ERR_APT
```
djsly: should we remove it during bootstrap if it's not GPU ?  I feel this is one more package customer we start reporting as

Why does my VHD have CVE for k8s-device-plugin, I don't have GPUs.

djsly: this package name is confusion, shoudnt it be `nvidida-k8s-device-plugin` ?

ganeshkumarashok: yes, we are discussing with the PMC and dalec teams for renaming this to nvidia-k8s-device-plugin. Concern is that there is already some usage of the package. So we're asking to learn more about the usage and see if they can mirror it in the new name.

ganeshkumarashok: Will switch to `nvidia-k8s-device-plugin` once that gets published

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: You're right. Made it exit

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Removed it during CSE, for CVE and space reduction

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Removed the two warnings. Now the only substantial log is going to `VHD_LOGS_FILEPATH`. The other two lines are not worth it for the file, and get encapsulated by the last log. We use a similar pattern for other logs too.

.gitignore
```diff
@@ -45,3 +45,9 @@ e2e/scenario-logs/
 
 # vhdbuilder generated prefetch scripts
 vhdbuilder/packer/prefetch.sh
+
+# Local development files
+CLAUDE.md
```
djsly: why not have a default one that works for the whole team ?

.gitignore
```diff
@@ -45,3 +45,9 @@ e2e/scenario-logs/
 
 # vhdbuilder generated prefetch scripts
 vhdbuilder/packer/prefetch.sh
+
+# Local development files
```
djsly: I dont think we need to add those to ignore, there are many files that users end up putting on their local dirve, we cant expect to have all of them added to the gitignore

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +78,13 @@ updateAptWithMicrosoftPkg() {
 
 cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST /opt/gpu
+    
+    # Remove k8s-device-plugin package on non-GPU nodes
+    if dpkg -l | grep -q "k8s-device-plugin"; then
```
djsly: cna we make the pacakge name a global var to help with the renaming ?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +78,13 @@ updateAptWithMicrosoftPkg() {
...
+        echo "Removing k8s-device-plugin package from non-GPU node..."
```
djsly: package name variable instead

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +78,13 @@ updateAptWithMicrosoftPkg() {
...
+        apt_get_purge 10 5 120 k8s-device-plugin || echo "Warning: Failed to remove k8s-device-plugin"
```
djsly: pacakge name var

djsly: when we fail to remove, we wan't to keep this package half installed ?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +78,13 @@ updateAptWithMicrosoftPkg() {
 
 cleanUpGPUDrivers() {
```
djsly, line 81: we dont have this in AzureLinux ?

.gitignore
ganeshkumarashok: That was a miss - removed it

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +459,19 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
   installAndConfigureArtifactStreaming acr-mirror-azurelinux3 rpm
 fi
 
+# Install GPU device plugin for Ubuntu 24.04 only and disable the systemd unit
+if [ "$OS" = "$UBUNTU_OS_NAME" ] && [ "${UBUNTU_RELEASE}" = "24.04" ]; then
```
djsly: is this a decision that we only support UBUNTUN24:04 for managed device plugin ?

.gitignore
ganeshkumarashok: True, we could have that

ganeshkumarashok: I have some custom instructions, but we could generalize it further

vhdbuilder/packer/install-dependencies.sh
djsly: isn't VHD_LOGS_FILEPATH the VHD release note ? how is this line helpful to have in the release note ? shouldnt we print the version that we installed ?

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: There isn't a deb package for device plugin that's available for 22.04 yet.

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Added the version

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: Original plan was to keep AzLinux and Ubuntu changes in separate PRs. With the new scope, I've added it.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: Isn't that better than failing entire CSE due to a relatively low-pri cleanup?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: I've kept it as variable per file. But since the scripts don't import from a common script, it's not global.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,52 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
...
+    pkg_version=$(dpkg-query -W -f='${Version}' "${pkg}" 2>/dev/null || echo "unknown")
+  elif [ "${OS}" = "${AZURELINUX_OS_NAME}" ]; then
+    pkg_version=$(rpm -q --qf '%{VERSION}-%{RELEASE}' "${pkg}" 2>/dev/null || echo "unknown")
+  fi
+  echo "  - ${pkg} version ${pkg_version}" >> ${VHD_LOGS_FILEPATH}
+}
+
+if should_install_device_plugin; then
+  install_device_plugin
+fi
```
djsly: might be good to print that we detected that we dont need to install ? the `should_install_` function isn't printing anything either.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,52 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
...
+
+  # Azure Linux 3.0 only
+  if [ "$OS" = "$AZURELINUX_OS_NAME" ] && [ "${OS_VERSION}" = "3.0" ]; then return 0; fi
+
+  return 1
+}
+
+install_device_plugin() {
+  local pkg="${K8S_DEVICE_PLUGIN_PKG}"
+  echo "Installing ${pkg} on ${OS} (Ubuntu ${UBUNTU_RELEASE:-n/a} / Version ${OS_VERSION:-n/a})..."
```
djsly: is `-n/a` a valid syntax ? 

also why are we printing Ubuntu and Version in the same log line ? aint those variable never set at the same time ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,52 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
   installAndConfigureArtifactStreaming acr-mirror-azurelinux3 rpm
 fi
 
+should_install_device_plugin() {
+  # Require x86_64 (skip arm64)
+  if [ "$(isARM64)" -eq 1 ]; then return 1; fi
```
surajssd: We have packages for arm64 as well.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,52 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
...
+
+  # Ubuntu 24.04 only
+  if [ "$OS" = "$UBUNTU_OS_NAME" ] && [ "${UBUNTU_RELEASE:-}" = "24.04" ]; then return 0; fi
```
surajssd: Both Ubuntu 22.04 and 24.04 are supported.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,52 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
...
+    return 1
+  fi
+
+  echo "Disabling nvidia-device-plugin systemd unit..."
+  systemctl disable nvidia-device-plugin.service || exit 1
+  systemctl mask nvidia-device-plugin.service || exit 1
+
+  # Capture installed package version for logging
+  local pkg_version="unknown"
+  if [ "${OS}" = "${UBUNTU_OS_NAME}" ]; then
```
surajssd: Can we combine this if else into the previous one, so we have everything in one place? Or a switch case is also readable since we are only the same variable `$OS`

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: I want to avoid adding all combinations in one release/PR in case there are issues that are not caught by tests.

ganeshkumarashok: Particularly given the huge usage of Ubuntu 22.04

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Yes, currently there's a separate GB 200 VHD so I don't think it's worth installing and removing for all ARM SKUs if won't be used broadly. Definitely worth coming back to this after we align on that GB 200 VHD plan.

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Added those logs.

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Syntax seems to be valid. 

And that's right: I've simplified the log line.

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: Sure, combined it and used a switch case instead

vhdbuilder/packer/install-dependencies.sh
djsly: let's update

vhdbuilder/packer/install-dependencies.sh
djsly: let's update

djsly: @ganeshkumarashok  if the package is present, and we want to be early GA. should we look at already adding the logic ? what are we gaining in merging this right now over augmenting the coverage right now ?

vhdbuilder/packer/install-dependencies.sh
djsly: @ganeshkumarashok could you clarify the "not caught by test" comment ? shouldn't this PR add the required tests as well at the same time ?  TDD ?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,59 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
...
+  if [ "$OS" = "$UBUNTU_OS_NAME" ] && [ "${UBUNTU_RELEASE:-}" = "24.04" ]; then return 0; fi
+
+  # Azure Linux 3.0 only
+  if [ "$OS" = "$AZURELINUX_OS_NAME" ] && [ "${OS_VERSION}" = "3.0" ]; then return 0; fi
+
+  echo "Skipping ${K8S_DEVICE_PLUGIN_PKG} installation: Not supported on ${OS} ${UBUNTU_RELEASE:-}${OS_VERSION:-}"
+  return 1
+}
+
+install_device_plugin() {
```
djsly: wrong function name casing.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,59 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
   installAndConfigureArtifactStreaming acr-mirror-azurelinux3 rpm
 fi
 
+should_install_device_plugin() {
```
djsly: wrong function name casing.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -459,6 +461,59 @@ elif [ "$OS" = "$MARINER_OS_NAME" ] && [ "$OS_VERSION" = "3.0" ] && [ "$(isARM64
...
+
+install_device_plugin() {
+  local pkg="${K8S_DEVICE_PLUGIN_PKG}"
+  echo "Installing ${pkg}..."
+  
+  # Install and get version based on OS
+  case "${OS}" in
+    "${UBUNTU_OS_NAME}")
+      apt_get_install 30 1 600 "${pkg}" || exit $ERR_APT
+      local pkg_version=$(dpkg-query -W -f='${Version}' "${pkg}" 2>/dev/null || echo "unknown")
```
djsly: where is ${Version} defined ?  

the package name/ version should be coming from components.json.

I'm wondering also if this shoulnt be using the same approach we are using for kubelet/kubectl   I'm not a fan of installing and uninstalling something that we don't need on 90% of the fleet. 

for GPU we normally on download the binary and remove them, we should follow the same approach here.

parts/common/components.json
```diff
@@ -1503,6 +1503,32 @@
           }
         }
       }
+    },
+    {
+      "name": "nvidia-device-plugin",
+      "downloadLocation": "/opt/nvidia-device-plugin/downloads",
+      "downloadURIs": {
+        "ubuntu": {
+          "r2404": {
```
surajssd, line 1481: This should be enabled for the ubuntu 2204 as well, right?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -400,6 +400,23 @@ function nodePrep {
...
+                    logs_to_events "AKS.CSE.installNvidiaDevicePlugin" "installNvidiaDevicePluginPkgFromCache ${DEVICE_PLUGIN_VERSION}"
+                else
+                    echo "ERROR: No cached nvidia-device-plugin package found" >&2
+                    exit $ERR_GPU_DEVICE_PLUGIN_START_FAIL
+                fi
+            else
+                echo "nvidia-device-plugin package already installed"
+            fi
+        fi
+
```
surajssd: Line # 426 right now hardcodes the MIG strategy.

surajssd: Ideally we should manage the installation + drop in login in one place. 

Drop in could include things like MIG, managing the resource constraints for the service, and any other modificaitons to the way the service is started.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -400,6 +400,23 @@ function nodePrep {
             logs_to_events "AKS.CSE.ensureMigPartition" ensureMigPartition
         fi
 
+        # Install GPU device plugin if needed and not already installed
+        if [ "${ENABLE_GPU_DEVICE_PLUGIN_IF_NEEDED}" = "true" ]; then
+            if ! systemctl list-unit-files | grep -q "nvidia-device-plugin.service"; then
+                echo "Installing nvidia-device-plugin package..."
+                # Determine version from components.json through available cached packages
+                DEVICE_PLUGIN_VERSION=$(find /opt/nvidia-device-plugin/downloads -name "nvidia-device-plugin-*" -type f | head -1 | grep -o '[0-9]\+\.[0-9]\+\.[0-9]\+[^/]*' | head -1)
```
surajssd: What's the point of sending the device plugin version if we are installing whatever is available?

I think we are finding version here of the first `*.rpm/*.deb` we find and then passing it to the `installNvidiaDevicePluginPkgFromCache` function. Why not install directly whatever package we find in the folder?

Do we anticipate multiple versions of the package? If yes then the logic to find the package version is only picking the first one that the tool `find` reports. And `find` is not known to report results in any specific order.

ganeshkumarashok: These latest changes are not yet ready for review

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +82,36 @@ updateAptWithMicrosoftPkg() {
...
     rm -Rf $GPU_DEST /opt/gpu
+    
+    # Remove GPU device plugin package on non-GPU nodes
+    if dpkg -l | grep -q "${K8S_DEVICE_PLUGIN_PKG}"; then
+        echo "Removing ${K8S_DEVICE_PLUGIN_PKG} package from non-GPU node..."
+        if apt_get_purge 10 5 120 "${K8S_DEVICE_PLUGIN_PKG}"; then
```
djsly: is this still valid ?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +82,36 @@ updateAptWithMicrosoftPkg() {
...
+            echo "WARNING: Failed to remove ${K8S_DEVICE_PLUGIN_PKG}. Package may still be present." >&2
+        fi
+    fi
+}
+
+installNvidiaDevicePluginPkgFromCache() {
+    echo "Installing cached nvidia-device-plugin package..."
+    
+    # Find the cached .deb file
+    debFile=$(find /opt/nvidia-device-plugin/downloads -name "nvidia-device-plugin*.deb" -type f | head -1)
```
djsly: which version are we installing ?  what is there are multiple versions cached ? shouldnt this be an error ? are we expecting to cache more than 1 ?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +82,36 @@ updateAptWithMicrosoftPkg() {
...
+    if [ -z "${debFile}" ]; then
+        echo "ERROR: No cached nvidia-device-plugin package found" >&2
+        exit $ERR_GPU_DEVICE_PLUGIN_START_FAIL
+    fi
+    
+    echo "Installing ${debFile}..."
+    dpkg -i "${debFile}" || exit $ERR_APT_INSTALL_TIMEOUT
+    
+    echo "Disabling nvidia-device-plugin systemd unit..."
```
djsly: we shoud wrap this around the nodebootstrapping config that will decide if we instlal it or not

ganeshkumarashok: Install case should be sufficient -- the no install case won't be needed (just the downloaded pkg will be cleaned up with the rest)

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: That was part of the clean up. Latest expanded scope of the PR wasn't ready for reviews.

ganeshkumarashok: I'm deleting the downloaded pkg instead

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: Since we only download one version, we shouldn't hit the case of multiple.  Switched it to using the latest version

e2e/scenario_nvidia_device_plugin_test.go
```diff
@@ -0,0 +1,116 @@
...
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				// Validate that the NVIDIA device plugin systemd service is running
+				ValidateNvidiaDevicePluginServiceRunning(ctx, s)
+				
+				// Validate that GPU resources are advertised by the device plugin
+				ValidateNodeAdvertisesGPUResources(ctx, s)
+				
+				// Validate that the NVIDIA device plugin binary was installed correctly
+				ValidateNvidiaDevicePluginBinaryInstalled(ctx, s)
```
surajssd: Ideally this should be the first test that runs in this validator function.

e2e/validators.go
```diff
@@ -974,3 +974,86 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
...
+}
+
+func ValidateNvidiaDevicePluginBinaryInstalled(ctx context.Context, s *Scenario) {
+	s.T.Helper()
+	s.T.Logf("validating that NVIDIA device plugin binary is installed")
+	
+	command := []string{
+		"set -ex",
+		"dpkg -l | grep nvidia-device-plugin || rpm -qa | grep nvidia-device-plugin",
+	}
```
surajssd: Can we do this instead for more resilient code, so we know this is failing reliably and not because of the side effect that `dpkg not found` or `rpm not found` on the other OSes:


```suggestion
    command := []string{"set -ex"}

	switch s.Config.VHD.OS {
	case config.OSAzureLinux:
		command = append(command, "rpm -qa | grep nvidia-device-plugin")
	case config.OSUbuntu:
		command = append(command, "dpkg -l | grep nvidia-device-plugin")
	default:
		s.T.Fatalf("unsupported OS for NVIDIA device plugin validation: %s", s.Config.VHD.OS)
	}
```

e2e/validators.go
```diff
@@ -974,3 +974,86 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
...
+					Args: []string{
+						"--max-steps", "1",
+					},
+					Resources: corev1.ResourceRequirements{
+						Limits: corev1.ResourceList{
+							"nvidia.com/gpu": resource.MustParse("1"),
+						},
+					},
+				},
+			},
```
surajssd, line 1043:
```suggestion
			},
			NodeName: s.Runtime.KubeNodeName,
```

I know we will only have one node per test, this just hammers the point that the pod is running on this node.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -400,17 +400,32 @@ function nodePrep {
...
+            # Ensure kubelet device-plugins directory exists
+            mkdir -p /var/lib/kubelet/device-plugins
+            
+            if [ "${MIG_NODE}" = "true" ]; then
                 tee "/etc/systemd/system/nvidia-device-plugin.service.d/10-mig_strategy.conf" > /dev/null <<'EOF'
 [Service]
 Environment="MIG_STRATEGY=--mig-strategy single"
 ExecStart=
-ExecStart=/usr/local/nvidia/bin/nvidia-device-plugin $MIG_STRATEGY
+ExecStart=/usr/bin/nvidia-device-plugin $MIG_STRATEGY
```
surajssd: Correct me if I am wrong here, but I think that the `MIG_STRATEGY` is being populated from the line # 421 with value `--mig-strategy single`, so no matter what the MIG strategy is, it is always gonna be single.

e2e/scenario_nvidia_device_plugin_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"context"
+	"testing"
+
+	"github.com/Azure/agentbaker/e2e/config"
+	"github.com/Azure/agentbaker/pkg/agent/datamodel"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v6"
+)
+
+func Test_Ubuntu2404_NvidiaDevicePluginRunning(t *testing.T) {
```
surajssd, line 15: Can we also add a test for MIG enabled nodes? And we should test it for mixed strategy.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -776,6 +780,11 @@ testPkgDownloaded() {
       if [ -z "${rpmFile}" ]; then
         err $test "Package ${packageName}-${packageVersion} does not exist, content of downloads dir is $(ls -al ${downloadLocation})"
       fi
+    elif [ $OS = $MARINER_OS_NAME ] || { [ $OS = $AZURELINUX_OS_NAME ] && [ $OS_VERSION = "2.0" ]; }; then
```
surajssd: We don't have package for Az Linux 2.0. This should've failed for that Az Linux 2.0.

surajssd (changes requested):

pkg/agent/baker_test.go
```diff
@@ -1987,6 +1987,24 @@ oom_score = -999
 
 				Expect(containerdConfigFileContent).To(Equal(expectedShimConfig))
 			}),
+		Entry("AKSUbuntu2204 with NVIDIA Device Plugin enabled", "AKSUbuntu2204+Containerd+DevicePlugin", "1.29.7",
```
djsly, line 1990: why 1.29.7 ? 1.31.X is end of support

djsly: ah this only verifies the logic I guess,

vhdbuilder/packer/test/linux-vhd-content-test.sh
djsly: right, confused as well why this if clause is needed

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -400,17 +400,32 @@ function nodePrep {
...
+            if ! systemctl list-unit-files | grep -q "nvidia-device-plugin.service"; then
+                echo "Installing nvidia-device-plugin package..."
+                logs_to_events "AKS.CSE.installNvidiaDevicePlugin" "installNvidiaDevicePluginPkgFromCache"
+            else
+                echo "nvidia-device-plugin package already installed"
+            fi
+        fi
+
         # Configure GPU device plugin
         if [ "${ENABLE_GPU_DEVICE_PLUGIN_IF_NEEDED}" = "true" ]; then
```
djsly: why two different if clause with the same comparaison ? line 414 and 404 should be combiened no ?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +82,29 @@ updateAptWithMicrosoftPkg() {
...
+        rm -rf /opt/nvidia-device-plugin/downloads
+        echo "Cached ${K8S_DEVICE_PLUGIN_PKG} downloads removed from non-GPU node"
+    fi
+}
+
+installNvidiaDevicePluginPkgFromCache() {
+    echo "Installing cached nvidia-device-plugin package..."
+    
+    # Find the cached .deb file with highest version
+    debFile=$(find /opt/nvidia-device-plugin/downloads -name "nvidia-device-plugin*.deb" -type f | sort -V | tail -1)
```
djsly: are we expecting to have multiple cached versions ?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -257,6 +261,29 @@ ensureRunc() {
...
+        rm -rf /opt/nvidia-device-plugin/downloads
+        echo "Cached ${K8S_DEVICE_PLUGIN_PKG} downloads removed from non-GPU node"
+    fi
+}
+
+installNvidiaDevicePluginPkgFromCache() {
+    echo "Installing cached nvidia-device-plugin package..."
+    
+    # Find the cached .rpm file with highest version
+    rpmFile=$(find /opt/nvidia-device-plugin/downloads -name "nvidia-device-plugin*.rpm" -type f | sort -V | tail -1)
```
djsly: are we expecting to have multiple cached versions ?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -257,6 +261,29 @@ ensureRunc() {
...
+    if [ -z "${rpmFile}" ]; then
+        echo "ERROR: No cached nvidia-device-plugin package found" >&2
+        exit $ERR_GPU_DEVICE_PLUGIN_START_FAIL
+    fi
+    
+    echo "Installing ${rpmFile}..."
+    rpm -i "${rpmFile}" || exit $ERR_APT_INSTALL_TIMEOUT
```
djsly: we  don't have a helper function with retries 

we have installDebPackageFromFile and installRPMPackageFromFile

Please have a look at the PR from Lily, which shows all the existing helper functions

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -78,6 +82,29 @@ updateAptWithMicrosoftPkg() {
...
+    if [ -z "${debFile}" ]; then
+        echo "ERROR: No cached nvidia-device-plugin package found" >&2
+        exit $ERR_GPU_DEVICE_PLUGIN_START_FAIL
+    fi
+    
+    echo "Installing ${debFile}..."
+    dpkg -i "${debFile}" || exit $ERR_APT_INSTALL_TIMEOUT
```
djsly: we don't have a helper function with retries

we have installDebPackageFromFile and installRPMPackageFromFile

Please have a look at the PR from Lily, which shows all the existing helper functions

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -257,6 +261,29 @@ ensureRunc() {
 
 cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST /opt/gpu
+    
+    # Remove cached GPU device plugin downloads on non-GPU nodes
+    if [ -d "/opt/nvidia-device-plugin/downloads" ]; then
```
djsly: we should always remove them no ?  we don't need cached packages even when we install them

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: No, we are not. This is just for handling in case we have multiple

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
ganeshkumarashok: You're right. It was removed in that case too, but the comments weren't right. Simplified that logic.

parts/linux/cloud-init/artifacts/cse_main.sh
ganeshkumarashok: This was part of the original script. I was considering that earlier too - it's a tradeoff between keeping that if condition shorter, with less depth vs keeping it as one. 

Different preferences but I refactored the original script like you prefer.

parts/linux/cloud-init/artifacts/cse_main.sh
ganeshkumarashok: That's desired. We are not supporting mixed since there's no way to specify that for users today.

e2e/scenario_nvidia_device_plugin_test.go
ganeshkumarashok: Mixed strategy is not applicable at this time, as we are not supporting that. Managed MIG will likely be supported via VMSS tags/another strategy later. Keeping the code for potential re-use at a later time.

vhdbuilder/packer/test/linux-vhd-content-test.sh
ganeshkumarashok: Removed that

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
ganeshkumarashok: Switched to a more verbose option but obtaining the version from components.json

e2e/validators.go
```diff
@@ -974,3 +974,93 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
...
+	gpuCapacity, exists := node.Status.Capacity["nvidia.com/gpu"]
+	require.True(s.T, exists, "node should advertise nvidia.com/gpu capacity")
+
+	gpuCount := gpuCapacity.Value()
+	require.Greater(s.T, gpuCount, int64(0), "node should advertise at least 1 GPU, but got %d", gpuCount)
+
+	s.T.Logf("node %s advertises %d nvidia.com/gpu resources", nodeName, gpuCount)
+}
+
+func ValidateNvidiaDevicePluginBinaryInstalled(ctx context.Context, s *Scenario) {
```
surajssd: Turns out there is a function for this and here is how to use it: 

https://github.com/Azure/AgentBaker/blob/3054e52fb50ef58310bfc6efc189583792e99271/e2e/scenario_test.go#L1292

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -405,15 +405,38 @@ function nodePrep {
...
+            if ! systemctl list-unit-files | grep -q "nvidia-device-plugin.service"; then
+                echo "Installing nvidia-device-plugin package..."
+                logs_to_events "AKS.CSE.installNvidiaDevicePlugin" "installNvidiaDevicePluginPkgFromCache"
+            else
+                echo "nvidia-device-plugin package already installed"
+            fi
+            
+            # Create systemd override directory and fix binary path
+            mkdir -p /etc/systemd/system/nvidia-device-plugin.service.d/
+            tee "/etc/systemd/system/nvidia-device-plugin.service.d/10-binary-path.conf" > /dev/null <<'EOF'
```
ganeshkumarashok, line 421: This is because the binary gets moved to /usr/local/bin/nvidia-device-plugin earlier

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
ganeshkumarashok: No, replaced it with the (more verbose) structure that obtains the version from components.json and checks for that.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1066,11 +1066,11 @@ configureSSHService() {
     if [ -f /etc/systemd/system/ssh.service.d/00-socket.conf ]; then
         rm /etc/systemd/system/ssh.service.d/00-socket.conf || echo "Warning: Could not remove 00-socket.conf"
     fi
-    
+
     if [ -f /etc/systemd/system/ssh.socket.d/addresses.conf ]; then
         rm /etc/systemd/system/ssh.socket.d/addresses.conf || echo "Warning: Could not remove addresses.conf"
     fi
-    
+
```
ganeshkumarashok, line 1073: Note: diffs like these are due to VS code config change for stripping out trailing whitespace (and adding a new line at EOF)

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -405,15 +405,38 @@ function nodePrep {
...
+            # Reload systemd to pick up the base path override
+            systemctl daemon-reload
+            
+            if [ "${MIG_NODE}" = "true" ]; then
                 tee "/etc/systemd/system/nvidia-device-plugin.service.d/10-mig_strategy.conf" > /dev/null <<'EOF'
 [Service]
 Environment="MIG_STRATEGY=--mig-strategy single"
 ExecStart=
-ExecStart=/usr/local/nvidia/bin/nvidia-device-plugin $MIG_STRATEGY
+ExecStart=/usr/local/bin/nvidia-device-plugin $MIG_STRATEGY
```
djsly, line 434: the rpm/deb installs the package under that folder ?

parts/common/components.json
```diff
@@ -1472,6 +1472,40 @@
           }
         }
       }
+    },
+    {
+      "name": "nvidia-device-plugin",
```
djsly, line 1477: can you augment the renovate.json/config file to add the support for this package, set the owner/reviewer

parts/linux/cloud-init/artifacts/cse_main.sh
ganeshkumarashok: it gets moved to that folder

parts/common/components.json
ganeshkumarashok: Added

djsly (approved):

e2e/scenario_nvidia_device_plugin_test.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.SKU.Name = to.Ptr("Standard_NC6s_v3")
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+
+				// Validate that the NVIDIA device plugin binary was installed correctly
+				versions := components.GetExpectedPackageVersions("nvidia-device-plugin", "ubuntu", "r2404")
+				if len(versions) > 0 {
+					ValidateInstalledPackageVersion(ctx, s, "nvidia-device-plugin", versions[0])
```
surajssd: Ideally this should fail if we can't find a package version for a given package, instead of not testing it. With this conditional we won't catch if something is wrong with the `components.json`.

e2e/scenario_nvidia_device_plugin_test.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.SKU.Name = to.Ptr("Standard_NC6s_v3")
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+
+				// Validate that the NVIDIA device plugin binary was installed correctly
+				versions := components.GetExpectedPackageVersions("nvidia-device-plugin", "ubuntu", "r2204")
+				if len(versions) > 0 {
+					ValidateInstalledPackageVersion(ctx, s, "nvidia-device-plugin", versions[0])
```
surajssd: Same as before.

e2e/scenario_nvidia_device_plugin_test.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.SKU.Name = to.Ptr("Standard_NC6s_v3")
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+
+				// Validate that the NVIDIA device plugin binary was installed correctly
+				versions := components.GetExpectedPackageVersions("nvidia-device-plugin", "azurelinux", "3.0")
+				if len(versions) > 0 {
+					ValidateInstalledPackageVersion(ctx, s, "nvidia-device-plugin", versions[0])
```
surajssd: same as before

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -257,6 +259,36 @@ ensureRunc() {
 
 cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST /opt/gpu
+    rm -rf /opt/nvidia-device-plugin/downloads
```
surajssd:
```suggestion
    rm -rf "/opt/${K8S_DEVICE_PLUGIN_PKG}/downloads"
```

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -70,14 +72,44 @@ updateAptWithMicrosoftPkg() {
...
-    
+
     retrycmd_silent 120 5 25 curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > /tmp/microsoft.gpg || exit $ERR_MS_GPG_KEY_DOWNLOAD_TIMEOUT
     retrycmd_if_failure 10 5 10 cp /tmp/microsoft.gpg /etc/apt/trusted.gpg.d/ || exit $ERR_MS_GPG_KEY_DOWNLOAD_TIMEOUT
     apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
 }
 
 cleanUpGPUDrivers() {
     rm -Rf $GPU_DEST /opt/gpu
+    rm -rf /opt/nvidia-device-plugin/downloads
```
surajssd:
```suggestion
    rm -rf "/opt/${K8S_DEVICE_PLUGIN_PKG}/downloads"
```

e2e/scenario_nvidia_device_plugin_test.go
ganeshkumarashok: It will fail subsequent checks in that same E2E. Not necessary but I have added it

PR #7004, opened 2025-09-11
Minor improvements to e2e logging

r2k1:
**What type of PR is this?**
/kind test

Minor improvements to error logs. Update missleading error messages.

### DIFF
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -367,18 +367,13 @@ func ValidateInstalledPackageVersion(ctx context.Context, s *Scenario, component
 		}
 	}()
 	execResult := execScriptOnVMForScenarioValidateExitCode(ctx, s, installedCommand, 0, "could not get package list")
-	containsComponent := func() bool {
-		for _, line := range strings.Split(execResult.stdout.String(), "\n") {
-			if strings.Contains(line, component) && strings.Contains(line, version) {
-				return true
-			}
+	for _, line := range strings.Split(execResult.stdout.String(), "\n") {
+		if strings.Contains(line, component) && strings.Contains(line, version) {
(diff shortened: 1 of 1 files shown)


e2e/validators.go
```diff
@@ -377,9 +376,9 @@ func ValidateInstalledPackageVersion(ctx context.Context, s *Scenario, component
 		return false
 	}()
 	if !containsComponent {
-		s.T.Logf("expected to find %s %s in the installed packages, but did not", component, version)
-		s.T.Fail()
+		s.T.Errorf("expected to find %s %s in the installed packages, but did not", component, version)
 	}
+	s.T.Logf("found %s %s in the installed packages", component, version)
```
jepio: Errorf continues, so this should be in an `else` branch?

r2k1: oops, good catch. Updated

e2e/validators.go
```diff
@@ -368,18 +367,13 @@ func ValidateInstalledPackageVersion(ctx context.Context, s *Scenario, component
...
+			s.T.Logf("found %s %s in the installed packages", component, version)
+			return
 		}
-		return false
-	}()
-	if !containsComponent {
-		s.T.Logf("expected to find %s %s in the installed packages, but did not", component, version)
-		s.T.Fail()
 	}
+	s.T.Errorf("expected to find %s %s in the installed packages, but did not", component, version)
```
timmy-wright, line 376: I assume that Errorf aborts the test in the same way Fail does?

r2k1: `Error` is `Log` followed by `Fail`.
Unlike `Fatal`, it fails the test without aborting it. (So multiple assertion are allowed to fail for example)

One of the issue I tried to address is we have a failed test with "Validation Succed" message

PR #7032, opened 2025-09-17
Compile aks-node-controller before running e2e

r2k1:
**What type of PR is this?**

/kind test


aks-node-controller/proto/aksnodeconfig/v1/config.proto
```diff
@@ -157,4 +157,8 @@ message Configuration {
 
   // Represents localdns profile for agentpool.
   LocalDnsProfile local_dns_profile = 41;
+
+  // If set, downloads and runs the specified binary to manage node lifecycle instead of the built-in logic.
+  // Useful for testing and development. Not intended for production use.
+  string aks_node_controller_url = 42;
```
Devinwong: I don't understand why we must add this variable to the aks-node-config if it's only for testing purposes? Why not just add a variable in our e2e somewhere so that we can provide the value on demand?

e2e/cluster.go
```diff
@@ -56,7 +56,7 @@ func (c *Cluster) MaxPodsPerNode() (int, error) {
 	return 0, fmt.Errorf("cluster agentpool profiles were nil or empty: %+v", c.Model)
 }
 
-func prepareCluster(ctx context.Context, cluster *armcontainerservice.ManagedCluster, isAirgap, isNonAnonymousPull, installNvidiaDevicePlugin bool) (*Cluster, error) {
```
Devinwong, line 59: Are we sure `installNvidiaDevicePlugin` can be removed?

aks-node-controller/proto/aksnodeconfig/v1/config.proto
r2k1: Sure, if you have an idea how to achieve it.

My issue is aks-node-controller is run on the VM boot and it's tricky to disable and run a separate binary.

e2e/cluster.go
r2k1: IDE highlights it as unused and there is no compilation errors. So it should be safe.

aks-node-controller/proto/aksnodeconfig/v1/config.proto
r2k1: I think I figure out how to avoid it. Check again.

r2k1: Flatcar + scriptless is unhappy now. Do you know how it works why it could fail?

e2e/cluster.go
djsly: it will be reused soonish lol

aks-node-controller/proto/aksnodeconfig/v1/config.proto
Devinwong: On a Flatcar+scritpless node, I don't see provisioning logs. It probably means aks-node-controller didn't execute the scripts.
```
azureuser@e1r8-2025-09-18-flatcarscriptless000000 /var/log/azure $ ls
Microsoft.Azure.Extensions.CustomScript  aks-node-controller.log  custom-script
```

Also I couldn't find aks-node-controller-hack too.
```
azureuser@e1r8-2025-09-18-flatcarscriptless000000 /var/log/azure $ ls /opt/azure/containers
aks-check-network.sh                  ensure-no-dup.sh                      provision.sh                      provision_start.sh
aks-log-collector-send.py             ensure_imds_restriction.sh            provision_cis.sh                  reconcilePrivateHosts.sh
aks-log-collector.sh                  image-bom.json                        provision_configs.sh              setup-custom-search-domains.sh
aks-node-controller                   init-aks-custom-cloud.sh              provision_installs.sh             sync-container-logs.sh
aks-node-controller-config-hack.json  kubelet.sh                            provision_installs_distro.sh      ubuntu-snapshot-update.sh
bind-mount.sh                         localdns                              provision_redact_cloud_config.py  validate-kubelet-credentials.sh
cleanup-vhd.sh                        measure-tls-bootstrapping-latency.sh  provision_send_logs.py
cloud-init-status-check.sh            mig-partition.sh                      provision_source.sh
enable-dhcpv6.sh                      prefetch.sh                           provision_source_distro.sh
```

Failed to create any file
```
azureuser@e1r8-2025-09-18-flatcarscriptless000000 /opt/azure/containers $ touch somefile
touch: cannot touch 'somefile': Permission denied
```
Maybe it's the nature of flatcar?

Devinwong: The VM is still kept. I can provide the ssh command to you if needed.

e2e/cluster.go
r2k1: I don't think so. If we need to run a device plugin pod for a scenario it should be done for a particular test. Not as a daemonset for all tests.

PR #7059, opened 2025-09-19
feat: support AAD SSH for Linux nodes

norshtein:
**What type of PR is this?**
/kind feature

This PR adds the support for AAD SSH(EntraID SSH). When AAD SSH is enabled, we should disable public key authenticate method on the node, in order to only keep AAD SSH method. This PR adds this logic and UT and E2E test for that. Note that the installation of "AADSSHLoginForLinux" extension will be managed by AKS RP.

Fixes #

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```
none
```


pkg/agent/baker.go
```diff
@@ -1105,6 +1105,9 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 		"ShouldDisableSSH": func() bool {
 			return config.SSHStatus == datamodel.SSHOff
 		},
+		"ShouldTurnOffPubkeyAuthSSH": func() bool {
+			return config.SSHStatus == datamodel.EntraIDSSH
```
djsly, line 1109: dont we need to check SSHOn as well ?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -960,6 +960,43 @@ disableSSH() {
     systemctlDisableAndStop ssh || exit $ERR_DISABLE_SSH
 }
 
+setPubkeyAuthSSH() {
+  local want="$1"
+  local SSHD_CONFIG="/etc/ssh/sshd_config"
```
djsly, line 961: this is valid on all OS family ?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -88,6 +88,12 @@ function basePrep {
 
     logs_to_events "AKS.CSE.installSecureTLSBootstrapClient" installSecureTLSBootstrapClient
 
+    if [ "${TURN_OFF_PUBKEY_AUTH_SSH}" = "true" ]; then
```
djsly: lets use a single function call in this file with logs_to_events, also have the conditional login within the function

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -88,6 +88,12 @@ function basePrep {
...
+        setPubkeyAuthSSH no || exit $ERR_CONFIG_PUBKEY_AUTH_SSH
+    else
+        setPubkeyAuthSSH yes || exit $ERR_CONFIG_PUBKEY_AUTH_SSH
+    fi
+
     if [ "${DISABLE_SSH}" = "true" ]; then
```
djsly, line 94: we shouldn't be altering the PubKey is disableSSH is enabled maybe ?

I feel we need to cleaner `configureSSH` function that properly hanled all the potential toggles in the right order to minimize unnecessary operations

e2e/validators.go
```diff
@@ -974,3 +974,22 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
 	s.T.Logf("waiting for Nvidia GPU resource to be available")
 	waitUntilResourceAvailable(ctx, s, "nvidia.com/gpu")
 }
+
+// ValidateEntraIDSSHConfigured validates that Entra ID SSH is properly configured
+// by verifying that SSH with private key authentication fails as expected
+func ValidateEntraIDSSHConfigured(ctx context.Context, s *Scenario) {
```
djsly: this is confusing, we are not really validating that EntraIDSSH is configured. We are simply acknoledging that SSH isn't working.  It could be that SSH is not configured/working for a different reason and we would think that it's cause EntraIDSSH is working.

pkg/agent/baker.go
norshtein: They're parallel relationship, the SSHStatus can only be one of `SSHOn`, `SSHOff`, `EntraIDSSH`. When specifying `EntraIDSSH`, it is ensured sshd is active(since sshd will only be turned down when `SSHOff`)

parts/linux/cloud-init/artifacts/cse_config.sh
norshtein: As far as I know, yes for all Linux distros. Searching `sshd_config` across the AgentBaker code base, all Linux distros are copying it to `/etc/ssh/sshd_config`

parts/linux/cloud-init/artifacts/cse_main.sh
norshtein: We won't. Since the SSHStatus can only be one of `SSHOn`, `SSHOff`, `EntraIDSSH` - if `EntraIDSSH`, then pubkey will be altered; if `SSHOff`(`DISABLE_SSH` will be true in this case), pubkey will not be altered but sshd will be turned down.

parts/linux/cloud-init/artifacts/cse_main.sh
norshtein: Thanks for reminding! updated.

e2e/validators.go
norshtein: yeah... since AgentBaker's E2E verification steps highly depends on SSH onto the node, we cannot perform much verification here. In AKS RP's E2E of this feature, we'll add full scenario E2E including AAD SSH onto the node. Renamed the function to `ValidatePubkeySSHDisabled` to make the name clearer.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: I dont see the udpated code ?

norshtein: sorry, I just noticed I pushed to wrong remote `origin` instead of `upstream`...

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -88,6 +88,14 @@ function basePrep {
 
     logs_to_events "AKS.CSE.installSecureTLSBootstrapClient" installSecureTLSBootstrapClient
 
+    if [ "${TURN_OFF_PUBKEY_AUTH_SSH}" = "true" ]; then
+        ssh_use_pubkey_auth="no"
+    else
+        ssh_use_pubkey_auth="yes"
+    fi
+    logs_to_events "AKS.CSE.setPubkeyAuthSSH" setPubkeyAuthSSH "${ssh_use_pubkey_auth}"
```
djsly: why not simply pass TURN_OFF_PUBKEY_AUTH_SSH instead of creating a new variable ? 

also, could we instead create a single function which would be call at this level ? 

`logs_to_events "AKS.CSE.configureSSH" configureSSH "${TURN_OFF_PUBKEY_AUTH_SSH}"`

If should handle both this logic and the one below maybe ?

```
 if [ "${DISABLE_SSH}" = "true" ]; thenExpand commentComment on line R99ResolvedCode has 
    fi
```

e2e/validators.go
```diff
@@ -974,3 +974,19 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
 	s.T.Logf("waiting for Nvidia GPU resource to be available")
 	waitUntilResourceAvailable(ctx, s, "nvidia.com/gpu")
 }
+
+// ValidatePubkeySSHDisabled validates that SSH with private key authentication fails as expected
+func ValidatePubkeySSHDisabled(ctx context.Context, s *Scenario) {
+	s.T.Helper()
+
+	connectionTest := fmt.Sprintf("%s echo 'THIS_SHOULD_FAIL'", sshString(s.Runtime.VMPrivateIP))
```
djsly: are we sure that the privilege POD has the right ssh credentials to normally make a connection over ssh ?

e2e/validators.go
```diff
@@ -974,3 +974,19 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
...
+	connectionResult, err := execOnPrivilegedPod(ctx, s.Runtime.Cluster.Kube, defaultNamespace, s.Runtime.Cluster.DebugPod.Name, connectionTest)
```
djsly: what are the connection results when trying to ssh using a key and EntraID is enabled ? should we ensure we get the specific expected error message ?

e2e/validators.go
```diff
@@ -974,3 +974,19 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
 	s.T.Logf("waiting for Nvidia GPU resource to be available")
 	waitUntilResourceAvailable(ctx, s, "nvidia.com/gpu")
 }
+
+// ValidatePubkeySSHDisabled validates that SSH with private key authentication fails as expected
+func ValidatePubkeySSHDisabled(ctx context.Context, s *Scenario) {
```
djsly, line 981: could we validate that the settings on the node are properly configured vs trying to simply ssh ?   

one intergration test with ssh and one unit test where the settings are configured as expected

parts/linux/cloud-init/artifacts/cse_main.sh
norshtein: Q: why not simply pass TURN_OFF_PUBKEY_AUTH_SSH instead of creating a new variable ?
A: basically we need to set the sshd_config to "PubkeyAuthentication yes" or "PubkeyAuthentication no". Current implementation is to let the function `setPubkeyAuthSSH` accepts "yes"/"no" as a parameter; we can definitely pass "true"/"false" to this function, and let this function then map "true"/"false" to "yes"/"no". In each implementation, we'll need to do the mapping somewhere, either outside the function or inside the function. Please let me know if you want me to moving the mapping inside the function.

Q: could we instead create a single function which would be call at this level?
A: yes, we can, but I'd suggest we do not touch existing logic in this PR in order to keep minimal code change and reduce the risk. We can refactor this part of code in another PR. WDYT?

e2e/validators.go
norshtein: I'm not sure... I'm actively updating the E2E test case to make sure it can pass...

e2e/validators.go
norshtein: Makes great sense. Updated the test to contain:
1. Unit test: rely on VMSS run command to check the sshd_config content containing specific text "PubkeyAuthentication no". 
2. Integration test: perform real SSH operation and assert the SSH will fail.

The E2E test passed in the pipeline:
 ✓ Ubuntu2204 entra IDSSH (163.41s)

e2e/validators.go
norshtein: yes, it will fail with "Permission denied" error, updated the E2E test to check returned error message contains this string.

e2e/validators.go
norshtein: Updated the E2E according to below comment.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: I mean when adding more logic, if we think it's better to combine, I prefer that we do it now, vs coming back.  What will be less risky in the future compared to if it's done now ?

e2e/validators.go
timmy-wright: I've been mulling this over recently. As long as the node joins the cluster, we can start a privileged container on it and run the test script there rather than running a container elsewhere and sshing to the node. We only really need to use another node in the cluster as a jumpbox when the node can't join the cluster - and that's to get logs. But that's a seperate PR :)

pkg/agent/datamodel/types.go
```diff
@@ -1836,6 +1836,10 @@ const (
 	SSHUnspecified SSHStatus = iota
 	SSHOff
 	SSHOn
+	// EntraIDSSH means the cluster has enabled Entra ID SSH feature on the cluster,
+	// AADSSHLoginForLinux extension will be installed on the node. In this case,
+	// we should disable Pubkey authentication in sshd_config.
+	EntraIDSSH
```
r2k1, line 1842: I think support for `scriptless` is missing

r2k1: It seems like instead of enum another way to configure the node is using booleans for enabling ssh:

https://github.com/Azure/AgentBaker/blob/master/aks-node-controller/proto/aksnodeconfig/v1/config.proto#L102

Maybe it would be easier to add another boolean instead of a new value to enum for consistency and to simplify transition.

cc @Devinwong @lilypan26

norshtein: what do you mean by "scriptless"?

norshtein: In the customer facing API, we defined `EntraIDSSH` as a new value to enum [here](https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp?path=/resourceprovider/server/microsoft.com/containerservice/datamodel/v20250902preview/types_security_profile.go&version=GBmaster&line=93&lineEnd=96&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), so in AgentBaker, we use the same implementation. We can for sure define it as a new boolean, since that's all about internal implementation. But I checked aks-node-controller's implementation, if I understand it correctly, the `enable_ssh` in proto file is manually defined, it is then mapped into "DISABLE_SSH" CSE var in [parser](https://github.com/Azure/AgentBaker/blob/master/aks-node-controller/parser/parser.go#L97). The way we define `EntraIDSSH` won't affect the parser: in the proto, define a new boolean `disablePubkeyAuthentication`, and then in praser, convert it to CSE var `TURN_OFF_PUBKEY_AUTH_SSH`. I think we may keep current implementation since it won't affect parser in aks-node-controller, please correct me if I'm wrong.

r2k1: You don't set ENV variables directly. The VM is configured either by modifying `NodeBootstrappingConfiguration` (current path) or by `AKSNodeConfig` (upcoming path, aka scriptless).

And you need to support both ways. There are e2e tests with `_Scriptless` suffix testing it.
You will need to add a new field to `AKSNodeConfig` and verify it works with e2e.

There is no enum for SSH in `AKSNodeConfig`. So I suggest same data type in both configs to simplify migration from one to another. Additionally, enums in `AKSNodeConfig` should be immutable (no new values or modifying existing, so I usually recommend avoiding them). Same config will be used on all VHDs, including ones released >6 months ago.

norshtein: Thanks for the explanation! I added scriptless implementation and UT and E2E test. But since the "cse_main.sh" containing disable pubkey is not built into the VHD image yet, so the E2E for scriptless failed. I checked its provision log, and confirmed it's caused by "cse_main.sh" does not contain the logic for handling disable pubkey. Is there any way to workaround this? Or we need to add the E2E for scriptless after latest "cse_main.sh" it built into the VHD image?

r2k1:
There are two pipelines running:
1) E2E using VHD from master.
2) Build VHD + E2E

The second pipeline build a subset of VHDs and runs E2E agains them. It's also blocks PR from merging on failure.
You can see the test results here. It seems to be passing


https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=138210258&view=logs&jobId=fa0bd643-15ef-5339-b190-0ae2e9563c80&j=c7a068aa-0008-5656-5cb7-beef58e331c1&t=f9c7131b-e96c-5ba9-ae38-8cb49f9c490c

e2e/validators.go
```diff
@@ -974,3 +976,50 @@ func ValidateEnableNvidiaResource(ctx context.Context, s *Scenario) {
...
+	require.NoError(s.T, err, "Failed to marshal run command response")
+	s.T.Logf("Run command output: %s", string(respJson))
+
+	// Parse the JSON response to extract the output and exit code
+	respString := string(respJson)
+
+	// Check if the command execution was successful by looking for our success message in the output
+	if !strings.Contains(respString, "SUCCESS: PubkeyAuthentication is disabled") {
+		s.T.Fatalf("PubkeyAuthentication is not properly disabled. Full response: %s", respString)
+	}
```
r2k1, line 1015: I usually use `require.Contains`, but it's not important

aks-node-controller/proto/aksnodeconfig/v1/config.proto
```diff
@@ -157,4 +157,10 @@ message Configuration {
 
   // Represents localdns profile for agentpool.
   LocalDnsProfile local_dns_profile = 41;
+
+  // Specifies whether to disable ssh public key authentication on the VM node
+  // This is labeled as optional (explicit presence) so that we know whether it's set or not.
+  // If it's not set, the default value will be nil, but will be set to false on the VHD.
+  // That is, the default behavior is to enable ssh public key authentication.
+  optional bool disable_pubkey_auth = 42;
```
Devinwong, line 165: IIUC, if you just want the default value to be `false`, you don't need to set it as `optional`. 
`explicit presence` is only needed if you want to know if the variable is in `unset` state and you have additional logic to handle it. In this case, it seems you don't need it. (Please correct me if I am wrong)

Without `optional`, if user sets it to `true`, you will get a `true`. If user sets it to `false` or unsets it, you will get `false`. And I think this is what you need.

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -72,6 +72,7 @@ RUNC_VERSION={{GetParameter "runcVersion"}}
 RUNC_PACKAGE_URL={{GetParameter "runcPackageURL"}}
 ENABLE_HOSTS_CONFIG_AGENT="{{EnableHostsConfigAgent}}"
 DISABLE_SSH="{{ShouldDisableSSH}}"
+TURN_OFF_PUBKEY_AUTH_SSH="{{ShouldTurnOffPubkeyAuthSSH}}"
```
Devinwong: Why not just use the same name `DisablePubkeyAuth`? Here it will be `DISABLE_PUBKEY_AUTH`. Then it will be clear between the aks-node-config and cse_cmd.sh

aks-node-controller/proto/aksnodeconfig/v1/config.proto
r2k1: With `optional` you can alter default behaviour in the future. You loose this opportunity without it.

pkg/agent/datamodel/types.go
Devinwong: The E2E test isn't tagged as `required` in the PR check in so it's fine. When the build VHD tests run, it will also run E2E with the newly built VHDs. As long as your test can pass there, it should be good to go.

aks-node-controller/proto/aksnodeconfig/v1/config.proto
Devinwong: yes. If that's needed or will be needed. No harm to keep it as `optional`. Just wanted to make it clear.

parts/linux/cloud-init/artifacts/cse_cmd.sh
norshtein: Good point! Renamed it.

parts/linux/cloud-init/artifacts/cse_main.sh
Devinwong: @norshtein, I will also suggest not adding a new variable outside your function `setPubkeyAuthSSH` if it's evitable. Having it self-contained will be easier to maintain and debug in long term and is the best practice in general.
Now that you have all required tests covering your changes, it should be quick to update this function. 🙂

Devinwong:
> Hey @r2k1 @Devinwong , I've updated the code according to code review feedback, and latest commit's E2E passed:
> 
> https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=138355619&view=results ✓ Ubuntu2204 entra IDSSH (189.54s) ✓ Ubuntu2204 entra IDSSH scriptless (183.35s)
> 
> Kindly request your approval on this PR, please let me know if more change is required.

Thanks for making the changes.
I left one last comment. All the others look good to me. Will approve once the remaining one is resolved. Thanks.

Devinwong (approved):
Thanks for all the changes. Approved!

PR #7064, opened 2025-09-20
Change katacc frozen version to 20250820.1

Bickor:
**What type of PR is this?**

Freezing the katacc version to the current publicly available: "202508.20.1"

This is the latest recorded version to be fully working, and the currently frozen version 0905 has been unable to be tested, as well as was not released fully to all regions. This PR ensures that we maintain the current working version to avoid any customer issue.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


cameronmeissner:
the existing image version `202509.05.0` was correctly released to all regions and tested, we can keep this the same and potentially look at cherry-picking in the release notes for these SKUs specifically

PR #7077, opened 2025-09-24
Don't recompile scriptless binary if an image hasn't been rebuild

r2k1:
/kind test

After #7032 we re-upload aks-node-controller on every E2E run for scriptless tests.
But we don't need if VHD has been rebuild in the same build run. Disable re-compilation for more accurate testing.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhd-template.yaml
@@ -103,7 +103,9 @@ stages:
     dependsOn: build_${{ parameters.stageName }}
     condition: and(succeeded(), eq('${{ parameters.build }}', True), ne(variables.SKIP_E2E_TESTS, 'true'))
     variables:
+      VHD_BUILD_ID: $(Build.BuildId)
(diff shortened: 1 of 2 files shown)


.pipelines/templates/.build-and-test-windows-vhd-template.yaml
```diff
@@ -103,7 +103,9 @@ stages:
     dependsOn: build_${{ parameters.stageName }}
     condition: and(succeeded(), eq('${{ parameters.build }}', True), ne(variables.SKIP_E2E_TESTS, 'true'))
     variables:
+      VHD_BUILD_ID: $(Build.BuildId)
       TAGS_TO_RUN: imageName=${{ parameters.imageName }}
+      DISABLE_SCRIPTLESS_COMPILATION: true
```
timmy-wright, line 108: I'd prefer to pass these around as template parameters, but guess it's too late to change these now.

timmy-wright (approved):

PR #7083, opened 2025-09-25
Security: migrating to oss/v2 for the pause image, aligning version to 3.6 for linux and 3.10.1 for windows

djsly:
Migrating pause image to come from oss/v2 (Dalec build) which are version of the image built using msft-go for better fips compliance.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/spec/vhdbuilder/packer/ensure_sig_vhd_exists_spec.sh
+++ b/spec/vhdbuilder/packer/ensure_sig_vhd_exists_spec.sh
@@ -27,6 +27,7 @@ Describe 'ensure_sig_vhd_exists function'
     MOCK_AZ_IMAGE_VERSIONS=""
 
     # Create mocks for external commands
+    # shellcheck disable=SC2329
     az() {
       case "$1 $2" in
         "sig show")
@@ -85,6 +86,7 @@ Describe 'ensure_sig_vhd_exists function'
       return 0
     }
 
+    # shellcheck disable=SC2329
     jq() {
       case "$*" in
         *provisioningState*)
@@ -102,6 +104,7 @@ Describe 'ensure_sig_vhd_exists function'
       esac
     }
 
+    # shellcheck disable=SC2329
     grep() {
       # Mock grep for cvm feature flag checking
       case "$*" in


cameronmeissner (approved):

cameronmeissner (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -628,19 +628,47 @@ retagAKSNodeCAWatcher() {
 retagAKSNodeCAWatcher
 capture_benchmark "${SCRIPT_NAME}_retag_aks_node_ca_watcher"
 
-pinPodSandboxImage() {
-  # This function pins the pod sandbox image to avoid Kubelet's Garbage Collector (GC) from removing it.
+pinPodSandboxImages() {
```
cameronmeissner, line 665: can we add something in VHD content tests to verify this is working as expected?

djsly: that's how I discovered it was failing ;) there is already a test

parts/common/components.json
```diff
@@ -115,6 +115,22 @@
         }
       ]
     },
+    {
+      "downloadURL": "mcr.microsoft.com/oss/v2/kubernetes/pause:*",
+      "amd64OnlyVersions": [],
+      "multiArchVersionsV2": [
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/pause",
+          "latestVersion": "3.6"
```
Devinwong, line 129: We should use `static` version instead of `floating` version, right? e.g. `3.6-2`
https://github.com/Azure/dalec-build-defs/blob/main/faq.md#how-do-floating-vs-static-tags-work

parts/common/components.json
```diff
@@ -115,6 +115,22 @@
...
+        }
+      ],
+      "windowsVersions": [
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/pause",
+          "latestVersion": "3.10.1"
```
Devinwong, line 135: Same. We should use static version

parts/common/components.json
Devinwong: Also, 3.6-2 doesn't contain `patch` version so the current regex in renovate.json line 67 wouldn't be able to detect it.
I asked copilot and this is its suggestion. FYR.
`^v?(?<major>\d+).(?<minor>\d+)(?:.(?<patch>\d+))?(?:-(?<prerelease>\d+))?$`
Tested with the following cases:
3.6.1 → major=3 minor=6 patch=1 (no prerelease)
3.6.1-1 → major=3 minor=6 patch=1 prerelease=1
3.6-1 → major=3 minor=6 (no patch) prerelease=1
3.9 → major=3 minor=9 (no patch, no prerelease)

djsly: humm we need to have the version stored as 3.6 , is this possible ?

djsly: talked with Devin, we will keep 3.6 since this version is 4 years old.

PR #7093, opened 2025-09-29
Log the aksNodeConfig parsing error but not return error, to provide backward compatibility

Devinwong:
**What type of PR is this?**
/kind bug

Log the aksNodeConfig parsing error but not return error, to provide backward compatibility.
Check the comments in codes for more info.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/aks-node-controller/app_test.go
+++ b/aks-node-controller/app_test.go
@@ -228,3 +228,48 @@ func TestApp_ProvisionWait(t *testing.T) {
 		})
 	}
 }
+func TestApp_Run_Integration(t *testing.T) {
+	t.Run("success case", func(t *testing.T) {
+		mc := &MockCmdRunner{}
+		app := &App{cmdRunner: mc.Run}
+		// Use a valid provision config file from testdata
+		exitCode := app.Run(context.Background(), []string{"aks-node-controller", "provision", "--provision-config=parser/testdata/test_aksnodeconfig.json"})
+		assert.Equal(t, 0, exitCode)
+	})
+
+	t.Run("failure case - unknown command", func(t *testing.T) {
+		mc := &MockCmdRunner{}
+		app := &App{cmdRunner: mc.Run}
+		exitCode := app.Run(context.Background(), []string{"aks-node-controller", "unknown"})
+		assert.Equal(t, 1, exitCode)
+	})
+
+	t.Run("failure case - missing command argument", func(t *testing.T) {
+		mc := &MockCmdRunner{}
+		app := &App{cmdRunner: mc.Run}
+		exitCode := app.Run(context.Background(), []string{"aks-node-controller"})
+		assert.Equal(t, 1, exitCode)
+	})
+
+	t.Run("failure case - command runner returns ExitError", func(t *testing.T) {
+		mc := &MockCmdRunner{
+			RunFunc: func(cmd *exec.Cmd) error {
+				return &ExitError{Code: 42}
+			},
+		}
+		app := &App{cmdRunner: mc.Run}
+		exitCode := app.Run(context.Background(), []string{"aks-node-controller", "provision", "--provision-config=parser/testdata/test_aksnodeconfig.json"})
+		assert.Equal(t, 42, exitCode)
+	})
+
+	t.Run("failure case - command runner returns generic error", func(t *testing.T) {
+		mc := &MockCmdRunner{
+			RunFunc: func(cmd *exec.Cmd) error {
+				return errors.New("generic error")
+			},
+		}
+		app := &App{cmdRunner: mc.Run}
+		exitCode := app.Run(context.Background(), []string{"aks-node-controller", "provision", "--provision-config=parser/testdata/test_aksnodeconfig.json"})
+		assert.Equal(t, 1, exitCode)
+	})
+}


aks-node-controller/app.go
```diff
@@ -77,8 +77,12 @@ func (a *App) run(ctx context.Context, args []string) error {
 		provisionOutput, err := a.ProvisionWait(ctx, provisionStatusFiles)
 		//nolint:forbidigo // stdout is part of the interface
 		fmt.Println(provisionOutput)
+		if err != nil {
+			slog.Error("aks-node-controller failed", "error", err)
```
r2k1: Usually, there is a no reason to log error and return it at the same time.

The same error will be logged multiple times and you will get a lot of noise.
I usually log errors only once it discarded. If any extra context is needed, then add it with `fmt.Errorf`

r2k1: It sounds like error handling in `App.Run()` isn't right?
And in some cases an error is silently ignored?

aks-node-controller/app.go
```diff
@@ -92,7 +96,17 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
...
 	config, err := nodeconfigutils.UnmarshalConfigurationV1(inputJSON)
 	if err != nil {
-		return fmt.Errorf("unmarshal provision config: %w", err)
+		// We try our best to continue unmarshal even if there are unexpected situations such as unknown fields.
+		// It usually happens when a newer version of aksNodeConfig is being parsed by an older version of aks-node-controller.
+		// This allows older versions of aks-node-controller to read configurations that may have fields added in newer versions.
+		// Log the error and continue processing.
+		// Note: This may result in loss of data if the unknown fields are critical but that's the best we can do for backward compatibility so far.
+
+		slog.Warn("Unmarshalling aksNodeConfigv1 encounters error but the process will continue."+
```
r2k1: I usually prefer to avoid WARN level and use either INFO or ERROR.
It's unlikely we're going to threat it any different to INFO or ERROR. But it often incorrectly filtered out in log queries.

aks-node-controller/pkg/nodeconfigutils/utils_test.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+
+			// Handle special test file case
+			if tt.name == "valid config from test file" {
+				validateTestFileResult(t, got)
+				return
+			}
+
+			// Validate normal test cases
+			validateTestResult(t, got, tt.want)
```
r2k1: I think it's better to inline all the logic here and use `assert` library for assertion.

Something like `assert.Equal(t, tt.want, got)` should do the job?

aks-node-controller/pkg/nodeconfigutils/utils_test.go
```diff
@@ -0,0 +1,179 @@
...
+				return
+			}
+
+			// Early return for error cases
+			if tt.wantErr {
+				return
+			}
+
+			// Handle special test file case
+			if tt.name == "valid config from test file" {
```
r2k1: I don't like checking the test name, looks like a hack. It's better to create a separate test if you can't integrate it in existing table.

aks-node-controller/pkg/nodeconfigutils/utils_test.go
```diff
@@ -0,0 +1,179 @@
...
+			// Check error expectation
+			if (err != nil) != tt.wantErr {
+				t.Errorf("UnmarshalConfigurationV1() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			// Early return for error cases
+			if tt.wantErr {
+				return
+			}
```
r2k1, line 131:
```
if tt.wantErr {
  require.Error(t, err)
} else {
  require.NoError(t, err)
}
```

I think something like this is simpler

aks-node-controller/app.go
r2k1: But I don't have strong opinion on it

aks-node-controller/app.go
Devinwong: Makes sense. Changed it back to the original code.

aks-node-controller/app.go
Devinwong: Kinda makes sense. I will use info too.

aks-node-controller/pkg/nodeconfigutils/utils_test.go
Devinwong: updated.

aks-node-controller/pkg/nodeconfigutils/utils_test.go
Devinwong: updated. Add a new test instead.

aks-node-controller/pkg/nodeconfigutils/utils_test.go
Devinwong: Copilot did the original ones. 😆
Yep updated to your pattern. Thanks.

r2k1 (approved):

r2k1 (approved):

PR #7044, opened 2025-09-18
Removed kata-cc artifacts.

Bickor:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Removing kata-cc build bits since they're no longer supported in build pipeline, and now ingesting kata VHD components from build pipeline instead of sign pipeline (since sign pipeline was only used for kata-cc).

[Successful Test](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=137712658&view=results)

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version
- [X] commits are GPG signed and Github marks them as verified

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #7102, opened 2025-09-30
Remove reference-info-base64 file for CVM

ganeshkumarashok:
**What type of PR is this?**
/kind cleanup

This is no longer needed. VHD builds fail if we don't remove this. Related PR: https://github.com/Azure/AgentBaker/pull/7044/files

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -49,7 +49,6 @@ steps:
         **/kata-containers-cc.img**
         **/kata-containers.img**
         **/kata-containers-initrd-base.img**
-        **/reference-info-base64**
       downloadPath: $(Build.SourcesDirectory)
 
   - task: DownloadPipelineArtifact@2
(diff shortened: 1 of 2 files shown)


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -49,7 +49,6 @@ steps:
         **/kata-containers-cc.img**
         **/kata-containers.img**
         **/kata-containers-initrd-base.img**
-        **/reference-info-base64**
```
Bickor, line 52: This task should no longer be built anyway since we have frozen the kata-cc image and set the default build to false for that VHD.

Bickor: I'm ok with either removing it or leaving it as is, but this change should not have any effect.

ganeshkumarashok: So this is not being referenced right? We should clean it up then.

PR #7104, opened 2025-09-30
Add a pause type command to windows nanoserver and server containers in e2e tests

timmy-wright:
**What type of PR is this?**

/kind flake

Currently when we startup the windows nanoserver and servercore containers in the VHD to check they start, we don't have a pause type command. Which means the pod completes, then gets restarted by the daemonset. 

So, on the odd occasion, we the pod is never in a "running" state when we check it and the test fails.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -36,9 +36,9 @@ func ValidatePodRunning(ctx context.Context, s *Scenario, pod *corev1.Pod) {
 
 	_, err = kube.WaitUntilPodRunning(ctx, pod.Namespace, "", "metadata.name="+pod.Name)
 	if err != nil {
-		jsonString, err := json.Marshal(pod)
-		if err != nil {
-			jsonString = []byte(err.Error())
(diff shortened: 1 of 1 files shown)


e2e/kube.go
```diff
@@ -521,8 +521,11 @@ func podWindows(s *Scenario, podName string, imageName string) *corev1.Pod {
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{
 				{
-					Name:  podName,
-					Image: imageName,
+					Name:            podName,
+					Image:           imageName,
+					ImagePullPolicy: "IfNotPresent",
+					// this should exist on both servercore and nanoserve
+					Command: []string{"cmd", "/c", "ping", "-t", "localhost"},
```
djsly, line 528: This is unlimited ping right ?

PR #7106, opened 2025-10-01
fix(cse): verifies oras login success without list repositories

### LINKED ISSUE #7105: [CSE] Overchecked access led to false access denial in `oras repo ls`
**What occurred**:  
During the node bootstrapping process, the used registry is authenticated first before images are pulled. If an ABAC-enabled ACR is used and the identity only has the `acrpull` role assigned, node bootstrapping will fail with errors such as:  

```console
Error response from registry: could not list repositories for \"$REGISTRY_NAME\"\n" "unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information.: [map[Action:* Name:catalog Type:registry]]'
```

**What was expected to happen**:  
An identity with the `acrpull` role should have the adequant permissions to pull images. There should be no additional checks on repository listing that result in false negatives.  

**Steps to reproduce**:  
- Create an ACR and assign the `acrpull` role to the bootstrap identity  
- Enable ABAC for the ACR  
- Bootstrap a new node

See ACR public documentation on [how ABAC effects existing role assignments](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-rbac-abac-repository-permissions?tabs=azure-portal#effect-on-existing-role-assignments)

**Environment**:

- AgentBaker version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

qweeah:
**What type of PR is this?**
/kind bug

This PR checks login success via checking NI package existence.

Fixes #7105 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -386,20 +386,31 @@ retrycmd_oras_login() {
     return $exit_code
 }
 
-retrycmd_can_oras_ls_acr() {
-    retries=$1; wait_sleep=$2; url=$3
+
+retrycmd_can_oras_fetch_image() {
+    retries=$1; wait_sleep=$2; registry_host=$3
+    
+    # List of image references to try fetching
+    local image_refs=(
+        "oss/binaries/kubernetes/kubernetes-node:v1.34.0-rc.1-linux-amd64"
+        "aks/packages/kubernetes/kubectl:v1.34.0"
+    )
+    
     for i in $(seq 1 $retries); do
-        output=$(timeout 60 oras repo ls "$url" --registry-config "$ORAS_REGISTRY_CONFIG_FILE" 2>&1)
-        if [ "$?" -eq 0 ]; then
-            echo "acr is reachable"
-            return 0
-        fi
-        # shellcheck disable=SC3010
-        if [[ "$output" == *"unauthorized: authentication required"* ]]; then
-            echo "ACR is not reachable: $output"
-            return 1
+        for image_ref in "${image_refs[@]}"; do
+            local full_image="${registry_host}/${image_ref}"
+            output=$(timeout 60 oras manifest fetch "$full_image" --registry-config "$ORAS_REGISTRY_CONFIG_FILE" --descriptor 2>&1)
+            if [ "$?" -eq 0 ]; then
+                echo "$registry_host is reachable"
+                return 0
+            fi
+        done
+        
+        if [ $i -lt $retries ]; then
+            sleep $wait_sleep
         fi
     done
+    
     echo "unexpected response from acr: $output"
     return $ERR_ORAS_PULL_NETWORK_TIMEOUT
 }
@@ -983,7 +994,7 @@ oras_login_with_kubelet_identity() {
         return
     fi
 
-    retrycmd_can_oras_ls_acr 10 5 $acr_url
+    retrycmd_can_oras_fetch_image 10 5 $acr_url
     ret_code=$?
     if [ "$ret_code" -eq 0 ]; then
         echo "anonymous pull is allowed for acr '$acr_url', proceeding with anonymous pull"
@@ -1032,7 +1043,7 @@ oras_login_with_kubelet_identity() {
     unset ACCESS_TOKEN REFRESH_TOKEN  # Clears sensitive data from memory
     set -x
 
-    retrycmd_can_oras_ls_acr 10 5 $acr_url
+    retrycmd_can_oras_fetch_image 10 5 $acr_url
     if [ "$?" -ne 0 ]; then
         echo "failed to login to acr '$acr_url', pull is still unauthorized"
         return $ERR_ORAS_PULL_UNAUTHORIZED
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -386,20 +386,31 @@ retrycmd_oras_login() {
...
+    retries=$1; wait_sleep=$2; registry_host=$3
+    
+    # List of image references to try fetching
+    local image_refs=(
+        "oss/binaries/kubernetes/kubernetes-node:v1.34.0-rc.1-linux-amd64"
+        "aks/packages/kubernetes/kubectl:v1.34.0"
+    )
+    
     for i in $(seq 1 $retries); do
-        output=$(timeout 60 oras repo ls "$url" --registry-config "$ORAS_REGISTRY_CONFIG_FILE" 2>&1)
```
cameronmeissner, line 422: I think the current logic makes more sense, what if these "dummy" images don't exist?

I will admit though, instead of using a regex to check the response string we should strictly be looking at the HTTP status code to detect unauthorized (401)

cameronmeissner: ah I just read the linked issue, I guess `acrpull` doesn't give the principal access to `ls` against the registry?

cameronmeissner: I just don't like the idea of using a set of "dummy" images since it feels quite brittle

qweeah: >  I guess acrpull doesn't give the principal access to ls against the registry?

Exactly. The main point is that listing a repository involves a metadata read, whereas pulling images involves a data read, and these are distinct actions as defined in ACR. It is **incorrect** to infer data read authorization success through metadata read authorization.

qweeah: I have updated the implementation to: 

1) explicitly use `oras repo ls` for ACR anonymous pulls, as anonymous pulls do not involve ABAC, making this approach still valid; 

2) Fast fail when access is insufficient by checking the `.permissions.actions` in the retrieved token. If `read` is missing, the process fails immediately without attempting login. This action check helps to quickly reject cases where login succeeds with non-read RBAC roles such as `acrDelete`.

In the case of an ABAC token with insufficient access, a separate issue will be created to enhance the error message returned.

fseldow (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -386,17 +386,20 @@ retrycmd_oras_login() {
...
 
-retrycmd_can_oras_ls_acr() {
-    retries=$1; wait_sleep=$2; url=$3
+retrycmd_can_oras_ls_acr_anonymously() {
+    retries=$1; wait_sleep=$2; acr_url=$3
+
     for i in $(seq 1 $retries); do
+        # Logout first to ensure insufficient ABAC token won't affect anonymous judging
+        oras logout "$acr_url" --registry-config "${ORAS_REGISTRY_CONFIG_FILE}" 2>/dev/null || true
         output=$(timeout 60 oras repo ls "$url" --registry-config "$ORAS_REGISTRY_CONFIG_FILE" 2>&1)
```
fseldow: the local variable is changed to `acr_url`, should update the `url` to `acr_url`

oras logout privateace2enonanonpullwestus3.azurecr.io --registry-config /etc/oras/config.yaml
++ timeout 60 oras repo ls '' --registry-config /etc/oras/config.yaml
+ output='Error: invalid reference: invalid registry ""'

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -973,6 +976,44 @@ update_base_url() {
   echo "$initial_url"
```
fseldow, line 1027: sorry i fail to find where defines `initial_url `

parts/linux/cloud-init/artifacts/cse_helpers.sh
qweeah: Good catch. Resolved.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -973,6 +976,44 @@ update_base_url() {
...
+                if [ "$action_exists" != "true" ]; then
+                    echo "Required action '$action' not found in token permissions"
+                    return $ERR_ORAS_PULL_UNAUTHORIZED
+                fi
+            done
+            echo "Token validation passed: all required actions present"
+        else
+            echo "No permissions field found in token. Assuming ABAC token, skipping permission validation"
+        fi
+    fi
```
djsly, line 1064: if no token, should we echo something ?

qweeah: @djsly Thanks for reviewing. The information provided here assumes it is an ABAC token and the fast-fail validation is bypassed.

Some further explanation: 

For ACR, access on the registry is presented in RBAC tokens like below:

```
{
...
  "permissions": {
     "actions": [
       "read"
     ],
     "notActions": null
  },
...
```

If this field is missing, it means ABAC is enabled and authorization cannot be performed since the scope of the access is not defined. For ACR, the scope refers to the repository. At this login stage, we only have the registry host name.

cameronmeissner (approved):

PR #7112, opened 2025-10-01
cleanup(stls): cleanup secure TLS bootstrapping configuration options in NodeBootstrappingConfiguration/AKSNodeConfig

cameronmeissner:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


aks-node-controller/parser/parser.go
```diff
@@ -32,146 +32,149 @@ func executeBootstrapTemplate(inputContract *aksnodeconfigv1.Configuration) (str
...
+		"CUSTOM_SEARCH_DOMAIN_FILEPATH":                      getCustomSearchDomainFilepath(),
+		"HTTP_PROXY_URLS":                                    config.GetHttpProxyConfig().GetHttpProxy(),
+		"HTTPS_PROXY_URLS":                                   config.GetHttpProxyConfig().GetHttpsProxy(),
+		"NO_PROXY_URLS":                                      getStringifiedStringArray(config.GetHttpProxyConfig().GetNoProxyEntries(), ","),
+		"PROXY_VARS":                                         getProxyVariables(config.GetHttpProxyConfig()),
+		"TLS_BOOTSTRAP_TOKEN":                                getTLSBootstrapToken(config.GetBootstrappingConfig()),
+		"ENABLE_SECURE_TLS_BOOTSTRAPPING":                    fmt.Sprintf("%v", getEnableSecureTLSBootstrap(config.GetBootstrappingConfig())),
+		"SECURE_TLS_BOOTSTRAPPING_DEADLINE":                  getSecureTLSBootstrappingDeadline(config.GetBootstrappingConfig()),
+		"SECURE_TLS_BOOTSTRAPPING_AAD_RESOURCE":              getSecureTLSBootstrappingAADResource(config.GetBootstrappingConfig()),
+		"SECURE_TLS_BOOTSTRAPPING_USER_ASSIGNED_IDENTITY_ID": getSecureTLSBootstrappingUserAssignedIdentityID(config.GetBootstrappingConfig()),
```
Devinwong: If the getter is simply returning a value, actually you can use the default public Getter. E.g., `config.GetBootstrappingConfig().GetSecureTlsBootstrappingAadResource()`.
In general, only when you want to add extra logic will you need to create a private getter.

r2k1: Agree, don't see much value in the new helpers functions

cameronmeissner: fair enough, I noticed we were doing this bootstrap token as well which is why I did it this way

aks-node-controller/proto/aksnodeconfig/v1/bootstrapping_config.proto
```diff
@@ -41,9 +41,22 @@ message BootstrappingConfig {
...
+  // Only used when secure TLS bootstrapping is enabled. This is the client ID of the user-assigned identity ID the node will use to perform secure TLS bootstrapping.
+  optional string secure_tls_bootstrapping_user_assigned_identity_id = 7;
+
+  // Only used when secure TLS bootstrapping is enabled. If specified, the bootstrap client installation will be replaced with the client version downloaded from this URL.
+  optional string secure_tls_bootstrapping_custom_client_download_url = 8;
+
+  // Only used when secure TLS bootstrapping is enabled. This is the bootstrapping deadline used to perform secure TLS bootstrapping.
+  // If the deadline is hit, the TLS bootstrap token will be used as a fall-back.
+  // This field will be deprecated once TLS bootstrap tokens are no longer used.
+  optional string secure_tls_bootstrapping_deadline = 9;
```
Devinwong, line 61: Note that when it is deprecated, the variable will still be in aksnodeconfig but you can remove the logic in your codes, to avoid breaking the JSON parsing.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -182,6 +182,7 @@ ORAS_REGISTRY_CONFIG_FILE=/etc/oras/config.yaml # oras registry auth config file
 # used by secure TLS bootstrapping to request AAD tokens - uniquely identifies AKS's Entra ID application.
 # more details: https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#how-to-use-kubelogin-with-aks
 AKS_AAD_SERVER_APP_ID="6dae42f8-4368-4678-94ff-3960e28e3630"
+DEFAULT_SECURE_TLS_BOOTSTRAPPING_DEADLINE="2m0s"
```
r2k1: Why define value in a separate file? Wouldn't it be easier to just include in `configureAndStartSecureTLSBootstrapping`?

aks-node-controller/parser/parser.go
cameronmeissner: addressed

aks-node-controller/proto/aksnodeconfig/v1/bootstrapping_config.proto
cameronmeissner: yep, noted

parts/linux/cloud-init/artifacts/cse_helpers.sh
cameronmeissner: was thinking the same

cameronmeissner: moved

PR #7120, opened 2025-10-02
Fail fast to unblock aks-node-controller provision-wait from waiting until timeout

Devinwong:
**What type of PR is this?**
/kind feature

Fail fast to unblock aks-node-controller provision-wait from waiting until timeout.
Now if the aksnodeconfig is invalid, it will fail immediately.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/aks-node-controller/README.md
+++ b/aks-node-controller/README.md
@@ -134,7 +134,12 @@ Key components:
 1. `aks-node-controller.service`: systemd unit that is triggered once cloud-init is complete (guaranteeing that config is present on disk) and then kickstarts bootstrapping.
 2. `aks-node-controller` go binary with two modes:
 
-- **provision**: parses the node config and triggers bootstrap process.
-  - `aks-node-controller` will try its best to parse `aksnodeconfigv1`, even with some unexpected situations, e.g., unknown fields in the config, unknown enum values, etc. This is to support backward compatibility where `aksnodeconfigv1` is newer than the one used in `aks-node-controller` in an older VHD node image.
-  - If `aks-node-controller` can't handle the `aksnodeconfigv1` properly, e.g., the config version is different from the expected version, it will return early by creating the `provision.complete` to notify the `provision-wait` to stop waiting and get the provision error. 
+- **provision**: Parses the node configuration and starts the bootstrap sequence.
+    - The controller performs a tolerant (forward‑compatible) parse of `aksnodeconfigv1.Configuration`: unknown fields, additional enum values, or future‑version knobs are ignored (and may be logged) so that a newer control‑plane can talk to an older VHD image.
+    - If the config cannot be safely interpreted (e.g. unsupported `Version`, malformed required field, or incompatible schema change), the controller fails fast. It writes the sentinel file `provision.complete` early so the `provision-wait` process stops polling and can surface an error instead of hanging indefinitely.
+    - In a fail‑fast path the normal bootstrap scripts never run, therefore `provision.json` (which would contain the serialized `CSEStatus`) is never created. A typical error looks like:
+        ```
+        failed to read provision.json: open /var/log/azure/aks/provision.json: no such file or directory. One reason could be that AKSNodeConfig is not properly set.
+        ```
+        This indicates the controller exited before emitting `provision.json`. Most commonly the rendered AKSNodeConfig was missing, had the wrong `Version` (expected `v1`), or was written to the wrong path (`/opt/azure/containers/aks-node-controller-config.json`). Fix the config generation, redeploy, and the bootstrap scripts will then populate `provision.json`.
 - **provision-wait**: waits for `provision.complete` to be present and reads `provision.json` which contains the provision output of type `CSEStatus` and is returned by CSE through capturing stdout.
--- a/aks-node-controller/README.md
+++ b/aks-node-controller/README.md
@@ -88,20 +88,20 @@ Clients need to provide CSE and Custom Data. [nodeconfigutils](pkg/nodeconfiguti
 
 1. Custom Data: Contains base64 encoded bootstrap configuration of type [aksnodeconfigv1.Configuration](pkg/gen/aksnodeconfig/v1) in json format which is placed on the node through cloud-init write directive.
 
-Format:
-```yaml
-#cloud-config
-write_files:
-- path: /opt/azure/containers/aks-node-controller-config.json
-  permissions: "0755"
-  owner: root
-  content: !!binary |
-   {{ encodedAKSNodeConfig }}`
-```
+    Format:
+    ```yaml
+    #cloud-config
+    write_files:
+    - path: /opt/azure/containers/aks-node-controller-config.json
+    permissions: "0755"
+    owner: root
+    content: !!binary |
+    {{ encodedAKSNodeConfig }}`
+    ```
 
 2. CSE: Script used to poll bootstrap status and return exit status once complete.
 
-CSE script: `/opt/azure/containers/aks-node-controller provision-wait`
+   CSE script: `/opt/azure/containers/aks-node-controller provision-wait`
 
 
 #### Provisioning flow diagram:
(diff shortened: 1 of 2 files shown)


aks-node-controller/app.go
```diff
@@ -152,7 +171,7 @@ func (a *App) ProvisionWait(ctx context.Context, filepaths ProvisionStatusFiles)
 			if event.Op&fsnotify.Create == fsnotify.Create && event.Name == filepaths.ProvisionCompleteFile {
 				data, err := os.ReadFile(filepaths.ProvisionJSONFile)
 				if err != nil {
-					return "", fmt.Errorf("failed to read provision.json: %w", err)
+					return "", fmt.Errorf("failed to read provision.json: %w. One reason could be that AKSNodeConfig is not properly set", err)
```
r2k1, line 178: I don't think this extra message is helpful.

Something like:
`fmt.Errorf("failed to read %s: %w", filepaths.ProvisionJSONFile, err)` should be good enough? (Maybe file path already included in the message?)

aks-node-controller/app.go
```diff
@@ -116,16 +119,35 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
...
+		slog.Error("failed to write provision.complete file", "error", writeErr)
+	}
+}
+
 func (a *App) ProvisionWait(ctx context.Context, filepaths ProvisionStatusFiles) (string, error) {
 	if _, err := os.Stat(filepaths.ProvisionCompleteFile); err == nil {
 		data, err := os.ReadFile(filepaths.ProvisionJSONFile)
 		if err != nil {
-			return "", fmt.Errorf("failed to read provision.json: %w", err)
+			return "", fmt.Errorf("failed to read provision.json: %w. One reason could be that AKSNodeConfig is not properly set", err)
```
r2k1, line 153: same

aks-node-controller/app.go
```diff
@@ -116,16 +119,35 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
...
-	// Is it ok to log a single line? Is it too much?
 	slog.Info("CSE finished", "exitCode", exitCode, "stdout", stdoutBuf.String(), "stderr", stderrBuf.String(), "error", err)
 	return err
 }
 
+// notifyProvisionFailure creates the provision.complete sentinel file if a provisioning
+// error occurred before the normal provisioning scripts had a chance to write it.
+// This prevents the provision-wait mode from blocking until timeout when we fail fast
+// (e.g. due to invalid or unsupported configuration).
+func (a *App) notifyProvisionFailure(runErr *error) {
```
r2k1: I think something like `writeCompleteFileOnError()` makes more sense

aks-node-controller/app.go
```diff
@@ -116,16 +119,35 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) error {
...
+	}
+	// Avoid clobbering an existing file if provisioning scripts actually created it before the error surfaced.
+	if _, statErr := os.Stat(provisionCompleteFilePath); statErr == nil {
+		return // file already exists
+	} else if !errors.Is(statErr, os.ErrNotExist) { // unexpected stat error
+		slog.Error("failed to stat provision.complete file", "error", statErr)
+		return
+	}
+	if writeErr := os.WriteFile(provisionCompleteFilePath, []byte{}, 0600); writeErr != nil {
+		slog.Error("failed to write provision.complete file", "error", writeErr)
```
r2k1: I would probably also log `provisionCompleteFilePath`

aks-node-controller/README.md
```diff
@@ -88,20 +88,20 @@ Clients need to provide CSE and Custom Data. [nodeconfigutils](pkg/nodeconfiguti
...
+    Format:
+    ```yaml
+    #cloud-config
+    write_files:
+    - path: /opt/azure/containers/aks-node-controller-config.json
+    permissions: "0755"
+    owner: root
+    content: !!binary |
+    {{ encodedAKSNodeConfig }}`
+    ```
```
r2k1, line 100: unnecessary spacing?

aks-node-controller/app.go
```diff
@@ -58,27 +58,30 @@ func (a *App) run(ctx context.Context, args []string) error {
...
+		var err error
+		// Ensure provision-wait is unblocked if we error before provisioning scripts create provision.complete
+		defer a.notifyProvisionFailure(&err)
 		fs := flag.NewFlagSet("provision", flag.ContinueOnError)
 		provisionConfig := fs.String("provision-config", "", "path to the provision config file")
 		dryRun := fs.Bool("dry-run", false, "print the command that would be run without executing it")
-		err := fs.Parse(args[2:])
-		if err != nil {
-			return fmt.Errorf("parse args: %w", err)
+		if parseErr := fs.Parse(args[2:]); parseErr != nil {
```
r2k1: This is confusing. Why `parseErr` is used here instead of `err`?

aks-node-controller/app.go
```diff
@@ -58,27 +58,30 @@ func (a *App) run(ctx context.Context, args []string) error {
 	}
 	switch args[1] {
 	case "provision":
+		var err error
+		// Ensure provision-wait is unblocked if we error before provisioning scripts create provision.complete
+		defer a.notifyProvisionFailure(&err)
```
r2k1: I'm not sure if using `defer` here is the best approach. Seems easy to make a mistake by something like `return a.Provision(ctx, ProvisionFlags{ProvisionConfig: *provisionConfig})`

Maybe a more robust solution would be extracting a function returning an error? And then handle the error?

aks-node-controller/app.go
```diff
@@ -134,12 +156,9 @@ func (a *App) ProvisionWait(ctx context.Context, filepaths ProvisionStatusFiles)
 	if err != nil {
 		return "", fmt.Errorf("failed to create watcher: %w", err)
 	}
-	defer watcher.Close()
```
r2k1, line 147: Why it's gone?

aks-node-controller/app.go
Devinwong: The original message was: `failed to read provision.json: open /var/log/azure/aks/provision.json: no such file or directory`. It doesn't indicate it's the aksnodeconfig issue. (In this case, I intentionally set the version to a wrong `v2`). That's why I added an extra message to help debugging. The actual error could be found in aks-node-controller.log for this case.

aks-node-controller/app.go
Devinwong: Oh it's an empty file. Just an indicator for provision complete

aks-node-controller/README.md
Devinwong: It's to make the indent of the code snippets in readme correct.
image

aks-node-controller/app.go
r2k1: When we can't write to the file it can be useful for the user to know what file we're trying to write

aks-node-controller/app.go
Devinwong: updated

aks-node-controller/app.go
Devinwong: updated

aks-node-controller/app.go
Devinwong: That's a good call out. I've updated the pattern in case people will forget to always assign error to `err`.

aks-node-controller/app.go
Devinwong: oh who did it. Added it back

aks-node-controller/app.go
Devinwong: oh you mean log the filepath. I thought you were saying the content. Updated.

aks-node-controller/app.go
Devinwong: btw, provision.json is provision scripts output, not config.json (which is the aksnodeconfig)

Devinwong: Later I will write a wiki to help debugging

r2k1 (approved):

aks-node-controller/app.go
```diff
@@ -58,27 +58,17 @@ func (a *App) run(ctx context.Context, args []string) error {
...
+		return err
 	case "provision-wait":
 		provisionStatusFiles := ProvisionStatusFiles{ProvisionJSONFile: provisionJSONFilePath, ProvisionCompleteFile: provisionCompleteFilePath}
-		provisionOutput, err := a.ProvisionWait(ctx, provisionStatusFiles)
+		provisionOutput, waitErr := a.ProvisionWait(ctx, provisionStatusFiles)
 		//nolint:forbidigo // stdout is part of the interface
 		fmt.Println(provisionOutput)
 		slog.Info("provision-wait finished", "provisionOutput", provisionOutput)
-		return err
+		return waitErr
```
r2k1: I guess theere is no need to rename it. Though I'm not sure if it's worth another round of approval

Devinwong: Originally there were different errors in the switch block so it's clearer to name them differently. But now it's not needed. Let me update it.

PR #7125, opened 2025-10-03
feat: install most components to /opt/bin instead of /usr/local/bin

chewi:
**What type of PR is this?**

/kind feature

/usr/local/bin is generally read-only on immutable distributions, so binaries cannot be installed there. This PR installs them to /opt/bin instead.
    
The packages published to PMC and the system extensions built from those install their binaries to /usr/bin. These were previously moved to /usr/local/bin, but since this is no longer possible, symlinks are now created in /opt/bin instead.
    
Symlinks are not used when installing from a tarball because the versioned binaries (e.g. kubelet-*) are removed, leading to dangling links.
    
The `install` command has been used to write these binaries because it removes the existing file first (rather than following a symlink) and takes care of making the new file executable.

~~As an aside, this also fixes a failing test that was introduced in #6995.~~

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

This change is part of a wider effort to install binaries to Flatcar from systemd system extensions (sysexts) rather than binary tarballs, which are being phased out. There are too many changes to submit everything in this one PR, so another will follow this one.

I could not find any relevant documentation to update.

Regarding upgrades, I am not experienced enough with AKS to know how to test that. I don't believe the changes would cause any issue for upgrades because they only change how and where things are installed. There are practically no configuration changes besides some paths in the systemd units.

**Release note**:

Is one needed here? I don't know whether end users will see or care about these changes.


parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -204,7 +204,8 @@ installRPMPackageFromFile() {
     if ! dnf_install 30 1 600 ${rpmFile}; then
         exit $ERR_APT_INSTALL_TIMEOUT
     fi
-    mv "/usr/bin/${packageName}" "/usr/local/bin/${packageName}"
+    mkdir -p /opt/bin
+    ln -snf "/usr/bin/${packageName}" "/opt/bin/${packageName}"
```
Devinwong, line 449: Does the requirement only retain components from installing to /usr/local/bin? If they are in /usr/bin, it's fine?

parts/common/components.json
```diff
@@ -1457,7 +1457,6 @@
     },
     {
       "name": "nvidia-container-toolkit",
-      "downloadLocation": "/usr/local/bin",
```
Devinwong, line 1613: Removed unintentionally?
@ganeshkumarashok to see if additional test is needed in content tests.

parts/common/components.json
```diff
@@ -853,7 +853,7 @@
     },
     {
       "name": "oras",
-      "downloadLocation": "/usr/local/bin",
```
Devinwong, line 937: If oras is not working, airgapped e2e test will catch it when downloading some components from MAR.

parts/common/components.json
```diff
@@ -870,7 +870,7 @@
     },
     {
       "name": "aks-secure-tls-bootstrap-client",
-      "downloadLocation": "/usr/local/bin",
+      "downloadLocation": "/opt/bin",
```
Devinwong, line 954: @cameronmeissner to confirm

parts/common/components.json
chewi: Not unintentional. As mentioned in that commit, it is only installed using dnf, so I dropped `downloadLocation` to avoid confusion. As you might imagine, I was scouring the codebase for instances of `/usr/local/bin`.

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
chewi: The point is that immutable distros cannot modify anything under /usr. This particular function applies to Mariner, which is not immutable, so it can install packages from PMC, which put the binaries in /usr/bin. It previously moved those binaries to /usr/local/bin for consistency with the others. Now it's symlinking them to /opt/bin instead. I don't like moving them because that's changing things behind the package manager's back.

parts/common/components.json
chewi: Yes, and Flatcar will rely on oras to fetch sysexts in the upcoming PR. If this fails at build time, it is considered fatal.

e2e/kubelet/generate-kubelet-flags.sh
```diff
@@ -29,7 +29,7 @@ for KUBE_BINARY_VERSION in $KUBE_BINARY_VERSIONS; do
     K8S_TGZ_TMP=${KUBE_BINARY_URL##*/}
     retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" ${KUBE_BINARY_URL} || exit 120
     tar --transform="s|.*|&-${KUBE_BINARY_VERSION}|" --show-transformed-names -xzvf "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" \
-        --strip-components=3 -C /usr/local/bin kubernetes/node/bin/kubelet kubernetes/node/bin/kubectl
+        --strip-components=3 -C /opt/bin kubernetes/node/bin/kubelet kubernetes/node/bin/kubectl
```
cameronmeissner, line 32: yeah... we're definitely not using this anymore

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -569,11 +569,10 @@ installKubeletKubectlFromURL() {
             fi
         fi
     fi
-    mv "/usr/local/bin/kubelet-${KUBERNETES_VERSION}" "/usr/local/bin/kubelet"
-    mv "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" "/usr/local/bin/kubectl"
+    install -m0755 "/opt/bin/kubelet-${KUBERNETES_VERSION}" /opt/bin/kubelet
```
cameronmeissner, line 638: I'm assuming `install` is available on all our linux distros

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -166,7 +166,7 @@ installCredentialProviderFromPMC() {
     mkdir -p "${CREDENTIAL_PROVIDER_BIN_DIR}"
     chown -R root:root "${CREDENTIAL_PROVIDER_BIN_DIR}"
     installRPMPackageFromFile "azure-acr-credential-provider" "${packageVersion}" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
-    mv "/usr/local/bin/azure-acr-credential-provider" "$CREDENTIAL_PROVIDER_BIN_DIR/acr-credential-provider"
+    ln -snf /usr/bin/azure-acr-credential-provider "$CREDENTIAL_PROVIDER_BIN_DIR/acr-credential-provider"
```
cameronmeissner, line 192: why changing from moving to linking?

cameronmeissner: nvm, I see your response to Devin's comment

parts/linux/cloud-init/artifacts/cse_install.sh
chewi: Yes, it's part of coreutils.

parts/common/components.json
```diff
@@ -934,7 +934,7 @@
     },
     {
       "name": "oras",
-      "downloadLocation": "/usr/local/bin",
+      "downloadLocation": "/opt/bin",
```
djsly, line 937: shouldnt this be installLocation, I don't really like that we download temp pacakges to /opt/bin ? maybe it's just me.

parts/linux/cloud-init/artifacts/azlosguard/cse_install_osguard.sh
```diff
@@ -13,7 +13,7 @@ installKubeletKubectlPkgFromPMC() {
 installRPMPackageFromFile() {
     local packageName="${1}"
     local desiredVersion="${2}"
-    local targetBinDir="${3:-"/usr/local/bin"}"
+    local targetBinDir="${3:-"/opt/bin"}"
```
djsly, line 16: @hbeberman mind reviwing changes to that file ?

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
```diff
@@ -79,7 +79,7 @@ rm /home/packer
 mkdir -p /_imageconfigs/out
 
 echo "kubelet/kubectl downloaded:" >> ${VHD_LOGS_FILEPATH}
-ls -ltr /usr/local/bin/* >> ${VHD_LOGS_FILEPATH}
+ls -ltr /opt/bin/* >> ${VHD_LOGS_FILEPATH}
```
djsly: why not using `/opt/bin/kube*` like the other place ?

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -1,5 +1,14 @@
 #cloud-config
 
+bootcmd:
```
djsly, line 3: all OS flavours will be getting this ?

parts/linux/cloud-init/artifacts/profile-d-path.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
djsly, line 1: I cannot find where this file is called.

parts/common/components.json
chewi: We don't, `installOras` downloads to /opt/oras/downloads and installs to the above. The naming is weird, but I haven't changed how it works.

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
chewi: I wondered that myself, but I was just preserving what it did before. Happy to change it.

parts/linux/cloud-init/nodecustomdata.yml
chewi: Yes. Flatcar doesn't natively support this, so the Go changes translate it to Ignition.

parts/linux/cloud-init/artifacts/profile-d-path.sh
chewi: It's installed to /etc/profile.d/path.sh, which is typically picked up by all login shells. See /etc/profile. It's not used by systemd units, which is why we need to explicitly reference /opt/bin in those.

parts/common/components.json
cameronmeissner: this should be fine

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -1,5 +1,14 @@
...
+- |
+  mkdir -p /opt/bin
+  for bin in aks-secure-tls-bootstrap-client ci-syslog-watcher.sh logrotate.sh teleportd; do
```
cameronmeissner, line 6: this is just to ensure that older VHDs will be compatible with the changes - e.g. creating the required symlinks at node provisioning time?

pkg/agent/baker.go
```diff
@@ -147,6 +147,17 @@ func (t *TemplateGenerator) getFlatcarLinuxNodeCustomDataJSONObject(config *data
 		}
 		newfiles = append(newfiles, *newfile)
 	}
+	if len(customData.BootCommands) > 0 {
```
cameronmeissner, line 133: do we have a unit test to cover this?

parts/linux/cloud-init/nodecustomdata.yml
chewi: Yes, the E2Es against the older VHD were almost entirely failing without it.

pkg/agent/baker.go
chewi: No, good call.

chewi: Added one now. I had to split out this part into a separate function to make it more easily testable.

vhdbuilder/packer/imagecustomizer/azlosguard/scripts/azlosguard-postinstall.sh
chewi: I went ahead and did that.

parts/linux/cloud-init/artifacts/azlosguard/cse_install_osguard.sh
hbeberman: Changes in this file look good to me!

YaoC (approved):

PR #7128, opened 2025-10-06
Merge GB200 changes into mainline for regular VHD release support

keith-ms:
**What type of PR is this?**

This PR merges in the support added for building the GB200-specific image. 

/kind feature

Previously, this VHD was built in an ad-hoc manner. So that new GB200 images are built automatically and can be referenced by RP under certain circumstances, add in the GB200 build process to the `master` branch.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests
- [X ] tested upgrade from previous version

Please check the `pkg/agent/testdata` directory. Do I need to regenerate those files? I accepted what was on `master`.

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -338,7 +338,11 @@ ExecStartPost=/sbin/iptables -P FORWARD ACCEPT
 EOF
 
   mkdir -p /etc/containerd
-  if [ "${GPU_NODE}" = "true" ]; then
+
+  # Skip containerd config generation if config already exists
+  if [ -f "/etc/containerd/config.toml" ]; then
+    echo "containerd config already exists at /etc/containerd/config.toml, skipping generation"
```
cameronmeissner, line 346: are we certain we always want to behave this way? generally our containerd config should take precedence since we're managing the node for the customer - I wonder if we should guard this behind a VMSS tag or something

parts/linux/cloud-init/artifacts/ubuntu/containerd-nvidia.toml
```diff
@@ -0,0 +1,39 @@
+oom_score = -999
```
cameronmeissner, line 1: probably don't need the new `ubuntu` sub-folder at this time

parts/linux/cloud-init/artifacts/ubuntu/doca.pub
```diff
@@ -0,0 +1,81 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
cameronmeissner, line 1: is there a way we can download these as needed at build-time instead?

parts/linux/cloud-init/artifacts/ubuntu/nvidia.pub
```diff
@@ -0,0 +1,29 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
cameronmeissner, line 1: same question here

vhdbuilder/packer/packer_source.sh
```diff
@@ -389,6 +389,41 @@ copyPackerFiles() {
 
   cpAndMode $NOTICE_SRC $NOTICE_DEST 444
 
+  # GB200-specific configuration
+  if grep -q "GB200" <<< "$FEATURE_FLAGS"; then
+    # Only applicable to Ubuntu 24.04 and ARM64
+    if [ ${UBUNTU_RELEASE} = "24.04" ] && [ ${CPU_ARCH} = "arm64" ]; then
```
cameronmeissner: we probably don't need these additional checks as you said in your comment within install-dependencies.sh

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -15,7 +15,7 @@ VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
 PERFORMANCE_DATA_FILE=/opt/azure/vhd-build-performance-data.json
 
 # Hardcode the desired size of the OS disk so we don't accidently rely on extra disk space
-MAX_BLOCK_COUNT=30298176 # 30 GB
+MAX_BLOCK_COUNT=60596352 # 60 GB
```
cameronmeissner: unless we know we can safely increase this for _all_ SKUs, we should make this GB200-specific

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
```diff
@@ -40,7 +40,7 @@
       "virtual_network_subnet_name": "{{user `subnet_name`}}",
       "ssh_read_write_timeout": "5m",
       "os_type": "Linux",
-      "os_disk_size_gb": 30,
+      "os_disk_size_gb": 60,
```
cameronmeissner: again, unless we know for this change is okay for all SKUs, we should have a separate packer template that's GB200-specific

parts/linux/cloud-init/artifacts/ubuntu/doca.pub
keith-ms: What are your concerns with checking this into the repository? If the key used to sign packages in the repository changes, hard coded the expected value in the AgentBaker code base means that the VHD build fails if the key changes. This is arguably a way to improve supply chain security. If the key used to sign the packages changes, we can at least confirm with NVIDIA that this happened.

A similar argument can be made about the key currently checked in, but I still think it would be helpful to know when the key changed in the future.

parts/linux/cloud-init/artifacts/ubuntu/nvidia.pub
keith-ms: See above, the same reasoning applies.

vhdbuilder/packer/packer_source.sh
keith-ms: I'll remove the ARM check. since the repositories contain references to both x86_64 and arm64, though the GB200 is only expected to ever have ARM. 

The release check is still valid because we'll conceivably build another image using a different Ubuntu version and the repos are all specific to 24.04.

parts/linux/cloud-init/artifacts/cse_config.sh
keith-ms: I added this because the changes to the `config.toml` file kept getting overwritten whenever the node provisioning would run (as I recall, even with scriptless enabled on the subscription).

Is the GB200 recognized as a GPU SKU yet? If so, then the `if` statement below this conditional will write the correct containerd configuration file.

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
keith-ms: That's going to require splitting out a separate packer file, isn't it? Or perhaps another builder in this list.

parts/linux/cloud-init/artifacts/ubuntu/doca.pub
cameronmeissner: that's a fair argument I suppose, curious what others think as well, I think we just generally try avoid checking public certs/keys into VCS if possible - we get could potentially get the best of both worlds by sticking this in a storage account, though that would be difficult to manage across testing/production tenants...

keith-ms: It's a plaintext file, what's the concern with checking it into a VCS?

parts/linux/cloud-init/artifacts/ubuntu/containerd-nvidia.toml
keith-ms: It keeps the artifacts folder more organized. I can rename it to gb200 if that helps.

vhdbuilder/packer/vhd-image-builder-arm64-gen2.json
keith-ms: Moving this to a different file and reverting the changes to this one. The GB200 will have its own file.

parts/linux/cloud-init/artifacts/ubuntu/containerd-nvidia.toml
keith-ms: The Ubuntu sub-folder already existed.

parts/linux/cloud-init/artifacts/cse_config.sh
abenn135: https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/pullrequest/13110877 was merged back in August so it should be recognized as GPU now. Haven't checked in a while though.

PR #7132, opened 2025-10-06
Fix kubelet flags for VHD Caching

r2k1:
**What type of PR is this?**

/kind bug

With the current bash scripts KUBELET_FLAGS env variable is modified in basePrep and results is writted in nodePrep.
With VHD Caching a reboot is introduced between basePrep and nodePrep phases causing the changes to be lost and result in invalid kubelet configuration.

Move all kubelet logic to nodePrep to avoid such issues.


cameronmeissner (approved):

e2e/node_config.go
```diff
@@ -687,7 +687,7 @@ func baseTemplateLinux(t *testing.T, location string, k8sVersion string, arch st
 			"--protect-kernel-defaults":           "true",
 			"--read-only-port":                    "0",
 			"--resolv-conf":                       "/run/systemd/resolve/resolv.conf",
-			"--rotate-certificates":               "false",
+			"--rotate-certificates":               "true",
```
timmy-wright, line 690: Is this change because aks-rp always sends "true"? I remember you mentioning it but have forgotten.

timmy-wright (approved):

e2e/node_config.go
r2k1: I don't think it's always. But it seems to be a default value that isn't tested properly otherwise

cameronmeissner: in prod RP will always set this to true (unless k8s RBAC is disabled), though in the context of abe2e RP isn't involved whatsoever

e2e/test_helpers.go
```diff
@@ -522,25 +520,6 @@ func createVMExtensionLinuxAKSNode(location *string) (*armcompute.VirtualMachine
 }
 
 func CreateImage(ctx context.Context, s *Scenario) *config.Image {
-	s.T.Log("Generalizing VM")
```
djsly, line 525: why dont we need this anymore ?

e2e/kube.go
```diff
@@ -179,7 +181,7 @@ func (k *Kubeclient) WaitUntilNodeReady(ctx context.Context, t *testing.T, vmssN
 	}
 
 	if node == nil {
-		t.Fatalf("failed to wait for %q to appear in the API server", vmssName)
+		t.Fatalf("ERROR: %q haven't appeared in k8s API server", vmssName)
```
djsly, line 184: can we print the describe ?

e2e/kube.go
```diff
@@ -145,9 +145,11 @@ func (k *Kubeclient) WaitUntilPodRunning(ctx context.Context, namespace string,
 	return pod, err
 }
 
-func (k *Kubeclient) WaitUntilNodeReady(ctx context.Context, t *testing.T, vmssName string) string {
+func (k *Kubeclient) WaitUntilNodeReady(ctx context.Context, t testing.TB, vmssName string) string {
+	startTime := time.Now()
 	var node *corev1.Node = nil
```
cameronmeissner, line 150: nit: redundant `= nil`

e2e/kube.go
cameronmeissner: we can't describe it if we don't have the node object right?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -180,6 +180,7 @@ EOF
 AZURE_JSON_PATH="/etc/kubernetes/azure.json"
 AKS_CUSTOM_CLOUD_JSON_PATH="/etc/kubernetes/${TARGET_ENVIRONMENT}.json"
 configureAzureJson() {
+    mkdir -p "$(dirname "${AZURE_JSON_PATH}")"
```
cameronmeissner, line 183: ah, configureK8s was previously creating this dir, good catch

I'd also consider moving `configureAzureJson` to node prep if there isn't anything else within basePrep after where it's currently called that relies on `/etc/kubernetes.azure.json` or the custom cloud config JSON

e2e/test_helpers.go
r2k1: Command for linux deletes SSH keys, machine name etc. We don't need any of it.

I'm not sure about windows yet. I may put it back at some point, but it's not what I'm using for the prototype, so updating tests to reflect it.

e2e/kube.go
r2k1: I haven't touched it. Maybe next time. Don't want to have another round of approvals.

PR #6714, opened 2025-07-29
feat: install kubelet kubectl packages from pmc

lilypan26:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version
- [ ] commits are GPG signed and Github marks them as verified

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #7134, opened 2025-10-07
feat: consume NI kubernetes packages on MAR

qweeah:
**What type of PR is this?**
/kind feature

This PR consumes Network-Isolated packages kubelet and kubectl binaries on MAR. Current implementation only covers Ubuntu scenario.

Fixes #

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -317,26 +317,38 @@ retrycmd_curl_file() {
...
-                cat $ORAS_OUTPUT
-            fi
+            cat $ORAS_OUTPUT
         fi
     done
 }
 
+retrycmd_get_tarball_from_registry_with_oras() {
+    tar_retries=$1; wait_sleep=$2; tarball=$3; url=$4
+    if [ -f "$tarball" ] && tar -tzf "$tarball" > /dev/null 2>&1; then
```
cameronmeissner, line 350: `tar -tzf "$tarball" > /dev/null 2>&1` will return exit code 0 if it succeeds, which will cause the conditional to evaluate to false incorrectly

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -553,15 +553,47 @@ installKubeletKubectlFromURL() {
...
-                updateKubeBinaryRegistryURL
-                
                 K8S_DOWNLOADS_TEMP_DIR_FROM_REGISTRY="/tmp/kubernetes/downloads" # /opt folder will return permission error
-                logs_to_events "AKS.CSE.installKubeletKubectlFromURL.extractKubeBinaries" extractKubeBinaries ${KUBERNETES_VERSION} "${KUBE_BINARY_REGISTRY_URL:-}" false ${K8S_DOWNLOADS_TEMP_DIR_FROM_REGISTRY}
-                # no egress traffic, default install will fail
-                # will exit if the download fails
+
+                # Try to pull distro-specific packages (e.g., .deb for Ubuntu) from registry
+                install_success=false
+                for TOOL_NAME in kubelet kubectl; do
```
cameronmeissner: would prefer we extract this to a helper func so we can unit test it and to make things more readable

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -290,7 +290,7 @@ evalPackageDownloadURL() {
 
 downloadAzureCNI() {
     mkdir -p ${1-$:CNI_DOWNLOADS_DIR}
-    # At VHD build time, the VNET_CNI_PLUGINS_URL is usually not set. 
+    # At VHD build time, the VNET_CNI_PLUGINS_URL is usually not set.
```
fseldow, line 290: revert auto format for unrelated part?

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -553,15 +553,47 @@ installKubeletKubectlFromURL() {
...
-                
                 K8S_DOWNLOADS_TEMP_DIR_FROM_REGISTRY="/tmp/kubernetes/downloads" # /opt folder will return permission error
-                logs_to_events "AKS.CSE.installKubeletKubectlFromURL.extractKubeBinaries" extractKubeBinaries ${KUBERNETES_VERSION} "${KUBE_BINARY_REGISTRY_URL:-}" false ${K8S_DOWNLOADS_TEMP_DIR_FROM_REGISTRY}
-                # no egress traffic, default install will fail
-                # will exit if the download fails
+
+                # Try to pull distro-specific packages (e.g., .deb for Ubuntu) from registry
+                install_success=false
+                for TOOL_NAME in kubelet kubectl; do
+                    tool_package_url="${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}/aks/packages/kubernetes/${TOOL_NAME}:${KUBERNETES_VERSION}"
```
fseldow: for package change, can we introduce new function or put logic in installKubeletKubectlPkgFromPMC.
It seems installUrl is the old method.

can refer to
https://github.com/Azure/AgentBaker/pull/6714/files#diff-4f9504893023c819ba00993d7db2a80f9883ef64f249e6e61b06d18094f3d425

parts/linux/cloud-init/artifacts/cse_helpers.sh
qweeah: Nice catch

qweeah: After struggling for a while I found that status 0 means `true` actually, run below code to find out:

```bash
# Example 1: Using the 'true' command
# The 'true' command always returns 0
true
echo "Exit code of 'true': $?"  # Prints: 0

if true; then
    echo "✓ This prints because 0 = true"
fi

# Example 2: Using the 'false' command
# The 'false' command always returns 1
false
echo "Exit code of 'false': $?"  # Prints: 1

if false; then
    echo "✗ This does NOT print because 1 = false"
fi
```

I would go with current implementation. I have added test cases to cover it.

parts/linux/cloud-init/artifacts/cse_install.sh
fseldow: for k8s < 1.34 we still use the old path because not all versions < 1.34.0 are published in pmc

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -101,4 +101,56 @@ installDebPackageFromFile() {
...
+    if ! apt-get ${opts} update 2>&1; then
+        echo "Failed to update apt cache from local repo ${local_repo_dir}"
+        rm -f "${tmp_list}"
+        rmdir "${tmp_dir}"
+        return 1
+    fi
+
+    # Install package from local repo (no download needed)
+    export DEBIAN_FRONTEND=noninteractive
+    if ! apt-get ${opts} install -y "${package_name}"; then
```
fseldow: can we share same download script with normal pmc installation? and put creating localrepo and cache update into download logic?

parts/linux/cloud-init/artifacts/cse_install.sh
qweeah: It is done by auto linter run on every save operation. Can we allow this? I need to manually revert the change on every save operation.

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
qweeah: Added reusable base helper functions that could be utilized by both PMC and MAR Distro package artifacts.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -529,6 +529,42 @@ extractKubeBinaries() {
     extractKubeBinariesToUsrLocalBin "${k8s_tgz_tmp}" "${k8s_version}" "${is_private_url}"
 }
 
+installK8sToolsFromProfileRegistry() {
```
fseldow: BootstrapRegistry or BootstrapProfileRegistry

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -554,9 +554,15 @@ EOF
 }
 
 configureKubeletAndKubectl() {
-    # Install kubelet and kubectl binaries from URL for Network Isolated, Custom Kube binary, and Private Kube binary
-    if [ -n "${CUSTOM_KUBE_BINARY_DOWNLOAD_URL}" ] || [ -n "${PRIVATE_KUBE_BINARY_DOWNLOAD_URL}" ] || [ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ]; then
+    # Install kubelet and kubectl binaries from URL for Custom Kube binary and Private Kube binary
+    if [ -n "${CUSTOM_KUBE_BINARY_DOWNLOAD_URL}" ] || [ -n "${PRIVATE_KUBE_BINARY_DOWNLOAD_URL}" ]; then
         logs_to_events "AKS.CSE.configureKubeletAndKubectl.installKubeletKubectlFromURL" installKubeletKubectlFromURL
+    elif [ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ]; then
```
fseldow: can we put this condition after the generate version check?
```
elif [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && ! semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.34.0"; then
        logs_to_events "AKS.CSE.configureKubeletAndKubectl.installKubeletKubectlFromURL" installKubeletKubectlFromURL
```

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -554,9 +554,15 @@ EOF
...
+        # For network isolated clusters, try distro packages for k8s >= 1.34.0, otherwise use binary installation
+        if ! semverCompare ${KUBERNETES_VERSION} "1.34.0" || \
```
fseldow: the e2e does not have 1.34.0 for now, we may need tag like SHOULD_ENFORCE_KUBE_PMC_INSTALL to enforce the e2e coverage

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -529,6 +529,42 @@ extractKubeBinaries() {
...
+        # retrycmd_pull_from_registry_with_oras will pull all artifacts to the directory with platform selection
+        if ! retrycmd_pull_from_registry_with_oras 10 5 "${tool_download_dir}" "${tool_package_url}" ${platform_flag}; then
+            echo "Failed to pull ${tool_name} package from registry"
+            return 1
+        fi
+
+        echo "Successfully pulled ${tool_name} package"
+
+        # Try to install using distro-specific package installer from local repo
+        if ! installKubeletKubectlPkgFromLocalRepo "${tool_name}" "${tool_download_dir}"; then
```
fseldow: if install successfully, should we remove the local repo/cache and download packages?

e2e/scenario_test.go
```diff
@@ -1941,6 +1941,7 @@ func Test_Ubuntu2204_PMC_Install(t *testing.T) {
 					vmss.Tags = map[string]*string{}
 				}
 				vmss.Tags["ShouldEnforceKubePMCInstall"] = to.Ptr("true")
+				vmss.Tags["ShouldEnforceBootstrapRegistryInstall"] = to.Ptr("true")
```
fseldow: do you think we could reuse `ShouldEnforceKubePMCInstall` for ni?

qweeah: If we proceed with that, we’ll also need to set `ShouldEnforceKubePMCInstall` to false for all the situations where the NI package isn’t forced, which would be quite complicated.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -325,26 +325,38 @@ retrycmd_curl_file() {
...
+    extra_flags="$*"
+    echo "${pull_retries} retries"
+    for i in $(seq 1 $pull_retries); do
+        if [ "$i" -eq "$pull_retries" ]; then
             return 1
+        fi
+        if [ "$i" -gt 1 ]; then
+            sleep $wait_sleep
+        fi
+        timeout 60 oras pull $url -o $target_folder --registry-config ${ORAS_REGISTRY_CONFIG_FILE} $extra_flags > $ORAS_OUTPUT 2>&1
```
fseldow: + timeout 60 oras pull privateace2enonanonpullwestus3.azurecr.io/aks/packages/kubernetes/kubelet:v1.32.7 -o /tmp/kubernetes/downloads/kubelet --registry-config /etc/oras/config.yaml '"--platform=linux/amd64:ubuntu' '22.04"'
+ '[' 1 -eq 0 ']'
+ cat /tmp/oras_verbose.out
Error: "oras pull" requires exactly 1 argument but got 3
Usage: oras pull [flags] <name>{:<tag>|@<digest>}
Please specify exactly 1 argument as the artifact reference you want to pull. Run "oras pull -h" for more options and examples

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -199,6 +199,32 @@ installKubeletKubectlPkgFromPMC() {
     installPkgWithAptGet "kubectl" "${k8sVersion}" || exit $ERR_KUBECTL_INSTALL_FAIL
 }
 
+installKubeletKubectlPkgFromLocalRepo() {
```
fseldow: the tool_name is in argument.
The function name should not include kubelet and kubectl

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -554,13 +554,22 @@ EOF
...
+    # Install kubelet and kubectl binaries from URL for Custom Kube binary and Private Kube binary
+    if [ -n "${CUSTOM_KUBE_BINARY_DOWNLOAD_URL}" ] || [ -n "${PRIVATE_KUBE_BINARY_DOWNLOAD_URL}" ]; then
         logs_to_events "AKS.CSE.configureKubeletAndKubectl.installKubeletKubectlFromURL" installKubeletKubectlFromURL
     # only install kube pkgs from pmc if k8s version >= 1.34.0 or skip_bypass_k8s_version_check is true
     elif [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && ! semverCompare ${KUBERNETES_VERSION:-"0.0.0"} "1.34.0"; then
         logs_to_events "AKS.CSE.configureKubeletAndKubectl.installKubeletKubectlFromURL" installKubeletKubectlFromURL
     else
+        if [ -n "${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER}" ] ; then
+            # For network isolated clusters, try distro packages first and fallback to binary installation
+            if ! logs_to_events "AKS.CSE.configureKubeletAndKubectl.installK8sToolsFromBootstrapProfileRegistry" "installK8sToolsFromBootstrapProfileRegistry ${BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER} ${KUBERNETES_VERSION}" && [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ]; then
```
fseldow, line 537: when there is no fallback with SHOULD_ENFORCE_KUBE_PMC_INSTALL==true,
installK8sToolsFromBootstrapProfileRegistry will not exit with ERR_ORAS_PULL_K8S_FAIL(207).
It will continue to run installFromPMC

qweeah: It's a bug. Not aborted after `installK8sToolsFromBootstrapProfileRegistry`. Fixed with test covered.

parts/linux/cloud-init/artifacts/cse_helpers.sh
qweeah: Seems `$*` is buggy, let's switch back to `$@`. I will add unit test coverage for it.

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -526,6 +526,45 @@ extractKubeBinaries() {
...
+    # Try to pull distro-specific packages (e.g., .deb for Ubuntu) from registry
+    local download_root="/tmp/kubernetes/downloads" # /opt folder will return permission error
+
+    for tool_name in $(get_kubernetes_tools); do
+        tool_package_url="${registry_server}/aks/packages/kubernetes/${tool_name}:v${kubernetes_version}"
+        tool_download_dir="${download_root}/${tool_name}"
+        mkdir -p "${tool_download_dir}"
+
+        # Construct platform string for ORAS pull
+        os_version=$(grep DISTRIB_RELEASE /etc/*-release| cut -f 2 -d "=") # todo: OS_VARIANT not working for ubuntu
```
fseldow: what about OS_VERSION?
this os_version not work for azure linux v3, it will return 3.0.master.xxxx, expected 3.0

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -526,6 +526,45 @@ extractKubeBinaries() {
     extractKubeBinariesToUsrLocalBin "${k8s_tgz_tmp}" "${k8s_version}" "${is_private_url}"
 }
 
+installK8sToolsFromBootstrapProfileRegistry() {
```
fseldow, line 529: kubelet and kubectl need to be installed in `/usr/local/bin`

qweeah: Added copy and validation for ubuntu

parts/linux/cloud-init/artifacts/cse_install.sh
qweeah: Made this OS specific

fseldow (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -40,22 +43,38 @@ apt_get_update() {
...
             sleep $wait_sleep
             apt_get_update
         fi
     done
-    echo Executed apt-get install --no-install-recommends -y \"$@\" $i times;
-    wait_for_apt_locks
+}
+apt_get_install() {
+    retries=$1; wait_sleep=$2; timeout=$3; shift && shift && shift
+    _apt_get_install $retries $wait_sleep '--no-install-recommends -o Dpkg::Options::="--force-confold"' "$@"
```
fseldow: it eventually runs `apt-get install --no-install-recommends -o 'Dpkg::Options::="--force-confold"' -y apparmor-utils`
will fail with error dpkg: error: need an action option

there should be no queto before Dpkg

qweeah: Resolved

parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -526,6 +526,46 @@ extractKubeBinaries() {
...
```
Devinwong, line 529: Is it expected that `installK8sToolsFromBootstrapProfileRegistry` always return 1 when it's on AzureLinux? Asking because I see that `getOsVersion` in Mariner scripts always return 1.

qweeah: Yes, AzureLinux and Mariner will be supported in additional PRs. At the moment, both `installToolFromLocalRepo` and `getOsVersion` return 1, which causes the installation fallback to `installKubeletKubectlFromURL`.

Devinwong (approved):
Approved. Thanks.

fseldow (approved):

PR #7142, opened 2025-10-07
fix: config latest patch for kube-proxy

jack4it:
**What type of PR is this?**

/kind bug

* adds configs for kube-proxy for the latest two patches
* adds similar configs for the v2 dalec image as well

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -708,23 +708,6 @@
         }
       ],
       "windowsVersions": []
-    },
-    {
-      "downloadURL": "mcr.microsoft.com/oss/v2/kubernetes/kube-proxy:*",
-      "amd64OnlyVersions": [],
-      "multiArchVersionsV2": [
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/kube-proxy",
-          "latestVersion": "v1.34.0-2",
-          "previousLatestVersion": "v1.34.0"
-        },
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/kube-proxy",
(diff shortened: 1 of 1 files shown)


jack4it:
@microsoft-github-policy-service agree [company="Microsoft"]

parts/common/components.json
```diff
@@ -698,8 +698,30 @@
...
+      "windowsVersions": []
+    },
+    {
+      "downloadURL": "mcr.microsoft.com/oss/v2/kubernetes/kube-proxy:*",
+      "amd64OnlyVersions": [],
+      "multiArchVersionsV2": [
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/kube-proxy",
+          "latestVersion": "v1.34.0-2",
+          "previousLatestVersion": "v1.34.0"
```
Devinwong: According to Dalec doc, https://github.com/Azure/dalec-build-defs/blob/main/faq.md#how-do-floating-vs-static-tags-work, it will be better to use either just `static` or `floating` (recommended static) for clarity. For this case, you can't really control what v1.34.0 is if you are using floating version.

parts/common/components.json
```diff
@@ -698,8 +698,30 @@
...
+        },
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/kube-proxy",
+          "latestVersion": "v1.34.1-2",
+          "previousLatestVersion": "v1.34.1"
```
Devinwong: Same. Floating version.

PR #7160, opened 2025-10-09
Add E2E for VHD Caching

r2k1:
**What type of PR is this?**

/kind test

Add test for Linux and Windows testign VHD caching functionality.
This creates a VHD, pushes it to a test image gallery and then provision VM from it.

The test is slow, taking extra 10-15 minutes to complete. Needs extra permissions to write to the test gallery. But without it there is a high chance the functionality will break at some point in the future.

### DIFF
--- a/e2e/test_helpers.go
+++ b/e2e/test_helpers.go
@@ -116,7 +116,8 @@ func runScenarioWithPreProvision(t *testing.T, original *Scenario) {
 		}
 		t.Log("=== Creating VHD Image ===")
 		customVHD = CreateImage(ctx, stage1)
-		t.Logf("Created custom VHD image: %#v", customVHD)
+		customVHDJSON, _ := json.MarshalIndent(customVHD, "", "  ")
+		t.Logf("Created custom VHD image: %s", string(customVHDJSON))
 	}
 	firstStage.Config.VMConfigMutator = func(vmss *armcompute.VirtualMachineScaleSet) {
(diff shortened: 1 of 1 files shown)


e2e/test_helpers.go
```diff
@@ -281,6 +277,7 @@ func maybeSkipScenario(ctx context.Context, t *testing.T, s *Scenario) {
 	s.Tags.OS = string(s.VHD.OS)
 	s.Tags.Arch = s.VHD.Arch
 	s.Tags.ImageName = s.VHD.Name
+	s.Tags.VHDCaching = true
```
timmy-wright: I'm confused by this line - it looks like we're always setting VHD caching to true. Rather than just having it for some tests.

r2k1: because that what it does ...


Fixed

e2e/exec.go
```diff
@@ -68,20 +56,19 @@ func execScriptOnVm(ctx context.Context, s *Scenario, vmPrivateIP, jumpboxPodNam
 	identifier := uuid.New().String()
 	var scriptFileName, remoteScriptFileName, interpreter string
 
-	switch script.interpreter {
-	case Powershell:
+	if s.IsWindows() {
```
timmy-wright, line 59: I think I prefer this as a switch statement because we might add different interpreters in future. But it's not a strong preference, so no need to change.

PR #7176, opened 2025-10-11
Update localdns to always execute networkreload in cleanup

SriHarsha001:
**What type of PR is this?**
/kind bug

This PR updates localdns script to always execute networkreload in cleanup

Fixes # In the scenario where network dropin file used by localdns is missing for whatever reason, eval "$NETWORKCTL_RELOAD_CMD" is not executed, so dhcp didn’t pick up custom DNS servers from vnet, localdns clusterlistener ip remained as nameserver in resolv.conf. This will lead to incorrectly forwarding requests to AzureDNSIP instead of VnetDNSIPs. 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/localdns.sh
+++ b/parts/linux/cloud-init/artifacts/localdns.sh
@@ -221,6 +221,45 @@ verify_network_dropin_dir() {
     return 0
 }
 
+# Initialize network variables that may not be set during restarts or cleanup traps.
+# This function can be called both during startup and in cleanup functions to ensure variables are available.
+initialize_network_variables() {
+    # Determine the default route interface if not already set.
+    # This will typically be eth0.
+    if [ -z "${DEFAULT_ROUTE_INTERFACE:-}" ]; then
+        DEFAULT_ROUTE_INTERFACE="$(ip -j route get "${AZURE_DNS_IP}" 2>/dev/null | jq -r 'if type == "array" and length > 0 then .[0].dev else empty end')"
+        if ! verify_default_route_interface; then
+            echo "Failed to determine default route interface during variable initialization."
+            return 1
+        fi
+    fi
+
+    # Get the network file associated with the default route interface if not already set.
+    # This will typically be /run/systemd/network/10-netplan-eth0.network.
+    if [ -z "${NETWORK_FILE:-}" ]; then
+        NETWORK_FILE="$(networkctl --json=short status "${DEFAULT_ROUTE_INTERFACE}" 2>/dev/null | jq -r '.NetworkFile')"
+        if ! verify_network_file; then
+            echo "Failed to determine network file during variable initialization."
+            return 1
+        fi
+    fi
+
+    # Get the network drop-in directory.
+    # This will typically be /run/systemd/network/10-netplan-eth0.network.d
+    if [ -z "${NETWORK_DROPIN_DIR:-}" ]; then
+        NETWORK_DROPIN_DIR="/run/systemd/network/${NETWORK_FILE##*/}.d"
+    fi
+
+    # Set the network drop-in file path if not already set.
+    # 70-localdns.conf file is written later in disable_dhcp_use_clusterlistener function.
+    # disable_dhcp_use_clusterlistener function is called after the cleanup function.
+    if [ -z "${NETWORK_DROPIN_FILE:-}" ]; then
+        NETWORK_DROPIN_FILE="${NETWORK_DROPIN_DIR}/70-localdns.conf"
+    fi
+
+    return 0
+}
+
 # Start localdns service.
 start_localdns() {
     echo "Starting localdns: ${COREDNS_COMMAND}."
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -353,28 +353,28 @@ cleanup_iptables_and_dns() {
...
-        fi        
-        eval "$NETWORKCTL_RELOAD_CMD"
-        if [ "$?" -ne 0 ]; then
-            echo "Failed to reload network after removing the DNS configuration."
-            return 1
-        fi
+    # Revert DNS configuration and network reload.
+    echo "Reverting DNS configuration by removing ${NETWORK_DROPIN_FILE}."
+    rm -f "$NETWORK_DROPIN_FILE"
+    if [ "$?" -ne 0 ]; then
```
yewmsft, line 408: please add happy path logging for debugbility

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -353,28 +353,28 @@ cleanup_iptables_and_dns() {
...
+        echo "Failed to remove network drop-in file ${NETWORK_DROPIN_FILE}."
+        return 1
     fi
+
+    eval "$NETWORKCTL_RELOAD_CMD"
```
yewmsft, line 415: like "start reloading network configurations..."

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: yes sure.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: Added echo statements before and after the execution.

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -469,6 +521,10 @@ verify_localdns_slicefile || exit $ERR_LOCALDNS_SLICEFILE_NOTFOUND
 # /opt/azure/containers/localdns/binary/coredns.
 verify_localdns_binary || exit $ERR_LOCALDNS_BINARY_ERR
 
+# Set required network environment variables.
+# ---------------------------------------------------------------------------------------------------------------------
+initialize_network_variables || exit $ERR_LOCALDNS_FAIL
```
yewmsft, line 526: pure question, why move this up? It should work at either location, right?

parts/linux/cloud-init/artifacts/localdns.sh
```diff
@@ -221,6 +221,45 @@ verify_network_dropin_dir() {
     return 0
 }
 
+# Initialize network variables that may not be set during restarts or cleanup traps.
+# This function can be called both during startup and in cleanup functions to ensure variables are available.
+initialize_network_variables() {
+    # Determine the default route interface if not already set.
+    # This will typically be eth0.
+    if [ -z "${DEFAULT_ROUTE_INTERFACE:-}" ]; then
+        DEFAULT_ROUTE_INTERFACE="$(ip -j route get "${AZURE_DNS_IP}" 2>/dev/null | jq -r 'if type == "array" and length > 0 then .[0].dev else empty end')"
```
yewmsft, line 230: even though very rare, but the node is not guaranteed to have a route for ${AZURE_DNS_IP}, right? What if the cx is using custom DNS server and blocks Azure DNS IP in route table?

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: But I think it is ok to move it here. Just checking if I missed something.

SriHarsha001: This was the actual bug. When watchdog triggers SIGABRT/SIGKILL on healthcheck failures and SIGABRT/SIGKILL is not trapped in bash, so cleanup is not invoked. localdns systemd unit will just restart, but variable holding the network dropin file path is not set and network dropin file exists in the path when cleanup is called again. So moving the variable before the cleanup will set all the network variable needed for removing or updating network dropin file.

parts/linux/cloud-init/artifacts/localdns.sh
SriHarsha001: ip -j route is asking kernel if I need to route a packet to 168.63.129.16 which nic will I take. From this we are getting eth0 nic. ip route get makes a routing table lookup and does not actually send a packet. So any firewall rules added by customer to drop the packets destined to 168.63.129.16 will not cause ip route get call failures. 

Inside a node that is pointing to custom DNS server - 

image

yewmsft: cx can block azure dns from route table
image

yewmsft: I think get route from vnet dns is safer

parts/linux/cloud-init/artifacts/localdns.sh
yewmsft: i c

PR #7179, opened 2025-10-12
regression: bring back storage account creation

timmy-wright:
**What type of PR is this?**

Brings back the storage account creation when windows VHDs are using a different base image version

So we can build pre-release VHDs for testing.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/spec/vhdbuilder/packer/ensure_sig_vhd_exists_spec.sh
+++ b/spec/vhdbuilder/packer/ensure_sig_vhd_exists_spec.sh
@@ -106,7 +106,8 @@ Describe 'ensure_sig_vhd_exists function'
       # Mock grep for cvm feature flag checking
       case "$*" in
         *cvm*)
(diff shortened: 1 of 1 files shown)


spec/vhdbuilder/packer/produce_ua_token_spec.sh
```diff
@@ -0,0 +1,333 @@
+#!/bin/bash
+
+# Tests for produce_ua_token function from produce-packer-settings-functions.sh
+
+Describe 'produce_ua_token function'
```
cameronmeissner, line 5: thanks for increasing test coverage!

timmy-wright: Thank Claude!

PR #7180, opened 2025-10-13
chore: use CI approach to run shell checks locally

timmy-wright:
**What type of PR is this?**

/kind chore

move docker modification from pipeline yaml to makefile so it can be used locally.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,31 +77,33 @@ validate-go:
 validate-shell:
 	@./.pipelines/scripts/verify_shell.sh
 
+.PHONY: shellspec-base-local
+shellspec-base-local:
+	# don't have access to aksdataplanedev.azurecr.io locally, so strip it
+	sed 's|aksdataplanedev.azurecr.io/||g' < ./spec/shellspec.Dockerfile | docker build -t shellspec-docker -
 
-.PHONY: shellspec-run
-shellspec-run:
+.PHONY: shellspec-base-ci
+shellspec-base-ci:
+	# don't have access to aksdataplanedev.azurecr.io in CI, so strip it
+	sed 's|aksdataplanedev.azurecr.io/||g' < ./spec/shellspec.Dockerfile | docker build -t shellspec-docker -
+	# docker build -t shellspec-docker - < ./spec/shellspec.Dockerfile
+
+.PHONY: shellspec
+shellspec: shellspec-base-local
 	docker run --rm \
 		-v $(CURDIR):/workspace \
 		-w /workspace \
 		shellspec-docker --shell bash --format d
 
-
-.PHONY: shellspec
-shellspec:
-	# don't have access to aksdataplanedev.azurecr.io locally, so strip it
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -77,17 +77,33 @@ validate-go:
...
+.PHONY: shellspec-base-local
+shellspec-base-local:
+	# don't have access to aksdataplanedev.azurecr.io locally, so strip it
+	sed 's|aksdataplanedev.azurecr.io/||g' < ./spec/shellspec.Dockerfile | docker build -t shellspec-docker -
+
+.PHONY: shellspec-base-ci
+shellspec-base-ci:
+	# don't have access to aksdataplanedev.azurecr.io in CI, so strip it
+	sed 's|aksdataplanedev.azurecr.io/||g' < ./spec/shellspec.Dockerfile | docker build -t shellspec-docker -
+	# docker build -t shellspec-docker - < ./spec/shellspec.Dockerfile
```
benjamin-brady, line 89: Did you mean to leave this comment here?

Makefile
```diff
@@ -77,17 +77,33 @@ validate-go:
 validate-shell:
 	@./.pipelines/scripts/verify_shell.sh
 
+.PHONY: shellspec-base-local
+shellspec-base-local:
+	# don't have access to aksdataplanedev.azurecr.io locally, so strip it
```
cameronmeissner, line 82: this is something @djsly created, we should be able to create team members access

Makefile
```diff
@@ -77,17 +77,33 @@ validate-go:
...
+shellspec-ci: shellspec-base-ci
 	docker run --rm \
 		-v $(CURDIR):/workspace \
 		-w /workspace \
 		shellspec-docker --shell bash --format d
 
 .PHONY: shellspec-focus
-shellspec-focus:
-	docker build -t shellspec-docker - < ./spec/shellspec.Dockerfile
+shellspec-focus: shellspec-base-local
```
cameronmeissner, line 106: I don't think we need to run shellspec-base-local here?

Makefile
timmy-wright: Yes - because that's what the code should do once we fix the pipeline permissions issue.

.github/workflows/shellspec.yaml
```diff
@@ -10,11 +10,10 @@ jobs:
       with:
         go-version: '1.24'
     - run: |
-        sed -i 's|aksdataplanedev.azurecr.io/||g' ./spec/shellspec.Dockerfile
```
Devinwong, line 13: agreed we should put the sed command in local make command instead of in the CI tests/actions

Makefile
timmy-wright: Still need to build the docker container to run shellspec inside it!

Makefile
timmy-wright: He's away this week - so let's do a fix next week once we have acccess.

PR #7191, opened 2025-10-14
fix: Remove pipeline warning on failing CIS rule 1.3.1.2 on CVM

jepio:
**What type of PR is this?**

/kind test

The rule can't be made to pass on CVM because CVM's rely on signed kernel EFI binaries where the signature includes the bootloader cmdline.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/vhd-scanning.sh
+++ b/vhdbuilder/packer/vhd-scanning.sh
@@ -237,6 +237,15 @@ isAzureLinuxOSGuard() {
     fi
     return 1
 }
+isUbuntuCVM() {
+    local os="$1"
+    local feature_flags="$2"
+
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -237,6 +237,15 @@ isAzureLinuxOSGuard() {
     fi
     return 1
 }
+isUbuntuCVM() {
+    local os="$1"
+    local feature_flags="$2"
+
+    if [ "$os" = "Ubuntu" ] && grep -q "cvm" <<< "$feature_flags"; then
+	return 0
```
cameronmeissner: nit: indentation is off

PR #7197, opened 2025-10-15
test: add tests to ensure that any iptables rules present are compatible with eBPF host routing

santhoshmprabhu:
**What type of PR is this?**

/kind test

We are currently rolling out a feature that allows customers to enable so-called eBPF host routing on clusters that use Cilium CNI. Enabling this feature causes iptables rules in the host network namespace to be bypassed by packets from non-hostNetwork pods. We have functionality such as wireserver firewall and IMDS restriction that are implemented using iptables rules, by Agent Baker. On clusters where customers enable eBPF host routing, this functionality has been ported to Cilium Clusterwide Network Policies.

If any new functionality is added to nodes using iptables rules, we need to make sure that we have an equivalent implementation that is compatible with eBPF host routing, or block the combination of new functionality with eBPF host routing. In fact, when eBPF host routing is enabled, our checks will cause the cluster to be inoperational if any rules are detected that we have not accounted for. In order to identify such cases, this PR adds a test to verify that only "known" iptables rules are present on the nodes. These rules are accounted for in our current implementation of eBPF host routing. The check is added to common linux validation, so as to catch any future changes that may introduce unsupported IPTables rules.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -127,7 +127,7 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 		ValidateLocalDNSResolution(ctx, s, "169.254.10.10")
 	}
 
-	ValidateIPTablesRules(ctx, s)
+	ValidateIPTablesCompatibleWithCiliumEBPF(ctx, s)
 }
 
 func ValidateSystemdWatchdogForKubernetes132Plus(ctx context.Context, s *Scenario) {
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -1202,8 +1202,9 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
 	s.T.Logf("MIG instances with profile %s are created", migProfile)
 }
 
-// ValidateIPTablesRules validates that all iptables rules in each table match the provided patterns
-func ValidateIPTablesRules(ctx context.Context, s *Scenario) {
+// ValidateIPTablesCompatibleWithCiliumEBPF validates that all iptables rules in each table match the provided patterns which are accounted for
+// when eBPF host routing is enabled.
+func ValidateIPTablesCompatibleWithCiliumEBPF(ctx context.Context, s *Scenario) {
 	s.T.Helper()
 	tablePatterns, globalPatterns := getIPTablesRulesCompatibleWithEBPFHostRouting()
 	tables := []string{"filter", "mangle", "nat", "raw", "security"}


e2e/scenario_test.go
```diff
@@ -2053,3 +2053,100 @@ func Test_AzureLinux3OSGuard_PMC_Install(t *testing.T) {
 		},
 	})
 }
+
+
+func Test_Ubuntu2404_IMDS_Restriction_IPTables_Rules(t *testing.T) {
```
Devinwong: Thanks for adding the tests.
Could you please add the same tests for scriptless, which is a new provisioning approach. Example: `Test_Ubuntu2204_ContainerdURL_IMDSRestrictionFilterTable_Scriptless`.

e2e/scenario_test.go
```diff
@@ -2053,3 +2053,100 @@ func Test_AzureLinux3OSGuard_PMC_Install(t *testing.T) {
...
+                       },
+                       Validator: func(ctx context.Context, s *Scenario) {
+                               tablePatterns, globalPatterns := getIPTablesRulesCompatibleWithEBPFHostRouting()
+                               ValidateIPTablesRules(ctx, s, tablePatterns, globalPatterns)
+                       },
+               },
+       })
+ }
+
+func Test_AzureLinuxV3_IMDS_Restriction_IPTables_Rules(t *testing.T) {
```
Devinwong: Please add test for scriptless too

e2e/scenario_test.go
```diff
@@ -2053,3 +2053,100 @@ func Test_AzureLinux3OSGuard_PMC_Install(t *testing.T) {
...
+                               ValidateIPTablesRules(ctx, s, tablePatterns, globalPatterns)
+                       },
+               },
+       })
+ }
+
+ // getIPTablesRulesCompatibleWithEBPFHostRouting returns the expected iptables patterns that are accounted for when EBPF host routing is enabled.
+ // If tests are failing due to unexpected iptables rules, please check with the Azure container networking team before updating these patterns,
+ // to ensure compatibility with EBPF host routing.
+func getIPTablesRulesCompatibleWithEBPFHostRouting() (map[string][]string, []string) {
```
cameronmeissner: let's move this into validators/validation.go instead

e2e/validators.go
```diff
@@ -1183,3 +1183,64 @@ func ValidateNvidiaDCGMExporterScrapeCommonMetric(ctx context.Context, s *Scenar
 	}
 	execScriptOnVMForScenarioValidateExitCode(ctx, s, strings.Join(command, "\n"), 0, "Nvidia DCGM Exporter is not returning DCGM_FI_DEV_GPU_UTIL")
 }
+// ValidateIPTablesRules validates that all iptables rules in each table match the provided patterns
+func ValidateIPTablesRules(ctx context.Context, s *Scenario, tablePatterns map[string][]string, globalPatterns []string) {
```
cameronmeissner: would prefer we make this a ubiquitous validator that runs on every scenario to make sure that no unexpected IP tables rules are added other than what's needed for IMDS restriction / wireserver firewall - since it seems like these patterns would never change across scenarios

e2e/scenario_test.go
santhoshmprabhu: Moved the checks to the common linux validation, so now the checks apply everywhere.

e2e/scenario_test.go
santhoshmprabhu: Moved the checks to the common linux validation, so now the checks apply everywhere.

e2e/validators.go
santhoshmprabhu: Moved to common linux validation.

e2e/validation.go
```diff
@@ -126,6 +126,9 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 		ValidateLocalDNSService(ctx, s, "enabled")
 		ValidateLocalDNSResolution(ctx, s, "169.254.10.10")
 	}
+
+	tablePatterns, globalPatterns := getIPTablesRulesCompatibleWithEBPFHostRouting()
```
r2k1: I would move this call inside ValidateIPTablesRules

e2e/validators.go
```diff
@@ -1201,3 +1201,128 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
...
+		}
+
+		s.T.Logf("All rules in table %s matched expected patterns", table)
+	}
+}
+
+// getIPTablesRulesCompatibleWithEBPFHostRouting returns the expected iptables patterns that are accounted for when EBPF host routing is enabled.
+// If tests are failing due to unexpected iptables rules, please check with the Azure container networking team before updating these patterns,
+// to ensure compatibility with EBPF host routing.
+func getIPTablesRulesCompatibleWithEBPFHostRouting() (map[string][]string, []string) {
```
r2k1, line 1284: Not critical, but I would either:
- create two functions 
- create two global variable
- create local variable inside validator

e2e/validators.go
```diff
@@ -1201,3 +1201,128 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
...
+	tablePatterns := map[string][]string{
+		"filter": {
+			`^.*--comment.*kubernetes load balancer firewall`,
+			`^.*--comment.*kubernetes health check service ports`,
+			`^.*--comment.*kubernetes externally visible service portals`,
+			`^.*--comment.*kubernetes forwarding rules`,
+			`^.*--comment.*kubernetes forwarding conntrack rule`,
+			`^.*--comment.*kubernetes service portals`,
+			`^.*--comment.*kubernetes externally-visible service portals`,
```
r2k1: This sounds fragile. Does it mean every time we modify comments we need to update e2e tests? I don't thin we should have any rules based on comments values

r2k1 (commented):
I don't think I like the set of rules. It basically prevents us from changing anything inside iptables.

So when someone is going to change something the test will just fail without any actionable items. So the only thing left for the developer is to include the new rule in the e2e.

I don't think it's helping us in validating we're compatible with ebpf. It just prevents any changes and doesn't explain how we actually should deal with them. At this state it just introduces another place that we need to manually update.

santhoshmprabhu:
@r2k1 I think I agree that a failure in this test is not directly actionable for the node sig in isolation. If we have a new iptables rule added to nodes, the test will fail, and that is designed that way because the rule will be skipped in eBPF host routing mode, and the feature will break when enabled alongside eBPF host routing. So, we either we need to prevent both features being enabled together (in RP validation), or have a non-iptables implementation that is compatible with eBPF host routing. 

Furthermore, our implementation of eBPF host routing currently has checks to catch "unaccounted" iptables rules, and if any are found, start raising kube events to warn the customer that their cluster is likely broken. The patterns in this PR come directly from the checks that are performed to find such rules. So, you're right that this makes the iptables config fairly rigid. Basically, if changes are made to iptables, we need to update our Cilium daemonset to allow the change, and that can be done only after we ensure that the change doesn't inadvertently enable a feature combination that breaks the cluster.

r2k1:
So you're saying we can't change IPTables from now? (It's quite annoying)

In this case the test failure should tell who should approve it. (Or what should be updated at the same time)
12 months from now nobody will know what we suppose to do with it.

I also see you just ignore/allow some rules that contains a comment. Can we have a single `AgentBaker` comment (need to figure out a better name, something like "AKSNode") that we can apply to all rules to allow us to add an IPTable rule without a need to change anything else? In this case we can just validate our rules contains the comment.

santhoshmprabhu:
@r2k1 @Devinwong @cameronmeissner Please take another look.

e2e/validators.go
Devinwong: Copilot said: 
`Not checking the comment in iptables rule validation is usually not recommended in environments like Kubernetes, especially when the validation intends to ensure security, interoperability, or compliance with known rule sets.` 
But @santhoshmprabhu  can confirm

Devinwong:
Since the validation applies to alls OS SKUs, I would be more cautious by running a full rebuild VHDs and E2e Tests. Triggered that here: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=142508120&view=results
Other than this, it looks good to me.

e2e/validators.go
```diff
@@ -1201,3 +1201,140 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
 	require.NotContains(s.T, stdout, "No MIG-enabled devices found", "no MIG devices were created.\nOutput:\n%s", stdout)
 	s.T.Logf("MIG instances with profile %s are created", migProfile)
 }
+
+// ValidateIPTablesRules validates that all iptables rules in each table match the provided patterns
+func ValidateIPTablesRules(ctx context.Context, s *Scenario) {
```
r2k1: I would rename it to something like

`ValidateIPTableCompatibleWithCiliumEBPF`

e2e/validators.go
r2k1: Still sounds weird that we validate comment on the rule instead of the rule itself.
The comment doesn't have any impact.

Devinwong (approved):

e2e/validators.go
santhoshmprabhu: Replaced the use of comments with chains. Mostly we were using comments to identify various functions that kuberenetes itself was implementing using iptables (those are anyway supported by Cilium in eBPF host routing mode).

e2e/validators.go
```diff
@@ -1201,3 +1201,124 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
...
+// ValidateIPTablesCompatibleWithCiliumEBPF validates that all iptables rules in each table match the provided patterns which are accounted for
+// when eBPF host routing is enabled.
+func ValidateIPTablesCompatibleWithCiliumEBPF(ctx context.Context, s *Scenario) {
+	s.T.Helper()
+	tablePatterns, globalPatterns := getIPTablesRulesCompatibleWithEBPFHostRouting()
+	tables := []string{"filter", "mangle", "nat", "raw", "security"}
+	success := true
+
+	for _, table := range tables {
+		s.T.Logf("Validating iptables rules for table: %s", table)
```
r2k1, line 1214: That's  noisy. I would remove it.

e2e/validators.go
```diff
@@ -1201,3 +1201,124 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
...
+				}
+			}
+
+			if !matched {
+				s.T.Logf("Rule in table %s did not match any pattern: %s", table, rule)
+				success = false
+			}
+		}
+
+		s.T.Logf("Checked rules in table %s against expected patterns", table)
```
r2k1, line 1266: That's noisy as well

r2k1 (approved):

PR #7210, opened 2025-10-16
Managed GPU experience AFEC enablement flag

ganeshkumarashok:
**What type of PR is this?**
/kind feature

Adds a field to NodeBootstrappingConfig for tracking AFEC flag enablement (passed via RP), that will be used to decide whether to enable/disable the feature during preview. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


pkg/agent/datamodel/types.go
```diff
@@ -1759,6 +1759,7 @@ type NodeBootstrappingConfiguration struct {
 	EnableKubeletConfigFile       bool
 	EnableNvidia                  bool
 	EnableAMDGPU                  bool
+	ManagedGPUExperienceAFECEnabled bool
```
cameronmeissner, line 1762: why not just `ManagedGPUExperienceEnabled`?

ganeshkumarashok: To make it clear that it's just the AFEC flag. The feature itself is enabled via the VMSS tag. In a separate PR (after aks-rp passes in this flag to NBC), the CSE change for managed GPU Experience will be conditioned on this AFEC flag too.

ganeshkumarashok: We may also add an additional field for just `ManagedGPUExperienceEnabled` if there are RP changes for GA.

cameronmeissner (approved):

PR #7216, opened 2025-10-17
Fix DNS resolution on Flatcar

johananl:
**What type of PR is this?**
/kind bug

This PR fixes DNS configuration for Flatcar by setting the content of `/etc/nsswitch.conf`.

The [stock](https://github.com/flatcar/baselayout/blob/5f3a7e45028a358b3406d3743777f353dcffb3e4/share/baselayout/nsswitch.conf#L7) Flatcar configuration attempts to use systemd-resolved in some cases, which doesn't seem to work with Azure CNI. This change aligns the DNS resolution behavior with Ubuntu and fixes the resolution problems.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


johananl:
The tests I've added should fail without the fix, however to test this manually create a cluster from a Flatcar VHD and run `ping google.com`. Before the fix you should get:

```
ping: google.com: Temporary failure in name resolution
```

After the fix ping should work.

cameronmeissner (approved):

cameronmeissner:
@johananl your commits must be signed to merge: https://github.com/Azure/AgentBaker/blob/master/CONTRIBUTING.md

PR #7223, opened 2025-10-20
Fix VHD Caching e2e for gen1 VHDs

r2k1:
**What type of PR is this?**
/kind test

Fix `HyperVGeneration` property on the created VHD. Pull it from VM record.
Update e2e logic to fetch VM record early, attach it to the Scenario and use it later. Avoid hard-coded "0" id for VMSS VM.
Avoid fetching same data multiple times.

### DIFF
--- a/e2e/test_helpers.go
+++ b/e2e/test_helpers.go
@@ -80,12 +80,14 @@ func newTestCtx(t testing.TB) context.Context {
 
 func RunScenario(t *testing.T, s *Scenario) {
 	t.Parallel()
+	// Special case for testing VHD caching. Not used by default.
 	if config.Config.TestPreProvision || s.VHDCaching {
 		t.Run("VHDCreation", func(t *testing.T) {
 			t.Parallel()
 			runScenarioWithPreProvision(t, s)
 		})
 	} else {
+		// Default path
 		runScenario(t, s)
 	}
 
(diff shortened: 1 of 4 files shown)


e2e/kube.go
```diff
@@ -147,10 +147,9 @@ func (k *Kubeclient) WaitUntilPodRunning(ctx context.Context, namespace string,
 
 func (k *Kubeclient) WaitUntilNodeReady(ctx context.Context, t testing.TB, vmssName string) string {
 	startTime := time.Now()
+	t.Logf("waiting for node %s to be ready in k8s API", vmssName)
+	defer t.Logf("waited for node %s to be ready in k8s API for %s", vmssName, time.Since(startTime))
```
timmy-wright: When is the "time.Since" executed with a defer? I'm unfamiliar with go semantics in this case.

e2e/scenario_win_test.go
```diff
@@ -342,21 +342,19 @@ func Test_Windows2025Gen2(t *testing.T) {
 	})
 }
 
-func Test_Windows2025Gen2_VHDCaching(t *testing.T) {
+func Test_Windows2022_VHDCaching(t *testing.T) {
```
timmy-wright, line 345: It'd be good to have seperate gen1 and a gen2 tests.

e2e/kube.go
r2k1: Too early. Fixed.

e2e/scenario_win_test.go
r2k1: There shouldn't be any difference between them (outside of a single important property that's fixed now and only set inside E2E). I want to limit number of e2e

djsly (approved):

timmy-wright (approved):

PR #6953, opened 2025-09-05
Enable jumbo frame for mana driver

anson627:
This pull request adds a new udev rule to automatically set the MTU to 9000 for Microsoft MANA Ethernet controllers on Linux systems. This helps ensure optimal network performance for accelerated networking in Azure environments.

Networking improvements:

* Added `/etc/udev/rules.d/99-microsoft-mana-mtu.rules` to set MTU to 9000 for network interfaces using the Microsoft MANA driver, leveraging a udev rule that triggers on interface addition.**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version
- [ ] commits are GPG signed and Github marks them as verified

**Special notes for your reviewer**:

**Release note**:

```
none
```

PR #7228, opened 2025-10-21
Revert "Enable jumbo frame for v6 VM SKU with mana driver (#6953)"

anson627:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


timmy-wright:
Windows diff is likely due to this PR branch not being merged with latest master so the diff command is comparing an older master with the latest master. Happy to ignore in this case.

PR #7233, opened 2025-10-22
test: Skip windows e2e for Linux testdata update

jepio:
**What type of PR is this?**
/kind test

The windows e2e pipeline had an exclude on windows testdata updates which looks like it was copied from the linux e2e pipeline definition. Replace this with some linux testdata paths that can safely skip windows e2e runs.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/e2e-windows.yaml
+++ b/.pipelines/e2e-windows.yaml
@@ -34,7 +34,10 @@ pr:
     exclude:
       - pkg/agent/datamodel/sig_config*.go # SIG config changes
       - pkg/agent/datamodel/*.json # SIG version changes
-      - pkg/agent/testdata/AKSWindows* # Windows test data
+      - pkg/agent/testdata/*Ubuntu* # Linux test data
+      - pkg/agent/testdata/*Flatcar* # Linux test data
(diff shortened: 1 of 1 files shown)


.pipelines/e2e-windows.yaml
```diff
@@ -34,7 +34,10 @@ pr:
     exclude:
       - pkg/agent/datamodel/sig_config*.go # SIG config changes
       - pkg/agent/datamodel/*.json # SIG version changes
-      - pkg/agent/testdata/AKSWindows* # Windows test data
```
djsly, line 37: this was there ?  should we keep ? I'll let Tim review

r2k1: testdata should only be used for unit testing. It should not have any affect on e2e.

I think we should ignore the whole folder.

r2k1: It can only used be as indirect signal "CSE has changed". Though it would be weird and we should have a more explicit ways to detect it

PR #7237, opened 2025-10-23
feat: fix cuda install & remove deprecated key

stanleyt10:
**What type of PR is this?**
/kind bug

Found that in AGC the dnf repoquery command was returning "" for CUDA_PACKAGES causing dnf install "" to fail. Issue is that gpg keys are not trusted by default in enviornment and we need to run repoquery command with a -y. Also removing deprecated key in azl3.0. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


henryli001 (approved):

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -79,9 +79,11 @@ downloadGPUDrivers() {
     # The proprietary driver will be used here in order to support older NVIDIA GPU SKUs like V100
     # Before installing cuda, check the active kernel version (uname -r) and use that to determine which cuda to install
     KERNEL_VERSION=$(uname -r | sed 's/-/./g')
-    CUDA_PACKAGE=$(dnf repoquery --available "cuda*" | grep -E "cuda-[0-9]+.*_$KERNEL_VERSION" | sort -V | tail -n 1)
+    CUDA_PACKAGE=$(dnf repoquery -y --available "cuda*" | grep -E "cuda-[0-9]+.*_$KERNEL_VERSION" | sort -V | tail -n 1)
 
-    if ! dnf_install 30 1 600 ${CUDA_PACKAGE}; then
+    if [ -z "$CUDA_PACKAGE" ]; then
+      echo "No cuda packages found"
```
djsly, line 85: here it's quite tricky, we would be expecting the pacakges to show up, so if they are not there, it will fail later on with a more obscure reason,  GPU driver wont be installed, something else will fail.   We should fail here right away instead and be clear about the reason, I would add a new exit $ERR_MISSING_PACKAGE

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -6,6 +6,7 @@ ERR_FILE_WATCH_TIMEOUT=6 # Timeout waiting for a file
 ERR_HOLD_WALINUXAGENT=7 # Unable to place walinuxagent apt package on hold during install
 ERR_RELEASE_HOLD_WALINUXAGENT=8 # Unable to release hold on walinuxagent apt package after install
 ERR_APT_INSTALL_TIMEOUT=9 # Timeout installing required apt packages
+ERR_MISSING_PACKAGES=10 # Unable to query / install required apt packages
```
cameronmeissner: I'd opt for having a more specific error code here, e.g. `ERR_MISSING_CUDA_PACKAGE` to make things more explicit without the need to go digging in provisioning logs to know exactly which package was missing

djsly (approved):

PR #7248, opened 2025-10-24
fix: retry when getting network adapter

timmy-wright:
**What type of PR is this?**

/kind regression

Retries the Get-IpAddress and Get-NetAdapter cmdlets a few times as these seem unreliable on WS2022 & WS23H2.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


benjamin-brady (approved):
Looks good to me. You could do an exponential backoff delay.

PR #7257, opened 2025-10-27
error handling

timmy-wright:
**What type of PR is this?**

/kind chore

Saw an odd intermittent error in a previous run of this check so am adding better error handling and logging.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/test/windows-files-check.ps1
+++ b/vhdbuilder/packer/test/windows-files-check.ps1
@@ -84,7 +84,7 @@ function Start-Job-To-Expected-State {
     }
 }
 
-function DownloadFileWithRetry {
+function DownloadFile {
     param (
         $URL,
         $Dest,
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/test/windows-files-check.ps1
```diff
@@ -86,22 +84,26 @@ function Start-Job-To-Expected-State {
...
         $retryDelay = 0,
         [Switch]$redactUrl = $false
     )
-    Write-Output "Downloading $URL"
-    curl.exe --silent -f --retry $retryCount --retry-delay $retryDelay -L $URL -o $Dest
-    if ($LASTEXITCODE) {
+
+    Write-Host "Downloading $URL"
+    try {
+        # We don't retry as that could hide flakey upstream servers that will cause issues in prod.
```
Devinwong, line 98: Why in prod, it doesn't retry the download? What is the concern there?

timmy-wright: In prod it does retry the download. This script is only used in pipeline checks - which I want to fail so we get an idea of upstream server reliability.

Devinwong: Trying to understand further: if it fails without retry but actually it could succeed with 1 retry, is there any action item?

timmy-wright: This is a preference thing. I've had some downloads from the upstream systems fail in prod (even with retrys) and would prefer to see them as a failure in agent baker rather than a flaky upstream being hidden by automated retrys. That way we're likely to notice the signals in AgentBaker PRs and can get them fixed before we have a prod incident.

(yes, I have a production incident for something like this :)

The production code, however, should retry.

PR #7258, opened 2025-10-27
feat: prepare mcr.azk8s.cn migration to mcr.azure.cn - retag

fseldow:
**What type of PR is this?**

/kind feature

mcr.azure.cn is ready, there is plan to deprecate mcr.azk8s.cn in the future

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -706,11 +706,15 @@ retagMCRImagesForChina() {
         return
     fi
     for mcrImage in ${allMCRImages[@]+"${allMCRImages[@]}"}; do
```
cameronmeissner, line 708: would prefer if we could add some unit testing around this

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -476,22 +476,35 @@ testImagesRetagged() {
   pulledImages=($(ctr -n k8s.io image ls))
   mcrImagesNumber=0
   mooncakeMcrImagesNumber=0
+  mooncakeLegacyMcrImagesNumber=0
```
fseldow, line 479: unit test

parts/linux/cloud-init/artifacts/cse_install.sh
fseldow: unit test lays in
https://github.com/Azure/AgentBaker/pull/7258/files#r2467817413

vhdbuilder/packer/windows/list-images.ps1
```diff
@@ -47,7 +47,7 @@ foreach($image in $imageList) {
     } else {
         # Get id and repoTags
         # Example:
-        # mcr.azk8s.cn/windows/servercore:ltsc2022
+        # mcr.azure.cn/windows/servercore:ltsc2022
```
qweeah, line 50: Should we keep the original case of mcr.azk8s.cn?

staging/cse/windows/configfunc.ps1
```diff
@@ -623,25 +623,26 @@ function Retag-ImagesForAzureChinaCloud {
...
-    
+
     $isExist=$false
     $imageList=$(ctr.exe -n k8s.io image ls | select -Skip 1)
     foreach ($imageInfo in $imageList) {
         $splitResult=($imageInfo -split '\s+')
         $image=$splitResult[0]
-        if ($image -like 'mcr.azk8s.cn*') {
+        # TODO(fseldow): remove azk8s when mcr.azk8s.cn is fully deprecated
+        if ($image -like 'mcr.azk8s.cn*' -or $image -like 'mcr.azure.cn*') {
```
AbelHu, line 633: do we expect to add tag `'mcr.azure.cn*'` for old vhds which only have 'mcr.azk8s.cn*'? if so, you need to check them separately.

AbelHu: it seems like that windows does not add tags in building vhd. so you can ignore it if you double confirm it.

vhdbuilder/packer/windows/list-images.ps1
fseldow: i think it is comment change, we do not need to remain the old examples

staging/cse/windows/configfunc.ps1
fseldow: i think we do not need for old vhds, new vhds should be ok.
I do not think we can rebuilt all old vhds

For adding tags logic, they are in line 666

PR #7275, opened 2025-10-28
feat: npd into vhd build

chmill-zz:
**What type of PR is this?**

/kind feature

this is getting npd into agentbaker by default. On the CSE side we'll restart the systemd service to make sure we're running the startup script that handles if the sku is gpu or not. 

Fixes #

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -46,32 +46,7 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 	ValidateRxBufferDefault(ctx, s)
 	ValidateKernelLogs(ctx, s)
 	ValidateScriptlessCSECmd(ctx, s)
-
-	// validate node-problem-detector is running and properly configured
-	// only if the VHD has NPD installed (skip_vhd_npd sentinel file exists)
-	// when old VHDs are used none of this will exist, test will fail and everyone is sad.
-	// So we're relying heavily on the validation in vhdbuild
-	// to make sure the skip_vhd_npd sentinel file exist going forward.
-	if fileExist(ctx, s, "/etc/node-problem-detector.d/skip_vhd_npd") {
-		logf(ctx, "skip_vhd_npd sentinel file found, validating node-problem-detector installation")
-		ValidateSystemdUnitIsRunning(ctx, s, "node-problem-detector.service")
-		ValidateSystemdUnitIsNotFailed(ctx, s, "node-problem-detector")
-		ValidateFileExists(ctx, s, "/etc/node-problem-detector.d/skip_vhd_npd")
-		ValidateFileExists(ctx, s, "/opt/bin/node-problem-detector-startup.sh")
-
-		// validate NPD config directories are installed
-		ValidateDirectoryContent(ctx, s, "/etc/node-problem-detector.d", []string{
-			"custom-plugin-monitor",
-			"plugin",
-			"system-log-monitor",
-			"system-stats-monitor",
-		})
-
-		// validate plugin directory is not empty (contains actual plugin scripts)
-		ValidateNonEmptyDirectory(ctx, s, "/etc/node-problem-detector.d/plugin")
-	} else {
-		logf(ctx, "Skipping node-problem-detector validation: VHD does not have NPD installed (skip_vhd_npd file not found)")
-	}
+	ValidateNodeProblemDetector(ctx, s)
 
 	ValidateSysctlConfig(ctx, s, map[string]string{
 		"net.ipv4.tcp_retries2":             "8",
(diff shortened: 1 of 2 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -354,8 +475,6 @@ while IFS= read -r p; do
           version=$(rpm -q containerd2)
         elif isMarinerOrAzureLinux "$OS"; then
           installStandaloneContainerd "${version}"
-        elif isFlatcar "$OS"; then
-          installStandaloneContainerd "${version}"
```
chewi: Why? As it happens, I will be effectively dropping this in another branch anyway, but I don't know why you've touched this here.

chmill-zz: good catch. I must have removed this accidently when playing locally with flatcar and npd config.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -239,6 +255,111 @@ unpackTgzToCNIDownloadsDIR() {
...
+            echo "Successfully copied and verified skip_vhd_npd sentinel file"
+            break
+        fi
+    done
+
+    # Install AKS startup script and systemd service
+    echo "Installing NPD startup script and systemd service"
+
+    local NPD_STARTUP_SCRIPT_SRC="/home/packer/node-problem-detector-startup.sh"
+    local NPD_STARTUP_SCRIPT_DEST="/usr/local/bin/node-problem-detector-startup.sh"
```
chewi: #7125 is not merged yet, but it should be soon. I am moving nearly everything from /usr/local/bin into /opt/bin because the former is not writeable on immutable distributions like Flatcar and Azure Linux OS Guard. If it works on OS Guard right now, it's because of a hack that should ideally be removed.

e2e/npd/fixtures/testdata/event-logs/npd-test-1761687776430/1761687777018.json
cameronmeissner, line 1: adding more testdata makes us sad, are we sure we need this?

e2e/validation.go
```diff
@@ -45,6 +45,32 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 	ValidateIPTablesCompatibleWithCiliumEBPF(ctx, s)
 	ValidateRxBufferDefault(ctx, s)
 
+	// validate node-problem-detector is running and properly configured
+	// only if the VHD has NPD installed (skip_vhd_npd sentinel file exists)
+	// when old VHDs are used none of this will exist, test will fail and everyone is sad.
+	// So we're relying heavily on the validation in vhdbuild
+	// to make sure the skip_vhd_npd sentinel file exist going forward.
+	if fileExist(ctx, s, "/etc/node-problem-detector.d/skip_vhd_npd") {
+		logf(ctx, "skip_vhd_npd sentinel file found, validating node-problem-detector installation")
```
cameronmeissner: do these tests give us the same level of confidence as the docker-based tests? mainly just wondering if we need to incorporate the docker-based tests into our E2E pipelines or just leave them for local dev

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -336,6 +353,8 @@ while IFS= read -r p; do
           installContainerd "${downloadDir}" "${evaluatedURL}" "${version}"
         elif isMarinerOrAzureLinux "$OS"; then
           installStandaloneContainerd "${version}"
+        elif isFlatcar "$OS"; then
+          installStandaloneContainerd "${version}"
```
cameronmeissner, line 406: is this change intended?

.gitignore
```diff
@@ -41,7 +42,6 @@ hack/tools/bin
 # to make sure we don't accidently add this back after the recent
 # migration off of go-bindata
 pkg/templates/templates_generated.go
-e2e/scenario-logs/
```
cameronmeissner, line 44: this definitely should stay

parts/linux/cloud-init/artifacts/node-problem-detector-startup.sh
```diff
@@ -0,0 +1,371 @@
...
+        echo "Error: container-runtime-endpoint is not set"
+        exit 1
+    fi
+    CONTAINER_RUNTIME_ENDPOINT=$(echo "${KUBELET_ARGS##*--container-runtime-endpoint=}" | cut -f1 -d" " | tr -d "\"")
+    export CONTAINER_RUNTIME_ENDPOINT
+}
+
+function get_sku() {
+    # Azure Instance Metadata Service endpoint for compute information.
+    IMDS_ENDPOINT="http://169.254.169.254/metadata/instance/compute?api-version=2021-02-01"
```
cameronmeissner, line 129: we can use the cache file like we do in CSE right?

PR #7276, opened 2025-10-29
Add a new validation for e2e to catch errors regarding exec command on a pod

Devinwong:
**What type of PR is this?**
/kind test

Add a new validation for e2e to catch errors regarding exec command on a pod. This will allow us to catch unexpected errors such as an issue caused by Runc 1.3.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -702,6 +702,8 @@ func Test_Ubuntu2204_Scriptless(t *testing.T) {
 			VHD:     config.VHDUbuntu2204Gen2Containerd,
 			Validator: func(ctx context.Context, s *Scenario) {
 				ValidateFileHasContent(ctx, s, "/var/log/azure/aks-node-controller.log", "aks-node-controller finished successfully")
+				// Validate exec command `sh` works on a specific test pod
+				ValidateExecCmdOnVM(ctx, s, "sh")
 			},
 			AKSNodeConfigMutator: func(config *aksnodeconfigv1.Configuration) {
 			},
@@ -717,6 +719,8 @@ func Test_Ubuntu2404_Scriptless(t *testing.T) {
 			VHD:     config.VHDUbuntu2404Gen2Containerd,
 			Validator: func(ctx context.Context, s *Scenario) {
 				ValidateFileHasContent(ctx, s, "/var/log/azure/aks-node-controller.log", "aks-node-controller finished successfully")
+				// Validate exec command `sh` works on a specific test pod
+				ValidateExecCmdOnVM(ctx, s, "sh")
 			},
 			AKSNodeConfigMutator: func(config *aksnodeconfigv1.Configuration) {
 			},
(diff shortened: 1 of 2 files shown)


e2e/validators.go
```diff
@@ -1201,3 +1201,64 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
 	require.NotContains(s.T, stdout, "No MIG-enabled devices found", "no MIG devices were created.\nOutput:\n%s", stdout)
 	s.T.Logf("MIG instances with profile %s are created", migProfile)
 }
+
+func execOnPodForScenario(ctx context.Context, s *Scenario, pod *corev1.Pod, cmd string) *podExecResult {
```
cameronmeissner, line 1205: we already have a number of funcs that do this, I don't really see the value in adding another super thin helper like this

e2e/validators.go
```diff
@@ -1201,3 +1201,64 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
...
+	s.T.Helper()
+	execResult, err := execOnUnprivilegedPod(ctx, s.Runtime.Cluster.Kube, pod.Namespace, pod.Name, cmd)
+	require.NoErrorf(s.T, err, "failed to execute command on pod: %v", cmd)
+	return execResult
+}
+
+// ValidateExecCmdOnVM creates a pod on the target node and executes the given command on that pod.
+func ValidateExecCmdOnVM(ctx context.Context, s *Scenario, command string) {
```
cameronmeissner, line 1213: nit: `ValidateExecCMDOnVM`

cameronmeissner, line 1213: don't we already create a testing pod on the e2e nodes? could we just look at adding a readiness/liveness exec probe to that pod? in which case no code changes would even be required

r2k1: It's gettings quite confusing. We have too many helpers to run a command on VM/Pod.
I think existing helpers already covers it? We just run things on a pod from a daemonset rather than explicitly creating one. (Though I prefer this way over daemonset, just don't want to duplicate things)

r2k1: We also have `ValidateNodeCanRunAPod` that seems to do something similar?

Should we improve the check if something is missing?

r2k1: `podHTTPServerLinux` seems to attempt to do the job, but doing it poorly?

I think we should just add probes there? I think I added them at some stage, but now they gone and I'm not sure why.

Devinwong: I don't know why yet. But not every base image can catch this issue. The one we were able to catch that issue is `nginx`. Let me check if `podHTTPServerLinux` can catch it.

Devinwong: 2 days ago I was still able to repro this with `nginx` latest image. But today I tried with the latest, last 2 latest but it no longer reproduces. The VHD is using Runc 1.3 which was the one having issue. Not sure why changed between. I will close this PR for now until it can be repro'd again.

PR #7277, opened 2025-10-29
bug: add DCGM CUDA13-compatible packages to GB200 image. Also add multinode DCGM package.

abenn135:
**What type of PR is this?**

/kind bug
/kind regression

Updates DGCM package to be compatible with cuda13, which is shipped on GB200 nodes.

not tracked

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
@@ -6,7 +6,7 @@ removeContainerd() {
 
 installDeps() {
     wait_for_apt_locks
-    retrycmd_silent 120 5 300 curl -fsSL https://packages.microsoft.com/config/ubuntu/${UBUNTU_RELEASE}/packages-microsoft-prod.deb > /tmp/packages-microsoft-prod.deb || exit $ERR_MS_PROD_DEB_DOWNLOAD_TIMEOUT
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -6,7 +6,7 @@ removeContainerd() {
 
 installDeps() {
     wait_for_apt_locks
-    retrycmd_silent 120 5 25 curl -fsSL https://packages.microsoft.com/config/ubuntu/${UBUNTU_RELEASE}/packages-microsoft-prod.deb > /tmp/packages-microsoft-prod.deb || exit $ERR_MS_PROD_DEB_DOWNLOAD_TIMEOUT
+    retrycmd_silent 120 5 300 curl -fsSL https://packages.microsoft.com/config/ubuntu/${UBUNTU_RELEASE}/packages-microsoft-prod.deb > /tmp/packages-microsoft-prod.deb || exit $ERR_MS_PROD_DEB_DOWNLOAD_TIMEOUT
```
keith-ms: 300 seconds seems high. Any reason this can't be one minute?

abenn135: So, during the outage recovery on Wednesday, I was seeing 30-90s delays talking to packages.microsoft.com. I wanted to give myself some headroom. But it's true that

a) this is pretty unusual, and
b) if there's a bona fide configuration problem, we want to give up before wasting too much time.

I'll lower this to 90s and we can make a patch release if the AFD issues recur.

PR #7284, opened 2025-10-29
components: Update DCGM exporter version for AzL3 to 4.6.0

surajssd:
**What type of PR is this?**

/kind test
/kind cleanup

- Update dcgm-exporter version of the azure linux to match the Ubuntu ones.
- Add a test to ensure that the versions are always consistent.

### DIFF
--- a/.github/workflows/validate-components.yml
+++ b/.github/workflows/validate-components.yml
@@ -1,7 +1,7 @@
 name: Validate Components
 on: pull_request
 
-jobs: 
+jobs:
   cue:
     runs-on: ubuntu-latest
     environment: test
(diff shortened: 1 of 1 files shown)


e2e/scenario_gpu_managed_experience_test.go
```diff
@@ -29,6 +30,81 @@ func getDCGMPackageNames(os string) []string {
...
+	return ""
+}
+
+type packageOSVariant struct {
+	pkgName   string
+	osName    string
+	osRelease string
+}
+
+func Test_Version_Consistency_GPU_Managed_Components(t *testing.T) {
```
ganeshkumarashok, line 60: Should we keep this in a separate file (and maybe folder) since this is not an E2E test per se? More like component validation

ganeshkumarashok (approved):
Lgtm besides moving it to a separate file/folder instead of e2e

PR #7290, opened 2025-10-30
feat(flatcar): disable autologin

ervcz:

This PR adds removes the kernel argument which enables autologin and adds a shell test to check that the kernel parameter is missing.

Fixes [33744380](https://msazure.visualstudio.com/CloudNativeCompute/_boards/board/t/Flatcar%20for%20AKS/Stories?workitem=33744380).

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Test carried out**:

- [x] Run the test without the butane config which would disable the autologin to test that the test fails
- [x] Apply the kernel option in the config to disable autologin and run `make -f packer.mk generate-flatcar...`
- [x] Create a new build and run the tests again, now all tests should pass

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -9,6 +9,7 @@ MARINER_KATA_OS_NAME="MARINERKATA"
 AZURELINUX_KATA_OS_NAME="AZURELINUXKATA"
 
 THIS_DIR="$(cd "$(dirname ${BASH_SOURCE[0]})" && pwd)"
+
 OS_VERSION="$1"
 ENABLE_FIPS="$2"
 OS_SKU="$3"
@@ -634,6 +635,42 @@ testLtsKernel() {
 
 }
 
+# Parse loginctl sessions to find console autologin sessions
+# Console autologin sessions have Remote=no and Service=login
+# Returns: Space-separated list of autologin session IDs
+parseAutologinSessions() {
+  loginctl list-sessions --no-legend 2>/dev/null | while read -r session_id rest; do
+    if [ -n "$session_id" ]; then
+      local session_info
+      session_info=$(loginctl show-session "$session_id" 2>/dev/null || true)
+
+      if [ -z "$session_info" ]; then
+        echo "ERROR: 'loginctl show-session' Command may have failed or session may not exist" >&2
+        echo "PARSE_ERROR:$session_id"
+        continue
+      fi
+
+      local is_remote=$(echo "$session_info" | sed -n "s/^Remote=//p")
+      local service=$(echo "$session_info" | sed -n "s/^Service=//p")
+
+      # Check if we can parse the required fields
+      if [ -z "$is_remote" ] && [ -z "$service" ]; then
+        echo "ERROR: Cannot parse loginctl output for session $session_id" >&2
+        echo "ERROR: loginctl show-session output format may have changed" >&2
+        echo "ERROR: Expected 'Remote=' and 'Service=' fields but found neither" >&2
+        echo "ERROR: Actual output: $session_info" >&2
+        echo "PARSE_ERROR:$session_id"
+        continue
+      fi
+
+      # Console autologin sessions have Remote=no and Service=login
+      if [ "$is_remote" = "no" ] && [ "$service" = "login" ]; then
+        echo "$session_id"
+      fi
+    fi
+  done
+}
+
 testAutologinDisabled() {
   local test="testAutologinDisabled"
   local os_sku=$1
@@ -648,21 +685,15 @@ testAutologinDisabled() {
     echo "$test: Checking for console autologin sessions using loginctl"
 
     local autologin_session_ids=""
-    autologin_session_ids=$(loginctl list-sessions --no-legend 2>/dev/null | while read -r session_id rest; do
-      if [ -n "$session_id" ]; then
-        local session_info
-        session_info=$(loginctl show-session "$session_id" 2>/dev/null || true)
-        local is_remote=$(echo "$session_info" | sed -n "s/^Remote=//p")
-        local service=$(echo "$session_info" | sed -n "s/^Service=//p")
+    autologin_session_ids=$(parseAutologinSessions)
 
-        # Console autologin sessions have Remote=no and Service=login
-        if [ "$is_remote" = "no" ] && [ "$service" = "login" ]; then
-          echo "$session_id"
-        fi
-      fi
-    done)
-
-    if [ -n "$autologin_session_ids" ]; then
+    # Check for parse errors (indicates loginctl output format changed)
+    if echo "$autologin_session_ids" | grep -q "^PARSE_ERROR:"; then
+      err $test "Failed to parse loginctl output - format may have changed"
+      echo "$test: Parse errors detected in session parsing" >&2
+      echo "$test: Sessions with errors: $autologin_session_ids" >&2
+      failed=1
+    elif [ -n "$autologin_session_ids" ]; then
       err $test "Found console autologin session(s) with Remote=no and Service=login"
       echo "$test: Autologin session IDs: $autologin_session_ids" >&2
       echo "$test: All sessions:" >&2
--- a/spec/vhdbuilder/packer/test/linux_vhd_content_test_helpers_spec.sh
+++ b/spec/vhdbuilder/packer/test/linux_vhd_content_test_helpers_spec.sh
@@ -1,0 +1,201 @@
+#!/bin/bash
+# shellcheck disable=SC2329
+
+# ShellSpec tests for parseAutologinSessions helper function
+
+Describe 'parseAutologinSessions helper function'
+  # Extract only the function definition using sed - clean approach!
+  BeforeAll "eval \"\$(sed -n '/^parseAutologinSessions()/,/^}/p' './vhdbuilder/packer/test/linux-vhd-content-test.sh')\""
+
+  Describe 'parseAutologinSessions'
+    It 'returns empty when no sessions exist'
+      loginctl() {
+        if [ "$1" = "list-sessions" ]; then
+          echo ""
+        fi
+      }
+      When call parseAutologinSessions
+      The output should equal ""
+    End
+
+    It 'returns empty when only SSH sessions exist (Remote=yes)'
+      loginctl() {
+        case "$1" in
+          "list-sessions")
+            echo "1 1000 user seat0"
+            ;;
+          "show-session")
+            cat <<EOF
+Remote=yes
+Service=sshd
+Type=tty
+EOF
+            ;;
+        esac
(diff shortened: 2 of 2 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,28 @@ testLtsKernel() {
 
 }
 
+testAutologinDisabled() {
+  test="testAutologinDisabled"
```
chewi:
```suggestion
  local test="testAutologinDisabled"
```

I'm adding `local` to all these in my branch, otherwise the output can be very confusing.

ervcz: Thanks! The log was indeed confusing. Curious to see how this improves it :)

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,28 @@ testLtsKernel() {
...
 
+testAutologinDisabled() {
+  local test="testAutologinDisabled"
+  os_sku=$1
+  echo "$test:Start"
+
+  if [ "${os_sku}" = "Flatcar" ]; then
+    echo "$test: Checking kernel command line for flatcar.autologin parameter"
+
+    if grep -q "flatcar.autologin" /proc/cmdline; then
```
johananl, line 719: From a security standpoint it would have probably been better to test that we actually **can't** log in without a password on the console (i.e. test the behavior rather than the code). But it probably isn't trivial to automate that.

chewi: Yeah, I was initially expecting that, but I'm not even sure how I'd actually go about it.

johananl: Can probably be done by running something like `expect` against the serial console, but I don't know if it's worth the effort :-/

chewi: Actually, autologin doesn't work like that. There is no login prompt, it just drops you straight to bash. If you were going to run `expect`, it would be to see if the prompt was there at all, but an easier way to check would be to run `sudo who` and see if more than one entry (the SSH session used by the test) appears. Also, the `getty` process runs while there is an active prompt, but not when autologin is enabled.

johananl: >it would be to see if the prompt was there at all

Yes, I meant this :slightly_smiling_face:

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,73 @@ testLtsKernel() {
...
+  echo "$test:Start"
+
+  if [ "${os_sku}" = "Flatcar" ]; then
+    local failed=0
+
+    # Test 1: Check actual behavior - verify no console autologin sessions exist
+    echo "$test: Checking for console autologin sessions using 'who'"
+    # Filter for console ttys (ttyS*, tty1, console) but exclude pts (SSH sessions)
+    local console_sessions
+    console_sessions=$(who | grep -E 'ttyS|tty1|console' | grep -v pts || true)
```
chewi: I don't think this works. When I try QEMU with or without `-nographic`, the autologin session gives me `pts/0`. I must admit that I find this surprising.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,73 @@ testLtsKernel() {
...
+    else
+      echo "$test: No console autologin sessions found (behavioral check passed)"
+    fi
+
+    # Test 2: Check getty processes - should be waiting for login, not auto-logged-in
+    echo "$test: Checking getty process status"
+    # If autologin is enabled, getty completes immediately (no waiting process)
+    # If autologin is disabled, getty should be running and waiting for login
+    local getty_status
+    getty_status=$(systemctl is-active serial-getty@ttyS0.service 2>/dev/null || echo "inactive")
```
chewi: I'd have just checked `pgrep getty`, but okay.

vhdbuilder/packer/test/linux-vhd-content-test.sh
ervcz: Thanks for the suggestions @johananl, @chewi. I agree, testing the actual behavior too makes the test more secure, so I extended the test with two more checks: one using `who` and the other verifying the `getty` status.

vhdbuilder/packer/test/linux-vhd-content-test.sh
ervcz: I will use loginctl instead of who. That specifies if a session is remote or not, so we won't have to rely on assuming session names only.

chewi: Annoyingly `list-sessions` doesn't seem to show that, even in its JSON output, although it does show the true TTY. `show-session` shows it, so I guess you'll need to iterate over the sessions. No JSON there, but easy enough to grep.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,80 @@ testLtsKernel() {
...
+    # Without autologin: we should see only SSH sessions (Remote=yes) or no sessions
+    echo "$test: Checking for console autologin sessions using loginctl"
+
+    local autologin_session_ids=""
+    autologin_session_ids=$(loginctl list-sessions --no-legend 2>/dev/null | while read -r session_id rest; do
+      if [ -n "$session_id" ]; then
+        local session_info
+        session_info=$(loginctl show-session "$session_id" 2>/dev/null || true)
+        local is_remote=$(echo "$session_info" | grep "^Remote=" | cut -d= -f2)
+        local service=$(echo "$session_info" | grep "^Service=" | cut -d= -f2)
```
chewi: This is just personal taste, you can decide.

```suggestion
        local is_remote=$(sed -n "s/^Remote=//p" <<< "$session_info")
        local service=$(sed -n "s/^Service=//p" <<< "$session_info")
```

chewi: shellcheck might complain about `<<<` being a bashism. If so:

```suggestion
        local is_remote=$(echo "$session_info" | sed -n "s/^Remote=//p")
        local service=$(echo "$session_info" | sed -n "s/^Service=//p")
```

I don't know why they don't just assume bash.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,80 @@ testLtsKernel() {
 
 }
 
+testAutologinDisabled() {
+  local test="testAutologinDisabled"
+  os_sku=$1
```
timmy-wright: Should this be "local" too?

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,80 @@ testLtsKernel() {
...
+    if grep -q "flatcar.autologin" /proc/cmdline; then
+      err $test "flatcar.autologin kernel parameter found in /proc/cmdline but should be absent for AKS security compliance"
+      echo "$test: Full kernel command line:" >&2
+      cat /proc/cmdline >&2
+      failed=1
+    else
+      echo "$test: flatcar.autologin parameter correctly absent from kernel command line"
+    fi
+
+    if [ $failed -eq 1 ]; then
```
timmy-wright, line 728: Probably easier to abort when the error occurs rather than having a "failed" variable. Then you can include a little more information in the error message

vhdbuilder/packer/test/linux-vhd-content-test.sh
timmy-wright: In fact, there are a few variables in this function that should be marked as local.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -634,6 +634,80 @@ testLtsKernel() {
 
 }
 
+testAutologinDisabled() {
```
timmy-wright, line 674: It'd be great if we could have unit tests on this function. I know it's testing a test file, but there is complex logic in the function. However, I can't see any existing tests of this file, so it's probably not worth figuring out how to test the file for this PR.

vhdbuilder/packer/test/linux-vhd-content-test.sh
ervcz: The reason to use the failed variable is to collect the info from all the checks. I thought it could help with debugging as in case of an error you would get immediate confirmation from the other two. These are the results I got after after local testing:

Successful:
```
testAutologinDisabled:Start
testAutologinDisabled: Checking for console autologin sessions using loginctl
testAutologinDisabled: No console autologin sessions found
testAutologinDisabled: Checking for agetty processes
testAutologinDisabled: Found agetty processes waiting for login (autologin disabled, correct)
testAutologinDisabled: Checking kernel command line for flatcar.autologin parameter
testAutologinDisabled: flatcar.autologin parameter correctly absent from kernel command line
testAutologinDisabled: PASSED - autologin is correctly disabled
testAutologinDisabled:Finish
``` 

Failed (second check shows the kernel parameter for autologin was detected, while the other checks verify the actual behavior):
```
testAutologinDisabled:Start
testAutologinDisabled: Checking for console autologin sessions using loginctl
testAutologinDisabled:Error: Found console autologin session(s) with Remote=no and Service=login
testAutologinDisabled: Autologin session IDs: 1
2
testAutologinDisabled: All sessions:
testAutologinDisabled: Checking for agetty processes
testAutologinDisabled:Error: No agetty processes found - getty may have completed due to autologin
testAutologinDisabled: Checking kernel command line for flatcar.autologin parameter
testAutologinDisabled:Error: flatcar.autologin kernel parameter found in /proc/cmdline but should be absent for AKS security compliance
testAutologinDisabled: Full kernel command line:
rootflags=rw mount.usrflags=ro BOOT_IMAGE=/flatcar/vmlinuz-a mount.usr=/dev/mapper/usr verity.usr=PARTUUID=7130c94a-213a-4e5a-8e26-6cce9662f132 rootflags=rw mount.usrflags=ro consoleblank=0 root=LABEL=ROOT consol3
testAutologinDisabled: FAILED - autologin security check failed
```

vhdbuilder/packer/test/linux-vhd-content-test.sh
ervcz: I quickly made an attempt to create a test file for linux-vhd-content-test.sh but I found that I would need to source/mock many things.

In my second attempt, (that's what you can see in my latest commit as well) I factored out the parsing part and wrote a test only for that. Tell me your opinion about it :)

IMO having three different checks for the same thing should already give us confidence.

timmy-wright (approved):

PR #7294, opened 2025-10-31
chore: adding function to allow azcli override in the future

djsly:
**What type of PR is this?**
/kind cleanup

Removing hardcoded AzCli version that we don't need anymore

### DIFF
--- a/vhdbuilder/packer/produce-packer-settings-functions.sh
+++ b/vhdbuilder/packer/produce-packer-settings-functions.sh
@@ -1,5 +1,20 @@
 #!/bin/bash
 
+function enforce_azcli_version() {
+	AZ_VER_REQUIRED=$1
+	AZ_DIST=$(lsb_release -cs)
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/produce-packer-settings-functions.sh
```diff
@@ -1,5 +1,20 @@
 #!/bin/bash
 
+function enforce_azcli_version() {
+	AZ_VER_REQUIRED=$1
```
awesomenix, line 4: Do you need this `wait_for_apt_locks`

Devinwong: yeah I think we need it.

djsly: @awesomenix this is to avoid having failure where the apg-get command fails since there is another apg command ongoing.  most probably you are right since this is a sequencial things.  but I think there is auto apt-get command getting excuted on new node bootup

PR #7310, opened 2025-11-04
Extract all coredns binaries for localdns

SriHarsha001:
**What type of PR is this?**
feature

This PR updates updates the code to extract CoreDNS binaries from all cached coredns images and copies them to organized directories:
- Latest version (n): /opt/azure/containers/localdns/binary/coredns
- Previous version (n-1): /opt/azure/containers/localdns/binary/version/coredns,
- Older version (n-2): /opt/azure/containers/localdns/binary/version/coredns, etc.

This is done so that we can send in a flag to use one of the binaries for localdns. Default will be the latest binary, but if there is a bug/regression in upstream coredns or CVEs, we will have an option to quickly fallback to previous binaries. 

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -900,16 +900,16 @@ extractAndCacheCoreDnsBinary() {
   local coredns_image_list=($(ctr -n k8s.io images list -q | grep coredns))
   if [ "${#coredns_image_list[@]}" -eq 0 ]; then
     echo "Error: No coredns images found."
-    exit 1
+    return 1
   fi
 
   # Clean up existing binary directories.
   if [ -d "${LOCALDNS_BINARY_PATH}" ]; then
-    rm -rf "${LOCALDNS_BINARY_PATH}" || exit 1
+    rm -rf "${LOCALDNS_BINARY_PATH}" || return 1
   fi
 
   # Ensure the main localdns directory and binary path exist.
-  mkdir -p "${LOCALDNS_BINARY_PATH}" || exit 1
+  mkdir -p "${LOCALDNS_BINARY_PATH}" || return 1
 
   local ctr_temp=""
   cleanup_coredns_imports() {
@@ -957,7 +957,7 @@ extractAndCacheCoreDnsBinary() {
 
     if [ "$retry_count" -eq "$max_retries" ]; then
       echo "Error: Failed to mount ${coredns_image_url} after ${max_retries} attempts." >> "${VHD_LOGS_FILEPATH}"
-      exit 1
+      return 1
     fi
 
     # This path is based on the standard CoreDNS image structure.
@@ -969,19 +969,19 @@ extractAndCacheCoreDnsBinary() {
       if [ $version_index -eq 0 ]; then
         cp "${coredns_binary}" "${LOCALDNS_BINARY_PATH}/coredns" || {
           echo "Error: Failed to copy latest coredns binary to ${LOCALDNS_BINARY_PATH}/coredns" >> "${VHD_LOGS_FILEPATH}"
-          exit 1
+          return 1
         }
 
         # Validate that the main binary was extracted successfully.
         if [ ! -f "${LOCALDNS_BINARY_PATH}/coredns" ]; then
           echo "Error: Latest coredns binary not found at ${LOCALDNS_BINARY_PATH}/coredns after extraction" >> "${VHD_LOGS_FILEPATH}"
-          exit 1
+          return 1
         fi
 
-        chmod +x "${LOCALDNS_BINARY_PATH}/coredns" || exit 1
+        chmod +x "${LOCALDNS_BINARY_PATH}/coredns" || return 1
         if [ ! -x "${LOCALDNS_BINARY_PATH}/coredns" ]; then
           echo "Error: Latest coredns binary at ${LOCALDNS_BINARY_PATH}/coredns is not executable" >> "${VHD_LOGS_FILEPATH}"
-          exit 1
+          return 1
         fi
 
         # Verify the binary is functional.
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -903,13 +903,24 @@ extractAndCacheCoreDnsBinary() {
...
 
-  rm -rf "${LOCALDNS_BINARY_PATH}" || exit 1
+  # Clean up existing binary directories only (preserve other files like localdns.sh)
+  if [ -d "${LOCALDNS_BINARY_PATH}" ]; then
+    rm -rf "${LOCALDNS_BINARY_PATH}" || exit 1
+  fi
+
+  # Remove any existing n-X versioned directories
+  if [ -d "/opt/azure/containers/localdns" ]; then
+    find "/opt/azure/containers/localdns" -maxdepth 1 -type d -name "n-*" -exec rm -rf {} + 2>/dev/null || true
```
yewmsft: what are the directory structure under /opt/azure/containers/localdns?
n-1, n-2, ..., n-x
Is there a n?

yewmsft: oh, i c. n is under /opt/azure/containers/localdns/binary/coredns

Devinwong: I was thinking the same thing. Maybe consider using the ones defined in components.json? `latest` and `previouslatest`? But if they are just existing temporarily, then it's fine.

SriHarsha001: Latest version will be without 'n' in the binary path - /opt/azure/containers/localdns/binary/coredns. I am doing this so that existing clusters and existing code works as is. Previous version will have 'n-1' and older version will have 'n-2'.

SriHarsha001: yeah, I had considered 'latest', 'previouslatest' earlier, but I noticed that the syntax in components.json file is not consistent and if we make any mistake in the file, then the binary extraction logic will fail.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -929,35 +940,53 @@ extractAndCacheCoreDnsBinary() {
...
-  local latest_coredns_tag="${sorted_coredns_tags[0]}"
-  # Extract major.minor.patch (removes -revision. eg - v1.12.0).
-  local latest_vMajorMinorPatch="${latest_coredns_tag%-*}"
+  # Group versions by major.minor.patch to get unique versions
+  declare -A unique_versions
+  local version_order=()
 
-  local previous_coredns_tag=""
-  # Iterate through the sorted list to find the next highest major-minor version.
   for tag in "${sorted_coredns_tags[@]}"; do
```
yewmsft: where is sorted_coredns_tags from?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -903,13 +903,24 @@ extractAndCacheCoreDnsBinary() {
     exit 1
   fi
 
-  rm -rf "${LOCALDNS_BINARY_PATH}" || exit 1
+  # Clean up existing binary directories only (preserve other files like localdns.sh)
+  if [ -d "${LOCALDNS_BINARY_PATH}" ]; then
+    rm -rf "${LOCALDNS_BINARY_PATH}" || exit 1
```
Devinwong: I'd suggest not to exit 1 directly in a function, although I understand your function is less likely be called by others.
I'd suggest `return 1` here and then at the caller you exit with 1, if that makes sense to you too. This applies to other exit 1 in this function

vhdbuilder/packer/install-dependencies.sh
yewmsft: nvm, found it

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -889,12 +889,12 @@ if [ -n "${PRIVATE_PACKAGES_URL:-}" ]; then
...
-
 LOCALDNS_BINARY_PATH="/opt/azure/containers/localdns/binary"
-# This function extracts CoreDNS binary from cached coredns images (n-1 image version and latest revision version)
-# and copies it to - /opt/azure/containers/localdns/binary/coredns.
-# The binary is later used by localdns systemd unit.
+# This function extracts CoreDNS binaries from all cached coredns images
+# and copies them to organized directories:
+# - Latest version: /opt/azure/containers/localdns/binary/coredns
+# - Previous versions: /opt/azure/containers/localdns/n-1/binary/coredns,
+# /opt/azure/containers/localdns/n-2/binary/coredns, etc.
```
Devinwong: for the coredns binaries stored in n-1 and n-2 folders, what is the purpose?

Devinwong: I see it in the pr description now

vhdbuilder/packer/install-dependencies.sh
SriHarsha001: This variable is defined a few lines earlier. 

 Extract available coredns image tags (v1.12.0-1 format) and sort them in descending order.
  local sorted_coredns_tags=($(for image in "${coredns_image_list[@]}"; do echo "${image##*:}"; done | sort -V -r))

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -903,13 +903,24 @@ extractAndCacheCoreDnsBinary() {
     exit 1
```
yewmsft: can we download?

yewmsft: instead of failing

vhdbuilder/packer/install-dependencies.sh
SriHarsha001: Yes, sure I can do that. Capture the return value from the function and then exit with non-zero ?

vhdbuilder/packer/install-dependencies.sh
SriHarsha001: By default we will use the latest coredns binary for localdns after this change. If we need to revert back to n-1 or n-2 binaries for localdns, we will use the binaries from this path. That decision will come from AKS RP via a flag. This will enable us to fallback easily and faster.

vhdbuilder/packer/install-dependencies.sh
yewmsft: oh, this is during vhd building. I think we are ok

SriHarsha001: Ah yes sure sure.

vhdbuilder/packer/install-dependencies.sh
yewmsft: Consider keeping the latest under /opt/azure/containers/localdns/binary/coredns, and then put all versions, including the latest version under /opt/azure/containers/localdns/[version], instead of n-1, n-2, etc. Because in real world, nodes will be on different image versions, and each n-1, n-2 on each nodes will contain different binary versions, which will make it very hard to manage. Calling out the version number in the directory name makes it much easier

SriHarsha001: Yeah, sure I can do that, but my concern is - I will have to send a specific coredns version to be used by localdns from AKS RP when we need to fallback. And that version may not be available on the node (if the node is using an older image). 

If we send 'n', 'n-1' or 'n-2' from AKS RP, localdns can use whatever corresponding version (n or n-1 or n-2) is available on the node and nothing breaks. If customer is using an older image, customer will have to update the node image. My idea was to use whatever binary version (n, n-1,n-2) is available in the VHD, instead of tying it down with a specific version. 

Please let me know your thoughts.

yewmsft: Good question, actually this is more of a design question, what functionalities do we want to achieve later on. I have some ideas. One of them is integrating with security patch channel to do localdns patching. We can discuss about those offline.

SriHarsha001: I have updated the code to save previous coredns binaries in /opt/azure/containers/localdns/binary/version/coredns

image

vhdbuilder/packer/install-dependencies.sh
SriHarsha001: I have updated the code to save previous coredns binaries in /opt/azure/containers/localdns/binary/version/coredns

image

vhdbuilder/packer/install-dependencies.sh
Devinwong: correct.

SriHarsha001: Ok, made the suggested changes.

PR #7312, opened 2025-11-04
feat: add fallback to kube binary install if cached deb/rpm not found

lilypan26:
**What type of PR is this?**

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -266,20 +266,26 @@ installPkgWithAptGet() {
...
+            return 0
+        fi
+
         # query all package versions and get the latest version for matching k8s version
         updateAptWithMicrosoftPkg
         fullPackageVersion=$(apt list ${packageName} --all-versions | grep ${packageVersion}- | awk '{print $2}' | sort -V | tail -n 1)
         if [ -z "${fullPackageVersion}" ]; then
             echo "Failed to find valid ${packageName} version for ${packageVersion}"
-            exit 1
+            return 1
```
djsly, line 280: i traced back this return1, and I can't seem to find where this would cause an issue. I think this return code isn't used anywhere ?

Devinwong: I think the return value here is not used as exit code. It's just the let the caller knows it fails and the caller will exit with proper exit code.
e.g.,
```
installPkgWithAptGet "kubelet" "${k8sVersion}" || exit $ERR_KUBELET_INSTALL_FAIL
installPkgWithAptGet "kubectl" "${k8sVersion}" || exit $ERR_KUBECTL_INSTALL_FAIL
installPkgWithAptGet "azure-acr-credential-provider" "${packageVersion}" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
```

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -911,6 +911,24 @@ getLatestPkgVersionFromK8sVersion() {
     echo $PACKAGE_VERSION
 }
 
+# installs kube binary if cached
+fallbackToKubeBinaryInstall() {
```
Devinwong, line 915: nit: could you add some tests in cse_helpers_spec.sh to cover this function? (or ask AI to do that 😄)

djsly (approved):

parts/linux/cloud-init/artifacts/cse_helpers.sh
lilypan26: not sure if worth adding a testing for this since main purpose of this test is to check whether the binary version is cached with `-f /usr/local/bin/${packageName}-${packageVersion}`, and we can't really test that

Devinwong: okay. No problem.

Devinwong (approved):

PR #7313, opened 2025-11-04
Fix windows network configuration for VHD caching feature

r2k1:
**What type of PR is this?**

/kind bug

Move network configuration from "basePrep" to "nodePrep" phase. 

Currently network configuration (including IP address configuration) is baked into VHD for caching feature. Multiple VMs created from such VHD attempts to use same IP.  A node in this case joins k8s cluster, but pods are failing on such node.

### DIFF
--- a/pkg/agent/testdata/AKSWindows2019+CustomCloud+ootcredentialprovider/CustomData
+++ b/pkg/agent/testdata/AKSWindows2019+CustomCloud+ootcredentialprovider/CustomData
@@ -411,6 +411,50 @@ function BasePrep {
         New-HostsConfigService
     }
 
+    Set-Explorer
+    Adjust-PageFileSize
+    Logs-To-Event -TaskName "AKS.WindowsCSE.PreprovisionExtension" -TaskMessage "Start preProvisioning script"
+    
+    Update-ServiceFailureActions
+    Adjust-DynamicPortRange
+    Register-LogsCleanupScriptTask
+    Register-NodeResetScriptTask
+
+    Update-DefenderPreferences
+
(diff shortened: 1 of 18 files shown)


e2e/validation.go
```diff
@@ -97,10 +98,18 @@ func ValidateCommonLinux(ctx context.Context, s *Scenario) {
 	_ = execScriptOnVMForScenarioValidateExitCode(ctx, s, "sudo curl http://168.63.129.16:32526/vmSettings", 0, "curl to wireserver failed")
 
 	execResult = execOnVMForScenarioOnUnprivilegedPod(ctx, s, "curl https://168.63.129.16/machine/?comp=goalstate -H 'x-ms-version: 2015-04-05' -s --connect-timeout 4")
-	require.Equal(s.T, "28", execResult.exitCode, "curl to wireserver should fail")
+	if !assert.Equal(s.T, "28", execResult.exitCode, "curl to wireserver expected to fail, but it didn't") {
+		// Log debug information. This validator seems to be flaky, hard to catch
+		s.T.Logf("host IPTABLES: %s", execScriptOnVMForScenario(ctx, s, "sudo iptables -t filter -L FORWARD -v -n --line-numbers").String())
```
djsly, line 103: should we get / print hte logs from the service that it supposed to apply those settings ?

r2k1: Oh, this change supposed to be in another PR #7281

r2k1: It's probably nice to have, but I don't know how to do it on top of my mind

timmy-wright (approved):

PR #7319, opened 2025-11-05
fix: installation of Azure ACR Credential Provider on Flatcar

chewi:
**What type of PR is this?**

/kind bug

Flatcar is currently falling into an `else` (Ubuntu) clause when installing the 1.34 credential provider, but it cannot install PMC packages using a package manager. Work around this for now by forcing a URL-based installation in this case.

Fixes [Flatcar for AKS work item #35564514](https://msazure.visualstudio.com/CloudNativeCompute/_boards/board/t/Flatcar%20for%20AKS/Stories?workitem=35564514)

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

Flatcar will use sysexts later, but that work isn't quite ready yet. #7125 is the first part of that work.


parts/common/components.json
```diff
@@ -1456,6 +1456,38 @@
             ],
             "downloadURL": "mcr.microsoft.com/oss/binaries/kubernetes/azure-acr-credential-provider:${version}-linux-${CPU_ARCH}"
           }
+        },
+        "flatcar": {
+          "current": {
+            "versionsV2": [
+              {
+                "k8sVersion": "1.30",
+                "renovateTag": "OCI_registry=https://mcr.microsoft.com, name=oss/binaries/kubernetes/azure-acr-credential-provider",
```
djsly, line 1465: shouldnt we download the /v2/ versions ?

djsly: ah those are the older OCI binaries

PR #7342, opened 2025-11-10
feat: add mooncake mcr containerd config host - linux

fseldow:
**What type of PR is this?**

/kind feature

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -1856,7 +1856,7 @@ func Test_Ubuntu2404Gen2_McrChinaCloud_Scriptless(t *testing.T) {
 		Config: Config{
 			Cluster: ClusterKubenet,
 			VHD:     config.VHDUbuntu2404Gen2Containerd,
-			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
+			AKSNodeConfigMutator: func(config *aksnodeconfigv1.Configuration) {
 			},
 			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
 				if vmss.Tags == nil {


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -359,6 +359,10 @@ EOF
     logs_to_events "AKS.CSE.ensureContainerd.configureContainerdRegistryHost" configureContainerdRegistryHost
   fi
 
+  if [ "${TARGET_CLOUD}" = "AzureChinaCloud" ]; then
+    logs_to_events "AKS.CSE.ensureContainerd.configureContainerdLegacyMooncakeMcrHost" configureContainerdLegacyMooncakeMcrHost
+  fi
```
qweeah: It would be better to use 'else' here, even though `TARGET_CLOUD == AzureChinaCloud` cannot occur when `BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER` is set.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -384,6 +388,22 @@ configureContainerdRegistryHost() {
...
 
+configureContainerdLegacyMooncakeMcrHost() {
+    LEGACY_MCR_REPOSITORY_BASE="mcr.azk8s.cn"
+    CONTAINERD_CONFIG_REGISTRY_HOST_MCR="/etc/containerd/certs.d/${LEGACY_MCR_REPOSITORY_BASE}/hosts.toml"
+    mkdir -p "$(dirname "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}")"
+    touch "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}"
+    chmod 0644 "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}"
+
+    TARGET_MCR_REPOSITORY_BASE="mcr.azure.cn"
+    CONTAINER_REGISTRY_URL=$(sed 's@/@/v2/@1' <<< "${TARGET_MCR_REPOSITORY_BASE}/")
```
qweeah: Should we remove the /v2/ since `mcr.azure.cn` is OCI compliant?

Some context here: `/v2/` is used for NI since ACR artifact cache doesn't support [prefixed repository mapping](https://dev.azure.com/msazure/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/678808/image-pull-through-mirror-config?anchor=container-config-doesn%27t-support-prefix), `/some-prefix/...` -> `mcr.microsoft.com/some-prefix/...`.

Since such feature doesn't need to supported for MAR, I think we don't need the /v2/ here

qweeah (changes requested):

parts/linux/cloud-init/artifacts/cse_config.sh
fseldow: added

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -384,6 +386,21 @@ configureContainerdRegistryHost() {
...
+configureContainerdLegacyMooncakeMcrHost() {
+    LEGACY_MCR_REPOSITORY_BASE="mcr.azk8s.cn"
+    CONTAINERD_CONFIG_REGISTRY_HOST_MCR="/etc/containerd/certs.d/${LEGACY_MCR_REPOSITORY_BASE}/hosts.toml"
+    mkdir -p "$(dirname "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}")"
+    touch "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}"
+    chmod 0644 "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}"
+
+    TARGET_MCR_REPOSITORY_BASE="mcr.azure.cn"
+    tee "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}" > /dev/null <<EOF
+[host."https://${TARGET_MCR_REPOSITORY_BASE%/}"]
```
qweeah: From ctrd document (see https://github.com/containerd/containerd/blob/main/docs/hosts.md#setup-a-local-mirror-for-docker) , we don't need trailing slash

```suggestion
[host."https://${TARGET_MCR_REPOSITORY_BASE}"]
```

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -384,6 +386,21 @@ configureContainerdRegistryHost() {
...
+  capabilities = ["pull", "resolve"]
+  override_path = true
```
qweeah: Not needed for OCI-complaint regitries

https://github.com/containerd/containerd/blob/main/docs/hosts.md#override_path-field
```suggestion
```

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -384,6 +386,20 @@ configureContainerdRegistryHost() {
 EOF
 }
 
+configureContainerdLegacyMooncakeMcrHost() {
```
karataliu, line 395: will this conflict for NI case?

for NI, above func will add the hosts using acr.

so this one should only be called for non NI?

fseldow: the ni will be skipped see more in 
https://github.com/Azure/AgentBaker/pull/7342#discussion_r2510004108

or do you mean we should put the condition within the function?

karataliu: skipping is enough

parts/linux/cloud-init/artifacts/manifest.json
```diff
@@ -22,3 +22,4 @@
         "versions": []
     }
 }
+#EOF
```
karataliu, line 25: no need to touch this file?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -384,6 +386,20 @@ configureContainerdRegistryHost() {
...
+    LEGACY_MCR_REPOSITORY_BASE="mcr.azk8s.cn"
+    CONTAINERD_CONFIG_REGISTRY_HOST_MCR="/etc/containerd/certs.d/${LEGACY_MCR_REPOSITORY_BASE}/hosts.toml"
+    mkdir -p "$(dirname "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}")"
+    touch "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}"
+    chmod 0644 "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}"
+
+    TARGET_MCR_REPOSITORY_BASE="mcr.azure.cn"
+    tee "${CONTAINERD_CONFIG_REGISTRY_HOST_MCR}" > /dev/null <<EOF
+[host."https://${TARGET_MCR_REPOSITORY_BASE}"]
```
karataliu, line 404: make sure this file is verified working when passing correct cloud env.

parts/linux/cloud-init/artifacts/manifest.json
fseldow: wired, it is not change in my commits.

parts/linux/cloud-init/artifacts/cse_config.sh
fseldow: added e2e

e2e/kube.go
```diff
@@ -471,6 +471,9 @@ func getClusterSubnetID(ctx context.Context, mcResourceGroupName string) (string
 
 func podHTTPServerLinux(s *Scenario) *corev1.Pod {
 	image := "mcr.microsoft.com/cbl-mariner/busybox:2.0"
+	if s.Tags.MockAzureChinaCloud {
+		image = "mcr.azk8s.cn/cbl-mariner/busybox:2.0"
```
djsly, line 475: how is this supposed to work ?  e2e will never run in Mooncake, how are we expecting this image to be pulled ?

e2e/aks_model.go
```diff
@@ -219,14 +219,29 @@ func getFirewall(ctx context.Context, location, firewallSubnetID, publicIPID str
 		TargetFqdns: []*string{to.Ptr(blobStorageFqdn)},
 	}
 
+	// needed for Mock Azure China Cloud tests
```
djsly, line 222: same question, are we expecting to run e2e in Mooncake ? is the mooncake ACR available from public ?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -384,6 +388,20 @@ configureContainerdRegistryHost() {
...
```
djsly, line 404: this is confusing, why are setting `mcr.azure.cn` inside a file which as `mcr.azk8s.cn` in the file name. 

This function needs more comments to explain why we are doing something like this.  Sounds like this is temporary and we wouldnt need this for ever.

e2e/aks_model.go
fseldow: mooncake MAR `mcr.azure.cn` is available from public, while `mcr.azk8s.cn` is not allowed.
So the related e2e runs in public cloud, but change the busybox pod from `mcr.microsoft.com` to `mcr.azk8s.cn` to verify the containerd host does pull image from `mcr.azure.cn`

e2e/kube.go
fseldow: answered in https://github.com/Azure/AgentBaker/pull/7342#discussion_r2560274187

e2e/scenario_test.go
```diff
@@ -1846,6 +1846,35 @@ func Test_Ubuntu2404Gen2(t *testing.T) {
 	})
 }
 
+func Test_Ubuntu2404Gen2_McrChinaCloud(t *testing.T) {
```
Devinwong, line 1874: Could you also add 1 test for scriptless? Please search for any test with `_scriptless` in this .go file as example.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -687,6 +687,17 @@ should_enforce_kube_pmc_install() {
...
 }
 
+e2e_mock_azure_china_cloud() {
+    set -x
+    body=$(curl -fsSL -H "Metadata: true" --noproxy "*" "http://169.254.169.254/metadata/instance?api-version=2021-02-01")
+    ret=$?
+    if [ "$ret" -ne 0 ]; then
+      return $ret
+    fi
+    should_enforce=$(echo "$body" | jq -r '.compute.tagsList[] | select(.name == "E2EMockAzureChinaCloud") | .value')
```
awesomenix, line 697: I hope we can save these instance response in a json file on disk at startup somewhere.

Anybody who wants to refer this wont keep querying instance metadata server or copy pasting this function everywhere.

fseldow: since this change will change functions of other features. I may change this requirement in another pr 
https://github.com/Azure/AgentBaker/pull/7473

PR #7349, opened 2025-11-11
test: Add forward and backward compatibility tests to aks-node-controller config file

r2k1:
**What type of PR is this?**

Maintaing configuration file compatible is tricky and can be broken in unexpected ways. Some people may not be aware of such contract. 

The code is mostly AI generated, but it seems to catch issues.

Add tests preventing such issues in the future.


aks-node-controller/pkg/nodeconfigutils/compatibility_test.go
```diff
@@ -0,0 +1,148 @@
...
+	require.NoError(t, err, "Old code should unmarshal current JSON")
+	require.NotNil(t, cfg)
+	t.Logf("✓ Successfully unmarshaled %d bytes", len(data))
+}
+`
+	testFile := filepath.Join(pkgPath, "compat_test.go")
+	require.NoError(t, os.WriteFile(testFile, []byte(testCode), 0644))
+
+	// Run test with old code
+	cmd := exec.Command("go", "test", "-v", "-run", "TestUnmarshalCompat", "./pkg/nodeconfigutils")
```
Devinwong, line 116: This is quite creative! Codes in codes!

report.xml
Devinwong, line 1: unintentionally uploaded?

tmp.json.log
Devinwong, line 1: unintentionally uploaded?

PR #7350, opened 2025-11-11
chore(e2e): add firewall to e2es

lilypan26:
**What type of PR is this?**

Adds firewall to AB e2es to block all non-essential endpoints. This allows us to catch cases where provisioning is making necessary outbound calls which could break customers with strict egress.
 Example failure when script calls apt-get update:
 
image

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -57,7 +57,7 @@ func (c *Cluster) MaxPodsPerNode() (int, error) {
 	return 0, fmt.Errorf("cluster agentpool profiles were nil or empty: %+v", c.Model)
 }
 
-func prepareCluster(ctx context.Context, cluster *armcontainerservice.ManagedCluster, isAirgap, isNonAnonymousPull, addFirewall bool) (*Cluster, error) {
+func prepareCluster(ctx context.Context, cluster *armcontainerservice.ManagedCluster, isAirgap, isNonAnonymousPull bool) (*Cluster, error) {
 	ctx, cancel := context.WithTimeout(ctx, config.Config.TestTimeoutCluster)
 	defer cancel()
 	cluster.Name = to.Ptr(fmt.Sprintf("%s-%s", *cluster.Name, hash(cluster)))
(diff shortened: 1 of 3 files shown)


e2e/scenario_test.go
```diff
@@ -21,7 +21,7 @@ func Test_AzureLinux3OSGuard(t *testing.T) {
 	RunScenario(t, &Scenario{
 		Description: "Tests that a node using an Azure Linux V3 OS Guard VHD can be properly bootstrapped",
 		Config: Config{
-			Cluster: ClusterKubenet,
```
cameronmeissner, line 24: are we ever still using ClusterKubenet?

e2e/aks_model.go
```diff
@@ -182,6 +182,258 @@ func getBaseClusterModel(clusterName, location, k8sSystemPoolSKU string) *armcon
 	}
 }
 
+func getFirewallRules(ctx context.Context, location, firewallSubnetID, publicIPID string) *armnetwork.AzureFirewall {
```
cameronmeissner: this creates the firewall, I'd probably rename to `createFirewall`/`getFirewall`

e2e/cluster.go
```diff
@@ -57,7 +57,7 @@ func (c *Cluster) MaxPodsPerNode() (int, error) {
 	return 0, fmt.Errorf("cluster agentpool profiles were nil or empty: %+v", c.Model)
 }
 
-func prepareCluster(ctx context.Context, cluster *armcontainerservice.ManagedCluster, isAirgap, isNonAnonymousPull bool) (*Cluster, error) {
+func prepareCluster(ctx context.Context, cluster *armcontainerservice.ManagedCluster, isAirgap, isNonAnonymousPull, addFirewall bool) (*Cluster, error) {
```
cameronmeissner: nit: the parameter count here is getting a little unwieldy, it's also difficult to tell what's being configured at the call-sites, might be worth creating a new `ClusterPreperation` struct or something which has fields for airgap nonAnonymous (I think it should be "authenticated" instead...), and addFIrewall

non-blocking, can do this later

lilypan26: oh i was toggling based on this variable for now, but was planning enable it by default if we think its ready to merge otherwise

e2e/scenario_test.go
lilypan26: like below, was planning add the firewall to `ClusterKubenet` and get rid of the new one if no outstanding comments/changes to make

cameronmeissner: I think that's probably fine, since in theory this should give us hardened coverage

PR #7357, opened 2025-11-12
ci: Temporarily disable CIS scanning

jepio:
**What type of PR is this?**

/kind failing-test

There has been a delay in getting a new artifact published with updated license file, and as a result the scanning may need to be disabled for a few days. Revert this afterwards.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/packer/vhd-scanning.sh
+++ b/vhdbuilder/packer/vhd-scanning.sh
@@ -274,6 +274,18 @@ if ! requiresCISScan "${OS_SKU}" "${OS_VERSION}"; then
     exit 0
 fi
 
+SKIP_CIS=${SKIP_CIS:-true}
+if [ "${SKIP_CIS,,}" = "true" ]; then
+    # For artifacts
+    touch cis-report.txt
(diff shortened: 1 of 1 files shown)


vhdbuilder/packer/vhd-scanning.sh
```diff
@@ -274,6 +274,18 @@ if ! requiresCISScan "${OS_SKU}" "${OS_VERSION}"; then
     exit 0
 fi
 
+SKIP_CIS=true
```
djsly: Could we expose this as a pipeline var ? Or set to true if not defined ?

djsly (approved):

vhdbuilder/packer/vhd-scanning.sh
jepio: Yes good idea - I think this works:
```
SKIP_CIS=${SKIP_CIS:-true}
```
take a look

PR #7360, opened 2025-11-12
test: add AzureLinuxOSGuardV3gen2fipsTL to Linux VHD check-in tests

cameronmeissner:
**What type of PR is this?**

test: add AzureLinuxOSGuardV3gen2fipsTL to Linux VHD check-in tests so we don't accidentally break OSGuard builds

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/.vsts-vhd-builder.yaml
+++ b/.pipelines/.vsts-vhd-builder.yaml
@@ -31,7 +31,6 @@ pr:
     - vhdbuilder/packer/windows
     - vhdbuilder/packer/*windows*
     - vhdbuilder/packer/**/*windows*
-    - /**/*.md
     - .github/**
     - e2e/scenario_win_test.go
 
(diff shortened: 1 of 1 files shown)


djsly (approved):

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -31,7 +31,6 @@ pr:
     - vhdbuilder/packer/windows
     - vhdbuilder/packer/*windows*
     - vhdbuilder/packer/**/*windows*
-    - /**/*.md
```
djsly, line 34: why ?

djsly (approved):

.pipelines/.vsts-vhd-builder.yaml
cameronmeissner: unnecessary imo, and the regex wasn't correct to begin with

PR #7362, opened 2025-11-12
test(e2e): add validation for secure + server TLS bootstrapping

cameronmeissner:
**What type of PR is this?**

this PR adds validtion for secure + server TLS bootstrappoing in e2e tests - this is done through:
1. setting the default kubelet client bootstrapping method to secure TLS bootstrapping
2. always enabling server TLS bootstrapping (a.k.a kubelet serving certificate rotation)

we still maintain coverage on all code paths, as we have scenarios which disable both secure TLS bootstrapping and kubelet serving certificate rotation

this PR also include a bit of general code cleanup and refactoring

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -172,7 +172,7 @@ func Test_Flatcar_SecureTLSBootstrapping_BootstrapToken_Fallback(t *testing.T) {
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
 				nbc.SecureTLSBootstrappingConfig = &datamodel.SecureTLSBootstrappingConfig{
 					Enabled:     true,
-					Deadline:    (30 * time.Second).String(),
+					Deadline:    (10 * time.Second).String(),
 					AADResource: "https://management.azure.com/", // use an unexpected AAD resource to force a secure TLS bootstrapping failure
 				}
 			},
@@ -218,7 +218,7 @@ func Test_AzureLinuxV3_SecureTLSBootstrapping_BootstrapToken_Fallback(t *testing
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
 				nbc.SecureTLSBootstrappingConfig = &datamodel.SecureTLSBootstrappingConfig{
 					Enabled:                true,
-					Deadline:               (30 * time.Second).String(),
+					Deadline:               (10 * time.Second).String(),
 					UserAssignedIdentityID: "invalid", // use an unexpected user-assigned identity ID to force a secure TLS bootstrapping failure
 				}
 			},
@@ -1858,7 +1858,7 @@ func Test_Ubuntu2204_SecureTLSBootstrapping_BootstrapToken_Fallback(t *testing.T
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
 				nbc.SecureTLSBootstrappingConfig = &datamodel.SecureTLSBootstrappingConfig{
 					Enabled:                true,
-					Deadline:               (30 * time.Second).String(),
+					Deadline:               (10 * time.Second).String(),
 					UserAssignedIdentityID: "invalid", // use an unexpected user-assigned identity ID to force a secure TLS bootstrapping failure
 				}
 			},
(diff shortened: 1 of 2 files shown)


r2k1 (approved):
It seems like you added about 10 tests?

Every test exponentially increases chance of overall failure due to general flakiness. Try to avoid unnessary tests. Combine multiple tests into one when possible.

A general request, not a concrete suggestion. Use common sense on importance of the checks.

In the future I want us to test more aks-node-controller and rely less on e2e.

cameronmeissner:
> It seems like you added about 10 tests?
> 
> Every test exponentially increases chance of overall failure due to general flakiness. Try to avoid unnessary tests. Combine multiple tests into one when possible.
> 
> A general request, not a concrete suggestion. Use common sense on importance of the checks.
> 
> In the future I want us to test more aks-node-controller and rely less on e2e.

I agree that adding more tests incurs large cost - let me revisit whether we can enable secure TLS bootstrapping as the "default" bootstrapping method while making sure that we still have coverage on the standard bootstrap token path

e2e/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/agentbaker/e2e
 
-go 1.23.7
+go 1.24.0
```
r2k1, line 3: That's strange version upgrade. I would jump to latest

e2e/validators.go
```diff
@@ -16,19 +20,209 @@ import (
...
+	ValidateDirectoryContent(ctx, s, "/var/lib/kubelet", []string{"kubeconfig"})
+	ValidateDirectoryContent(ctx, s, "/var/lib/kubelet/pki", []string{"kubelet-client-current.pem"})
+	kubeletLogs := execScriptOnVMForScenarioValidateExitCode(ctx, s, "sudo journalctl -u kubelet", 0, "could not retrieve kubelet logs with journalctl").stdout.String()
+	switch {
+	case s.SecureTLSBootstrappingEnabled() && s.Tags.BootstrapTokenFallback:
+		s.T.Logf("will validate bootstrapping mode: secure TLS bootstrapping failure with bootstrap token fallback")
+		ValidateSystemdUnitIsNotRunning(ctx, s, "secure-tls-bootstrap")
+		require.True(
+			s.T,
+			!strings.Contains(kubeletLogs, "unable to validate bootstrap credentials") && strings.Contains(kubeletLogs, "kubelet bootstrap token credential is valid"),
```
r2k1, line 52: I don't like validating log messages.
Feels like a wrong level of testing.

Though I can live with it.

e2e/validators.go
```diff
@@ -16,19 +20,209 @@ import (
...
+		validateTLSBootstrappingLinux(ctx, s)
+	}
+}
+
+func validateTLSBootstrappingLinux(ctx context.Context, s *Scenario) {
+	ValidateDirectoryContent(ctx, s, "/var/lib/kubelet", []string{"kubeconfig"})
+	ValidateDirectoryContent(ctx, s, "/var/lib/kubelet/pki", []string{"kubelet-client-current.pem"})
+	kubeletLogs := execScriptOnVMForScenarioValidateExitCode(ctx, s, "sudo journalctl -u kubelet", 0, "could not retrieve kubelet logs with journalctl").stdout.String()
+	switch {
+	case s.SecureTLSBootstrappingEnabled() && s.Tags.BootstrapTokenFallback:
```
r2k1, line 47: Sometimes I wonder whether we should perform validation based on node config. 
Or is it better to perform it explictly in the test scenario. (Who is going to test our test?)

Additionally, I initially planned to use Tags only for running subset of tests and not for branching. (Though it's probably been misused like this for some time)

r2k1 (approved):

e2e/go.mod
r2k1: Does it come from a dependency upgrade?

I'm actually not sure if it's used. We probably use different go binary to compile the code.

e2e/validators.go
cameronmeissner: I agree it's brittle, though eventually we'll be able to remove this

e2e/validators.go
cameronmeissner: agreed that we should only be using NBC/nodeConfig, but doing this for now since it's simple and as you said tags are already being used in this way - I think we can come back in a separate PR to clean this all up at once

e2e/go.mod
cameronmeissner: it came from bumping client-go to 0.34.2, I think we can revisit this in a separate PR

PR #7366, opened 2025-11-13
feat: use the official Pause image TAG instead of specifying the LTCS Windows version tag

djsly:
**What type of PR is this?**
/kind cleanup

Use the official Pause image TAG instead of specifying the LTCS Windows version tag

### DIFF
--- a/vhdbuilder/packer/windows/list-images.ps1
+++ b/vhdbuilder/packer/windows/list-images.ps1
@@ -15,6 +15,7 @@ $bomList = @()
 
 # starting containerd for printing containerD info, the same way as we pre-pull containerD images in configure-windows-vhd.ps1
 Start-Job -Name containerd -ScriptBlock { containerd.exe }
+ctr.exe -n k8s.io image ls
 $imageList=$(ctr.exe -n k8s.io image ls | select -Skip 1)
 foreach($image in $imageList) {
     $splitResult=($image -split '\s+')


parts/windows/windowscsehelper.ps1
```diff
@@ -426,19 +426,6 @@ function Get-WindowsVersion {
     }
 }
 
-function Get-WindowsPauseVersion {
```
timmy-wright, line 429: So:
* This file is included in the script produced by AgentBaker
* The functions in it are used by scripts included in the VHD
* So removing this function means that AgentBaker will not be able to provision older windows VHDs. 
* It also won't be able to provision the current windows VHD until we rollout a new VHD

So let's leave this function here please.

timmy-wright (changes requested):
As per comment :)

parts/linux/cloud-init/artifacts/manifest.json
```diff
@@ -22,3 +22,4 @@
         "versions": []
     }
 }
+#EOF
```
cameronmeissner, line 25: was this by mistake? also, are we even using this file anymore?

staging/cse/windows/containerdfunc.ps1
```diff
@@ -171,18 +171,16 @@ function ProcessAndWriteContainerdConfig {
...
   $podAnnotations = 'pod_annotations = ["io.microsoft.container.processdumplocation","io.microsoft.wcow.processdumptype", "io.microsoft.wcow.processdumpcount"]'
-  $pauseWindowsVersion = Get-WindowsPauseVersion
 
   # Replace all placeholders
   $processedTemplate = $template.Replace('{{sandboxIsolation}}', $sandboxIsolation).
     Replace('{{pauseImage}}', $pauseImage).
     Replace('{{hypervisors}}', $hypervRuntimes).
     Replace('{{cnibin}}', $formatedBin).
     Replace('{{cniconf}}', $formatedConf).
-    Replace('{{currentversion}}', $pauseWindowsVersion).
```
djsly, line 182: @timmy-wright  should I keep this to now break older VHD ?

timmy-wright: This file is cached on the VHD, so you shouldn't break older VHDs.

timmy-wright: Actually, I'm wrong. It's included in the generated agentbaker script.

PR #7368, opened 2025-11-13
feat: incorporate linux prefetch optimization into VHD builds

cameronmeissner:
**What type of PR is this?**

perform linux prefetch optimization during VHD builds - this will support our effort to have a release-candidate built in our testing tenant which can then be promoted directly to our production tenant upon successful E2E validation. previously, this optimization step was performed directly as a part of the official release process.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/vhdbuilder/prefetch/scripts/optimize.sh
+++ b/vhdbuilder/prefetch/scripts/optimize.sh
@@ -160,7 +160,7 @@ prepare_source() {
 # To a managed image which can be used as a source image for the image builder template.
 # This is needed since image builder templates do not support SIG image version sources that
 # have a "Specialized" OS state. As of writing, this only applies to TrustedLaunch and CVM SKUs,
-# sonce those SKUs must be built on special hardware, and thus must be captured within a Specialized
+# since those SKUs must be built on special hardware, and thus must be captured within a Specialized
 # SIG image definition after being built with Packer.
 # This function performs the following steps to create a suitable source image based on an image version
 # coming from a Specialized image definition:


vhdbuilder/prefetch/scripts/optimize.sh
```diff
@@ -0,0 +1,284 @@
...
+    echo "temp VHD URI is ${temp_vhd_uri}"
+    TEMP_VHD_URI="${temp_vhd_uri}"
+}
+
+wait_for_vhd_copy() {
+    while [ "$(az storage blob show --blob-url "${VHD_URI}" --auth-mode login | jq -r '.properties.copy.status')" = "pending" ]; do
+        echo "VHD ${VHD_URI} is still undergoing a pending copy operation"
+        sleep 30s
+    done
+    echo "pending copy operation over ${VHD_URI} has completed"
```
awesomenix: is there a way to capture failure here? like the status could have been failed.

vhdbuilder/prefetch/scripts/optimize.sh
```diff
@@ -0,0 +1,284 @@
+#!/bin/bash
+set -uxo pipefail
```
awesomenix, line 2: overall looks good, would like to see some retry handling built-in in steps which are retriable.

vhdbuilder/prefetch/scripts/optimize.sh
zachary-bailey: yeah should we do a check to see if it has a status other than success, and if thats the case, call delete_vhd() and exit accordingly?

vhdbuilder/prefetch/scripts/optimize.sh
```diff
@@ -0,0 +1,284 @@
...
+    SOURCE_TYPE="SharedImageVersion"
+    SOURCE_ID_KEY="imageVersionId"
+    SOURCE_ID="${CAPTURED_SIG_VERSION_ID}"
+}
+
+# This function is needed to convert SIG image versions within a specialized image defintion
+# To a managed image which can be used as a source image for the image builder template.
+# This is needed since image builder templates do not support SIG image version sources that
+# have a "Specialized" OS state. As of writing, this only applies to TrustedLaunch and CVM SKUs,
+# sonce those SKUs must be built on special hardware, and thus must be captured within a Specialized
```
zachary-bailey: nit, since*

vhdbuilder/prefetch/scripts/optimize.sh
cameronmeissner: at the moment I'd like to keep this as simple as possible - we'll get added retry logic for each individual step for free once we migrate this to golang internally, though for now making the entire script idempotent and using the `retryTaskOnfailure` directive on the ADO step should be enough

vhdbuilder/prefetch/scripts/optimize.sh
cameronmeissner: good catch

cameronmeissner: fixed

awesomenix (approved):

vhdbuilder/prefetch/scripts/optimize.sh
```diff
@@ -0,0 +1,293 @@
...
+            if [ "${copy_status,,}" = "success" ]; then
+                echo "VHD ${VHD_URI} has been successfully copied from image builder storage, nothing to do"
+                exit 0
+            fi
+            if [ "${copy_status,,}" = "pending" ]; then
+                echo "echo VHD ${VHD_URI} is currently being copied from image builder storage, will wait for copy completion"
+                if wait_for_vhd_copy; then
+                    exit 0
+                fi
+                echo "pending copy operation was not successful, will delete existing blob and attempt to retry optimization and VHD creation"
```
zachary-bailey, line 60: I dont think that the `delete_vhd` function call under this comment will ever execute if `wait_for_vhd_copy` function call returns 1?

possible alternative:

`wait_for_vhd_copy && exit 0`
`delete_vhd`
`exit $?`

zachary-bailey: in this edge case `copy_status` could be pending, which would start the logic on line 56, but it later failed, since it was originally pending I dont think we would ever get to the logic in line 62 to delete the vhd?

zachary-bailey: or we could just add `delete_vhd` into the `wait_for_vhd_copy` command

cameronmeissner: I'll add the missing `delete_vhd`, I think we technically don't even need it since subsequent copy operations to the same VHD will just overwrite it... but you're right that it's missing here

cameronmeissner: fixed

vhdbuilder/prefetch/scripts/optimize.sh
```diff
@@ -0,0 +1,294 @@
...
+            -e "s#<VHD_URI>#${VHD_URI}#g" \
+            "${IMAGE_BUILDER_TEMPLATE_PATH}" > input.json || return $?
+
+        if [ ! -f "input.json" ]; then
+            echo "unable to create input image template for ${IMAGE_BUILDER_TEMPLATE_NAME}"
+            return 1
+        fi
+
+        echo "creating image builder template ${IMAGE_BUILDER_TEMPLATE_NAME} in resource group ${IMAGE_BUILDER_RG_NAME}"
+        az resource create -n "${IMAGE_BUILDER_TEMPLATE_NAME}" \
```
zachary-bailey, line 101: Just want to double check that you are overriding the default resource group created for the template here: [image builder docs](https://learn.microsoft.com/en-us/azure/virtual-machines/image-builder-overview?tabs=azure-powershell#:~:text=During%20the%20image%2Dcreation%20process%2C%20files%20are%20downloaded%20and%20stored%20in%20the%20IT_%3CDestinationResourceGroup%3E_%3CTemplateName%3E%20resource%20group%2C%20which%20incurs%20a%20small%20storage%20cost.%20If%20you%20don%27t%20want%20to%20keep%20these%20files%2C%20delete%20the%20image%20template%20after%20you%27ve%20built%20the%20image.)

so when we clean up the resource group later, it deletes any associated resources or files that were persisted with the template run

zachary-bailey: and so cleaning up the RG deletes the template

cameronmeissner: it will delete all child resources (the image builder template + generated staging resource group + ephemeral storage account/disk/managed image in CVM/TL case)

PR #7371, opened 2025-11-13
feat: inspektor gadget into vhd build

chmill-zz:
**What type of PR is this?**

/kind feature

this is getting inspektor gadget into agentbaker by default.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -414,7 +414,7 @@ func ValidateInspektorGadget(ctx context.Context, s *Scenario) {
 	servicePath := "/usr/lib/systemd/system/" + serviceName
 
 	// Check if IG is installed on this VHD by looking for the skip sentinel file.
-	// The skip file is only present on VHDs that have IG installed (Ubuntu, Mariner, Azure Linux).
+	// The skip file is only present on VHDs that have IG installed (Ubuntu and Azure Linux non-OSGuard).
 	// Flatcar, OSGuard, and older VHDs do not have IG installed and will not have the skip file.
 	if !fileExist(ctx, s, skipFile) {
 		s.T.Logf("Skipping Inspektor Gadget validation: sentinel file %s not found (VHD does not have IG installed)", skipFile)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -29,6 +29,7 @@ source /home/packer/provision_source_benchmarks.sh
 source /home/packer/provision_source_distro.sh
 source /home/packer/tool_installs.sh
 source /home/packer/tool_installs_distro.sh
+source "${THIS_DIR}/install-ig.sh"
```
cameronmeissner: can just hardcode to `/home/packer/install-ig.sh`?

parts/linux/cloud-init/artifacts/inspektor-gadget/skip_vhd_ig
cameronmeissner, line 1: again, would prefer we create this file within install-deps rather than check it into version control

vhdbuilder/packer/install-ig.sh
```diff
@@ -0,0 +1,292 @@
...
+    cp -f "${skip_source}" "${IG_SKIP_FILE}"
+
+    chmod 644 "${IG_SKIP_FILE}"
+}
+
+ig_install_deb_stack() {
+    local download_dir="${IG_BUILD_ROOT}/downloads"
+    mkdir -p "${download_dir}"
+
+    local ig_tag="${IG_VERSION}-ubuntu18.04u${IG_REVISION}"
```
cameronmeissner, line 108: we don't support 18.04 anymore :D

chmill-zz: ah yeah this is fun... IG is agnostic here. So they've got their generic in 18.04 for ig then ig-gadgets live in 20.04, but they're maintained. This was exciting to learn.

chmill-zz: this is how extension is currently pulling in ig components.

cameronmeissner: hrm... so we're forced to take from 18.04 repo... wondering if we could create a dalec spec for each OS version so we're not pinned

in the meantime if this works on all OS versions then I guess it's not blocking

PR #7382, opened 2025-11-14
surfacing the error from provision-wait mode

Devinwong:
**What type of PR is this?**
/kind bug

Surfacing the error from provision-wait mode.
The original issue was that when there is non-zero exit code in the provisioning scripts, for non-scriptless, CRP operation can surface it but scriptless couldn't. 
Validated in local AB e2e. Would double check with the issue reporter once this is released next week.

Fixes #

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


aks-node-controller/app.go
```diff
@@ -65,6 +67,9 @@ func (a *App) run(ctx context.Context, args []string) error {
 	case "provision-wait":
 		provisionStatusFiles := ProvisionStatusFiles{ProvisionJSONFile: provisionJSONFilePath, ProvisionCompleteFile: provisionCompleteFilePath}
 		provisionOutput, err := a.ProvisionWait(ctx, provisionStatusFiles)
+		if err != nil {
+			slog.Error("provision-wait failed", "error", err)
+		}
```
r2k1: Usually I avoid logging error and returning it at the same time. Typically it just lead to logging same error twice.

aks-node-controller/app.go
```diff
@@ -158,11 +163,9 @@ func (a *App) writeCompleteFileOnError(err error) {
 
 func (a *App) ProvisionWait(ctx context.Context, filepaths ProvisionStatusFiles) (string, error) {
 	if _, err := os.Stat(filepaths.ProvisionCompleteFile); err == nil {
-		data, err := os.ReadFile(filepaths.ProvisionJSONFile)
-		if err != nil {
-			return "", fmt.Errorf("failed to read provision.json: %w. One reason could be that AKSNodeConfig is not properly set", err)
-		}
-		return string(data), nil
+		// Fast path: provision.complete already exists when we enter. Avoid watcher overhead.
```
r2k1: A side note. It seems like there is a potential for a race.

What if the file created after we run `os.Stat` but before `NewWatcher`?

I think we should check it after watcher is started.

I would probably try to add a new channel (capcity one? not sure). Send to it a dummy event and add a new case statement.

r2k1: Actually, solution is simpler:

Just move `os.Stat(filepaths.ProvisionCompleteFile)` after `watcher.Add`

r2k1: It's not to avoid watcher overhead.

I think the watcher isn't going to produce an event if the file already exists.

Devinwong: That's a good point. Updated and added unit tests.

aks-node-controller/app.go
Devinwong: makes sense. removed.

aks-node-controller/app.go
```diff
@@ -157,19 +159,17 @@ func (a *App) writeCompleteFileOnError(err error) {
...
 	if err != nil {
 		return "", fmt.Errorf("failed to create watcher: %w", err)
 	}
 	defer watcher.Close()
+
+	if _, statErr := os.Stat(filepaths.ProvisionCompleteFile); statErr == nil {
+		// Fast path: provision.complete already exists when we enter. Avoid watcher overhead.
+		// We read and evaluate once and return immediately. Only this branch executes in this scenario.
+		return readAndEvaluateProvision(filepaths.ProvisionJSONFile)
+	}
```
r2k1: I think it should be moved after `watcher.Add`

PR #7387, opened 2025-11-15
Extract latest coredns binary for localdns

SriHarsha001:
**What type of PR is this?**
design

In this PR, updating the code logic to extract latest coredns binary that will be used for localdns. Currently, we were using n-1 coredns version for localdns and this one has AWS CVE. 

Fixes # CVE

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Release note**:

```
none
```


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1514,29 +1514,10 @@ testCorednsBinaryExtractedAndCached() {
 
   # Determine latest version (eg. v1.12.0-1).
   local latest_coredns_tag="${sorted_coredns_tags[0]}"
```
yewmsft, line 1621: so this is the list of all coredns images we cache on the vhd, right? So will this latest_coredns_tag always be 1.13.*?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -972,10 +935,10 @@ extractAndCacheCoreDnsBinary() {
     local coredns_binary="${ctr_temp}/usr/bin/coredns"
     if [ -f "${coredns_binary}" ]; then
       cp "${coredns_binary}" "${LOCALDNS_BINARY_PATH}/coredns" || {
-        echo "Error: Failed to copy coredns binary of ${previous_coredns_tag}" >> "${VHD_LOGS_FILEPATH}"
+        echo "Error: Failed to copy coredns binary of ${latest_coredns_tag}" >> "${VHD_LOGS_FILEPATH}"
```
yewmsft, line 938: which binary version do we choose when we are enabling localdns on 1.31 clusters?

vhdbuilder/packer/test/linux-vhd-content-test.sh
SriHarsha001: yes, this will be the list of all coredns images cached in the VHD. Components.json file will be updated periodically when we bump coredns images in AgentBaker. Yes, latest_coredns_tag will be 1.13.* for now as this is a the latest coredns image version we have updated in components.json file. But in the future, whatever is the latest version updated in components.json file, that versions binary will be extracted.

vhdbuilder/packer/install-dependencies.sh
SriHarsha001: Currently, binary version of coredns used in localdns is not tied to K8s version. We need to do a design for deciding on which version will be used on which nodes and how to patch the CVEs and how to fallback. Pushing the binaries from DALEC to ACR and then pulling the binary in AgentBaker seem to be best option - but need to do POC and design review for this. For now, idea is to use the latest coredns binary for all localdns enabled nodes.

vhdbuilder/packer/test/linux-vhd-content-test.sh
SriHarsha001: image

PR #7397, opened 2025-11-18
fix: pin publisher base images

zachary-bailey:
**What type of PR is this?**

/kind bug

This PR pins the base image versions used for Ubuntu Packer builds in order to prevent boot issues.

Boot errors.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


vhdbuilder/publisher_base_image_version.json
```diff
@@ -0,0 +1,12 @@
+{
+  "20_04-lts": "20.04.202505200",
```
cameronmeissner, line 2: this won't be picked up on builds off master

cameronmeissner: nvm, logic doesn't care about the branch thankfully

vhdbuilder/publisher_base_image_version.json
```diff
@@ -0,0 +1,12 @@
+{
```
djsly, line 1: where can we add note, / reminder to clean this up ? outside of this JSON file ?

djsly (approved):

PR #7401, opened 2025-11-19
feat: add PR lint workflow

cameronmeissner:
**What type of PR is this?**

feat: add PR lint workflow to ensure that PR titles follow conventional commits: https://commitlint.js.org/#/concepts-commit-conventions, which allows us to automatically label each PR so it can correctly be categorized within AgentBaker release notes

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.github/workflows/pr-lint.yaml
+++ b/.github/workflows/pr-lint.yaml
@@ -42,7 +42,7 @@ jobs:
           # Configure which scopes are allowed (newline-delimited).
           # These are regex patterns auto-wrapped in `^ $`.
           scopes: |
-            *
+            .*
 
           # Configure scope as optional.
           requireScope: false


.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+          #  |____________ Type
+
+          # Configure which types are allowed (newline-delimited).
+          # These are regex patterns auto-wrapped in `^ $`.
+          # Default: https://github.com/commitizen/conventional-commit-types
+          types: |
+            feat
+            fix
+            refactor
+            chore
```
djsly: I never understood when `chore` should be used

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+            test
```
djsly, line 40: test/perf are the same ?   when should `perf` be used ?

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+            perf
+            ci
+            docs
+
+          # Configure which scopes are allowed (newline-delimited).
+          # These are regex patterns auto-wrapped in `^ $`.
+          scopes: |
```
djsly, line 44: if we are missing a scope, will it be allowed ?

djsly: saw the next line!

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+          # the default error message that is shown when the pattern doesn't match.
+          # The variables `subject` and `title` can be used within the message.
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            didn't match the configured pattern. Please ensure that the subject
+            doesn't start with an uppercase character.
+
+          # If the PR only contains a single commit, the action will validate that
+          # it matches the configured pattern.
+          validateSingleCommit: true
```
djsly: I think they recommand to not use this, and instead use the official GitHub solution 

https://github.com/marketplace/actions/semantic-pull-request#legacy-configuration-for-validating-single-commits

However, [GitHub has introduced an option to streamline this behaviour](https://github.blog/changelog/2022-05-11-default-to-pr-titles-for-squash-merge-commit-messages/), so using that instead should be preferred.

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+          # Related to `validateSingleCommit` you can opt-in to validate that the PR
+          # title matches a single commit to avoid confusion.
+          validateSingleCommitMatchesPrTitle: true
+
+        continue-on-error: true
```
djsly, line 63: this wou;dnt enforce ?

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+            - \`ci: update build pipeline\` - for CI/CD changes
+            - \`perf: improve performance\` - for performance improvements
+
+            ### Guidelines:
+            - Use lowercase for the type and description
+            - Keep the description concise but descriptive
+            - Use imperative mood (e.g., "add" not "adds" or "added")
+            - Don't end with a period
+
+            ### Examples:
```
djsly, line 127: what about example with scopes and listing the allowed scopes ?

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+on:
+  pull_request:
+    branches:
+      - master
+      - dev
+    types:
+      - opened
+      - edited
+      - synchronize
```
r2k1: I don't think you need `synchronize` if you only care about title?

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+jobs:
+  lint-pr-title:
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Lint PR Title
+        id: lint-pr-title
+        uses: amannn/action-semantic-pull-request@v6
```
r2k1, line 20: It's a bit concerning to give write access to external action.

Is it limited only writing to PR? A human need to approve it, right?

.github/workflows/pr-lint.yaml
r2k1: I would say when you try to improve VM provision time rather then E2E/Build time.

r2k1: Though I agree in general we should use minimal possible list of options.

Are we going to behave differently for different scenarios?

.github/workflows/pr-lint.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+            Your PR title doesn't follow the expected format. Please update your PR title to follow one of these patterns:
+
+            ### Conventional Commits Format:
+            - \`feat: add new feature\` - for new features
+            - \`fix: resolve bug in component\` - for bug fixes
+            - \`docs: update README\` - for documentation changes
+            - \`refactor: improve code structure\` - for refactoring
+            - \`test: add unit tests\` - for test additions
+            - \`chore(deps): update dependencies\` - for maintenance tasks
```
r2k1: What's the differemce between `chore` vs `chore(deps)`?

.github/workflows/pr-lint.yaml
cameronmeissner: yeah I saw the same, just figured it would be good to cover our bases

.github/workflows/pr-lint.yaml
cameronmeissner: there's an `exit 1` as the last workflow step, after updating the PR with a comment for guidance, check below

.github/workflows/pr-lint.yaml
cameronmeissner: will update the exmaple

.github/workflows/pr-lint.yaml
cameronmeissner: `chore` could be removing unused code, for example, `chore(deps)` would specifically be for updating a dependency in components.json

.github/workflows/pr-lint.yaml
cameronmeissner: what exactly is considered a synchronization? it's one of the recommended triggers according to the action docs

.github/workflows/pr-lint.yaml
cameronmeissner: it should be fine, the action won't actually be doing anything, it's the github script below which adds the comment assuming the linter fails which needs the write action

since we're using `pull_request` instead of `pull_request_target`, we should also be protected from fork attacks

.github/workflows/pr-lint.yaml
cameronmeissner: we use it for our automated PRs and it's used by renovate

.github/workflows/pr-lint.yaml
cameronmeissner: it will dictate how the PR is auto-labeled, which will subsequently dictate how it's categorized within AgentBaker release notes

.github/workflows/pr-lint.yaml
cameronmeissner: removed, let's make sure to update that repo setting

.github/workflows/pr-lint.yaml
cameronmeissner: ah,  triggered by new commits - we can remove that

PR #7403, opened 2025-11-19
switch blobfuse2 version to 2.5.0

ganeshkumarashok:
**What type of PR is this?**
/kind bug

Reverts the blobfuse version to 2.5.0 from 2.5.1. This is to fix an issue that will be hit only in sovereign clouds where AAD endpoint is different than standard public aad endpoint for Azure and customer is using SPN or MSI based authentication.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/manifest.json
```diff
@@ -22,3 +22,4 @@
         "versions": []
     }
 }
+#EOF
```
djsly, line 25: why ?

ganeshkumarashok: fixed

PR #7437, opened 2025-11-21
fix: managed gpu experience tests

jepio:
**What type of PR is this?**
/kind failing-test

Managed GPU experience tests were failing, the reasons:
* GRID checks on non-GRID VM sizes
* GRID instance size with AL3 (unsupported)
* NPD metrics not showing up because NPD init races with GPU driver installation -> restart NPD after GPU driver is installed.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -910,6 +910,9 @@ configGPUDrivers() {
     # This prevents automatic updates from upgrading Nvidia repo dependencies and eliminates
     # the need for users to allowlist NVIDIA repository URLs in their firewalls
     removeNvidiaRepos
+
+    # Restart node-problem-detector to pick up GPU related metrics
+    systemctl restart node-problem-detector || true
```
djsly, line 917: since NPD is coming from an extension, this is expected to fail normally correct ? this will just cover the case when NPD is installed before CSEVMSS runs ? could we augment the comment to explicitely call this out ? We willsoon be bringing NPD to be packaged as part of the VHD, so we should expect this logic to be udpated.

e2e/scenario_gpu_managed_experience_test.go
```diff
@@ -262,14 +262,14 @@ func Test_AzureLinux3_NvidiaDevicePluginRunning(t *testing.T) {
 			Cluster: ClusterKubenet,
 			VHD:     config.VHDAzureLinuxV3Gen2,
 			BootstrapConfigMutator: func(nbc *datamodel.NodeBootstrappingConfiguration) {
-				nbc.AgentPoolProfile.VMSize = "Standard_NV6ads_A10_v5"
+				nbc.AgentPoolProfile.VMSize = "Standard_NC6s_v3"
```
djsly, line 265: The NCv3-series will be retired on September 30, 2025. 

any reason we are moving to this SKU, is it because we are running out of A10 ?

Or maybe cause A10 (Grid drvier) don't support the Managed GPU experience ?

jepio: The Azure Linux driver reports it doesn't support A10, I'm not sure if we have a GRID driver for Azure Linux.

I reverted to NC6s_v3 because that's what was used last week, we can switch to anything that is supported. What do we have quota for: A100? T4?

parts/linux/cloud-init/artifacts/cse_config.sh
jepio: Let me update the comment. Yes, this might fail depending on timing: the driver installation might take a while and NPD installation will be happen either before, during, or after.
Once NPD is in the VHD we'll need to decide the order and make this line required or not.

e2e/scenario_gpu_managed_experience_test.go
djsly: it this was a miss, good to revert for now. we should indeed revisit our GPU quotas for e2e and our coverage 

@ganeshkumarashok  / @awesomenix

awesomenix: We have lot of quota for A10 (about 500 cores). A miss on my part of this, sorry about that.

>The Azure Linux driver reports it doesn't support A10

Its ok to keep it at A10, but for this PR I am ok to revert to get this in good state.

parts/linux/cloud-init/artifacts/cse_config.sh
jepio: updated the comment

jepio:
@djsly: might want to merge this with override after the CI runs because there are still some non-gpu e2e's failing

PR #7446, opened 2025-11-25
fix: kubelet requires containerd

Devinwong:
**What type of PR is this?**
/kind fix

kubelet now follows containerd’s lifecycle more strictly:

- If containerd crashes and restarts, kubelet will also restart.

This guards against an issue where containerd crashes while kubelet remains running, which previously prevented the node remediator from noticing the failure and taking action (for example, rebooting the node).

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -1211,13 +1211,13 @@ func Test_Ubuntu2204_CustomSysctls_Kubelet_Lifecycle(t *testing.T) {
 			Validator: func(ctx context.Context, s *Scenario) {
 				ValidateUlimitSettings(ctx, s, customContainerdUlimits)
 				ValidateSysctlConfig(ctx, s, customSysctls)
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/kubelet.service
```diff
@@ -1,8 +1,9 @@
 [Unit]
 Description=Kubelet
 ConditionPathExists=/usr/local/bin/kubelet
-Wants=network-online.target containerd.service
+Wants=network-online.target
 After=network-online.target containerd.service
+Requires=containerd.service
```
timmy-wright: It'd be good to see an e2e test that restarts containerd and checks kubelet also restarts. We have a similar test for chrony - where we kill the process and check it restarts.

PR #7447, opened 2025-11-25
feat: add node label for dcgm exporter

lilypan26:
**What type of PR is this?**

Add node label to indicate when dcgm exporter is enabled on GPU nodes so Prometheus can discover nodes for scraping.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -403,10 +403,20 @@ function nodePrep {
     elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "true" ]; then
         logs_to_events "AKS.CSE.installNvidiaManagedExpPkgFromCache" "installNvidiaManagedExpPkgFromCache" || exit $ERR_NVIDIA_DCGM_INSTALL
         logs_to_events "AKS.CSE.startNvidiaManagedExpServices" "startNvidiaManagedExpServices" || exit $ERR_NVIDIA_DCGM_EXPORTER_FAIL
-    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "false" ]; then
+
```
cameronmeissner: I feel like L398-420 should be in a helper (maybe in cse_install.sh) that we can unit test with shellspec, though I see you have the E2E coverage which is fine for now

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -403,10 +403,20 @@ function nodePrep {
...
+        # Add node label to indicate DCGM exporter is enabled
+        DCGM_EXPORTER_LABEL="kubernetes.azure.com/dcgm-exporter=enabled"
+        echo "adding node label $DCGM_EXPORTER_LABEL"
+        addKubeletNodeLabel $DCGM_EXPORTER_LABEL
```
djsly: would we need to ensure that ensure Kubelet is call after ?  since this is part of `nodePrep` If the start of Kubelet occurs after we would not have those labels correct ?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -403,10 +403,20 @@ function nodePrep {
...
+    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" != "true" ]; then
         logs_to_events "AKS.CSE.stop.nvidia-device-plugin" "systemctlDisableAndStop nvidia-device-plugin"
         logs_to_events "AKS.CSE.stop.nvidia-dcgm" "systemctlDisableAndStop nvidia-dcgm"
         logs_to_events "AKS.CSE.stop.nvidia-dcgm-exporter" "systemctlDisableAndStop nvidia-dcgm-exporter"
+
+        # Add node label to indicate DCGM exporter is disabled
+        DCGM_EXPORTER_LABEL="kubernetes.azure.com/dcgm-exporter=disabled"
+        echo "adding node label $DCGM_EXPORTER_LABEL"
+        addKubeletNodeLabel $DCGM_EXPORTER_LABEL
```
djsly: could we move the echo inside `addKubeletNodeLabel` and simply pass the string as an argument to simplify the number of added lines between the two conditiions?

lilypan26: oops there's already a logging statement in the function, i removed the excess ones

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -403,10 +403,20 @@ function nodePrep {
     elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "true" ]; then
         logs_to_events "AKS.CSE.installNvidiaManagedExpPkgFromCache" "installNvidiaManagedExpPkgFromCache" || exit $ERR_NVIDIA_DCGM_INSTALL
         logs_to_events "AKS.CSE.startNvidiaManagedExpServices" "startNvidiaManagedExpServices" || exit $ERR_NVIDIA_DCGM_EXPORTER_FAIL
-    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "false" ]; then
+
+        # Add node label to indicate DCGM exporter is enabled
+        DCGM_EXPORTER_LABEL="kubernetes.azure.com/dcgm-exporter=enabled"
+        echo "adding node label $DCGM_EXPORTER_LABEL"
+        addKubeletNodeLabel $DCGM_EXPORTER_LABEL
+    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" != "true" ]; then
```
surajssd:
```suggestion
    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "false" ]; then
```

Shouldn't this be false, was it added by mistake?

surajssd: Oh I see it now, it is flipping to not being not true.

e2e/scenario_gpu_managed_experience_test.go
```diff
@@ -161,6 +161,7 @@ func Test_Ubuntu2404_NvidiaDevicePluginRunning(t *testing.T) {
 				ValidateNvidiaDCGMExporterSystemDServiceRunning(ctx, s)
 				ValidateNvidiaDCGMExporterIsScrapable(ctx, s)
 				ValidateNvidiaDCGMExporterScrapeCommonMetric(ctx, s, "DCGM_FI_DEV_GPU_UTIL")
+				ValidateFileHasContent(ctx, s, "/etc/default/kubelet", "kubernetes.azure.com/dcgm-exporter=enabled")
```
surajssd: Ideally the tests should verify the node labels than existence of label in the label initiation conifg. Because ultimately the thing that will get consumed is the actual node label.

To repharse, this test is testing the preconditions that will lead to the ideal outcome, instead of testing the ideal outcome. Sometimes things could be misconfigured and we may not get that ideal outcome but our 
tests will keep passing.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1148,6 +1148,28 @@ EOF
     echo "Enable localdns succeeded."
 }
 
+configureManagedGPUExperience() {
+    if [ "$?" -ne 0 ] && [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ]; then
```
surajssd: IDK what the value of `$?` would be since `$?` returns return code of the previous command that ran. Earlier we were running `ENABLE_MANAGED_GPU_EXPERIENCE=$(retrycmd_silent 10 1 10 bash -cx enableManagedGPUExperience)` before this `if` block so no error code passing was needed.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1148,6 +1148,28 @@ EOF
...
+        echo "failed to determine if managed GPU experience should be enabled by nodepool tags"
+        exit $ERR_LOOKUP_ENABLE_MANAGED_GPU_EXPERIENCE_TAG
+    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "true" ]; then
+        logs_to_events "AKS.CSE.installNvidiaManagedExpPkgFromCache" "installNvidiaManagedExpPkgFromCache" || exit $ERR_NVIDIA_DCGM_INSTALL
+        logs_to_events "AKS.CSE.startNvidiaManagedExpServices" "startNvidiaManagedExpServices" || exit $ERR_NVIDIA_DCGM_EXPORTER_FAIL
+
+        # Add node label to indicate DCGM exporter is enabled
+        DCGM_EXPORTER_LABEL="kubernetes.azure.com/dcgm-exporter=enabled"
```
djsly: do we need this variable ?

lilypan26: removed

parts/linux/cloud-init/artifacts/cse_config.sh
lilypan26: moved this check outside of the function

parts/linux/cloud-init/artifacts/cse_main.sh
lilypan26: confirmed it's called after, and have tests ensure labels are persisted

lilypan26:
> What was the conclusion on our discussion about tradeoffs for label logic in aks-rp vs Agentbaker?

I asked cameron and he mentioned a few benefits of adding it in AB as opposed to RP:
- ensures there's no modifications to labels between when AB gets the labels from RP to when kubelet is actually started
- having it in on the node image ensure we don't need to worry about various bootstrappers calling AB

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1148,6 +1148,21 @@ EOF
...
+        logs_to_events "AKS.CSE.installNvidiaManagedExpPkgFromCache" "installNvidiaManagedExpPkgFromCache" || exit $ERR_NVIDIA_DCGM_INSTALL
+        logs_to_events "AKS.CSE.startNvidiaManagedExpServices" "startNvidiaManagedExpServices" || exit $ERR_NVIDIA_DCGM_EXPORTER_FAIL
+
+        addKubeletNodeLabel "kubernetes.azure.com/dcgm-exporter=enabled"
+    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "false" ]; then
+        logs_to_events "AKS.CSE.stop.nvidia-device-plugin" "systemctlDisableAndStop nvidia-device-plugin"
+        logs_to_events "AKS.CSE.stop.nvidia-dcgm" "systemctlDisableAndStop nvidia-dcgm"
+        logs_to_events "AKS.CSE.stop.nvidia-dcgm-exporter" "systemctlDisableAndStop nvidia-dcgm-exporter"
+
+        addKubeletNodeLabel "kubernetes.azure.com/dcgm-exporter=disabled"
```
djsly: Didn’t we decide to not put a label when disabled ?

ganeshkumarashok:
> > What was the conclusion on our discussion about tradeoffs for label logic in aks-rp vs Agentbaker?
> 
> I asked cameron and he mentioned a few benefits of adding it in AB as opposed to RP:
> 
> * ensures there's no modifications to labels between when AB gets the labels from RP to when kubelet is actually started
> * having it in on the node image ensure we don't need to worry about various bootstrappers calling AB

To do for future: we'll need to add labeling logic for karpenter and NAP afterwards, separately

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -400,14 +400,8 @@ function nodePrep {
...
         exit $ERR_LOOKUP_ENABLE_MANAGED_GPU_EXPERIENCE_TAG
-    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "true" ]; then
-        logs_to_events "AKS.CSE.installNvidiaManagedExpPkgFromCache" "installNvidiaManagedExpPkgFromCache" || exit $ERR_NVIDIA_DCGM_INSTALL
-        logs_to_events "AKS.CSE.startNvidiaManagedExpServices" "startNvidiaManagedExpServices" || exit $ERR_NVIDIA_DCGM_EXPORTER_FAIL
-    elif [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ] && [ "${ENABLE_MANAGED_GPU_EXPERIENCE}" = "false" ]; then
-        logs_to_events "AKS.CSE.stop.nvidia-device-plugin" "systemctlDisableAndStop nvidia-device-plugin"
-        logs_to_events "AKS.CSE.stop.nvidia-dcgm" "systemctlDisableAndStop nvidia-dcgm"
-        logs_to_events "AKS.CSE.stop.nvidia-dcgm-exporter" "systemctlDisableAndStop nvidia-dcgm-exporter"
     fi
+    logs_to_events "AKS.CSE.configureManagedGPUExperience" configureManagedGPUExperience || exit $ERR_ENABLE_MANAGED_GPU_EXPERIENCE
```
surajssd, line 404: This error `ERR_ENABLE_MANAGED_GPU_EXPERIENCE` does not exists already, has to be defined similar to other errors, like this: https://github.com/Azure/AgentBaker/blob/135d009b7dd6ee610027d9c2839e926da738cdcb/parts/linux/cloud-init/artifacts/cse_helpers.sh#L145

lilypan26: fixed

ganeshkumarashok (approved):

PR #7449, opened 2025-11-25
chore: move e2e and gc to use service connection

awesomenix:
**What type of PR is this?**

Move E2E to use service connection, this removes the need for attaching managed identity to build agents.

/kind cleanup

Fixes #

**Requirements**:

- [X] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```

### DIFF
--- a/.pipelines/templates/e2e-template.yaml
+++ b/.pipelines/templates/e2e-template.yaml
@@ -24,10 +24,15 @@ jobs:
         fetchTags: false
         fetchDepth: 1
 
-      - bash: bash .pipelines/scripts/e2e_run.sh
+      - task: AzureCLI@2
+        inputs:
+          azureSubscription: $(E2E_ARM_SERVICE_CONNECTION)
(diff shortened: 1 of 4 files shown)


.pipelines/templates/e2e-template.yaml
```diff
@@ -24,10 +24,15 @@ jobs:
         fetchTags: false
         fetchDepth: 1
 
-      - bash: bash .pipelines/scripts/e2e_run.sh
+      - task: AzureCLI@2
```
djsly, line 27: what do we get extra with this task type ?

djsly (approved):

.pipelines/templates/e2e-template.yaml
awesomenix: i think its logs in using federated identity, i am pretty sure you can achieve the same with cli manually as well, but the automatic way of doing it makes it easy(especially when things can change in future)

PR #7463, opened 2025-11-28
fix: set network plugin to "none" in Azure CNI test cases

johananl:
**What type of PR is this?**

/kind bug

The Azure CNI plugin is installed nowadays by the `azure-cns` DaemonSet. When the user specifies `--network-plugin=azure`, RP ensures `azure-cns` installs the plugin and passes `none` to AB to avoid a duplicate installation process.

However, the e2e test cases aren't up to date and still specify `azure` as the plugin type in the VMSS config. This causes two installation processes to compete, thus leading to inconsistent test results.

Fixes #7460 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


johananl:
This is likely obsolete by now. Closing since this has no traction and I'm not actively working on AgentBaker at the moment.

PR #7504, opened 2025-12-08
fix: skip update repo if install has opt to define source

fseldow:
**What type of PR is this?**

/kind bug

we observed ni cluster with for azure linux node pool has latency issue for vmss creation. The cse will stuck at 'dnf makecache --disablerepo='*' --enablerepo="${repo_name}"' It will stuck over 10 minutes and finally caused the cse exit with timeout. However, the next instance or ssh into failed node both can work. Thus i make this pr to add timeout option for 'dnf makecache' and 'apt update'.

Fixes #

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```
none
```


qweeah (approved):
LGTM

qweeah (approved):
LGTM

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -231,15 +231,15 @@ EOF
 
     # Update DNF cache for the new repository
     echo "Updating DNF cache for local repository"
-    dnf makecache --disablerepo='*' --enablerepo="${repo_name}" || {
+    _dnf_makecache 6 10 30 --disablerepo='*' --enablerepo="${repo_name}" || {
```
lilypan26, line 235: can we add shellspec tests for this?

parts/linux/cloud-init/artifacts/mariner/cse_helpers_mariner.sh
```diff
@@ -32,8 +36,11 @@ dnf_install() {
         if [ $i -eq $retries ]; then
             return 1
         else
-            sleep $wait_sleep
-            dnf_makecache
+            sleep "$wait_sleep"
+            case " $* " in
+                *" --disablerepo "*) ;;    # skip dnf_makecache when --disablerepo is present
```
lilypan26: doesn't `--disablerepo` have a parameter? so matching with the trailing space would not match? e.g. with `--disablerepo='*'`

fseldow: remove empty char thx for catch

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
fseldow: added in spec/parts/linux/cloud-init/artifacts/cse_install_mariner_spec.sh

spec/parts/linux/cloud-init/artifacts/cse_install_mariner_spec.sh
```diff
@@ -31,4 +31,41 @@ Describe 'cse_install_mariner.sh'
             The output line 1 should include "Installing azurelinux-repos-cloud-native"
         End
     End
-End
\ No newline at end of file
+
+    Describe 'installToolFromLocalRepo'
```
lilypan26, line 35: can you mock `dnf_makecache` and have it `echo "dnf_makecache"`? That way we can confirm it is not called when --disablerepo is specified

fseldow: add ut 'cse_helpers_ubuntu.sh-_apt_get_install' and `cse_helpers_mariner.sh - dnf_install` to cover the scenario

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -232,15 +232,15 @@ EOF
...
         echo "Failed to update DNF cache for local repository"
         rm -f "${repo_file}"
         return 1
     }
 
     # Install the package from the local repository
     echo "Installing ${tool_name} from local repository"
-    if ! dnf_install 30 1 600 ${tool_name} --disablerepo='*' --enablerepo="${repo_name}"; then
+    if ! dnf_install 4 1 120 ${tool_name} --disablerepo='*' --enablerepo="${repo_name}"; then
```
lilypan26: why are we adjusting the retry and timeout values for the install?

fseldow: 600s means 10min, which already exceed the cse timeout. (though it seems timeout is not implement in dnf_install)

if the cse timeout, the extension error msg will only left csetimeout without correct exit code or helpful log.

But considering the timeout is not working for now, i will revert this retry and timeout values

fseldow:
too old pr and too many conflicts

PR #7826, opened 2026-02-07
chore(e2e): trying new listWatch logic for WaitNodeReady

djsly:
Try to augment the waitNodeReady logic to reconnect when the connection to the apiserer breaks

Fixes #

### DIFF
--- a/e2e/kube.go
+++ b/e2e/kube.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/Azure/agentbaker/e2e/config"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
-	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


Devinwong (approved):
Approved. It's always good when it's trying to make api server connection less flaky

e2e/kube.go
```diff
@@ -143,60 +142,126 @@ func (k *Kubeclient) WaitUntilPodRunning(ctx context.Context, namespace string,
...
 
-		var nodeFromEvent *corev1.Node
-		switch v := event.Object.(type) {
-		case *corev1.Node:
-			nodeFromEvent = v
-
-		default:
-			t.Logf("skipping object type %T", event.Object)
+		// Step 1: List existing nodes to get current state and resourceVersion
+		nodeList, err := k.Typed.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
r2k1, line 175: I think LIST+WATCH pattern been important for older versions. Where WATCH didn't produce initial events.

There is `SendInitialEvents` that seems to be set to True by default.
I don't think you need LIST.

djsly:
Merged in a different PR

PR #7990, opened 2026-02-27
feat: disable kernel lockdown mode for azurelinux 3.0 aks image by default

miz060:
Disable kernel lockdown by default to enable AKS customer installing their own custom GPU driver when using azurelinux 3.0. For trusted launch scenario, the AzureLinux 3.0 kernel auto-enables kernel lockdown mode when secure boot is detected via an OOT patch, so this won't introduce regression. Per PM discussion, we will follow up with:
- Add note in our public docs
- Add note in the AKS release notes
- Update our TSG with instructions on how to check if kernel lockdown is enabled or not

Fixes #

Validation:
- aks-image-build: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=1058779&view=results
- manually validated in AKS cluster that after this change, kernel lockdown is disabled when Secure Boot is off, and set to integrity when Secure Boot is on.


cameronmeissner (approved):

vhdbuilder/scripts/linux/mariner/tool_installs_mariner.sh
```diff
@@ -172,6 +172,15 @@ activateNfConntrack() {
     echo nf_conntrack >> /etc/modules-load.d/contrack.conf
 }
 
+# Remove lockdown=integrity from kernel cmdline for Azure Linux 3.0.
+# The AzureLinux 3.0 kernel has an OOT patch that automatically enables
+# lockdown when secure boot is detected.
+disableKernelLockdownCmdline() {
+    echo "Removing lockdown=integrity from kernel cmdline..."
+    dnf_install 120 5 25 grubby || exit $ERR_APT_INSTALL_TIMEOUT
+    grubby --update-kernel=ALL --remove-args="lockdown"
```
djsly: should you uninstall grubby after ?

djsly (approved):

PR #8056, opened 2026-03-10
feat: skip cloud-init ready report and add standalone report_ready script

awesomenix:
Bake experimental_skip_ready_report into VHD via cloud.cfg.d to skip
 cloud-init's built-in health ready report to Azure fabric. Add a
 standalone Python script (report_ready.py) that can be invoked from
 CSE to report ready at the appropriate time during node provisioning.

 Depends on canonical/cloud-init#6771.

I tried importing the cloud init library directly and copilot suggested this

The problems:

   1. Heavy import chain — importing cloudinit.sources.helpers.azure pulls in cloudinit.url_helper → requests, cloudinit.distros, cloudinit.subp, and dozens 
  more. It works on the VM (cloud-init's deps are installed), but it's a large dependency surface.
   2. get_metadata_from_fabric() requires a distro object — it's a full cloud-init Distro class instance (needed for ISO ejection). You'd have to either 
  construct one or monkey-patch it.
   3. report_success_to_host() depends on cloud-init's reporting handler registry — kvp.get_kvp_handler() looks up 
  instantiated_handler_registry.registered_items["telemetry"], which is only populated during cloud-init's normal boot. Outside of cloud-init's lifecycle, this 
  returns None and silently skips KVP reporting.
   4. Version coupling — your script would break if cloud-init refactors these internal APIs (they're not public API). Different VHDs may ship different 
  cloud-init versions.

  A middle ground: you could import just the KVP handler class directly and the low-level wireserver pieces, bypassing the high-level functions:

   from cloudinit.reporting.handlers import HyperVKvpReportingHandler
   handler = HyperVKvpReportingHandler()
   handler.write_key("PROVISIONING_REPORT", report_string)

  But the wireserver reporting (GoalState + health POST) has deep entanglement with url_helper, distro objects, and telemetry decorators that make it hard to
  call standalone.

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_start.sh
+++ b/parts/linux/cloud-init/artifacts/cse_start.sh
@@ -103,6 +103,15 @@ EVENT_JSON=$( jq -n \
 )
 echo ${EVENT_JSON} > ${EVENTS_LOGGING_DIR}${EVENTS_FILE_NAME}.json
 
+
+if [ -x /opt/azure/containers/report_ready.py ]; then
+    if [ "$EXIT_CODE" -eq 0 ]; then
+        python3 /opt/azure/containers/report_ready.py -v || echo "WARNING: Failed to report ready to Azure fabric"
+    else
+        python3 /opt/azure/containers/report_ready.py -v --failure --description "ExitCode: ${EXIT_CODE}, ${message_string}" || echo "WARNING: Failed to report failure to Azure fabric"
+    fi
+fi
+
 # force a log upload to the host after the provisioning script finishes
 # if we failed, wait for the upload to complete so that we don't remove
 # the VM before it finishes. if we succeeded, upload in the background
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -122,6 +122,15 @@ if ! isMarinerOrAzureLinux "$OS"; then
   disableNtpAndTimesyncdInstallChrony || exit 1
 fi
 
+skipCloudInitReadyReport || exit 1
+
+if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+  install -m 0644 /home/packer/DataSourceAzure.py \
+    /usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py || exit 1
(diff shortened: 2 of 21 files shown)


vhdbuilder/packer/install-dependencies.sh
```diff
@@ -122,6 +122,15 @@ if ! isMarinerOrAzureLinux "$OS"; then
   disableNtpAndTimesyncdInstallChrony || exit 1
 fi
 
+skipCloudInitReadyReport || exit 1
+
+if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
+  install -m 0644 /home/packer/DataSourceAzure.py \
+    /usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py || exit 1
```
djsly, line 129: this was extracted from the cloud-init repo ?

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -97,6 +97,15 @@ EVENT_JSON=$( jq -n \
 )
 echo ${EVENT_JSON} > ${EVENTS_LOGGING_DIR}${EVENTS_FILE_NAME}.json
 
+
+if [ -x /opt/azure/containers/report_ready.py ]; then
+    if [ "$EXIT_CODE" -eq 0 ]; then
+        python3 /opt/azure/containers/report_ready.py -v || echo "WARNING: Failed to report ready to Azure fabric"
```
djsly, line 109: this script was extracted from cloud-init repo ?

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1, line 137: Is it the main slow part that you skip before reporting node health and to optimise VM provision time?

Can we finish script execution early and execute some parts of the logic in the background?
Should it be part of the main blocking execution flow on error to avoid loosing valuable telemetry?

parts/linux/cloud-init/artifacts/cse_start.sh
awesomenix: AI generated, asked it to copy paste from cloud init repo and also implement the KVP logging as well.

PR #8178, opened 2026-03-26
feat: pass cse cmd generated from nbc to node controller

lilypan26:
Implements scriptless phase 2:

- both aksnodeconfig and cse cmd generated from NBC are passed to aks node controller
- only cse cmd is used to start provisioning
- logs when cse generated from nbc vs aksnodeconfig differ

Fixes #


parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -30,10 +30,14 @@ SYSTEMD_SUMMARY=$(systemd-analyze || true)
 CSE_ENDTIME_FORMATTED=$(date +"%F %T.%3N")
 EVENTS_FILE_NAME=$(date +%s%3N)
 EXECUTION_DURATION=$(($(date +%s) - $(date -d "$CSE_STARTTIME" +%s)))
-SCRIPTLESS_MODE=false
+SCRIPTLESS_MODE=""
```
awesomenix, line 33: `none`

aks-node-controller/app.go
```diff
@@ -171,6 +172,16 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) (*ProvisionRe
 		provisionResult.Error = fmt.Sprintf("build CSE command: %v", err)
 		return provisionResult, errors.New(provisionResult.Error)
 	}
+
+	if flags.NBCCMD != "" {
+		nbcCMD := exec.CommandContext(ctx, "/bin/bash", flags.NBCCMD)
+		if nbcCMD != cmd {
+			slog.Info("Detected difference between AKSNodeConfig command and NBC command, ", "originalCmd: ", cmd.String(), "nbcCmd: ", nbcCMD.String())
```
Devinwong: Is the plan also to do this check in `rp`?

Devinwong: I see. This is the app.go. Not e2e. Yes then we will also log this in rp.

aks-node-controller/app.go
```diff
@@ -171,6 +172,16 @@ func (a *App) Provision(ctx context.Context, flags ProvisionFlags) (*ProvisionRe
 		provisionResult.Error = fmt.Sprintf("build CSE command: %v", err)
 		return provisionResult, errors.New(provisionResult.Error)
 	}
+
+	if flags.NBCCMD != "" {
+		nbcCMD := exec.CommandContext(ctx, "/bin/bash", flags.NBCCMD)
+		if nbcCMD != cmd {
```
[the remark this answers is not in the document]
Devinwong: This comment looks valid. It's checking if it is the same pointer but I guess you wanted to check if the values are the same? On the other hand, the order of variables in aksnodeconfig could be different. Might need another way to check it, assumed this was what you were checking.

lilypan26:
continued here: https://github.com/Azure/AgentBaker/pull/8291

PR #8228, opened 2026-04-02
feat: skip MIG reboot for H100 and newer GPU architectures

ganeshkumarashok:
## Summary
- A100 (Ampere) requires a VM reboot after enabling MIG mode due to infoROM bit changes needing a platform-level reset on Azure. H100 (Hopper) and newer architectures support dynamic MIG mode changes without reboot — a GPU reset via `nvidia-smi --gpu-reset` is sufficient.
- Adds `MIG_NEEDS_REBOOT` CSE variable driven by a per-SKU map (`MIGRebootRequiredSizes`) in `gpu_components.go`. Unknown SKUs default to reboot-required for safety.
- H200 is also mapped as no-reboot-required (Hopper architecture). Future Blackwell SKUs just need `false` entries in the map.

## Changes
- **`pkg/agent/datamodel/gpu_components.go`** — `MIGRebootRequiredSizes` map (A100 → `true`, H100/H200 → `false`) and `MIGNeedsReboot()` helper
- **`pkg/agent/baker.go`** — `MIGNeedsReboot` template function
- **`parts/linux/cloud-init/artifacts/cse_cmd.sh`** — `MIG_NEEDS_REBOOT` template variable
- **`parts/linux/cloud-init/artifacts/cse_main.sh`** — conditional: A100 reboots, H100+ does `nvidia-smi --gpu-reset`
- **`e2e/validators.go`** — `ValidateNodeDidNotReboot` (checks `journalctl --list-boots` count == 1)
- **`e2e/scenario_gpu_managed_experience_test.go`** — `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG_H100_NoReboot`

## Test plan
- [ ] Existing `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG` (A100) still reboots and passes
- [ ] New `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG_H100_NoReboot` passes — MIG enabled, no reboot, GPU resources advertised
- [ ] Unit tests pass (`go test ./pkg/agent/...`)


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -459,12 +459,26 @@ function nodePrep {
...
-            # Commands such as `nvidia-smi --gpu-reset` may succeed,
-            # while commands such as `nvidia-smi -q` will show mismatched current/pending mig mode.
-            # this will not be required per nvidia for next gen H100.
-            REBOOTREQUIRED=true
+            # Determine if a reboot is needed to activate MIG mode based on GPU
+            # compute capability. Ampere (SM 8.x, e.g. A100) requires a VM reboot
+            # because changing the MIG mode bit in the infoROM needs a platform-level
+            # reset on Azure. Hopper and newer (SM >= 9.0, e.g. H100/H200) support
+            # dynamic MIG mode changes — nvidia-smi -mig 1 takes effect immediately.
+            gpu_compute_cap=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -1 | xargs)
```
timmy-wright, line 467: **`head -1` makes reboot decision from first GPU only (Simon + Kaylee)**

The node-wide `REBOOTREQUIRED` gate is set based on the first GPU's compute capability only. On a multi-GPU node with heterogeneous architectures (or if the first `nvidia-smi` output line is a transient bad read), this can silently skip a needed reboot or force an unnecessary one.

Suggestion: evaluate **all** output lines. If any line fails to parse or any value is `< 90`, set `REBOOTREQUIRED=true`. Log mixed-capability nodes explicitly.

```bash
while IFS= read -r cap; do
    cap_int=$(echo "${cap}" | tr -d '.')
    case "${cap_int}" in
        ""|*[!0-9]*)
            echo "Unable to parse compute_cap: '${cap}'"
            REBOOTREQUIRED=true ;;
        *)
            if [ "${cap_int}" -lt 90 ]; then REBOOTREQUIRED=true; fi ;;
    esac
done < <(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null)
```

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -459,12 +459,26 @@ function nodePrep {
...
+            # Strip the dot for integer comparison (e.g. "8.0" -> "80", "9.0" -> "90").
+            gpu_compute_cap_int=$(echo "${gpu_compute_cap}" | tr -d '.')
+            case "${gpu_compute_cap_int}" in
+                ""|*[!0-9]*)
+                    echo "Unable to determine GPU compute capability from nvidia-smi output: '${gpu_compute_cap}'"
```
timmy-wright, line 472: **Magic number `90` needs a named constant (Kaylee)**

`90` represents the first SM architecture that supports dynamic MIG enablement (Hopper, SM 9.0). A future reader has to reverse-engineer this from the string-munging context.

Suggestion: name it:
```bash
# SM >= 9.0 (Hopper) supports dynamic MIG mode changes without reboot.
MIG_DYNAMIC_MODE_MIN_SM=90
if [ "${gpu_compute_cap_int}" -lt "${MIG_DYNAMIC_MODE_MIN_SM}" ]; then
```
Blackwell (SM 10.0 → `100`) will naturally satisfy `100 >= 90`, so the logic is forward-compatible; the name just makes the intent explicit.

e2e/validators.go
```diff
@@ -2521,6 +2521,33 @@ func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile st
...
+
+	command := []string{
+		"set -exo pipefail",
+		// Diagnostics: dump the raw boot list so failures show what was counted.
+		"echo '=== journalctl --list-boots ==='",
+		"sudo journalctl --list-boots --no-pager || true",
+		"echo '=== uptime ==='",
+		"uptime || true",
+		// Count actual boot records via JSON output: one JSON object per boot, no header.
+		"boot_count=$(sudo journalctl --list-boots --no-pager --output=json | wc -l)",
```
timmy-wright, line 2543: **`boot_count=0` gives a false/misleading failure (Simon + Wash)**

If `journald --list-boots --output=json` returns no records (journal trimmed, or journald not running), `boot_count=0` and the assertion fails with the message "Node rebooted unexpectedly" — which is a false diagnosis. A rebooted node with a trimmed journal also produces `boot_count=1` and **passes** — which is a false negative.

Suggestion: branch explicitly:
```bash
case "$boot_count" in
  0) echo 'journald returned no boot records — cannot verify'; exit 1 ;;
  1) echo 'ok: exactly one boot' ;;
  *) echo "node rebooted: $boot_count boots recorded"; exit 1 ;;
esac
```
Or use `/proc/sys/kernel/random/boot_id` (stable per boot, readable without journald) as a more reliable indicator.

e2e/scenario_gpu_managed_experience_test.go
```diff
@@ -543,6 +543,107 @@ func Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG(t *testing.T) {
...
+				nbc.MigStrategy = "Single"
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.SKU.Name = to.Ptr("Standard_NC40ads_H100_v5")
+
+				extension, err := createVMExtensionLinuxAKSNode(t.Context(), vmss.Location)
+				require.NoError(t, err, "creating AKS VM extension")
+				vmss.Properties = addVMExtensionToVMSS(vmss.Properties, extension)
+			},
+			// This scenario is focused on H100 MIG no-reboot behavior. The default Linux
```
timmy-wright, line 579: **`SkipDefaultValidation: true` drops critical host health checks (Simon + Wash)**

Skipping default validation removes: `ValidateNodeCanRunAPod`, `ValidateAKSLogCollector`, `ValidateDiskQueueService`, `ValidateNoFailedSystemdUnits`, `ValidateKernelLogs`, `ValidateWaagentLog`, TLS bootstrapping, iptables/eBPF, stale kube binary checks, and more. The H100 no-reboot path can regress host health silently.

The comment justifies skipping due to transient failures on slow-booting H100 SKUs — but the fix for transient failures is an allowlist/retry, not a wholesale skip.

Suggestion: re-add at minimum:
```go
ValidateNodeCanRunAPod(ctx, s)
ValidateAKSLogCollector(ctx, s)
ValidateDiskQueueService(ctx, s)
ValidateNoFailedSystemdUnits(ctx, s)
ValidateKernelLogs(ctx, s)
ValidateWaagentLog(ctx, s)
```

timmy-wright (commented):
## Squad Review — AgentBaker PR #8228

**Reviewers:** Simon (defects) · Wash (testing) · Kaylee (design) · Zoe (security: no findings)

The compute-capability-based runtime detection is the right approach — cleaner than a static SKU map and naturally forward-compatible with Blackwell (SM 10.0 satisfies `>= 90`). No security concerns (Zoe). Two high-severity issues to address before merge.

### 🔴 High

- **`head -1` single-GPU decision wrong for multi-GPU nodes** (`cse_main.sh` ~line 467, inline) — node-wide reboot gate from first GPU only; heterogeneous or bad first-line read can skip a needed reboot.
- **`SkipDefaultValidation: true` drops critical host health checks** (`scenario_gpu_managed_experience_test.go` ~line 579, inline) — missing `ValidateNodeCanRunAPod`, kernel logs, failed systemd units, AKS log collector, waagent log, and more. H100 no-reboot path can regress silently.

### 🟡 Medium

- **`boot_count=0` gives false diagnosis** (`validators.go` ~line 2543, inline) — trimmed journal returns 0, fails with wrong message; rebooted node with trimmed journal passes with count=1.
- **`nvidia-smi --gpu-reset` path described but absent** (Wash) — PR description says H100 uses `nvidia-smi --gpu-reset` after MIG mode change; code only does `nvidia-smi -mig 1`. Neither the code nor the test validates `current == pending` MIG mode. Either implement/assert the reset path, or correct the PR description.
- **Runtime detect collapses driver failure into reboot-required** (Kaylee) — `nvidia-smi` not available at boot time collapses into `REBOOTREQUIRED=true`, mixing bootstrap health with hardware capability. Consider a hybrid: runtime detect first, VM-size fallback second.

### 🟢 Low

- **Magic number `90` undocumented** (`cse_main.sh` ~line 472, inline) — name it `MIG_DYNAMIC_MODE_MIN_SM=90`.
- **`compute_cap` parser untested** (Wash + Simon) — no unit tests for `8.0`, `9.0`, empty, malformed, or multi-GPU output. Extract to a helper + bats/shunit tests.
- **Single-GPU test cannot catch multi-line parsing bugs** (Simon) — PR changes generic logic for all Hopper+ but test covers one single-GPU SKU. Add shellspec unit tests for multi-line output.
- **`ValidateNodeDidNotReboot` too specialised for shared `validators.go`** (Kaylee) — co-locate with GPU scenario until genuinely reused.

PR #8268, opened 2026-04-09
fix(e2e): add DAG dependencies to prevent route table race condition

djsly:
## Problem

PR #8149 refactored `prepareCluster` to use concurrent DAG execution. This introduced a race condition between three operations that mutate the VNet/subnet:

1. **`getOrCreateBastion`** — creates AzureBastionSubnet
2. **`addFirewallRules`** — creates AzureFirewallSubnet + associates `abe2e-fw-rt` with `aks-subnet`
3. **`collectGarbageVMSS`** — deletes old VMSS → triggers AKS cloud-controller route reconciliation

Running them concurrently causes the AKS pod route table (`aks-agentpool-*-routetable`) to be **detached** from the subnet. The firewall route table (`abe2e-fw-rt`) wins the association, but has no pod routes. Result:

- `NetworkUnavailable: Waiting for cloud routes` — never cleared
- `cni plugin not initialized` — CNI config never written
- **65/154 E2E tests fail** (100% of kubenet/overlay, 0% of AzureCNI)

### Evidence from Azure

| Route Table | Routes | Subnet Association |
|---|---|---|
| `aks-agentpool-13663576-routetable` | 63 pod routes | **0 subnets** ❌ |
| `abe2e-fw-rt` | 2 (vnet-local + firewall) | 1 (`aks-subnet`) |

## Fix

Add DAG dependency edges to restore the sequential ordering for VNet-mutating operations:

- `addFirewallRules` / `addNetworkIsolatedSettings` → depends on `bastion` (serialize subnet writes)
- `collectGarbageVMSS` → depends on network setup (prevent VMSS deletion from racing with route table association)

All other tasks remain fully concurrent (kube client, identity, maintenance, subnet read, extract params).

## DAG Before vs After

```
BEFORE (#8149):  bastion ──┐
                            ├──► g.Wait()
                 firewall ─┤
                            │
                 gcVMSS ───┘   (all concurrent, race on VNet)

AFTER (this PR): bastion ──► firewall ──► gcVMSS
                                          (serialized VNet mutations)
                 kube ─────┐
                 identity ─┤
                 subnet ───┤  (still concurrent — no VNet writes)
                 maint ────┘
```

Fixes the E2E infrastructure failures observed on PR #8230 and the scriptless E2E gate.

### DIFF
--- a/.pipelines/.vsts-vhd-builder-pr-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-pr-windows.yaml
@@ -34,13 +34,14 @@ pr:
       - parts/windows
       - go.mod
       - go.sum
-      - e2e/scenario_win_test.go
+      - e2e/
       - staging/cse/windows/
 
     exclude:
       - vhdbuilder/release-notes
       - /**/*.md
       - .github/**
+      - e2e/scenario_test.go
       - .pipelines/scripts/verify_shell.sh
 
 pool:
--- a/.pipelines/.vsts-vhd-builder-pr-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-pr-windows.yaml
@@ -34,14 +34,13 @@ pr:
       - parts/windows
       - go.mod
       - go.sum
-      - e2e/
+      - e2e/scenario_win_test.go
       - staging/cse/windows/
 
     exclude:
       - vhdbuilder/release-notes
       - /**/*.md
       - .github/**
-      - e2e/scenario_test.go
       - .pipelines/scripts/verify_shell.sh
 
 pool:
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -78,6 +78,9 @@ func prepareCluster(ctx context.Context, clusterModel *armcontainerservice.Manag
 
 	g := dag.NewGroup(ctx)
 
+	// bastion creates AzureBastionSubnet — a VNet-level mutation that must
+	// finish before other subnet writes (firewall / network-isolated setup)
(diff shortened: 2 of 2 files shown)


timmy-wright (approved):

r2k1 (approved):

.pipelines/.vsts-vhd-builder-pr-windows.yaml
```diff
@@ -34,14 +34,13 @@ pr:
       - parts/windows
       - go.mod
       - go.sum
-      - e2e/
+      - e2e/scenario_win_test.go
```
r2k1: Any e2e change can affect windows tests

timmy-wright: Can you also add the validators code, and all the underlying vmss.go and other supporting files? 

In fact, it probably makes sense to add an exclude rule on scenario_test.go and trigger the windows builds on all other fiies.

timmy-wright (changes requested):
Not sure this is sensible - see comment.

.pipelines/.vsts-vhd-builder-pr-windows.yaml
r2k1: That's what we had before this PR.

Probably just revert changes in the file

djsly: why are we bulding new VHD when we make a change in the test logic ? doesn't make sense

r2k1 (approved):

timmy-wright (approved):

PR #8272, opened 2026-04-10
fix(e2e): retry WireServer blocked validation to eliminate flakes

djsly:
## Problem

The WireServer iptables FORWARD DROP rules blocking pod→WireServer traffic can be transiently absent during node setup. When kube-proxy or CNI flush/recreate iptables chains, the validation runs before the rules are re-applied, causing the unprivileged pod `curl` to get a non-28 exit code (connection succeeded instead of timing out).

The existing code already acknowledged this: `// This validator seems to be flaky, hard to catch`

## Fix

Extract the two WireServer checks into `validateWireServerBlocked()` with retry logic:
- Uses `wait.PollUntilContextTimeout` (10s interval, 1 min per check) — same pattern used elsewhere in E2E
- Retries until `exit code 28` (connection timeout = blocked) is observed
- On persistent failure: logs host iptables state and last curl result for debugging
- Keeps the same validation semantics — just tolerates transient iptables state
- Guards against nil result on early context cancellation

## Impact

- Eliminates false-negative E2E failures on WireServer validation  
- No change to what's being validated — still asserts pods can't reach WireServer
- Adds ~0s overhead on healthy nodes (first check passes immediately)
- Adds up to 2 min retry window on nodes with transient iptables state (1 min per check, 2 checks sequential)

### DIFF
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -14,7 +14,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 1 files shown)


e2e/validation.go
```diff
@@ -247,3 +235,48 @@ func getIPTablesRulesCompatibleWithEBPFHostRouting() (map[string][]string, []str
...
+
+	type wireServerCheck struct {
+		cmd  string
+		desc string
+	}
+
+	checks := []wireServerCheck{
+		{
+			cmd:  "curl http://168.63.129.16/machine/?comp=goalstate -H 'x-ms-version: 2015-04-05' -s --connect-timeout 4",
+			desc: "wireserver port 80 goalstate",
```
pdamianov-dev, line 254: nit: i dont see a port on the url, am I missing something?

djsly: http's default port is 80

https's defualt port is 443. 

default protocol ports, they don't need to be typed

pdamianov-dev (approved):

PR #8273, opened 2026-04-10
fix(e2e): strip --pod-infra-container-image for Kubernetes >= 1.35

djsly:
## Summary

Kubernetes 1.35 removed the `--pod-infra-container-image` kubelet flag (deprecated since 1.24). When AKS RP sends this flag for 1.35+ clusters, kubelet crash-loops with:

```
"command failed" err="failed to parse kubelet flag: unknown flag: --pod-infra-container-image"
```

The sandbox image is already configured separately via `containerd.toml` (`PodInfraContainerSpec`), so removing the CLI flag has no functional impact.

## Root Cause

Discovered during E2E gate failures where Random VHD tests selected AzureLinuxV3gen2 with kubelet 1.35.1. CSE reported exit 0 (kubelet hadn't crash-looped yet), but nodes never joined the cluster.

## Changes

| File | Change |
|------|--------|
| `pkg/agent/baker.go` | Version-gated `delete(kubeletFlags, "--pod-infra-container-image")` for K8s >= 1.35, both Linux and Windows paths |
| `aks-node-controller/parser/helper.go` | Strip flag in `getKubeletFlags()` for >= 1.35 |
| `parts/linux/cloud-init/artifacts/cse_config.sh` | Defensive strip before writing `KUBELET_FLAGS` (forward-compat: old AgentBaker + new kubelet) |
| `e2e/node_config.go` | Remove hardcoded flag from test configs (not needed, containerd config handles sandbox image) |

## Why the flag is safe to remove

- The sandbox (pause) image is configured via `containerd.toml` → `sandbox_image` setting
- `PodInfraContainerSpec` in `KubeBinaryConfig` populates containerd config independently
- `get_sandbox_image()` in `cse_helpers.sh` already has a 1.35 guard — only falls back to kubelet flags for < 1.35
- Flag was deprecated since Kubernetes 1.24; containerd has been the default since then

## Testing

- All `pkg/agent` tests pass
- All `aks-node-controller/parser` tests pass
- Snapshot test data regenerated

## Linked Work Items

AB#37487803

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -21,9 +21,8 @@ func baseKubeletConfig() *aksnodeconfigv1.KubeletConfig {
 	return &aksnodeconfigv1.KubeletConfig{
 		EnableKubeletConfigFile: true,
 		KubeletFlags: map[string]string{
-			"--cloud-provider":            "external",
-			"--kubeconfig":                "/var/lib/kubelet/kubeconfig",
-			"--pod-infra-container-image": "mcr.microsoft.com/oss/v2/kubernetes/pause:3.6",
+			"--cloud-provider": "external",
+			"--kubeconfig":     "/var/lib/kubelet/kubeconfig",
 		},
 		KubeletNodeLabels: map[string]string{
 			"agentpool":                               "nodepool2",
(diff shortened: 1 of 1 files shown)


cameronmeissner (approved):

e2e/node_config.go
```diff
@@ -21,9 +21,8 @@ func baseKubeletConfig() *aksnodeconfigv1.KubeletConfig {
 	return &aksnodeconfigv1.KubeletConfig{
 		EnableKubeletConfigFile: true,
 		KubeletFlags: map[string]string{
-			"--cloud-provider":            "external",
-			"--kubeconfig":                "/var/lib/kubelet/kubeconfig",
-			"--pod-infra-container-image": "mcr.microsoft.com/oss/v2/kubernetes/pause:3.6",
```
cameronmeissner, line 26: should we gate this on k8s version?

cameronmeissner (approved):

PR #8287, opened 2026-04-12
feat: cache versioned kubelet kubectl package binaries

awesomenix:
## What this does
 
 This change avoids unnecessary kubelet/kubectl package installation work during CSE when the corresponding binaries are already 
available on the VHD.
 
 Today, Ubuntu and Mariner/Azure Linux cache the kubelet/kubectl PMC packages on the VHD, but CSE still installs them via the package 
manager/runtime package flow. That has a few downsides:
 - it re-runs package installation logic during provisioning
 - on Ubuntu, `dpkg -i` triggers package `postinst` behavior for `kubelet.service`
 - we pay extra provisioning latency even though AKS already owns kubelet service configuration and startup
 
 This PR changes the flow so VHD build materializes versioned kubelet/kubectl binaries from the cached package artifacts into:
 - `/opt/bin/kubelet-<k8sVersion>`
 - `/opt/bin/kubectl-<k8sVersion>`
 
 Then, during CSE, if the requested version is already present in `/opt/bin`, we reuse the existing cache-first path and do the final 
rename into place instead of reinstalling from the package.
 
 ## Changes
 
 ### VHD build
 - extend `vhdbuilder/packer/install-dependencies.sh` so cached `kubelet`/`kubectl` package artifacts also produce versioned binaries 
under `/opt/bin`
 - support both:
   - Ubuntu: extract `/usr/bin/<tool>` from cached `.deb`
   - Mariner/Azure Linux: extract `/usr/bin/<tool>` from cached `.rpm`
 
 ### CSE
 - add a shared helper to detect whether versioned kubelet/kubectl binaries already exist in `/opt/bin`
 - update Ubuntu and Mariner package-based install paths to:
   - use the cached versioned binaries when available
   - fall back to existing package install behavior when the cache is not present
   - continue respecting `SHOULD_ENFORCE_KUBE_PMC_INSTALL=true`
 
 ### Tests
 - add/extend coverage for the cache-first kubelet/kubectl flow
 - extend Linux VHD content validation to verify package-backed kubelet/kubectl versions also materialize versioned binaries in 
`/opt/bin`
 
 ## Why
 
 This keeps kubelet/kubectl aligned with the existing `kubernetes-binaries` flow:
 - cache versioned binaries on the image
 - do a cheap final move during provisioning
 
 Expected benefits:
 - lower CSE latency
 - avoid unnecessary package-manager work during provisioning
 - avoid Ubuntu kubelet package `postinst` side effects when the VHD already has the requested binary
 
 ## Notes
 
 - this does **not** remove the existing package fallback path
 - if the versioned binary is missing, CSE still falls back to the current package installation behavior
 - `SHOULD_ENFORCE_KUBE_PMC_INSTALL=true` still forces the package path for validation / test scenarios

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Masterminds/semver"
 
 	"github.com/Azure/agentbaker/e2e/config"
+	"github.com/Azure/agentbaker/e2e/toolkit"
 	"github.com/Azure/agentbaker/pkg/agent"
 	"github.com/Azure/agentbaker/pkg/agent/datamodel"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
@@ -306,6 +307,14 @@ func nbcToAKSNodeConfigV1(nbc *datamodel.NodeBootstrappingConfiguration) *aksnod
 // this is what we previously used for bash e2e from e2e/nodebootstrapping_template.json.
 // which itself was extracted from baker_test.go logic, which was inherited from aks-engine.
 func baseTemplateLinux(t testing.TB, location string, k8sVersion string, arch string) *datamodel.NodeBootstrappingConfiguration {
+	customKubeProxyImage := fmt.Sprintf("mcr.microsoft.com/oss/kubernetes/kube-proxy:v%s", k8sVersion)
+	customKubeBinaryURL := fmt.Sprintf("https://packages.aks.azure.com/kubernetes/v%s/binaries/kubernetes-node-linux-%s.tar.gz", k8sVersion, arch)
+	is134OrAbove, pErr := toolkit.CheckK8sConstraint(k8sVersion, ">=1.34.0")
+	require.NoError(t, pErr, "failed to parse Kubernetes version")
+	if is134OrAbove {
+		customKubeProxyImage = ""
+		customKubeBinaryURL = ""
+	}
 	config := &datamodel.NodeBootstrappingConfiguration{
 		ContainerService: &datamodel.ContainerService{
 			ID:       "",
@@ -336,8 +345,8 @@ func baseTemplateLinux(t testing.TB, location string, k8sVersion string, arch st
 						UserAssignedID:                    "",
 						UserAssignedClientID:              "",
 						CustomHyperkubeImage:              "",
-						CustomKubeProxyImage:              fmt.Sprintf("mcr.microsoft.com/oss/kubernetes/kube-proxy:v%s", k8sVersion),
-						CustomKubeBinaryURL:               fmt.Sprintf("https://packages.aks.azure.com/kubernetes/v%s/binaries/kubernetes-node-linux-%s.tar.gz", k8sVersion, arch),
+						CustomKubeProxyImage:              customKubeProxyImage,
+						CustomKubeBinaryURL:               customKubeBinaryURL,
 						MobyVersion:                       "",
 						ContainerdVersion:                 "",
 						WindowsNodeBinariesURL:            "",
--- a/parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
+++ b/parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
@@ -213,8 +213,7 @@ installCredentialProviderFromPkg() {
 	echo "installing azure-acr-credential-provider package version: $packageVersion"
     mkdir -p "${CREDENTIAL_PROVIDER_BIN_DIR}"
     chown -R root:root "${CREDENTIAL_PROVIDER_BIN_DIR}"
-    installRPMPackageFromFile "azure-acr-credential-provider" "${packageVersion}" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
-    ln -snf /usr/bin/azure-acr-credential-provider "$CREDENTIAL_PROVIDER_BIN_DIR/acr-credential-provider"
+    installRPMPackageFromFile "azure-acr-credential-provider" "${packageVersion}" "${CREDENTIAL_PROVIDER_BIN_DIR}/acr-credential-provider" || exit "$ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT"
 }
 
   getPackageCacheRoot() {
@@ -233,8 +232,9 @@ installCredentialProviderFromPkg() {
 
 installKubeletKubectlFromPkg() {
     local desiredVersion="${1}"
-	  installRPMPackageFromFile "kubelet" $desiredVersion || exit $ERR_KUBELET_INSTALL_FAIL
-    installRPMPackageFromFile "kubectl" $desiredVersion || exit $ERR_KUBECTL_INSTALL_FAIL
+
+	  installRPMPackageFromFile "kubelet" "${desiredVersion}" "/opt/bin/kubelet" || exit "$ERR_KUBELET_INSTALL_FAIL"
+    installRPMPackageFromFile "kubectl" "${desiredVersion}" "/opt/bin/kubectl" || exit "$ERR_KUBECTL_INSTALL_FAIL"
 }
 
 installToolFromLocalRepo() {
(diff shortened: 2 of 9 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -221,8 +221,13 @@ testPackagesInstalled() {
...
         # kubernetes-binaries, namely, kubelet and kubectl are installed in a different way so we test them separately
-        # Intentionally remove leading 'v' from each element in the array
-        testKubeBinariesPresent "${PACKAGE_VERSIONS[@]#v}"
+        # Intentionally remove leading 'v' from each element in the array.
+        # ACL consumes kubelet/kubectl differently, so restrict this validation to Ubuntu and Mariner.
+        if [ "$OS" = "$UBUNTU_OS_NAME" ] || [ "$OS" = "$MARINER_OS_NAME" ]; then
+          testKubeBinariesPresent "${PACKAGE_VERSIONS[@]#v}"
+        else
+          echo "Skipping testKubeBinariesPresent for ${OS}${OS_VARIANT:+ ${OS_VARIANT}}"
+        fi
```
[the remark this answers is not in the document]
awesomenix: true, but that package doesnt exist anymore for > 1.34 which we are moving towards

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -262,6 +262,62 @@ unpackTgzToCNIDownloadsDIR() {
...
+    local rpm_file
+
+    rpm_file=$(find "${download_dir}" -maxdepth 1 -name "${package_name}-${version_no_epoch}*" -print -quit 2>/dev/null) || rpm_file=""
+    if [ -z "${rpm_file}" ]; then
+      echo "Failed to locate cached ${package_name} rpm for ${package_version}"
+      return 1
+    fi
+
+    rpm2cpio "${rpm_file}" | cpio -i --to-stdout "./usr/bin/${package_name}" "./usr/local/bin/${package_name}" | install -m0755 /dev/stdin "${binary_path}"
+  else
```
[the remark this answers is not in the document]
awesomenix: VHD build seems to be fine

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -222,7 +222,13 @@ installCredentialProviderFromPkg() {
 }
 
 installKubeletKubectlFromPkg() {
-    k8sVersion="${1}"
+    local k8sVersion="${1}"
+    if [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && hasCachedKubeBinariesForVersion "${k8sVersion}"; then
+        echo "Found cached kubernetes-binaries for ${k8sVersion}; installing from /opt/bin cache"
```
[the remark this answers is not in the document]
awesomenix: thats ok

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -233,6 +233,12 @@ installCredentialProviderFromPkg() {
 
 installKubeletKubectlFromPkg() {
     local desiredVersion="${1}"
+    if [ "${SHOULD_ENFORCE_KUBE_PMC_INSTALL}" != "true" ] && hasCachedKubeBinariesForVersion "${desiredVersion}"; then
+      echo "Found cached kubernetes-binaries for ${desiredVersion}; installing from /opt/bin cache"
```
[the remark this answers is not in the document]
awesomenix: thats ok

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -262,6 +262,62 @@ unpackTgzToCNIDownloadsDIR() {
...
+  elif isMarinerOrAzureLinux "$OS"; then
+    local rpm_file
+
+    rpm_file=$(find "${download_dir}" -maxdepth 1 -name "${package_name}-${version_no_epoch}*" -print -quit 2>/dev/null) || rpm_file=""
+    if [ -z "${rpm_file}" ]; then
+      echo "Failed to locate cached ${package_name} rpm for ${package_version}"
+      return 1
+    fi
+
+    rpm2cpio "${rpm_file}" | cpio -i --to-stdout "./usr/bin/${package_name}" "./usr/local/bin/${package_name}" | install -m0755 /dev/stdin "${binary_path}"
```
cameronmeissner, line 312: `rpm2cpio` `cpio` are always available on azlinux? I'm assuming so from looking at the man pages, but just want to double-check

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -231,15 +231,23 @@ testPackagesInstalled() {
...
         testPkgDownloaded "${name%-pmc}" "${downloadLocation}" "${PACKAGE_VERSIONS[@]}"
         continue
         ;;
+      "kubelet"|\
+      "kubectl")
+        testPkgDownloaded "${name}" "${downloadLocation}" "${PACKAGE_VERSIONS[@]}"
+        if [ "$OS" = "$UBUNTU_OS_NAME" ] || [ "$OS" = "$MARINER_OS_NAME" ]; then
+          testVersionedKubernetesPackageBinariesPresent "${name}" "${PACKAGE_VERSIONS[@]}"
+        else
+          echo "Skipping testVersionedKubernetesPackageBinariesPresent for ${OS}${OS_VARIANT:+ ${OS_VARIANT}}"
```
cameronmeissner, line 247: are the flatcar + ACL cases covered elsewhere?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -213,8 +213,7 @@ installCredentialProviderFromPkg() {
 	echo "installing azure-acr-credential-provider package version: $packageVersion"
     mkdir -p "${CREDENTIAL_PROVIDER_BIN_DIR}"
     chown -R root:root "${CREDENTIAL_PROVIDER_BIN_DIR}"
-    installRPMPackageFromFile "azure-acr-credential-provider" "${packageVersion}" || exit $ERR_CREDENTIAL_PROVIDER_DOWNLOAD_TIMEOUT
-    ln -snf /usr/bin/azure-acr-credential-provider "$CREDENTIAL_PROVIDER_BIN_DIR/acr-credential-provider"
```
cameronmeissner, line 217: are we removing the symlinking logic? is that intentional?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -233,8 +232,9 @@ installCredentialProviderFromPkg() {
 
 installKubeletKubectlFromPkg() {
     local desiredVersion="${1}"
-	  installRPMPackageFromFile "kubelet" $desiredVersion || exit $ERR_KUBELET_INSTALL_FAIL
-    installRPMPackageFromFile "kubectl" $desiredVersion || exit $ERR_KUBECTL_INSTALL_FAIL
+
+	  installRPMPackageFromFile "kubelet" "${desiredVersion}" "/opt/bin/kubelet" || exit "$ERR_KUBELET_INSTALL_FAIL"
```
cameronmeissner, line 236: nit: indentation seems off

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -403,22 +403,60 @@ installNvidiaManagedExpPkgFromCache() {
...
+
+    if [ -z "${binaryPath}" ]; then
+        echo "Failed to locate ${packageName} binary in ${rpmFile}"
+        rm -rf "${extractDir}"
+        return 1
+    fi
+
+    mkdir -p "$(dirname "${targetPath}")"
+
+    mv "${sourceBinary}" "${targetPath}"
```
cameronmeissner: copilot is also complaining about this - but it looks like `sourceBinary` should be `binaryPath` here

e2e/node_config.go
```diff
@@ -990,3 +998,15 @@ func pruneKubeletConfig(kubernetesVersion string, datamodel *datamodel.NodeBoots
 	}
 	return datamodel, nil
 }
+
+func checkK8sConstraint(kubernetesVersion string, constraintStr string) (bool, error) {
```
cameronmeissner: nit: can move this to toolkit/helpers?

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
awesomenix: yeah because install already handles moving the binary to right destination path

vhdbuilder/packer/install-dependencies.sh
awesomenix: yep there are other places which does this already https://github.com/Azure/AgentBaker/blob/main/parts/linux/cloud-init/artifacts/azlosguard/cse_install_osguard.sh#L55

vhdbuilder/packer/test/linux-vhd-content-test.sh
awesomenix: They use sysext so binaries will not be present in those cases

parts/linux/cloud-init/artifacts/mariner/cse_install_mariner.sh
```diff
@@ -403,22 +403,60 @@ installNvidiaManagedExpPkgFromCache() {
   done
 }
 
+extractRPMBinaryFromFile() {
```
cameronmeissner: nit: I'd rename to `extractBinaryFromRPM` or `extractBinaryFromRPMFile`

cameronmeissner (approved):

cameronmeissner (approved):

PR #8310, opened 2026-04-14
disable minor update for all packages by default

Devinwong:
Disable minor update for all packages by default (using wildcards). We unintentionally enabled some packages earlier.

Fixes #


cameronmeissner (approved):

djsly:
@Devinwong  not sure this makes sense, I would revert and revisit how we want to do it.

PR #8319, opened 2026-04-15
fix: use single e2e and run all flavours

awesomenix:
Use single e2e to run all scenarios, dont know if this will work

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -549,7 +549,7 @@ func Test_Ubuntu2204_Scriptless(t *testing.T) {
 }
 
 func Test_Ubuntu2204_Failure_Scriptless(t *testing.T) {
-	err := RunScenario(t, &Scenario{
+	RunScenario(t, &Scenario{
 		Description: "tests that a new ubuntu 2204 node using self contained installer can be properly bootstrapped",
 		Tags: Tags{
 			Scriptless: true,
@@ -567,16 +567,13 @@ func Test_Ubuntu2204_Failure_Scriptless(t *testing.T) {
 				config.BootstrappingConfig = nil
 				config.KubernetesCaCert = ""
 			},
-			ReturnErrorOnVMSSCreation: true,
+			ExpectedError: "API server connection check code: 51",
 		},
 	})
-
-	// Expect the error to contain API server connection failure since we provided invalid config
-	require.ErrorContains(t, err, "API server connection check code: 51")
 }
 
 func Test_Ubuntu2204_Early_Failure_Scriptless(t *testing.T) {
-	err := RunScenario(t, &Scenario{
+	RunScenario(t, &Scenario{
 		Description: "tests that a new ubuntu 2204 node using self contained installer can be properly bootstrapped",
 		Tags: Tags{
 			Scriptless: true,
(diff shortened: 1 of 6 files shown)


r2k1:
This looks way more complicated than it can be.


For reference, the existing VHD caching pattern at [test_helpers.go#L76-L89](https://github.com/Azure/AgentBaker/blob/main/e2e/test_helpers.go#L76-L89) already does exactly this — `copyScenario(original)`, mutate the copy, run it. The scriptless fanout should follow the same simple approach rather than introducing a profile/options resolution layer.

Something like:

```go
func RunScenario(t *testing.T, s *Scenario) error {
      t.Parallel()
      t.Run("original", func(t *testing.T) {
          t.Parallel()
          runScenario(t, copyScenario(s))
      })
      if config.EnableScriptless && s.AKSNodeConfigMutator != nil { // scriptless scenarios
          t.Run("scriptless", func(t *testing.T) {
              t.Parallel()
              s := copyScenario(s)
              s.DisableScriptlessCompilation = true
              runScenario(t, s)
          })
      }
      if config.EnableScriptlessNBC && s.AKSNodeConfigMutator == nil && !s.IsWindows() {
          t.Run("scriptless_nbc", func(t *testing.T) {
              t.Parallel()
              s := copyScenario(s)
              s.EnableScriptlessNBCCSECmd = true
              s.DisableScriptlessCompilation = true
              runScenario(t, s)
          })
      }
      return nil
    // ... existing code
}
```

No new types, no new files — flags go directly on `Scenario` (like `VHDCaching` already does), and `prepareAKSNode`/`vmss.go` read from `s.DisableScriptlessCompilation` instead of `config.Config.DisableScriptLessCompilation`.

Devinwong (approved):

Devinwong (approved):

e2e/test_helpers.go
```diff
@@ -236,10 +258,8 @@ func prepareAKSNode(ctx context.Context, s *Scenario) (*ScenarioVM, error) {
 	nbc, err := getBaseNBC(s.T, s.Runtime.Cluster, s.VHD)
 	require.NoError(s.T, err)
 
-	if config.Config.EnableScriptlessCSECmd {
-		nbc.EnableScriptlessCSECmd = true
-	}
-	if config.Config.EnableScriptlessNBCCSECmd {
+	nbc.EnableScriptlessCSECmd = true
```
r2k1, line 261: So, it's something we set always to true now?
Should it be inside NBC initialiser?

e2e/types.go
```diff
@@ -124,6 +124,9 @@ type Scenario struct {
 	// used to override the default location.
 	Location string
 
+	// EnableScriptlessNBCCSECmd indicates whether to enable scriptless NBCCSECmd in the scenario.
+	EnableScriptlessNBCCSECmd bool
```
r2k1, line 128: Maybe move it to `ScenarioRuntime` to separate configurable and non-configurable options

e2e/scenario_test.go
```diff
@@ -567,16 +567,13 @@ func Test_Ubuntu2204_Failure_Scriptless(t *testing.T) {
...
 		},
 	})
-
-	// Expect the error to contain API server connection failure since we provided invalid config
-	require.ErrorContains(t, err, "API server connection check code: 51")
 }
 
 func Test_Ubuntu2204_Early_Failure_Scriptless(t *testing.T) {
-	err := RunScenario(t, &Scenario{
+	RunScenario(t, &Scenario{
```
r2k1, line 576: Will these tests run in "default" and "scriptless" mode?

e2e/test_helpers.go
r2k1, line 76: I guess this now may have no return value?

r2k1 (approved):

e2e/test_helpers.go
awesomenix: Yep eventually will remove this parameter, will wait until then

e2e/scenario_test.go
awesomenix: default only if they have AKSNodeConfigMutator set

e2e/test_helpers.go
awesomenix: Agreed. Will change this in follow up PR

PR #8338, opened 2026-04-17
fix: speed up provisioning even more by 10s with localdns enabled

awesomenix:
- Configure skipping WAagent hold for scriptless case.
- Capture localdns logs for debugging
- prewarm coredns so its faster by 6s


parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -71,6 +71,11 @@ function basePrep {
         systemctl restart systemd-timesyncd
     fi
 
+    # pre-warm coredns by checking its version.
```
djsly, line 74: I'm a bit puzzled, do you know what this is doing in the background ? how is this supposed to accelerate/get ride of the 6sec delay you saw ?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -71,6 +71,11 @@ function basePrep {
...
+    if [ "${SHOULD_ENABLE_LOCALDNS}" = "true" ]; then
+        nohup /bin/sh -c '/opt/azure/containers/localdns/binary/coredns --version >/dev/null 2>&1' >/dev/null 2>&1 &
```
djsly, line 76: is there is an error with this command, should we care ? maybe log on a local filesystem ?

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -192,4 +192,5 @@ LOCALDNS_MEMORY_LIMIT="{{GetLocalDNSMemoryLimitInMB}}"
 LOCALDNS_GENERATED_COREFILE="{{GetGeneratedLocalDNSCoreFile}}"
 PRE_PROVISION_ONLY="{{GetPreProvisionOnly}}"
 CSE_TIMEOUT="{{GetCSETimeout}}"
+SKIP_WAAGENT_HOLD="{{GetSkipWaAgentHold}}"
```
djsly, line 195: should this be part of another PR ?  I'm confused how we can make the `GetSkipWaAgentHold` work ? isn't this risky ?

awesomenix: We already disable waagent hold for aks-node-controller, because of phase 2 i just reimplemented in CSE which will be only enabled (or skipped hold) in phase 2 https://github.com/Azure/AgentBaker/blob/44399f6ef74194132289eace4a237fa8e234d60f/aks-node-controller/parser/parser.go#L189

parts/linux/cloud-init/artifacts/cse_main.sh
awesomenix: dont care, since we are just using it for warming up the binary so in this case either warm it from azure storage or just paging it into memory

parts/linux/cloud-init/artifacts/cse_main.sh
awesomenix: it just reading the binary from azure storage into memory or just paging into memory, considering 6s slowness my guess is the hydration of the binary itself.

parts/linux/cloud-init/artifacts/cse_cmd.sh
Devinwong: walinuxagent is already centralized in components.json. I remember weeks ago there was a PR that handled this logic and should have had sufficient test coverage even on lower end VM. (Nishchay can confirm). If this is the case, then it should be safe to set SKIP_WAAGENT_HOLD to true completely in Scriptless phase 2 (EnableScriptlessNBCCSECmd = true)

awesomenix: i confirmed on A and B series on the lowest end as well

Devinwong (approved):
LGTM

PR #8344, opened 2026-04-17
fix: fall back to reinstall command to download security patch package if the package is already installed

YaoC:
…e if the package is already installed

fall back to reinstall command to download security patch package if the package is already installed

- `tdnf install` only works when a package with same version is not installed
- `tdnf reinstall` only works when a package with same version is installed

Fixes #


yewmsft (approved):
nit: `$dnf_download_output` is unquoted in a few places (e.g. `tee $dnf_download_output`, `grep -qi "already installed" $dnf_download_output`). Should be `"$dnf_download_output"` for shell best practice. Pre-existing issue so not blocking.

PR #8367, opened 2026-04-21
fix: containerd not ready on kata with artifactstreaming

SriHarsha001:

Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -421,8 +421,11 @@ ensureNoDupOnPromiscuBridge() {
 }
 
 ensureArtifactStreaming() {
-  waitForContainerdReady || exit $ERR_ARTIFACT_STREAMING_INSTALL
+  # Start overlaybd services before waiting for containerd, as containerd
+  # may block on connecting to the overlaybd snapshotter proxy plugin socket
+  # when artifact streaming is configured in the containerd config.
   retrycmd_if_failure 120 5 25 systemctl --quiet enable --now acr-mirror overlaybd-tcmu overlaybd-snapshotter
+  waitForContainerdReady || exit $ERR_ARTIFACT_STREAMING_INSTALL
   /opt/acr/bin/acr-config --enable-containerd 'azurecr.io'
 }
 
--- a/aks-node-controller/parser/templates/containerd.toml.gtpl
+++ b/aks-node-controller/parser/templates/containerd.toml.gtpl
@@ -5,12 +5,11 @@ root = "{{.KubeletConfig.GetContainerDataDir}}"{{- end}}
   sandbox_image = "{{ .KubeBinaryConfig.GetPodInfraContainerImageUrl }}"
   enable_cdi = true
   [plugins."io.containerd.grpc.v1.cri".containerd]
-    {{- if .GetIsKata }}
+    {{- if or .GetIsKata .GetEnableArtifactStreaming }}
     disable_snapshot_annotations = false
     {{- end}}
     {{- if .GetEnableArtifactStreaming }}
     snapshotter = "overlaybd"
-    disable_snapshot_annotations = false
     {{- end}}
     {{- if getEnableNvidia . }}
     default_runtime_name = "nvidia-container-runtime"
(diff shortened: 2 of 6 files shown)


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -421,8 +421,11 @@ ensureNoDupOnPromiscuBridge() {
 }
 
 ensureArtifactStreaming() {
-  waitForContainerdReady || exit $ERR_ARTIFACT_STREAMING_INSTALL
```
awesomenix, line 424: is it possible to call `ensureArtifactStreaming` before even containerd starts up? so that way we dont have failing containerd restarting unnecessarily

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -421,8 +421,11 @@ ensureNoDupOnPromiscuBridge() {
...
+  # Start overlaybd services before waiting for containerd, as containerd
+  # may block on connecting to the overlaybd snapshotter proxy plugin socket
+  # when artifact streaming is configured in the containerd config.
   retrycmd_if_failure 120 5 25 systemctl --quiet enable --now acr-mirror overlaybd-tcmu overlaybd-snapshotter
+  waitForContainerdReady || exit $ERR_ARTIFACT_STREAMING_INSTALL
   /opt/acr/bin/acr-config --enable-containerd 'azurecr.io'
```
awesomenix, line 429: this could be a problem, does it depend on containerd? i dont have any information on this binary, will check on VHD

awesomenix:
i looked into the binary on the VHD and checked that all it does only config modification, which means i had incorrectly added waitForContainerdReady.

``` 
/opt/acr/bin/acr-config --help
Configures acr-mirror related settings
Usage: acr-config [OPTIONS] [SUFFIX]
Arguments:
  [SUFFIX]
          List of image reference suffixes that should be handled by the mirror,
          **Note** Only relevant if using `--ns=_default`
Options:
  -d, --debug
          Enables debug logging
      --custom-root-dir <CUSTOM_ROOT_DIR>
          Custom root directory
          This will be used as the root of the config directory `./<custom_root>/etc/containerd`
          Note: This will always root the custom_root in the current directory, meaning this tool cannot create arbitrary directories
      --ns <NS>
          Namespace of images to mirror,
          ex. example.azurecr.io (Note that this **MUST** be the exact registry host)
          If left unset this will enable the mirror to handle all namespaces.
          Depending on the other settings the mirror will automatically recognize registry suffixes that will be enabled, and those that will not be enabled. For the latter a `hosts.toml` will be written in order for those regsitries to skip calling the mirror.
          [default: _default]
      --stream-format <STREAM_FORMAT>
          Streaming format the mirror will configure the hosts.toml file with.
          When clients make requests they will pass this information as a header so that the mirror can automatically recognize, if it should upgrade the request to resolve an image reference to it's streamable format.
          [default: overlaybd]
      --enable-containerd
          If set, this will update `/etc/containerd/config.toml` by:
          - setting the default `snapshotter`
          - setting the default `hosts-dir` for `cri` clients
          - enabling the proxy snapshotter plugin for the corresponding stream-format
      --mirror-port <MIRROR_PORT>
          Port acr-mirror is listening to
          [default: 8578]
  -h, --help
          Print help (see a summary with '-h')
 ```
 
```
Yes — that command looks like a config mutator, not a runtime client of containerd.
/opt/acr/bin/acr-config --enable-containerd 'azurecr.io'
from its own help means:
 - it edits /etc/containerd/config.toml
 - it writes registry hosts.toml files under containerd’s config tree
 - it enables containerd to use the overlaybd snapshotter and mirror settings
So it is integrating with containerd indirectly through config files, not by being the thing that handles container operations over the containerd socket.
In other words:
 - acr-config: writes config
 - containerd: later reads that config and changes behavior
 - acr-mirror / overlaybd: the services containerd is then configured to use
So “does it connect with containerd?” 
 - Not in the normal gRPC/runtime sense: probably no
 - Does it wire containerd up to use the mirror/snapshotter?: yes, by changing containerd’s config files
 
 
this means we should move the call ensureArtifactStreaming before containerd starts up and remove waitForContainerdReady
```

PR #8373, opened 2026-04-22
fix(e2e): wait on RegionalReplicationStatus before declaring SIG image usable

ganeshkumarashok:
## Summary

- Fix `GalleryImageNotFound` cascading failures in AgentBaker E2E. `ensureReplication` now waits on per-region `RegionalReplicationStatus.State == Completed` instead of trusting the parent provisioning LRO. The `vmss.go` retry loop classifies the 404 against actual replication state — failing fast on terminal `Failed`, blocking on `Replicating`, retrying once on the genuinely transient fabric-consistency case.

## Background

On PR #8228's GPU E2E run ([build 161141895](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=161141895)), 22 of 368 tests failed with the same shape:

```
failed to create VMSS after 10 retries:
RESPONSE 404: GalleryImageNotFound
"The gallery image .../2404gen2containerd/versions/1.1776726423.26642
 is not available in westus2 region."
```

The pipeline log shows `ensureReplication` had already emitted `Replication took: 12m2.7s` for that region — i.e. it returned success — and then VMSS create kept 404'ing for ~10 more minutes.

### Root cause

`replicatedToCurrentRegion` checks `PublishingProfile.TargetRegions`, which records *intent to replicate*, not the actual regional state. The Azure Compute Gallery Get API returns `provisioningState=Succeeded` once the parent LRO finishes, but `RegionalReplicationStatus.State` for the new region can remain `Replicating` for several more minutes before it transitions to `Completed`. During that window, the `imageReference` is unresolvable from `Microsoft.Compute/virtualMachineScaleSets` create — yielding the 404.

The existing retry loop in `vmss.go` (10 attempts × 5s = ~50 s) was orders of magnitude shorter than that window, and it surfaced the failure as `failed to create VMSS after 10 retries` — visually indistinguishable from a real test failure. PR authors had no way to tell that the change-under-test was never actually exercised.

### Why fix here, not in the pipeline YAML

Considered and rejected:

- **Per-pipeline YAML pre-flight gate** — duplicates the framework's responsibility, and tests in one E2E run pull different images for different regions, so a static gate would either over-replicate or under-gate.
- **`t.Skip` on `GalleryImageNotFound`** — silently masks real bugs. PR authors would think their change was validated when it wasn't.
- **Bumping the retry budget** — doesn't distinguish `Failed` from `Replicating`, wastes CRP quota, and still produces a misleading error message.

The framework already owns "ensure image is usable in this region" (`ensureReplication`). The right fix is to make that primitive honor its contract.

### `e2e/config/azure.go`

- New `waitForRegionalReplicationCompleted`: polls the gallery image version with `Expand=ReplicationStatusTypesReplicationStatus` (15s interval, 20 min cap) until `RegionalReplicationStatus.State == Completed`. Treats `Replicating`/`Unknown` as transient. Fails fast with attached `Details` on terminal `Failed`.
- `ensureReplication` now calls it on **both** paths:
  - "Region is already a target": the LRO may have completed but the regional replica may still be `Replicating` (this is exactly what bit PR #8228).
  - "Newly added target": the new replicate LRO can return `Succeeded` with the regional state still `Replicating`.
- New public `WaitForImageVersionReplicatedToRegion` for the retry loop in `vmss.go` — fetches the live version and delegates to the same poller.

### `e2e/vmss.go`

- `CreateVMSSWithRetry`'s classifier now distinguishes `GalleryImageNotFound` from `AllocationFailed`. On the 404 it consults the actual regional replication state via `WaitForImageVersionReplicatedToRegion`:
  - **Completed** → fall through, retry once (true fabric-consistency race, sub-minute window).
  - **Replicating** → block until **Completed** under the poller's 20-min cap, then retry.
  - **Failed** → return immediately with the precise infra error wrapped, instead of retrying nine more times.

### `e2e/config/azure_replication_test.go`

- New table-driven unit tests for `findRegionalReplicationStatus` covering nil status, case/whitespace normalization both ways (ARM returns `"West US 2"`, our location strings are `"westus2"`), nil-entry tolerance, and missing-region.

## Behavior changes

- **Before**: 22 noisy red `failed to create VMSS after 10 retries` failures every time a fresh VHD ran multi-region E2E. Test logs blamed the test code.
- **After**: `ensureReplication` blocks until the regional replica is actually serving — usual case, no retry pressure on CRP at all. If something is genuinely wrong with replication, you get a clear `regional replication of <id> to <region> failed: <Azure-provided details>` instead of `failed to create VMSS after 10 retries`. If it's the rare fabric-eventual-consistency case (LRO+ReplicationStatus both Completed but compute fabric not yet propagated), the existing retry catches it without a long fixed sleep.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./config/ -run TestFindRegionalReplicationStatus -v -vet=off` passes (vet-off only because of an unrelated pre-existing nit in `config/config.go:182`)
- [ ] CI: full e2e + unit-test gates on this PR
- [ ] Post-merge: rerun GPU E2E on PR #8228 and confirm the 22 GalleryImageNotFound failures don't recur and the MIG H100 change gets a real signal

### DIFF
--- a/e2e/config/azure.go
+++ b/e2e/config/azure.go
@@ -564,8 +564,13 @@ func (a *AzureClient) ensureReplication(ctx context.Context, image *Image, versi
 	}
 
 	if replicatedToCurrentRegion(version, location) {
-		toolkit.Logf(ctx, "Image version %s is already in region %s", *version.ID, location)
-		return nil
+		// Intent-to-replicate is registered in PublishingProfile.TargetRegions, but that
+		// does NOT mean the regional replica is actually serving traffic. The regional
+		// replication state can still be "Replicating" while the parent ProvisioningState
+		// is "Succeeded" — this is the source of GalleryImageNotFound 404s on VMSS create.
+		// Wait for the regional state itself before declaring success.
+		toolkit.Logf(ctx, "Image version %s is already a target of region %s; verifying regional replication state", *version.ID, location)
+		return a.waitForRegionalReplicationCompleted(ctx, image, version, location)
 	}
 	regions := make([]string, 0, len(version.Properties.PublishingProfile.TargetRegions))
 	for _, targetRegion := range version.Properties.PublishingProfile.TargetRegions {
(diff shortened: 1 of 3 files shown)


ganeshkumarashok:
Superseded by #8374 — recreated from a branch on Azure/AgentBaker so validate-pull-request-source passes (this repo blocks PRs from forks).

PR #8379, opened 2026-04-22
feat: implement budget timeout for apt_get_install

Devinwong:
The 3rd argument 'timeout' in apt_get_install was accepted but never used. Following the pattern from PR #8230, this change makes it work as a per-operation wall-clock budget:

- _apt_get_install: new maxBudget parameter (default 0). When > 0, tracks elapsed time and exits with code 2 if budget exceeded. Each apt-get install attempt is wrapped with 'timeout' capped to remaining budget. CSE timeout guards (check_cse_timeout) are added before each attempt, gated on CSE_STARTTIME_SECONDS to avoid noise during VHD build.

- apt_get_install: passes the timeout value as maxBudget only during real CSE runs (CSE_STARTTIME_SECONDS set). VHD build callers are unaffected.

- apt_get_install_from_local_repo: passes explicit 0 budget to _apt_get_install to maintain existing behavior.

Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
@@ -58,20 +58,23 @@ _apt_get_install() {
     shift && shift && shift && shift
     local packages=("${@}")
 
+    local hasBudget=false
+    if [ "${maxBudget}" -gt 0 ]; then
+        hasBudget=true
+    fi
+
     local opStartTime
     opStartTime=$(date +%s)
 
     for i in $(seq 1 $retries); do
-        # CSE timeout guard — only during real CSE runs
-        if [ -n "${CSE_STARTTIME_SECONDS:-}" ]; then
-            if ! check_cse_timeout; then
-                echo "CSE timeout approaching, exiting apt_get_install early." >&2
-                return 2
-            fi
+        # CSE timeout guard
+        if ! check_cse_timeout; then
+            echo "CSE timeout approaching, exiting apt_get_install early." >&2
+            return 2
         fi
 
         # Per-operation budget check
-        if [ "${maxBudget}" -gt 0 ]; then
+        if [ "$hasBudget" = true ]; then
             local opElapsed
             opElapsed=$(( $(date +%s) - opStartTime ))
             if [ "$opElapsed" -ge "$maxBudget" ]; then
@@ -86,17 +89,19 @@ _apt_get_install() {
 
         # Cap per-attempt timeout to the remaining budget so a single attempt
         # cannot overrun the operation window.
-        local aptTimeout=""
-        if [ "${maxBudget}" -gt 0 ]; then
+        local install_ok=false
+        if [ "$hasBudget" = true ]; then
             local remaining=$(( maxBudget - ( $(date +%s) - opStartTime ) ))
             if [ "$remaining" -lt 1 ]; then
                 echo "apt_get_install budget of ${maxBudget}s exceeded, exiting early." >&2
                 return 2
             fi
-            aptTimeout="timeout $remaining"
+            timeout "$remaining" apt-get install ${apt_opts} -o Dpkg::Options::="--force-confold" --no-install-recommends "${packages[@]}" && install_ok=true
+        else
+            apt-get install ${apt_opts} -o Dpkg::Options::="--force-confold" --no-install-recommends "${packages[@]}" && install_ok=true
         fi
 
-        if ${aptTimeout} apt-get install ${apt_opts} -o Dpkg::Options::="--force-confold" --no-install-recommends "${packages[@]}"; then
+        if [ "$install_ok" = true ]; then
             echo "Executed apt-get install \"${packages[*]}\" $i times"
             wait_for_apt_locks
             DEBIAN_FRONTEND=noninteractive apt-get clean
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -54,15 +54,50 @@ _apt_get_install() {
...
+        # CSE timeout guard — only during real CSE runs
+        if [ -n "${CSE_STARTTIME_SECONDS:-}" ]; then
+            if ! check_cse_timeout; then
+                echo "CSE timeout approaching, exiting apt_get_install early." >&2
+                return 2
+            fi
+        fi
+
+        # Per-operation budget check
+        if [ "${maxBudget}" -gt 0 ]; then
```
cameronmeissner: can `maxBudget` ever be <= 0?

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -54,15 +54,50 @@ _apt_get_install() {
...
 
-        if apt-get install ${apt_opts} -o Dpkg::Options::="--force-confold" --no-install-recommends "${@}"; then
-            echo "Executed apt-get install \"${packages[@]}\" $i times"
+        # Cap per-attempt timeout to the remaining budget so a single attempt
+        # cannot overrun the operation window.
+        local aptTimeout=""
+        if [ "${maxBudget}" -gt 0 ]; then
+            local remaining=$(( maxBudget - ( $(date +%s) - opStartTime ) ))
+            if [ "$remaining" -lt 1 ]; then
+                echo "apt_get_install budget of ${maxBudget}s exceeded, exiting early." >&2
```
cameronmeissner, line 96: is this only piping to stderr? we should ensure that these messages also show up in stdout I'd say

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -54,15 +54,50 @@ _apt_get_install() {
...
+                return 2
+            fi
+            aptTimeout="timeout $remaining"
```
cameronmeissner: nit: would prefer we just use `timeout $remaining apt-get install ...` directly

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -72,14 +107,31 @@ _apt_get_install() {
         if [ $i -eq $retries ]; then
             return 1
         else
+            # Check budget/CSE again before sleeping
+            if [ -n "${CSE_STARTTIME_SECONDS:-}" ] && ! check_cse_timeout; then
+                echo "CSE timeout approaching, exiting apt_get_install early." >&2
+                return 2
+            fi
+            if [ "${maxBudget}" -gt 0 ]; then
```
cameronmeissner: seems redundant to check `maxBudget` multiple times like this

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -72,14 +107,31 @@ _apt_get_install() {
         if [ $i -eq $retries ]; then
             return 1
         else
+            # Check budget/CSE again before sleeping
+            if [ -n "${CSE_STARTTIME_SECONDS:-}" ] && ! check_cse_timeout; then
```
cameronmeissner: can we bake the logic of checking whether `CSE_STARTTIME_SECONDS` is non-empty directly into `check_cse_timeout`?

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
Devinwong: <= 0 (0 if unset) means no budget cap. Technically someone can call this function with <= 0.

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
Devinwong: Good call. These messages go to stderr because they're warnings/errors, following the same convention as Sylvain's PR #8230's budget messages and check_cse_timeout. In practice, cse_start.sh captures both streams together (>> /var/log/azure/cluster-provision.log 2>&1), so these
 messages do appear in the provisioning log. Writing to both stdout and stderr would duplicate them in that combined log.
But if you think this is helpful for debugging, I can add it back.

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
Devinwong: makes sense. Will update it.

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
Devinwong: Agreed. Updated to use a bool to save the check

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
Devinwong: There is already that checking logic inside. I will remove this one.

djsly (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -72,14 +112,31 @@ _apt_get_install() {
...
+            fi
+            if [ "$hasBudget" = true ]; then
+                local postElapsed=$(( $(date +%s) - opStartTime ))
+                if [ "$postElapsed" -ge "$maxBudget" ]; then
+                    echo "apt_get_install budget of ${maxBudget}s exceeded after ${postElapsed}s, exiting early." >&2
+                    return 2
+                fi
+            fi
             sleep $wait_sleep
             apt_get_update
```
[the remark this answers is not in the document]
Devinwong: `apt_get_update` slight runs over max budget should be acceptable. Otherwise over-engineered.

PR #8414, opened 2026-04-28
fix: [NI]Use cloud-specific ARM endpoint for IMDS token in ORAS login

charleswool:
The oras_login_with_kubelet_identity (Linux) and Invoke-OrasLogin (Windows) functions hardcoded https://management.azure.com/ as the ARM resource endpoint in the IMDS token request URL. This causes authentication failures in sovereign clouds (e.g. Fairfax) where the correct endpoint differs.

Added get_arm_resource_endpoint / Get-ArmResourceEndpoint helpers that resolve the ARM endpoint from the cloud environment variable (TARGET_CLOUD on Linux, $TargetEnvironment on Windows):
- AzureUSGovernmentCloud -> https://management.usgovcloudapi.net/
- AzureChinaCloud        -> https://management.chinacloudapi.cn/
- USNatCloud             -> https://management.azure.eaglex.ic.gov/
- USSecCloud             -> https://management.azure.microsoft.scloud/
- default (public)       -> https://management.azure.com/

Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -1259,6 +1259,20 @@ assert_refresh_token() {
     return 0
 }
 
+# Returns the ARM resource endpoint for the current cloud environment.
+# Uses TARGET_CLOUD (set by cse_cmd.sh from GetTargetEnvironment) to
+# determine the correct endpoint. Each sovereign cloud has its own ARM
+# management endpoint required for IMDS token requests.
+get_arm_resource_endpoint() {
+    case "${TARGET_CLOUD:-}" in
+        AzureUSGovernmentCloud)  echo "https://management.usgovcloudapi.net/" ;;
+        AzureChinaCloud)         echo "https://management.chinacloudapi.cn/" ;;
+        USNatCloud)              echo "https://management.azure.eaglex.ic.gov/" ;;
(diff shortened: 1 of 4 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1259,6 +1259,20 @@ assert_refresh_token() {
     return 0
 }
 
+# Returns the ARM resource endpoint for the current cloud environment.
+# Uses TARGET_CLOUD (set by cse_cmd.sh from GetTargetEnvironment) to
+# determine the correct endpoint. Each sovereign cloud has its own ARM
+# management endpoint required for IMDS token requests.
+get_arm_resource_endpoint() {
```
fseldow, line 1266: can we generate one go function instead of in script like `GetOutBoundCmd` then both linux and windows could share `GetArmResourceEndpoint`

fseldow: or like
```
"GetMCRRepositoryBase": func() string {
			if config.CloudSpecConfig.KubernetesSpecConfig.MCRKubernetesImageBase == "" {
				return "mcr.microsoft.com"
			}
			return config.CloudSpecConfig.KubernetesSpecConfig.MCRKubernetesImageBase
		},
```

PR #8424, opened 2026-04-29
fix: use cloud-specific ARM endpoint for IMDS token in ORAS login

charleswool:
The oras_login_with_kubelet_identity (Linux) and Invoke-OrasLogin (Windows) functions hardcoded https://management.azure.com/ as the ARM resource endpoint in the IMDS token request URL. This causes authentication failures in sovereign clouds (e.g. Fairfax) where the correct endpoint differs.

Added get_arm_resource_endpoint / Get-ArmResourceEndpoint helpers that resolve the ARM endpoint from the cloud environment variable (TARGET_CLOUD on Linux, $TargetEnvironment on Windows):
- AzureUSGovernmentCloud -> https://management.usgovcloudapi.net/
- AzureChinaCloud        -> https://management.chinacloudapi.cn/
- default (public)       -> https://management.azure.com/

Fixes #

### DIFF
--- a/aks-node-controller/parser/helper.go
+++ b/aks-node-controller/parser/helper.go
@@ -570,22 +570,34 @@ func getTargetCloud(v *aksnodeconfigv1.Configuration) string {
 }
 
 // getArmResourceEndpoint returns the ARM resource endpoint to use as the IMDS
-// "resource" parameter when acquiring an AAD token. The endpoint is sourced
-// from the resourceManagerEndpoint field of CustomEnvJsonContent, which is
-// populated by AKS RP. Returns empty string if not present; callers handle
-// the empty case (e.g. cse_helpers.sh defaults to the public ARM endpoint).
+// "resource" parameter when acquiring an AAD token.
+//   - For AKS custom clouds (Azure Stack): sourced from
+//     CustomEnvJsonContent.resourceManagerEndpoint, which is populated by AKS RP.
+//   - For public sovereign clouds (Fairfax / Mooncake): mapped by cloud name.
+//     These endpoints are public knowledge so hardcoding is acceptable.
+//   - For Azure public cloud (and any unknown): returns empty; cse_helpers.sh
+//     defaults to https://management.azure.com/.
 func getArmResourceEndpoint(v *aksnodeconfigv1.Configuration) string {
-	raw := v.GetCustomCloudConfig().GetCustomEnvJsonContent()
-	if raw == "" {
-		return ""
+	if getIsAksCustomCloud(v.GetCustomCloudConfig()) {
+		raw := v.GetCustomCloudConfig().GetCustomEnvJsonContent()
+		if raw == "" {
+			return ""
+		}
+		var env struct {
+			ResourceManagerEndpoint string `json:"resourceManagerEndpoint"`
+		}
+		if err := json.Unmarshal([]byte(raw), &env); err != nil {
+			return ""
+		}
+		return env.ResourceManagerEndpoint
 	}
-	var env struct {
-		ResourceManagerEndpoint string `json:"resourceManagerEndpoint"`
+	switch getCloudTargetEnv(v) {
+	case "AzureUSGovernmentCloud":
+		return "https://management.usgovcloudapi.net/"
+	case "AzureChinaCloud":
+		return "https://management.chinacloudapi.cn/"
 	}
-	if err := json.Unmarshal([]byte(raw), &env); err != nil {
-		return ""
-	}
-	return env.ResourceManagerEndpoint
+	return ""
 }
 
 func getAzureEnvironmentFilepath(v *aksnodeconfigv1.Configuration) string {
--- a/aks-node-controller/parser/helper.go
+++ b/aks-node-controller/parser/helper.go
@@ -575,8 +575,7 @@ func getTargetCloud(v *aksnodeconfigv1.Configuration) string {
 //     CustomEnvJsonContent.resourceManagerEndpoint, which is populated by AKS RP.
 //   - For public sovereign clouds (Fairfax / Mooncake): mapped by cloud name.
 //     These endpoints are public knowledge so hardcoding is acceptable.
-//   - For Azure public cloud (and any unknown): returns empty; cse_helpers.sh
-//     defaults to https://management.azure.com/.
+//   - Default (Azure public cloud and any unknown): https://management.azure.com/.
 func getArmResourceEndpoint(v *aksnodeconfigv1.Configuration) string {
 	if getIsAksCustomCloud(v.GetCustomCloudConfig()) {
 		raw := v.GetCustomCloudConfig().GetCustomEnvJsonContent()
@@ -597,7 +596,7 @@ func getArmResourceEndpoint(v *aksnodeconfigv1.Configuration) string {
 	case "AzureChinaCloud":
 		return "https://management.chinacloudapi.cn/"
 	}
-	return ""
+	return "https://management.azure.com/"
 }
 
 func getAzureEnvironmentFilepath(v *aksnodeconfigv1.Configuration) string {
--- a/aks-node-controller/parser/helper.go
+++ b/aks-node-controller/parser/helper.go
@@ -569,6 +569,25 @@ func getTargetCloud(v *aksnodeconfigv1.Configuration) string {
 	return getTargetEnvironment(v)
 }
 
+// getArmResourceEndpoint returns the ARM resource endpoint to use as the IMDS
+// "resource" parameter when acquiring an AAD token. For AKS custom clouds (Azure
+// Stack), it extracts resourceManagerEndpoint from the custom env JSON. For all
+// other clouds, it maps the cloud name to the corresponding sovereign endpoint.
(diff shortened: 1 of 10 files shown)


pkg/agent/datamodel/sig_config.go
```diff
@@ -76,6 +76,23 @@ func GetCloudTargetEnv(location string) string {
 	}
 }
 
+// GetArmResourceEndpoint returns the ARM resource endpoint for the given cloud name.
+// This is used as the "resource" parameter in IMDS token requests.
+func GetArmResourceEndpoint(cloudName string) string {
```
fseldow: seems not necessary to put the function under datamodel.
What about move the function to baker.go and keep it as private function?

staging/cse/windows/networkisolatedclusterfunc.ps1
```diff
@@ -172,7 +172,7 @@ function Invoke-OrasLogin {
     }
 
     # Get AAD Access Token using Managed Identity Metadata Service
-    $accessUrl = "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/&client_id=$ClientID"
+    $accessUrl = "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=${ArmResourceEndpoint}&client_id=$ClientID"
```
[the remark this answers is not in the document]
fseldow: same suggestion as the AI. to set one default value if global variable is empty.
Because in windows, the release of kuberneteswindowssetup.ps1 and staging/* could be in different release

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1280,7 +1280,7 @@ oras_login_with_kubelet_identity() {
     fi
 
     set +x
-    access_url="http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/&client_id=$client_id"
+    access_url="http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=${ARM_RESOURCE_ENDPOINT}&client_id=$client_id"
```
fseldow: same suggestion as the windows, leave one default local variable to setup default value as public cloud arm endpoint to avoid suprise

awesomenix: +1 something like

`${ARM_RESOURCE_ENDPOINT-https://management.azure.com/}`

pkg/agent/datamodel/sig_config.go
charleswool: good point, updated

parts/linux/cloud-init/artifacts/cse_helpers.sh
charleswool: added default fallback for both windows and linux

charleswool:
@copilot apply changes based on the comments in [this thread](https://github.com/Azure/AgentBaker/pull/8424#pullrequestreview-4193902233)

charleswool:
@copilot there are some test failure in remote, can you check: https://github.com/Azure/AgentBaker/pull/8424

pkg/agent/baker.go
```diff
@@ -612,6 +612,27 @@ func validateAndSetWindowsNodeBootstrappingConfiguration(config *datamodel.NodeB
 	}
 }
 
+// getArmResourceEndpoint returns the ARM resource endpoint for the given cloud name.
+// This is used as the "resource" parameter in IMDS token requests.
```
fseldow: no need for this line comment
This is used as the "resource" parameter in IMDS token requests.

pkg/agent/baker.go
```diff
@@ -612,6 +612,27 @@ func validateAndSetWindowsNodeBootstrappingConfiguration(config *datamodel.NodeB
...
+func getArmResourceEndpoint(cloudName string) string {
```
fseldow: put it after `getContainerServiceFuncMap`? since other FuncMap put their helper function after `getContainerServiceFuncMap`

fseldow (approved):
/lgtm

charleswool:
@copilot fseldow has two more comments on remote, please address and check Agentbaker E2E / GPU E2E: failure

charleswool:
@copilot some e2e may just be flacky failure. I will add the test log for your reference here

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -114,6 +114,7 @@ CONTAINERD_ULIMITS="{{GetContainerdUlimitString}}"
 {{/* see GetCustomEnvironmentJSON for more weirdness. */}}
 TARGET_CLOUD="{{- if IsAKSCustomCloud -}} AzureStackCloud {{- else -}} {{GetTargetEnvironment}} {{- end -}}"
 TARGET_ENVIRONMENT="{{GetTargetEnvironment}}"
+ARM_RESOURCE_ENDPOINT="{{GetArmResourceEndpoint}}"
```
cameronmeissner, line 117: we should update env vars for aks-node-controller as well: https://github.com/Azure/AgentBaker/blob/4828603cb378958ab198a8e87163e397705db2d8/aks-node-controller/parser/parser.go#L33

cameronmeissner: cc @awesomenix @Devinwong - maybe worth adding a new GH action that validates any changes within cse_cmd.sh are reconciled within aks-node-controller?

charleswool: addressed, kindly revisit

cameronmeissner (approved):

cameronmeissner (changes requested):
would shouldn't hardcode resource manager endpoints

pkg/agent/baker.go
```diff
@@ -1016,6 +1016,12 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return GetCloudTargetEnv(cs.Location)
 		},
+		"GetArmResourceEndpoint": func() string {
+			if cs.IsAKSCustomCloud() && cs.Properties.CustomCloudEnv.ResourceManagerEndpoint != "" {
```
cameronmeissner: `cs.Properties.CustomCloudEnv.ResourceManagerEndpoint` should always be getting set by the resource provider, thus we shouldn't need the fallback mechanism

pkg/agent/baker.go
```diff
@@ -1016,6 +1016,12 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			}
 			return GetCloudTargetEnv(cs.Location)
 		},
+		"GetArmResourceEndpoint": func() string {
+			if cs.Properties != nil && cs.Properties.CustomCloudEnv != nil {
```
cameronmeissner, line 1021: I guess this will only be populated in custom clouds - you still need to handle the FF/MC case explicitly based on the cloud name (also need to handle public, but you already default to the public endpoint in scripts directly which should be fine)

charleswool: Good catch, updated to keep the custom-cloud path and add explicit mapping for ff/mc. Also mirror the same logic in help.go

aks-node-controller/parser/helper.go
```diff
@@ -569,6 +569,37 @@ func getTargetCloud(v *aksnodeconfigv1.Configuration) string {
...
+			return ""
+		}
+		return env.ResourceManagerEndpoint
+	}
+	switch getCloudTargetEnv(v) {
+	case "AzureUSGovernmentCloud":
+		return "https://management.usgovcloudapi.net/"
+	case "AzureChinaCloud":
+		return "https://management.chinacloudapi.cn/"
+	}
```
cameronmeissner, line 598: nit: looking at it, just for safety, I'd probably explicitly default to public cloud here

pkg/agent/baker.go
```diff
@@ -1016,6 +1016,22 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
...
+			}
+			// Public sovereign clouds (FF/MC) — endpoints are public knowledge so
+			// it's safe to map by cloud name. Public cloud falls through to empty;
+			// scripts default to https://management.azure.com/.
+			switch GetCloudTargetEnv(cs.Location) {
+			case datamodel.AzureUSGovernmentCloud:
+				return "https://management.usgovcloudapi.net/"
+			case datamodel.AzureChinaCloud:
+				return "https://management.chinacloudapi.cn/"
+			}
```
cameronmeissner, line 1031: same nit here

cameronmeissner (approved):

aks-node-controller/parser/helper.go
charleswool: SGTM, updated

pkg/agent/baker.go
charleswool: SGTM, updated

fseldow (approved):

cameronmeissner (approved):

PR #8426, opened 2026-04-29
fix: enable hairpin mode for ubuntu 24.04 with kubenet bridge plugin

levimm:
This is the bug report: https://github.com/Azure/AKS/issues/5669#issuecomment-4335041448

We can repro the issue with kubenet + Ubuntu 24.04 (kernel 6.8), but couldn't repro for kubenet + azlinux3 (kernel 6.6) or other lower version of Ubuntu.

Why it worked before with hairpinmode off: when kube-proxy DNAT'd the ClusterIP to the same PodIP, kernels ≤ 6.6 would "promote" the bridged frame to the L3 path (ip_forward()), which ran nat POSTROUTING / KUBE-POSTROUTING, which SNAT'd src to 10.244.0.1 (cbr0's IP), then re-emitted the packet. By the time the frame returned to br_forward(), src MAC = host MAC (≠ veth peer MAC), so split-horizon no longer applied even without hairpin_mode.

Below investigation from AI, I don't have enough kernel knowledge to tell:
> There are two plausible mechanisms (both consistent with the dmesg data, hard to disambiguate without kernel source diving):
6.6 takes a different route earlier: After DNAT, instead of continuing through br_forward → should_deliver, the older code path notices the DNAT changed the dst and re-classifies the frame, sending it through ip_forward() → IPv4 POSTROUTING. POSTROUTING runs, MASQUERADE rewrites src → 10.244.0.1 (cbr0 IP), then the packet is re-sent as a fresh L3 packet, which on emit through cbr0 has src MAC = host MAC (different from veth_A's peer MAC) → hairpin no longer same-port → frame delivers.
6.6 treats DNAT'd frames as eligible to bypass split-horizon: The pre-rework br_nf_pre_routing_finish_bridge may have explicitly handled the DNAT-changed-dst case by skipping should_deliver's split-horizon check (or by re-injecting via L3).

Although it looks like it's kernel related, we still want to scope the fix to ubuntu only for now.

Fixes #

### DIFF
--- a/aks-node-controller/parser/helper.go
+++ b/aks-node-controller/parser/helper.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
@@ -39,6 +40,8 @@ import (
 var (
 	//go:embed templates/kubenet-cni.json.gtpl
 	kubenetTemplateContent []byte
+	//go:embed templates/kubenet-cni-hairpin.json.gtpl
+	kubenetTemplateHairpinContent []byte
 	//go:embed  templates/containerd.toml.gtpl
 	containerdConfigTemplateText string
 	//nolint:gochecknoglobals
@@ -144,8 +147,36 @@ func getStringifiedStringArray(arr []string, delimiter string) string {
 }
 
 // getKubenetTemplate returns the base64 encoded Kubenet template.
+// On Ubuntu 24.04+ (kernel 6.8+), promiscuous mode no longer implicitly enables
+// hairpin forwarding in the bridge. Use explicit hairpinMode for those nodes.
 func getKubenetTemplate() string {
+	if needsExplicitHairpinMode() {
+		return base64.StdEncoding.EncodeToString(kubenetTemplateHairpinContent)
+	}
 	return base64.StdEncoding.EncodeToString(kubenetTemplateContent)
+}
+
+// needsExplicitHairpinMode checks /etc/os-release to determine if the node is
+// running Ubuntu 24.04+, where kernel 6.8+ requires explicit BR_HAIRPIN_MODE.
+func needsExplicitHairpinMode() bool {
+	data, err := os.ReadFile("/etc/os-release")
+	if err != nil {
+		return false
+	}
+	var id, versionID string
+	for _, line := range strings.Split(string(data), "\n") {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(line, "ID=") {
+			id = strings.Trim(strings.TrimPrefix(line, "ID="), `"`)
+		} else if strings.HasPrefix(line, "VERSION_ID=") {
+			versionID = strings.Trim(strings.TrimPrefix(line, "VERSION_ID="), `"`)
+		}
+	}
+	if strings.ToLower(id) != "ubuntu" {
+		return false
+	}
+	// Compare VERSION_ID >= "24.04" lexicographically (works for Ubuntu version format)
+	return versionID >= "24.04"
 }
 
 // getContainerdConfigBase64 returns the base64 encoded containerd config depending on whether the node is with GPU or not.
--- a/aks-node-controller/parser/templates/kubenet-cni-hairpin.json.gtpl
+++ b/aks-node-controller/parser/templates/kubenet-cni-hairpin.json.gtpl
@@ -1,0 +1,24 @@
+{
+	"cniVersion": "0.3.1",
+	"name": "kubenet",
+	"plugins": [{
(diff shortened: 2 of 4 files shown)


pkg/agent/baker.go
```diff
@@ -1582,6 +1587,38 @@ const kubenetCniTemplate = `
...
+    "plugins": [{
+    "type": "bridge",
+    "bridge": "cbr0",
+    "mtu": 1500,
+    "addIf": "eth0",
+    "isGateway": true,
+    "ipMasq": false,
+    "promiscMode": false,
+    "hairpinMode": true,
+    "ipam": {
```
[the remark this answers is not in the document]
levimm: cannot set hairpin mode and promiscuous mode at the same time.
https://github.com/containernetworking/plugins/blob/main/plugins/main/bridge/bridge.go#L553

aks-node-controller/parser/templates/kubenet-cni-hairpin.json.gtpl
```diff
@@ -0,0 +1,24 @@
...
+	"cniVersion": "0.3.1",
+	"name": "kubenet",
+	"plugins": [{
+		"type": "bridge",
+		"bridge": "cbr0",
+		"mtu": 1500,
+		"addIf": "eth0",
+		"isGateway": true,
+		"ipMasq": false,
+		"promiscMode": false,
```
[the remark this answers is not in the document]
levimm: cannot set hairpin mode and promiscuous mode at the same time.
https://github.com/containernetworking/plugins/blob/main/plugins/main/bridge/bridge.go#L553

pkg/agent/baker.go
```diff
@@ -969,6 +969,11 @@ func getContainerServiceFuncMap(config *datamodel.NodeBootstrappingConfiguration
 			return ""
 		},
 		"GetKubenetTemplate": func() string {
+			// On Ubuntu 24.04+ (kernel 6.8+), promiscuous mode no longer implicitly
+			// enables hairpin forwarding in the bridge. Use explicit hairpinMode instead.
+			if profile.Is2404VHDDistro() {
```
djsly, line 974: we need to explain that this is for all future OS version ? or a specific Kernel versions ? 2404 doesnt tell us the kernel vesions? trying to protect us from the same issue in the future when we introduce 2604, and we start using new HWE kernels.

aks-node-controller/parser/helper.go
```diff
@@ -144,10 +147,38 @@ func getStringifiedStringArray(arr []string, delimiter string) string {
...
+	if needsExplicitHairpinMode() {
+		return base64.StdEncoding.EncodeToString(kubenetTemplateHairpinContent)
+	}
 	return base64.StdEncoding.EncodeToString(kubenetTemplateContent)
 }
 
+// needsExplicitHairpinMode checks /etc/os-release to determine if the node is
+// running Ubuntu 24.04+, where kernel 6.8+ requires explicit BR_HAIRPIN_MODE.
+func needsExplicitHairpinMode() bool {
+	data, err := os.ReadFile("/etc/os-release")
```
djsly, line 162: I'm sure we have this data already in aks-node-controller ?

aks-node-controller/parser/helper.go
```diff
@@ -144,10 +147,38 @@ func getStringifiedStringArray(arr []string, delimiter string) string {
...
 func getKubenetTemplate() string {
+	if needsExplicitHairpinMode() {
+		return base64.StdEncoding.EncodeToString(kubenetTemplateHairpinContent)
+	}
 	return base64.StdEncoding.EncodeToString(kubenetTemplateContent)
 }
 
+// needsExplicitHairpinMode checks /etc/os-release to determine if the node is
+// running Ubuntu 24.04+, where kernel 6.8+ requires explicit BR_HAIRPIN_MODE.
+func needsExplicitHairpinMode() bool {
```
djsly, line 161: @Devinwong  dont we have a beter way already to know which OS we are in ?

aks-node-controller/parser/helper.go
```diff
@@ -144,10 +147,38 @@ func getStringifiedStringArray(arr []string, delimiter string) string {
 }
 
 // getKubenetTemplate returns the base64 encoded Kubenet template.
+// On Ubuntu 24.04+ (kernel 6.8+), promiscuous mode no longer implicitly enables
+// hairpin forwarding in the bridge. Use explicit hairpinMode for those nodes.
 func getKubenetTemplate() string {
+	if needsExplicitHairpinMode() {
```
djsly, line 153: should we maybe look for the kernel as well ? once we move to 6.12, if the issue if fixed in the kernel would we still want to change this ?

PR #8430, opened 2026-04-29
fix: disable scriptless phase2 for subsets of overlapping tests

awesomenix:
Disable few subsets of overlapping GPU tests to save quota


e2e/types.go
```diff
@@ -36,6 +36,7 @@ type Tags struct {
 	VHDCaching             bool
 	MockAzureChinaCloud    bool
 	VMSeriesCoverageTest   bool
+	DisableScriptlessNBC   bool
```
Devinwong: Can we leave a comment to let ourselves know when full scriptless (phase3) is enabled, what toggles/flags we can clean up and which should be left?

lilypan26: nit: could we just repurpose the `SkipScriptlessNBC` variable instead of adding a new tag?

lilypan26 (approved):

Devinwong (approved):

ganeshkumarashok (approved):

PR #8453, opened 2026-05-05
test: (scriptless) Enable scriptless phase 3 in AB e2es

lilypan26:
- Enables scriptless phase 3 in ab e2es under tag `EnableScriptlessANC`
    - This tag means both ANC and NBC cse cmd will be provided, tests should validate that there are no diffs between the generated provisioning env vars
    - For now, NBC cse cmd will be used for provisioning until there are no more diffs, after which we can switch to using ANC
- Adds `CustomDataPhase3` to provide both ANC and NBC cse cmd to AKS node controller

Fixes #


e2e/vmss.go
```diff
@@ -90,26 +90,67 @@ func ConfigureAndCreateVMSS(ctx context.Context, s *Scenario) (*ScenarioVM, erro
...
+
+	// Build provision flags conditionally based on what's provided.
+	var flags []string
+	var encodedNBCCSECmd string
+	if s.Runtime.AKSNodeConfig != nil {
+		flags = append(flags, "--provision-config="+configPath)
+	}
+	if nbcCmdScript != "" {
+		flags = append(flags, "--nbc-cmd="+nbcCmdPath)
+		encodedNBCCSECmd = base64.StdEncoding.EncodeToString([]byte(nbcCmdScript))
```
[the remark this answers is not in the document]
lilypan26: we want to keep it like this until there are no more diffs in the env var, then we can switch to using ANC

e2e/types.go
```diff
@@ -140,6 +140,7 @@ type ScenarioRuntime struct {
 	VM                        *ScenarioVM
 	VMSSName                  string
 	EnableScriptlessNBCCSECmd bool
+	EnableScriptlessANC       bool
```
Devinwong: Does that mean if we move into phase 3 in production, we can safely remove the phase 2 tag `EnableScriptlessNBCCSECmd`? Trying to sort this out here because there will be 2 similar tags.

lilypan26: Yes, honestly wondering if we can remove phase 2 tests now, since phase 3 is still using nbc cse cmd, only difference is it is passed through ANC instead of NBC.

pkg/agent/baker.go
```diff
@@ -1742,15 +1740,14 @@ root = "{{GetDataDir}}"{{- end}}
...
-
 [plugins."io.containerd.cri.v1.runtime".containerd]
   {{- if IsNSeriesSKU }}
   default_runtime_name = "nvidia-container-runtime"
   [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.nvidia-container-runtime]
     runtime_type = "io.containerd.runc.v2"
-    [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.nvidia-container-runtime.options]
-      BinaryName = "/usr/bin/nvidia-container-runtime"
-      SystemdCgroup = true
+  [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.nvidia-container-runtime.options]
```
Devinwong, line 1798: I don't know why there is no icm on this before your change? maybe no one is using containerd.cri.v1?

pkg/agent/baker.go
```diff
@@ -1916,10 +1906,10 @@ root = "{{GetDataDir}}"{{- end}}
     type = "snapshot"
     address = "/run/containerd/tardev-snapshotter.sock"
 [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.kata-cc]
-	pod_annotations = ["io.katacontainers.*"]
   snapshotter = "tardev"
   runtime_type = "io.containerd.kata-cc.v2"
   privileged_without_host_devices = true
+  pod_annotations = ["io.katacontainers.*"]
```
Devinwong, line 1962: Many indents error are related to Kata. I am wondering if kata is never in production or these configs are never used.

e2e/node_config.go
```diff
@@ -81,19 +83,16 @@ func baseKubeletConfig() *aksnodeconfigv1.KubeletConfig {
 				"nodefs.inodesFree": "5%",
 			},
 			ProtectKernelDefaults: true,
-			FeatureGates:          map[string]bool{},
-			FailSwapOn:            to.Ptr(false),
+			FeatureGates: map[string]bool{
+				"RotateKubeletServerCertificate": true,
```
Devinwong, line 87: We already have this set to true in `KubeletFlags["--feature-gates"`. 
If we've set `EnableKubeletConfigFile: false,`, do we need this?

e2e/types.go
Devinwong: > Yes, honestly wondering if we can remove phase 2 tests now, since phase 3 is still using nbc cse cmd, only difference is it is passed through ANC instead of NBC.

I think we can. @awesomenix what do you think?

pkg/agent/baker.go
```diff
@@ -1481,10 +1481,7 @@ func isMariner(osSku string) bool {
 	return osSku == datamodel.OSSKUCBLMariner || osSku == datamodel.OSSKUMariner || osSku == datamodel.OSSKUAzureLinux
 }
 
-const sysctlTemplateString = `# This is a partial workaround to this upstream Kubernetes issue:
-# https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731
```
Devinwong, line 1487: No one dared to remove this although it's 9 years ago. 😂
Indeed, do you know if the issue no longer exists?

e2e/scenario_test.go
```diff
@@ -232,6 +253,9 @@ func Test_ACL_ARM64(t *testing.T) {
 				nbc.AgentPoolProfile.VMSize = "Standard_D2pds_v6"
 				nbc.IsARM64 = true
 			},
+			AKSNodeConfigMutator: func(_ *Cluster, config *aksnodeconfigv1.Configuration) {
+				config.VmSize = "Standard_D2pds_v6"
```
Devinwong: Why the vmsize needs explicit set for some cases only?

aks-node-controller/parser/parser.go
```diff
@@ -194,7 +194,8 @@ func getCSEEnv(config *aksnodeconfigv1.Configuration) map[string]string {
 		"SERVICE_ACCOUNT_IMAGE_PULL_DEFAULT_TENANT_ID": config.GetServiceAccountImagePullProfile().GetDefaultTenantId(),
 		"IDENTITY_BINDINGS_LOCAL_AUTHORITY_SNI":        config.GetServiceAccountImagePullProfile().GetLocalAuthoritySni(),
 		"CSE_TIMEOUT":                                  getCSETimeout(config),
-		"SKIP_WAAGENT_HOLD":                            "true",
+		"SKIP_WAAGENT_HOLD":                            "false",
```
awesomenix: this needs to be true, because scriptless mode we always never hold waagent

awesomenix: Yep this would be a difference compared to non scriptless path, may be we can have a list of keys which are ok to ignore

e2e/types.go
awesomenix: Yeah you can remove phase2. I am wondering what latency does diff add to the provisioning. Can we perform the diff after we have provisioned so we wont affect the provisioning path?

awesomenix: Do we need a new flag. Because RP needs to provide aksnodeconfig input somewhere which means we are changing the signature of GetBootstrappingData(). We will probably be switching over to aks-node-controller.utils.CustomData instead based on some feature flag.

pkg/agent/baker.go
lilypan26: im just fixing the indentation here, i dont think it actually changes the functionality

pkg/agent/baker.go
lilypan26: same here, the indentations are just for readablity, but fixing here to make it match ANC template

pkg/agent/baker.go
lilypan26: yea i checked the issue and im still not sure whether it is fixed or not but regardless im just removing the comment here because it was causing diff with the anc content. the actual sysctl content should be the same, just reordered some stuff because the anc config is sorted

e2e/types.go
lilypan26: this is just e2e flag to distinguish between scriptless phases. No need to add NBC flag because like you said we will be using anc.CustomData()

e2e/scenario_test.go
lilypan26: In any scenario where we set vmsize in NBC, we also set it in ANC to match. When we switch over to use ANC instead of NBC we will need it so the scenario uses the right vm size.

e2e/node_config.go
lilypan26: yes we set it in NBC, but we need to set it in ANC so it matches when converted to kubelet flags

e2e/types.go
lilypan26: Moved the diff to after provisioning. Will make phase 2 cleanup a separate PR since this one is already getting big

pkg/agent/baker.go
Devinwong: I see. It's written into a .conf file, not a yaml file. Otherwise indent would matter.

Devinwong (approved):

djsly:
**AgentBaker Linux PR gate — E2E mass-failure (change-caused, HIGH confidence)**

- Run: [167125948](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=167125948) (`failed`)
- Failed task: `Run AgentBaker E2E` → AzureCLI → exit 1
- Net new failures: **~13 leaves**, all under `Test_LocalDNSHostsPlugin*`.

**Two failure shapes, same root cause:**
1. **Loud / fast (1 leaf):** `Test_LocalDNSHostsPlugin_Scriptless/ACL/default` (155.83s) → `e2e/validation.go:52` 🔴 `expected no env var differences between provision-config and nbc-cmd, but found differences:` — the new `compareEnvs` validator this PR added fires with a non-empty diff list.
2. **Slow / derivative (7 leaves):** `Test_LocalDNSHostsPlugin{,_Scriptless}/{Ubuntu2204,Ubuntu2404,AzureLinuxV3,ACL}/{default,scriptless_nbc}` → `kube.go:189` after ~742s 🔴 `"<pod>" haven't appeared in k8s API server: context deadline exceeded`. Identical signature across every distro × variant → systemic node-bring-up degradation, not infra.

**Likely root cause (`(b)` + `(c)` confirmed, `(a)` ruled out):**
- This PR flips `nbc.EnableScriptlessNBCCSECmd = true` for every non-`Scriptless` scenario in `e2e/node_config.go`, and `e2e/scenario_test.go` strips `Tags: { Scriptless: true }` from each `_Scriptless` test and adds a `BootstrapConfigMutator` — so **every LocalDNS scenario now provisions via the new scriptless NBC-CSE-Cmd path, including the `default` (bash-CSE) variants**.
- `parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh` was changed from `if/elif` to two independent `if`s, so when both files exist `--provision-config` and `--nbc-cmd` are both passed to `aks-node-controller provision`, which then runs `compareEnvs`.
- `e2e/node_config.go::nbcToAKSNodeConfigV1` has **no `LocalDNSProfile` mapping** — every `LocalDNS` occurrence in the PR diff is the unrelated `GenerateLocalDNSCoreFile` template. The localdns systemd unit therefore never gets configured/started on nodes provisioned via the converter, DNS on the node breaks, debug pods can't reach the API server → the `kube.go:189` timeout.
- The 2 cosmetic `localdns.toml.gtpl` diff lines (blank-line removal + trailing newline) are not the cause.

**Build-vs-test:** test-code-caused (e2e converter + scenario tagging) with a tightly-coupled product-code enabler (`baker.go` scriptless plumbing + `aks-node-controller-wrapper.sh`). Not a `main` regression.

**Confidence:** HIGH — three independent indicators converge: PR's own validator firing loudly on ACL/default, the `EnableScriptlessNBCCSECmd` flip mechanism, and the perfectly-symmetric distro×variant timeout pattern.

**Strongest alternative (less likely):** `sysctlTemplateString` rewrite in `baker.go` degrading DNS/conntrack reliability — refuted because every branch still sets `tcp_retries2=8`, LocalDNS scenarios don't customize sysctls, and a sysctl change wouldn't produce a 100% deterministic per-distro pattern lined up with files this PR touched.

**Side-channel (not the cause, FYI):** `build2204arm64gen2containerd` and `build2404gen2containerd` flagged `succeededWithIssues` with `CIS regressions detected (1)` — same 6.1.4.1 pattern recurring across renovate PRs the last 48h (#8652, #8294); non-gating here.

**Recommended next action (owner: @lilypan26):**
1. Look at the ACL/default `validation.go:52` env-var diff list first — it names the exact env keys the converter mis-maps.
2. Audit `e2e/node_config.go::nbcToAKSNodeConfigV1` for missing NBC field coverage — start with `LocalDNSProfile`, then any field that drives a systemd unit / cloud-init artifact. Use `pkg/agent/baker.go` rendering and `aks-node-controller/parser` as the reference.
3. Decide if enabling `EnableScriptlessNBCCSECmd = true` for `default` variants of every scenario is intended in Phase 3 — if yes, converter coverage must be exhaustive before merge; if not, scope the flip to opted-in scenarios.
4. Re-run the LocalDNS slice (Ubuntu2204/2404 + AzureLinuxV3 + ACL × {default, scriptless_nbc}) before re-requesting review.

_Posted by Clawpilot AgentBaker gate detective._

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
```diff
@@ -42,13 +42,12 @@ command=("$BIN_PATH" provision)
...
     command+=("--provision-config=$CONFIG_PATH")
-elif [ -f "$NBC_CMD_PATH" ]; then
+fi
+if [ -f "$NBC_CMD_PATH" ]; then
     log "Launching aks-node-controller with nbc cmd ${NBC_CMD_PATH}"
     command+=("--nbc-cmd=$NBC_CMD_PATH")
-else
-    log "Gracefully exit aks-node-controller without provision config or nbc cmd"
-    exit 0
 fi
```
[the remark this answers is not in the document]
lilypan26: the controller return appropriate error when neither flags are passed

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/Azure/agentbaker
 go 1.25.10
 
 require (
+	github.com/Azure/agentbaker/aks-node-controller v0.0.0-20241215075802-f13a779d5362
```
awesomenix: oh! this version wont exist when we import this to RP isnt it? because we import v20260527.0 types

awesomenix (approved):

PR #8466, opened 2026-05-07
Add echo statements for versioning information

chaitanya51906312:

Fixes #


chaitanya51906312:
this is only a test,so good to close

PR #8468, opened 2026-05-07
fix: always add aks custom cloud until we do better

awesomenix:
There is no easy way to implement this as of now. We will wait for better solution with rcv1p work?

### DIFF
--- a/parts/linux/cloud-init/nodecustomdata.yml
+++ b/parts/linux/cloud-init/nodecustomdata.yml
@@ -25,6 +25,15 @@ write_files:
     Any overridden files will be listed here - Hotfix mode
     Example: {{GetCSEHelpersScriptFilepath}}
 
+{{if IsAKSCustomCloud}}
+- path: {{GetInitAKSCustomCloudFilepath}}
+  permissions: "0744"
+  encoding: gzip
+  owner: root
+  content: !!binary |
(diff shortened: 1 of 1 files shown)


parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -25,6 +25,13 @@ write_files:
     Any overridden files will be listed here - Hotfix mode
     Example: {{GetCSEHelpersScriptFilepath}}
 
+- path: {{GetInitAKSCustomCloudFilepath}}
+  permissions: "0744"
+  encoding: gzip
+  owner: root
+  content: !!binary |
+    {{GetVariableProperty "cloudInitData" "initAKSCustomCloud"}}
+
```
Devinwong, line 36: should we add a condition
```
 {{if IsAKSCustomCloud}}
 - path: {{GetInitAKSCustomCloudFilepath}}
   permissions: "0744"
   encoding: gzip
   owner: root
   content: !!binary |
     {{GetVariableProperty "cloudInitData" "initAKSCustomCloud"}}
 {{end}}
```
to avoid writing an empty file for public cloud because `cloudInitData["initAKSCustomCloud"] is empty`

Devinwong (approved):

cameronmeissner (approved):

PR #8470, opened 2026-05-07
feat(e2e): add HTTPS_PROXY + private DNS test scenario

r2k1:
## Summary

Add E2E test for node bootstrapping with HTTPS_PROXY configured and private DNS zone for the API server FQDN. Regression coverage for IcM 603699115 / ADO#31707996.

### New test: `Test_Ubuntu2204_HTTPSProxy_PrivateDNS`
Validates that a node bootstraps successfully when `HTTPS_PROXY` is set and the API server FQDN resolves via a private DNS zone.

### Proxy infrastructure (all non-isolated clusters)
- Python-based HTTP CONNECT proxy deployed as DaemonSet on system pool (`mcr.microsoft.com/cbl-mariner/base/python:3`)
- Private DNS zone for API server FQDN linked to cluster VNet
- Both run in `prepareCluster` DAG, consistent across clusters

### Signature refactor
- `BootstrapConfigMutator` and `AKSNodeConfigMutator` now accept `*Cluster` as first parameter
- Enables scenarios to access cluster properties (e.g., `ProxyURL`)

### Bug fix
- `createNewAKSClusterWithRetry` now retries on managed identity reconciliation `NotFound` errors after cluster deletion

## Test Result
```
--- PASS: Test_Ubuntu2204_HTTPSProxy_PrivateDNS/default (251.99s)
PASS
```

Fixes: ADO#31707996

### DIFF
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"net/netip"
 	"strings"
@@ -21,6 +22,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v8"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v7"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/privatedns/armprivatedns"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources/v3"
 	"github.com/google/uuid"
 	corev1 "k8s.io/api/core/v1"
@@ -42,6 +44,7 @@ type Cluster struct {
 	SubnetID        string
 	ClusterParams   *ClusterParams
 	Bastion         *Bastion
+	ProxyURL        string
 }
 
 // Returns true if the cluster is configured with Azure CNI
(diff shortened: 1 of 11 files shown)


e2e/cluster.go
```diff
@@ -107,7 +110,21 @@ func prepareCluster(ctx context.Context, clusterModel *armcontainerservice.Manag
...
+	proxyURL := dag.Go1(g, kube, func(ctx context.Context, k *Kubeclient) (string, error) {
+		if err := k.EnsureDebugDaemonsets(ctx, isNetworkIsolated, config.GetPrivateACRName(true, *cluster.Location)); err != nil {
+			return "", err
+		}
+		if isNetworkIsolated {
+			return "", nil
+		}
+		return k.GetProxyURL(ctx)
+	}, debugDeps...)
+	if !isNetworkIsolated {
```
timmy-wright, line 123: This surprises me. Probably due to me not quite understanding things. I thought for network isolated clusters, we would need a private DNS for the API server setup. But this code only sets up the private DNS for non-network isolated clusters.

timmy-wright (approved):

r2k1:
Re: timmy-wright's question about private DNS only for non-isolated clusters:

Network-isolated clusters already have their own private DNS setup via `addPrivateEndpointForACR` which creates a `privatelink.azurecr.io` zone. The private DNS zone we add here is for the **API server FQDN** — it mirrors public DNS resolution so that nodes with HTTP_PROXY configured can resolve the API server via the system resolver (nslookup) rather than through the proxy.

Network-isolated clusters block outbound entirely and use a different connectivity model (private endpoints + NSG rules for specific IPs), so the proxy + private DNS pattern doesn't apply to them.

e2e/cluster.go
r2k1: Network-isolated clusters block all outbound and use private endpoints with NSG allow-rules for specific IPs. They don't use HTTP_PROXY — the proxy pattern is for clusters that have internet access through a forward proxy. The private DNS zone here mirrors public DNS for the API server FQDN, which only matters when nodes use a system resolver + proxy combination.

PR #8475, opened 2026-05-08
fix: blacklist rxrpc/esp4/esp6 modules to mitigate DirtyFrag LPE

djsly:
## Summary

Blacklist rxrpc, esp4, and esp6 kernel modules to mitigate the DirtyFrag universal Linux LPE. The rxrpc path bypasses AppArmor `restrict_unprivileged_userns` — confirmed exploitable on Ubuntu 24.04 AKS nodes (kernel 6.8.0-1052-azure).

## Evidence

Exploit tested live on AKS node: unprivileged user (uid=1001) overwrote `/etc/passwd` root line via rxrpc.ko page-cache write, removing shadow reference (`root:x:` → `root::`). No user namespaces needed.

## Changes

- `modprobe-CIS.conf`: add `install/blacklist` rules for esp4, esp6, rxrpc (same pattern as algif_aead for CVE-2026-31431)
- `cse_main.sh`: add CSE hotfix for existing VHDs that don't have the new modprobe-CIS.conf baked in
- `linux-vhd-content-test.sh`: add `testDirtyFragModulesDisabled` VHD content test

## Testing

- Verified DirtyFrag exploit fails when modules are blacklisted
- VHD content test validates modprobe config and module state
- Pattern matches existing algif_aead mitigation (proven in production)

## References

- https://github.com/V4bel/dirtyfrag
- https://github.com/0xdeadbeefnetwork/Copy_Fail2-Electric_Boogaloo

## Related

- AB#37878493

### DIFF
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -2781,61 +2781,21 @@ func ValidateCollectWindowsLogsScript(ctx context.Context, s *Scenario) {
 		"collect-windows-logs.ps1 failed or did not produce a zip file")
 }
 
-// ValidateAlgifAeadMitigation verifies CVE-2026-31431 mitigation is active:
-// the algif_aead kernel module must be blocked via modprobe config, not loaded,
-// and modprobe must refuse to load it.
-func ValidateAlgifAeadMitigation(ctx context.Context, s *Scenario) {
-	s.T.Helper()
-
-	// Flatcar uses a different kernel module setup
-	if s.VHD.Flatcar {
-		s.T.Log("Skipping algif_aead validation: not applicable for Flatcar")
-		return
-	}
-
-	script := strings.Join([]string{
-		`# Check modprobe config blocks the module`,
-		`if ! grep -qs 'install algif_aead /bin/false' /etc/modprobe.d/*.conf 2>/dev/null; then`,
-		`  echo "FAIL: algif_aead disable rule not found in /etc/modprobe.d/*.conf"`,
-		`  exit 1`,
-		`fi`,
-		`echo "PASS: modprobe config blocks algif_aead"`,
-		``,
-		`# Check module is not loaded`,
-		`if grep -qE '^algif_aead ' /proc/modules 2>/dev/null; then`,
-		`  echo "FAIL: algif_aead module is loaded"`,
-		`  exit 1`,
-		`fi`,
-		`echo "PASS: algif_aead module is not loaded"`,
-		``,
-		`# Check modprobe refuses to load it`,
-		`if sudo modprobe algif_aead 2>/dev/null; then`,
-		`  echo "FAIL: modprobe algif_aead succeeded, should be blocked"`,
-		`  sudo rmmod algif_aead 2>/dev/null || true`,
-		`  exit 1`,
-		`fi`,
-		`echo "PASS: modprobe algif_aead correctly refused"`,
-	}, "\n")
-
-	execScriptOnVMForScenarioValidateExitCode(ctx, s, script, 0,
-		"CVE-2026-31431 (algif_aead) mitigation validation failed")
-}
-
-// ValidateDirtyFragMitigation verifies DirtyFrag LPE mitigation is active:
-// the esp4, esp6, and rxrpc kernel modules must be blocked via modprobe config,
-// not loaded, and modprobe must refuse to load them. The rxrpc path bypasses
-// AppArmor userns restrictions and was confirmed exploitable on Ubuntu 24.04.
-func ValidateDirtyFragMitigation(ctx context.Context, s *Scenario) {
+// ValidateVulnerableKernelModulesDisabled verifies that kernel modules with known
+// LPE vulnerabilities are blocked via modprobe config, not loaded, and cannot be loaded.
+// Covers: CVE-2026-31431 (algif_aead), DirtyFrag (esp4, esp6, rxrpc).
+// To add a new CVE mitigation, append the module name to the list below.
+func ValidateVulnerableKernelModulesDisabled(ctx context.Context, s *Scenario) {
 	s.T.Helper()
 
 	if s.VHD.Flatcar {
-		s.T.Log("Skipping DirtyFrag validation: not applicable for Flatcar")
+		s.T.Log("Skipping vulnerable kernel module validation: not applicable for Flatcar")
 		return
 	}
 
 	script := strings.Join([]string{
 		`failed=0`,
-		`for mod in esp4 esp6 rxrpc; do`,
+		`for mod in algif_aead esp4 esp6 rxrpc; do`,
 		`  if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then`,
 		`    echo "FAIL: ${mod} disable rule not found in /etc/modprobe.d/*.conf"`,
 		`    failed=1`,
@@ -2860,5 +2820,5 @@ func ValidateDirtyFragMitigation(ctx context.Context, s *Scenario) {
 	}, "\n")
 
 	execScriptOnVMForScenarioValidateExitCode(ctx, s, script, 0,
-		"DirtyFrag (esp4/esp6/rxrpc) mitigation validation failed")
+		"Vulnerable kernel module mitigation validation failed (algif_aead/esp4/esp6/rxrpc)")
 }
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -2796,7 +2796,7 @@ func ValidateVulnerableKernelModulesDisabled(ctx context.Context, s *Scenario) {
 	script := strings.Join([]string{
 		`failed=0`,
 		`for mod in algif_aead esp4 esp6 rxrpc; do`,
-		`  if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then`,
+		`  if ! grep -qsE "^install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then`,
 		`    echo "FAIL: ${mod} disable rule not found in /etc/modprobe.d/*.conf"`,
 		`    failed=1`,
 		`  else`,
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -2820,3 +2820,45 @@ func ValidateAlgifAeadMitigation(ctx context.Context, s *Scenario) {
 	execScriptOnVMForScenarioValidateExitCode(ctx, s, script, 0,
 		"CVE-2026-31431 (algif_aead) mitigation validation failed")
 }
+
+// ValidateDirtyFragMitigation verifies DirtyFrag LPE mitigation is active:
+// the esp4, esp6, and rxrpc kernel modules must be blocked via modprobe config,
+// not loaded, and modprobe must refuse to load them. The rxrpc path bypasses
+// AppArmor userns restrictions and was confirmed exploitable on Ubuntu 24.04.
+func ValidateDirtyFragMitigation(ctx context.Context, s *Scenario) {
+	s.T.Helper()
+
+	if s.VHD.Flatcar {
+		s.T.Log("Skipping DirtyFrag validation: not applicable for Flatcar")
+		return
+	}
+
+	script := strings.Join([]string{
+		`failed=0`,
+		`for mod in esp4 esp6 rxrpc; do`,
+		`  if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then`,
+		`    echo "FAIL: ${mod} disable rule not found in /etc/modprobe.d/*.conf"`,
+		`    failed=1`,
+		`  else`,
+		`    echo "PASS: modprobe config blocks ${mod}"`,
+		`  fi`,
+		`  if grep -qE "^${mod} " /proc/modules 2>/dev/null; then`,
+		`    echo "FAIL: ${mod} module is loaded"`,
+		`    failed=1`,
+		`  else`,
+		`    echo "PASS: ${mod} module is not loaded"`,
+		`  fi`,
+		`  if sudo modprobe "${mod}" 2>/dev/null; then`,
+		`    echo "FAIL: modprobe ${mod} succeeded, should be blocked"`,
+		`    sudo modprobe -r "${mod}" 2>/dev/null || true`,
+		`    failed=1`,
+		`  else`,
+		`    echo "PASS: modprobe ${mod} correctly refused"`,
+		`  fi`,
+		`done`,
+		`exit $failed`,
+	}, "\n")
+
+	execScriptOnVMForScenarioValidateExitCode(ctx, s, script, 0,
+		"DirtyFrag (esp4/esp6/rxrpc) mitigation validation failed")
+}
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -861,53 +861,6 @@ configureNodeExporter() {
     echo "Node Exporter started successfully"
 }
 
-# Disable kernel modules with known LPE vulnerabilities.
-# Writes modprobe blacklist rules and unloads any already-loaded modules.
-# Applies to existing VHDs that don't yet have the fix baked into modprobe-CIS.conf.
-# Safe to run unconditionally — idempotent (overwrites with same content if already present).
-#
-# To add a new CVE mitigation, append to the MODULES array below.
-disableVulnerableKernelModules() {
-    # Each entry: "module_name:config_file:description"
-    local MODULES=(
-        "algif_aead:disable-algif_aead.conf:CVE-2026-31431 (Copy Fail)"
-        "esp4:disable-dirtyfrag.conf:DirtyFrag (xfrm-ESP page-cache write)"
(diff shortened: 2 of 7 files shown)


awesomenix (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -311,6 +311,25 @@ EOF
         fi
     fi
 
+    # DirtyFrag: Mitigate xfrm-ESP + RxRPC page-cache write LPE vulnerabilities.
+    # rxrpc path bypasses AppArmor userns restrictions — does NOT require user namespaces.
+    # Confirmed exploitable on Ubuntu 24.04 (kernel 6.8). No upstream patch exists yet.
+    # Applies to existing VHDs that don't yet have the modprobe-CIS.conf fix baked in.
+    if [ "$OS" = "$UBUNTU_OS_NAME" ] || isMarinerOrAzureLinux "$OS"; then
+        if ! grep -qs "install rxrpc /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — Now iterates over all three modules individually and writes the config file if **any** rule is missing. See b041fd3.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -311,6 +311,25 @@ EOF
...
+            printf "install esp4 /bin/false\nblacklist esp4\ninstall esp6 /bin/false\nblacklist esp6\ninstall rxrpc /bin/false\nblacklist rxrpc\n" > /etc/modprobe.d/disable-dirtyfrag.conf
+        fi
+        for mod in rxrpc esp4 esp6; do
+            if grep -q "^${mod} " /proc/modules 2>/dev/null; then
+                if rmmod "$mod" 2>/dev/null; then
+                    echo "DirtyFrag: successfully unloaded ${mod} module"
+                else
+                    echo "DirtyFrag: failed to unload ${mod} (in use), reboot required for full mitigation"
+                fi
+            fi
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — Switched from `rmmod` to `modprobe -r` for safer dependency-aware unloading. See b041fd3.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1294,6 +1294,37 @@ testAlgifAeadDisabled() {
...
+      echo "$test: modprobe config correctly blocks ${mod}"
+    fi
+
+    if grep -qE "^${mod} " /proc/modules 2>/dev/null; then
+      err "$test" "${mod} kernel module is loaded despite being disabled"
+      failed=1
+    else
+      echo "$test: ${mod} module is not loaded"
+    fi
+  done
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — Added `modprobe ${mod}` negative test in the VHD test loop (verify load is blocked, not just config exists). Also added a full E2E validator (`ValidateDirtyFragMitigation`) in `e2e/validators.go` registered alongside `ValidateAlgifAeadMitigation`. See b041fd3.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -311,6 +311,32 @@ EOF
         fi
     fi
 
+    # DirtyFrag: Mitigate xfrm-ESP + RxRPC page-cache write LPE vulnerabilities.
+    # rxrpc path bypasses AppArmor userns restrictions — does NOT require user namespaces.
+    # Confirmed exploitable on Ubuntu 24.04 (kernel 6.8). No upstream patch exists yet.
+    # Applies to existing VHDs that don't yet have the modprobe-CIS.conf fix baked in.
+    if [ "$OS" = "$UBUNTU_OS_NAME" ] || isMarinerOrAzureLinux "$OS"; then
```
timmy-wright: It's odd that the two parts of this test are done in different ways. Do we have a "isUbuntu" function like the isMarinerOrAzureLinux function?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -311,6 +311,32 @@ EOF
...
+        local dirtyfrag_needs_update=false
+        for mod in esp4 esp6 rxrpc; do
+            if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then
+                dirtyfrag_needs_update=true
+                break
+            fi
+        done
+        if [ "$dirtyfrag_needs_update" = "true" ]; then
+            printf "install esp4 /bin/false\nblacklist esp4\ninstall esp6 /bin/false\nblacklist esp6\ninstall rxrpc /bin/false\nblacklist rxrpc\n" > /etc/modprobe.d/disable-dirtyfrag.conf
```
timmy-wright: Seems this would be easier to read and modify using a HEREDOC rather than a printf.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -311,6 +311,32 @@ EOF
...
 
+    # DirtyFrag: Mitigate xfrm-ESP + RxRPC page-cache write LPE vulnerabilities.
+    # rxrpc path bypasses AppArmor userns restrictions — does NOT require user namespaces.
+    # Confirmed exploitable on Ubuntu 24.04 (kernel 6.8). No upstream patch exists yet.
+    # Applies to existing VHDs that don't yet have the modprobe-CIS.conf fix baked in.
+    if [ "$OS" = "$UBUNTU_OS_NAME" ] || isMarinerOrAzureLinux "$OS"; then
+        local dirtyfrag_needs_update=false
+        for mod in esp4 esp6 rxrpc; do
+            if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then
+                dirtyfrag_needs_update=true
```
timmy-wright: Why do we need this initial check? Is there any problem with always creating the dirtyfrag file?

Also, there's a logic error in the check:
* The check looks to see if any of the modules are missing
* If any are missing, then we block all of them

If the check is important, then shouldn't we just block the missing ones?

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: Good catch — switched to `isUbuntu "$OS"` for consistency with the rest of the codebase. See 8667af0.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: Agreed — replaced printf with a heredoc. Much cleaner. See 8667af0.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: You're right, no reason for the check — the file is small and writing it unconditionally is simpler and eliminates the partial-config edge case entirely. Removed the loop/check, now always writes the file (idempotent). See 8667af0.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -311,6 +311,31 @@ EOF
         fi
     fi
 
+    # DirtyFrag: Mitigate xfrm-ESP + RxRPC page-cache write LPE vulnerabilities.
+    # rxrpc path bypasses AppArmor userns restrictions — does NOT require user namespaces.
+    # Confirmed exploitable on Ubuntu 24.04 (kernel 6.8). No upstream patch exists yet.
+    # Applies to existing VHDs that don't yet have the modprobe-CIS.conf fix baked in.
+    # Safe to run unconditionally — idempotent, overwrites with same content if already present.
+    if isUbuntu "$OS" || isMarinerOrAzureLinux "$OS"; then
+        cat > /etc/modprobe.d/disable-dirtyfrag.conf <<EOF
```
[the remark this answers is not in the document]
djsly: @hbeberman do we need this mitigation for ACL ?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -861,6 +861,53 @@ configureNodeExporter() {
     echo "Node Exporter started successfully"
 }
 
+# Disable kernel modules with known LPE vulnerabilities.
```
awesomenix: if you move all this to cse_main.sh, your hotfix story will be easier

parts/linux/cloud-init/artifacts/cse_main.sh
awesomenix, line 52: shouldnt that function be here so it can be called below in right place. The location of function seems odd

djsly: fixed

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -294,21 +294,53 @@ EOF
...
-            printf "install algif_aead /bin/false\nblacklist algif_aead\n" > /etc/modprobe.d/disable-algif_aead.conf
-        fi
-        if grep -q '^algif_aead ' /proc/modules 2>/dev/null; then
-            if rmmod algif_aead 2>/dev/null; then
-                echo "CVE-2026-31431: successfully unloaded algif_aead module"
-            else
-                echo "CVE-2026-31431: failed to unload algif_aead (in use), reboot required for full mitigation"
+    # Defined inline in cse_main.sh (not sourced from cse_config.sh) to support scriptless provisioning.
+    # To add a new CVE mitigation, append to the MODULES array below.
+    disableVulnerableKernelModules() {
```
awesomenix: please add a test in _spec.sh so we wont accidently break this in future, also makes it easier for reviewing on how this is used.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -294,21 +294,53 @@ EOF
...
+        # Unload any already-loaded vulnerable modules
+        for entry in "${MODULES[@]}"; do
+            local mod="${entry%%:*}"
+            local desc="${entry#*:}"
+            desc="${desc#*:}"
+            if grep -q "^${mod} " /proc/modules 2>/dev/null; then
+                if modprobe -r "$mod" 2>/dev/null; then
+                    echo "${desc}: successfully unloaded ${mod}"
+                else
+                    echo "${desc}: failed to unload ${mod} (in use), reboot required for full mitigation"
```
awesomenix: would be nice to have a guestagent log here so we know which/how many customers faced this path

parts/linux/cloud-init/artifacts/cse_config.sh
djsly: Done ✅ — Moved `disableVulnerableKernelModules()` to top-level in `cse_main.sh` (line 60, before `basePrep`). No longer nested inside a function. Also added ShellSpec test (`cse_main_disable_modules_spec.sh`). See 90589fb.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -50,6 +50,54 @@ get_ubuntu_release() {
...
 }
 
+# Disable kernel modules with known LPE vulnerabilities.
+# Writes modprobe blacklist rules and unloads any already-loaded modules.
+# Applies to existing VHDs that don't yet have the fix baked into modprobe-CIS.conf.
+# Safe to run unconditionally — idempotent (overwrites with same content if already present).
+# Defined in cse_main.sh (not sourced) to support scriptless provisioning.
+#
+# To add a new CVE mitigation, append to the MODULES array below.
+disableVulnerableKernelModules() {
```
awesomenix: very stupid question, i had to read it again, why cant this be simple

```
disableVulnerableKernelModule() {
    local mod=$1; shift
    local desc=$1; shift

    printf "# $desc\ninstall $mod /bin/false\nblacklist $mod\n" > /etc/modprobe.d/disable-$mod.conf

     if grep -q "^${mod} " /proc/modules 2>/dev/null; then
          if modprobe -r "$mod" 2>/dev/null; then
              echo "${desc}: successfully unloaded ${mod}"
          else
              echo "${desc}: failed to unload ${mod} (in use), reboot required for full mitigation"
          fi
      fi
}

disableVulnerableKernelModule "algif_aead" "CVE-2026-31431 (Copy Fail)"
disableVulnerableKernelModule "esp4" "DirtyFrag (xfrm-ESP page-cache write)"
disableVulnerableKernelModule "esp6" "DirtyFrag (xfrm-ESP6 page-cache write)"
disableVulnerableKernelModule "rxrpc" "DirtyFrag (RxRPC page-cache write, bypasses AppArmor userns)"
```

djsly: Love it — adopted exactly. `disableVulnerableKernelModule()` now takes `(mod, desc)` and the call site reads like a manifest. No more associative arrays or colon parsing. See 6eaef92.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: Done ✅ — function is now at top-level (line 60), called from inside `basePrep`. See 6eaef92.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: Added `cse_main_disable_modules_spec.sh` with 5 test cases: config file creation, content validation, idempotency, loaded module unload, and no-op when not loaded. See 6eaef92.

parts/linux/cloud-init/artifacts/cse_main.sh
djsly: Good point — the function already echoes success/failure messages which get captured by `logs_to_events`. These show up in the GA extension logs. We could add explicit `logs_to_events` wrapping per-module for more structured telemetry — will consider in a follow-up.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1262,35 +1262,41 @@ testNfsServerService() {
...
 
-  # Verify modprobe config blocks the module
-  if ! grep -qs "install algif_aead /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then
-    err "$test" "algif_aead disable rule not found in /etc/modprobe.d/*.conf"
-    return 1
-  fi
-  echo "$test: modprobe config correctly blocks algif_aead"
+  local failed=0
+  for mod in algif_aead esp4 esp6 rxrpc; do
+    if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — grep now uses `^install` anchor. See 6eaef92.

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1262,35 +1262,41 @@ testNfsServerService() {
...
+    if grep -qE "^${mod} " /proc/modules 2>/dev/null; then
+      err "$test" "${mod} kernel module is loaded despite being disabled"
+      failed=1
+    else
+      echo "$test: ${mod} module is not loaded"
+    fi
+
+    if modprobe "${mod}" 2>/dev/null; then
+      err "$test" "modprobe ${mod} succeeded — module should be blocked"
+      rmmod "${mod}" 2>/dev/null || true
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — switched to `modprobe -r`. See 6eaef92.

e2e/validators.go
```diff
@@ -2781,42 +2781,44 @@ func ValidateCollectWindowsLogsScript(ctx context.Context, s *Scenario) {
...
-		`fi`,
-		`echo "PASS: modprobe algif_aead correctly refused"`,
+		`failed=0`,
+		`for mod in algif_aead esp4 esp6 rxrpc; do`,
+		`  if ! grep -qs "install ${mod} /bin/false" /etc/modprobe.d/*.conf 2>/dev/null; then`,
+		`    echo "FAIL: ${mod} disable rule not found in /etc/modprobe.d/*.conf"`,
+		`    failed=1`,
+		`  else`,
+		`    echo "PASS: modprobe config blocks ${mod}"`,
+		`  fi`,
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — E2E validator now uses `^install` anchor. See 6eaef92.

spec/parts/linux/cloud-init/artifacts/cse_main_disable_modules_spec.sh
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    BeforeEach 'setup'
+    AfterEach 'cleanup'
+
+    # Mock grep for /proc/modules to simulate no modules loaded
+    grep() {
+        if [ "$2" = "/proc/modules" ]; then
+            return 1
+        fi
```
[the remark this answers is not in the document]
djsly: Fixed ✅ — Rewrote spec to use a temp file for `/proc/modules` via sed substitution instead of a broken grep mock. See 6eaef92.

awesomenix (approved):

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
cameronmeissner: maybe im blind, I don't see that we wrapped the calls to `disableVulnerableKernelModule` with `logs_to_events`?

PR #8479, opened 2026-05-08
perf(e2e): skip infrastructure DAG for existing clusters

r2k1:
## Summary

Skip heavy infrastructure setup (firewall, ACR, daemonsets, DNS, maintenance) when `prepareCluster` finds an existing cluster. Only run lightweight tasks needed for every test (kube client, bastion, subnet, identity, garbage collection, proxy URL resolution).

## Problem

`prepareCluster` runs the full infrastructure DAG on every test run, even when the cluster and all its infrastructure already exist from a previous run. This causes:
- **409 Conflict errors** when concurrent ADO pipeline runs create the same resources simultaneously
- **Unnecessary latency** — firewall, ACR, DNS zone creation adds minutes to cluster setup
- **Wasted Azure API calls** for idempotent operations that are already complete

## Solution

`getOrCreateCluster` now returns a `isNewCluster` bool. When false, `prepareCluster` skips:
- Maintenance configuration
- Firewall rules / network-isolated NSG
- ACR setup
- Debug daemonsets + proxy
- Private DNS zone

It still runs:
- Bastion (getOrCreate, fast for existing)
- Kube client, subnet ID, kubelet identity
- Cluster params extraction
- VMSS garbage collection
- Proxy URL resolution (reads existing pod)

## Risk

Low — all skipped tasks are idempotent `CreateOrUpdate` operations that were no-ops for existing clusters anyway. If infra is missing (e.g., someone manually deleted the bastion), the individual `getOrCreate` functions in the always-run tasks handle recreation.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,6 +57,8 @@ When making changes, reason whether the file is used in VHD building stage, or p
 
 One way to debug / explore / just for fun is to run [e2e](./e2e/) tests. To run locally, follow the readme file under that folder. 
 
+When modifying E2E cluster infrastructure (daemonsets, firewall rules, DNS zones, ACR, proxy) in `prepareCluster`, bump `clusterInfraVersion` in `e2e/cluster.go` so existing cached clusters get re-configured.
+
 The SRE guidelines ground other coding guidelines and practices.
 
 ### Golang Guidelines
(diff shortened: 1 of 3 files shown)


r2k1:
Closing — the idempotent 409 handling in PR #8470 is the better approach. Running the full DAG every time with proper idempotency is simpler and self-healing, with no version constant to maintain.

PR #8481, opened 2026-05-08
chore(deps): update inspektor-gadget

burak-ok:

This updates Inspektor Gadget to v0.52.0

Supersedes #8045

### DIFF
--- a/spec/vhdbuilder/packer/install_ig_spec.sh
+++ b/spec/vhdbuilder/packer/install_ig_spec.sh
@@ -17,18 +17,3 @@ Describe 'ig_extract_upstream_version function'
     The stderr should include "[ig] Could not parse upstream version from 'not-a-version'"
   End
 End
-
-Describe 'ig_validate_version_compatibility function'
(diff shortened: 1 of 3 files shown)


spec/vhdbuilder/packer/install_ig_spec.sh
```diff
@@ -17,18 +17,3 @@ Describe 'ig_extract_upstream_version function'
     The stderr should include "[ig] Could not parse upstream version from 'not-a-version'"
   End
 End
-
-Describe 'ig_validate_version_compatibility function'
```
cameronmeissner, line 21: why the test removal?

burak-ok: IMO keeping this test up to date with version bumps might become annoying. As copilot also mentioned a mismatched version of `ig` and `ig-gadgets` will be caught in a later (but heavier) test.

If its preferred to also have this I can include this again

cameronmeissner (approved):

PR #8499, opened 2026-05-12
fix: bump windows base image version for 5B & improve windows image bump script output

timmy-wright:
For some reason the bump image version script broke, so this PR fixes the script and bumps the image versions for 5B release.

Fixes #

### DIFF
--- a/vhdbuilder/packer/windows/update_windows_base_versions.sh
+++ b/vhdbuilder/packer/windows/update_windows_base_versions.sh
@@ -66,7 +66,7 @@ for key in "${sku_keys[@]}"; do
 	printf '%q ' "${az_image_list_cmd[@]}"
 	echo ""
 
-	latest_version=$(eval "${az_image_list_cmd[@]}" 2>/dev/null | sort -uV | tail -n 1) || true
+	latest_version=$("${az_image_list_cmd[@]}" 2>/dev/null | sort -uV | tail -n 1) || true
 	status=$? # Capture the exit status immediately
 	if [ $status -ne 0 ]; then
 		echo "  ERROR: Command to get latest versions failed with exit code $status"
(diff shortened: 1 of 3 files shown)


djsly (approved):

vhdbuilder/packer/windows/update_windows_base_versions.sh
```diff
@@ -46,57 +46,55 @@ updated=0
...
+	azimagelistcommand="az vm image list -p \"${publisher}\" -f \"${offer}\" -s \"${sku}\" --all --query [].version -o tsv"
+	echo "  ${azimagelistcommand}"
+
+	latest_version=$(eval "${azimagelistcommand}" 2>/dev/null | sort -uV | tail -n 1) || true
+	status=$? # Capture the exit status immediately
+	if [ $status -ne 0 ]; then
+		echo "  ERROR: Command to get latest versions failed with exit code $status"
+		errors=$((errors + 1))
+		continue
+	fi
```
sinmentis: I think we should avoid `eval` here.

A safer pattern is to build the command as a Bash array, print it with shell quoting for debugging, and execute the array directly:

```suggestion
az_image_list_cmd=(
  az vm image list
  -p "${publisher}"
  -f "${offer}"
  -s "${sku}"
  --all
  --query "[].version"
  -o tsv
)

printf ' %q' "${az_image_list_cmd[@]}"
echo

if ! latest_version=$("${az_image_list_cmd[@]}" | sort -uV | tail -n 1); then
  echo " ERROR: Command to get latest versions failed with exit code $?"
  errors=$((errors + 1))
  continue
fi
```

timmy-wright: Changed.

sinmentis (approved):

benjamin-brady (approved):

PR #8505, opened 2026-05-13
ci: declare least-privilege permissions on the remaining 11 workflows

arpitjain099:
This repo already has explicit `permissions:` blocks on the major workflows (`codeql-analysis.yml`, `golangci-lint.yml`, `hotfix-generate.yml`, `labeler.yaml`, `pr-lint.yaml`, `tidy.yaml`, etc.). This PR finishes the job by adding scoped permissions to the 11 remaining workflows so none of them implicitly inherit a write-capable `GITHUB_TOKEN`.

The bulk of the changes are `permissions: contents: read` because the workflows just lint, vet, or test against a checkout:

- `go-test.yml`, `no-sudo-check.yml`, `shellcheck.yml`, `shellspec.yaml`
- `validate-components.yml`, `validate-image-version.yml`
- `validate-windows-binary-signature.yaml`, `validate-windows-ut.yml`

Three workflows get a different scope because they actually do something with a token:

- `auto-update.yml` invokes `tibdex/auto-update@v2` with `secrets.GITHUB_TOKEN` to refresh PR branches off `main`, so it gets `contents: write` + `pull-requests: read`.
- `generate-kubelet-flags.yaml` performs its push + PR creation using `secrets.PERSONAL_ACCESS_TOKEN` (an explicit user PAT, not the default token). The default `GITHUB_TOKEN` only needs `contents: read` for the checkout; an inline comment was added documenting that.
- `validate-pull-request-source.yml` doesn't even check out the repo; it only inspects `github.event.pull_request.head.repo.full_name`. `permissions: {}` (deny all) is the most accurate.

Each edited file was validated with `python3 -c "import yaml; yaml.safe_load(...)"`.

### DIFF
--- a/.github/workflows/auto-update.yml
+++ b/.github/workflows/auto-update.yml
@@ -6,6 +6,10 @@ on:
     branches:
       - main
 
+permissions:
+  contents: write
+  pull-requests: read
+
 jobs:
   Auto:
     name: Auto-update
--- a/.github/workflows/generate-kubelet-flags.yaml
+++ b/.github/workflows/generate-kubelet-flags.yaml
@@ -2,6 +2,11 @@ name: Generate valid kubelet flags per supported k8s version
 on:
   workflow_dispatch: {}
 
+# The push + PR creation steps below explicitly use secrets.PERSONAL_ACCESS_TOKEN,
+# so the default GITHUB_TOKEN does not need write scopes here.
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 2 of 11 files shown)


arpitjain099:
Hi @awesomenix, gentle ping on this. PR has been open for 4 days without review. I noticed you've been on the recent-merger side of recent merges in this repo. When you have a moment, would you mind giving it a quick look? No urgency. Happy to address any feedback.

PR #8521, opened 2026-05-14
feat(linux): add build support for GB200/300 image series

keith-ms:
This change pulls in the commits specific to the `release-gb200` branch into `main` rather than trying to merge commit from `main` into `release-gb200`.

I've modified the `vhdbuilder/packer/pre-install-dependencies.sh` from the `release-gb200` branch so that the NVIDIA kernel path from the PPA depends on the presence of the `GB200` feature flag.


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -634,6 +639,44 @@ ensurePodInfraContainerImage() {
     rm -f ${POD_INFRA_CONTAINER_IMAGE_TAR}
 }
 
+validateKubeletNodeLabels() {
```
cameronmeissner, line 642: where is this used?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -720,6 +720,68 @@ if [ $OS = $UBUNTU_OS_NAME ] && [ "$(isARM64)" -ne 1 ]; then  # No ARM64 SKU wit
 EOF
 fi
 
+if grep -q "GB200" <<< "$FEATURE_FLAGS"; then
+  # The GB200 feature flag should only be set for arm64 and Ubuntu 24.04, but validate
+  if [ ${UBUNTU_RELEASE} = "24.04" ]; then
```
cameronmeissner: nit: we're not checking CPU architecture here like the comment above is suggesting

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -720,6 +720,68 @@ if [ $OS = $UBUNTU_OS_NAME ] && [ "$(isARM64)" -ne 1 ]; then  # No ARM64 SKU wit
...
+    # Need to replicate all functionality from github.com/azure/aks-gpu/install.sh.
+    # aks-gpu is designed to run at node boot/join time, whereas the GB200 VHD is set up
+    # to have all drivers installed at VHD build time.
+    #
+    # TODO(abenn135): move all GPU installation logic back into the AgentBaker repo, and
```
cameronmeissner: can we remove the TODO?

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -720,6 +720,68 @@ if [ $OS = $UBUNTU_OS_NAME ] && [ "$(isARM64)" -ne 1 ]; then  # No ARM64 SKU wit
...
+
+    # Now we are off-piste: enable DCGM, DCGM exporter, container device plugin, and the NVIDIA containerd config.
+    systemctl enable nvidia-dcgm
+    systemctl enable nvidia-dcgm-exporter
+    systemctl enable nvidia-device-plugin
+    systemctl enable openibd
+
+    # One additional request from MAI: Disable the AKS node problem detector. When this file is present, the Azure AKS VM Extension assumes the NPD has been installed on the VHD and skips installing it at provision time.
+    mkdir -p /etc/node-problem-detector.d/
+    touch /etc/node-problem-detector.d/skip_vhd_npd
```
cameronmeissner, line 778: did you confirm already with the observability team that this skip file is respected by AKS VM extension? I don't think we've moved NPD into the VHD yet

cc @chmill-zz

packer.mk
```diff
@@ -12,8 +12,13 @@ build-packer: setup-golang generate-prefetch-scripts build-image-fetcher build-a
 ifeq (${ARCHITECTURE},ARM64)
 	@echo "${MODE}: Building with Hyper-v generation 2 ARM64 VM"
 ifeq (${OS_SKU},Ubuntu)
+ifeq ($(findstring GB200,$(FEATURE_FLAGS)),GB200)
```
cameronmeissner, line 15: @copilot is this syntax correct?

vhdbuilder/packer/install-dependencies.sh
keith-ms: I ingested this from the `release-gb200` branch. During the testing I did, the expected version of node problem detector got installed. I'd recommend keeping this.

parts/linux/cloud-init/artifacts/cse_config.sh
keith-ms: This isn't used. I can remove it, but it probably should be integrated because a label over 63 characters will cause kubelet to fail to start.

vhdbuilder/packer/install-dependencies.sh
chmill-zz: the skip file for npd is in place on the extension side, just not implemented broadly on agentbaker side. So the behavior keith tested is working as expected which is fabulous news.

keith-ms: @copilot Don't remove the NPD change. Please revert that change in your commit.

parts/linux/cloud-init/artifacts/ubuntu/doca.list
cameronmeissner, line 1: not gonna block on this, though I'd prefer we move these artifacts that are only uploaded to GB200/300 VHDs into a subfolder, maybe called `graceblackwell` or something

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -39,6 +39,11 @@ if [ $OS = $UBUNTU_OS_NAME ]; then
     else
       dpkg --get-selections | grep -e "linux-\(headers\|modules\|image\)" | grep -v "linux-\(headers\|modules\|image\)-azure" | grep -v "$current_kernel" | tr -s '[[:space:]]' | tr '\t' ' ' | cut -d' ' -f1 | xargs -I{} apt-get --purge remove -yq {}
     fi
+  else
+    # However, for the 24.04 ARM images, we MUST have both -azure and -azure-nvidia kernels, so that we can run on either vanilla ARM64 hardware or GB200.
+    if [ $(dpkg --get-selections | grep -c "linux-image") -lt 2 ]; then
+      echo "ERROR: Ubuntu 24.04 ARM image is missing either the -azure or -azure-nvidia kernel, cannot continue!" && exit 1
```
cameronmeissner, line 45: nit: personally I'd move this sort of thing to content tests, though I understand the rationale of leaving it here for now to speed up build times

cameronmeissner (approved):

vhdbuilder/packer/gb200-mai-bom.json
```diff
@@ -0,0 +1,46 @@
+{
```
ganeshkumarashok, line 1: Do you plan to manually maintain these versions? Or would you like to auto-update with Renovate? I assume manual given the difficulty in obtaining quota to test post updates.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -345,22 +345,27 @@ LimitNOFILE=1048576
...
-  rm -f /etc/containerd/config.toml
-  if [ "${GPU_NODE}" = "true" ]; then
-    # Check VM tag directly to determine if GPU drivers should be skipped
-    export -f should_skip_nvidia_drivers
-    should_skip=$(should_skip_nvidia_drivers)
-    if [ "$?" -eq 0 ] && [ "${should_skip}" = "true" ]; then
-      echo "Generating non-GPU containerd config for GPU node due to VM tags"
-      echo "${CONTAINERD_CONFIG_NO_GPU_CONTENT}" | base64 -d > /etc/containerd/config.toml || exit $ERR_FILE_WATCH_TIMEOUT
+
+  if grep -q 'BinaryName = "/usr/bin/nvidia-container-runtime"' /etc/containerd/config.toml 2>/dev/null; then
```
ganeshkumarashok, line 349: Should there be an explicit GB 200/300 check here as well?

keith-ms: This happens at runtime, not build time, and I don't believe there's a way to check this at that point (the feature flag isn't present, though perhaps you can by SKU via IMDS, but at the risk of failure due to IMDS access problems).

parts/linux/cloud-init/artifacts/ubuntu/doca.list
```diff
@@ -0,0 +1,2 @@
+deb [arch=arm64 signed-by=/etc/apt/keyrings/doca-net.pub] https://linux.mellanox.com/public/repo/doca/latest/ubuntu24.04/arm64-sbsa/ ./
+deb [arch=amd64 signed-by=/etc/apt/keyrings/doca-net.pub] https://linux.mellanox.com/public/repo/doca/latest/ubuntu24.04/x86_64/ ./
```
ganeshkumarashok, line 2: Is this amd64 block correct? Thought it was supposed to be arm64?

keith-ms: It is correct, I added both repositories for completeness.

vhdbuilder/packer/gb200-mai-bom.json
keith-ms: This is done because the image requires very specific versions of packages. This isn't meant to be a broadly useful image, despite the `gb200` in the name implying something generic.

vhdbuilder/packer/post-install-dependencies.sh
```diff
@@ -39,6 +39,11 @@ if [ $OS = $UBUNTU_OS_NAME ]; then
     else
       dpkg --get-selections | grep -e "linux-\(headers\|modules\|image\)" | grep -v "linux-\(headers\|modules\|image\)-azure" | grep -v "$current_kernel" | tr -s '[[:space:]]' | tr '\t' ' ' | cut -d' ' -f1 | xargs -I{} apt-get --purge remove -yq {}
     fi
+  else
+    # However, for the 24.04 ARM images, we MUST have both -azure and -azure-nvidia kernels, so that we can run on either vanilla ARM64 hardware or GB200.
+    if [ $(dpkg --get-selections | grep -c "linux-image") -lt 2 ]; then
```
ganeshkumarashok, line 44: Would it be worth including grep -q "GB200" <<<"$FEATURE_FLAGS" for future safety, when there's no dual booting?

ganeshkumarashok (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
ganeshkumarashok: Fair. Could also consider adding a sentinel flag during build time, such as `/etc/aks/gpu-config-baked.marker`, which can be checked here. Just an optional suggestion

vhdbuilder/packer/post-install-dependencies.sh
mbsabath: added

cameronmeissner (approved):

cameronmeissner (approved):

PR #8527, opened 2026-05-14
fix: remove old kube-proxy images and updated cloud manager to match RP

awesomenix:
match versions in RP


Devinwong (approved):
always happy to approve clean-up

PR #8580, opened 2026-05-25
fix(e2e): strict wireserver validation — fail fast on unexpected curl exits

r2k1:

Tightens the e2e wireserver-block validator so it fails fast on any unexpected curl exit code from an unprivileged pod, instead of retrying for a minute and only failing if no acceptable exit code ever showed up.

### Why

`validateWireServerBlocked` is a security check: pods must not be able to reach the wireserver IP (168.63.129.16). The previous implementation retried for 1 minute, passing the check if curl eventually returned exit 28 (timeout). Two problems with that:

1. It treats the result of a security check as something to wait out — but a successful curl from a pod is **never** a transient condition. It means FORWARD DROP/REJECT rules are missing right now, and the test should surface that immediately.
2. The retry loop bounded the budget by time, not by observations. If the first curl returned exit 0 (reachable) but the last one returned 28 (timeout), the test would pass while a real regression was present somewhere in the window.

### What

- Whitelist exit codes 28 (FORWARD DROP timeout) and 7 (FORWARD REJECT refused) as the only valid "wireserver blocked" signals.
- Anything else fails loudly with full diagnostics: FORWARD chain, KUBE-FORWARD chain, iptables-save filter, and conntrack entries for the wireserver IP. This makes regressions trivial to triage from the test log.
- Retry the exec call only on transient kube-apiserver exec failures, never on the curl result itself — one observation of an unexpected exit code is enough to fail.

This is strictly more defensive than the original (which only accepted exit 28) because it also accepts REJECT-based blocks, while failing on every other class of regression instead of swallowing them.

### Scope

`e2e/validation.go` only. Test-only change, no product code touched. Extracted from #8480.

N/A — test-only hardening, no linked issue.

### DIFF
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -10,7 +10,6 @@ import (
 	"github.com/Azure/agentbaker/e2e/config"
 	"github.com/Azure/agentbaker/e2e/toolkit"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


e2e/validation.go
```diff
@@ -298,30 +309,37 @@ func validateWireServerBlocked(ctx context.Context, s *Scenario) {
 		},
 	}
 
+	allowedExitCodes := map[string]bool{"28": true, "7": true}
```
timmy-wright, line 312: This would be easier to read with a list of allowable exit codes rather than a map. The constant time lookup benefit we get below doesn't seem worth it given the length of time these tests take to run.

r2k1: Converting to slice won't improve much. You will trade it for slightly more comlicated lookup.
Either way it isolated to function and doesn't seem important

awesomenix (approved):

PR #8593, opened 2026-05-27
fix(windows): don't use version in e2e tests

timmy-wright:

Removes the hard coded windows scripts version from the e2e tests - that way we actually use the scripts from the repo.

Fixes #


janenotjung-hue (approved):
test- LGTM

r2k1 (approved):

PR #8594, opened 2026-05-27
chore: upgrade azuredisk-csi-driver to v1.33.10, v1.34.4 and blob-csi-driver to v1.26.12, v1.27.5

andyzhangx:
Upgrade azuredisk-csi-driver image versions:
- v1.34.3 → v1.34.4
- v1.33.9 → v1.33.10

Upgrade blob-csi-driver image versions:
- v1.26.11 → v1.26.12
- v1.27.4 → v1.27.5

Including corresponding windows-hp image updates.

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -505,21 +505,25 @@
       "multiArchVersionsV2": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes-csi/azuredisk-csi",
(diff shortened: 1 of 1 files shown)


timmy-wright:
image

Looks like this adds 80MB to the VHD as we're doing both latest and prev latest version now. That's OK, but might remove the prev latest version later if the VHDs get full again.

timmy-wright (approved):

PR #8595, opened 2026-05-27
fix: update containerd versions on Ubuntu to fix CVEs

awesomenix:
Fixes bunch of CVE


djsly (approved):

parts/common/components.json
djsly, line 1043: shouldnt we bump this one as well ?

djsly: not yet ready..

djsly (approved):

PR #8598, opened 2026-05-27
feat: reduce node impact by aks-log-collector

awesomenix:
## Summary of Changes

1. *aks-log-collector.sh* — Reduce node impact by splitting normal/full collection

Disk I/O reduction:

 - Added "waagent_full" config flag (default false) that gates the heavy WALinuxAgent-style manifest (rotated syslogs, journal, kern, dpkg, boot, secure, etc.)
 - Default (normal) mode only collects: AKS configs, GPU logs, current syslog, auth.log, dmesg, cloud-init, azure extension logs
 - Files >1MB are tail-truncated via fifo — avoids reading multi-GB logs
 - .gz (already-compressed rotated logs) are skipped entirely
 - Truncated files preserve their original path in the zip

Disruptive commands gated behind sysinfo flag:

 - conntrack -L / conntrack -S (locks conntrack spinlock)
 - ss -anoempiO --cgroup (locks socket tables)
 - systemctl status --all -fr (queries all units)
 - find /dev /etc /var/lib/waagent /var/log -ls (heavy I/O)
 - crictl stats / crictl statsp (can stall containerd)
 - blkid, sysctl -a, lspci -vkPP, etc.

Per-netns disruptive calls (conntrack, ss) also gated behind sysinfo within the COLLECT_NETNS block.

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

2. *aks-log-collector.timer* — Delay first run, simplify triggers

 - Removed OnActiveSec=0m (no longer runs immediately on timer activation)
 - Changed OnBootSec=5min → OnBootSec=10min (gives node more time to stabilize before first collection)
 - Keeps OnUnitActiveSec=60m (hourly thereafter)

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

3. *cse_main.sh* — Start log collector after kubelet is healthy

 - Added systemctlEnableAndStartNoBlock aks-log-collector.timer 30 || true after the kubelet health check
 - Ensures log collection only begins once the node is provisioned and kubelet is running

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

4. *cse_start.sh* — Remove background log upload on success

 - Removed upload_logs & on successful CSE exit
 - Logs are still uploaded on CSE failure (EXIT_CODE != 0)
 - Reduces unnecessary background I/O on the happy path

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

5. *pre-install-dependencies.sh* — Don't start timer during VHD build

 - Changed from systemctlEnableAndStart aks-log-collector.timer to systemctl disable --now aks-log-collector.service
 - Timer is no longer started during VHD build (it gets enabled at provisioning time via CSE instead)
 - Still disables WALA log collection via waagent.conf

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

## Net effect

By default (all flags false), the log collector now:

 - Waits 10 minutes after boot
 - Only collects lightweight AKS-specific configs + last 1MB of key logs
 - Skips all expensive system introspection commands
 - Doesn't read rotated/compressed logs
 - Runs hourly with minimal disk and kernel impact

Full diagnostics are available on-demand via IMDS tags (sysinfo, waagent_full, netns, iptables, nftables).

### DIFF
--- a/parts/linux/cloud-init/artifacts/aks-log-collector.sh
+++ b/parts/linux/cloud-init/artifacts/aks-log-collector.sh
@@ -19,11 +19,13 @@ shopt -s nullglob nocaseglob extglob
 # JSON body:
 # {
 #   "disable": false,
+#   "waagent_full": false,
 #   "files": [ "/etc/skel/.bashrc", "/etc/skel/.bash_profile" ],
 #   "pod_log_namespaces": [ "default", "pahealy" ],
 #   "iptables": false,
 #   "nftables": false,
-#   "netns": true
+#   "netns": true,
+#   "sysinfo": false
 # }
 CONFIG=$(curl -s -H Metadata:true --noproxy '*' 'http://169.254.169.254/metadata/instance/compute?api-version=2021-02-01' | jq '.tagsList[] | select(.name=="aks-log-collector") | .value | fromjson')
 
@@ -36,6 +38,8 @@ CONFIG=$(curl -s -H Metadata:true --noproxy '*' 'http://169.254.169.254/metadata
 COLLECT_IPTABLES=$(<<<"$CONFIG" jq -esRr 'try fromjson catch null | .iptables? // false')
 COLLECT_NFTABLES=$(<<<"$CONFIG" jq -esRr 'try fromjson catch null | .nftables? // false')
 COLLECT_NETNS=$(<<<"$CONFIG" jq -esRr 'try fromjson catch null | .netns? // false')
+COLLECT_SYSINFO=$(<<<"$CONFIG" jq -esRr 'try fromjson catch null | .sysinfo? // false')
+COLLECT_WAAGENT_FULL=$(<<<"$CONFIG" jq -esRr 'try fromjson catch null | .waagent_full? // false')
 
 ### START CONFIGURATION
 ZIP="aks_logs.zip"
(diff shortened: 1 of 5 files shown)


parts/linux/cloud-init/artifacts/aks-log-collector.sh
```diff
@@ -75,63 +79,73 @@ GLOBS+=(/var/log/nvidia*.log)
...
-GLOBS+=(/var/log/dpkg*)
-GLOBS+=(/var/log/yum*)
-GLOBS+=(/var/log/boot*)
-GLOBS+=(/var/log/auth*)
-GLOBS+=(/var/log/secure*)
-GLOBS+=(/var/log/journal*)
 
+# Based on MANIFEST_FULL from Azure Linux Agent's log collector
+# https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/ga/logcollector_manifests.py
+if [ "$COLLECT_WAAGENT_FULL" = "true" ]; then
```
cameronmeissner, line 113: nit: might be a good idea to send the flags to lower-case

parts/linux/cloud-init/artifacts/aks-log-collector.sh
```diff
@@ -75,63 +79,73 @@ GLOBS+=(/var/log/nvidia*.log)
...
+  GLOBS+=(/etc/sysconfig/network-scripts/route-eth*)
+  GLOBS+=(/etc/ufw/ufw.conf)
+  GLOBS+=(/var/lib/hyperv/.kvp_pool_*)
+  GLOBS+=(/var/lib/dhcp/dhclient.eth0.leases)
+  GLOBS+=(/var/lib/dhclient/dhclient-eth0.leases)
+  GLOBS+=(/var/lib/wicked/lease-eth0-dhcp-ipv4.xml)
+  GLOBS+=(/var/lib/waagent/SharedConfig.xml)
+  GLOBS+=(/var/lib/waagent/ManagedIdentity-*.json)
+  # Rotated and additional log files
+  GLOBS+=(/var/log/syslog*)
```
cameronmeissner: I think we should always include the rotated syslog* + messages*

awesomenix: we include the main ones, these are just rotated ones

parts/linux/cloud-init/artifacts/aks-log-collector.timer
```diff
@@ -2,8 +2,7 @@
 Description=AKS Log Collector Timer
 
 [Timer]
-OnActiveSec=0m
-OnBootSec=5min
+OnBootSec=10min
```
cameronmeissner: can we just remove this?

awesomenix: this is for successful and failures cases, my assumption is that on failure RP would have deleted the node anyways, so it wont run?

cameronmeissner: if cse fails: synchronous invocation via CSE start will give us a log upload
if cse succeeds: we enable the timer and should just get log uploads every 60 minutes

I think either way we wouldn't need an extra automatic invocation 10 minutes (or X minutes) after boot

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -586,6 +586,8 @@ function nodePrep {
 
     checkServiceHealth kubelet || exit $ERR_KUBELET_FAIL
 
+    systemctlEnableAndStartNoBlock aks-log-collector.timer 30 || true
```
cameronmeissner: can we at least log a warning if we fail to start the log collector instead of just `|| true`

parts/linux/cloud-init/artifacts/aks-log-collector.sh
cameronmeissner: we need the rotated ones, there's often times where some issue happened some time in the past and by the time we upload that issue's syslog entries were rotated our to .1/.2/.3/etc.

parts/linux/cloud-init/artifacts/aks-log-collector.sh
```diff
@@ -279,27 +308,42 @@ if [ "${COLLECT_NETNS}" = "true" ]; then
...
-    # The API for the log bundle has a max file size (defined above, usually 100MB), so if
-    # adding this last file made the zip go over that size, remove that file and stop processing.
-    FILE_SIZE=$(stat --printf "%s" ${ZIP})
-    if [ "$FILE_SIZE" -ge "$MAX_SIZE" ]; then
-      echo "WARNING: ZIP file size $FILE_SIZE >= $MAX_SIZE; removing last log file and terminating adding more files."
-      zip -d "${ZIP}" $file
-      break
-    fi
+MAX_FILE_SIZE=$((1 * 1024 * 1024))
+echo "Adding log files to zip archive with max file size: $MAX_FILE_SIZE bytes..."
```
cameronmeissner, line 318: this is pretty restrictive, I'd expect syslogs can be larger than 1MB

awesomenix: we do collect the file, it just collects the last 1MB which is quite a bit isnt it?

cameronmeissner (approved):

PR #8599, opened 2026-05-28
fix: limit log collection to latest 10MB for all the files

awesomenix:
limit max collection size to 10MB so we dont overflow memory or the log file

### DIFF
--- a/parts/linux/cloud-init/artifacts/aks-log-collector.sh
+++ b/parts/linux/cloud-init/artifacts/aks-log-collector.sh
@@ -46,6 +46,7 @@ ZIP="aks_logs.zip"
 
 # Log bundle upload max size is limited to 100MB
 MAX_SIZE=104857600
+MAX_COLLECT_SIZE=$((10 * 1024 * 1024))
 
 # File globs to include
 # Smaller and more critical files are closer to the top so that we can be certain they're included.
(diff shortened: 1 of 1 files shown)


parts/linux/cloud-init/artifacts/aks-log-collector.sh
```diff
@@ -314,20 +315,19 @@ fi
...
-echo "Adding log files to zip archive with max file size: $MAX_FILE_SIZE bytes..."
+echo "Adding log files to zip archive with max file size: $MAX_COLLECT_SIZE bytes..."
 for file in "${GLOBS[@]}"; do
   # shellcheck disable=SC3010
   [[ "$file" == *.gz ]] && continue
   test -e "$file" || continue
 
   fsize=$(stat --printf "%s" "$file")
-  if [ "$fsize" -gt "$MAX_FILE_SIZE" ]; then
+  if [ "$fsize" -gt "$MAX_COLLECT_SIZE" ]; then
```
djsly, line 325: xcould we log the file name that we are droppping ?

djsly (approved):

parts/linux/cloud-init/artifacts/aks-log-collector.sh
awesomenix: Unfortunately nobody looks at aks-log-collector logs, its not even collected.

PR #8603, opened 2026-05-28
test(acl): add UsePreviouslyBuiltVHD for AB update test

bfjelds:
Add UsePreviouslyBuiltVHD flag to scenario Config that resolves the VHD to the most recent prior main-branch build in the build gallery, excluding the current pipeline run. This avoids the UUID collision where the current build's VHD and COSI share filesystem UUIDs.

New PreviouslyBuiltSIGVersion function on AzureClient:
- Lists all versions in the build gallery for the image definition
- Filters to branch=refs/heads/main, provisioning Succeeded/Updating
- Excludes current build by matching Config.BuildID against buildId tag
- Sorts by parsed 'now' tag (unix epoch) for deterministic ordering
- Supports PREVIOUSLY_BUILT_VHD_BUILD_ID env var override for debugging

Test_ACL_ABUpdate now sets UsePreviouslyBuiltVHD: true to boot from a prior build before applying the current pipeline's COSI.

Fixes #


bfjelds:
ACL/trident has addressed the need to use previous VHD.  This PR is longer needed.

PR #8604, opened 2026-05-28
fix: prewarm containerd in boothook

awesomenix:
- prewarm containerd, containerd metadata, snapshots during boothook

### DIFF
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -52,6 +52,8 @@ logger -t aks-boothook "boothook start $(date -Ins)"
 
 mkdir -p /opt/azure/containers
 
+nohup /bin/bash /opt/azure/containers/provision_preload.sh >/dev/null 2>&1 &
+
 cat <<'EOF' | base64 -d | gzip -d >%[1]s
 %[2]s
 EOF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -868,7 +868,7 @@ EOF
 }
 
 ensureSnapshotUpdate() {
-    systemctlEnableAndStart snapshot-update.timer 30 || exit $ERR_SNAPSHOT_UPDATE_START_FAIL
+    systemctlEnableAndStartNoBlock snapshot-update.timer 30 || exit $ERR_SNAPSHOT_UPDATE_START_FAIL
 }
 
 ensureMigPartition(){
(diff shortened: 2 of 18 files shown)


parts/linux/cloud-init/artifacts/disk_queue.service
```diff
@@ -3,6 +3,7 @@ Description=Set nr_requests and queue_depth based on experimental tuning
 
 [Service]
 Type=oneshot
+ExecStartPre=/bin/sleep 300
```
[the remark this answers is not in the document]
awesomenix: accident

pkg/agent/baker.go
```diff
@@ -52,6 +52,12 @@ logger -t aks-boothook "boothook start $(date -Ins)"
 
 mkdir -p /opt/azure/containers
 
+nohup /bin/sh -c '/usr/bin/containerd --version >/dev/null 2>&1' >/dev/null 2>&1 &
```
Devinwong: you only prewarms containerd, not kubelet, right? The PR title says both.

awesomenix: Fixed

pkg/agent/baker.go
```diff
@@ -52,6 +52,12 @@ logger -t aks-boothook "boothook start $(date -Ins)"
...
+nohup cat /var/lib/containerd/io.containerd.metadata.v1.bolt/meta.db > /dev/null 2>&1 &
+find /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots -maxdepth 1 > /dev/null 2>&1 &
```
lilypan26: just curious is this only useful when artifact streaming is enabled?

pkg/agent/baker.go
lilypan26: will moving this to boothook cause slowdown for nodes not using scriptless? e.g. PIS doesn't use scriptless yet right

pkg/agent/baker.go
awesomenix: all the time, i think it will not be useful during artifact streaming becase we use a different overlay and that copies over all the snapshots

pkg/agent/baker.go
awesomenix: unforunately the only time we can do this is in boothook since you can see we were calling it in cse_main but it was no use. so we dont use scriptless only during preprovisioning, where speed doesnt matter. We still use scriptless during deployment of PIS nodes.

Devinwong (approved):

PR #8613, opened 2026-05-31
perf: trim GPU provisioning critical path (skip redundant pull, async cleanup, defer DCGM)

ganeshkumarashok:
## Summary

Three small, independent CSE-time optimizations that trim the GPU node provisioning critical path on Ubuntu. None change the default driver install behavior; each is low-risk and self-contained.

### 1. Skip the redundant driver image pull (`configGPUDrivers`)
The `aks-gpu-cuda` image is normally **pre-pulled into the VHD**, but `configGPUDrivers()` unconditionally ran `ctr image pull` at boot — a wasted manifest/layer round trip to MCR (and exposure to MCR throttling). Now we only pull when the image is genuinely absent locally; otherwise we go straight to `ctr run`.

### 2. Async image cleanup (drop `--sync`)
The post-install `ctr images rm --sync` blocked CSE waiting for containerd GC to finish. Dropping `--sync` removes the image reference immediately and lets GC reclaim space asynchronously — same disk outcome, no blocking.

### 3. Defer DCGM telemetry off the critical path
`nvidia-dcgm` and `nvidia-dcgm-exporter` are **monitoring only** and don't gate GPU workload scheduling, yet they were started with the blocking `systemctlEnableAndStart` and hard-`exit`ed CSE on a slow/failed start. They now use `systemctlEnableAndStartNoBlock` and are non-fatal. The `nvidia-device-plugin` start stays **blocking and fatal** because it gates the node advertising GPUs to the scheduler.

## Tests
- New shellspec coverage for `startNvidiaManagedExpServices`: asserts device-plugin stays blocking while dcgm/dcgm-exporter are enqueued off the critical path and don't fail provisioning. (7/7 GPU-service examples pass.)
- `go test ./pkg/agent/...` — pass.
- `make generate` — no snapshot diffs.

## Risk / behavior notes
- #1 keeps a full fallback pull when the image is missing — no regression on non-prebaked images.
- #2 still removes the image; only the synchronous wait is gone.
- #3 makes DCGM/exporter start non-fatal. This is the intended trade-off (telemetry should never block or fail node provisioning); reviewers who want DCGM failures to remain fatal should flag it.

## Coordination
Touches `configGPUDrivers()`, which the held PR #8612 (prebuild GPU kernel module) also edits — these two will need a light rebase against each other whenever both land. This PR is independent and can merge on its own.

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1003,14 +1003,20 @@ configGPUDrivers() {
     if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
         waitForContainerdReady || exit $ERR_GPU_DRIVERS_START_FAIL
         mkdir -p /opt/{actions,gpu}
-        ctr -n k8s.io image pull $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG
+        # The driver image is normally pre-pulled into the VHD; only hit the registry when it is
+        # actually missing so provisioning doesn't pay a redundant manifest/layer round trip.
+        if ! ctr -n k8s.io images ls -q | grep -qx "$NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG"; then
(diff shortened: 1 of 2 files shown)


ganeshkumarashok:
Closing in favor of #8615, which is the identical change from a same-repo branch. AgentBaker requires PRs from `Azure/AgentBaker` branches (fork PRs don't get CI/secrets).

PR #8618, opened 2026-06-01
feat(linux): refactor aks-secure-tls-bootstrap-client installation to use PMC/MCR and bump to v1.1.4-1

cameronmeissner:

refactor aks-secure-tls-bootstrap-client installation to use PMC/MCR now that the client is being built/published by dalec - this enable secure TLS bootstrapping support on FIPS images through dalec's out-of-the-box FIPS support. As such, this PR also updates AgentBaker E2Es to enable secure TLS bootstrapping on ALL VHDs according to user-specified E2E configuration.

This PR also onboard secure TLS bootstrap client to renovate for Linux so we can start consuming updates automatically.

Fixes #


.github/renovate.json
```diff
@@ -456,6 +440,20 @@
         "surajssd"
       ]
     },
+    {
+      "matchPackageNames": [
+        "aks-secure-tls-bootstrap-client"
+      ],
+      "groupName": "aks-secure-tls-bootstrap-client",
+      "assignees": [
```
djsly, line 448: why not all of us ? :D

cameronmeissner: fair enough, just figured you and I have the most context

parts/linux/cloud-init/artifacts/azlosguard/azurelinux-ms-oss.repo
```diff
@@ -0,0 +1,9 @@
+[azurelinux-official-ms-oss]
```
djsly, line 1: I wonder id there is a way to not enable this by default during bootstrapping ? if this only needed during the build ? should we remove it after ?

cameronmeissner: we'll eventually need it for aks-node-controller at runtime right?

djsly (approved):

.github/renovate.json
cameronmeissner: fixed

lilypan26 (approved):

PR #8640, opened 2026-06-04
fix: widen fs.file-max sysctl from int32 to int64

fcher:

Sets the right data type(int64) for fs.file-max sysctl parameter, matching the Linux kernel default that customers already see at runtime. Also matches the documentation here: https://learn.microsoft.com/en-us/azure/aks/custom-node-configuration?tabs=linux-node-pools

`buf breaking` flags this as a source-level field-type change. This is intentional and coordinated. The next step would be update the validator in aks-rp repo

Parent Work Item: https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/37310362


fcher:
> @fcher please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

awesomenix (approved):

PR #8646, opened 2026-06-05
fix: use one big vnet and attach AKS clusters to it to avoid creating bastion multiple times

awesomenix:
- Make everyone's life simple
- README has all the details, but idea is to use shared VNET and stop doing unnecessary activities multiple times.

### DIFF
--- a/e2e/node_config.go
+++ b/e2e/node_config.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"context"
 	"encoding/base64"
 	"fmt"
 	"testing"
@@ -65,7 +66,7 @@ func baseKubeletConfig() *aksnodeconfigv1.KubeletConfig {
 			EventRecordQps: to.Ptr(int32(0)),
 			ClusterDomain:  "cluster.local",
 			ClusterDns: []string{
-				"10.0.0.10",
+				"172.16.0.10",
 			},
 			StreamingConnectionIdleTimeout: "4h",
 			NodeStatusUpdateFrequency:      "10s",
@@ -98,7 +99,7 @@ func baseKubeletConfig() *aksnodeconfigv1.KubeletConfig {
 	}
 }
 
-func getBaseNBC(t testing.TB, cluster *Cluster, vhd *config.Image) (*datamodel.NodeBootstrappingConfiguration, error) {
+func getBaseNBC(ctx context.Context, t testing.TB, cluster *Cluster, vhd *config.Image) (*datamodel.NodeBootstrappingConfiguration, error) {
 	var nbc *datamodel.NodeBootstrappingConfiguration
 
 	if vhd.Distro.IsWindowsDistro() {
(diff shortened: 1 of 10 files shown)


Devinwong (commented):
e

e2e/shared_infra.go
```diff
@@ -0,0 +1,615 @@
...
+func ensureSubnet(ctx context.Context, rg, vnetName, subnetName, cidr string) error {
+	_, err := config.Azure.Subnet.Get(ctx, rg, vnetName, subnetName, nil)
+	if err == nil {
+		return nil
+	}
+	if !isNotFoundError(err) {
+		return fmt.Errorf("checking subnet %s: %w", subnetName, err)
+	}
+
+	return retryOn409(ctx, fmt.Sprintf("creating subnet %s", subnetName), func() error {
```
Devinwong, line 227: Condsider adding a retry on 400 too?
Azure returns HTTP 400 (Bad Request) for overlapping subnet address spaces, not 409 Conflict. 

if two tests try to create the same subnet name(deterministic from cluster name), only one wins. But if two different clusters hash to the same starting slot and list subnets at the same time, they could both pick the same free CIDR and one will get aconflict on a different subnet name with the same address space. Azure will reject this with a "subnet address space overlaps" error, which is not a 400

e2e/shared_infra.go
```diff
@@ -0,0 +1,615 @@
...
+		return nil, fmt.Errorf("ensuring shared firewall: %w", err)
+	}
+
+	identityID, tenantID, err := ensureClusterIdentity(ctx, rg, location)
+	if err != nil {
+		return nil, fmt.Errorf("ensuring cluster identity: %w", err)
+	}
+
+	// Best-effort cleanup of orphaned cluster subnets
+	cleanupOrphanedSubnets(ctx, rg)
```
Devinwong, line 85: async for unblocking every test?

e2e/node_config.go
```diff
@@ -98,7 +99,7 @@ func baseKubeletConfig() *aksnodeconfigv1.KubeletConfig {
 	}
 }
 
-func getBaseNBC(t testing.TB, cluster *Cluster, vhd *config.Image) (*datamodel.NodeBootstrappingConfiguration, error) {
+func getBaseNBC(ctx context.Context, t testing.TB, cluster *Cluster, vhd *config.Image) (*datamodel.NodeBootstrappingConfiguration, error) {
```
Devinwong, line 102: ctx it not really used?

e2e/shared_infra.go
awesomenix: its impossible to have 2 different clusters to have same hash, because the hash is determined by the public SSH key. WE should be ok here

Devinwong (approved):

e2e/shared_infra.go
awesomenix: deferring to next PR

PR #8651, opened 2026-06-05
fix(vhd): remove bundled overlaybd packages after artifact streaming install

ganeshkumarashok:

The `acr-mirror` package (artifact streaming) bundles the overlaybd installer packages under `/opt/acr/tools/overlaybd/bin/` as install-time inputs:

- `.deb` variant: `overlaybd.deb` (~14 MB) + `overlaybd-snapshotter.deb` (~44 MB)
- `.rpm` variant: `overlaybd.rpm` (~11 MB) + `overlaybd-snapshotter.rpm` (~45 MB)

`/opt/acr/tools/overlaybd/install.sh` `dpkg`/`tdnf`-installs these at VHD bake time, and nothing references them at runtime (the only reference to `bin/*.{deb,rpm}` is `install.sh` itself). They were left on disk and shipped on **every** VHD where artifact streaming is installed — ~55-58 MB of dead weight — regardless of whether a customer ever enables the feature.

This change removes them right after install, mirroring the existing cleanup of the downloaded `acr-mirror` package (`rm "$MIRROR_DOWNLOAD_PATH"` a few lines above). It covers both `.deb` and `.rpm` so it applies to Ubuntu and Azure Linux. A `linux-vhd-content-test` assertion verifies no leftover `.deb`/`.rpm` packages remain under the overlaybd `bin` directory (guarded so it is skipped on VHDs where artifact streaming is not installed).

No runtime behavior change: the overlaybd binaries are already installed into their system locations by `install.sh`; only the now-redundant installer packages are deleted.

Fixes #

### DIFF
--- a/vhdbuilder/packer/install-dependencies.sh
+++ b/vhdbuilder/packer/install-dependencies.sh
@@ -646,6 +646,8 @@ installAndConfigureArtifactStreaming() {
   /opt/acr/tools/overlaybd/config.sh exporterConfig.enable true
   /opt/acr/tools/overlaybd/config.sh exporterConfig.port 9863
   systemctl link /opt/overlaybd/overlaybd-tcmu.service /opt/overlaybd/snapshotter/overlaybd-snapshotter.service
+  # Remove the bundled overlaybd installer packages (~55-58 MB); install.sh already installed them and they're unused at runtime.
+  rm -f /opt/acr/tools/overlaybd/bin/*.deb /opt/acr/tools/overlaybd/bin/*.rpm
   echo "  - acr-mirror version ${version}" >> ${VHD_LOGS_FILEPATH}
 }
 
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -538,6 +538,27 @@ testAuditDNotPresent() {
   echo "$test:Finish"
 }
 
+testArtifactStreamingPackagesCleanedUp() {
+  local test="testArtifactStreamingPackagesCleanedUp"
+  echo "$test:Start"
+  local overlaybdBinDir="/opt/acr/tools/overlaybd/bin"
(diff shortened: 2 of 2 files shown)


mxj220 (approved):

djsly:
## RCA: Linux PR gate sample (build 166952419)

**Summary**
The Linux PR gate failure for PR #8651 appears to be an infra / PR merge-ref lifecycle issue, not a regression from the overlaybd cleanup change in this PR.

**Evidence**
- Build: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=166952419
- Failing log: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=166952419&view=logs&j=22b1d4aa-9805-56d1-3c23-ee6230498383&t=0745f6fc-e50a-5b8d-2fe0-5939d66e0d0b
- PR files: https://github.com/Azure/AgentBaker/pull/8651/files
- Source commit: https://github.com/Azure/AgentBaker/commit/994159ea685d521bd01db663248bbcb00cc48eac

**Three-level RCA**
1. **Failure surface**: the gate failed while bootstrapping the E2E test VM, before the test payload could validate PR-specific behavior.
2. **Immediate cause**: `run-test.sh` attempted to fetch `refs/pull/8651/merge` inside the test VM.
3. **Deeper cause**: that GitHub PR merge ref was no longer available after the PR merged, so the bootstrap dependency disappeared during/after the gate run. This points to PR merge-ref lifecycle / infra timing rather than the AgentBaker code change.

**Likely RCA / confidence**
- Likely RCA: PR merge-ref lifecycle race in the test VM bootstrap path.
- Confidence: high that this is infra/bootstrap-related; low likelihood that it was caused by the overlaybd cleanup changes in PR #8651.

**Next actions**
- Treat this sample as non-actionable for PR #8651 code changes.
- For infra hardening, prefer fetching a durable commit SHA for post-merge/retry scenarios instead of relying only on `refs/pull/<id>/merge`.

PR #8653, opened 2026-06-07
Cap per-VM STLS retry attempts (AB#38327355)

djsly:
## What & why

ADO: [AB#38327355](https://dev.azure.com/msazure/CloudNativeCompute/_workitems/edit/38327355)

The `aks-secure-tls-bootstrap` client has **no native long-running retry loop** — each invocation makes a single attempt and exits. Re-invocation is driven by kubelet's `Restart=always` pulling the failed STLS unit via its `Wants=` relation.

On **2026-06-04**, a single stuck VM (`b409a057-44a0-4b06-a6de-2e24e59a90a5` in sub `37deca37-...`) hit `GetNonceFailure: produced zero addresses` (DNS) and emitted **2,837 STLS failure events in 53+ hours**. That single VM was **96% of all STLS fail events** in eastus2euap on Jun 5–6 and dropped per-event QoS to ~93% even though only 68 VMs actually failed.

## Approach

New Linux wrapper `parts/linux/cloud-init/artifacts/secure-tls-bootstrap-retry-cap.sh` that the systemd unit invokes instead of the binary directly. The wrapper:

- Tracks attempt count + first-attempt timestamp in `/var/lib/aks-secure-tls-bootstrap/`, wiped per provisioning session by `cse_config.sh` and reset across reboots via boot-id check.
- Applies **exponential backoff** (1 s → 300 s default cap) between attempts.
- Caps per-session retries: **50 attempts OR 7200 s wall-clock**, whichever fires first. Both bounds are env-var configurable from the apiserver.
- On cap, emits a distinct guest-agent event with **TaskName `AKS.Bootstrap.SecureTLSBootstrapping.RetryCapReached`** carrying the last `FinalErrorType`, attempt count, elapsed seconds, and client version. Once capped, further invocations exit `0` immediately without re-running the binary or re-emitting events.

Defaults are conservative — a typical transient blip recovers well within the budget. A stuck VM under these defaults emits ~30 binary-failure events + 1 `RetryCapReached` event over ~2 h, then goes quiet (vs. the 2,837 events / 53 h observed today).

## Full plumbing — 4 new knobs in both deployment paths

**Legacy CSE**: `SecureTLSBootstrappingConfig` fields in `types.go` → template funcs in `baker.go` → env vars in `cse_cmd.sh` → read by `cse_config.sh` and written to `/etc/default/secure-tls-bootstrap`.

[...]

PR #8654, opened 2026-06-07
stls: resolve apiserver IP locally and inject as APISERVER_IP env var

djsly:
## Summary

Eliminates infinite STLS bootstrap retries caused by node-local DNS failures by injecting a pre-resolved API server IP into the STLS client via an `APISERVER_IP` environment variable.

Tracked by **AB#38327357** (parent feature **AB#34681743**, Cameron Meissner).

## Background

When DNS is broken on a node at STLS bootstrap time (CoreDNS not ready, systemd-resolved race, custom DNS misconfig), the STLS client emits

```
rpc error: code = DeadlineExceeded ... name resolver error: produced zero addresses
```

and retries forever (`retry.WithMax(math.MaxUint)` in the gRPC interceptor). Root cause: `client/internal/bootstrap/grpc.go::getServiceClient` builds the dial target as `fmt.Sprintf("%s:443", cfg.APIServerFQDN)`, so gRPC's built-in `dns` resolver re-resolves the FQDN on every retry attempt.

Reference incident: VMId `b409a057-44a0-4b06-a6de-2e24e59a90a5`, 53+ hours of retries on Jun 5–6.

## What this PR does

`configureAndStartSecureTLSBootstrapping` in `parts/linux/cloud-init/artifacts/cse_config.sh` now resolves the API server IP at provisioning time and writes it as `APISERVER_IP=<addr>` into `/etc/default/secure-tls-bootstrap` (the systemd `EnvironmentFile` for `secure-tls-bootstrap.service`).

Resolution order:

1. If `API_SERVER_NAME` is already an IPv4 literal → use it as-is.
2. For `*.privatelink.*` hosts → IMDS instance tag `aksAPIServerIPAddress` (same source as `reconcile-private-hosts.sh`; survives cluster DNS failures).
3. `getent ahostsv4 ${API_SERVER_NAME}` (IPv4 preferred).
4. `getent ahostsv6 ${API_SERVER_NAME}` (IPv6 fallback).
5. Any non-`[0-9a-fA-F:.]` result is discarded.

If every step fails (e.g., DNS already down on a public cluster at CSE time), `APISERVER_IP` stays empty, the line is **not** emitted, and STLS falls back to its existing FQDN-dial behaviour — no regression.

The companion **[Azure/aks-secure-tls-bootstrap#180](https://github.com/Azure/aks-secure-tls-bootstrap/pull/180)** PR:
- Adds `APIServerIP` to `bootstrap.Config`, populated from `os.Getenv("APISERVER_IP")`.
- Switches `getServiceClient` to a `passthrough:///` dial target when the IP is set, with `tls.Config.ServerName = cfg.APIServerFQDN` and `grpc.WithAuthority(net.JoinHostPort(cfg.APIServerFQDN, "443"))` so TLS SAN validation and HTTP/2 `:authority` still resolve against the FQDN. The gRPC `dns` resolver is bypassed entirely.

## Backward / forward compatibility (6-month VHD window)

| AgentBaker (CSE) | STLS client binary | Behaviour |
| --- | --- | --- |
| Old (no env var) | Old (no env reader) | Status quo — DNS dial on every retry. |
| Old (no env var) | New (env reader) | `APISERVER_IP` unset → STLS falls back to FQDN dial. Identical to status quo. |
| New (writes env var) | Old (no env reader) | Old binary silently ignores unknown env vars. FQDN dial. Identical to status quo. |
| New (writes env var) | New (env reader) | **Fix activates**: STLS dials the IP literal, no DNS at gRPC time. |

All four cells are safe — the fix only activates when both sides are new.

## Why this lives in `cse_config.sh` and not RP / parser

STLS is started at `cse_main.sh` line ~390 — **before** the cluster-validation `nslookup` at line ~515. We cannot rely on a side effect of that later DNS call. Resolving and writing the env var inside `configureAndStartSecureTLSBootstrapping`, right before the `EnvironmentFile` write, is the only place that gives us both:
- A live cluster context (so the IMDS tag and `getent` can run).
- An atomic write of `APISERVER_IP` and `BOOTSTRAP_FLAGS` into the same `EnvironmentFile`.

No RP-side / `pkg/agent/datamodel/types.go` / proto changes — IP is resolved locally by CSE.

## aks-node-controller parity

The ANC provisioning path invokes the same `configureAndStartSecureTLSBootstrapping` function via `cse_main.sh`, so no additional ANC code change is required. Confirmed by grep: no separate STLS bootstrap path exists in `aks-node-controller/`.

### Unit (ShellSpec)

`spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh` adds 6 new `It` blocks covering every resolver branch:

1. `API_SERVER_NAME` is an IPv4 literal → echoed as-is.
2. `getent ahostsv4` returns an IPv4 → captured, written.
3. `ahostsv4` fails, `ahostsv6` returns an IPv6 → captured.
4. `*.privatelink.*` host + IMDS returns valid tag → IP from IMDS, no `getent` call.
5. `*.privatelink.*` host + IMDS returns no tag → falls through to DNS.
6. All resolvers fail → `APISERVER_IP=` line absent, CSE does **not** fail.
7. IMDS returns garbage → discarded by sanity-strip, falls through to DNS.

Full suite: **97/97 green** under podman + the standard `aksdataplanedev.azurecr.io/shellspec` base image.

### E2E

New test `Test_Ubuntu2204_SecureTLSBootstrapping_APIServerIPEnvVar` in `e2e/scenario_test.go` provisions an STLS-enabled Ubuntu 22.04 node and validates that `/etc/default/secure-tls-bootstrap` contains `APISERVER_IP=`. This is a smoke test of the AgentBaker side.

The full DNS-blackhole e2e (assert STLS bootstrap succeeds with DNS broken) requires the new STLS client binary baked into the VHD and will be added in a follow-up PR once #180 lands.

### Lint / generate

- `make validate-shell`: no new shellcheck issues introduced (pre-existing SC3014 warnings unchanged).
- `make generate`: no snapshot drift (verified via `GOPROXY=https://proxy.golang.org,direct GENERATE_TEST_DATA="true" go test ./pkg/agent/...`).

## Risks / open items

- **CSE-time DNS already broken on a public cluster**: no IMDS tag (private only), `getent` fails, `APISERVER_IP=""`, STLS falls back to FQDN dial — same as today. Not worse, but the fix doesn't help this narrow case. Future RP-side enhancement is the only complete fix; out of scope.
- **API server IP changes mid-bootstrap**: extremely rare within the short STLS window; would surface as TCP/TLS error rather than DNS, eventually retries succeed. Acceptable; orthogonal to AB#38327355 (retry cap).
- **Windows**: out of scope.
- **`kube-apiserver-proxy` literal naming** (from the ADO item): the string does not appear in either repo. The real dial target is `${APIServerFQDN}`. Flagging for cross-check with **@cameronmeissner** before landing.

## Related work items

- **AB#38327357** — this PR
- **AB#34681743** — parent STLS Phase 1 feature
- **AB#38327355** — STLS retry cap (sibling)
- **AB#38327356** — per-VM STLS QoS metric (sibling)

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -2429,6 +2429,41 @@ func Test_Ubuntu2404_SecureTLSBootstrapping_BootstrapToken_Fallback(t *testing.T
 	})
 }
 
+// Test_Ubuntu2204_SecureTLSBootstrapping_APIServerIPEnvVar validates that the
+// CSE shell code (configureAndStartSecureTLSBootstrapping in cse_config.sh)
+// resolves the API server IP at provisioning time and writes it as
+// APISERVER_IP=<addr> into /etc/default/secure-tls-bootstrap.
+//
+// Tracking: AB#38327357. The companion STLS client change in
+// Azure/aks-secure-tls-bootstrap reads this env var and dials the IP literal
+// directly so the gRPC dns:/// resolver is never consulted on retries.
+//
+// This test validates only the AgentBaker side (the env var is correctly
+// populated). The end-to-end "DNS blackhole, STLS still succeeds" test
+// requires the new STLS client binary baked into the VHD and is tracked as
+// a follow-up.
+func Test_Ubuntu2204_SecureTLSBootstrapping_APIServerIPEnvVar(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "validates that CSE writes APISERVER_IP into /etc/default/secure-tls-bootstrap so STLS can dial without DNS",
+		Config: Config{
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDUbuntu2204Gen2Containerd,
+			BootstrapConfigMutator: func(_ *Cluster, nbc *datamodel.NodeBootstrappingConfiguration) {
+				nbc.SecureTLSBootstrappingConfig = &datamodel.SecureTLSBootstrappingConfig{
+					Enabled: true,
+				}
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				// The env-var line must be present and assign a non-empty value.
+				// The resolver block falls back through IMDS tag -> getent ahostsv4
+				// -> getent ahostsv6 and writes nothing if all sources fail, so a
+				// missing line would indicate a hard regression.
+				ValidateFileHasContent(ctx, s, "/etc/default/secure-tls-bootstrap", "APISERVER_IP=")
+			},
+		},
+	})
+}
+
 func Test_Ubuntu2404Gen2_GPUNoDriver(t *testing.T) {
 	RunScenario(t, &Scenario{
 		Description: "Tests that a GPU-enabled node using the Ubuntu 2404 VHD opting for skipping gpu driver installation can be properly bootstrapped",
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -548,12 +548,58 @@ configureAndStartSecureTLSBootstrapping() {
         BOOTSTRAP_CLIENT_FLAGS="${BOOTSTRAP_CLIENT_FLAGS} --deadline=${SECURE_TLS_BOOTSTRAPPING_DEADLINE}"
     fi
 
+    # AB#38327357: resolve the apiserver IP locally and hand it to STLS via the
+    # APISERVER_IP env var so the client can dial the literal IP and skip the
+    # gRPC dns:/// resolver. If anything fails the var stays empty, the line
+    # is omitted, and STLS falls back to its existing FQDN dial path. Best
+    # effort only — must never fail CSE.
+    APISERVER_IP=""
+    case "${API_SERVER_NAME}" in
+        ''|*[!0-9.]*)
+            # Not a plain IPv4 literal. Try the IMDS aksAPIServerIPAddress tag
+            # (private clusters only — same source reconcile-private-hosts.sh
+            # uses for privatelink FQDNs), then DNS via getent.
+            case "${API_SERVER_NAME}" in
+                *.privatelink.*)
+                    APISERVER_IP=$(curl -sSL -m 5 -H "Metadata: true" \
+                        "http://169.254.169.254/metadata/instance/compute/tags?api-version=2019-03-11&format=text" 2>/dev/null \
+                        | tr ';' '\n' \
+                        | awk -F: 'tolower($1) == "aksapiserveripaddress" { print $2; exit }')
+                    # Discard IMDS values that are not plausible IP literals so
+                    # we fall through to getent instead of short-circuiting on
+                    # an invalid (or absent) tag.
+                    case "${APISERVER_IP}" in
+                        ''|*[!0-9a-fA-F:.]*) APISERVER_IP="" ;;
+                    esac
+                    ;;
+            esac
+            if [ -z "${APISERVER_IP}" ] && [ -n "${API_SERVER_NAME}" ]; then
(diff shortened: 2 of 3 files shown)


djsly:
**AgentBaker Linux PR gate — E2E failure (needs deeper triage)**

- Run: [167080536](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=167080536) (`failed`)
- Failed task: `Run AgentBaker E2E` → AzureCLI → exit 1 (DONE 460 tests, 95 skipped, **3 failures** in 1501.866s)

**Failing leaves (all 3 on the same parent test):**
- `Test_Ubuntu2204Gen2_ImagePullIdentityBinding_NetworkIsolated/default` (2.76s) — `test_helpers.go:227` 🔴 (early-fail, empty error string in build log)
- `Test_Ubuntu2204Gen2_ImagePullIdentityBinding_NetworkIsolated/scriptless_nbc` (0.24s) — same
- `Test_Ubuntu2204Gen2_ImagePullIdentityBinding_NetworkIsolated` (root container, 0.10s)

**Signal — strongly correlated with sibling PR #8653:** Build [167080476](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=167080476) (PR #8653) for the same author and overlapping STLS workstream shows the **same** `Test_Ubuntu2204Gen2_ImagePullIdentityBinding_NetworkIsolated/{default,scriptless_nbc}` failure shape with empty error strings at `test_helpers.go:227`. Two PRs failing the same network-isolated scenario at sub-3s with empty error strings points away from a PR-specific behavioural regression and toward either (a) a shared early-setup error in this scenario (likely test-helper precondition / VMSS setup raising before the scenario body runs) or (b) a transient infra issue in the NetworkIsolated subnet path during this window.

**This PR's changes:**
- `e2e/scenario_test.go` (+35), `parts/linux/cloud-init/artifacts/cse_config.sh` (+46, adds `resolveAPIServerIP` + `APISERVER_IP` env var injection), `spec/.../cse_config_spec.sh` (+145 ShellSpec).
- The PR does NOT modify any image-pull or NetworkIsolated test wiring; the `APISERVER_IP` env injection runs in `cse_config.sh` before kubelet bootstrap and shouldn't gate image-pull identity binding.

**Confidence:** Medium-low on root cause without the actual error message at `test_helpers.go:227` (build log shows the marker `🔴 FAIL:` with an empty body — the real reason is in the per-scenario VM/log artifact). High that this is **not** caused by the `APISERVER_IP` injection in `cse_config.sh` (no path from APISERVER_IP env to ImagePullIdentityBinding test fixture).

**Strongest alternative (likely the truth):** shared early-setup failure in `Test_Ubuntu2204Gen2_ImagePullIdentityBinding_NetworkIsolated` (test fixture / VMSS-with-private-cluster precondition) — refuted as "PR-caused" by the same shape on the sibling PR #8653 that touches completely different STLS code.

**Recommended next action (owner: PR author):**
1. Open the scenario-logs artifact for `Test_Ubuntu2204Gen2_ImagePullIdentityBinding_NetworkIsolated/default` — `test_helpers.go:227` is the per-scenario error printer; the real assertion is one frame above. The early <3s timing strongly suggests a pre-VMSS precondition (e.g. private-cluster networking setup, ACR-with-private-endpoint, identity binding RBAC).
2. Rerun the failing job once. If it reproduces with the same sub-3s shape on both this PR and #8653, file a flake/infra tracker on the `ImagePullIdentityBinding_NetworkIsolated` scenario rather than blocking either PR.
3. PR-quality note (not a fix request): the `APISERVER_IP` injection in `cse_config.sh` looks unrelated to this failure.

_Posted by Clawpilot AgentBaker gate detective._

cameronmeissner:
not needed

PR #8661, opened 2026-06-08
feat(gpu): pre-bake NVIDIA CUDA kernel module into the VHD

ganeshkumarashok:
## What
Move the ~80–150s in-CSE NVIDIA **DKMS kernel-module compile** off the node-provisioning critical path by compiling it at VHD build time.

- **VHD build** (`install-dependencies.sh`), opt-in via `FEATURE_FLAGS=NVIDIA_CUDA_PREBAKE`: after pre-pulling the `aks-gpu-cuda` image, run it in **build-only** mode. aks-gpu compiles + DKMS-registers the kernel module and stages userspace libs against the VHD's target kernel with **no device access** (safe on the GPU-less Packer builders), and writes `/opt/azure/aks-gpu/dkms-marker`. The build-only path first installs `gcc make libc6-dev` — the standard non-GPU builder ships gcc/make but **not** `libc6-dev`, without which `nvidia-installer` cannot compile.
- **Node boot** (`cse_config.sh`): `configGPUDrivers` passes **install-skip-build** when the marker is present **and the marker's `driver_kind` matches this node's `NVIDIA_GPU_DRIVER_TYPE`**, so aks-gpu skips the recompile and runs only device-init. aks-gpu independently re-validates the marker (kernel + version + kind) and falls back to a full build on any remaining mismatch (e.g. kernel drift), so behaviour is unchanged on non-prebaked VHDs (no marker → `install`). Integrated with the timed `installGPUDriverImage` wrapper on `main`.
- **Non-GPU / `gpu-driver=none` teardown** (`cse_install_ubuntu.sh::cleanUpPrebakedGPUDriver`): when a shared Ubuntu VHD that was prebaked is used for a non-GPU node, deregister the DKMS module and remove the staged driver userspace **libs and binaries** (e.g. `nvidia-smi`) so the node is genuinely driver-free.

## Why
The DKMS compile dominates GPU driver install on every GPU node boot. Pre-baking removes it from the critical path. Works transparently on GPU pools (which default to `ConfigGPUDriverIfNeeded=true`).

## Key correctness decisions (surfaced by e2e)
- **`driver_kind` guard (GRID safety):** a CUDA-prebaked marker on a *shared* VHD must **not** short-circuit a **GRID** node's install — the GRID image may not support `install-skip-build` and would fail to stage its own userspace files (observed as **CSE exit 84** in e2e). The guard restricts skip-build to nodes whose driver kind matches the baked marker.
- **Teardown removes binaries + skips slow `dkms remove`:** leaving `nvidia-smi` on `PATH` with its libs removed makes it *error* instead of being "command not found" (a non-GPU node should look driver-free). Replacing the per-version `dkms remove --all` (~35s) with source-tree + built-module removal keeps the non-GPU provisioning path fast.
- **Secure Boot:** the baked module is unsigned, exactly like today's runtime-DKMS-built module. AKS Ubuntu nodes run with Secure Boot **off**, so this is **no regression** vs the existing flow (Secure Boot ON is already incompatible with both the prebake and the runtime-build path; out of scope here).

## Validation (real AgentBaker e2e, `[TEST All VHDs]` pipeline)
Validated end-to-end by baking prebaked VHDs (a test ACR carried the unmerged aks-gpu#162 build-only image) and running the relevant GPU e2e scenarios on Ubuntu 22.04 + 24.04 Gen2:
- **CUDA skip-build proven on a real GPU:** marker exact-match (kernel + driver version + kind) → prebuilt module loads → "recompile skipped" → `nvidia-smi` healthy.
- **GRID** nodes provision via full install (driver_kind guard).
- **GPUNoDriver / gpu-driver=none** node ends up genuinely driver-free.
- **CSE cached-performance** teardown back under budget.
- Bake proven cross-kernel (2204 = 5.15.x, 2404 = 6.8.x); the builder runs an older kernel than the VHD boots, so aks-gpu#162 compiles against the VHD's **target** kernel (newest with installed headers), not `uname -r`.

## Notes / sequencing
- **Depends on Azure/aks-gpu#162** (build-only / install-skip-build + target-kernel selection). The `aks-gpu-cuda` image in `components.json` must carry that support before the flag is enabled.
- The flag is **off by default** here. A follow-up enables `FEATURE_FLAGS=NVIDIA_CUDA_PREBAKE` on the GPU SKU build jobs once the aks-gpu image has shipped — so this PR cannot break existing VHD builds.
- The driver image is intentionally **kept** in the VHD — boot device-init still sources the container toolkit, fabric manager, containerd runtime config and udev rules from it. Dropping it for VHD-size savings is a separate, deferred follow-up.
- The boot-time fallback recompile remains fully intact: nodes get `gcc/libc6-dev` via `installDeps`, so a marker mismatch still compiles at boot.

## Tests
- `make generate-testdata`: no drift (the CSE scripts are dropped via cloud-init, not embedded in the customdata snapshot).
- `verify_shell` POSIX gate: changed lines introduce no new `SC3010/SC3014`.
- ShellSpec (full suite green): new examples for the marker → action selection **including the driver_kind mismatch guard** (match → skip-build, CUDA-marker-on-GRID → full install), plus the faster libs+binaries+marker teardown.

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1221,11 +1221,16 @@ configGPUDrivers() {
         mkdir -p /opt/{actions,gpu}
         # When the kernel module was pre-built into the VHD (build-only at image-bake time),
         # a marker is present. Ask aks-gpu to skip the ~100s DKMS recompile and run only the
-        # device-dependent steps. aks-gpu independently re-validates the marker (kernel +
-        # driver_version + driver_kind) and falls back to a full build on any mismatch, so this
-        # is safe even after a kernel upgrade or on a shared VHD with a different driver kind.
+        # device-dependent steps -- but ONLY when the marker's driver_kind matches THIS node's
+        # driver (NVIDIA_GPU_DRIVER_TYPE). A CUDA-prebaked marker on a GRID node (or vice-versa)
+        # must request a full "install": the other driver image may not even support
+        # install-skip-build and would fail to stage its userspace files (e.g. /opt/gpu/config.sh).
+        # aks-gpu still independently re-validates the marker (kernel + driver_version +
+        # driver_kind) and falls back to a full build on any remaining mismatch (e.g. kernel drift).
         GPU_INSTALL_ACTION="install"
-        if [ -f "${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}" ]; then
+        GPU_DKMS_MARKER="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
+        if [ -f "$GPU_DKMS_MARKER" ] && \
+           [ "$(sed -n 's/^driver_kind=//p' "$GPU_DKMS_MARKER" | head -n1)" = "$NVIDIA_GPU_DRIVER_TYPE" ]; then
             GPU_INSTALL_ACTION="install-skip-build"
         fi
         # The driver image is normally pre-pulled into the VHD; only hit the registry when it is
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1211,20 +1211,30 @@ pullGPUDriverImage() {
 }
 
 installGPUDriverImage() {
-    retrycmd_if_failure 5 10 600 bash -c "$CTR_GPU_INSTALL_CMD $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG gpuinstall /entrypoint.sh install"
+    local gpuInstallAction="${1:-install}"
+    retrycmd_if_failure 5 10 600 bash -c "$CTR_GPU_INSTALL_CMD $NVIDIA_DRIVER_IMAGE:$NVIDIA_DRIVER_IMAGE_TAG gpuinstall /entrypoint.sh ${gpuInstallAction}"
 }
 
 configGPUDrivers() {
     if [ "$OS" = "$UBUNTU_OS_NAME" ]; then
         waitForContainerdReady || exit $ERR_GPU_DRIVERS_START_FAIL
         mkdir -p /opt/{actions,gpu}
+        # When the kernel module was pre-built into the VHD (build-only at image-bake time),
+        # a marker is present. Ask aks-gpu to skip the ~100s DKMS recompile and run only the
+        # device-dependent steps. aks-gpu independently re-validates the marker (kernel +
+        # driver_version + driver_kind) and falls back to a full build on any mismatch, so this
+        # is safe even after a kernel upgrade or on a shared VHD with a different driver kind.
+        GPU_INSTALL_ACTION="install"
+        if [ -f "${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}" ]; then
+            GPU_INSTALL_ACTION="install-skip-build"
+        fi
         # The driver image is normally pre-pulled into the VHD; only hit the registry when it is
         # actually missing so provisioning doesn't pay a redundant manifest/layer round trip.
         # Use containerd's native exact-name filter rather than text-matching `images ls` output.
         if [ -z "$(ctr -n k8s.io images ls -q "name==${NVIDIA_DRIVER_IMAGE}:${NVIDIA_DRIVER_IMAGE_TAG}")" ]; then
             logs_to_events "AKS.CSE.configGPUDrivers.pullGPUDriverImage" pullGPUDriverImage
         fi
-        logs_to_events "AKS.CSE.configGPUDrivers.installGPUDriverImage" installGPUDriverImage
+        logs_to_events "AKS.CSE.configGPUDrivers.installGPUDriverImage" installGPUDriverImage "$GPU_INSTALL_ACTION"
         ret=$?
         if [ "$ret" -ne 0 ]; then
             echo "Failed to install GPU driver, exiting..."
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1227,8 +1227,8 @@ configGPUDrivers() {
         # install-skip-build and would fail to stage its userspace files (e.g. /opt/gpu/config.sh).
         # aks-gpu still independently re-validates the marker (kernel + driver_version +
         # driver_kind) and falls back to a full build on any remaining mismatch (e.g. kernel drift).
-        GPU_INSTALL_ACTION="install"
-        GPU_DKMS_MARKER="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
+        local GPU_INSTALL_ACTION="install"
+        local GPU_DKMS_MARKER="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
         if [ -f "$GPU_DKMS_MARKER" ] && \
            [ "$(sed -n 's/^driver_kind=//p' "$GPU_DKMS_MARKER" | head -n1)" = "$NVIDIA_GPU_DRIVER_TYPE" ]; then
             GPU_INSTALL_ACTION="install-skip-build"
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
@@ -247,9 +247,11 @@ cleanUpPrebakedGPUDriver() {
     # depmod/initramfs refresh is required.
     rm -rf /var/lib/dkms/nvidia || true
     rm -f /lib/modules/*/updates/dkms/nvidia*.ko* 2>/dev/null || true
-    # aks-gpu relocates the userspace libs under GPU_DEST/lib64; on Ubuntu GPU_DEST=/usr/bin.
+    # aks-gpu stages the driver userspace libs under its container's GPU_DEST/lib64. NOTE: that is
+    # the aks-gpu *container's* GPU_DEST=/usr/bin (aks-gpu config.sh), NOT this CSE script's
+    # GPU_DEST=/usr/local/nvidia -- the prebake writes to /usr/bin, so the teardown clears /usr/bin.
     rm -rf /usr/bin/lib64 || true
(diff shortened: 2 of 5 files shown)


ganeshkumarashok:
Folded in the non-GPU teardown (commit a70a1ae) so this stays one PR.

**Why:** regular x86 CUDA shares its Ubuntu VHD with non-GPU nodes (no x86 GPU-dedicated VHD; only GB200 is). Pre-baking the driver leaves non-GPU nodes carrying an *installed*, DKMS-registered driver that recompiles `nvidia.ko` on every kernel patch. The existing `cleanUpGPUDrivers` only removed cached dirs, not an installed driver.

**Change:** `cleanUpGPUDrivers` (Ubuntu) now calls `cleanUpPrebakedGPUDriver` — `dkms remove` + drop `/usr/bin/lib64`, `nvidia.conf`, `/var/lib/dkms/nvidia`, marker. Gated on the prebake marker (no-op on non-prebaked VHDs → backward compatible). Module is never loaded on non-GPU nodes, so deregistration is safe. ShellSpec coverage added (3 examples pass); both shellcheck gates pass.

Open decision in-thread: sync (wired) vs async `systemd-run` teardown. GB200/300 left untouched.

ganeshkumarashok:
## ✅ End-to-end validation summary (real AgentBaker e2e, not just unit tests)

This feature was validated by baking **real prebaked VHDs** through the `[TEST All VHDs]` ADO pipeline (172842) and running the **GPU e2e suite** against them on real GPU hardware. A test ACR carried an `aks-gpu-cuda` image built from the unmerged Azure/aks-gpu#162 (build-only / install-skip-build), at the **production driver version `580.126.09`**.

### 🎯 Capstone: CUDA prebake proven on a real Tesla V100
`Test_Ubuntu2204_GPUNC` node CSE log (build `168578710`):
```
GPU_INSTALL_ACTION=install-skip-build              # driver_kind guard matched (cuda == cuda)
aks-gpu: using kernel module prebuilt in the VHD for kernel 5.15.0-1114-azure (recompile skipped)
NVIDIA-SMI 580.126.09   Driver Version: 580.126.09   CUDA Version: 13.0
0  Tesla V100-PCIE-16GB
```
Marker matched → **DKMS recompile skipped** → driver loaded → `nvidia-smi` healthy on the GPU. The ~80–150 s in-CSE compile is removed from the node-provisioning critical path.

### What was validated
| Area | Result | Evidence |
|------|--------|----------|
| **In-pipeline bake** (build-only) | ✅ both SKUs, cross-kernel | 2204 marker `kernel=5.15.0-1114-azure, driver_version=580.126.09, driver_kind=cuda`; 2404 `kernel=6.8.0-1059-azure` (builds `168578710`/`168580527`) |
| **`libc6-dev` toolchain** before bake | ✅ required & works | builder ships gcc/make but **not** libc6-dev → `nvidia-installer` can't compile without it |
| **CUDA skip-build** on real GPU | ✅ | V100 above; marker exact-match (kernel+version+kind) |
| **`driver_kind` guard (GRID safety)** | ✅ | GRID nodes (prod `aks-gpu-grid` image) correctly take a **full install**, not skip-build (a CUDA-baked marker must not short-circuit a GRID node → was exit 84 before the guard) |
| **non-GPU / `gpu-driver=none` teardown** | ✅ | `cleanUpPrebakedGPUDriver` removes nvidia binaries + fast DKMS deregister; `GPUNoDriver` + `CSE_CachedPerformance` green |
| **Backward compatibility** | ✅ | no marker → `install` (unchanged); boot-time fallback recompile intact (nodes get gcc/libc6-dev via installDeps) |
| **ShellSpec / `make generate`** | ✅ | 117 examples 0 failures; testdata no-drift; POSIX shellcheck clean |

### About the e2e "failures"
The GPU e2e run reported failures, but **none are caused by this PR**. Every GPU scenario that failed did so **only** on `validators.go ValidateNoFailedSystemdUnits` because `nvidia-cdi-refresh.{path,service}` were stuck in a `failed` state — a **pre-existing race introduced by nvidia-container-toolkit 1.18.0** (the toolkit auto-starts `nvidia-ctk cdi generate` before the driver libs are staged). Proof it's unrelated to prebake: GRID nodes on the **stock production** `aks-gpu-grid` image fail identically, and the units ship in the toolkit deb (not in this repo). This is tracked/fixed separately in **#8736** (a systemd drop-in that waits for the driver) — with that drop-in, the GPU scenarios are expected to go fully green. The remaining non-GPU failures (LocalDNSHostsPlugin, etc.) are unrelated flakes; the 2404 A100/H100 also hit the known SIG image-replication infra timeout (#8374).

### Traceability
- 2204 CUDA/V100 proof: build `168578710` · 2404 (kernel 6.8 bake + GRID guard): `168580527`
- Companion fix for the toolkit cdi-refresh race: #8736
- aks-gpu dependency: Azure/aks-gpu#162

_Note: the `cse_helpers.sh` test-ACR image override and the pipeline feature-flag enablement used to drive this validation are **test-only** and intentionally **not** part of this PR; the flag stays off by default until the aks-gpu image ships._

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1211,20 +1211,35 @@ pullGPUDriverImage() {
...
+        # When the kernel module was pre-built into the VHD (build-only at image-bake time),
+        # a marker is present. Ask aks-gpu to skip the ~100s DKMS recompile and run only the
+        # device-dependent steps -- but ONLY when the marker's driver_kind matches THIS node's
+        # driver (NVIDIA_GPU_DRIVER_TYPE). A CUDA-prebaked marker on a GRID node (or vice-versa)
+        # must request a full "install": the other driver image may not even support
+        # install-skip-build and would fail to stage its userspace files (e.g. /opt/gpu/config.sh).
+        # aks-gpu still independently re-validates the marker (kernel + driver_version +
+        # driver_kind) and falls back to a full build on any remaining mismatch (e.g. kernel drift).
+        GPU_INSTALL_ACTION="install"
+        GPU_DKMS_MARKER="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
```
xuexu6666: These two are assigned without `local`, so they leak into the global shell scope (bash treats any in-function assignment without `local` as global). The sibling `installGPUDriverImage` just above already uses `local gpuInstallAction="${1:-install}"` — suggest matching that here:

```sh
local GPU_INSTALL_ACTION="install"
local GPU_DKMS_MARKER="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
```

The later `GPU_INSTALL_ACTION="install-skip-build"` reassignment is unaffected. Not a functional bug (both are set unconditionally before use and passed explicitly as a positional arg), but CLAUDE.md calls this out: "use local variables rather than constants when their scoping allows for it" and "avoid using variables declared inside another function, even they are visible. It is hard to reason and might introduce subtle bugs."

runzhen (approved):
The aks-gpu repo should have the corresponding change to support "install-skip-build" command?

nvm, I saw the changes in https://github.com/Azure/aks-gpu/pull/162

awesomenix (approved):

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -737,6 +737,33 @@ if [ $OS = $UBUNTU_OS_NAME ] && [ "$(isARM64)" -ne 1 ]; then  # No ARM64 SKU wit
     cat << EOF >> ${VHD_LOGS_FILEPATH}
   - nvidia-cuda-driver=${NVIDIA_DRIVER_IMAGE_TAG}
 EOF
+
+  # Opt-in: pre-build the NVIDIA kernel module into the VHD so node provisioning skips the
```
Devinwong, line 741: question about this: the Opt-in here means if the feature flag FEATURE_FLAGS=NVIDIA_CUDA_PREBAKE is provided in the GPU SKU build pipeline, right?
Does that mean a non-GPU node could still use this GPU SKU VHD?

Devinwong: I don't fully understand the scenario. If that's the case, why not only GPU node using GPU SKU VHD?

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -223,12 +223,56 @@ removeNvidiaRepos() {
...
+    # aks-gpu relocates the userspace libs under GPU_DEST/lib64; on Ubuntu GPU_DEST=/usr/bin.
+    rm -rf /usr/bin/lib64 || true
+    # nvidia-installer also drops driver userspace BINARIES under GPU_DEST (=/usr/bin on Ubuntu).
+    # Remove them too so a non-GPU node looks genuinely driver-free: otherwise e.g. `nvidia-smi`
+    # remains on PATH and, with its libs (lib64) gone, errors instead of being "command not found".
+    for nvidiaBin in nvidia-smi nvidia-debugdump nvidia-persistenced nvidia-cuda-mps-control \
+                     nvidia-cuda-mps-server nvidia-modprobe nvidia-bug-report.sh nvidia-powerd \
+                     nvidia-ngx-updater nvidia-sleep.sh; do
+        rm -f "/usr/bin/${nvidiaBin}" || true
+    done
```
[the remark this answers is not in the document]
Devinwong: @ganeshkumarashok  Is this a valid question?

vhdbuilder/packer/install-dependencies.sh
ganeshkumarashok: >the Opt-in here means if the feature flag FEATURE_FLAGS=NVIDIA_CUDA_PREBAKE is provided in the GPU SKU build pipeline, right?

The flag will be set in the VHD build pipeline.

>Does that mean a non-GPU node could still use this GPU SKU VHD?

There is no separate GPU VHD today. GPU and non-GPU x86 nodes boot from the same Ubuntu VHD (the CUDA driver image is  already cached in it today). Only GB200 (arm64) has its own dedicated image.

So when  NVIDIA_CUDA_PREBAKE  is on, the prebaked module lands in that shared Ubuntu VHD, and a non-GPU node that boots from it inherits an installed driver it doesn't need. That's exactly why `cleanUpPrebakedGPUDrive`r  runs on non-GPU nodes — to remove the dead weight (and stop  nvidia.ko  rebuilding on every  │ kernel patch). It's a no-op on VHDs that weren't prebaked.
 
"Why not a GPU-only VHD?" is fair — it would avoid the non-GPU footprint entirely, but it's a bigger change (new build target + RP logic to pick it for GPU  │ pools), and higher storage costs with more SKUs to manage.

It's on the table as a future optimization; for now, shared VHD + teardown keeps the footprint minimal without splitting the image matrix.

Devinwong: Got it. Thanks for the explanations

PR #8697, opened 2026-06-12
feat: add secondary nic count in NBC and NetworkConfig

jumpinthefire:
See https://github.com/Azure/AgentBaker/pull/8642 for more details.
This PR adds a configuration field that AKS RP will set when non-delegated secondary NICs are provisioned within a node pool. When this field is set, node provision will configure and bring up secondary NICs found in the OS. This is to leave existing agent pools intact and only turn on secondary NICs when explicitly desired by the customer, using the RP API.

Both scripted and scriptless variants are set.


djsly (approved):

jumpinthefire:
Confirmed the same e2e failures occur on the main branch.

djsly (approved):

PR #8703, opened 2026-06-15
fix: if niether nbc cmd or aks node config exists, exit gracefully

awesomenix:
We automatically run aks node controller on all VHD, the only difference is that we check if check existence of configs to whether to proceed or not.

### DIFF
--- a/parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
+++ b/parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
@@ -22,6 +22,11 @@ log() {
 # this is to ensure that shellspec won't interpret any further lines below
 ${__SOURCED__:+return}
 
+if [ ! -f "$CONFIG_PATH" ] && [ ! -f "$NBC_CMD_PATH" ]; then
+    log "Gracefully exit aks-node-controller without provision config or nbc cmd"
+    exit 0
+fi
+
 if [ -f "$HOTFIX_JSON" ]; then
     log "Found ANC hotfix config at ${HOTFIX_JSON}; running download-hotfix"
     if "$BIN_PATH" download-hotfix; then
--- a/spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
+++ b/spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env shellspec
+
+Describe 'aks-node-controller-wrapper.sh'
+    SCRIPT="./parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh"
+
+    setup_wrapper_test() {
+        TEST_DIR="${SHELLSPEC_WORKDIR}/aks-node-controller-wrapper"
+        BIN_DIR="${TEST_DIR}/bin"
+        mkdir -p "$BIN_DIR"
+
+        cat >"${BIN_DIR}/hostname" <<'EOF'
+#!/bin/sh
+printf 'test-host\n'
+EOF
+        chmod +x "${BIN_DIR}/hostname"
+
+        cat >"${BIN_DIR}/cat" <<'EOF'
+#!/bin/sh
+if [ "$1" = "/etc/hostname" ]; then
+    printf 'test-host\n'
+else
+    /bin/cat "$@"
+fi
+EOF
+        chmod +x "${BIN_DIR}/cat"
+
+        cat >"${BIN_DIR}/logger" <<'EOF'
+#!/bin/sh
+exit 0
+EOF
+        chmod +x "${BIN_DIR}/logger"
+
+        export PATH="${BIN_DIR}:$PATH"
+        export TEST_DIR
+        export BIN_PATH="${TEST_DIR}/aks-node-controller"
+        export CONFIG_PATH="${TEST_DIR}/aks-node-controller-config.json"
+        export NBC_CMD_PATH="${TEST_DIR}/aks-node-controller-nbc-cmd.sh"
+    }
+
+    cleanup_wrapper_test() {
+        rm -rf "$TEST_DIR"
+        unset BIN_PATH CONFIG_PATH NBC_CMD_PATH TEST_DIR BIN_DIR
+    }
+
+    create_fake_aks_node_controller() {
+        cat >"$BIN_PATH" <<'EOF'
+#!/bin/sh
+printf '%s\n' "$@" >"${TEST_DIR}/args"
+exit 0
+EOF
+        chmod +x "$BIN_PATH"
+    }
+
+    BeforeEach setup_wrapper_test
+    AfterEach cleanup_wrapper_test
+
+    It 'exits successfully without invoking aks-node-controller when config and nbc cmd are absent'
+        When run bash "$SCRIPT"
+        The status should be success
+        The output should include "Gracefully exit aks-node-controller without provision config or nbc cmd"
+        The output should not include "Spawned aks-node-controller"
+    End
+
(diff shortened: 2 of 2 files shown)


parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
```diff
@@ -39,6 +39,11 @@ else
 fi
 
 command=("$BIN_PATH" provision)
+if [ ! -f "$CONFIG_PATH" ] && [ ! -f "$NBC_CMD_PATH" ]; then
+    log "Gracefully exit aks-node-controller without provision config or nbc cmd"
+    exit 0
```
timmy-wright: 🔴 **[BLOCKER] Provision sentinel not written on early exit**

This `exit 0` bypasses ANC writing `provision.complete`/`provision.json`. Any `provision-wait` that polls for those files now hangs until timeout rather than fast-failing, while `aks-node-controller.service` reports clean success (`RemainAfterExit=yes`, no `Restart=`).

A real config-delivery failure is now indistinguishable from an intentional old-VHD skip.

**Options:**
1. Write the sentinel files before exiting here (if that is safe to do without a full provision run)
2. Confirm and document that `provision-wait` is already guarded against ANC being skipped
3. Exit non-zero so upstream callers can distinguish skip from failure (if that is acceptable)

timmy-wright (commented):
## Squad Code Review

**Reviewers:** Simon (code defects), Wash (tests), Book (resilience), Zoe (security)

### Summary

| Reviewer | Verdict |
|----------|---------|
| 🛠️ Book | 🔴 BLOCKER — `exit 0` skips sentinel write; `provision-wait` will timeout instead of fast-fail |
| 🔬 Simon | 🟡 CONCERN — early-exit guard fires after hotfix download, not before |
| 🧪 Wash | 🟡 CONCERN — missing `config-only` and `nbc-cmd-only` test cases; hotfix path not isolated in fixture |
| 🔒 Zoe | 🟢 PASS — no security issues introduced |

### Core issue (Book + Zoe agree)

`exit 0` bypasses ANC writing `provision.complete`/`provision.json`. The `provision-wait` path downstream still blocks waiting for those sentinel files. With `RemainAfterExit=yes` and no `Restart=` on `aks-node-controller.service`, this means:

- `aks-node-controller.service` → reports **success**
- `provision-wait` → **hangs until timeout**

A real config-delivery failure (not just a planned old-VHD skip) is now indistinguishable from a successful skip. The intent is right; the missing piece is either writing the sentinel on early exit, or confirming that `provision-wait` is guarded against ANC being skipped.

See inline comments for specifics.

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
```diff
@@ -39,6 +39,11 @@ else
 fi
 
 command=("$BIN_PATH" provision)
+if [ ! -f "$CONFIG_PATH" ] && [ ! -f "$NBC_CMD_PATH" ]; then
```
timmy-wright: 🟡 **[CONCERN] Guard fires after hotfix download, not before**

If `HOTFIX_JSON` exists, `"$BIN_PATH" download-hotfix` runs before this check. A node with neither `CONFIG_PATH` nor `NBC_CMD_PATH` can still invoke the ANC binary, hit network/disk, or emit a binary failure log before exiting 0 — the exit is not as "graceful" as intended.

**Fix:** Move this guard above the hotfix block (before line 25), or gate the hotfix download on the same file-presence condition.

spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
```diff
@@ -0,0 +1,72 @@
...
+
+    cleanup_wrapper_test() {
+        rm -rf "$TEST_DIR"
+        unset BIN_PATH CONFIG_PATH NBC_CMD_PATH TEST_DIR BIN_DIR
+    }
+
+    BeforeEach setup_wrapper_test
+    AfterEach cleanup_wrapper_test
+
+    It 'exits successfully without invoking aks-node-controller when config and nbc cmd are absent'
```
timmy-wright, line 57: 🟡 **[CONCERN] `HOTFIX_JSON` not isolated in fixture**

The setup does not unset or stub `HOTFIX_JSON`. If the environment has a `HOTFIX_JSON` at the default path, the hotfix download block runs before the early-exit guard, potentially invoking `BIN_PATH` — which does not exist in this fixture — before the script reaches the guard.

**Fix:** Add `export HOTFIX_JSON="${TEST_DIR}/nonexistent-hotfix.json"` (or equivalent) in `setup_wrapper_test` to ensure the hotfix path is inert during this test.

spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
```diff
@@ -0,0 +1,72 @@
...
+        When run bash "$SCRIPT"
+        The status should be success
+        The output should include "Gracefully exit aks-node-controller without provision config or nbc cmd"
+        The output should not include "Spawned aks-node-controller"
+    End
```
timmy-wright, line 62: 🟡 **[CONCERN] Missing `config-only` and `nbc-cmd-only` test cases**

The suite covers `none` (both absent) and `both` (both present) but not:
- `CONFIG_PATH` present, `NBC_CMD_PATH` absent
- `NBC_CMD_PATH` present, `CONFIG_PATH` absent

A regression that flips `&&` to `||` in the guard condition would go completely undetected by the current suite. Please add both single-file cases.

spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
```diff
@@ -0,0 +1,72 @@
...
+printf '%s\n' "$@" >"${TEST_DIR}/args"
+exit 0
+EOF
+        chmod +x "$BIN_PATH"
+
+        When run bash "$SCRIPT"
+        The status should be success
+        The output should include "Launching aks-node-controller with config ${CONFIG_PATH}"
+        The output should include "Launching aks-node-controller with nbc cmd ${NBC_CMD_PATH}"
+        The contents of file "${TEST_DIR}/args" should include "provision"
```
timmy-wright: 🔵 **[MINOR] Assertion too loose**

`should include "provision"` is satisfied by `--provision-config=...` as well, so this assertion does not actually verify the positional `provision` subcommand arg independently.

**Fix:** Assert the exact first line/arg, e.g. `The first line of file "${TEST_DIR}/args" should eq "provision"`, to distinguish the subcommand from the flag values.

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
awesomenix: Fixed.

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
awesomenix: This is intentional

spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
awesomenix: fixed

spec/parts/linux/cloud-init/artifacts/aks_node_controller_wrapper_spec.sh
awesomenix: fixed

timmy-wright (approved):

PR #8710, opened 2026-06-15
fix: fix artifact arm64 params, anc hotfix to handle boothook

awesomenix:
- Fixed Test_Ubuntu2404_ArtifactStreaming_ARM64 to actually enable artifact streaming for ARM64 nodes, so overlaybd/containerd validators match the scenario intent.
 - Updated customData write file injection to handle #cloud-boothook scriptless NBC customData, not only gzipped write_files: cloud-config.
 - This lets Test_Ubuntu2204_ANCHotfix_BinarySelection pre-seed the ANC hotfix JSON/binary before aks-node-controller.service starts, so wrapper binary selection can detect the hotfix path.
 
 Fixes tests failed here: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=168054626&view=logs&j=10f66467-59e5-53bc-0c8a-42cc89770de4&t=85db0a6d-4633-56aa-cb86-ea559630d4e6

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -763,7 +763,7 @@ func Test_Ubuntu2204_ANCHotfix_BinarySelection(t *testing.T) {
 				},
 			},
 			BootstrapConfigMutator: func(_ *Cluster, nbc *datamodel.NodeBootstrappingConfiguration) {
-				nbc.EnableScriptlessCSECmd = true
+				nbc.EnableScriptlessNBCCSECmd = true
 			},
 			Validator: func(ctx context.Context, s *Scenario) {
 				// Wrapper found the pre-seeded hotfix binary and selected it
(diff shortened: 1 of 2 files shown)


e2e/scenario_test.go
```diff
@@ -1389,6 +1389,11 @@ func Test_Ubuntu2404_ArtifactStreaming_ARM64(t *testing.T) {
 			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
 				vmss.SKU.Name = to.Ptr("Standard_D2pds_V5")
 			},
+			BootstrapConfigMutator: func(_ *Cluster, nbc *datamodel.NodeBootstrappingConfiguration) {
+				nbc.EnableArtifactStreaming = true
+				nbc.AgentPoolProfile.VMSize = "Standard_D2pds_V5"
```
djsly, line 1394: we hardcode those skus ? we don't have a const/var ?

djsly (approved):

e2e/scenario_test.go
awesomenix: Not for arm64 i see. We can probably take it as input?

mxj220 (approved):

PR #8711, opened 2026-06-15
fix: prevent CSE hang when curl verbose output blocks on unstable disks

pdamianov-dev:
- Redirect CURL_OUTPUT to /dev/shm (tmpfs) instead of /tmp to avoid blocking on unstable OS disks. /dev/shm is kernel-mounted tmpfs (CONFIG_TMPFS=y on all Azure images) across all VM SKUs including CVM and ARM64.
- Add --max-time to curl in _retry_file_curl_internal so curl enforces its own deadline even if shell timeout cannot deliver signals.
- Add -k 5 to timeout in _retrycmd_internal and _retry_file_curl_internal to escalate to SIGKILL if SIGTERM is ignored (e.g. process in D-state).

[Bug 36680094](https://msazure.visualstudio.com/CloudNativeCompute/_workitems/edit/36680094): [Repair Item] Improve CSE script to handle curl timeouts and prevent blocking by redirecting verbose output away from unstable disks and enforcing strict timeout with forced kill signals.
Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -399,7 +399,7 @@ _retry_file_curl_internal() {
             fi
         fi
 
-        timeout $effectiveTimeout curl -fsSLv $url -o $filePath > $CURL_OUTPUT 2>&1
+        timeout -k 5 $effectiveTimeout curl --max-time $effectiveTimeout -fsSLv $url -o $filePath > $CURL_OUTPUT 2>&1
         if [ "$?" -ne 0 ]; then
             cat $CURL_OUTPUT
         fi
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -679,6 +679,36 @@ func Test_Ubuntu2204_Early_Failure_Scriptless(t *testing.T) {
 	})
 }
 
+// Test_Ubuntu2204_CSETimeoutOnUnreachableDownload validates that the per-curl timeout (-k 5)
+// in cse_helpers.sh _retry_file_curl_internal properly kills hung download attempts.
+// It sets CustomKubeBinaryURL to a non-routable IP (RFC 5737 TEST-NET-1) which causes curl
+// to hang until the timeout fires, and sets a short CSETimeout so the test completes quickly.
+// The global timeout in cse_start.sh (timeout -k5s) kills the entire provisioning script.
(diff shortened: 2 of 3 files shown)


parts/linux/cloud-init/artifacts/cse_install.sh
```diff
@@ -25,7 +25,14 @@ MANIFEST_FILEPATH="/opt/azure/manifest.json"
 COMPONENTS_FILEPATH="/opt/azure/components.json"
 VHD_LOGS_FILEPATH="/opt/azure/vhd-install.complete"
 MAN_DB_AUTO_UPDATE_FLAG_FILEPATH="/var/lib/man-db/auto-update"
-CURL_OUTPUT=/tmp/curl_verbose.out
+# Prefer /dev/shm (tmpfs) for verbose logs to avoid blocking on unstable OS disks.
+# /dev/shm is kernel-mounted tmpfs (CONFIG_TMPFS=y on all Azure images) and available on
+# all VM SKUs including CVM and ARM64. Fallback to /tmp for non-Azure environments.
+if [ -d "/dev/shm" ] && [ -w "/dev/shm" ]; then
+    CURL_OUTPUT=/dev/shm/curl_verbose.out
```
cameronmeissner: can we log something out when /dev/shm is selected?

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -203,7 +203,14 @@ export DOCKER_GPU_INSTALL_CMD="docker run --privileged --net=host --pid=host -v
 APT_CACHE_DIR=/var/cache/apt/archives/
 PERMANENT_CACHE_DIR=/root/aptcache/
 EVENTS_LOGGING_DIR=/var/log/azure/Microsoft.Azure.Extensions.CustomScript/events/
-CURL_OUTPUT=/tmp/curl_verbose.out
+# Prefer /dev/shm (tmpfs) for curl verbose logs to avoid blocking on unstable OS disks.
+# /dev/shm is kernel-mounted tmpfs (CONFIG_TMPFS=y on all Azure images) and available on
+# all VM SKUs including CVM and ARM64. Fallback to /tmp for non-Azure environments.
+if [ -d "/dev/shm" ] && [ -w "/dev/shm" ]; then
```
awesomenix: can we use `mktemp`?

pdamianov-dev:
It was determined that this PR and the associated work item were not needed to address the issue and did not fully match the result of the linked ticket

PR #8712, opened 2026-06-15
fix: vhdbuilder phase 2.5 should still use nbccsecmd hack

lilypan26:
For phase 2.5, we should still use nbccsecmd hack, not scriptless compilation

Fixes #

### DIFF
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -2670,7 +2670,7 @@ func ValidateScriptlessNBCCSECmd(ctx context.Context, s *Scenario) {
 	nbc := s.Runtime.NBC
 	if nbc != nil && nbc.EnableScriptlessNBCCSECmd && s.VHD.SupportsScriptless() {
 		fileNameToCheck := "/opt/azure/containers/aks-node-controller-nbc-cmd.sh"
-		if s.Runtime.AKSNodeConfig != nil || (!config.Config.DisableScriptLessCompilation && !s.Tags.NetworkIsolated) {
+		if !config.Config.DisableScriptLessCompilation && !s.Tags.NetworkIsolated {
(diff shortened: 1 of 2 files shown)


djsly (approved):

e2e/validators.go
```diff
@@ -2670,7 +2670,7 @@ func ValidateScriptlessNBCCSECmd(ctx context.Context, s *Scenario) {
 	nbc := s.Runtime.NBC
 	if nbc != nil && nbc.EnableScriptlessNBCCSECmd && s.VHD.SupportsScriptless() {
 		fileNameToCheck := "/opt/azure/containers/aks-node-controller-nbc-cmd.sh"
-		if !config.Config.DisableScriptLessCompilation && !s.Tags.NetworkIsolated {
+		if s.Runtime.AKSNodeConfig != nil || (!config.Config.DisableScriptLessCompilation && !s.Tags.NetworkIsolated) {
```
awesomenix: i think there are 2 cases

1. AgentBaker E2E - Where we compile the AKS node controller and this uses -hach.sh
2. Regular PR E2E - Where we use production scenario because we us whatever is on disk

The check before seems fine isnt it?

e2e/vmss.go
```diff
@@ -376,7 +376,7 @@ func createVMSSModel(ctx context.Context, s *Scenario) armcompute.VirtualMachine
 		}
 
 		customData = func() string {
-			if config.Config.DisableScriptLessCompilation {
+			if config.Config.DisableScriptLessCompilation && !s.Runtime.NBC.EnableScriptlessNBCCSECmd {
```
awesomenix: hmm seems odd, are we saying this is old scriptless?

awesomenix (approved):

PR #8713, opened 2026-06-15
feat(dra):  DRA driver installation

runzhen:
Initial checkin for dra-driver-nvidia-gpu 

Below items will be handled in another PR
1. add "enable DRA" filed in proto buffer. 
2. support AzureLinux

[unit_test.txt](https://github.com/user-attachments/files/29019796/unit_test.txt)
```

Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  37s   default-scheduler  Successfully assigned default/pod to 3718-2026-06-16-ubuntu2404dradrivernvidiagpurunningdefaul000000
  Normal  Created    13s   kubelet            spec.containers{ctr0}: Created container: ctr0
  Normal  Started    13s   kubelet            spec.containers{ctr0}: Started container ctr0
  Normal  Pulled     13s   kubelet            spec.containers{ctr1}: Container image "mcr.microsoft.com/azuredocs/samples-tf-mnist-demo:gpu" already present on machine
  Normal  Created    13s   kubelet            spec.containers{ctr1}: Created container: ctr1
  Normal  Started    13s   kubelet            spec.containers{ctr1}: Started container ctr1

k logs -f pod --all-containers
GPU 0: NVIDIA A10-4Q (UUID: GPU-920b9300-69c2-11f1-9f17-62dd377f7a08)
GPU 0: NVIDIA A10-4Q (UUID: GPU-920b9300-69c2-11f1-9f17-62dd377f7a08)
```

Fixes #


parts/common/components.json
```diff
@@ -1976,6 +1976,40 @@
...
+        "azurelinux": {
+          "v3.0": {
+            "versionsV2": [
+              {
+                "renovateTag": "RPM_registry=https://packages.microsoft.com/azurelinux/3.0/prod/cloud-native/x86_64/repodata, name=dra-driver-nvidia-gpu, os=azurelinux, release=3.0",
+                "latestVersion": "0.4.0-1.azl3"
+              }
+            ]
+          }
+        }
```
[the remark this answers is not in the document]
runzhen: will support az linux later

parts/common/components.json
```diff
@@ -1976,6 +1976,40 @@
         }
       }
     },
+    {
```
sulixu, line 1979: the first PR should land with only the components.json change, so that future references in the e2es to the .deb packages can be downloaded successfully.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -583,6 +568,33 @@ function nodePrep {
 
     checkServiceHealth kubelet || exit $ERR_KUBELET_FAIL
 
+    # Defer Configure managed GPU experience after `checkServiceHealth kubelet`
```
sulixu, line 571: are you deferring the non-DRA experience as well? 
so this means fully managed gpu experience will start after kubelet, i feel only dra needs kubelet up as prerequisite, kubelet-up is not required to install nvidia device-plugin package

runzhen: yes, just like the comment said, dcgm, dcgm-exporter do not require kubeconfig but are kept here so the whole function stays in one place.

sulixu (commented):
lets break down this into smaller PRs, i feel 

kubelet-up is not required to install the NVIDIA device-plugin package, so delay all managed-GPU install/start until after kubelet is not necessary.
kubelet-up maybe required to install DRA, so that `/var/lib/kubelet/kubeconfig` exists before install DRA. 
Defer only the DRA branch worth a separate PR.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -191,8 +201,20 @@ managedGPUPackageList() {
...
     # Ensure kubelet device-plugins directory exists BEFORE package installation
     mkdir -p /var/lib/kubelet/device-plugins
+    mkdir -p /var/lib/kubelet/plugins_registry
+    mkdir -p /var/lib/kubelet/plugins
 
-    for packageName in $(managedGPUPackageList); do
+    packageList="$(managedGPUPackageList)"
+
+    if [ "${ENABLE_MANAGED_GPU_EXPERIENCE_DRA}" = "true" ]; then
+        packageList="$(managedGPUPackageListDRA)"
```
surajssd, line 210: Why not just inject device plugin or dra-driver-nvidia-gpu as needed, instead of repeat the dcgm packages in two places like `managedGPUPackageListDRA` and `managedGPUPackageList`?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1809,13 +1809,67 @@ configureManagedGPUExperience() {
         # EnableManagedGPUExperience is mutable, so services may have been
         # installed on a previous CSE run. Stop them if they exist.
         logs_to_events "AKS.CSE.stop.nvidia-device-plugin" "systemctlDisableAndStop nvidia-device-plugin"
+        logs_to_events "AKS.CSE.stop.dra-driver-nvidia-gpu" "systemctlDisableAndStop dra-driver-nvidia-gpu"
```
surajssd, line 1812: Isn't this needed only when `ENABLE_MANAGED_GPU_EXPERIENCE_DRA` is set?

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -583,6 +568,33 @@ function nodePrep {
...
+    # The dra-driver-nvidia-gpu service needs that /var/lib/kubelet/kubeconfig, so it must be started
+    # only after kubelet is up. The other managed GPU services (device-plugin, dcgm,
+    # dcgm-exporter) do not require kubeconfig but are kept here so the whole flow
+    # stays in one place.
+    if [ "${GPU_NODE}" = "true" ] && [ "${skip_nvidia_driver_install}" != "true" ]; then
```
surajssd, line 576: Note, out of scope of this PR, we should stop calling this as a `skip_nvidia_driver_install` but simply `skip_driver_install`. Right now, publicly the API is skipping gpu driver installation and makes no distinction between AMD or Nvidia.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -191,8 +201,20 @@ managedGPUPackageList() {
 installNvidiaManagedExpPkgFromCache() {
     # Ensure kubelet device-plugins directory exists BEFORE package installation
     mkdir -p /var/lib/kubelet/device-plugins
+    mkdir -p /var/lib/kubelet/plugins_registry
```
surajssd, line 204: Add similar changes to `mariner/cse_install_mariner.sh`

**Azure Linux can cache the package but cannot install/start it.** `components.json` + `install-dependencies.sh` download the RPM for AzLinux, but `managedGPUPackageListDRA` and the `plugins_registry`/`plugins` mkdirs live only in `ubuntu/cse_install_ubuntu.sh`. A DRA-enabled AzLinux node would start the service with no `gpu-kubelet-plugin` installed → guaranteed failure. Latent today (flag hardcoded off), but recommend an explicit non-Ubuntu guard rather than relying on the flag never being set.

e2e/scenario_gpu_managed_experience_test.go
```diff
@@ -783,3 +783,40 @@ func Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG_Mixed(t *testing.T) {
...
+				require.NoError(t, err, "creating AKS VM extension")
+				vmss.Properties = addVMExtensionToVMSS(vmss.Properties, extension)
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				containerdVersions := components.GetExpectedPackageVersions("containerd", "ubuntu", "r2404")
+				runcVersions := components.GetExpectedPackageVersions("runc", "ubuntu", "r2404")
+				ValidateContainerd2Properties(ctx, s, containerdVersions)
+				ValidateRuncVersion(ctx, s, runcVersions)
+				ValidateContainerRuntimePlugins(ctx, s)
+				ValidateDraDriverNvidiaGpuServiceRunning(ctx, s)
```
surajssd, line 818: **e2e validates liveness, not function.** `ValidateDraDriverNvidiaGpuServiceRunning` only checks `systemctl is-active`/`is-enabled` — a service can be active while DRA is misconfigured. No test schedules a Pod with a `ResourceClaim` or checks a published ResourceSlice / plugin socket. Please deploy a sample pod and see if it comes up and uses the GPU.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1856,50 +1910,33 @@ EOF
...
 
-    # 3. Start the nvidia-dcgm-exporter service.
-    # Create systemd drop-in directory for nvidia-dcgm-exporter service
-    DCGM_EXPORTER_OVERRIDE_DIR="/etc/systemd/system/nvidia-dcgm-exporter.service.d"
-    mkdir -p "${DCGM_EXPORTER_OVERRIDE_DIR}"
+    # TODO: get version from components.json
+    IMAGE_NAME="mcr.microsoft.com/oss/v2/nvidia/dra-driver-nvidia-gpu:v0.4.0-1"
+
+    DRA_DRIVER_SVC_DIR="/etc/systemd/system/"
+    tee "${DRA_DRIVER_SVC_DIR}/dra-driver-nvidia-gpu.service" > /dev/null <<EOF
```
surajssd, line 1925: This will generate double slashes in the final file path

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1809,13 +1809,67 @@ configureManagedGPUExperience() {
...
         logs_to_events "AKS.CSE.stop.nvidia-dcgm" "systemctlDisableAndStop nvidia-dcgm"
         logs_to_events "AKS.CSE.stop.nvidia-dcgm-exporter" "systemctlDisableAndStop nvidia-dcgm-exporter"
         rm -f "${managed_gpu_marker}"
     fi
 }
 
 startNvidiaManagedExpServices() {
+    # 1. Start device plugin or DRA driver
+    if [ "${ENABLE_MANAGED_GPU_EXPERIENCE_DRA}" = "true" ]; then
+        startDRADriverNvidiaGpu
```
surajssd, line 1822: **Mode-switch is not idempotent** If both `ENABLE_MANAGED_GPU_EXPERIENCE` and `..._DRA` are true, the enable-branch starts only the DRA driver (`startDRADriverNvidiaGpu` wins). A cluster transitioning `device-plugin` → `DRA` could leave both active. No e2e covers the both-true case.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
surajssd: Also this is not cleaned up right now. DRA package can be left behind.

vhdbuilder/packer/install-dependencies.sh
```diff
@@ -556,6 +556,14 @@ while IFS= read -r p; do
         echo "  - ${K8S_DEVICE_PLUGIN_PKG} version ${version}" >> ${VHD_LOGS_FILEPATH}
       done
       ;;
+    "dra-driver-nvidia-gpu")
+      for version in ${PACKAGE_VERSIONS[@]}; do
+        if [ "${OS}" = "${UBUNTU_OS_NAME}" ] || isAzureLinux "$OS"; then
+          downloadPkgFromVersion "dra-driver-nvidia-gpu" "${version}" "${downloadDir}"
+        fi
+        echo "  - dra-driver-nvidia-gpu version ${version}" >> ${VHD_LOGS_FILEPATH}
```
surajssd, line 564: Shouldn't this echo be inside the if block? Why log if something is not downloaded?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1856,50 +1910,33 @@ EOF
...
+After=kubelet.service
 
-    # Create drop-in file to override service configuration
-    tee "${DCGM_EXPORTER_OVERRIDE_DIR}/10-aks-override.conf" > /dev/null <<EOF
 [Service]
-# Remove file-based logging - let systemd handle logs
-StandardOutput=journal
-StandardError=journal
-# Change default port from 9400 to 19400 so that it does not conflict with user installed dcgm-exporter
 ExecStart=
```
surajssd, line 1930: This empty execstart is only needed if you are creating a drop-in config not in the service itself.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1856,50 +1910,33 @@ EOF
...
-    # Create systemd drop-in directory for nvidia-dcgm-exporter service
-    DCGM_EXPORTER_OVERRIDE_DIR="/etc/systemd/system/nvidia-dcgm-exporter.service.d"
-    mkdir -p "${DCGM_EXPORTER_OVERRIDE_DIR}"
+    # TODO: get version from components.json
+    IMAGE_NAME="mcr.microsoft.com/oss/v2/nvidia/dra-driver-nvidia-gpu:v0.4.0-1"
+
+    DRA_DRIVER_SVC_DIR="/etc/systemd/system/"
+    tee "${DRA_DRIVER_SVC_DIR}/dra-driver-nvidia-gpu.service" > /dev/null <<EOF
+[Unit]
+After=kubelet.service
```
surajssd, line 1927: Why are we generating the systemd service by hand doesn't the package already ship this?

surajssd (changes requested):

PR #8723, opened 2026-06-16
fix: dont use hack stuff for ANC hotfix test

awesomenix:
-hack stuff doesnt have things to download hotfix, dont use it for this test

### DIFF
--- a/e2e/vmss.go
+++ b/e2e/vmss.go
@@ -406,7 +406,7 @@ func createVMSSModel(ctx context.Context, s *Scenario) armcompute.VirtualMachine
 	} else {
 		cse = nodeBootstrapping.CSE
 		customData = nodeBootstrapping.CustomData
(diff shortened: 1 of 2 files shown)


e2e/vmss.go
```diff
@@ -473,6 +473,10 @@ func createVMSSModel(ctx context.Context, s *Scenario) armcompute.VirtualMachine
 	return model
 }
 
+func enableScriptlessCCompilation(s *Scenario) bool {
+	return s.Runtime.NBC.EnableScriptlessNBCCSECmd && len(s.Config.CustomDataWriteFiles) <= 0 && !config.Config.DisableScriptLessCompilation && !s.Tags.NetworkIsolated && !s.Runtime.NBC.PreProvisionOnly
```
lilypan26, line 477: nit: i would leave out `s.Runtime.NBC.EnableScriptlessNBCCSECmd` here, and instead do `if s.Runtime.NBC.EnableScriptlessNBCCSECmd && enableScriptlessCCompilation(s)`. I think this makes more sense and also you can reuse it `validators.go`

lilypan26 (approved):

PR #8767, opened 2026-06-23
refactor: write ANC hotfix.json via boothook and rename target_version to scripts_version

Devinwong:
## Summary

Write the ANC hotfix config (`aks-node-controller-hotfix.json`) directly from the boothook
instead of via a `write_files` entry in `nodecustomdata.yml`. This eliminates the double-parse
of `nodecustomdata.yml` in `download-hotfix`, and gates an optional script hotfix payload by
`scripts_version`.

### Files / source of truth

- **Source of truth (committed, edited by operators):** `hotfix/anc-hotfix-version.json`
  - `version` = ANC binary hotfix version to download/install
  - `scripts_version` = ANC/VHD version base that should receive the script hotfix `write_files`
- **Embedded config read by `baker.go`:** `parts/hotfix/aks-node-controller-hotfix.json`
  - Generated from `hotfix/anc-hotfix-version.json` by `hotfix/anc_hotfix_generate.py`.
  - Embedded into the ABSVC binary via `parts.Templates` and shipped in custom data as
    `/opt/azure/containers/aks-node-controller-hotfix.json`.
- **Embedded script hotfix payload (optional):** `parts/hotfix/anc-scripts-hotfix-files.yml`
  - Generated by `hotfix/hotfix_generate.py`; shipped as
    `/opt/azure/containers/anc-scripts-hotfix-files.yml`.

Both generated files are only committed on `official/*` hotfix branches. On normal builds they
are absent, which `baker.go` treats as "no active hotfix" (ships nothing).

## Flow

```mermaid
flowchart TD
    A["ABSVC / baker.go\nreads parts/hotfix/aks-node-controller-hotfix.json\n(embedded at build time)"]
    B["Boothook\nwrite nodecustomdata.yml\nwrite aks-node-controller-hotfix.json\nwrite anc-scripts-hotfix-files.yml (if active)"]
    C["download-hotfix\nread aks-node-controller-hotfix.json"]
    D{"ANC binary\nhotfix applies?"}
    E["Download & stage\nhotfix binary"]
    F["Skip binary download"]
    S{"scripts_version\nmatches running version?"}
    T["apply script hotfix\nwrite_files"]
    U["skip script hotfix"]
    G["provision\nread nodecustomdata.yml\napply write_files\nrun NBC cmd"]

    A --> B --> C --> D
    D -- Yes --> E --> S
    D -- No --> F --> S
    S -- Yes --> T --> G
    S -- No --> U --> G
```

Note: the script hotfix `write_files` are applied inside `download-hotfix` (via
`applyScriptHotfix`), which the wrapper runs **before** `provision`. `provision` then
materializes the main `nodecustomdata.yml` and runs the NBC command.

## Changes

| File | Change |
|------|--------|
| `pkg/agent/baker.go` | Add `loadHotfixConfigJSON()` and `getLinuxNodeScriptsHotfixFilesPayload()`; ship the gzipped hotfix config + optional script payload via the shared `gzipFileBlockFmt` boothook block / Flatcar Ignition entries. Absent embed files are treated as "no active hotfix". |
| `hotfix/anc-hotfix-version.json` | Source-of-truth config (`{version, scripts_version}`), edited by operators on `official/*`. |
| `parts/hotfix/aks-node-controller-hotfix.json` | Generated embed (not committed on `main`); produced by `hotfix/anc_hotfix_generate.py`. |
| `parts/hotfix/anc-scripts-hotfix-files.yml` | Generated script-hotfix `write_files` payload (not committed on `main`); produced by `hotfix/hotfix_generate.py`. |
| `hotfix/anc_hotfix_generate.py` / `hotfix/hotfix_generate.py` | Emit standalone payload files (no marker injection into `nodecustomdata.yml`); delete the generated file when no hotfix is active. |
| `aks-node-controller/hotfix.go` | `download-hotfix` reads the config, conditionally downloads the binary, and applies the script hotfix `write_files` (gated by `scripts_version`). `target_version` ΓåÆ `scripts_version`. |
| `aks-node-controller/nodecustomdata.go` | `applyWriteFiles` no-ops on a missing (optional) payload; `applyNodeCustomData` stays strict (errors on missing required `nodecustomdata.yml`). |
| `.github/workflows/*hotfix-generate.yml` | Regenerate + commit/update/delete the generated embed files on hotfix PRs. |
| Tests / snapshots | Updated renames; regenerated test data. |


aks-node-controller/nodecustomdata_test.go
```diff
@@ -79,3 +79,37 @@ write_files:
...
+  content: |
+    normal-content
+# ---- hotfix: auto-generated by hotfix-generate GH Action ----
+- path: %s
+  permissions: "0644"
+  owner: root
+  content: |
+    hotfix-content
+# ---- end hotfix ----
+`, normalPath, hotfixPath)
```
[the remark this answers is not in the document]
Devinwong: Addressed in the latest update. `TestApplyNodeCustomDataWithHotfixConfig_TargetVersionMismatchSkipsScriptHotfixBlock` now places the hotfix block in the middle of `nodecustomdata.yml`, includes an additional `write_files` entry after `# ---- end hotfix ----`, and asserts that post-block entry is still applied when `target_version` mismatches. This specifically guards against newline-trimming regressions in `stripScriptHotfixBlock` that could merge adjacent YAML/content lines.

aks-node-controller/nodecustomdata_test.go
```diff
@@ -79,3 +79,109 @@ write_files:
 	require.NoError(t, err)
 	assert.Equal(t, "0", result.ExitCode)
 }
+
+func TestApplyNodeCustomDataWithHotfixConfig_TargetVersionMismatchSkipsScriptHotfixBlock(t *testing.T) {
+	tempDir := t.TempDir()
+	normalPath := filepath.Join(tempDir, "normal.txt")
+	hotfixPath := filepath.Join(tempDir, "hotfix.txt")
```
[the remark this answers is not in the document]
Devinwong: Addressed in latest commit: this test now sets `Version = "202604.01.1"` (with restore defer) so the skip path is validated via an explicit target_version mismatch rather than an unparsable default version.

hotfix/anc_hotfix_generate.py
```diff
@@ -122,10 +139,10 @@ def remove_hotfix():
...
-    version = read_hotfix_version()
-    if version:
-        inject(version)
-        print(f"\nDone. Injected ANC hotfix version {version}.")
+    payload = read_hotfix_config()
+    if payload:
+        inject(payload)
+        print(f"\nDone. Injected ANC hotfix config {payload}.")
     else:
         # No version set — remove any stale hotfix entry
```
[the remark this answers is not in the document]
Devinwong: Updated in latest commit: the comment now reflects both supported semantics (`version` or `target_version`) before stale entry cleanup.

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
```diff
@@ -27,13 +26,11 @@ if [ ! -f "$CONFIG_PATH" ] && [ ! -f "$NBC_CMD_PATH" ]; then
...
-        log "ANC download-hotfix completed; binary selection follows"
-    else
-        log "ANC download-hotfix failed; binary selection follows"
-    fi
+log "Running ANC download-hotfix pre-check"
+if "$BIN_PATH" download-hotfix; then
+    log "ANC download-hotfix completed; binary selection follows"
+else
+    log "ANC download-hotfix failed; binary selection follows"
 fi
```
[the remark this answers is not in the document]
Devinwong: Already covered by current spec updates: ShellSpec now asserts `download-hotfix` runs before `provision` when either config or nbc cmd exists, and confirms it is skipped on early-exit when both are absent.

aks-node-controller/nodecustomdata.go
```diff
@@ -54,6 +87,25 @@ func applyNodeCustomData(path string) error {
...
+		return content
+	}
+	end := start + endOffset + len(scriptHotfixEndMarker)
+	if end < len(content) && content[end] == '\n' {
+		end++
+	}
+	if start > 0 && content[start-1] == '\n' {
+		start--
+	}
+	return content[:start] + content[end:]
```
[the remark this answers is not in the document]
Devinwong: Addressed by the newline-trimming guard now in `stripScriptHotfixBlock`: we only trim the preceding newline for a true blank separator (double-newline), preventing adjacent YAML/content line merges.

aks-node-controller/hotfix.go
```diff
@@ -77,27 +92,68 @@ func (a *App) downloadHotfix(ctx context.Context) error {
 // hotfixConfig is the JSON structure of the hotfix configuration file.
 // Using JSON allows future extension (e.g., adding checksum, source URL) without format changes.
 type hotfixConfig struct {
-	Version string `json:"version"`
+	Version       string `json:"version"`
+	TargetVersion string `json:"target_version"`
```
awesomenix: we should have a proper name, target_version is not intuitive, may be `cse_version` would be more appropriate

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
```diff
@@ -27,13 +26,11 @@ if [ ! -f "$CONFIG_PATH" ] && [ ! -f "$NBC_CMD_PATH" ]; then
...
 
-if [ -f "$HOTFIX_JSON" ]; then
-    log "Found ANC hotfix config at ${HOTFIX_JSON}; running download-hotfix"
-    if "$BIN_PATH" download-hotfix; then
-        log "ANC download-hotfix completed; binary selection follows"
-    else
-        log "ANC download-hotfix failed; binary selection follows"
-    fi
+log "Running ANC download-hotfix pre-check"
```
awesomenix: this would cause increased latency in startup with or without hotfix since we need to parse nodecustomdata, would be a good idea to make it part of boothook

aks-node-controller/nodecustomdata.go
```diff
@@ -32,6 +36,36 @@ type nodeCustomData struct {
 }
 
 func applyNodeCustomData(path string) error {
+	hotfixCfg, err := readHotfixConfig(defaultHotfixVersionPath)
```
awesomenix: i feel this will become incredibly complex to parse specific files, would be better to make it part of boothook to make understanding easier

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
Devinwong: Just to clarify the suggestion: Currently nodecustomdata.yml is parsed twice: once in `download-hotfix` (to extract hotfix metadata) and once in `provision` (to apply all write_files). Do you mean we should parse it once at boothook, store the extracted hotfix config to disk, and let download-hotfix and provision just consume the pre-extracted results rather than re-parsing the YAML themselves?

aks-node-controller/nodecustomdata.go
```diff
@@ -17,6 +19,8 @@ const (
 	defaultNodeCustomDataPath = "/opt/azure/containers/nodecustomdata.yml"
 	encodingGZIP              = "gzip"
 	encodingBase64            = "base64"
+	scriptHotfixBeginMarker   = "# ---- hotfix: auto-generated by hotfix-generate GH Action ----"
```
awesomenix: why do you need this? its ok if the node customdata wants to replace hotfix.json isnt it?

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
```diff
@@ -27,13 +26,11 @@ if [ ! -f "$CONFIG_PATH" ] && [ ! -f "$NBC_CMD_PATH" ]; then
     exit 0
 fi
 
-if [ -f "$HOTFIX_JSON" ]; then
```
awesomenix, line 30: my thinking here is that run download-hotfix for both cases here only based on existence of HOTFIX_JSON

1. Check if version is present then download ANC from PMC
2. Check if scripts_version is present then unpack nodecustomdata.

So this way the logic will be encapsulated as hotfix and clearer for reader.

parts/hotfix/aks-node-controller-hotfix.json
```diff
@@ -0,0 +1 @@
+{}
```
awesomenix: Why need this empty file?

pkg/agent/baker.go
```diff
@@ -98,10 +98,24 @@ chmod 0600 %[1]s
...
+       "path": "/opt/azure/containers/aks-node-controller-hotfix.json",
+       "mode": 384,
+       "contents": { "source": "data:;base64,%s" }
        }`
 	nodeCustomDataPath = "/opt/azure/containers/nodecustomdata.yml"
 	nbcCmdFilePath     = "/opt/azure/containers/aks-node-controller-nbc-cmd.sh"
 	aksNodeConfigPath  = "/opt/azure/containers/aks-node-controller-config.json"
+	hotfixConfigPath   = "/opt/azure/containers/aks-node-controller-hotfix.json"
+	// hotfixConfigBlockFmt writes a plain base64-encoded JSON file (no gzip — hotfix.json is small).
+	hotfixConfigBlockFmt = `
```
awesomenix: we can probably reuse the aks node config block fmt and make it general?

pkg/agent/baker.go
```diff
@@ -193,6 +220,30 @@ func (t *TemplateGenerator) getLinuxNodeCustomDataJSONObject(config *datamodel.N
 	return fmt.Sprintf("{\"customData\": \"%s\"}", str)
 }
 
+// loadHotfixConfigJSON reads the embedded hotfix version config and returns
+// the raw JSON if a hotfix is active (version or scripts_version is set).
+// Returns an empty string when no hotfix is active.
+func loadHotfixConfigJSON() (string, error) {
+	data, err := parts.Templates.ReadFile("hotfix/aks-node-controller-hotfix.json")
```
awesomenix, line 296: Check existence of the file before reading it.

Why do you need to massage this file so much, i see that trimspace everywhere? Cant we just blindly trust the file and read it?

pkg/agent/baker.go
awesomenix: Something like `fileBlockFmt`

aks-node-controller/nodecustomdata.go
awesomenix: as previously mentioned move this to `download-hotfix` command itself. So everything is setup for anc to run.

These markers seems very fragile would recommend not using it.

aks-node-controller/hotfix.go
```diff
@@ -77,27 +83,68 @@ func (a *App) downloadHotfix(ctx context.Context) error {
...
 	if err := json.Unmarshal(data, &cfg); err != nil {
-		return "", fmt.Errorf("parsing hotfix config %s: %w", path, err)
+		return hotfixConfig{}, true, fmt.Errorf("parsing hotfix config %s: %w", path, err)
+	}
+	cfg.Version = strings.TrimSpace(cfg.Version)
+	cfg.ScriptsVersion = strings.TrimSpace(cfg.ScriptsVersion)
+	return cfg, true, nil
+}
+
+func shouldApplyScriptsVersion(currentVersion, targetVersion string) bool {
```
awesomenix, line 163: if you move Applynodecustomdata here then we can just reuse whatever is present here.

aks-node-controller/hotfix.go
```diff
@@ -77,27 +83,68 @@ func (a *App) downloadHotfix(ctx context.Context) error {
...
 	if len(strings.TrimSpace(string(data))) == 0 {
-		return "", nil
+		return hotfixConfig{}, true, nil
 	}
 	var cfg hotfixConfig
 	if err := json.Unmarshal(data, &cfg); err != nil {
-		return "", fmt.Errorf("parsing hotfix config %s: %w", path, err)
+		return hotfixConfig{}, true, fmt.Errorf("parsing hotfix config %s: %w", path, err)
+	}
+	cfg.Version = strings.TrimSpace(cfg.Version)
```
awesomenix, line 158: What is with this `TrimSpace` everywhere?

awesomenix (changes requested):

parts/linux/cloud-init/artifacts/aks-node-controller-wrapper.sh
Devinwong: okay. I moved the logic to ANC's `download-hotfix`. But I agree that it's clearer in this wrapper.sh and it could save some ms from spawning ANC binary when it's not needed. 
Added it back.

pkg/agent/baker.go
Devinwong: Added the check a few lines above.

Any agreed it was over defensive as we fully control the file. Removed unnecessary ones.

aks-node-controller/nodecustomdata.go
Devinwong: Updated. Now download-hotfix will replace both scripts hotfix and download ANC. 
Agreed the markers are fragile and removed the usage

parts/hotfix/aks-node-controller-hotfix.json
Devinwong: removed.

pkg/agent/baker.go
Devinwong: updated

aks-node-controller/hotfix.go
Devinwong: Extracted applyWriteFiles as the shared core; both callers use it; no markers now. Let me know if this is not addressed.

aks-node-controller/hotfix.go
Devinwong: agreed. It was over-defensive. Dropped most of them because the file files are fully controlled by us.

aks-node-controller/nodecustomdata.go
```diff
@@ -32,6 +32,13 @@ type nodeCustomData struct {
 }
 
 func applyNodeCustomData(path string) error {
+	return applyWriteFiles(path)
+}
```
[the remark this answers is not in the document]
Devinwong: Fixed in 2e786a70a8: `applyNodeCustomData` now `os.Stat`-guards the required `nodecustomdata.yml` and returns an error on absence (the exit-240 path in `Provision`), while `applyWriteFiles` keeps the optional no-op-on-missing behavior for the script hotfix payload. Added `TestApplyNodeCustomDataMissingFileErrors`.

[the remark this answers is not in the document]
Devinwong: Fixed in 2e786a70a8: `applyNodeCustomData` now `os.Stat`-guards the required `nodecustomdata.yml` and returns an error on absence (the exit-240 path in `Provision`), while `applyWriteFiles` keeps the optional no-op-on-missing behavior for the script hotfix payload. Added `TestApplyNodeCustomDataMissingFileErrors`.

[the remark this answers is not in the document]
Devinwong: Fixed in 2e786a70a8: `applyNodeCustomData` now `os.Stat`-guards the required `nodecustomdata.yml` and returns an error on absence (the exit-240 path in `Provision`), while `applyWriteFiles` keeps the optional no-op-on-missing behavior for the script hotfix payload. Added `TestApplyNodeCustomDataMissingFileErrors`.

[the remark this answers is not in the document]
Devinwong: Fixed in 2e786a70a8: `applyNodeCustomData` now `os.Stat`-guards the required `nodecustomdata.yml` and returns an error on absence (the exit-240 path in `Provision`), while `applyWriteFiles` keeps the optional no-op-on-missing behavior for the script hotfix payload. Added `TestApplyNodeCustomDataMissingFileErrors`.

pkg/agent/baker.go
```diff
@@ -193,6 +249,58 @@ func (t *TemplateGenerator) getLinuxNodeCustomDataJSONObject(config *datamodel.N
...
+	parameters := getParameters(config)
+	variables := getCustomDataVariables(config)
+	rendered, err := t.getSingleLine(hotfixFilesTemplatePath, config.AgentPoolProfile, getBakerFuncMap(config, parameters, variables), true)
+	if err != nil {
+		return "", err
+	}
+	// After rendering, conditionals may have elided all entries for this distro.
+	if !strings.Contains(rendered, "- path:") {
+		return "", nil
+	}
```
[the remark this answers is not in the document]
Devinwong: Fixed in 2e786a70a8: write_files detection is now line-anchored via `hasWriteFileEntries` (a line whose first non-space token is `- path:`), so a stray `- path:` inside a comment or file content can no longer be a false positive. The default placeholder file is also no longer committed (absent = no active hotfix), so the original comment-substring case can't occur either.

pkg/agent/baker.go
```diff
@@ -143,37 +166,63 @@ func (t *TemplateGenerator) getScriptlessNBCCustomData(config *datamodel.NodeBoo
...
 	if encodedAKSNodeConfig != "" {
-		aksNodeConfigBlock = fmt.Sprintf(aksNodeConfigBlockFmt, aksNodeConfigPath, encodedAKSNodeConfig)
+		extraBlocks += fmt.Sprintf(gzipFileBlockFmt, aksNodeConfigPath, encodedAKSNodeConfig)
+	}
+	if encodedHotfixConfig != "" {
+		extraBlocks += fmt.Sprintf(gzipFileBlockFmt, hotfixConfigPath, encodedHotfixConfig)
+	}
+	if encodedScriptsHotfixFiles != "" {
+		extraBlocks += fmt.Sprintf(gzipFileBlockFmt, scriptsHotfixFilesPath, encodedScriptsHotfixFiles)
 	}
```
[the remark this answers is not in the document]
Devinwong: Not applicable to our deployment model. AKS nodes are immutable: the boothook runs once on a fresh VHD where these `/opt/azure/containers` paths don't pre-exist, and a node's custom data is fixed at VM creation. The optional files are only ever written when a hotfix is active in that node's own custom data, so file presence reliably reflects active state. There's no same-disk re-provision path that could leave stale files from a previous custom-data version, so explicit removal would only add churn to every node's boothook. Resolving.

aks-node-controller/hotfix.go
```diff
@@ -29,20 +33,41 @@ const (
...
 
-// downloadHotfix installs the requested hotfix and stages it alongside the VHD-baked binary.
-// The wrapper script decides which binary to execute after this command returns.
+// downloadHotfix applies any active hotfix for the running ANC/VHD version. It is the
+// single place where version-gated remediation happens:
+//   - the ANC binary hotfix (gated by `version`), staged alongside the VHD-baked binary
+//   - the script hotfix write_files (gated by `scripts_version`), materialized to disk
+//
+// Running both here means that by the time `provision` runs, the correct binary and the
+// correct scripts are already in place — no marker-based stripping of nodecustomdata.yml.
```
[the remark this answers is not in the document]
Devinwong: Updated the PR description: the flow now states the script hotfix `write_files` are applied inside `download-hotfix` (via `applyScriptHotfix`), which the wrapper runs before `provision` -- matching the implementation.

.github/workflows/anc-hotfix-generate.yml
```diff
@@ -1,6 +1,9 @@
 name: ANC Hotfix Template Update
-# Injects the ANC hotfix version into nodecustomdata.yml when
-# hotfix/anc-hotfix-version.json is updated in a PR targeting an official/* branch.
+# Injects the ANC hotfix version into parts/hotfix/aks-node-controller-hotfix.json when
+# hotfix/anc-hotfix-version.json is updated in a PR targeting an official/* branch. baker
+# embeds that file into the boothook, which writes hotfix.json to disk at provisioning time.
```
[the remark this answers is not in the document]
Devinwong: Updated the PR description to use the actual file names: source of truth is `hotfix/anc-hotfix-version.json` (operator-edited), and the generated/embedded file written here and read by baker.go is `parts/hotfix/aks-node-controller-hotfix.json`.

pkg/agent/baker.go
```diff
@@ -193,6 +242,64 @@ func (t *TemplateGenerator) getLinuxNodeCustomDataJSONObject(config *datamodel.N
...
+}
+
+// loadHotfixConfigJSON reads the embedded hotfix config and returns the raw JSON when a
+// hotfix is active. The file is only committed on official/* hotfix branches, so an absent
+// file (normal builds) or an empty config (neither version nor scripts_version set) means
+// "no active hotfix" and returns "".
+func loadHotfixConfigJSON() (string, error) {
+	data, err := parts.Templates.ReadFile("hotfix/aks-node-controller-hotfix.json")
+	if errors.Is(err, fs.ErrNotExist) {
+		return "", nil
```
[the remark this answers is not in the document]
Devinwong: Updated the PR description to reference the actual embed path `parts/hotfix/aks-node-controller-hotfix.json` (read via embedded path `hotfix/aks-node-controller-hotfix.json`), distinct from the `hotfix/anc-hotfix-version.json` source of truth.

aks-node-controller/hotfix.go
```diff
@@ -74,30 +99,90 @@ func (a *App) downloadHotfix(ctx context.Context) error {
...
-		return "", nil
+	if len(data) == 0 {
+		return hotfixConfig{}, true, nil
 	}
 	var cfg hotfixConfig
 	if err := json.Unmarshal(data, &cfg); err != nil {
-		return "", fmt.Errorf("parsing hotfix config %s: %w", path, err)
+		return hotfixConfig{}, true, fmt.Errorf("parsing hotfix config %s: %w", path, err)
+	}
+	return cfg, true, nil
```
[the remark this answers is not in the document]
Devinwong: Fixed in cf33f8fb66: `readHotfixConfig` now treats whitespace-only content as an empty/no-op config (`len(strings.TrimSpace(...)) == 0`), so a partial/truncated write no longer errors `download-hotfix`.

aks-node-controller/hotfix.go
```diff
@@ -302,11 +387,11 @@ func copyBinaryAlongside(src, dst, refPath string) error {
...
 func shouldUpgradeToHotfix(current, hotfix string) (bool, error) {
-	cv, err := semver.NewVersion(strings.TrimSpace(current))
+	cv, err := semver.NewVersion(current)
 	if err != nil {
 		return false, fmt.Errorf("parsing current version %q: %w", current, err)
 	}
-	hv, err := semver.NewVersion(strings.TrimSpace(hotfix))
+	hv, err := semver.NewVersion(hotfix)
 	if err != nil {
 		return false, fmt.Errorf("parsing hotfix version %q: %w", hotfix, err)
```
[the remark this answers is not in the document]
Devinwong: Fixed in cf33f8fb66, but at the boundary rather than in the comparator: `readHotfixConfig` now trims `version`/`scripts_version` once after unmarshal, so `shouldUpgradeToHotfix`/`shouldApplyScriptsVersion` consume already-normalized values and stay trim-free. This handles incidental operator whitespace while avoiding scattered TrimSpace calls.

pkg/agent/baker.go
```diff
@@ -70,7 +70,9 @@ systemctl start --no-block aks-node-controller.service
 	// aksNodeConfigBlock is appended to the boothook when AKSNodeConfig is provided.
 	// It writes the gzipped+base64-encoded JSON config to disk so the wrapper script
 	// can pass --provision-config alongside --nbc-cmd for env comparison.
-	aksNodeConfigBlockFmt = `
+	// gzipFileBlockFmt writes a gzipped+base64-encoded file to disk. It is general-purpose
+	// and reused for any boothook payload that is gzip-compressed (AKSNodeConfig, hotfix files).
+	gzipFileBlockFmt = `
```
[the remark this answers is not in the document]
Devinwong: Fixed earlier in de8c0a45d1: the comment above `gzipFileBlockFmt` was rewritten to describe the general-purpose name/usage and no longer references `aksNodeConfigBlock`. Resolving (thread is on an outdated revision).

aks-node-controller/nodecustomdata.go
Devinwong: Then I have to create a new parts/hotfix/anc-scripts-hotfix-files.yml which stores the hotfix scripts contents.

The clean separation is:

| Subcommand | Parses | Never touches |
|------------|--------|---------------|
| `download-hotfix` | `aks-node-controller-hotfix.json` + `anc-scripts-hotfix-files.yml` | `nodecustomdata.yml` |
| `provision` | `nodecustomdata.yml` | the hotfix files |

`aks-node-controller-hotfix.json` only tells `download-hotfix` if there is anc (and which version) or scripts hotfix, and if there is scripts hotfix, the contents are in `parts/hotfix/anc-scripts-hotfix-files.yml`.

aks-node-controller/hotfix.go
```diff
@@ -29,20 +33,41 @@ const (
...
+		return fmt.Errorf("read hotfix config from %s: %w", hotfixPath, err)
+	}
+	if !exists {
+		return nil
 	}
 
+	if err := a.downloadHotfixBinary(ctx, hotfixCfg); err != nil {
+		return err
+	}
+	return a.applyScriptHotfix(hotfixCfg)
```
awesomenix, line 60: Recommend applying scripts hotfix before binary download, greater chance of success for scripts not to fail

aks-node-controller/hotfix.go
awesomenix: i meant shouldnt this function be same as 

`func shouldUpgradeToHotfix(current, hotfix string) (bool, error) {` 

Why not just reuse that?

aks-node-controller/hotfix.go
```diff
@@ -29,20 +33,41 @@ const (
...
 		hotfixPath = defaultHotfixVersionPath
 	}
-	hotfixVersion, err := readHotfixVersion(hotfixPath)
+
+	hotfixCfg, exists, err := readHotfixConfig(hotfixPath)
 	if err != nil {
-		return fmt.Errorf("read hotfix version from %s: %w", hotfixPath, err)
+		return fmt.Errorf("read hotfix config from %s: %w", hotfixPath, err)
+	}
+	if !exists {
```
awesomenix, line 53: Why do you need this `exists` flag here?

Isnt `Version == "" && Scripts_Version == ""` mean the same or just fall through and everything should work out since we check for emptyness in individual functions anyways. Or simple

```
if hotfixCfg.Version != "" {
   if err := a.downloadHotfixBinary(ctx, hotfixCfg.Version); err != nil {
		return err
	}
}

if hotfixCfg.Scripts_Version != "" {
   if err := a.applyScriptHotfix(ctx, hotfixCfg.Scripts_version); err != nil {
		return err
	}
}

return nil
```

aks-node-controller/nodecustomdata.go
```diff
@@ -32,6 +32,19 @@ type nodeCustomData struct {
 }
 
 func applyNodeCustomData(path string) error {
+	// nodecustomdata.yml is required input for the NBCCmd provisioning path, so a missing
+	// file must be a hard error. applyWriteFiles itself treats absence as a no-op for
+	// optional payloads (e.g. the script hotfix), so guard existence here before delegating.
+	if _, err := os.Stat(path); err != nil {
```
awesomenix, line 38: we already perform the check inside applyWriteFiles this is not needed here

pkg/agent/baker.go
```diff
@@ -193,6 +242,77 @@ func (t *TemplateGenerator) getLinuxNodeCustomDataJSONObject(config *datamodel.N
...
+	}
+	return false
+}
+
+// loadHotfixConfigJSON reads the embedded hotfix config and returns the raw JSON when a
+// hotfix is active. The file is only committed on official/* hotfix branches, so an absent
+// file (normal builds) or an empty config (neither version nor scripts_version set) means
+// "no active hotfix" and returns "".
+func loadHotfixConfigJSON() (string, error) {
+	data, err := parts.Templates.ReadFile("hotfix/aks-node-controller-hotfix.json")
```
lilypan26, line 296: shoulnd;t this be `/opt/azure/containers/aks-node-controller-hotfix.json` ?

PR #8768, opened 2026-06-24
feat(windows): add WS2025 Gen2 Trusted Launch VHD build

sinmentis:
Add a new `2025-gen2-tl` Windows SKU that builds a WS2025 Gen2 VHD whose SIG image definition is stamped with SecurityType=TrustedLaunch, so AKS Windows node pools can run on a Trusted Launch image.

The VHD content is identical to `2025-gen2` (same base image, components, CSE). The only difference is the image definition carries the TrustedLaunch feature, driven by ENABLE_TRUSTED_LAUNCH=True threaded through the Windows template chain into produce-packer-settings (which is already OS-agnostic).

- windows_settings.json / windows_settings.cue: new `2025-gen2-tl` entry, reusing the existing TL-capable base SKU 2025-datacenter-core-smalldisk-g2
- windows VHD templates: new build2025gen2tl stage + enableTrustedLaunch param threaded to ENABLE_TRUSTED_LAUNCH in the buildVhd / run-test / convert-sig env blocks
- run-test.sh / convert-sig-to-classic-storage-account-blob.sh: extend the Trusted Launch guard to Windows so the cache-test VM boots with --security-type TrustedLaunch (this also makes it the Secure Boot ground-truth check) and the production disk is created with the TrustedLaunch security profile

PR build defaults off; release build defaults on.

Fixes #

### DIFF
--- a/.pipelines/.vsts-vhd-builder-pr-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-pr-windows.yaml
@@ -77,7 +77,7 @@ stages:
       # 2025 gen2 is the default, so build that for PRs
       build2025: False
       build2025gen2: True
-      build2025gen2tl: False
+      build2025gen2tl: True
 
   - stage: delete_old_windows_vhds
     dependsOn: [ ]
--- a/.pipelines/.vsts-vhd-builder-pr-windows.yaml
+++ b/.pipelines/.vsts-vhd-builder-pr-windows.yaml
@@ -77,6 +77,7 @@ stages:
       # 2025 gen2 is the default, so build that for PRs
       build2025: False
       build2025gen2: True
+      build2025gen2tl: False
 
   - stage: delete_old_windows_vhds
     dependsOn: [ ]
--- a/.pipelines/templates/.build-and-test-windows-vhds-template.yaml
+++ b/.pipelines/templates/.build-and-test-windows-vhds-template.yaml
@@ -11,6 +11,9 @@ parameters:
   - name: build2025gen2
     displayName: Build 2025 Gen 2
     type: boolean
+  - name: build2025gen2tl
+    displayName: Build 2025 Gen 2 Trusted Launch
+    type: boolean
   - name: windowsContainerImageJsonUrl
     displayName: Windows container image JSON URL Override
     type: string
@@ -151,6 +154,30 @@ stages:
       architecture: X86_64
       buildVmSize: ${{ parameters.buildVmSize }}
       build: ${{ parameters.build2025gen2 }}
+      vhddebug: ${{ parameters.vhddebug }}
+      dryrun: ${{ parameters.dryrun }}
+      installOpenSshServer: ${{ parameters.installOpenSshServer }}
+      skipExtensionCheck: ${{ parameters.skipExtensionCheck }}
+      windowsBaseImageUrl: $(WINDOWS_2025_GEN2_BASE_IMAGE_URL)
+      windowsNanoImageUrl: $(WINDOWS_2025_NANO_IMAGE_URL)
+      windowsCoreImageUrl: $(WINDOWS_2025_CORE_IMAGE_URL)
+      windowsContainerImageJsonUrl: ${{ parameters.windowsContainerImageJsonUrl }}
+      overrideBranch: ${{ parameters.overrideBranch }}
+      useOverrides: ${{ parameters.useOverrides }}
+      csePackageDir: ${{ parameters.csePublishDir }}
+      cseFileName: ${{ parameters.cseFileName }}
+
+  - template: ./.build-and-test-windows-vhd-template.yaml
+    parameters:
+      stageName: win_2025_v2_tl
+      artifactName: 2025-gen2-tl
+      imageName: windows-2025-gen2-tl
+      windowsSku: 2025-gen2-tl
+      hyperVGeneration: V2
+      architecture: X86_64
+      enableTrustedLaunch: True
+      buildVmSize: ${{ parameters.buildVmSize }}
+      build: ${{ parameters.build2025gen2tl }}
       vhddebug: ${{ parameters.vhddebug }}
       dryrun: ${{ parameters.dryrun }}
       installOpenSshServer: ${{ parameters.installOpenSshServer }}
--- a/.pipelines/templates/.builder-release-template-windows.yaml
+++ b/.pipelines/templates/.builder-release-template-windows.yaml
@@ -61,7 +61,6 @@ parameters:
   - name: enableTrustedLaunch
     type: boolean
     displayName: Enable Trusted Launch on the SIG image definition.
-    default: False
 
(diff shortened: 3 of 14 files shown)


.pipelines/templates/.builder-release-template-windows.yaml
```diff
@@ -58,6 +58,10 @@ parameters:
   - name: csePackageFileName
     type: string
     displayName: CSE package filename.
+  - name: enableTrustedLaunch
+    type: boolean
+    displayName: Enable Trusted Launch on the SIG image definition.
+    default: False
```
timmy-wright: My preference is to not have a default and explicitly set it in the places that use this. Which isn't many places so it's not a big change.

.pipelines/.vsts-vhd-builder-pr-windows.yaml
```diff
@@ -77,6 +77,7 @@ stages:
       # 2025 gen2 is the default, so build that for PRs
       build2025: False
       build2025gen2: True
+      build2025gen2tl: True
```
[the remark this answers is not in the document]
timmy-wright: Disagree - running the agentbaker e2es for ws2025 and trusted launch is sensible.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
```diff
@@ -54,7 +54,7 @@ disk_resource_id="/subscriptions/${SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GR
 capture_benchmark "${SCRIPT_NAME}_set_variables_for_converting_to_disk"
 
 echo "Converting $sig_resource_id to $disk_resource_id"
-if [ "${OS_TYPE}" = "Linux" ] && [ "${ENABLE_TRUSTED_LAUNCH}" = "True" ]; then
+if { [ "${OS_TYPE}" = "Linux" ] || [ "${OS_TYPE}" = "Windows" ]; } && [ "${ENABLE_TRUSTED_LAUNCH}" = "True" ]; then
```
timmy-wright: We can drop the "os==linux || os==windows" part of this test - just need to check trusted launch is set to true.

.pipelines/templates/.build-and-test-windows-vhds-template.yaml
```diff
@@ -164,6 +167,30 @@ stages:
       csePackageDir: ${{ parameters.csePublishDir }}
       cseFileName: ${{ parameters.cseFileName }}
 
+  - template: ./.build-and-test-windows-vhd-template.yaml
+    parameters:
+      stageName: win_2025_v2_tl
+      artifactName: 2025-gen2-tl
+      imageName: windows-2025-gen2-tl
```
timmy-wright, line 174: Just checking: is the "-tl" suffix consistent with the naming of the Linux trusted launch VHDs?

vhdbuilder/packer/test/run-test.sh
```diff
@@ -72,7 +72,7 @@ else
   TARGET_COMMAND_STRING="--size Standard_D2ds_v5"
 fi
 
-if [ "${OS_TYPE}" = "Linux" ] && [ "${ENABLE_TRUSTED_LAUNCH,,}" = "true" ]; then
+if { [ "${OS_TYPE}" = "Linux" ] || [ "${OS_TYPE}" = "Windows" ]; } && [ "${ENABLE_TRUSTED_LAUNCH,,}" = "true" ]; then
```
timmy-wright: We can drop the "os==linux || os==windows" part of this test - just need to check trusted launch is set to true.

timmy-wright (commented):
Core of changes looks good - I wrote some nitpicky comments. It'll need an e2e test. Base that on the ws2025-gen2 e2e test.

.pipelines/templates/.builder-release-template-windows.yaml
sinmentis: Done - dropped the default. The release template is only ever called from the singular template, which always passes enableTrustedLaunch explicitly, so it no longer needs a default.

.pipelines/.vsts-vhd-builder-pr-windows.yaml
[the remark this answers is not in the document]
sinmentis: Agreed, keeping it True so the PR gate builds the WS2025 Gen2 TL VHD and runs the agentbaker e2es against it.

vhdbuilder/packer/convert-sig-to-classic-storage-account-blob.sh
sinmentis: Done - dropped the OS_TYPE check, now gating on ENABLE_TRUSTED_LAUNCH only.

vhdbuilder/packer/test/run-test.sh
sinmentis: Done - dropped the OS_TYPE check, now gating on ENABLE_TRUSTED_LAUNCH only.

.pipelines/templates/.build-and-test-windows-vhds-template.yaml
sinmentis: It keeps the same `tl` token the Linux TL VHDs use (`acl-tl-gen2`, `acl-fips-tl-gen2`). The order is `gen2-tl` rather than Linux's `tl-gen2` because the existing Windows SKUs already suffix `gen2` (`2025-gen2`, `2022-containerd-gen2`), so `-tl` is just appended to the existing Windows name. The resulting image-definition `windows-2025-gen2-tl` contains `tl`, which is what the RP-side getSecurityType matches on. Happy to switch to `2025-tl-gen2` if you'd prefer the exact Linux ordering.

r2k1 (approved):

e2e/scenario_win_test.go
```diff
@@ -215,6 +215,33 @@ func Test_Windows2025Gen2(t *testing.T) {
...
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.Properties = addTrustedLaunchToVMSS(vmss.Properties)
+			},
+			BootstrapConfigMutator: func(_ *Cluster, configuration *datamodel.NodeBootstrappingConfiguration) {
+				Windows2025BootstrapConfigMutator(t, configuration)
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				ValidateWindowsVersionFromWindowsSettings(ctx, s, "2025-gen2-tl")
+				ValidateWindowsProductName(ctx, s, "Windows Server 2025 Datacenter")
+				ValidateWindowsDisplayVersion(ctx, s, "24H2")
```
r2k1, line 233: 24h2 sounds strange

sinmentis: Server 2025 is built on the Windows 11 24H2 codebase, same way Server 2022 reads "21H2"

PR #8769, opened 2026-06-24
fix: cleanup cluster resources if aks cluster is being deleted

awesomenix:
- cleanup attached resources before deleting the cluster for successful deletion
- seems like the aks dev infra deletes the MC resource group, causing nuisance, since we used shared resources we are indefinitely stuck.

Fix it

### DIFF
--- a/e2e/cluster.go
+++ b/e2e/cluster.go
@@ -351,6 +351,10 @@ func getExistingCluster(ctx context.Context, location, clusterName string) (*arm
 			return &existingCluster.ManagedCluster, nil
 		}
 		toolkit.Logf(ctx, "##vso[task.logissue type=warning;]Cluster %s has deleting or missing node resource group %s, deleting cluster", clusterName, *existingCluster.Properties.NodeResourceGroup)
+		nodeResourceGroup := *existingCluster.Properties.NodeResourceGroup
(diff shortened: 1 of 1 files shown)


timmy-wright (approved):

e2e/cluster.go
```diff
@@ -351,6 +351,10 @@ func getExistingCluster(ctx context.Context, location, clusterName string) (*arm
 			return &existingCluster.ManagedCluster, nil
 		}
 		toolkit.Logf(ctx, "##vso[task.logissue type=warning;]Cluster %s has deleting or missing node resource group %s, deleting cluster", clusterName, *existingCluster.Properties.NodeResourceGroup)
+		nodeResourceGroup := *existingCluster.Properties.NodeResourceGroup
+		if cleanupErr := detachNodeResourceGroupReferencesFromClusterSubnet(ctx, *existingCluster.Location, *existingCluster.Name, nodeResourceGroup); cleanupErr != nil {
```
r2k1, line 355: I'm a bit confused what is cleaned here.

The RG is deleted and cluster will be deleted. Where are these resources live?

r2k1 (approved):

e2e/cluster.go
awesomenix: so before we used to have vnet in each RG, which means AKS associates subnet to the NSG/RouteTable/ etc. But now since the vnet lives in shared resource group, AKS skip disassociating it because the vnet is not owned, which in my opinions not cool. But we have to work around it.

lilypan26 (approved):

PR #8785, opened 2026-06-29
docs: add ACL troubleshooting skills file for AKS Node SIG

Ankita13-code:
Adds a dedicated Copilot skills file (doc/skills/acl-troubleshooting.md) for Azure Container Linux (ACL) troubleshooting, and a pointer to it from .github/copilot-instructions.md.

This enables AKS Node SIG engineers to use GitHub Copilot for diagnosing ACL-specific issues. The skills file covers:

ACL architecture (immutable host, UKI, sysexts, Ignition provisioning)
ACL-specific error codes (231, 232) and related ORAS errors
Component delivery via sysexts (kubelet, kubectl, GPU drivers) with registry paths
Log locations (CSE, Ignition, coreos-cloudinit, kubelet)
Troubleshooting workflows (debug pods, sysext failures, GPU issues, SELinux, package version identification)
Development guidelines for making ACL-safe changes in AgentBaker
Content is derived from the [ACL TSG](vscode-file://vscode-app/c:/Users/ankitapareek/AppData/Local/Programs/Microsoft%20VS%20Code/7e7950df89/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and AgentBaker source code.

Fixes N/A — new documentation, no linked issue.

### DIFF
--- a/.github/skills/acl-troubleshooting/SKILL.md
+++ b/.github/skills/acl-troubleshooting/SKILL.md
@@ -1,0 +1,328 @@
+---
+name: acl-troubleshooting
+description: >
+  Use when diagnosing or resolving issues related to Azure Container Linux (ACL)
+  nodes in AKS clusters. Triggers: "ACL node issue", "sysext failure",
+  "ERR_ORAS_PULL_SYSEXT_FAIL", "immutable host", "no package manager on node",
+  "Ignition provisioning failure", "SELinux denial on ACL", "GPU driver ACL",
+  "coreos-cloudinit", "systemd-sysext", "ACL debug pod", "ACL package versions".
+---
+
+# Azure Container Linux (ACL) Troubleshooting for AKS Node SIG
+
+ACL is fundamentally different from Ubuntu and traditional Azure Linux
+(Mariner) — many debugging workflows that work on those distros will **not**
+work on ACL.
+
+## What is ACL?
+
+Azure Container Linux is a container-host distro composed of Azure Linux 3.0
+packages, assembled into a UKI-based (Unified Kernel Image) immutable image
+using Flatcar Container Linux build scripts and composition tooling.
+
+Key properties:
+- **Immutable `/usr`** — `/usr` is mounted read-only and changes only via image updates or sysexts; `/etc`, `/opt`, and `/var` remain writable. No package manager (`dnf`, `rpm`, `tdnf`) is available on the host.
+- **Image-based updates** — no in-place package updates; the entire OS image is replaced (currently via AKS node image upgrade; A/B Trident upgrades are in progress).
+- **UKI boot** — ACL boots from a Unified Kernel Image. systemd-boot loads UKI addons from `/boot/EFI/Linux/<uki>.extra.d/`; a `firstboot.addon.efi` sets `flatcar.first_boot=detected` on the kernel command line, which triggers Ignition. The VHD build consumes this addon during packer provisioning and `cleanup-vhd.sh` restores it so every VM created from the VHD sees a genuine first boot.
+- **Ignition-based provisioning** — first-boot config runs as systemd units in the initramfs, before switch-root (not cloud-init as primary).
+- **SELinux enforcing by default**.
+- **Sysexts for extensibility** — additional packages are layered via systemd system extensions rather than package installs.
+
+## ACL in AgentBaker
+
+### Relevant paths
+
+| Purpose | Path |
+|---------|------|
+| ACL-specific CSE helpers | `parts/linux/cloud-init/artifacts/acl/cse_helpers_acl.sh` |
+| ACL-specific install logic | `parts/linux/cloud-init/artifacts/acl/cse_install_acl.sh` |
+| ACL cloud-init/Ignition config | `parts/linux/cloud-init/acl.yml` |
+| ACL VHD packer customdata (Butane) | `vhdbuilder/packer/acl-customdata.yaml` |
+| ACL VHD packer customdata (JSON) | `vhdbuilder/packer/acl-customdata.json` |
+| OS detection helpers | `parts/linux/cloud-init/artifacts/cse_helpers.sh` — `isACL()` |
+| Sysext merge logic | `parts/linux/cloud-init/artifacts/acl/cse_install_acl.sh` — `mergeSysexts()` |
+| UKI firstboot addon restore | `vhdbuilder/packer/cleanup-vhd.sh` |
+
+### OS detection
+
+In shell scripts, use the helper functions:
+```bash
+isACL()                  # true on Azure Container Linux
+isMarinerOrAzureLinux()  # true on Mariner/AzL — NOT on ACL
+isUbuntu()               # true on Ubuntu
+```
+
+**Important:** `isMarinerOrAzureLinux()` explicitly excludes ACL. When adding
+logic that should apply to ACL, you must handle the `isACL()` case separately.
+
+### ACL stubs
+
+Many package-management functions are stubbed out in `cse_helpers_acl.sh`
+because ACL has no package manager. Functions like `apt_get_update()`,
+`holdWALinuxAgent()`, etc. are no-ops on ACL. If you need to install software
+on ACL, it must be done via sysexts or baked into the VHD image.
+
+## Sysext-related error codes on ACL
+
+These CSE exit codes are relevant when troubleshooting ACL; code 231 is also used by Flatcar:
+
+| Code | Constant | Meaning |
+|------|----------|---------|
+| 231 | `ERR_ORAS_PULL_SYSEXT_FAIL` | Failed to pull a systemd sysext via ORAS from the registry |
+| 232 | `ERR_SYSEXT_VERSION_ID_NOT_FOUND` | `VERSION_ID` not found in `/etc/os-release`; required for sysext tag resolution |
+
+Related ORAS errors (not ACL-specific but frequently seen on ACL):
+
+| Code | Constant | Meaning |
+|------|----------|---------|
+| 45 | `ERR_ORAS_DOWNLOAD_ERROR` | Unable to install/run ORAS |
+| 207 | `ERR_ORAS_PULL_K8S_FAIL` | Failed to pull kube-node artifact via ORAS |
+| 210 | `ERR_ORAS_IMDS_TIMEOUT` | Timeout waiting for IMDS response during ORAS auth |
+| 211 | `ERR_ORAS_PULL_NETWORK_TIMEOUT` | Timeout pulling ORAS tokens for login |
+| 212 | `ERR_ORAS_PULL_UNAUTHORIZED` | Authorization failure pulling artifact via ORAS |
+
+## How components are delivered on ACL
+
+Unlike Ubuntu/Mariner where binaries are installed via packages, ACL delivers
+them as **sysexts pulled from OCI registries via ORAS**:
+
+| Component | Registry path | Installed when | Notes |
+|-----------|--------------|----------------|-------|
+| kubelet | `mcr.microsoft.com/oss/v2/kubernetes/kubelet-sysext` | Node CSE | Tagged by k8s version |
+| kubectl | `mcr.microsoft.com/oss/v2/kubernetes/kubectl-sysext` | Node CSE | Tagged by k8s version |
+| azure-acr-credential-provider | `mcr.microsoft.com/oss/v2/kubernetes/azure-acr-credential-provider-sysext` | Node CSE | Tagged by version |
+| aks-secure-tls-bootstrap-client | `mcr.microsoft.com/aks-secure-tls-bootstrap/v2/aks-secure-tls-bootstrap-client-sysext` | **VHD build only** | Pull failures surface in the VHD build pipeline, not in node provisioning logs |
+| GPU drivers (NVIDIA) | `mcr.microsoft.com/azurelinux/<major.minor>/azure-container-linux/<sysext-name>` | Node CSE | Tagged by `VERSION_ID` from `/etc/os-release` |
+
+After sysext merge, compatibility symlinks are created (e.g., `/opt/bin/kubelet` →
+`/usr/bin/kubelet`) and `systemd-sysext --no-reload refresh` activates them.
+
+**Network-isolated clusters:** When `BOOTSTRAP_PROFILE_CONTAINER_REGISTRY_SERVER`
+is set, ACR cache rules don't support ORAS repo tag listing. The code falls back
+to a fixed tag format `v{version}-1-azlinux3-{arch}` instead of querying tags.
+
+### GPU sysext selection, pull, and activation
+
+GPU sysexts are selected based on the VM SKU and `NVIDIA_GPU_DRIVER_TYPE`
+(set by AgentBaker from GPU SKU maps in `gpu_components.go`):
+
+1. **Driver type resolution:**
+   - `grid` (converged sizes like NVads_A10_v5) → `nvidia-driver-vgpu` sysext
+   - `grid-v20` → **not supported on ACL** (Ubuntu-only); fails with `ERR_NVIDIA_DRIVER_INSTALL`
+   - Default: `nvidia-driver-cuda` (proprietary) for legacy GPUs (T4, V100), or `nvidia-driver-cuda-open` (OpenRM) for A100+
+
+2. **Version resolution:** `getACLVersionID()` reads `VERSION_ID` from `/etc/os-release` (e.g., `3.0.20260304`). GPU sysexts are tagged by the OS image version, not the driver version.
+
+3. **Registry path:** `mcr.microsoft.com/azurelinux/<major.minor>/azure-container-linux/<sysext-name>:<VERSION_ID>`
+
+4. **Pull and activation:** `installACLGPUSysext()` → `mergeSysexts()` → ORAS pull → symlink to `/etc/extensions/<name>.raw` → `systemd-sysext refresh`.
+
+## Where to find logs
+
+**CSE (Custom Script Extension) logs:**
+```bash
+/var/log/azure/aks/                              # main AKS log directory
+/var/log/azure/cluster-provision.log             # provisioning output
+/var/log/azure/cluster-provision-cse-output.log  # symlink to above
+```
+
+**Ignition logs** (first-boot only — check these first for provisioning failures):
+```bash
+sudo journalctl -u ignition-disks.service    # disk/partition provisioning
+sudo journalctl -u ignition-files.service    # file writes, user creation
+sudo journalctl -t ignition                  # all Ignition messages
+```
+
+**cloud-init (coreos-cloudinit) logs** — for AKS custom-data injection:
+```bash
+sudo journalctl -u oem-cloudinit.service
+sudo journalctl -u coreos-cloudinit.service  # alternative unit name
+```
+
+> ACL uses Flatcar's `coreos-cloudinit`, not upstream `cloud-init`.
+> There is no `/var/log/cloud-init.log`. All output goes through the journal.
+
+**kubelet logs:**
+```bash
+sudo journalctl -u kubelet.service
+```
+
+## Troubleshooting common issues
+
+### "Command not found" or "Package manager not available"
+
+**Root cause:** ACL has no `dnf`, `rpm`, `tdnf`, or `apt`. The host is immutable.
+
+**Resolution:** Run debug tools from a privileged container:
+```bash
+kubectl debug node/<node-name> -it \
+  --image=mcr.microsoft.com/azurelinux/base/core:3.0 \
+  --profile=sysadmin
+```
+Inside the debug pod (host PID/network/IPC namespaces and full capabilities;
+host filesystem at `/host`):
+```bash
+tdnf install -y strace tcpdump
+strace -p <pid>   # host PIDs visible via shared namespace
+```
+Delete the debug pod when done.
+
+### Sysext issues
+
+**Check active sysexts:**
+```bash
+systemd-sysext status
+systemd-sysext list
+```
+
+**Sysext download failures:** Look for `ERR_ORAS_PULL_SYSEXT_FAIL` (exit 231)
+in CSE logs. The `mergeSysexts()` function handles the flow:
+1. `matchLocalSysext()` — checks `/opt/<name>/downloads/` for a local `.raw` file
+2. `matchRemoteSysext()` — queries `oras repo tags` to find the best matching tag
+3. `downloadSysextFromVersion()` — `oras pull` from the registry
+4. Symlinks result to `/etc/extensions/<name>.raw`
+5. `systemd-sysext --no-reload refresh` activates all extensions
+
+### Provisioning failures (Ignition)
+
+ACL uses **Ignition** for first-boot provisioning, running as systemd units
+in the initramfs before switch-root. See the "Where to find logs" section
+for log locations.
+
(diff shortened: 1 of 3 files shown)


doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+| 210 | `ERR_ORAS_IMDS_TIMEOUT` | Timeout waiting for IMDS response during ORAS auth |
+| 211 | `ERR_ORAS_PULL_NETWORK_TIMEOUT` | Timeout pulling ORAS tokens for login |
+| 212 | `ERR_ORAS_PULL_UNAUTHORIZED` | Authorization failure pulling artifact via ORAS |
+
+## How components are delivered on ACL
+
+Unlike Ubuntu/Mariner where kubelet, kubectl, and other binaries are installed
+via packages, ACL delivers them as **sysexts pulled from OCI registries via ORAS**:
+
+| Component | Registry path | Notes |
```
jiria: Should we have the complete list here?

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+1. Checks for a local match in `/opt/<name>/downloads/`
+2. Falls back to querying the remote registry via `oras repo tags`
+3. Downloads and symlinks to `/etc/extensions/<name>.raw`
+
+### Provisioning failures (Ignition)
+
+ACL uses **Ignition** for first-boot provisioning (runs in initramfs, before
+systemd). See the "Where to find logs" section above for log locations.
+
+Ignition is **strictly one-shot** — it never re-runs. If provisioning failed,
```
jiria: Perhaps we should mention here the first boot UKI addon and its role.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+the node must be re-imaged.
+
+**Check order:** Always check Ignition logs first (it runs earlier and handles
+the critical path). `coreos-cloudinit` handles supplemental custom-data that
+runs after Ignition.
+
+### GPU driver failures on ACL
```
jiria: I think we should mention a bit more about how the GPU sysext is selected, pulled and activated on the first boot.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+getenforce           # check current mode
+```
+
+**Persistently change mode** (edit `/etc/selinux/config`):
+```ini
+SELINUX=permissive   # or enforcing
+```
+Requires reboot. Note: `/etc` changes may not survive image-based updates.
+
+**Set SELinux mode at AKS nodepool level:**
```
jiria: Perhaps highlight that this is even more durable and would survive updates as well.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+sudo setenforce 0   # permissive
+sudo setenforce 1   # back to enforcing
+getenforce           # check current mode
+```
+
+**Persistently change mode** (edit `/etc/selinux/config`):
+```ini
+SELINUX=permissive   # or enforcing
+```
+Requires reboot. Note: `/etc` changes may not survive image-based updates.
```
jiria: With the current model, /etc changes will survive image based update.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Set SELinux mode at AKS nodepool level:**
+```bash
+# Create nodepool in permissive mode:
+az aks nodepool add \
+  --resource-group myRG --cluster-name myCluster --name myPool \
+  --node-count 1 \
+  --tags acl-node-security-profile="selinux=permissive"
+
+# Update existing nodepool:
```
jiria: Probably add a note or a command here about reboot again.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
+# Skill: Azure Container Linux (ACL) Troubleshooting for AKS Node SIG
```
jiria: Curious why here and not under https://github.com/Azure/AgentBaker/blob/main/.github/copilot/skills. Also if it is a copilot skill, we should probably add frontmatter.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+  --resource-group myRG --cluster-name myCluster --name myPool \
+  --node-count 1 \
+  --tags acl-node-security-profile="selinux=permissive"
+
+# Update existing nodepool:
+az aks nodepool update \
+  --resource-group myRG --cluster-name myCluster --name myPool \
+  --tags acl-node-security-profile="selinux=permissive"
+```
+Verify: `kubectl debug` → `chroot /host` → `sestatus`.
```
jiria: ⚠️ Worth hardening this block before it lands — it's the one genuinely destructive command in the doc:

```bash
az aks nodepool update \
  --resource-group myRG --cluster-name myCluster --name myPool \
  --tags acl-node-security-profile="selinux=permissive"
```

Two issues:

1. **`--tags` replaces the whole tag set**, it isn't additive. Anyone copy-pasting this onto a real pool silently drops cost-center / ownership / team tags. Worth reading them back first with `az aks nodepool show ... --query tags` and re-supplying them.

2. **The tag is consumed at node provisioning time**, but the `Verify:` line right below sends the reader to `sestatus` on an existing node — where they'll still see `enforcing` and reasonably conclude the tag doesn't work. Could we add a note that the pool needs `az aks nodepool upgrade --node-image-only` (or a scale-out) first, and that verification should happen on a *newly provisioned* node?

Suggest leading with the `nodepool add` path as the recommended one, and putting the `update` path behind an explicit warning.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+
+Unlike Ubuntu/Mariner where kubelet, kubectl, and other binaries are installed
+via packages, ACL delivers them as **sysexts pulled from OCI registries via ORAS**:
+
+| Component | Registry path | Notes |
+|-----------|--------------|-------|
+| kubelet | `mcr.microsoft.com/oss/v2/kubernetes/kubelet-sysext` | Tagged by k8s version |
+| kubectl | `mcr.microsoft.com/oss/v2/kubernetes/kubectl-sysext` | Tagged by k8s version |
+| azure-acr-credential-provider | `mcr.microsoft.com/oss/v2/kubernetes/azure-acr-credential-provider-sysext` | Tagged by version |
+| aks-secure-tls-bootstrap-client | `mcr.microsoft.com/aks-secure-tls-bootstrap/v2/aks-secure-tls-bootstrap-client-sysext` | Tagged by version |
```
jiria: Small correctness fix on this row:

```markdown
| aks-secure-tls-bootstrap-client | `mcr.microsoft.com/.../aks-secure-tls-bootstrap-client-sysext` | Tagged by version |
```

Unlike the other three, this one is installed at **VHD build time**, not during node CSE — [`cse_install_acl.sh`](https://github.com/Azure/AgentBaker/blob/main/parts/linux/cloud-init/artifacts/acl/cse_install_acl.sh) even calls it out directly:

```bash
# Only called at build-time, unlike kubelet or credential provider installation.
installSecureTLSBootstrapClientSysext() {
```

Since it sits in the same table with the same "Tagged by version" note, someone triaging a TLS bootstrap failure will go hunting for an ORAS pull / `ERR_ORAS_PULL_SYSEXT_FAIL` in the node CSE logs and find nothing. Could we add a note so it points at the VHD build pipeline instead?

```markdown
Tagged by version — **installed at VHD build time only**, not during node CSE.
Pull failures surface in the VHD build pipeline, not in node provisioning logs.
```

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+4. Sysext package lists are also in artifacts (e.g., `nvidia-driver-cuda_packages.txt`).
+
+### Ignition vs cloud-init: key differences
+
+| | Ignition | cloud-init |
+|---|---|---|
+| Format | JSON (or Butane YAML transpiled) | YAML cloud-config |
+| When | initramfs — before switch-root | Post switch-root, as systemd service |
+| Re-runs | Never — provision once | Can re-run per-boot modules |
+| Failure mode | Atomic — fully provisions or fails clearly | Partial state possible |
```
jiria: Two small corrections in the Ignition comparison table and the prose above it:

1. **"Failure mode | Atomic — fully provisions or fails clearly"** — Ignition isn't transactional. It runs staged (`ignition-disks` → `ignition-files`) and a mid-stage failure leaves partitions created and some files already written before it drops to `emergency.target`. The "atomic" framing could lead someone triaging a failed provision to assume any file on disk came from a successful run, and skip looking for exactly the half-applied state they need to find. Maybe:

   ```markdown
   | Failure mode | Fails fast into `emergency.target`; writes from earlier stages are already applied (not transactional) |
   ```

2. **"runs in initramfs, before systemd"** — minor, but this is contradicted by the log commands a few sections up (`journalctl -u ignition-disks.service`, `-u ignition-files.service`), which are systemd units. systemd *is* running in the initramfs. "runs in the initramfs as systemd units, before switch-root" would be more precise.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+2. Falls back to querying the remote registry via `oras repo tags`
+3. Downloads and symlinks to `/etc/extensions/<name>.raw`
+
+### Provisioning failures (Ignition)
+
+ACL uses **Ignition** for first-boot provisioning (runs in initramfs, before
+systemd). See the "Where to find logs" section above for log locations.
+
+Ignition is **strictly one-shot** — it never re-runs. If provisioning failed,
+the node must be re-imaged.
```
jiria: This is the doc's guidance for its most severe failure class, so it might be worth expanding slightly:

> Ignition is **strictly one-shot** — it never re-runs. If provisioning failed, the node must be re-imaged.

Two things I'd add:

1. **Collect logs first.** This section is about *diagnosing* a provisioning failure, but the only remediation given destroys the sole copy of the evidence — the Ignition journal lives on that node. Worth a `journalctl -t ignition` + `/var/log/azure/` capture step before anything else, otherwise the same bug recurs and we destroy the evidence each time.

2. **Say how to re-image, safely.** In an AKS context "re-imaged" most naturally reads as `az vmss reimage` against the `MC_*` resource group, which is unsupported on AKS-managed scale sets and can leave the pool unreconcilable. Could we point explicitly at `az aks nodepool upgrade --node-image-only` (or cordon/drain + `kubectl delete node` and let it scale back), with a note to avoid direct VMSS operations?

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+those distros will **not** work on ACL.
+
+## What is ACL?
+
+Azure Container Linux is a container-host distro composed of Azure Linux 3.0
+packages, assembled into a UKI-based (Unified Kernel Image) immutable image
+using Flatcar Container Linux build scripts and composition tooling.
+
+Key properties:
+- **Immutable `/usr`** — the root filesystem is read-only; no package manager (`dnf`, `rpm`, `tdnf`) is available on the host.
```
jiria: Tiny wording nit on the very first bullet:

> **Immutable `/usr`** — the root filesystem is read-only

The label says `/usr` but the sentence widens it to the whole root filesystem. Only `/usr` is read-only — `/etc`, `/opt` and `/var` are writable, which the doc relies on later (`/etc/extensions/<name>.raw`, `/opt/<name>/downloads/`, "Write to `/opt` (writable)").

Minor, but it's the reader's first impression of the platform, and someone taking it literally might not think to go poke at `/etc/extensions/` and `/opt/<name>/downloads/` — which are the two most useful directories in the sysext section. Suggest:

> **Immutable `/usr`** — `/usr` is mounted read-only and changes only via image updates or sysexts; `/etc`, `/opt` and `/var` remain writable. No package manager (`dnf`, `rpm`, `tdnf`) is available on the host.

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+```bash
+sudo journalctl | grep avc
+```
+
+**Temporarily set permissive (does not persist across reboots):**
+```bash
+sudo setenforce 0   # permissive
+sudo setenforce 1   # back to enforcing
+getenforce           # check current mode
+```
```
jiria: A couple of small things in this section — neither blocking:

1. **`sudo journalctl | grep avc`** pipes the whole journal through grep (slow on a busy node) and substring-matches anything containing "avc". `sudo journalctl -t audit -g avc --since "1 hour ago"` is faster and more precise.

2. **`setenforce 0`** is listed as a plain step with no caveat. Given ACL ships enforcing by default and that's part of the point of the platform, worth framing it explicitly as a *diagnostic* — "confirm the failure disappears, then `setenforce 1` immediately" — so it doesn't quietly become the standard workaround.

jiria:
Tiny thing — the TSG link in the PR description came out as a VS Code webview URL rather than the real link:

```
[ACL TSG](vscode-file://vscode-app/c:/Users/ankitapareek/AppData/Local/Programs/Microsoft%20VS%20Code/...)
```

It won't resolve for anyone, and it has your local path in it on a public repo. The References section in the file itself has the correct URL — probably just worth pasting that one into the description instead.

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — moved to `.github/skills/acl-troubleshooting/SKILL.md` with YAML frontmatter, matching the `code-review` skill pattern.

doc/skills/acl-troubleshooting.md
Ankita13-code: Fixed — now reads: "Immutable `/usr` — `/usr` is mounted read-only and changes only via image updates or sysexts; `/etc`, `/opt`, and `/var` remain writable."

doc/skills/acl-troubleshooting.md
Ankita13-code: Added an "Installed when" column to the table — kubelet/kubectl/credential-provider are Node CSE, aks-secure-tls-bootstrap-client is VHD build only, GPU drivers are Node CSE.

doc/skills/acl-troubleshooting.md
Ankita13-code: Fixed — added "Installed when" column with **VHD build only** and a note that pull failures surface in the VHD build pipeline, not node provisioning logs.

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — added UKI boot as a key property with details on `firstboot.addon.efi`, systemd-boot addon loading, and the `cleanup-vhd.sh` restore step.

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — added a full "GPU sysext selection, pull, and activation" section covering driver type resolution (grid/grid-v20/cuda/cuda-open), version resolution via `getACLVersionID()`, registry path pattern, and the pull+activation flow.

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — nodepool tag section now explicitly says "this is the recommended approach as it is durable across image updates and node reimages".

doc/skills/acl-troubleshooting.md
Ankita13-code: Corrected — now reads: "This local `/etc` change is lost when AKS reimages or replaces the node; use the node-pool tag below for durable configuration."

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — added `az aks nodepool upgrade --node-image-only` command and a note that the tag is consumed at provisioning time, so existing nodes need reimage.

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — `nodepool add` is now the primary example. `nodepool update` is behind a warning about `--tags` replacing the entire tag set, with guidance to read existing tags first. Also added `az aks nodepool upgrade --node-image-only` for applying to existing pools.

doc/skills/acl-troubleshooting.md
Ankita13-code: Both fixed:
1. Failure mode now reads: "Fails fast into `emergency.target`; writes from earlier stages are already applied (not transactional)"
2. Changed to "running as systemd units in the initramfs before switch-root"

doc/skills/acl-troubleshooting.md
Ankita13-code: Done — added log collection step (`journalctl -t ignition` + `/var/log/azure/`) before any reimage, and replaced "re-imaged" with explicit `az aks nodepool upgrade --node-image-only` / `az aks nodepool delete-machines` with a warning against `az vmss reimage`.

doc/skills/acl-troubleshooting.md
Ankita13-code: Both fixed:
1. Changed to `sudo journalctl -t audit -g avc --since "1 hour ago"`
2. Reframed `setenforce 0` as diagnostic: "permissive — for diagnosis only" with "restore enforcing immediately" after.

Ankita13-code:
Fixed — removed all `dev.azure.com` URLs from the file (they were also failing the `no-internal-hostnames` CI check). The PR description has been updated to use plain text references instead.

.github/skills/acl-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
+---
```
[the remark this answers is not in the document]
Ankita13-code: Resolved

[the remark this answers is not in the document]
Ankita13-code: Addressed — file has been at `.github/skills/acl-troubleshooting/SKILL.md` since the previous push.

.github/skills/acl-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
...
+### ACL stubs
+
+Many package-management functions are stubbed out in `cse_helpers_acl.sh`
+because ACL has no package manager. Functions like `apt_get_update()`,
+`holdWALinuxAgent()`, etc. are no-ops on ACL. If you need to install software
+on ACL, it must be done via sysexts or baked into the VHD image.
+
+## ACL-specific error codes
+
+These CSE exit codes indicate ACL-specific failures:
```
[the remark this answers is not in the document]
Ankita13-code: Fixed — renamed to "Sysext-related error codes on ACL" and noted that code 231 is also used by Flatcar.

.github/skills/acl-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
...
+| Component | Registry path | Installed when | Notes |
+|-----------|--------------|----------------|-------|
+| kubelet | `mcr.microsoft.com/oss/v2/kubernetes/kubelet-sysext` | Node CSE | Tagged by k8s version |
+| kubectl | `mcr.microsoft.com/oss/v2/kubernetes/kubectl-sysext` | Node CSE | Tagged by k8s version |
+| azure-acr-credential-provider | `mcr.microsoft.com/oss/v2/kubernetes/azure-acr-credential-provider-sysext` | Node CSE | Tagged by version |
+| aks-secure-tls-bootstrap-client | `mcr.microsoft.com/aks-secure-tls-bootstrap/v2/aks-secure-tls-bootstrap-client-sysext` | **VHD build only** | Pull failures surface in the VHD build pipeline, not in node provisioning logs |
+| GPU drivers (NVIDIA) | `mcr.microsoft.com/azurelinux/<major.minor>/azure-container-linux/<sysext-name>` | Node CSE | Tagged by `VERSION_ID` from `/etc/os-release` |
+
+After sysext merge, symlinks are created (e.g., `/usr/bin/kubelet` →
+`/opt/bin/kubelet`) and `systemd-sysext --no-reload refresh` activates them.
```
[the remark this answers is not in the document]
Ankita13-code: Fixed — corrected to `/opt/bin/kubelet` → `/usr/bin/kubelet`.

.github/skills/acl-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
...
+
+Ignition is **strictly one-shot** — it never re-runs.
+
+**When a node fails to provision:**
+1. **Collect logs first** — the Ignition journal and `/var/log/azure/` live on
+   the node and will be destroyed if you re-image without capturing them:
+   ```bash
+   # From a debug pod or serial console:
+   journalctl -t ignition > /tmp/ignition.log
+   cp -r /var/log/azure/ /tmp/azure-logs/
```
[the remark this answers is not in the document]
Ankita13-code: Fixed — separated serial console and debug pod contexts. Debug pod path now uses `chroot /host`.

.github/skills/acl-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
...
+   ```
+2. **Re-image safely** — in AKS, use `az aks nodepool upgrade --node-image-only`
+   or cordon/drain + `kubectl delete node` and let the pool scale back.
```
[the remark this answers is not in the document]
Ankita13-code: Fixed — replaced with `az aks nodepool delete-machines` and clarified that `kubectl delete node` only removes the API object.

.github/skills/acl-troubleshooting/SKILL.md
```diff
@@ -0,0 +1,324 @@
...
+sudo setenforce 1   # restore enforcing immediately
+getenforce           # verify current mode
+```
+
+**Persistently change mode** (edit `/etc/selinux/config`):
+```ini
+SELINUX=permissive   # or enforcing
+```
+Requires reboot. With the current ACL update model, `/etc` changes survive
+image-based updates.
```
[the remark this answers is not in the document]
Ankita13-code: Fixed — now reads: "This local `/etc` change is lost when AKS reimages or replaces the node; use the node-pool tag below for durable configuration."

doc/skills/acl-troubleshooting.md
```diff
@@ -0,0 +1,273 @@
...
+
+| Component | Registry path | Notes |
+|-----------|--------------|-------|
+| kubelet | `mcr.microsoft.com/oss/v2/kubernetes/kubelet-sysext` | Tagged by k8s version |
+| kubectl | `mcr.microsoft.com/oss/v2/kubernetes/kubectl-sysext` | Tagged by k8s version |
+| azure-acr-credential-provider | `mcr.microsoft.com/oss/v2/kubernetes/azure-acr-credential-provider-sysext` | Tagged by version |
+| aks-secure-tls-bootstrap-client | `mcr.microsoft.com/aks-secure-tls-bootstrap/v2/aks-secure-tls-bootstrap-client-sysext` | Tagged by version |
+| GPU drivers (NVIDIA) | `mcr.microsoft.com/azurelinux/<major.minor>/azure-container-linux/<sysext-name>` | Tagged by `VERSION_ID` from `/etc/os-release` |
+
+After sysext merge, symlinks are created (e.g., `/usr/bin/kubelet` → `/opt/bin/kubelet`).
```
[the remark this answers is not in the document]
Ankita13-code: Fixed — symlink direction corrected.

PR #8777, opened 2026-06-26
docs(kubelet): clarify config file generation precedence with tests and variable naming

abigailliang-aks-sig-node:
## Summary

Document and test the existing kubelet config file generation precedence model in the legacy CSE path. Adds variable rename (kubeletConfig → flagConfig) for clarity, error handling for json.MarshalIndent, and unit tests explicitly validating the three-layer precedence: CustomKubeletConfig wins > flags backfill > v1beta1 defaults. No behavioral changes. Scriptless path flag→config file sync to follow in a separate PR.

### Why this matters

When kubelet config file mode is enabled, translated flags (e.g. `--max-pods`, `--cluster-dns`, `--event-qps`) are **removed from CLI** and only exist in the config file. The config file needs ~40 fields populated, but `CustomKubeletConfig` only covers ~12 user-configurable fields. The remaining ~28 fields must come from the RP-provided `KubeletConfig` flag map — otherwise kubelet falls back to v1beta1 defaults which are often not what AKS intends.

### Precedence model

```
┌─────────────────────────────────────────────────────────────────┐
│              kubelet config file generation                       │
│                                                                   │
│  Priority 1 (highest): CustomKubeletConfig                        │
│  ┌─────────────────────────────────────────────────────────┐      │
│  │ User-specified via AKS API (~12 fields)                 │      │
│  │ e.g. ImageGcHighThreshold=90, SeccompDefault=true       │      │
│  │                                                         │      │
│  │ → wins when both sources set the same field             │      │
│  └─────────────────────────────────────────────────────────┘      │
│                          │                                        │
│                          ▼                                        │
│  Priority 2: KubeletConfig flags                                  │
│  ┌─────────────────────────────────────────────────────────┐      │
│  │ RP-provided defaults from CLI flag map (~40 fields)     │      │
│  │ e.g. --max-pods=110, --cluster-dns=172.16.0.10          │      │
│  │                                                         │      │
│  │ → backfills the ~28 fields not set by CustomKubeletConfig│     │
│  └─────────────────────────────────────────────────────────┘      │
│                          │                                        │
│                          ▼                                        │
│  Priority 3 (lowest): kubelet v1beta1 defaults                    │
│  ┌─────────────────────────────────────────────────────────┐      │
│  │ Built-in kubelet defaults                               │      │
│  │ e.g. eventRecordQPS=5, imageGCHighThreshold=85          │      │
│  │                                                         │      │
│  │ → used when neither source sets the field               │      │
[...]

PR #8793, opened 2026-06-29
feat(kubelet): sync translated flags to config file in scriptless path

abigailliang-aks-sig-node:
## Summary

In the scriptless (aks-node-controller) path, sync translated flags from `KubeletFlags` into `KubeletConfigFileConfig` when the config file field is nil/unset. This prevents kubelet from falling back to v1beta1 defaults during the RP migration period as a **purely defensive backfill layer**. 

### Context: Enabling Kubelet Config File by Default

Previously, most kubelet settings were passed to kubelet through command-line flags. The RP side has moved most supported kubelet settings (except a small set that still need to remain as flags) into the kubelet configuration file.

The RP-side PR ([aks-rp #16468627](https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/16468627)) changes the default behavior for Kubernetes 1.36+ clusters by enabling the kubelet config file path by default. Instead of primarily relying on command-line flags, kubelet settings will now be applied through the kubelet configuration file. This aligns with the upstream Kubernetes direction of moving kubelet configuration away from command-line flags and toward the structured kubelet config file.

#### Relationship to Custom Node Configuration

`EnableKubeletConfigFile` is a lower-level mechanism that controls **how** kubelet settings are delivered to kubelet:

- **When enabled**: kubelet settings are written to the kubelet configuration file, and settings that have been translated to the config file are omitted from kubelet command-line flags.
- **When disabled**: kubelet settings are passed through command-line flags instead.

Custom Node Configuration is one consumer of this mechanism — custom kubelet settings need to be represented in the kubelet config file. That's why `EnableKubeletConfigFile` is already enabled whenever a customer specifies `CustomKubeletConfig` or `CustomLinuxOSConfig`.

The RP-side PR expands that mechanism beyond the custom node configuration scenario. For Kubernetes 1.36+ clusters, we enable the kubelet config file path by default, even when the customer isn't using custom node configuration.

#### Customer Impact

Customers shouldn't see any difference from a functionality perspective. They can already configure the same kubelet settings today. What's changing is the implementation under the hood — instead of passing settings via kubelet command-line flags by default, we write them to the kubelet config file. The customer-facing behavior stays the same; we're just changing the default delivery mechanism internally.

#### Kubelet Settings That Remain as CLI Flags

There are 12 kubelet options that still remain as CLI flags (detailed in [#8777](https://github.com/Azure/AgentBaker/pull/8777)). They fall into a few categories:

- **Bootstrap dependencies**: `--kubeconfig`, `--bootstrap-kubeconfig` — needed before kubelet can read its configuration
- **Settings not part of KubeletConfiguration schema**: `--node-ip`, `--node-labels`
- **System-level/runtime settings**: `--container-runtime-endpoint`

> The simple way to think about it: you can't put "where to find the config" inside the config itself (circular dependency).

---

### Rationale for This PR (AgentBaker Side)

RP is the authoritative source for the final kubelet config file. As stated in [node_config.go:353](https://github.com/Azure/AgentBaker/blob/4ebd7b12a31832c62a34cf8ce125f3a118c7eb92/e2e/node_config.go#L354-L355):
> "we require client (e.g. AKS-RP) to provide the final kubelet config that is ready to be written to the final kubelet config file on a node."

RP-side flow ([scriptless_direct.go](https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/commit/bfb6b309529f9a1e1518c5b577f3a62edf4c3022?refName=refs/heads/master&path=/resourceprovider/sharedlib/agentbaker/scriptless_direct.go&_a=contents&tab=details) — CreateAKSNodeConfig)
```
  Step 1: Build base KubeletConfig (flag map) from system defaults
          ├── staticLinuxKubeletConfig (non-overridable: --address, --authorization-mode, --client-ca-file...)
          └── defaultKubeletConfig (overridable: --event-qps, --image-gc-*, --cluster-domain...)

  Step 2: processKubeletConfigFileContent()
          ├── 2a. Map ~30 translated flags from flag map → KubeletConfigFileConfig proto fields
          └── 2b. CustomKubeletConfig overlays (non-nil fields win)

  Step 3: processKubeletFlags()
          └── filterKubeletFlags(kubeletConfig, enableKubeletConfigFile)
              When enableKubeletConfigFile=true:
                → strips all translatedKubeletConfigFlags from KubeletFlags (they already live in config file)
              When enableKubeletConfigFile=false:
                → keeps everything in KubeletFlags

  Step 4: Assemble proto + post-assembly opt functions
          └── Opts (e.g. withAKSNodeConfigDefaultContainerLogMaxSize) write BOTH sides simultaneously

  Key: enableKubeletConfigFile = (CustomKC != nil) || (CustomLinuxOSConfig != nil)
```
───────────────────────────────────────────┤

### Precedence model

1. `KubeletConfigFileConfig` field already set → keep it (RP explicitly set it)
2. `KubeletFlags` has the corresponding translated flag → backfill into config file
3. Neither set → kubelet uses v1beta1 default

Existing `KubeletConfigFileConfig` values are never overwritten. Flags remain in `KubeletFlags` (not removed) for backward compatibility since CLI flags have higher runtime precedence.

**Why no conflicts arise**
```
  ┌───────────────────────────────────┬───────────────────────────────────┬────────────────────────────────────┬────────────┬────────────────────────┐
  │             Scenario              │ Translated flags in KubeletFlags? │ KubeletConfigFileConfig has value? │ Conflict?  │     sync behavior      │
  ├───────────────────────────────────┼───────────────────────────────────┼────────────────────────────────────┼────────────┼────────────────────────┤
  │ CustomKC present (enableKCF=true) │ ❌ Stripped by Step 3             │ ✅ Set by CustomKC                 │ None       │ Nothing to backfill    │
  ├───────────────────────────────────┼───────────────────────────────────┼────────────────────────────────────┼────────────┼────────────────────────┤
  │ CustomKC absent (enableKCF=false) │ ✅ Retained                       │ May be nil                         │ None       │ Backfills nil ✅       │
  ├───────────────────────────────────┼───────────────────────────────────┼────────────────────────────────────┼────────────┼────────────────────────┤
  │ Post-assembly opt dual-writes     │ ✅ Opt writes flag                │ ✅ Opt writes config               │ Consistent │ Sees non-nil, skips ✅ │
  └───────────────────────────────────┴───────────────────────────────────┴────────────────────────────────────┴────────────┴────────────────────────┘
```
### Changes

- **`aks-node-controller/parser/helper.go`**: Add `syncTranslatedFlagsToConfigFile()` covering all ~40 translated flag types (string, *int32, int32, *bool, bool, []string, map[string]string, map[string]bool, nested auth/authz structs). Called in `getKubeletConfigFileContent()` before marshaling.
- **`aks-node-controller/parser/helper_test.go`**: Add 4 unit tests validating backfill, no-override, nil safety, and end-to-end integration.

### Depends on

- #8777 (legacy CSE path precedence documentation)

### Test Plan

```bash
cd aks-node-controller && go test ./parser/ -run "syncTranslatedFlags|getKubeletConfigFileContent" -v
```

E2E: Use `AKSNodeConfigMutator` to set `KubeletConfigFileConfig.EventRecordQps = nil` + `KubeletFlags["--event-qps"] = "0"`, verify `/etc/default/kubeletconfig.json` has `"eventRecordQPS": 0`.


aks-node-controller/parser/helper.go
```diff
@@ -677,6 +681,179 @@ func getKubeletConfigFileContent(kubeletConfig *aksnodeconfigv1.KubeletConfig) s
...
+		}
+	}
+	backfillInt32Ptr("--event-qps", &cfg.EventRecordQps)
+	backfillInt32Ptr("--image-gc-high-threshold", &cfg.ImageGcHighThresholdPercent)
+	backfillInt32Ptr("--image-gc-low-threshold", &cfg.ImageGcLowThresholdPercent)
+	backfillInt32Ptr("--max-pods", &cfg.MaxPods)
+	backfillInt32Ptr("--pod-max-pids", &cfg.PodPidsLimit)
+	backfillInt32Ptr("--container-log-max-files", &cfg.ContainerLogMaxFiles)
+
+	// Non-optional int32 fields (ReadOnlyPort, EvictionMaxPodGracePeriod):
```
Devinwong, line 752: the optional/ non-optional parts are complicated.
Is it just these 2 fields ReadOnlyPort, EvictionMaxPodGracePeriod not backfilled? or all non-optional fields not backfilled?

abigailliang-aks-sig-node: It comes down to the differences in the proto definitions:
```
// Non-optional — Go generates `int32` (the zero value `0` makes it impossible to distinguish between "unset" and "explicitly set to 0")
int32 read_only_port = 5;
int32 eviction_max_pod_grace_period = 43;

// Optional — Go generates `*int32` (`nil` = unset, `&0` = explicitly set to 0)
optional int32 event_record_qps = 13;
optional int32 max_pods = 24;
optional int32 image_gc_high_threshold_percent = 19;
```
This is a legacy issue with the proto definitions. The correct approach is for any field that might be `0`.

Devinwong: sounds good

Devinwong (dismissed):

aks-node-controller/parser/helper.go
```diff
@@ -677,6 +681,203 @@ func getKubeletConfigFileContent(kubeletConfig *aksnodeconfigv1.KubeletConfig) s
...
+	backfillBoolPtr("--cgroups-per-qos", &cfg.CgroupsPerQos)
+	backfillBoolPtr("--cpu-cfs-quota", &cfg.CpuCfsQuota)
+	backfillBoolPtr("--fail-swap-on", &cfg.FailSwapOn)
+	backfillBoolPtr("--serialize-image-pulls", &cfg.SerializeImagePulls)
+
+	// Non-optional bool fields (RotateCertificates, ServerTlsBootstrap, ProtectKernelDefaults):
+	// NOT backfilled. Proto3 non-optional bool uses false as zero value, making it
+	// impossible to distinguish "explicitly set to false" from "never set". Backfilling
+	// would violate the precedence model (e.g., RotateCertificates=false is a valid
+	// explicit choice that should not be overwritten by a flag value of true).
```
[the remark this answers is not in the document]
abigailliang-aks-sig-node: add a to do comment

awesomenix (approved):

aks-node-controller/parser/helper.go
```diff
@@ -677,6 +689,232 @@ func getKubeletConfigFileContent(kubeletConfig *aksnodeconfigv1.KubeletConfig) s
...
+	backfillString("--node-status-update-frequency", &cfg.NodeStatusUpdateFrequency)
+	backfillString("--node-status-report-frequency", &cfg.NodeStatusReportFrequency)
+	backfillString("--kube-reserved-cgroup", &cfg.KubeReservedCgroup)
+	backfillString("--system-reserved-cgroup", &cfg.SystemReservedCgroup)
+
+	// Optional *int32 fields — backfill if nil.
+	backfillInt32Ptr := func(flag string, field **int32) {
+		if *field == nil {
+			if v, ok := flags[flag]; ok && v != "" {
+				if n, err := strconv.ParseInt(v, 10, 32); err == nil {
```
awesomenix, line 736: silently failing here could be a problem, what if there was expectation on this being synced but its no longer present. either we can error here or log a warning.

abigailliang-aks-sig-node: done.

noted this is a pure parse error handler — if `strconv.ParseInt/strconv.ParseBool` fails. AgentBaker's stance is to  accept whatever RP sends without validation(correct me if i goes wrong); it doesn't own the definition of  what should be in the config file or what the correct values are. 

Correctness of  flag values is RP's responsibility and should be covered by RP-side unit tests.  If we want additional defensive coverage on our side, that would be an e2e test — but it's not this function's job to enforce.

lilypan26:
im a bit skeptical the effectiveness of the backfilling. in the non-scriptless path (in agentbaker) and in the scriptless path (in RP), we remove `TranslatedKubeletConfigFlags` from kubeletflags when kubeletconfigfile is enabled. So won't the backfilling always be a no-op?

Devinwong:
> im a bit skeptical the effectiveness of the backfilling. in the non-scriptless path (in agentbaker) and in the scriptless path (in RP), we remove `TranslatedKubeletConfigFlags` from kubeletflags when kubeletconfigfile is enabled. So won't the backfilling always be a no-op?

This is a good call.
Seems it will be an no-op. 
This is a decision tree diagram that copilot generated.
{098C5839-B85D-4338-B8C2-F6E482D33943}

Curious: where did this request come from?

abigailliang-aks-sig-node:
> im a bit skeptical the effectiveness of the backfilling. in the non-scriptless path (in agentbaker) and in the scriptless path (in RP), we remove `TranslatedKubeletConfigFlags` from kubeletflags when kubeletconfigfile is enabled. So won't the backfilling always be a no-op?

@lilypan26 

yes.

According to the [1st pr](https://github.com/Azure/AgentBaker/pull/8777) (also on the 1st pharagraph of this pr summary), rp side will do the fields translated and filtered out. So this logic is no-oped. 

It's defensive against deployment ordering: 
RP and AgentBaker deploy independently, so i concern there's a window where new AgentBaker code is live but RP hasn't yet started populating `KubeletConfigFileConfig` (translated flags are still in KubeletFlags). The sync ensures nodes  provisioned during that window don't end up with an incomplete config file. Once RP fully rolls out,  it becomes a **no-op** — a few map lookups with no side effects.

reply to **Curious: where did this request come from?** this a trend that move fields from flags to file but after investigation, we found out that the migration job has been done on the rp side and only left few fields hard to migrate (eg `--config-path`). its deserve to make a deep dive and get an alignment on this job through these 2 prs.
CC @Devinwong

lilypan26:
> > im a bit skeptical the effectiveness of the backfilling. in the non-scriptless path (in agentbaker) and in the scriptless path (in RP), we remove `TranslatedKubeletConfigFlags` from kubeletflags when kubeletconfigfile is enabled. So won't the backfilling always be a no-op?
> 
> @lilypan26
> 
> yes.
> 
> According to the [1st pr](https://github.com/Azure/AgentBaker/pull/8777) (also on the 1st pharagraph of this pr summary), rp side will do the fields translated and filtered out. So this logic is no-oped.
> 
> It's defensive against deployment ordering: RP and AgentBaker deploy independently, so i concern there's a window where new AgentBaker code is live but RP hasn't yet started populating `KubeletConfigFileConfig` (translated flags are still in KubeletFlags). The sync ensures nodes provisioned during that window don't end up with an incomplete config file. Once RP fully rolls out, it becomes a **no-op** — a few map lookups with no side effects.
> 
> reply to **Curious: where did this request come from?** this a trend that move fields from flags to file but after investigation, we found out that the migration job has been done on the rp side and only left few fields hard to migrate (eg `--config-path`). its deserve to make a deep dive and get an alignment on this job through these 2 prs. CC @Devinwong

The code to remove `TranslatedKubeletConfigFlags` from kubeletFlags has present for quite some time so I personally think it would be a bit overkill to include this extra precaution. If you find flags that are not included in `TranslatedKubeletConfigFlags` then it would be worth covering those, but I think it's all covered. Can let @awesomenix  weigh in as well.

Devinwong (approved):

awesomenix:
>The code to remove TranslatedKubeletConfigFlags from kubeletFlags has present for quite some time so I personally think it would be a bit overkill to include this extra precaution. If you find flags that are not included in TranslatedKubeletConfigFlags then it would be worth covering those, but I think it's all covered

Agree with lily here didnt realize there is something like this on Agentbaker side, lets focus on trying to enable kubeletconfig file and making it default fo scriptless.

PR #8795, opened 2026-06-30
fix(windows): write TLS bootstrap-config in NodePrep, not BasePrep

benjamin-brady:
Work item: AB#38630884

## Summary

Fixes [AB#38630884](https://dev.azure.com/msazure/CloudNativeCompute/_workitems/edit/38630884): PIS / VHD-caching Windows nodes fail to register with `cannot create certificate signing request: Unauthorized`.

## Root cause

On PIS/VHD-cached clusters the Windows VHD is baked from a temporary VMSS with TLS bootstrapping enabled. During the bake, `BasePrep` writes `c:\k\bootstrap-config` with the **bake-time** TLS bootstrap token and drops a `C:\AzureData\base_prep.complete` marker — both captured into the cached VHD.

On the real node the CSE **skips `BasePrep`** because that marker exists, so `bootstrap-config` keeps the stale baked token instead of the fresh `$global:TLSBootstrapToken` delivered via CustomData. The apiserver then rejects the kubelet's first CSR as `Unauthorized`.

The bootstrap-config write was simply left in `BasePrep` when the script was split into `BasePrep`/`NodePrep` (commit `63b9d9b9`, #6601).

## Fix

Move `Write-BootstrapKubeConfig` from `BasePrep` into `NodePrep`, before `Install-KubernetesServices`. kubelet (installed/started by `Install-KubernetesServices`) is the only consumer of `bootstrap-config`, so writing it in `NodePrep` guarantees the **live** token is used on every real provision.

- **No VHD rebake required** — the Windows CSE script is rendered fresh into CustomData by the RP on every provision; only the on-disk `bootstrap-config` file is baked. The corrected `NodePrep` write takes effect on new nodes from existing cached VHDs.
- **A move (not keep-and-refresh) is correct** — `BasePrep` never starts kubelet nor connects to the API server, so nothing there consumes the token-based `bootstrap-config`. Calico in `BasePrep` uses the separate **cert-based** `config` from `Write-KubeConfig`, which is unchanged.

## Validation

- `go test ./pkg/agent/...` passes; `make generate-testdata` produced no snapshot drift.
- Suggested manual validation (from the work item):
  - Provision a Windows node from a PIS cached VHD; confirm `c:\k\bootstrap-config` token == `$global:TLSBootstrapToken` and the node reaches Ready.
  - Rotate the agent pool TLS bootstrap token, provision again from the same cached VHD, and confirm the node still joins.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/e2e/scenario_win_test.go
+++ b/e2e/scenario_win_test.go
@@ -390,7 +390,19 @@ func Test_Windows2025Gen2_VHDCaching(t *testing.T) {
 // to use the legacy bootstrap-token path. Catches regressions in the two-stage
 // CSE flow that only surface when no secure-tls-bootstrap client is around to
 // overwrite the temporary kubeconfig.
+//
+// It also positively guards the BasePrep->NodePrep kubeconfig fix: a stale
+// sentinel bootstrap token is baked during the pre-provision (BasePrep) stage,
+// while the real cluster token is used at provision time. If bootstrap-config
+// were written in BasePrep (the buggy behaviour), the cached VHD would carry the
+// stale token and the node would fail to register; because it is written in
+// NodePrep, the live token wins and the sentinel must never reach the node.
 func Test_Windows2022_VHDCaching_LegacyTLSBootstrap(t *testing.T) {
+	// Deliberately bogus but correctly-formatted ([a-z0-9]{6}.[a-z0-9]{16}) token.
+	// Baked into the VHD at BasePrep time only; must be overwritten by the live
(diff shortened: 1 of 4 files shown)


timmy-wright (approved):

timmy-wright (approved):

benjamin-brady:
The E2E tests that cover windows and windows VHD caching passed.

r2k1 (approved):

PR #8812, opened 2026-07-01
fix(kubelet): remove streamingConnectionIdleTimeout for k8s >= 1.34

abigailliang-aks-sig-node:
## Summary

Remove `streamingConnectionIdleTimeout` from both the kubelet command line and the kubelet config file for Kubernetes >= 1.34, where this field was removed from `KubeletConfiguration`.

## Context: RP-side flow (scriptless path)

```
Step 1: defaults_kubelet.go — removeKubeletFlags()
        └── k8s >= 1.34: delete(kubeletConfig, "--streaming-connection-idle-timeout")
                ↓
        kubeletConfig map (flag is already gone)
                ↓
Step 2: scriptless_direct.go — processKubeletConfigFileContent()
        └── cfg.StreamingConnectionIdleTimeout = kubeletFlags["--streaming-connection-idle-timeout"]
            → empty string (key was deleted in step 1)
                ↓
Step 3: scriptless_direct.go — processKubeletFlags()
        └── filterKubeletFlags() strips translated flags when enableKCF=true
            → flag not on command line either way
                ↓
Step 4: Proto sent to aks-node-controller
        └── KubeletConfigFileConfig.StreamingConnectionIdleTimeout = "" (omitted by protojson)
```

RP already handles this in `removeKubeletFlags()`. However, the field may still arrive in `KubeletConfigFileConfig` from:
- Older RP versions that haven't rolled out yet
- The `GetAKSNodeConfig` (NBC conversion) path where the removal may not be applied

## Changes (defense-in-depth)

| File | Change |
|------|--------|
| `aks-node-controller/helpers/utils.go` | `ValidateAndSetLinuxKubeletFlags()`: delete `--streaming-connection-idle-timeout` for >= 1.34 |
| `aks-node-controller/parser/parser.go` | `getCSEEnv()`: clear `KubeletConfigFileConfig.StreamingConnectionIdleTimeout` for >= 1.34 before marshaling |
| `pkg/agent/baker.go` | `ValidateAndSetLinuxNodeBootstrappingConfiguration()`: same flag deletion for the old NBC path |

## Why both flag deletion AND config file clearing?

- **Flag deletion** (`helpers/utils.go`, `baker.go`): covers `enableKCF=false` case where translated flags stay on the command line
- **Config file clearing** (`parser.go`): covers `enableKCF=true` case where the field would appear in `kubeletconfig.json` if RP sets it

## Test plan
- [x] unit tests
- [x] local e2e tests
- [x] CI passing: https://github.com/Azure/AgentBaker/actions/runs/29296756647/job/86971871744?pr=8812


aks-node-controller/parser/helper_test.go
```diff
@@ -1380,18 +1378,17 @@ func Test_getKubeletConfigFileContent(t *testing.T) {
 						ClusterDns: []string{
 							"10.0.0.10",
 						},
-						StreamingConnectionIdleTimeout: "4h0m0s",
-						NodeStatusUpdateFrequency:      "10s",
-						ImageGcHighThresholdPercent:    to.Ptr(int32(90)),
-						ImageGcLowThresholdPercent:     to.Ptr(int32(70)),
-						CgroupsPerQos:                  to.Ptr(true),
-						CpuManagerPolicy:               "static",
```
abigailliang-aks-sig-node, line 1549: this change block due to  gofmt/goimports column alignment.

When a struct literal contains multiple fields, gofmt attempts to align the values ​​following the colons into the same column. The width of the aligned column is determined by the longest field name in the group:

aks-node-controller/parser/helper_test.go
```diff
@@ -1449,247 +1446,6 @@ func Test_getKubeletConfigFileContent(t *testing.T) {
...
-		"--enforce-node-allocatable": "pods",
-		// map[string]string
-		"--eviction-hard":   "memory.available<750Mi,nodefs.available<10%",
-		"--system-reserved": "cpu=100m,memory=1Gi",
-		// map[string]bool
-		"--feature-gates": "RotateKubeletServerCertificate=true,DynamicKubeletConfig=false",
-		// nested
-		"--client-ca-file":               "/etc/kubernetes/certs/ca.crt",
-		"--authentication-token-webhook": "true",
-		"--authorization-mode":           "Webhook",
```
abigailliang-aks-sig-node: same gofmt behave

Devinwong:
could you clarify why k8s>1.34 needs to remove `streamingConnectionIdleTimeout`. Did it block something or just to make it right?

abigailliang-aks-sig-node:
> could you clarify why k8s>1.34 needs to remove `streamingConnectionIdleTimeout`. Did it block something or just to make it right?

I found this pr remove this argument for 1.34.0+, not for all versions:
[Pull Request 14602085](https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/pullrequest/14602085): remove deprecated --streaming-connection-idle-timeout kubelet flag for >= 1.3...

aks-node-controller/helpers/utils.go
```diff
@@ -227,6 +227,12 @@ func ValidateAndSetLinuxKubeletFlags(kubeletFlags map[string]string, cs *datamod
 		!IsKubernetesVersionGe(cs.Properties.OrchestratorProfile.OrchestratorVersion, "1.25.0") {
 		kubeletFlags["--feature-gates"] = addFeatureGateString(kubeletFlags["--feature-gates"], "DisableAcceleratorUsageMetrics", false)
 	}
+
+	// streamingConnectionIdleTimeout was removed from KubeletConfiguration in k8s 1.34+.
+	// It must not appear on the command line or in the config file for those versions.
+	if IsKubernetesVersionGe(cs.Properties.OrchestratorProfile.OrchestratorVersion, "1.34.0") {
```
lilypan26, line 236: nit: i think this whole file isn't used anymore since its all done on the RP side, but can leave that as a future TODO to clean this up

lilypan26 (dismissed):

aks-node-controller/proto/aksnodeconfig/v1/kubelet_config.proto
```diff
@@ -123,198 +123,199 @@ message KubeletConfigFileConfig {
...
-     Default: "4h"
-     +optional. */
+   can be idle before the connection is automatically closed.
+   Deprecated since k8s 1.25, removed from KubeletConfiguration in k8s 1.34.
+   This field is cleared by aks-node-controller for k8s >= 1.34 and will not
+   appear in the marshaled config file. Retained in proto for backward
+   compatibility with k8s < 1.34.
+   Default: "4h"
+   +optional. */
   string streaming_connection_idle_timeout = 16;
```
awesomenix: i think there is a way to mark proto field as deprecated.

```
deprecated (field option): If set to true, indicates that the field is deprecated and should not be used by new code. In most languages this has no actual effect. In Java, this becomes a @Deprecated annotation. For C++, clang-tidy will generate warnings whenever deprecated fields are used. In the future, other language-specific code generators may generate deprecation annotations on the field’s accessors, which will in turn cause a warning to be emitted when compiling code which attempts to use the field. If the field is not used by anyone and you want to prevent new users from using it, consider replacing the field declaration with a [reserved](https://protobuf.dev/programming-guides/proto3/#fieldreserved) statement.
```

`int32 old_field = 6 [deprecated = true];`

pkg/agent/baker.go
```diff
@@ -638,6 +638,12 @@ func ValidateAndSetLinuxNodeBootstrappingConfiguration(config *datamodel.NodeBoo
 		!IsKubernetesVersionGe(config.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion, "1.25.0") {
 		kubeletFlags["--feature-gates"] = addFeatureGateString(kubeletFlags["--feature-gates"], "DisableAcceleratorUsageMetrics", false)
 	}
+
+	// streamingConnectionIdleTimeout was removed from KubeletConfiguration in k8s 1.34+.
+	// It must not appear on the command line or in the config file for those versions.
+	if IsKubernetesVersionGe(config.ContainerService.Properties.OrchestratorProfile.OrchestratorVersion, "1.34.0") {
```
awesomenix, line 678: There is also this place

https://github.com/Azure/AgentBaker/blob/b9e6be3f4e9345d3e6a21d967eaddf36c865873c/pkg/agent/datamodel/types.go#L136

abigailliang-aks-sig-node: It’s already handled in[ aks-node-controller/parser/parser.go ](https://github.com/Azure/AgentBaker/blob/remove-streaming-connection-idle-timeout/aks-node-controller/parser/parser.go#L38)as well. Baker removes the flag before generating the CSE, and ANC removes it again during node-side parsing for defense in depth. The field in types.go is retained for backward compatibility with older Kubernetes versions, so we don’t need to change it.

awesomenix (approved):

yongkangchen (approved):

abigailliang-aks-sig-node:
> ### AgentBaker Linux gate detective
> **Run:** https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=172264217 **Failed job/stage/task:** �2e / Run AgentBaker E2E / Run AgentBaker E2E
> 
> **TL;DR:** Ubuntu 24.04 Gen2 E2E package validation reached the node and then failed because expected moby-containerd 2.3.2-ubuntu24.04u1 was missing from installed packages.
> 
> **Likely cause/signature:** deterministic VHD/package-content or expectation skew ($e2eSig).
> 
> **Confidence:** High.
> 
> **Recommended owner/action:** VHD/package owner should compare Ubuntu 24.04 package feed/build inputs against installed package inventory and either fix the package source/image content or correct expected metadata.
> 
> **Strongest alternative:** test expectation advanced before package/feed availability; less likely to be PR-change-caused because PR #8812 changes kubelet streaming idle timeout logic, not containerd packaging.
> 
> **Evidence:** E2E est-log.json reports the package assertion; sibling Ubuntu 24.04 Gen2 leaves show the same signature; retry evidence is deterministic rather than transient.
> 
> **Wiki signature:** [ubuntu2404-gen2-moby-containerd-232-ubuntu2404u1-missing](https://dev.azure.com/msazure/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki?pagePath=%2FAKS%2FSIGs%20and%20Teams%2FAKS%20Components%2FSIG%3A%20Node%20Lifecycle%2FAI%20Agent%20Knowledge%2FAgentBaker%20Gate%20PR%20Pipeline%20Flakiness%2FNoisy%20Active%20Items)

fix at https://github.com/Azure/AgentBaker/pull/8959

aks-node-controller/helpers/utils_test.go
```diff
@@ -397,6 +397,37 @@ func TestValidateAndSetLinuxKubeletFlags_FeatureGatesByVersion(t *testing.T) {
...
 
+func TestValidateAndSetLinuxKubeletFlags_StreamingConnectionIdleTimeout(t *testing.T) {
+	testCases := []struct {
+		name          string
+		version       string
+		expectRemoved bool
+	}{
+		{"k8s 1.33 keeps flag", "1.33.0", false},
+		{"k8s 1.34 removes flag", "1.34.0", true},
+		{"k8s 1.35 removes flag", "1.35.0", true},
```
cameronmeissner, line 408: nit: I'd probably also add a test case for 1.36

cameronmeissner (approved):

PR #8823, opened 2026-07-02
test(e2e): serialize gallery image-version replication to fix GalleryImageNotFound

ganeshkumarashok:
## Problem

The GPU E2E suite has been failing with **`GalleryImageNotFound`** (20 failures in the latest `main` run) — e.g.:

```
The gallery image .../images/2404gen2containerd/versions/1.1783016979.17372
is not available in southeastasia region. Please contact image owner to
replicate to this region, or change your requested region.
```

All 20 were the **same version** in the **same region** (southeastasia).

## Root cause — a concurrent lost-update race on `TargetRegions`

`CachedPrepareVHD` memoizes VHD resolution per **(image, region)**, so different regions resolve the *same* freshly-built SIG image version **in parallel**. Adding a region is a **read-modify-write PUT** of the version's `TargetRegions` (`replicateImageVersionToCurrentRegion`), and there was **no locking**:

1. southeastasia reads the version (regions `[eastus]`), appends → PUT `[eastus, southeastasia]`.
2. uaenorth reads the version (also `[eastus]`, before step 1 lands), appends → PUT `[eastus, uaenorth]`.
3. uaenorth's stale PUT **drops southeastasia** → the image is no longer replicated there → every cached southeastasia scenario fails with `GalleryImageNotFound`.

The run log confirms it: southeastasia **and** uaenorth replicated version `1.1783016979.17372` concurrently.

## Fix

Serialize replication **per image-version ID** with an in-process mutex, and **re-read the live `TargetRegions` inside the lock** before each update, so every PUT builds on the current region set instead of clobbering a concurrent addition. Distinct versions still replicate in parallel — only same-version region additions serialize (which Azure requires anyway, since a version resource only accepts one update at a time).

## Scope / testing

- e2e-only change (`e2e/config/azure.go`); no provisioning/runtime code touched.
- `go build ./config/` + `go vet ./config/` pass.
- Trade-off: same-version replications now run sequentially (only on the first run after a new VHD build, for the ~2–3 regions a version is used in) — correctness over a small first-run latency; the tests already block on replication today.

### DIFF
--- a/e2e/config/azure.go
+++ b/e2e/config/azure.go
@@ -12,6 +12,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Azure/agentbaker/e2e/toolkit"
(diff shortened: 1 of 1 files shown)


ganeshkumarashok:
Rebased onto current `main` and refreshed — this is still the highest-value fix for GPU E2E.

**Fresh data (2026-07-27, def 443878):** `main` is nominally green on 07-25/26/27 but every "green" run still carries **3–5 persistent GPU test failures** that don't gate the pipeline. The dominant signature is now `GalleryImageNotFound` on Ubuntu2404 GPU scenarios (H100, A100, RTXPro6000_GridV20, MIG/MIG_Mixed) running in remote regions (westus2, southcentralus, southeastasia, UAENorth) — verified in builds 173908127, 173825038, 173761954.

Root cause is exactly what this PR fixes: `ensureReplication` does a read-modify-write of the shared image version's `TargetRegions`, and with `-parallel 60` (single job, 60 goroutines) scenarios in different regions clobber each other's region additions → the image ends up unavailable in a dropped region. The per-version mutex + re-read-under-lock closes that race.

Verified: cherry-picks clean onto current `main`, `go build ./...` passes. Complements #8374 (regional-replication-state wait), which addresses the separate eventual-consistency gap.

PR #8834, opened 2026-07-04
fix: localdns readiness curl must bypass HTTP proxy on http-proxy clusters

yewmsft:
> **Incident repair item** for ICM [51000001093225](https://portal.microsofticm.com/imp/v5/incidents/details/51000001093225/summary).

## Summary

On AKS clusters configured with an **HTTP proxy**, enabling **LocalDNS** causes node bootstrap (CSE) to hang until timeout, failing node provisioning (create / scale / node-image upgrade / surge). Root cause is a shell-quoting bug in the LocalDNS readiness health check.

- `localdns.service` inherits the system-wide proxy env (via systemd `DefaultEnvironment` on http-proxy clusters). The readiness probe curls the **node-local** listener `169.254.10.10:8181/ready`, which must **not** be sent through the proxy — a link-local address is not routable by an external HTTP proxy.
- `CURL_COMMAND` was a **string** invoked unquoted (`$($CURL_COMMAND)`). Word-splitting happens **without quote-removal**, so any attempt to pass `--noproxy '*'` reached curl as a literal 3-char argument `'*'`, leaving the proxy bypass ineffective. The probe was proxied, never returned `OK`, and CSE blocked until timeout.
- Reproduces regardless of outbound type (**UDR or LoadBalancer**) — the only conditions are "HTTP proxy configured" + "LocalDNS enabled".

## Fix

- Declare `CURL_COMMAND` as a **bash array** and expand as `"${CURL_COMMAND[@]}"` at both call sites (`wait_for_localdns_ready`, `start_localdns_watchdog`) so arguments are passed verbatim with correct quoting.
- Add an explicit `--noproxy "${LOCALDNS_NODE_LISTENER_IP}"` so the node-local readiness request always bypasses the proxy, plus `--connect-timeout`/`--max-time` so a probe can't hang.
- Update the `wait_for_localdns_ready` ShellSpec stubs to array form (`CURL_COMMAND=(echo OK)`), required by the new expansion.

## Validation

- Reproduced the failure and confirmed the fix end-to-end on http-proxy clusters (both UDR and LoadBalancer outbound): with the fix, the surge node bootstraps and joins with LocalDNS enabled — **without** adding link-local IPs to `noProxy`.
- `make generate-testdata` — all `pkg/agent/...` tests pass, no generated-file diff (localdns.sh is not embedded in the CSE snapshot).
- ShellSpec `localdns_spec.sh`: **85 examples, 0 failures**.
- `bash -n` syntax check on both files.

## Test plan

- [x] ShellSpec `localdns_spec.sh` passes (85/0)
- [x] `pkg/agent/...` go tests pass, no snapshot drift
- [x] e2e: LocalDNS-enabled nodepool on an http-proxy cluster bootstraps successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -22,10 +22,10 @@ spec/parts/linux/cloud-init/artifacts/ubuntu-snapshot-update_spec.sh @yewmsft @Y
 spec/parts/linux/cloud-init/artifacts/mariner-package-update_spec.sh @yewmsft @YaoC
 
 # Code owners for localdns.
-parts/linux/cloud-init/artifacts/localdns.sh @SriHarsha001 @yewmsft
-parts/linux/cloud-init/artifacts/localdns.service @SriHarsha001 @yewmsft
-parts/linux/cloud-init/artifacts/localdns-delegate.conf @SriHarsha001 @yewmsft
-spec/parts/linux/cloud-init/artifacts/localdns_spec.sh @SriHarsha001 @yewmsft
+parts/linux/cloud-init/artifacts/localdns.sh @saewoni @yewmsft
+parts/linux/cloud-init/artifacts/localdns.service @saewoni @yewmsft
+parts/linux/cloud-init/artifacts/localdns-delegate.conf @saewoni @yewmsft
+spec/parts/linux/cloud-init/artifacts/localdns_spec.sh @saewoni @yewmsft
 
 # Set components.json & windows_settings.json to no code-owner so that it can be approved and merged by component owner.
 parts/common/components.json
--- a/parts/linux/cloud-init/artifacts/localdns.sh
+++ b/parts/linux/cloud-init/artifacts/localdns.sh
@@ -54,7 +54,10 @@ RESOLV_CONF="/run/systemd/resolve/resolv.conf"
 
 # Curl check if localdns is running.
 # This is used by start_localdns_watchdog and wait_for_localdns_ready.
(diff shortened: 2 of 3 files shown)


awesomenix (approved):

cameronmeissner (approved):

cameronmeissner (approved):

Devinwong (approved):

saewoni (approved):

awesomenix (approved):

saewoni:
### Code review

No issues found.

Checks performed:
- Reviewed current PR head `8bea7633906b7739317d7e2ede86f1d337d13e5f`; PR is open and not draft.
- Reviewed changed files: `CODEOWNERS`, `parts/linux/cloud-init/artifacts/localdns.sh`, and `spec/parts/linux/cloud-init/artifacts/localdns_spec.sh`.
- Checked `CLAUDE.md` guidance for provisioning-script changes, especially shell portability and regression risk.
- Verified the `CURL_COMMAND` conversion to a bash array is applied consistently at both call sites and in ShellSpec stubs.
- Checked `bash -n` syntax for `localdns.sh` and `localdns_spec.sh`.
- Verified the introduced curl flags (`--noproxy`, `--connect-timeout`, `--max-time`) are supported by curl on the MLVM.
- Reviewed recent LocalDNS history and prior review comments for applicable regression patterns.

SriHarsha001 (approved):

PR #8839, opened 2026-07-06
feat(gpu): enable Grace-Blackwell (GB200/GB300) via IMEX, Fabric Mana…

xuexu6666:
…ger off

GB has no on-board NVSwitch (switch trays live in the rack, run by NVOS), so host Fabric Manager must stay off and cross-node NVLink is set up via IMEX. Everything is keyed off the VM size:

- datamodel: IMEXGPUSizes for the two GB SKUs; GB explicitly false in FabricManagerGPUSizes (starting fabricmanager on GB fails provisioning).
- GPUNeedsIMEX + template func; GPU_NEEDS_IMEX plumbed on BOTH provisioning paths -- classic baker (cse_cmd.sh) and aks-node-controller (parser).
- installNvidiaIMEX: install nvidia-imex from a baked cache, create the IMEX channel (NVreg_CreateImexChannel0=1, triggers reboot), leave the daemon DISABLED (per-domain peer list is owned at runtime by ComputeDomains); cache removed after install and on teardown.
- VHD build: cache nvidia-imex, and cache the aks-gpu-cuda-lts (open R580) driver image on arm64 too -- arm64 reaches parity with amd64 so GB runs on the vanilla arm64 image, no bespoke VHD.

Tests: unit guard (gpu_gb_test: GB -> FM off + IMEX on; H100 -> inverse) and an e2e scenario (GB300/GB200 on vanilla arm64) with GB-specific validators (FM not running, nvidia-imex installed-but-disabled, IMEX channel device, open R580 module, cache cleaned).

Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
@@ -223,7 +223,7 @@ removeNvidiaRepos() {
     fi
 }
 
-# installNvidiaIMEX installs + enables nvidia-imex for Grace-Blackwell NVL (GB200/GB300). Called from
+# installNvidiaIMEX installs nvidia-imex (daemon left DISABLED) for Grace-Blackwell NVL (GB200/GB300). Called from
 # cse_main.sh only for IMEXGPUSizes, AFTER the GPU driver is installed by aks-gpu (.run installer).
 # GB has no on-board NVSwitch (so no host Fabric Manager); IMEX is its host-side fabric piece for
 # cross-node NVLink. This step only (a) installs the nvidia-imex binary and (b) creates the IMEX channel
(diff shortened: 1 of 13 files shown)


parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -223,6 +223,54 @@ removeNvidiaRepos() {
...
+# cross-node NVLink. This step only (a) installs the nvidia-imex binary and (b) creates the IMEX channel
+# device (a driver/kernel thing pods cannot create). It does NOT run the daemon or write a peer list:
+# both are a runtime, topology-dependent concern owned by whatever brings up the NVLink domain --
+# ComputeDomains (nvidia-dra-driver-gpu, which runs its own imex daemons in pods) on Kubernetes, or a
+# manual `nodes_config.cfg` + `systemctl start` for hostNetwork/testing.
+# EXPERIMENT open question: with ComputeDomains running imex in pods, the host may only need the CHANNEL
+# -- confirm on GB HW whether the host nvidia-imex install is required or just the channel.
+# nvidia-imex is BAKED into the VHD (install-dependencies.sh caches it to /opt/nvidia-imex/downloads via
+# the components.json entry) -- CSE installs it from that cache, no runtime download.
+installNvidiaIMEX() {
```
awesomenix, line 238: might need to define this in other OS as well to avoid regression.

PR #8845, opened 2026-07-06
test(e2e): verify artifact streaming actually streams an image on pod launch

ganeshkumarashok:
## What

Adds `Test_Ubuntu2204_ArtifactStreaming_ImagePull`, an e2e that verifies artifact streaming **actually streams an image on pod launch**, not just that the streaming services come up.

## Why

Every existing `*_ArtifactStreaming*` scenario only asserts node bootstrap: that `overlaybd-snapshotter`, `overlaybd-tcmu`, and `acr-mirror` are running and `/etc/overlaybd` is non-empty. The per-scenario pod (`podHTTPServerLinux`) pulls a plain OCI image (`cbl-mariner/busybox:2.0`). With streaming enabled the CRI snapshotter is `overlaybd`, but overlaybd only *streams* overlaybd-format images — a plain OCI image falls back to overlayfs. So **the on-demand (TCMU) streaming path is never exercised** by the current tests.

## How

The new test:
1. Ensures an overlaybd **artifact-streaming referrer** exists for an image in the e2e private ACR — `az acr import` + `az acr artifact-streaming create`, made idempotent by checking the referrer end-state via `az acr manifest list-referrers` (so cached-ACR re-runs are a no-op and it does not depend on CLI "already exists" wording).
2. Launches a pod from that ACR image and waits for `Running` (a successful pull through the overlaybd snapshotter).
3. **While the pod is still running**, asserts overlaybd opened a TCMU-backed block device for it — a backstore device under `/sys/kernel/config/target/core/user_*/*/`. Plain OCI images (overlayfs fallback) create none, so a non-zero count is the definitive proof the image was streamed on demand.

Uses `ClusterAzureBootstrapProfileCache` (Premium private ACR attached, **not** network isolated) with `NonAnonymousACR` so the ACR pull secret is present.

### Notes / review focus
- **Node-side proof timing:** the TCMU check runs before pod teardown on purpose. The backstore only exists while the container rootfs is mounted, so the test does not use `ValidatePodRunning*` (which force-deletes the pod on return).
- **False-positive hardening:** counts backstore **devices** under `user_*/*/`, not the HBA directory (which `overlaybd-tcmu` may pre-create empty).
- **`az` dependency:** the setup shells out to `az` (as the suite already does in `types.go`). The runner must be subscription-authenticated with `az acr artifact-streaming`/`az acr manifest` available. There is no armcontainerregistry SDK surface for creating a streaming artifact today; swap for an SDK call once one exists.

## Test
- `go build ./...`, `go vet .` clean; test compiles and is registered.
- Full run requires Azure infra (Premium ACR + streaming-enabled node) — not run in this change.

Draft: seeking feedback on the streaming-artifact prep approach and the TCMU signal before a full pipeline run.

### DIFF
--- a/e2e/artifact_streaming.go
+++ b/e2e/artifact_streaming.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"regexp"
 	"strings"
 	"time"
 
@@ -22,6 +23,16 @@ const artifactStreamingE2ERepoTag = "artifact-streaming-e2e/base-core:2.0"
 // artifactStreamingSourceImage is a small, always-available public image we convert to an
 // overlaybd (artifact streaming) image so the pod pull exercises the streaming path.
 const artifactStreamingSourceImage = "mcr.microsoft.com/cbl-mariner/base/core:2.0"
+
+// overlaybdStreamingArtifactType is the artifactType of the ACR overlaybd streaming referrer that
+// `az acr artifact-streaming create` produces. Verified against a live ACR streaming artifact
+// (annotations include streaming.format=overlaybd). Filtering `list-referrers` by this type avoids
+// matching unrelated referrers such as signatures or SBOMs.
+const overlaybdStreamingArtifactType = "application/vnd.azure.artifact.streaming.v1"
+
+// streamingOperationIDRegex extracts the operation UUID from the async `az acr artifact-streaming
+// create` output, e.g. "... operation show ... --id 1a410a07-d2d5-4f3a-a386-a4a2e75c1e40".
+var streamingOperationIDRegex = regexp.MustCompile(`--id\s+([0-9a-fA-F-]{36})`)
 
 // ValidateArtifactStreamingImagePull verifies that artifact streaming actually streams an image
 // on pod launch, rather than merely bootstrapping the overlaybd/acr-mirror services.
@@ -95,9 +106,15 @@ func ValidateArtifactStreamingImagePull(ctx context.Context, s *Scenario) {
 }
 
 // ensureStreamingArtifactForImage imports the source image into the private ACR and ensures its
-// overlaybd artifact-streaming referrer exists. Success is defined by the end state (a streaming
-// referrer is present), not by the CLI exit code, so it is fully idempotent across cached-ACR
-// re-runs without depending on the exact wording of "already exists" messages.
+// overlaybd artifact-streaming referrer is fully created and ready to pull. It is idempotent across
+// cached-ACR re-runs.
+//
+// IMPORTANT: `az acr artifact-streaming create` is ASYNCHRONOUS — it returns immediately with an
+// operation ID while ACR converts the image to overlaybd format server-side. We MUST wait for that
+// operation to succeed before pulling: if the node pulls before conversion finishes, containerd
+// pulls the plain OCI image, caches it as overlayfs, and no streaming ever happens (and re-pulling
+// on the same node won't fix it, because the image is already cached). This async behaviour is the
+// reason the first version of this test failed.
 //
 // NOTE: this shells out to `az` (as the e2e suite already does in types.go). The E2E runner must be
 // authenticated to the subscription and have the `az acr artifact-streaming`/`az acr manifest`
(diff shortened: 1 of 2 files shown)


ganeshkumarashok:
## Update: fixed the real E2E failure (async ACR conversion race)

The first E2E run had my test fail correctly — it was a genuine bug in the test, not infra. Root cause from the node diagnostics:

```
az acr artifact-streaming create output:
WARNING: The operation is in progress. Use 'az acr artifact-streaming operation show ... --id ...' to check the status.
```

`az acr artifact-streaming create` is **asynchronous**. The test launched the pod immediately, before ACR finished converting the image to overlaybd server-side, so containerd pulled the plain OCI image and cached it as overlayfs (0 TCMU backstores; the overlaybd exporter confirmed `Read_Throughput=0`, `QPS=0`). Re-pulling on the same node can't recover — the non-streamed image is already cached.

Fix (commit 4b17a7a): `ensureStreamingArtifactForImage` now waits for readiness before any pull — it polls `az acr artifact-streaming operation show --id <op>` until `Succeeded`, then confirms the referrer is listable.

### Re: the two review comments (thanks @copilot)

**1. `list-referrers` argument order** — verified this is actually **correct** as written. Per `az acr manifest list-referrers --help`, `--name/-n` is the image (`repo:tag`) and `--registry/-r` is the registry. Confirmed live against the e2e ACR:
```
$ az acr manifest list-referrers -r abe2eprivatenonanonwestus3 -n artifact-streaming-e2e/base-core:2.0
{ "manifests": [ { "artifactType": "application/vnd.azure.artifact.streaming.v1",
                   "annotations": { "streaming.format": "overlaybd", ... } }, ... ] }
```
(The `default` sub-test also empirically found the referrer with this ordering.)

**2. `artifactType` substring too loose** — good catch, fixed. The readiness check now filters `list-referrers --artifact-type application/vnd.azure.artifact.streaming.v1` (the verified overlaybd streaming type above), so signatures/SBOMs can't be mistaken for the streaming referrer.

Also expanded node-side diagnostics (acr-mirror + overlaybd-snapshotter journals, containerd image list, azurecr.io `hosts.toml`) to triage any residual overlayfs fallback on the re-run.

Note: unrelated to this PR, ~half the E2E suite failed on a shared `wireserver` check (`unexpected curl exit code "0"`) in the same run.

ganeshkumarashok:
## Root cause of the 2nd run failure: acr-mirror can't auth to the non-anon ACR

The async fix worked (this run correctly skipped `create` because the referrer was already fully converted), but the pod still fell back to overlayfs. The new node diagnostics found why — in the **acr-mirror journal**:

```
ERROR lifec_registry: Error with azure sdk, request token error
ERROR ...login: Could not authn request, falling back to anonymous auth
ERROR ...authenticate: Could not authn w/ registry
WARN  ...proxy_route: Context did not generate a response, returning a 503
```

The standalone e2e VMSS node has **no managed identity**, so acr-mirror can't obtain an AAD token to authenticate to the **non-anonymous** ACR → returns 503 → `overlaybd-snapshotter` falls back to `writeType: overlayfs`. (The containerd `hosts.toml` + `x-ms-upgrade-if-streamable: overlaybd` header were correct — node config was never the problem.)

**Fix (commit 1fbaa64):** point the validator at the cluster's **anonymous-pull** ACR (also provisioned by `ClusterAzureBootstrapProfileCache`; verified Premium + `anonymousPullEnabled: true`). acr-mirror's anonymous path can then serve the streaming manifest without a node identity. Dropped the now-unneeded pod image pull secret.

Re-running E2E now.

ganeshkumarashok:
## ✅ E2E green — the test now streams for real

Agentbaker E2E passed ([build 171274258](https://msazure.visualstudio.com/09706533-03bf-4b43-9a9b-b49c75429646/_build/results?buildId=171274258), 48m):

```
✓ Ubuntu2204 artifact streaming image pull default (222.97s)
✓ Ubuntu2204 artifact streaming image pull scriptless nbc (218.65s)
DONE 475 tests, 62 skipped
```

Both variants provision a streaming-enabled node, prepare the overlaybd artifact in the anonymous ACR, launch a pod from it, and confirm a TCMU-backed block device is mounted on the node — i.e. the image was streamed on demand, not unpacked into overlayfs. The three iterations to get here each caught a real streaming-path bug (async ACR conversion race; acr-mirror auth → anonymous ACR), which is exactly what this test is meant to catch.

Ready for review.

SriHarsha001 (approved):

PR #8863, opened 2026-07-08
chore(deps): bump go to 1.26.4 across all modules

calvin197:
## What this PR does / why we need it

Bump the `go` directive in all modules (agentbaker service, aks-node-controller, e2e, hack/tools, image-fetcher, vhdbuilder/lister, vhdbuilder/prefetch) from `1.25.11` to `1.26.4`, and move the CI `actions/setup-go` pins to `'1.26'`. This picks up the go1.26.5 security fixes (crypto/tls, os) and aligns absvc with e2e/STLS, which are already on the 1.26 line.

### Why 1.26.4 and not 1.26.5?

The `go X.Y.Z` line in `go.mod` is a **minimum-version floor**, not the compiler that's actually used:

- **GitHub CI** installs Go via `actions/setup-go` and runs with `GOTOOLCHAIN=local`. The upstream setup-go manifest currently tops out at **1.26.4** (go1.26.5 was only released 2026-07-07 and isn't published there yet). A `go 1.26.5` directive is therefore unsatisfiable in CI and fails hard with `go.mod requires go >= 1.26.5 (running go 1.26.4; GOTOOLCHAIN=local)`.
- **The release/VHD build** compiles with the latest `msft-golang` toolchain (1.26.5), which comfortably satisfies a `1.26.4` floor — so the shipped binaries still carry the go1.26.5 crypto/tls and os security fixes regardless of the directive value.

Setting the floor to `1.26.4` keeps CI green today while still delivering the security fixes at build time. The workflow pins stay on the `'1.26'` major.minor line, so they'll auto-pick 1.26.5 (and a trivial `.4 → .5` directive bump can follow) once it lands in the setup-go manifest.

## Which issue(s) this PR fixes

Fixes #

## Special notes for your reviewer

`e2e/go.mod` is unchanged — it was already on `1.26.4`. Net diff is 13 files (6 `go.mod` + 7 workflow pins).

### DIFF
--- a/.github/workflows/check-coverage.yml
+++ b/.github/workflows/check-coverage.yml
@@ -14,7 +14,7 @@ jobs:
         if: success()
         uses: actions/setup-go@v6
         with:
-          go-version: '1.25'
+          go-version: '1.26'
       - name: Checkout code
         uses: actions/checkout@v7
       - name: Run unit tests
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -30,7 +30,7 @@ jobs:
 
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.25'
+          go-version: '1.26'
 
       - run: |
           set -ex
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -13,7 +13,7 @@ jobs:
         fetch-depth: 0
     - uses: actions/setup-go@v6
       with:
-        go-version: '1.25'
+        go-version: '1.26'
     - run: |
         set -ex
         make test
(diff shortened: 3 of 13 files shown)


cameronmeissner (approved):

awesomenix:
has RP moved to 1.26? you will have a hard time importing this into RP if thats not the case

awesomenix (changes requested):
just make sure RP has moved to 1.26

cameronmeissner:
> has RP moved to 1.26? you will have a hard time importing this into RP if thats not the case

yes good catch, RP hasn't been bumped yet - let's just bump the service go mod specifically and revisit this later

rhamitarora (approved):

PR #8868, opened 2026-07-08
feat: add aks-live-patching module with live patching service proto

YaoC:
add aks-live-patching module with live patching service proto

Fixes #


aks-live-patching/Makefile
```diff
@@ -0,0 +1,17 @@
+# Run buf in docker, mounting the full repo into the container
+# Emulate running "buf" in the current directory
+BUF_VERSION = 1.71.0
+BUF = docker run --rm --platform linux/amd64 --volume "$(CURDIR)/../:$(CURDIR)/../" --workdir $(CURDIR) bufbuild/buf:$(BUF_VERSION)
```
djsly, line 4: we dont have docker...

aks-live-patching/go.mod
```diff
@@ -0,0 +1,15 @@
+module github.com/Azure/agentbaker/aks-live-patching
+
+go 1.25.11
```
djsly, line 3: I think this will need msft-go...  have we consider moving to managed-dalec ?

aks-live-patching/Makefile
djsly: maye we do during build...

Devinwong: yes we do. aks-node-controller proto also uses docker

Devinwong:
One gap vs existing aks-node-controller pattern: this PR adds a new proto module but doesn’t wire it into .github/workflows/buf.yaml.

Please add aks-live-patching/proto/** and aks-live-patching/buf.yaml to workflow triggers, and add a Buf run for aks-live-patching (same pattern as aks-node-controller: read BUF_VERSION from module Makefile + input: aks-live-patching).

Otherwise proto lint/breaking checks won’t run for this new module.

Devinwong:
Another parity gap: aks-live-patching is a new Go submodule but current CI wiring only explicitly covers existing modules (e.g. aks-node-controller).

Please add aks-live-patching to repo Go CI coverage:

golangci-lint workflow matrix
Go unit test path (workflow and/or root Makefile target chain) so go build + go test run in this module too.
This keeps quality gates consistent with existing module patterns before AKS-RP starts consuming it.

aks-live-patching/go.mod
Devinwong: I believe we are already using msft-go because there is 
`sudo apt-get -y install msft-golang` in `hack/setup_golang.sh`?

YaoC:
Added the `buf skip breaking` label intentionally. This PR *introduces* the new `aks-live-patching` proto module, so it doesn't exist yet on the base branch (`main`). Buf's breaking-change check compares against the base commit, finds no `.proto` files there, and fails with `Module "path: "aks-live-patching"" had no .proto files`.

Build, lint, and format all pass — only the breaking step is affected, and there's genuinely no baseline to compare against for a brand-new module.

The label skips only the breaking check for this one PR. Once this merges, `main` will contain the protos, and breaking-change detection runs normally on every future PR automatically — no follow-up cleanup needed.

aks-live-patching/go.mod
YaoC: I think whether this is the case should depend on the one using this module?

YaoC:
@djsly @Devinwong thanks for reviewing my PR! addressed your comments, PTAL

aks-live-patching/go.mod
Devinwong: Yao, you are right.
- go 1.25.11 means the consumer needs a compatible Go 1.25.11+ toolchain.
- Whether that toolchain is generic Go or msft-go is decided by the environment building/importing the module.

aks-live-patching/Makefile
```diff
@@ -0,0 +1,16 @@
+# Run buf in docker, mounting the full repo into the container
+# Emulate running "buf" in the current directory
+BUF = docker run --platform linux/amd64 --volume "$(CURDIR)/../:$(CURDIR)/../" --workdir $(CURDIR) bufbuild/buf:1.47.2
+
```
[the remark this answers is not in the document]
Devinwong: this comment looks valid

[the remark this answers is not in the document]
YaoC: This has been done

aks-live-patching/pkg/gen/akslivepatching/v1/live_patching_grpc.pb.go
```diff
@@ -0,0 +1,127 @@
...
+
+// UnsafeLivePatchingServiceServer may be embedded to opt out of forward compatibility for this service.
+// Use of this interface is not recommended, as added methods to LivePatchingServiceServer will
+// result in compilation errors.
+type UnsafeLivePatchingServiceServer interface {
+	mustEmbedUnimplementedLivePatchingServiceServer()
+}
+
+func RegisterLivePatchingServiceServer(s grpc.ServiceRegistrar, srv LivePatchingServiceServer) {
+	// If the following call pancis, it indicates UnimplementedLivePatchingServiceServer was
```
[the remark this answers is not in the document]
YaoC: This is auto generated file, we'd better not change it.

aks-live-patching/Makefile
```diff
@@ -0,0 +1,17 @@
...
+	@($(BUF) format -w)
+	rm -rf pkg/gen/akslivepatching/v1
+	docker build --platform linux/amd64 -t protoc-docker - < protoc.Dockerfile
+	docker run --platform linux/amd64 --rm -v $(shell pwd):/$(shell pwd) --workdir=$(shell pwd) protoc-docker protoc --go_opt=module=github.com/Azure/agentbaker/aks-live-patching --go_out=./ --go-grpc_opt=module=github.com/Azure/agentbaker/aks-live-patching --go-grpc_out=./ --proto_path=proto $(shell find proto/akslivepatching/v1 -name '*.proto')
+	$(MAKE) proto-lint
+
+.PHONY: proto-lint
+proto-lint:
+	@($(BUF) lint)
+	@($(BUF) breaking --against '../.git#branch=main,subdir=aks-live-patching')
```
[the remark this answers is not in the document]
YaoC: added PR label `buf skip breaking` to ignore the failure

Devinwong (approved):
LGTM

PR #8874, opened 2026-07-10
fix: scriptless is always enabled on prod, match the same in e2e

awesomenix:
Fix our e2e to test things properly


e2e/scenario_cse_perf_test.go
```diff
@@ -199,8 +199,6 @@ func Test_Ubuntu2204_CSE_CachedPerformance(t *testing.T) {
 			SkipScriptlessNBC:        true,
 			EagerCSETimingExtraction: true,
 			BootstrapConfigMutator: func(_ *Cluster, nbc *datamodel.NodeBootstrappingConfiguration) {
-				// Disable scriptless CSE so traditional CSE scripts run and emit timing events
```
djsly, line 202: we will be getting timing event again now that scriptless is enabled ?

awesomenix: yes.

djsly (approved):

PR #8893, opened 2026-07-10
feat: produce vhd info and consume them in e2e so we dont get throttled

awesomenix:
We are constantly throttled trying to get VHD info using tags in our E2E for PR checkin

1. Publish VHD metadata from VHD builds
2. Consume this metadata to get VHD information to be used in tests.

This saves going to azure for VHD info.

### DIFF
--- a/.pipelines/templates/.builder-release-template.yaml
+++ b/.pipelines/templates/.builder-release-template.yaml
@@ -271,6 +271,46 @@ steps:
       RESOURCE_GROUP_NAME: $(AZURE_RESOURCE_GROUP_NAME)
       DRY_RUN: $(DRY_RUN)
 
+  - bash: |
+      set -euo pipefail
+
+      metadata_dir="$(Build.ArtifactStagingDirectory)/e2e-vhd-metadata"
+      metadata_file="${metadata_dir}/${VHD_ARTIFACT_NAME}.json"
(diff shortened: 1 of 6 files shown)


.pipelines/templates/.builder-release-template.yaml
```diff
@@ -271,6 +271,46 @@ steps:
...
+      CAPTURED_SIG_VERSION: $(CAPTURED_SIG_VERSION)
+      MANAGED_SIG_ID: $(MANAGED_SIG_ID)
+      PACKER_BUILD_LOCATION: $(PACKER_BUILD_LOCATION)
+      REPLICATIONS: $(REPLICATIONS)
+
+  - task: PublishPipelineArtifact@0
+    condition: succeeded()
+    displayName: Publish E2E VHD metadata
+    inputs:
+      artifactName: 'e2e-vhd-metadata-${{ parameters.artifactName }}'
```
djsly: if we rerun, will it faill ? I recall seeing error where we cannot reupload the same artifact

djsly (approved):
please review the rerun case, to ensure it doesnt fail

.pipelines/templates/.builder-release-template.yaml
awesomenix: done

lilypan26 (approved):

PR #8895, opened 2026-07-11
feat: support localdns hosts plugin refresh override

saewoni:

This change lets AgentBaker carry a LocalDNS hosts plugin refresh interval from node bootstrapping config down to provisioning, instead of relying on the hardcoded 15 minute timer cadence.

It adds the new LocalDNS field to the datamodel and aks-node-controller proto/parser paths, emits the value into both legacy CSE and scriptless env, and updates `cse_config.sh` to rewrite `aks-localdns-hosts-setup.timer` when a positive interval is provided. When the field is unset or invalid, provisioning keeps the existing default behavior.

This complements the corresponding AKS-RP change in PR 16401188.

Validation run:
- `make validate-shell`
- `docker run --rm -v /home/sakwa/AgentBaker:/workspace -w /workspace shellspec-docker --shell bash --format d spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh`
- `go test ./pkg/agent/...`
- `cd aks-node-controller && go test ./parser`
- `./hack/tools/bin/golangci-lint run ./pkg/agent/...`
- `cd aks-node-controller && ../hack/tools/bin/golangci-lint run ./parser/...`

Fixes #


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1804,8 +1806,33 @@ EOF
...
-    # This will update the hosts file with fresh IPs from live DNS
+    local hosts_plugin_refresh_interval="${LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS:-}"
+    if [ -n "${hosts_plugin_refresh_interval}" ]; then
+        case "${hosts_plugin_refresh_interval}" in
+            *[!0-9]*)
+                echo "Warning: LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS must be a positive integer, got '${hosts_plugin_refresh_interval}'. Using default timer interval."
+                ;;
+            *)
+                if [ "${hosts_plugin_refresh_interval}" -gt 0 ]; then
+                    sed -i "s/^OnUnitActiveSec=.*/OnUnitActiveSec=${hosts_plugin_refresh_interval}s/" "${hosts_setup_timer}"
```
awesomenix: Seems very hacky. I would think that there would be some form of `timer.template` and then you render it either on agentbaker side or aks-node-controller side?

yewmsft (commented):
nice, right direction — configurable over the hardcoded 15min, additive optional field, empty preserves default. one blocking thing (AccuracySec, else sub-minute intervals silently don't fire) + a few asks inline.

parts/linux/cloud-init/artifacts/cse_config.sh
yewmsft: this only rewrites `OnUnitActiveSec`, but the shipped `aks-localdns-hosts-setup.timer` also has `AccuracySec=1min`. systemd won't fire more accurately than the accuracy window, so `OnUnitActiveSec=5s` gets coalesced to ~1min — sub-minute cadence won't actually happen. RP (16401188) is passing seconds, so this silently won't deliver. need to drop AccuracySec too:

```sh
sed -i "s/^AccuracySec=.*/AccuracySec=1s/" "${hosts_setup_timer}"
```

did you verify on a real node that a small value actually fires at that cadence? i don't think it does as-is.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1804,8 +1806,33 @@ EOF
...
-    # Enable the timer for periodic refresh (every 15 minutes)
-    # This will update the hosts file with fresh IPs from live DNS
+    local hosts_plugin_refresh_interval="${LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS:-}"
+    if [ -n "${hosts_plugin_refresh_interval}" ]; then
+        case "${hosts_plugin_refresh_interval}" in
+            *[!0-9]*)
+                echo "Warning: LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS must be a positive integer, got '${hosts_plugin_refresh_interval}'. Using default timer interval."
+                ;;
+            *)
+                if [ "${hosts_plugin_refresh_interval}" -gt 0 ]; then
```
yewmsft: this is the only bound — RP validates `>0`, so `1` is legal and would hit live DNS every second on every localdns node. do we want a floor? at least call out a sane min.

aks-node-controller/parser/helper.go
```diff
@@ -889,6 +889,19 @@ func getLocalDnsCriticalFqdns(config *aksnodeconfigv1.Configuration) string {
 	return getStringifiedStringArray(trimmed, ",")
 }
 
+// getLocalDnsHostsPluginRefreshIntervalInSeconds returns the refresh interval in seconds
+// for the LocalDNS hosts plugin timer. Empty string means use the default timer cadence.
+func getLocalDnsHostsPluginRefreshIntervalInSeconds(config *aksnodeconfigv1.Configuration) string {
+	if config == nil || config.GetLocalDnsProfile() == nil {
```
yewmsft: `GetLocalDnsProfile()` and `GetHostsPluginRefreshIntervalInSeconds()` are nil-safe on nil receivers. drop this guard and just call the getters:

```go
refresh := config.GetLocalDnsProfile().GetHostsPluginRefreshIntervalInSeconds()
if refresh <= 0 {
    return ""
}
```

(baker.go is a plain datamodel struct, not proto — keep the nil checks there.)

spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh
```diff
@@ -1493,6 +1504,15 @@ SETUP_EOF
             The output should include "systemctlEnableAndStartNoBlock aks-localdns-hosts-setup.timer 30"
         End
 
+        It 'should update the timer refresh interval when provided'
```
yewmsft, line 1594: only the happy path is covered. the `*[!0-9]*` warning branch and the empty/`0` -> default-preserved path never run in test. add a bad-format case (e.g. `"abc"`) asserting the timer stays `15min`, and an unset case. otherwise would this short circuit your failure code path?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1650,6 +1650,7 @@ LOCALDNS_COREFILE_BASE=${corefile_base}
 LOCALDNS_COREFILE_WITH_HOSTS=${LOCALDNS_COREFILE_WITH_HOSTS:-}
 SHOULD_ENABLE_HOSTS_PLUGIN=${SHOULD_ENABLE_HOSTS_PLUGIN:-false}
 LOCALDNS_CRITICAL_FQDNS=${LOCALDNS_CRITICAL_FQDNS:-}
+LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS=${LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS:-}
```
yewmsft: who reads this? the interval only drives the timer rewrite at provision time — the periodic `aks-localdns-hosts-setup` service doesn't consume it. same for the write at L1799. if nobody reads them, drop both.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1804,8 +1806,33 @@ EOF
     touch "${hosts_file}"
     chmod 0644 "${hosts_file}"
 
-    # Enable the timer for periodic refresh (every 15 minutes)
-    # This will update the hosts file with fresh IPs from live DNS
+    local hosts_plugin_refresh_interval="${LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS:-}"
+    if [ -n "${hosts_plugin_refresh_interval}" ]; then
+        case "${hosts_plugin_refresh_interval}" in
+            *[!0-9]*)
```
yewmsft, line 1871: nit: this warning and the `-gt 0` else below print the same message — collapse. also the "Refresh every 15 minutes..." comment in `aks-localdns-hosts-setup.timer` goes stale once you sed the value; make it generic.

parts/linux/cloud-init/artifacts/cse_config.sh
saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

Instead of rewriting the shipped timer unit in place, the refresh override now uses a systemd drop-in:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1824-L1848

The base timer file stays generic/default-only here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/aks-localdns-hosts-setup.timer#L7-L13

saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

The override now uses a systemd drop-in and sets both `OnUnitActiveSec` and `AccuracySec=1s` here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1824-L1848

ShellSpec coverage for the generated override is here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh#L1507-L1515

I did not run a live-node cadence check in this branch; the validation here was ShellSpec/shellcheck rather than a real node.

parts/linux/cloud-init/artifacts/cse_config.sh
saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

The warning is collapsed to the single default-preserving branch here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1860-L1862

And the shipped timer comments are generic/default-only here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/aks-localdns-hosts-setup.timer#L7-L13

parts/linux/cloud-init/artifacts/cse_config.sh
saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

The interval is no longer written into either localdns environment file. The generated localdns env now stops at `LOCALDNS_CRITICAL_FQDNS` here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1647-L1653

And the minimal hosts-setup env write is only:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1808-L1815

The interval is only consumed at provisioning time for the timer drop-in here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1824-L1848

spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh
saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

Added coverage for the provided interval, unset/default-restore, invalid-format/default-preserved, and below-min/default-preserved cases here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh#L1507-L1545

parts/linux/cloud-init/artifacts/cse_config.sh
saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

Added a 5s floor here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/parts/linux/cloud-init/artifacts/cse_config.sh#L1824-L1862

aks-node-controller/parser/helper.go
saewoni: Addressed in commit https://github.com/Azure/AgentBaker/commit/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2.

Dropped the explicit nil guard and now rely on the proto getters nil-safe behavior here:
https://github.com/Azure/AgentBaker/blob/8366dcd0f7b1b776e6f4090c441eb573eb4cd2c2/aks-node-controller/parser/helper.go#L892-L899

parts/linux/cloud-init/artifacts/cse_config.sh
yewmsft: Following up on the floor: the current impl treats a *provided but below-min* numeric value (e.g. `3`) the same as invalid input and falls through to the 15min default via `removeAKSLocalDNSHostsSetupTimerOverride`. That's the wrong direction.

If RP explicitly passes a sub-floor value it's asking for a *fast* refresh — silently handing back the *slowest* cadence (15min) inverts intent. The 15min default belongs only to the unset/empty case (no override requested); it has nothing to do with a provided value that happens to be below the floor. For a provided numeric value under the min, clamp **up** to the 5s floor instead of dropping to default:

```sh
if [ "${hosts_plugin_refresh_interval}" -lt "${min_hosts_plugin_refresh_interval_in_seconds}" ]; then
    hosts_plugin_refresh_interval="${min_hosts_plugin_refresh_interval_in_seconds}"
fi
```

(Non-numeric garbage can still warn + fall through to default — that's a genuinely malformed input, a separate case.) Please also update the ShellSpec below-min test to assert the timer ends up at `5s`, not `15min`.

yewmsft:
### Empirical verification on a live localdns node ✅

Provisioned a 1-nodepool/1-node cluster with localdns enabled (sub `2b03bf…`), then on the node applied **exactly this PR's drop-in** (`/etc/systemd/system/aks-localdns-hosts-setup.timer.d/10-refresh-interval.conf` → `OnUnitActiveSec=5s` + `AccuracySec=1s`), populated `LOCALDNS_CRITICAL_FQDNS` (as RP would), `daemon-reload` + `systemctl enable --now aks-localdns-hosts-setup.timer`, and observed the timer live. RP's real `/etc/localdns/environment` was left intact.

**Cadence — confirmed ~5s**
- 8 firings in a 50s window (expected 8–10)
- Inter-firing intervals: `6 6 6 6 6 5 6` seconds

The steady ~6s (vs. 5s nominal) is expected and correct: `OnUnitActiveSec` measures from **run finish**, and each refresh does 5 sequential `dig` lookups (~1s total) → 5s idle + ~1s runtime ≈ 6s wall-clock.

**Refresh calls — all succeeding**
```
Received 5 critical FQDNs from RP
Using upstream DNS servers: 168.63.129.16
Dispatching resolution for <5 FQDNs>...
All FQDN resolutions completed
Found 27 valid IP address mappings
service: Deactivated successfully / Finished
```
`ExecMainStatus=0`, 27 A/AAAA records written across the 5 FQDNs (HCP apiserver, mcr, management.azure.com, login.microsoftonline.com, packages.microsoft.com), and `/etc/localdns/hosts` mtime advances every cycle.

This validates the full RP-driven path end-to-end: **interval → env var → `cse_config.sh` drop-in → 5s timer → successful dig refresh** — using the pre-PR base timer (15min/1min) on the VHD, overridden by this PR's drop-in.

parts/linux/cloud-init/artifacts/cse_config.sh
saewoni: Updated in commit 842c07356e78f0bbafd097bc7b32a40523cf4807: https://github.com/Azure/AgentBaker/commit/842c07356e78f0bbafd097bc7b32a40523cf4807

The below-min numeric path now clamps to `5s` and still writes the systemd override instead of dropping back to the `15min` default: https://github.com/Azure/AgentBaker/blob/842c07356e78f0bbafd097bc7b32a40523cf4807/parts/linux/cloud-init/artifacts/cse_config.sh#L1836-L1856

The invalid non-numeric path still warns and falls back to default here: https://github.com/Azure/AgentBaker/blob/842c07356e78f0bbafd097bc7b32a40523cf4807/parts/linux/cloud-init/artifacts/cse_config.sh#L1862-L1864

The ShellSpec below-min case now asserts the timer override ends up at `5s` with `AccuracySec=1s`: https://github.com/Azure/AgentBaker/blob/842c07356e78f0bbafd097bc7b32a40523cf4807/spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh#L1542-L1550

yewmsft (approved):
Approving. Code correctly implements RP-driven localdns hosts-plugin refresh override with a 5s floor (sub-floor values clamp to 5s), AccuracySec=1s, and clean unset/invalid handling. Verified end-to-end on a live localdns node: ~5s cadence and successful dig refreshes. Conflict resolved, functional CI green. Remaining unresolved threads are outdated/cosmetic.

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1842,8 +1859,52 @@ EOF
...
+        removeAKSLocalDNSHostsSetupTimerOverride "${hosts_setup_timer_override}"
+    else
+        local should_override_refresh_interval="false"
+        case "${hosts_plugin_refresh_interval}" in
+            *[!0-9]*)
+                ;;
+            *)
+                should_override_refresh_interval="true"
+                if [ "${hosts_plugin_refresh_interval}" -lt "${min_hosts_plugin_refresh_interval_in_seconds}" ]; then
+                    echo "Warning: LOCALDNS_HOSTS_PLUGIN_REFRESH_INTERVAL_IN_SECONDS must be >= ${min_hosts_plugin_refresh_interval_in_seconds}, got '${hosts_plugin_refresh_interval}'. Clamping to ${min_hosts_plugin_refresh_interval_in_seconds}s."
```
cameronmeissner, line 1876: is the goal that customers will be able to set this? if so why not validate this on the RP side?

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
yewmsft: No, not intended for customer to set. This is reserved for incident mitigation just in case the customer is being throttled by azure dns or hit vfp limit.

saewoni: is it not good practice to validate on the AgentBaker side? aka, all values that're passed to the AgentBaker should be assumed to have passed validation?

cameronmeissner: `aka, all values that're passed to the AgentBaker should be assumed to have passed validation?` - yes for values configurable by the customer

PR #8902, opened 2026-07-11
fix: make anc even faster by 7s

awesomenix:
- warm up aks node controller saves about 3s.
- warm up coredns and stls bootstrap binaries saving about 13s.
- disable pub key auth only when requested, so regular path remains fast 4s.
- disable ig service because we have already imported the gadgets we dont need to do it again at startup, its unnecessarily competing for resources.
- add cse_preload.sh to prefetch

```
 ig service only imports OCI artifacts.

 ig image import :

1. Opens the gadget tar bundle.
2. Copies its blobs using ORAS.
3. Stores them under  /var/lib/ig/oci-store .
4. Updates the OCI index and tracking file.

It does not run gadgets, load eBPF programs, call  modprobe , or load kernel modules. eBPF programs are loaded later by commands such as  ig run .

Disabling the boot service is safe provided  ig image list  works directly from the newly booted VHD. It also avoids  ExecStop  removing the baked images during shutdown.
```

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1469,17 +1469,8 @@ disableSSH() {
     systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
 }
 
-configureSSHPubkeyAuth() {
-  local disable_pubkey_auth="$1"
-  local ssh_use_pubkey_auth
-
-  # Determine the desired pubkey auth setting
-  if [ "${disable_pubkey_auth}" = "true" ]; then
-    ssh_use_pubkey_auth="no"
-  else
-    ssh_use_pubkey_auth="yes"
-  fi
-  local SSHD_CONFIG="/etc/ssh/sshd_config"
+disableSSHPubkeyAuth() {
+  local SSHD_CONFIG="${SSHD_CONFIG_FILE:-/etc/ssh/sshd_config}"
   local TMP
   TMP="$(mktemp)"
 
@@ -1490,7 +1481,7 @@ configureSSHPubkeyAuth() {
   # PubkeyAuthentication yes
   # AuthorizedKeysCommand /usr/sbin/aad_certhandler %u %k
   # AuthorizedKeysCommandUser root
-  awk -v desired="$ssh_use_pubkey_auth" '
+  awk -v desired="no" '
     BEGIN { in_match=0; replaced=0; inserted=0 }
     /^Match([[:space:]]|$)/ {
       if (!replaced && !inserted) { print "PubkeyAuthentication " desired; inserted=1 }
@@ -1503,6 +1494,12 @@ configureSSHPubkeyAuth() {
     END { if (!replaced && !inserted) print "PubkeyAuthentication " desired }
   ' "$SSHD_CONFIG" > "$TMP"
 
+  local ssh_service="sshd.service"
+  if systemctl cat ssh.service >/dev/null 2>&1; then
+    ssh_service="ssh.service"
+  fi
+  systemctl is-active --quiet "$ssh_service" || systemctl start "$ssh_service" || exit $ERR_CONFIG_PUBKEY_AUTH_SSH
+
   # Validate the candidate config
   sshd -t -f "$TMP" || { rm -f "$TMP"; exit $ERR_CONFIG_PUBKEY_AUTH_SSH; }
 
@@ -1510,8 +1507,8 @@ configureSSHPubkeyAuth() {
   install -m 644 -o root -g root "$TMP" "$SSHD_CONFIG"
   rm -f "$TMP"
 
-  # Reload sshd
-  systemctl reload sshd || systemctl restart sshd || exit $ERR_CONFIG_PUBKEY_AUTH_SSH
+  # Reload or restart ssh service
+  systemctl reload-or-restart "$ssh_service" || exit $ERR_CONFIG_PUBKEY_AUTH_SSH
 }
 
 # Internal function that writes credential provider config to a specified path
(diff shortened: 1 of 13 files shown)


parts/linux/cloud-init/artifacts/cse_preload.sh
```diff
@@ -4,10 +4,12 @@
...
-
+/opt/azure/containers/aks-node-controller version >/dev/null 2>&1 &
 /usr/bin/containerd --version >/dev/null 2>&1 &
 cat /var/lib/containerd/io.containerd.metadata.v1.bolt/meta.db >/dev/null 2>&1 &
 find /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots -maxdepth 1 >/dev/null 2>&1 &
 /sbin/modprobe overlay >/dev/null 2>&1 &
+cat /opt/bin/aks-secure-tls-bootstrap-client >/dev/null 2>&1 &
+/opt/azure/containers/localdns/binary/coredns --version >/dev/null 2>&1 &
 
 wait
```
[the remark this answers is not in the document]
awesomenix: should be ok since we call this script in background

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -2357,12 +2357,6 @@ testInspektorGadgetAssets() {
     err $test "Unit file missing at $unit_file"
   fi
 
-  local service_state
-  service_state=$(systemctl is-enabled "$service_name" 2>/dev/null || true)
-  if [ "$service_state" != "enabled" ]; then
-    err $test "$service_name not enabled, state: ${service_state:-absent}"
-  fi
-
   # Verify gadgets were imported during VHD build (tracking file should exist and have content)
```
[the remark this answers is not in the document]
awesomenix: dont need it

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1466,6 +1457,8 @@ configureSSHPubkeyAuth() {
     END { if (!replaced && !inserted) print "PubkeyAuthentication " desired }
   ' "$SSHD_CONFIG" > "$TMP"
 
+  systemctl is-active --quiet "sshd.service" || systemctl start "sshd.service" || exit $ERR_CONFIG_PUBKEY_AUTH_SSH
+
```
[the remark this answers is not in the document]
awesomenix: its fine

parts/linux/cloud-init/artifacts/cse_preload.sh
```diff
@@ -4,10 +4,16 @@
 # boot so that node provisioning runs against an already-warm page cache.
 # This is best-effort: every command is backgrounded and its output and exit
 # status are intentionally ignored. It must never block or fail provisioning.
+preload() {
```
cameronmeissner, line 7: I wasn't even aware of this file - this is essentially what workload prefetch optimization was built for :) have you considered moving this into build-time prefetch optimization? it should be feasible since it seems we're always pre-warming a static set of file paths, regardless of live node config

cameronmeissner: I'd be really curious to run some tests, using what you currently vs. calling this script at build time as a part of workload prefetch optimization

awesomenix: Yep thats the plan for next week :)

parts/linux/cloud-init/artifacts/cse_preload.sh
```diff
@@ -4,10 +4,16 @@
...
+    "$@" >/dev/null 2>&1 &
+}
 
-/usr/bin/containerd --version >/dev/null 2>&1 &
-cat /var/lib/containerd/io.containerd.metadata.v1.bolt/meta.db >/dev/null 2>&1 &
-find /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots -maxdepth 1 >/dev/null 2>&1 &
-/sbin/modprobe overlay >/dev/null 2>&1 &
+preload /opt/azure/containers/aks-node-controller version
```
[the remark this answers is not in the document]
awesomenix: there is 

Name:  "version",
				Usage: "Print the version",

parts/linux/cloud-init/artifacts/cse_preload.sh
awesomenix: nevermind it was straight forward, added it.

abigailliang-aks-sig-node (approved):

vhdbuilder/packer/install-ig.sh
```diff
@@ -220,8 +241,8 @@ installIG() {
         fi
     fi
 
-    # Enable the systemd service (baseline files copied by packer_source.sh)
-    ig_enable_service_unit || echo "[ig] Failed to enable ${IG_SERVICE_NAME}"
+    # disable the systemd service (baseline files copied by packer_source.sh)
+    ig_disable_service_unit || echo "[ig] Failed to disable ${IG_SERVICE_NAME}"
     ig_import_gadgets || echo "[ig] Gadget import failed during build"
```
Devinwong, line 246: If ig_import_gadgets fails, we should fail the VHD build instead of only logging, because with the service now disabled there’s no boot-time fallback import. I’d change it to a hard failure (e.g. ig_import_gadgets || return 1) or gate creation of skip_vhd_ig on successful import only.

awesomenix: i could argue that nobody is monitoring the service on VHD itself. But i need to check with chase what is the expectation now that the service is disabled with NPD fail?

awesomenix: ok we already tests `ig image list` in our e2e we should be fine

djsly:
@awesomenix how can we ensure that we don't have regression on the timing improvements ? can we augment the cse perf suites to ensure we don't regress ?

awesomenix:
>how can we ensure that we don't have regression on the timing improvements ? can we augment the cse perf suites to ensure we don't regress ?

@djsly Good question, the idea was to leverage telescope pipeline run daily, takes an average of 10 runs and alerts us if there is a regression. Reason why i am hesitant with E2E is it runs a single time and we cannot set tighter limits because we dont want flakiness.

Another approach could be make it part of regular E2E tests we have and run a check at the end so average, take 99 percentile which should be same as telescope pipeline

Devinwong (approved):

PR #8910, opened 2026-07-13
feat(compliance): integrate compliance-engine-assessor scanning

robertwoj-microsoft:
Initial integration of the Compliance Engine Assessor which will replace the current scripted CIS Assessor scanning.

PR descript will be extended later after the initial wiring is set up.


robertwoj-microsoft:
Follow-up: https://github.com/Azure/AgentBaker/pull/9587

PR #8911, opened 2026-07-13
fix: disable aks node controller and only start at boot and wait for network online target

awesomenix:
dont wait for network-online.target start in parallel, this saves about 4s

### DIFF
--- a/parts/linux/cloud-init/artifacts/aks-node-controller.service
+++ b/parts/linux/cloud-init/artifacts/aks-node-controller.service
@@ -1,7 +1,6 @@
 [Unit]
 Description=Parse contract and run csecmd
-After=network-online.target
-Wants=network-online.target
+DefaultDependencies=no
 
 [Service]
 Type=oneshot
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -175,8 +175,9 @@ func Test_ACL_Scriptless(t *testing.T) {
 	RunScenario(t, &Scenario{
 		Description: "Tests that a node using ACL and the self-contained installer can be properly bootstrapped",
 		Config: Config{
-			Cluster: ClusterKubenet,
-			VHD:     config.VHDACLGen2TL,
+			Cluster:                ClusterKubenet,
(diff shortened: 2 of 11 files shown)


parts/linux/cloud-init/artifacts/aks-node-controller.service
```diff
@@ -1,12 +1,8 @@
...
-Wants=network-online.target
+DefaultDependencies=no
 
 [Service]
 Type=oneshot
 ExecStart=/opt/azure/containers/aks-node-controller-wrapper.sh
 RemainAfterExit=yes
-
-[Install]
-WantedBy=basic.target
```
Devinwong, line 12: you probably were aware of this too

🔴 High Risk — 🏗️ Architecture / Backward Compatibility
parts/linux/cloud-init/artifacts/aks-node-controller.service removes [Install] and WantedBy, and vhdbuilder/packer/pre-install-dependencies.sh now disables the unit. That drops the long-standing boot fallback path (service auto-start) and makes provisioning depend entirely on boothook start timing.
This aligns with the failing PR E2E signal: multiple scenarios are stuck in provision-wait loop ([ -f /opt/azure/containers/provision.complete ]) and never reach expected CSE output (ADO build 172002687, Run AgentBaker E2E).
Mitigation: restore a fallback start path (or equivalent robust trigger) so ANC still starts when boothook path misses/races.

awesomenix: The problem is that we drop in the cse-cmd.sh or the aks node config only as part of boothook.

So we want to wait for it to be written before we can startup aks node controller otherwise we will have an issue where the service starts up before that and misses the file completely.

In future when we exclusively move to aks node controller we can add an explicity wait for aks node config before service starts that way everything is synchronized.

Devinwong (approved):
LGTM. Just the PR title seems conflicting with description.

awesomenix (commented):
will hold off on this change, feels a bit high risk

awesomenix:
this PR is actually useless because netowrk is always online before boothook

PR #8914, opened 2026-07-13
fix(ubuntu): don't treat benign ESM apt-hook stderr as apt-get failure (CSE exit 99)

xuexu6666:
## Problem

Node bootstrap can fail with **CSE `exit 99`** (`ERR_APT_UPDATE_TIMEOUT` / VMSS `AptUpdateTimeoutVMExtensionError`) **even though `apt-get update` actually succeeded**, so the node never joins the cluster.

`_apt_get_update()` / `apt_get_dist_upgrade()` in `parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh` decide success/failure by grepping the **merged stdout+stderr** (`2>&1`) for `^W:`, `^E:`, or the over-broad `^Error.*`:

```bash
! (apt-get ${apt_opts} update 2>&1 | tee $out | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && ...
```

The Ubuntu Pro / ESM apt hook (`apt-news` / `esm-cache`) logs this to stderr when it can't reach its snapd/dbus socket during **early boot**:

```
Error connecting: Error sending credentials: Error sending message: Broken pipe
```

That line starts with `Error`, so the grep flags the whole run as failed, retries 10×, `return 1` → caller `apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT` → **exit 99**. apt itself succeeded (`Hit:1 …/ubuntu/24.04/prod noble InRelease`, exit 0). It's intermittent because it depends on the early-boot dbus timing race.

Observed live on a real GB300 node (k8s 1.36.1, `2404gen2arm64containerd/202606.19.0`): `aks-node-controller` `exitCode=99`, `apt-get …/microsoft-prod.list update` shows `Hit:1 … noble InRelease` with the benign broken-pipe line as the *only* "error", and `KubeletStartTime` empty.

## Fix

Strip only the known-benign ESM/D-Bus line **before** the existing error grep, at both apt call sites. Intentionally minimal — the existing error regex is unchanged, so real apt failures (`E:` / `W:` / `Err:` prefixed) are still detected. Avoids regressing error detection on VHDs that live in production for 6 months.

- New file-scope `APT_BENIGN_STDERR_REGEX` (documented; not `readonly` so re-sourcing stays safe).
- `| grep -vE "${APT_BENIGN_STDERR_REGEX}"` inserted ahead of the error grep in `_apt_get_update` and `apt_get_dist_upgrade`.

## Tests

Adds a `_apt_get_update error detection` ShellSpec block:
- benign ESM line only → **succeeds** (regression test for this bug)
- clean update → succeeds
- real `E:`/`Err:` output → still **fails** after retries

`shellspec spec/parts/linux/cloud-init/artifacts/cse_helpers_ubuntu_spec.sh` → **12 examples, 0 failures**. TDD: the benign-line test failed before the fix, passes after.

## Notes

- `make generate` produces no diff (scripts are `go:embed`'d; no committed testdata copy of this script).

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
@@ -2,6 +2,16 @@
 
 echo "Sourcing cse_helpers_distro.sh for Ubuntu"
 
+# The Ubuntu Pro / ESM apt hooks (apt-news, esm-cache) log connection errors to stderr when they
+# cannot reach their snapd/dbus socket during early boot, e.g.:
+#   Error connecting: Error sending credentials: Error sending message: Broken pipe
+# These are cosmetic — the apt operation itself still succeeds (the repo line shows "Hit: ... InRelease").
+# The generic error grep used by the apt wrappers below matches any line starting with "Error", so it
+# would otherwise treat the whole apt-get run as failed and CSE would exit 99 (the node never joins).
+# Strip these known-benign lines before error detection; real apt failures (E:/W:/Err: prefixed) are
+# still caught. Not marked readonly so the file stays safe to re-source.
+APT_BENIGN_STDERR_REGEX='^Error (connecting|sending (credentials|message)):.*Broken pipe'
+
 
 holdWALinuxAgent() {
     echo $(date),$(hostname), startHoldWALinuxAgent "$1"
@@ -36,7 +46,7 @@ _apt_get_update() {
         export DEBIAN_FRONTEND=noninteractive
         dpkg --configure -a --force-confdef
         apt-get ${apt_opts} -f -y install
-        ! (apt-get ${apt_opts} update 2>&1 | tee $apt_update_output | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
+        ! (apt-get ${apt_opts} update 2>&1 | tee $apt_update_output | grep -vE "${APT_BENIGN_STDERR_REGEX}" | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
         cat $apt_update_output && break || \
         cat $apt_update_output
         if [ $i -eq $retries ]; then
@@ -174,7 +184,7 @@ apt_get_dist_upgrade() {
     # inline (not via /etc/apt/apt.conf.d) so it does not leak into the captured VHD or change node
     # runtime apt behavior. apt_get_dist_upgrade is only invoked during VHD build.
     # https://ubuntu.com/server/docs/explanation/software/about-apt-upgrade-and-phased-updates/#how-do-i-turn-off-phased-updates
-    ! (apt-get -o Dpkg::Options::="--force-confnew" -o APT::Get::Always-Include-Phased-Updates=true dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
+    ! (apt-get -o Dpkg::Options::="--force-confnew" -o APT::Get::Always-Include-Phased-Updates=true dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -vE "${APT_BENIGN_STDERR_REGEX}" | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
     cat $apt_dist_upgrade_output && break || \
     cat $apt_dist_upgrade_output
     if [ $i -eq $retries ]; then
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
@@ -2,14 +2,19 @@
 
 echo "Sourcing cse_helpers_distro.sh for Ubuntu"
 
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -36,7 +46,7 @@ _apt_get_update() {
         export DEBIAN_FRONTEND=noninteractive
         dpkg --configure -a --force-confdef
         apt-get ${apt_opts} -f -y install
-        ! (apt-get ${apt_opts} update 2>&1 | tee $apt_update_output | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
+        ! (apt-get ${apt_opts} update 2>&1 | tee $apt_update_output | grep -vE "${APT_BENIGN_STDERR_REGEX}" | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
```
cameronmeissner, line 54: wouldn't this be dangerous in the sense that we'd be silently ignoring potentially legitimate issues with being able to reach PMC/Canonical repos remotely?

xuexu6666: The line we strip is a **local** D-Bus/IPC error, not a remote one.

`Error sending credentials … Broken pipe` is the ubuntu-pro-client hook (`20apt-esm-hook.conf`) doing `systemctl start apt-news.service esm-cache.service` and failing the local systemd handshake in early boot — nothing to do with HTTP/DNS to PMC. A real PMC/Canonical outage shows up as apt's own `Err:` / `E: Failed to fetch` / `Could not connect`, which don't match the filter and still fail. Added a test (`still FAILS when PMC/Canonical is genuinely unreachable`) to lock that in.

Verified on live **GB300** hardware (same vanilla image `202606.19.0`): on non-baked `1.36.1` the CSE hit exit 99 while `apt-get update` had actually succeeded (`Hit:1 … noble InRelease`, only the broken-pipe line present); on baked `1.35.5` the nodes joined fine. Also confirmed the hook fires unattached (`is_attached=False`, `ubuntu-pro-client` installed).

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -174,7 +184,7 @@ apt_get_dist_upgrade() {
     # inline (not via /etc/apt/apt.conf.d) so it does not leak into the captured VHD or change node
     # runtime apt behavior. apt_get_dist_upgrade is only invoked during VHD build.
     # https://ubuntu.com/server/docs/explanation/software/about-apt-upgrade-and-phased-updates/#how-do-i-turn-off-phased-updates
-    ! (apt-get -o Dpkg::Options::="--force-confnew" -o APT::Get::Always-Include-Phased-Updates=true dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
+    ! (apt-get -o Dpkg::Options::="--force-confnew" -o APT::Get::Always-Include-Phased-Updates=true dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -vE "${APT_BENIGN_STDERR_REGEX}" | grep -E "^([WE]:.*)|^([Ee][Rr][Rr][Oo][Rr].*)$") && \
```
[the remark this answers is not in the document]
xuexu6666: Addressed in 4e12b9a544 — added a `Describe 'apt_get_dist_upgrade error detection (VHD build)'` block with two tests: the benign ESM/D-Bus line is ignored on dist-upgrade, and a real apt error still fails the run.

spec/parts/linux/cloud-init/artifacts/cse_helpers_ubuntu_spec.sh
```diff
@@ -143,4 +143,62 @@ Describe 'apt_get_install budget timeout'
...
+        It "succeeds when apt-get update prints only the benign Ubuntu Pro/ESM hook line"
+            # The apt-news / esm-cache apt hook logs this to stderr during early boot when it cannot
+            # reach its snapd/dbus socket. The apt operation itself still succeeds (Hit: ... InRelease),
+            # so it must NOT be treated as a failure. Regression test for the exit-99 false positive:
+            # apt-get update succeeded yet CSE exited 99 because the error grep matched "Error ...".
+            apt-get() {
+                case "$*" in
+                    *update*)
+                        echo "Hit:1 https://packages.microsoft.com/ubuntu/24.04/prod noble InRelease"
+                        echo "Error connecting: Error sending credentials: Error sending message: Broken pipe"
```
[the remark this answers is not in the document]
xuexu6666: Addressed in 4e12b9a544 — the ESM hook line is now emitted to stderr (`>&2`) in the stubs, matching the real hook, which the implementation merges via `2>&1`.

cameronmeissner (approved):

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
djsly: I thought I removed the ESM service in a previous PR

cameronmeissner (changes requested):
after offline discussion we don't want to move forward with this without further investigation into what's actually causing apt to produce these errors to stderr

PR #8915, opened 2026-07-13
fix: clean up mismatched prebaked GPU drivers

djsly:

Fixes Ubuntu GRID GPU nodes that can otherwise keep a CUDA driver prebaked into the VHD when CSE dynamically installs a GRID driver. The CSE now reads the prebake marker before managed GPU driver installation, normalizes the prebaked and requested driver families, and invokes the existing prebaked driver cleanup path when they do not match.

Legacy markers without `driver_kind` are treated as CUDA prebake only for GRID requests. This is intentionally driver-family aware only; CUDA-vs-CUDA version drift still needs separate marker/version comparison work.

Validation:
- Focused ShellSpec coverage for absent markers, CUDA-to-GRID cleanup, legacy GRID cleanup, and matching CUDA no-op passed with bash.
- `GENERATE_TEST_DATA=true go test ./pkg/agent...` passed.
- Full `make generate` is blocked by internal Go proxy authorization while installing `ginkgo`.

N/A

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1295,23 +1295,35 @@ validateGPUDrivers() {
     fi
 }
 
-# logGPUDriverPrebakeReadiness emits a stage-1 observability signal on a managed GPU node: whether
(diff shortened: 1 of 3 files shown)


djsly:
fixed in https://github.com/Azure/AgentBaker/pull/8919

PR #8918, opened 2026-07-13
fix(gpu): tear down cuda-lts VHD prebake on GRID nodes before managed install

ganeshkumarashok:
## Summary
Fixes `nvidia-smi` failing with **"Failed to initialize NVML: Driver/library version mismatch"** on agentpool **GRID / converged A10 (NVv5)** nodes and **NAP GRID** nodes booting the CUDA-prebaked shared Ubuntu VHD.

**Root cause:** the shared Ubuntu VHD prebakes only the `cuda(-lts)` driver + a DKMS marker (#8786/#8803). The existing `cleanUpPrebakedGPUDriver` only runs on nodes that do **not** install a managed driver (`GPU_NODE != true` OR skip). A GRID node installs the `grid` driver on top of the stale prebaked `cuda` module + `/usr/bin/lib64` libs, and the two collide → NVML mismatch.

## Fix
Add `cleanUpGridNodeCudaPrebake`, invoked from `ensureGPUDrivers` **before** the `configGPUDrivers` / `validateGPUDrivers` dispatch (so both install paths are covered). It tears down the prebake when:
- this node's driver type maps to **grid** (`grid` / `grid-v20`), **and**
- the prebake marker is **not** grid — i.e. a `cuda` marker, or a legacy marker with no `driver_kind=` line (the only kind the VHD bakes today).

This is a pure driver-**kind** mismatch, so no version comparison is needed. Reuses the existing `cleanUpPrebakedGPUDriver` for removal (which also clears `/usr/bin/lib64` and the DKMS registration).

## Scope
GRID mismatch only. **CUDA nodes are intentionally untouched** — the NAP cuda same-kind / different-version case (`580.126.09` vs baked `580.159.04`) is a separate issue, addressed by aligning the driver version NAP requests with the baked image rather than by teardown.

## Why this is safe on `official/v20260702`
Consume/skip-build is not in this branch, so a kept prebake buys no provisioning benefit today; the only effect of the guard is removing a driver that would otherwise break a GRID node. On a matching (grid) or absent marker, or on a CUDA node, it is a no-op.

## Test plan
- [x] shellcheck clean on the new function
- [x] Isolated logic harness: teardown on grid+cuda-marker, grid-v20+cuda, grid+legacy-marker; no-op on cuda-lts node, grid+grid-marker, no marker, non-Ubuntu
- [x] Added ShellSpec coverage in `cse_config_spec.sh` (`cleanUpGridNodeCudaPrebake`)
- [ ] e2e on an agentpool GRID/A10 node + a NAP GRID node booting the prebaked VHD

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sulixu (approved):

ganeshkumarashok:
Superseded by #8921, which is an exact cherry-pick of the squashed `main` merge (#8919) onto `official/v20260702` — byte-identical GRID-fix content. Closing in favor of that backport.

PR #8925, opened 2026-07-13
fix(gpu): tear down cuda-lts VHD prebake on GRID nodes before managed…

awesomenix:
… install (#8919)
testing cherry-pick

DO NOT MERGE


awesomenix:
hotfixing works

PR #8930, opened 2026-07-14
Remove 'dev' from ANC version content-test regex

Devinwong:
## Summary
- update `testAKSNodeControllerVersion` in `vhdbuilder/packer/test/linux-vhd-content-test.sh` to require only `YYYYMM.DD.PATCH`
- remove the `dev` branch from the regex
- update the failure message to match the stricter validation

## Why this is safe
This content test validates VHDs produced by `packer.mk`, and those builds always stamp a real datetime ANC version via:
- `ANC_VERSION="${IMAGE_VERSION:-$(date +%Y%m.%d.0)}"`
- `-ldflags "-X main.Version=${ANC_VERSION}"`

So a `dev` version on a production VHD indicates ldflags version stamping failed, which should be rejected by the content test. `dev` remains valid for unit tests/local e2e flows that do not use ldflags, and those flows do not run through this VHD content test.

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -1935,8 +1935,8 @@ testAKSNodeControllerVersion() {
     return 1
   fi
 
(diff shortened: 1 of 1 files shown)


Devinwong:
Closing as duplicate.

PR #8952, opened 2026-07-15
feat: update snapshot to be generic

chmill-zz:

updating ubuntu-snapshot-updater.sh to be generic and allow multiple component update through it in the future

splitting the original security updates into their own file to be called by ubuntu-snapshot-updater. So we can keep component specific needs separate. 

why not a new service? dropping the old one would remove the ability to hotfix those on older vhd using legacy service. aks will support both update methods. same concerns with a new file called by the existing service. 

if we hotfix an old vhd they'll get the new generic .sh along with the security updater. so this seems the safest route to old/new 

Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
@@ -15,13 +15,20 @@ set -e
 : "${KNEAD_COMPONENT_GOAL_ANNOTATION:=kubernetes.azure.com/live-patching-config-goal-hash}"
 : "${KNEAD_COMPONENT_STATUS_ANNOTATION:=kubernetes.azure.com/live-patching-status}"
 : "${KNEAD_COMPONENT_STATE_FILE:=/var/lib/aks/live-patching/current.json}"
+: "${KNEAD_KUBECONFIG_WAIT_TIMEOUT_SECONDS:=600}"
 
 KNEAD_COMPONENT_RESULTS='{}'
 KNEAD_COMPONENT_RESULTS_VALID=true
 
 # Waits for kubelet credentials so kubectl can read Node and ConfigMap state.
 knead_wait_for_kubeconfig() {
+    local wait_started_at="${SECONDS}"
+
     while [ ! -f "${KUBECONFIG}" ]; do
+        if [ $((SECONDS - wait_started_at)) -ge "${KNEAD_KUBECONFIG_WAIT_TIMEOUT_SECONDS}" ]; then
+            echo "timed out waiting for kubelet kubeconfig after ${KNEAD_KUBECONFIG_WAIT_TIMEOUT_SECONDS}s" >&2
+            return 1
+        fi
         echo "waiting for kubelet kubeconfig"
         sleep 3
     done
@@ -297,7 +304,9 @@ knead_main() {
     local payload
     local result=0
 
-    knead_wait_for_kubeconfig
+    if ! knead_wait_for_kubeconfig; then
+        return 1
+    fi
     if ! node_json="$(knead_read_node)"; then
         echo "failed to read node"
         return 1
--- a/parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
@@ -16,9 +16,47 @@ set -e
 : "${KNEAD_COMPONENT_STATUS_ANNOTATION:=kubernetes.azure.com/live-patching-status}"
 : "${KNEAD_COMPONENT_STATE_FILE:=/var/lib/aks/live-patching/current.json}"
 : "${KNEAD_KUBECONFIG_WAIT_TIMEOUT_SECONDS:=600}"
+: "${KNEAD_EVENTS_LOGGING_DIR:=/var/log/azure/Microsoft.Azure.Extensions.CustomScript/events}"
 
 KNEAD_COMPONENT_RESULTS='{}'
 KNEAD_COMPONENT_RESULTS_VALID=true
+
+# Writes a Guest Agent event using the established logs_to_events schema. This
+# remains available to component handlers sourced into this shell process.
+knead_emit_event() {
+    local task="$1"
+    local message="$2"
+    local level="${3:-Informational}"
+    local events_file_name
+    events_file_name="$(date +%s%3N)"
+    local timestamp
+    timestamp="$(date +"%F %T.%3N")"
+    local event_json
+    event_json="$(jq -n \
+        --arg Timestamp   "${timestamp}" \
+        --arg OperationId "${timestamp}" \
+        --arg Version "1.23" \
+        --arg TaskName    "${task}" \
+        --arg EventLevel  "${level}" \
+        --arg Message     "${message}" \
+        --arg EventPid    "0" \
+        --arg EventTid    "0" \
+        '{Timestamp: $Timestamp, OperationId: $OperationId, Version: $Version, TaskName: $TaskName, EventLevel: $EventLevel, Message: $Message, EventPid: $EventPid, EventTid: $EventTid}')"
+
+    mkdir -p "${KNEAD_EVENTS_LOGGING_DIR}"
+    printf '%s\n' "${event_json}" > "${KNEAD_EVENTS_LOGGING_DIR%/}/${events_file_name}.json"
+}
+
+# Component scripts should follow this pattern with a uniquely named
+# knead_emit_<component>_event wrapper that delegates transport to knead_emit_event.
+# This wrapper emits events for the generic live-patching reconciliation loop.
+knead_emit_reconcile_event() {
+    local outcome="$1"
+    local reason="$2"
+    local level="${3:-Informational}"
+
+    knead_emit_event "AKS.LivePatching.reconcile" "${outcome}: ${reason}" "${level}" || true
+}
 
 # Waits for kubelet credentials so kubectl can read Node and ConfigMap state.
 knead_wait_for_kubeconfig() {
@@ -305,18 +343,22 @@ knead_main() {
     local result=0
 
     if ! knead_wait_for_kubeconfig; then
+        knead_emit_reconcile_event "Failed" "kubeconfig wait failed" "Error"
         return 1
     fi
     if ! node_json="$(knead_read_node)"; then
         echo "failed to read node"
+        knead_emit_reconcile_event "Failed" "node read failed" "Error"
         return 1
     fi
     if ! node_name="$(printf '%s' "${node_json}" | jq -er '.metadata.name')"; then
         echo "failed to read node name"
+        knead_emit_reconcile_event "Failed" "node name read failed" "Error"
         return 1
     fi
     if ! goal="$(knead_get_node_annotation "${node_json}" "${KNEAD_COMPONENT_GOAL_ANNOTATION}")"; then
         echo "failed to read live-patching goal annotation"
+        knead_emit_reconcile_event "Failed" "goal annotation read failed" "Error"
         return 1
     fi
     if [ -z "${goal}" ]; then
@@ -327,6 +369,7 @@ knead_main() {
 
     if ! status="$(knead_get_node_annotation "${node_json}" "${KNEAD_COMPONENT_STATUS_ANNOTATION}")"; then
         echo "failed to read live-patching status annotation"
+        knead_emit_reconcile_event "Failed" "status annotation read failed, goal=${goal}" "Error"
         return 1
     fi
     # Skip reconciliation only when this goal was processed and every component succeeded.
@@ -355,6 +398,9 @@ knead_main() {
 
     if [ "${result}" -eq 0 ]; then
         echo "knead-component completed successfully"
+        knead_emit_reconcile_event "Succeeded" "goal=${goal}"
+    else
+        knead_emit_reconcile_event "Failed" "goal=${goal}" "Error"
     fi
     return "${result}"
 }
--- a/parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
@@ -55,7 +55,7 @@ knead_get_node_annotation() {
     local node_json="$1"
     local annotation="$2"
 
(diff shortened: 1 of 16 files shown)


parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -3,151 +3,357 @@
...
+# nodeConfig schema. The goal must be the bare sha256 digest of the exact
+# ConfigMap value read by this node.
+knead_read_configmap() {
+    local goal="$1"
+    local payload
+    local payload_hash
+
+    # shellcheck disable=SC2086
+    if ! payload="$($KUBECTL get cm -n "${KNEAD_COMPONENT_CONFIG_NAMESPACE}" "${KNEAD_COMPONENT_CONFIGMAP}" -o "jsonpath={.data.${KNEAD_COMPONENT_CONFIG_KEY_JSONPATH}}")"; then
+        echo "failed to read live-patching-config ConfigMap" >&2
```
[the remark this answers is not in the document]
chmill-zz: I'm not sure what copilot is getting at here. Looks good from here

KUBECTL=kubectl \
KNEAD_COMPONENT_CONFIG_NAMESPACE=kube-system \
KNEAD_COMPONENT_CONFIGMAP=live-patching-config \
KNEAD_COMPONENT_CONFIG_KEY_JSONPATH='live-patching-config\.json' \
__SOURCED__=1 \
bash -c '
    source parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
    knead_read_configmap \
        a39c2e4dd1f80854162ad925fdb6b74c4a7b415c5d75c8ee39094453e9df68b9 |
        jq -e .
'
{
  "components": [
    {
      "name": "securityPatch",
      "nodeConfig": "{\"agentPools\":{\"nodepool1\":{\"goldenTimestamp\":\"20260710T000000Z\"}}}"
    }
  ]
}

and just full output - 

k get cm -n kube-system live-patching-config -ojson
{
    "apiVersion": "v1",
    "data": {
        "live-patching-config.json": "{\"components\":[{\"name\":\"securityPatch\",\"nodeConfig\":\"{\\\"agentPools\\\":{\\\"nodepool1\\\":{\\\"goldenTimestamp\\\":\\\"20260710T000000Z\\\"}}}\"}]}"
    },
    "kind": "ConfigMap",
    "metadata": {
        "creationTimestamp": "2026-07-15T05:51:14Z",
        "name": "live-patching-config",
        "namespace": "kube-system",
        "resourceVersion": "90353",
        "uid": "19ba9af1-da09-4504-9a34-67ad3b73259d"
    }
}

spec/parts/linux/cloud-init/artifacts/security-update_spec.sh
```diff
@@ -0,0 +1,185 @@
...
+    End
+
+    It 'retries unattended upgrade until it succeeds'
+        eval 'unattended-upgrade() {
+            TEST_UNATTENDED_ATTEMPT=$((TEST_UNATTENDED_ATTEMPT + 1))
+            echo "unattended-upgrade called: ${TEST_UNATTENDED_ATTEMPT}"
+            [ "${TEST_UNATTENDED_ATTEMPT}" -ge 3 ]
+        }'
+        TEST_UNATTENDED_ATTEMPT=0
+        export TEST_UNATTENDED_ATTEMPT
```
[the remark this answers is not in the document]
chmill-zz: huh. So we have apt-get() in cse_helpers and systemd-sysext() in cse_install_spec

so this is following pattern of "-" for executable wrapping.

spec/parts/linux/cloud-init/artifacts/security-update_spec.sh
```diff
@@ -0,0 +1,185 @@
...
+    End
+
+    It 'fails after ten unattended upgrade attempts'
+        eval 'unattended-upgrade() {
+            TEST_UNATTENDED_ATTEMPT=$((TEST_UNATTENDED_ATTEMPT + 1))
+            echo "unattended-upgrade called: ${TEST_UNATTENDED_ATTEMPT}"
+            return 1
+        }'
+        TEST_UNATTENDED_ATTEMPT=0
+        export TEST_UNATTENDED_ATTEMPT
```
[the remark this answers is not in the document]
chmill-zz: same https://github.com/Azure/AgentBaker/pull/8952#discussion_r3591207931

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -3,151 +3,357 @@
...
+    return "${result}"
 }
 
 ${__SOURCED__:+return}
 # --------------------------------------- Main Execution starts here --------------------------------------------------
 
-# source apt_get_update
-source /opt/azure/containers/provision_source_distro.sh
+# shellcheck disable=SC1091
+source /opt/azure/containers/security-update.sh
```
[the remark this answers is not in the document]
chmill-zz: This handler is intentionally a hard dependency. The VHD and hotfix paths deliver `ubuntu-snapshot-update.sh` and `security-update.sh` together, and `snapshot-update-service_spec.sh` validates both delivery paths. If the handler is missing, source already exits nonzero and reports the exact missing path, so an additional existence check would duplicate the same failure behavior without improving recovery.

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -3,151 +3,357 @@
...
+    fi
+    if ! status="$(printf '%s' "${KNEAD_COMPONENT_RESULTS}" | jq -c --arg currentHash "${goal}" '{currentHash: $currentHash, components: .}')" || [ -z "${status}" ]; then
+        echo "failed to render live-patching status annotation"
+        return 1
     fi
 
-    code_name=$(lsb_release -cs)
-    generate_sources_list "${repo_endpoint}" "${golden_timestamp}" "${code_name}"
+    # shellcheck disable=SC2086
+    $KUBECTL annotate --overwrite node "${node_name}" "${KNEAD_COMPONENT_STATUS_ANNOTATION}=${status}"
```
YaoC, line 334: Can we keep writing `kubernetes.azure.com/live-patching-current-timestamp` for a while (in addition to the new `live-patching-status` annotation)? The server-side (RP) still has logic that verifies security-patch status via `current-timestamp`, so dropping it here means new-VHD/hotfixed nodes report convergence through a channel the RP doesn't read yet — patching can look stalled server-side even when it succeeded.

Alternatively, we need to guarantee the RP change that reads `live-patching-status` is rolled out to **every region before this VHD releases** (and stays ahead for the full 6-month VHD support window). Otherwise there's a window where the node's reported state and the RP's verifier disagree. Keeping the old annotation as a temporary dual-write is the lower-risk path until the RP rollout is confirmed complete everywhere.

chmill-zz: Added the temporary dual-write in c9c5b3e2dd. I placed the legacy `live-patching-current-timestamp` write in the `securityPatch` handler rather than the generic reconciler, since the handler owns `goldenTimestamp` and knows whether patching succeeded. `ubuntu-snapshot-update.sh` continues to write the generic `live-patching-status`. A legacy annotation failure is treated as a component failure so we do not report false convergence to the existing RP reader.

parts/linux/cloud-init/nodecustomdata.yml
```diff
@@ -152,6 +152,28 @@ write_files:
...
+{{- else if IsFlatcar }}
+{{- else }}
+- path: /opt/azure/containers/ubuntu-snapshot-update.sh
+  permissions: "0544"
+  encoding: gzip
+  owner: root
+  content: !!binary |
+    {{GetVariableProperty "cloudInitData" "snapshotUpdateScript"}}
+
+- path: /opt/azure/containers/security-update.sh
```
saewoni, line 171: is this part only supposed to be for Ubuntu?

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -3,151 +3,357 @@
...
+    local node_name
+    local node_json
+    local goal
+    local status
+    local payload
+    local result=0
 
-    export APT_CONFIG="${SECURITY_PATCH_CONFIG_DIR}/apt.conf"
+    knead_wait_for_kubeconfig
+    if ! node_json="$(knead_read_node)"; then
```
saewoni, line 349: Is `knead_read_node` intended only for the LPC/generic ConfigMap path, not the LPS/bootstrap path? `knead_main` waits for `/var/lib/kubelet/kubeconfig` before calling `knead_read_node`, and then `knead_read_node` does `$KUBECTL get node "${node_name}" -o json`. That means this flow only starts after kubelet credentials exist and assumes the VM can read its Kubernetes Node object, which seems incompatible with the LPS pre-bootstrap use case where config is needed before kubelet registration/API access.

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -3,151 +3,357 @@
...
+
+KNEAD_COMPONENT_RESULTS='{}'
+KNEAD_COMPONENT_RESULTS_VALID=true
 
-main() {
-    # At startup, we need to wait for kubelet to finish TLS bootstrapping to create the kubeconfig file.
-    while [ ! -f ${KUBECONFIG} ]; do
-        echo 'Waiting for TLS bootstrapping'
+# Waits for kubelet credentials so kubectl can read Node and ConfigMap state.
+knead_wait_for_kubeconfig() {
```
saewoni, line 62: Should this wait have a timeout? As written, `knead_wait_for_kubeconfig` can block forever if kubelet never writes `/var/lib/kubelet/kubeconfig`. Since this runs from the snapshot-update timer/service, a stuck invocation could prevent clean retry behavior and make failures harder to observe. The previous behavior also waited indefinitely, but if this is becoming the generic live-patching entrypoint, it may be worth bounding the wait and returning failure so systemd/timer retries handle it.

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
```diff
@@ -3,151 +3,357 @@
...
+
+    printf '%s' "${node_json}" | jq -r --arg annotation "${annotation}" '.metadata.annotations[$annotation] // empty'
+}
+
+# Reads and validates the generic ConfigMap contract before any component handler runs.
+#
+# Knead validates the component envelope. Each handler owns its decoded
+# nodeConfig schema. The goal must be the bare sha256 digest of the exact
+# ConfigMap value read by this node.
+knead_read_configmap() {
```
saewoni, line 104: If LPC emits an invalid top-level live-patching-config envelope, this fails before dispatching any component, so one bad component entry could block the whole generic node-side flow, including securityPatch. Is that intentional? Should invalid component entries be isolated/skipped like unsupported components, or should LPC guarantee this envelope is fully validated before stamping the goal hash?

parts/linux/cloud-init/nodecustomdata.yml
chmill-zz: yeah ubuntu only. the mariner side has mariner-package-update.sh that we'll update in a different pr

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
chmill-zz: this update of knead has been written for LPC which is what the original service was doing here. And that has to wait for kubelet before anything else can be done. I lean towards a separate service for LPS

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
chmill-zz: the service is set with a 60min timeout. and then snapshot-update.timer sets a OnUnitActiveSec=10min which schedules the service to act every 10min. But yeah if getting the kubeconfig gets stuck we'll just hang out for 60min. failing faster is better, will add something

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
chmill-zz: its intentional. we want the top level validated before updating goal hash. If we can validate the top we independently attempt component updates so we could have mixed success/failure there. If the envelope itself isn't valid we can't safely identify the component set or set a claim on hash configuration

parts/linux/cloud-init/artifacts/ubuntu/ubuntu-snapshot-update.sh
chmill-zz: yeah lps things look like they're over in aks-node-controller.service / aks-node-controller-launcher.sh and check-hotfix in checkhotfix.go

spec/parts/linux/cloud-init/artifacts/snapshot-update-service_spec.sh
```diff
@@ -0,0 +1,83 @@
...
+
+Describe 'snapshot update service commands'
+    validate_ubuntu_packer_inputs() {
+        local packer_file
+
+        for packer_file in \
+            vhdbuilder/packer/vhd-image-builder-base.json \
+            vhdbuilder/packer/vhd-image-builder-cvm.json \
+            vhdbuilder/packer/vhd-image-builder-arm64-gen2.json \
+            vhdbuilder/packer/vhd-image-builder-arm64-gb.json
```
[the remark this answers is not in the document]
chmill-zz: The ACL templates upload these shared inputs to `/home/packer`, but ACL does not install or run the Ubuntu updater. `packer_source.sh` has a dedicated `isACL` branch (lines 605-612); only the Ubuntu `else` branch installs both `security-update.sh` and `ubuntu-snapshot-update.sh` (lines 613-621). Provisioning likewise enables `snapshot-update.timer` only for Ubuntu or Mariner/Azure Linux (`cse_main.sh` lines 656-659), excluding ACL. The VHD content test also scopes these assets to `OS_SKU=Ubuntu` (`linux-vhd-content-test.sh` lines 2495-2509). Adding the Ubuntu apt handler to ACL would be incorrect, so no change is needed here.

PR #8960, opened 2026-07-15
fix: add strict deb version match in VHD containerd content test

abigailliang-aks-sig-node:
## Summary
Add strict full-version matching in the VHD content test for containerd and cri-tools, catching version drift at VHD build time rather than in e2e.

## Problem
The VHD content test (`testContainerd`/`testCriCtl`) previously only compared `major.minor.patch` (stripping the hotfix suffix via `cut -d- -f1`). This meant:
- `components.json` says `2.3.2-ubuntu24.04u1`
- VHD build installs `2.3.2-ubuntu24.04u2` (due to apt wildcard `moby-containerd=2.3.2*`)
- Content test passes (only checks `2.3.2` == `2.3.2`)
- E2e validator fails (exact match of full version string)

## Changes
- Add `assertPackageVersion` helper (inline in `linux-vhd-content-test.sh`): uses `dpkg-query` (deb) or `rpm` (RPM-based) to compare the full installed version against components.json
- Use explicit hardcoded package name mapping per OS (not renovateTag parsing, since renovateTag is metadata for Renovate bot and its format is not guaranteed stable)
  - `kubernetes-cri-tools` → `kubernetes-cri-tools` (same across all OS)
  - `containerd` → `moby-containerd` (Ubuntu/Mariner) or `containerd2` (Azure Linux 3.0+)
- Fail fast when package name cannot be resolved
- Fix `err()` call signatures to match the two-argument format (`$1`=test name, `$2`=message)
- Rename `debPackageName` → `installedPackageName` since the check now covers both deb and rpm
- Script is fully self-contained — no external helper files needed (runs on VM without dependencies)

## Follow-up items
1. **ARM64 support**: `testPackagesInstalled` currently skips entirely on ARM64 (line 209), meaning ARM64 VHDs have no package version validation. Opening this up to ARM64 would require verifying that all packages in `components.json` have correct ARM64 entries and that the download/install paths work for arm64.

## Test plan
- [x] VHD build with mismatched version fails at content test (not at e2e) — verified by intentionally setting containerd version to `u99`:
  ```
  testContainerd:Error: installed moby-containerd version '2.3.2-ubuntu24.04u2' does not match expected '2.3.2-ubuntu24.04u99' from components.json
  ```
  Failed in "Test, Scan, and Cleanup" stage as expected.
- [x] shellspec CI passes locally with no regressions (804 examples, 0 failures)
- [x] VHD build with matching `components.json` version passes content test (will be verified once CI runs with correct version)


vhdbuilder/packer/test/linux-vhd-content-test.sh
cameronmeissner, line 238: I was just noticing that we skip these checks on ARM64 yesterday while working on 26.04 - maybe you might consider opening this up to ARM64 as well?

abigailliang-aks-sig-node: Good point — ARM64 skips testPackagesInstalled entirely Opening it up would require verifying that all packages in components.json have correct ARM64 entries and that the download/install paths work for arm64.

 I'd prefer to keep this PR focused on the strict  version match and tackle ARM64 support as a follow-up.

cameronmeissner: let's definitely take care of this as a follow-up

vhdbuilder/packer/test/test_pkg_name_extraction.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
Devinwong: This script file is not called or referenced by anyone, right?

abigailliang-aks-sig-node: this a local test file starting as `test_...`.

I originally planned to add a ShellSpec test, but `ces_hepler.sh` does not work well in the ShellSpec Docker environment. As a temporary solution, I’m keeping this as a local test script for manual validation.

 And one of the follow-up is to figure out a way to unit test vhdbuilder/packer/test/linux-vhd-content-test.sh in ci

vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -277,11 +277,17 @@ testPackagesInstalled() {
...
-            ;;
-          "containerd")
-            testContainerd "$version"
+          "kubernetes-cri-tools"|"containerd")
+            # Resolve the actual installed package name from components.json renovateTag.
+            # The components.json "name" field may differ from the real package name
+            # (e.g. "containerd" -> moby-containerd on Ubuntu, containerd2 on Azure Linux 3.0).
+            local pkgName
+            pkgName=$(getPackageJSON "$p" "${OS}" "${OS_VERSION}" "${OS_VARIANT}" | jq -r '.versionsV2[0].renovateTag // empty' | sed -n 's/.*name=\([^,]*\).*/\1/p' | xargs)
+            if [ "$name" = "kubernetes-cri-tools" ]; then
```
Devinwong: We already have case switch at line 279. I will try not to add another if else block to filter which package name again.

abigailliang-aks-sig-node: good catch.

first line 279`"kubernetes-cri-tools"|"containerd"` is not a **package name**. they are `name` field in conponents.json. in 
[line 217](https://github.com/Azure/AgentBaker/blob/44a8c6513b463774d22e7d6aa05d4fa6f2049b36/vhdbuilder/packer/test/linux-vhd-content-test.sh#L217)

The actually package could either be `moby-containerd` or `containerd2`.

and in kubernetes-cri-tools case. it is happenes that the package name is the same as what in conponents.json name.

The way we get pkgname at line 285, which shared by both "kubernetes-cri-tools"|"containerd".

vhdbuilder/packer/test/test_pkg_name_extraction.sh
Devinwong: If it's not added any CI, or local test bundle (like `make generate`). I am afraid that a few weeks later you will be the only one who knows this presence of this test script😅 and it will become dangling. Why did you say cse_helper.sh doesn't work well? There is an existing `cse_helper_spec.sh`.

vhdbuilder/packer/test/linux-vhd-content-test.sh
Devinwong: Oh I see your purpose.
However, I will suggest we should not parse the renovateTag here because its purpose was for Renovate to get the metadata, which acts more like a comment and the format is not that strict that could be break your parsing here unintentionally in the future. For example, if someone doesn't put a renovateTag, Renovate will still work (by not updating it) but your test will break.
Maybe we can keep both case like
case "kubernetes-cri-tools": $pkgName=$name
case "containerd": if ubuntu then $pkgName = moby-containerd, etc.
The is content test so I think it's fine to keep it more verbose (or explicit)

vhdbuilder/packer/test/linux-vhd-content-test.sh
djsly: created work item to track

vhdbuilder/packer/test/test_pkg_name_extraction.sh
lilypan26: Agreed, i think it would be best to add as a shellspec. If not possible, i honestly don't think its necessary since its a test for a test, not super vital if we don't have it since the vhd content test will catch it

abigailliang-aks-sig-node: The reason I initially thought integrating `cse_helpers.sh` with ShellSpec would be tricky was that I was trying to use **source** to load the script, which failed in the Docker environment due to **missing system dependencies** like `lsb_release`. So I left it as a local-only test script and planned to figure out the CI integration in a follow-up PR.

but your concerns here make sense. i could figure out a way to run a shell spec as the sample you mentioned here. thx.

also one thing i dont fully understand here. `make generate `doesn't corelated to this change —you mean we have an `cse_helpers_spec.sh` file can run the `cse_helpers.sh`

abigailliang-aks-sig-node: Update: I removed the ShellSpec test since it had become redundant. It was only comparing hardcoded strings, and now that the package mapping lives directly in the content test cases, there isn't much left for it to test.

Thanks for the big-picture perspective—that was really helpful! 🙏 @Devinwong and @lilypan26

Devinwong: Thanks for making the changes!
But as Lily suggested, it's rare to see unit testing a test-only function.
IMO, we generally just test production codes. Although it's no harm to test a test-only function, however if one day your `spec/vhdbuilder/packer/test/assert_package_version_spec.sh` fails, you will need to check 2 places (your test codes and the production codes). You can have local test to make sure your test-only function is valid. But once it's confirmed, we will trust it's always reliable.
So I will just remove the `spec/vhdbuilder/packer/test/assert_package_version_spec.sh` if that makes sense to you.

Devinwong (approved):
LGTM. Thanks for making the changes

PR #9004, opened 2026-07-21
test: extend gpu e2e pipeline jobTimeoutInMinutes to 2 hours

sulixu:
 GPU-only 120-minute timeout.

  - Added jobTimeoutInMinutes, default 90, to .pipelines/templates/e2e-template.yaml:21.
  - Set GPU pipeline to 120 in .pipelines/e2e-gpu.yaml:39.
  - Other E2E pipelines remain at 90 minutes.
  
Fixes #

### DIFF
--- a/.pipelines/e2e-gpu.yaml
+++ b/.pipelines/e2e-gpu.yaml
@@ -36,4 +36,5 @@ jobs:
     parameters:
       name: Linux GPU Tests
       IgnoreScenariosWithMissingVhd: false
+      jobTimeoutInMinutes: 120
       failedTestsRetryCount: ${{ variables.E2E_FAILED_TESTS_RETRY_COUNT }}
--- a/.pipelines/templates/e2e-template.yaml
+++ b/.pipelines/templates/e2e-template.yaml
@@ -18,6 +18,10 @@ parameters:
     type: number
     displayName: Number of times to retry failed E2E tests
     default: 0
+  - name: jobTimeoutInMinutes
+    type: number
+    displayName: Maximum duration of the E2E job
+    default: 90
   - name: useVhdMetadataArtifacts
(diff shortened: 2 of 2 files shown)


.pipelines/e2e-gpu.yaml
```diff
@@ -3,10 +3,9 @@ variables:
   TAGS_TO_RUN: "gpu=true"
   TAGS_TO_SKIP: "os=windows"
   SKIP_E2E_TESTS: false
-  SKIP_TESTS_WITH_SKU_CAPACITY_ISSUE: true
+  SKIP_TESTS_WITH_SKU_CAPACITY_ISSUE: false
```
ganeshkumarashok: let's get capacity first before switching it back on?

.pipelines/templates/e2e-template.yaml
```diff
@@ -18,6 +18,10 @@ parameters:
     type: number
     displayName: Number of times to retry failed E2E tests
     default: 0
+  - name: jobTimeoutInMinutes
+    type: number
+    displayName: Maximum duration of the E2E job
+    default: 90
```
ganeshkumarashok, line 24: increase the time here?

.pipelines/e2e-gpu.yaml
sulixu: done

.pipelines/templates/e2e-template.yaml
cameronmeissner: shouldn't matter since we're providing an override from the gpu yaml - I see @sulixu is just matching what we previously were providing directly to E2E job below

cameronmeissner (approved):

ganeshkumarashok (approved):

PR #8096, opened 2026-03-16
feat(rcv1p): unify cert bootstrap flow and add Windows CA refresh task

rchincha:
## What this PR does / why we need it

This PR implements **RCV1P (Robust Certificate Validation for 1P)** — the next-generation mechanism for distributing Azure root CA certificates to AKS nodes. Instead of hardcoding certificate bundles, RCV1P queries the Azure wireserver at provisioning time to download and install the latest root certificates into the OS trust store.

**Reference**: https://eng.ms/docs/products/onecert-certificates-key-vault-and-dsms/onecert-customer-guide/autorotationandecr/rcv1ptsg

---

#### 1. Linux: Unified cert bootstrap flow (`init-aks-custom-cloud.sh`)
- **Consolidated 3 scripts into 1**: Removed `init-aks-custom-cloud-mariner.sh`, `init-aks-custom-cloud-operation-requests-mariner.sh`, and `init-aks-custom-cloud-operation-requests.sh`. All cert logic now flows through a single `init-aks-custom-cloud.sh` that detects the distro (Ubuntu, Mariner, AzureLinux, Flatcar, ACL) at runtime.
- **Extracted repo depot logic**: Moved Ubuntu repo depot initialization and chrony configuration into a new `init-aks-custom-cloud-repos.sh` to keep base `customData` size small for non-custom-cloud scenarios (critical for Flatcar/ACL which have tight size limits).
- **Two cert endpoint modes**: `legacy` (ussec/usnat regions) and `rcv1p` (all other regions), selected by cloud location at runtime.
- **Fatal failure handling**: Wireserver cert retrieval failures are fatal on both Linux and Windows — after exhausting retries, provisioning fails rather than continuing without certificates. The cert mode (legacy vs rcv1p) is always determined and must succeed.

#### 2. Windows: CA cert refresh task and rcv1p support (`kubernetesfunc.ps1`)
- **New `Get-CACertificates` with `-Location` parameter**: Determines cert endpoint mode from location, uses legacy endpoint for ussec/usnat, rcv1p for all others.
- **New `Register-CACertificatesRefreshTask`**: Registers a daily scheduled task to refresh CA certificates, with backward compatibility for older VHDs that don't accept `-Location`.
- **New `Should-InstallCACertificatesRefreshTask`**: Gates refresh task registration on wireserver opt-in status.
- **Pester tests**: 260 lines of tests covering URI construction, endpoint mode selection, and refresh task gating.

[...]

PR #9011, opened 2026-07-22
fix: preserve location for linux provisioning

djsly:
## Summary

Preserve `LOCATION` for the legacy Linux CSE provisioning path so `configureAzureJson` writes a non-empty `/etc/kubernetes/azure.json` `location` value. This fixes the VirtualNodeAddon regression where `aci-connector-linux` crashed because ACI region was empty.

The observed customData already contains `LOCATION=<region>`, but it is not exported. After the shell assignment was moved away from the final provisioning env block, the legacy `nohup /bin/bash ... provision_start.sh` child process no longer inherited `LOCATION`.

## Changes

- Pass scoped `REPO_DEPOT_ENDPOINT` and `LOCATION` into `init-aks-cloud` when that script is present.
- Restore `LOCATION` to the contiguous legacy CSE env assignment block that is flattened into the final `nohup` provisioning command.
- Add source-only template comments warning that the env assignment block must remain contiguous through `nohup`.
- Add a parser assertion confirming the scriptless/aks-node-controller path already carries `LOCATION` through `cmd.Env`; no scriptless runtime command change is needed.

## Validation

- `cd aks-node-controller && GENERATE_TEST_DATA=true go test ./parser/...`
- `cd aks-node-controller && go test ./parser/...`
- `go test ./pkg/agent...`
- `git diff --check`

Note: `make generate` was attempted, but the repo tool install path hit an internal Go proxy `401 Unauthorized` while installing `github.com/onsi/ginkgo/v2/ginkgo@v2.1.1`, so I used the targeted generation/test commands above.

## Related

- AB#38891323


parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -17,7 +17,7 @@ else
 fi;
 {{end}}
 REPO_DEPOT_ENDPOINT="{{AKSCustomCloudRepoDepotEndpoint}}"
-LOCATION={{GetVariable "location"}}
```
awesomenix, line 20: Hmm! but there are other variables here which is passed in, what makes this variable special? 

Example:

ServicePrincipalID, which is also written in azure.json

djsly: Good question. The difference is the move in #8096: `LOCATION` was moved out of the later inline env block that launches `provision_start.sh` and into a standalone shell assignment before `init-aks-cloud`. Standalone shell assignments are not inherited by the later child bash unless exported, so `configureAzureJson` sees empty `LOCATION`. Variables like the service principal fields stayed in the later provisioning env block, so they continue to be inherited by the provisioning child.

I pushed follow-up commit 494dc4ed9c to make the shell behavior explicit and also preserve `REPO_DEPOT_ENDPOINT` for the init path.

parts/linux/cloud-init/artifacts/cse_cmd.sh
```diff
@@ -17,7 +17,7 @@ else
 fi;
 {{end}}
 REPO_DEPOT_ENDPOINT="{{AKSCustomCloudRepoDepotEndpoint}}"
```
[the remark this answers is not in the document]
djsly: Fixed in 494dc4ed9c by exporting `REPO_DEPOT_ENDPOINT` before `init-aks-cloud` runs. I also added semicolons so the rendered one-line command cannot accidentally fold later assignments into the `export` builtin.

aks-node-controller/parser/parser_test.go
```diff
@@ -586,6 +590,16 @@ func assertHasKeyWithValue[K comparable, V any](t *testing.T, m map[K]V, key K,
...
 
+func assertExportsLocationBeforeProvisionStart(t *testing.T, cmd *exec.Cmd) {
+	command := cmd.String()
+	locationExportIndex := strings.Index(command, `export LOCATION="${LOCATION}";`)
+	provisionStartIndex := strings.Index(command, "/opt/azure/containers/provision_start.sh")
+
+	require.NotEqual(t, -1, locationExportIndex, "expected generated command to export LOCATION")
+	require.NotEqual(t, -1, provisionStartIndex, "expected generated command to start provision_start.sh")
+	assert.Less(t, locationExportIndex, provisionStartIndex, "LOCATION must be exported before provision_start.sh starts")
+}
```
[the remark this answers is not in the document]
djsly: Fixed in 494dc4ed9c. The helper now asserts `LOCATION` is exported before both the `init-aks-cloud` invocation and `provision_start.sh`.

djsly:
Updated the PR to match the review/Teams direction:

- Removed the broad `export LOCATION` / `export REPO_DEPOT_ENDPOINT` approach.
- `init-aks-cloud` now receives only the scoped env it needs: `REPO_DEPOT_ENDPOINT` and `LOCATION`.
- `provision_start.sh` now receives scoped `LOCATION`, preserving the region for `configureAzureJson` without globally exporting it.
- The extracted node customData confirmation (`LOCATION=eastus2` present, no `export LOCATION`) supports the refined RCA: RP/customData input is correct; the failure is shell-scope/process propagation after the LOCATION assignment was moved away from the provisioning env block.

Validated with:
- `cd aks-node-controller && GENERATE_TEST_DATA=true go test ./parser/...`
- `cd aks-node-controller && go test ./parser/...`
- `go test ./pkg/agent...`
- `git diff --check`

awesomenix (approved):

aks-node-controller/parser/templates/cse_cmd.sh.gtpl
```diff
@@ -1,5 +1,8 @@
 echo $(date),$(hostname) > ${PROVISION_OUTPUT};
 if [ -f "${INIT_AKS_CLOUD_FILEPATH}" ]; then
-	"${INIT_AKS_CLOUD_FILEPATH}" >> /var/log/azure/cluster-provision.log 2>&1 || exit $?;
+	REPO_DEPOT_ENDPOINT="${REPO_DEPOT_ENDPOINT}" LOCATION="${LOCATION}" "${INIT_AKS_CLOUD_FILEPATH}" >> /var/log/azure/cluster-provision.log 2>&1 || exit $?;
 fi;
-/usr/bin/nohup /bin/bash -c "/bin/bash /opt/azure/containers/provision_start.sh"
+{{/*
+Keep LOCATION inline with nohup; BuildCSECmd flattens this template into one shell command,
+so inserting runtime control flow or command separators here can stop LOCATION from reaching provision_start.sh.
+*/}}LOCATION="${LOCATION}" /usr/bin/nohup /bin/bash -c "/bin/bash /opt/azure/containers/provision_start.sh"
```
djsly: why is location added here.  ?

aks-node-controller/parser/templates/cse_cmd.sh.gtpl
```diff
@@ -1,5 +1,8 @@
 echo $(date),$(hostname) > ${PROVISION_OUTPUT};
 if [ -f "${INIT_AKS_CLOUD_FILEPATH}" ]; then
-	"${INIT_AKS_CLOUD_FILEPATH}" >> /var/log/azure/cluster-provision.log 2>&1 || exit $?;
+	REPO_DEPOT_ENDPOINT="${REPO_DEPOT_ENDPOINT}" LOCATION="${LOCATION}" "${INIT_AKS_CLOUD_FILEPATH}" >> /var/log/azure/cluster-provision.log 2>&1 || exit $?;
 fi;
-/usr/bin/nohup /bin/bash -c "/bin/bash /opt/azure/containers/provision_start.sh"
+{{/* Keep LOCATION inline with nohup. */ -}}
+{{/* BuildCSECmd flattens this template into one shell command, so this assignment is passed to nohup. */ -}}
+{{/* Be careful not to add runtime control flow or command separators that break the flattening logic. */ -}}
+LOCATION="${LOCATION}" /usr/bin/nohup /bin/bash -c "/bin/bash /opt/azure/containers/provision_start.sh"
```
djsly: is this really needed ? isn't the issue affecting CSE_CMD.sh only ?

aks-node-controller/parser/templates/cse_cmd.sh.gtpl
djsly: Agreed — it should not be added in the scriptless template. I pushed `51eb9f936b` to remove the inline `LOCATION` from `aks-node-controller/parser/templates/cse_cmd.sh.gtpl` and keep only the existing `cmd.Env` coverage/assertion for scriptless. The actual RCA/fix remains limited to legacy `parts/linux/cloud-init/artifacts/cse_cmd.sh`, where the flattened inline env block feeds the final `nohup` provisioning command.

aks-node-controller/parser/templates/cse_cmd.sh.gtpl
djsly: Correct, the runtime issue affects the legacy `cse_cmd.sh` path. Scriptless already passes `LOCATION` through `BuildCSECmd` via `cmd.Env`, so the inline template change was redundant. I removed it in `51eb9f936b` and left only the `cmd.Env` regression assertion.

awesomenix (approved):

cameronmeissner (approved):

rchincha (approved):

PR #9019, opened 2026-07-23
chore: add node-exporter validation before version bump

chmill-zz:
## What this does
- parses the node-exporter scrape as Prometheus exposition data instead of checking for any `node_` text
- validates every node-exporter metric family retained by the AKS default Prometheus profile
- verifies the service is persistently enabled and remains active after scraping
- adds focused unit coverage for valid and missing metric families

## Why
Node-exporter is currently pinned to 1.9.1 and PMC now carries 1.12.1. This establishes the existing metrics contract before taking the multi-minor version update in a separate PR.

## Validation
- `go test . -run TestValidateNodeExporterMetrics`
- `go test . -run ^`
- `go test ./components`
- `go vet .`
- `git diff --check`

### DIFF
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/coder/websocket v1.8.14
 	github.com/joho/godotenv v1.5.1
 	github.com/klauspost/compress v1.18.5
-	github.com/prometheus/common v0.67.5
 	github.com/samber/lo v1.52.0
 	github.com/sanity-io/litter v1.5.5
 	github.com/stretchr/testify v1.11.1
@@ -88,7 +87,6 @@ require (
 	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/tidwall/match v1.1.1 // indirect
 	github.com/tidwall/pretty v1.2.1 // indirect
--- a/e2e/go.mod
+++ b/e2e/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/coder/websocket v1.8.14
 	github.com/joho/godotenv v1.5.1
 	github.com/klauspost/compress v1.18.5
+	github.com/prometheus/common v0.67.5
 	github.com/samber/lo v1.52.0
 	github.com/sanity-io/litter v1.5.5
 	github.com/stretchr/testify v1.11.1
@@ -87,6 +88,7 @@ require (
 	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/tidwall/match v1.1.1 // indirect
 	github.com/tidwall/pretty v1.2.1 // indirect
--- a/e2e/nodeexporter/metrics.go
+++ b/e2e/nodeexporter/metrics.go
@@ -1,0 +1,47 @@
+package nodeexporter
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ValidateMetrics verifies that a node-exporter scrape contains every metric
+// retained by the AKS default Prometheus profile.
+func ValidateMetrics(metricsText string) error {
+	metricNames := make(map[string]struct{})
+	for line := range strings.SplitSeq(metricsText, "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		if end := strings.IndexAny(line, "{ \t"); end > 0 {
+			metricNames[line[:end]] = struct{}{}
+		}
+	}
+
+	requiredMetrics := []string{
+		"node_disk_read_time_seconds_total",
+		"node_disk_reads_completed_total",
+		"node_disk_write_time_seconds_total",
+		"node_disk_writes_completed_total",
+		"node_memory_MemAvailable_bytes",
+		"node_network_receive_bytes_total",
+		"node_network_receive_errs_total",
+		"node_network_receive_packets_total",
+		"node_network_transmit_bytes_total",
+		"node_network_transmit_errs_total",
+		"node_network_transmit_packets_total",
+		"node_netstat_Tcp_RetransSegs",
+		"node_pressure_cpu_waiting_seconds_total",
+		"node_filesystem_free_bytes",
+		"node_filesystem_size_bytes",
+	}
+	for _, name := range requiredMetrics {
+		if _, exists := metricNames[name]; !exists {
+			return fmt.Errorf("required metric %q is missing", name)
+		}
+	}
+
+	return nil
+}
--- a/e2e/nodeexporter/metrics.go
+++ b/e2e/nodeexporter/metrics.go
@@ -37,10 +37,14 @@ func ValidateMetrics(metricsText string) error {
 		"node_filesystem_free_bytes",
 		"node_filesystem_size_bytes",
 	}
+	var missingMetrics []string
 	for _, name := range requiredMetrics {
 		if _, exists := metricNames[name]; !exists {
-			return fmt.Errorf("required metric %q is missing", name)
+			missingMetrics = append(missingMetrics, name)
 		}
+	}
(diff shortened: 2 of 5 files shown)


e2e/validators.go
```diff
@@ -2292,25 +2294,63 @@ func ValidateNodeExporter(ctx context.Context, s *Scenario) {
...
-	}
-	execScriptOnVMForScenarioValidateExitCode(ctx, s, strings.Join(command, "\n"), 0, "node-exporter should be listening on port 19100 and serving metrics over HTTP")
+	// Validate the metrics contract consumed by the AKS Prometheus default profile. Scrape the node IP directly
+	// so this also verifies that the endpoint is reachable on the address used by monitoring infrastructure.
+	s.T.Logf("Validating node-exporter metrics on port 19100")
+	metricsURL := fmt.Sprintf("http://%s:19100/metrics", s.Runtime.VM.PrivateIP)
+	scrapeAndValidateNodeExporter(ctx, s, metricsURL)
+
+	execScriptOnVMForScenarioValidateExitCode(ctx, s, "systemctl is-active node-exporter.service", 0,
+		"node-exporter should remain active after scraping")
```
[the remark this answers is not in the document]
chmill-zz: Fixed in 5eecc36ac1 by reusing `serviceName` for the post-scrape liveness check.

e2e/validators.go
```diff
@@ -2292,25 +2294,63 @@ func ValidateNodeExporter(ctx context.Context, s *Scenario) {
...
+	execScriptOnVMForScenarioValidateExitCode(ctx, s, fmt.Sprintf("systemctl is-active %s", serviceName), 0,
+		"node-exporter should remain active after scraping")
 
 	s.T.Logf("node-exporter validation passed")
 }
 
+func scrapeAndValidateNodeExporter(ctx context.Context, s *Scenario, metricsURL string) {
+	s.T.Helper()
+
+	result := execScriptOnVMForScenario(ctx, s, fmt.Sprintf("curl --noproxy '*' -fsS --max-time 10 %s", metricsURL))
```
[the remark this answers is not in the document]
chmill-zz: Good catch. The original validator used `-sS`; `-f` was introduced in this PR and would hide useful HTTP error bodies. Restored `-sS` and quoted the URL in a2e31bc66f.

e2e/validators.go
```diff
@@ -2292,25 +2294,63 @@ func ValidateNodeExporter(ctx context.Context, s *Scenario) {
...
 
+func scrapeAndValidateNodeExporter(ctx context.Context, s *Scenario, metricsURL string) {
+	s.T.Helper()
+
+	result := execScriptOnVMForScenario(ctx, s, fmt.Sprintf("curl --noproxy '*' -sS --max-time 10 %q", metricsURL))
+	require.Equal(s.T, "0", result.exitCode,
+		"node-exporter scrape failed\nstdout: %s\nstderr: %s", result.stdout, result.stderr)
+
+	err := validateNodeExporterMetrics(result.stdout)
+	require.NoError(s.T, err, "node-exporter scrape did not satisfy the AKS Prometheus metrics contract")
```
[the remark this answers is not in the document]
chmill-zz: Added a bounded 2 KB response preview for parse or missing-metric failures in 84a58a676a, while avoiding dumping the full scrape payload.

e2e/go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/coder/websocket v1.8.14
 	github.com/joho/godotenv v1.5.1
 	github.com/klauspost/compress v1.18.5
+	github.com/prometheus/common v0.67.5
```
r2k1: I would prefer avoiding adding dependency without a strong reason.

Prom format is simple and it's easy to extract metric name

chmill-zz: Agreed. Replaced `expfmt` with a small scanner that ignores comments and extracts each sample metric name before labels or whitespace. Removed both added module requirements in 8f9fae379e; focused tests and vet pass.

e2e/validators.go
```diff
@@ -2280,7 +2280,7 @@ func ValidateNodeExporter(ctx context.Context, s *Scenario) {
 	ValidateSystemdUnitIsNotFailed(ctx, s, serviceName)
 
 	// Validate service is enabled
-	execScriptOnVMForScenarioValidateExitCode(ctx, s, fmt.Sprintf("systemctl is-enabled %s", serviceName), 0, fmt.Sprintf("%s should be enabled", serviceName))
+	execScriptOnVMForScenarioValidateExitCode(ctx, s, fmt.Sprintf("systemctl is-enabled %s | grep -qx enabled", serviceName), 0, fmt.Sprintf("%s should be enabled", serviceName))
```
djsly: what was missing before ? I feel is systemctl updates their output this could break, are we saying that `is-enabled` isn't doing what it says it does, fail when not enabled ?

e2e/validators_node_exporter_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestValidateNodeExporterMetricsRejectsMissingMetric(t *testing.T) {
+	metrics := strings.Replace(validNodeExporterMetrics(), `node_network_receive_bytes_total{device="eth0"} 1`, "", 1)
+
+	err := validateNodeExporterMetrics(metrics)
+
+	require.ErrorContains(t, err, `required metric "node_network_receive_bytes_total" is missing`)
+}
+
+func validNodeExporterMetrics() string {
```
djsly: seems flaky ? won't this fails when we bump node-exporter and they start changing metrics ? ordering, etc ?

e2e/validators.go
```diff
@@ -2292,25 +2292,73 @@ func ValidateNodeExporter(ctx context.Context, s *Scenario) {
...
 
 	s.T.Logf("node-exporter validation passed")
 }
 
+func scrapeAndValidateNodeExporter(ctx context.Context, s *Scenario, metricsURL string) {
+	s.T.Helper()
+
+	result := execScriptOnVMForScenario(ctx, s, fmt.Sprintf("curl --noproxy '*' -sS --max-time 10 %q", metricsURL))
+	require.Equal(s.T, "0", result.exitCode,
+		"node-exporter scrape failed\nstdout: %s\nstderr: %s", result.stdout, result.stderr)
```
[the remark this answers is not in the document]
chmill-zz: `require.Equal` does support printf-style `msgAndArgs` in testify v1.11.1. It forwards to `assert.Equal`, and `messageFromMsgAndArgs` calls `fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)`, so stdout/stderr are interpolated here. Keeping this as-is.

e2e/validators_node_exporter_test.go
chmill-zz: The fixture parser is order-independent: it builds a set of metric names from the scrape. Metric-name changes are intentionally expected to fail this test because these are the exact names retained by the aks-operator default Prometheus profile; a rename would otherwise silently break AKS recording/alerting inputs. When the profile intentionally changes, this contract list and fixture should change with it.

e2e/validators.go
chmill-zz: Agreed. `systemctl is-enabled` already returns non-zero for disabled units, and exact output matching was unrelated to this PR while adding output-coupling. Reverted the `grep -qx enabled` change in 70ea128f2c.

e2e/validators_node_exporter_test.go
```diff
@@ -0,0 +1,58 @@
+package e2e
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateNodeExporterMetrics(t *testing.T) {
```
cameronmeissner: are these offline tests that don't require a live VM? curious why we'd add new tests here instead of just enhancing the existing validation

chmill-zz: yeah these are offline/no vm tests. Idea here is we sanity check the metirc name parsing. So if this is healthy but real e2e fails its due to exporter output change/failure rather than parse logic

r2k1: Ideally. I want them to be separate from "Scenario" tests. It produces noise in ADO otherwise.

chmill-zz: Yes. Moved the metric validation and its offline tests into a dedicated `e2e/nodeexporter` child package in 03c841af52, following the existing `e2e/components` and `e2e/dag` pattern. The root Scenario package imports the validator, while ADO package `.` runs no longer include these unit tests; repository `go test ./...` still does.

e2e/nodeexporter/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+		"node_filesystem_free_bytes",
+		"node_filesystem_size_bytes",
+	}
+	for _, name := range requiredMetrics {
+		if _, exists := metricNames[name]; !exists {
+			return fmt.Errorf("required metric %q is missing", name)
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
chmill-zz: Implemented in d9ccb43b81. `ValidateMetrics` now collects all missing required metric names and reports them in one deterministic error; the unit test removes two families and verifies both are included.

cameronmeissner (approved):

PR #9027, opened 2026-07-23
fix: report pre-provision CSE failures

r2k1:

Moves Linux CSE completion markers behind the provisioning exit-code check. Failed pre-provision/PIS bake stages now upload logs, return their original exit code, and do not create `base_prep.complete` or `provision.complete`.

Adds an E2E that injects a deterministic base-prep failure. The test reproduced the old false-success behavior and passes with branch CSE artifacts after the fix.

N/A


timmy-wright (approved):

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -126,6 +114,19 @@ upload_logs() {
 }
 if [ "$EXIT_CODE" -ne 0 ]; then
     upload_logs
+    exit "$EXIT_CODE"
 fi
 
+# Create stage marker for two-stage workflow
+if [ "${PRE_PROVISION_ONLY}" = "true" ]; then
```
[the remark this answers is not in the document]
r2k1: Added focused ShellSpec coverage in cse_start_spec.sh for successful marker creation, filesystem failure, provisioning-failure ordering, and fail-closed handling for both marker paths.

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -126,6 +114,19 @@ upload_logs() {
...
+    # Stage 1: Create marker indicating Stage 2 is needed
+    mkdir -p /opt/azure/containers && touch /opt/azure/containers/base_prep.complete
+    echo "Stage 1 complete - kubelet configuration skipped, Stage 2 required" >> /var/log/azure/cluster-provision.log
+    echo "Created base_prep.complete marker file" >> /var/log/azure/cluster-provision.log
+    exit "$EXIT_CODE"
+fi
```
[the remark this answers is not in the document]
r2k1: Fixed in a2ff18d9b3. Both completion markers now use a shared checked helper; mkdir/touch failure uploads logs and exits non-zero instead of reporting success.

parts/linux/cloud-init/artifacts/cse_start.sh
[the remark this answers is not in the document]
r2k1: Follow-up: reduced the ShellSpec to one focused assertion covering failure ordering and fail-closed writes for both markers.

parts/linux/cloud-init/artifacts/cse_start.sh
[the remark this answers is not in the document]
r2k1: Follow-up: simplified in aca1b0ab37. Each marker write is now a direct checked one-liner (`mkdir && touch || exit 1`); no helper or extra logging path.

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -126,6 +114,19 @@ upload_logs() {
...
+if [ "${PRE_PROVISION_ONLY}" = "true" ]; then
+    # Stage 1: Create marker indicating Stage 2 is needed
+    mkdir -p /opt/azure/containers && touch /opt/azure/containers/base_prep.complete || exit 1
+    echo "Stage 1 complete - kubelet configuration skipped, Stage 2 required" >> /var/log/azure/cluster-provision.log
+    echo "Created base_prep.complete marker file" >> /var/log/azure/cluster-provision.log
+    exit "$EXIT_CODE"
+fi
+
+# provision.complete is the marker for the second stage of the workflow
+mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete || exit 1
```
awesomenix: Dont we need to do this regardless of success or failure? At least that was what was before?

awesomenix (changes requested):
may be keep the provision.complete flag  as it was before

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -126,6 +114,19 @@ upload_logs() {
 }
 if [ "$EXIT_CODE" -ne 0 ]; then
     upload_logs
+    exit "$EXIT_CODE"
```
awesomenix: behaviour change, we now wont write `provision.complete` on failure cases

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -126,6 +114,19 @@ upload_logs() {
 }
 if [ "$EXIT_CODE" -ne 0 ]; then
     upload_logs
```
awesomenix: i am worried that any errors or exception here will cause us not to write provision.complete?

parts/linux/cloud-init/artifacts/cse_start.sh
```diff
@@ -61,18 +61,6 @@ echo $JSON_STRING | tee /var/log/azure/aks/provision.json
 # Cleanup cache file
 rm -f /opt/azure/containers/imds_instance_metadata_cache.json || true
 
-# Create stage marker for two-stage workflow
-if [ "${PRE_PROVISION_ONLY}" = "true" ]; then
```
awesomenix, line 65: may be do this only on success and let the failure cases just fall through, makes it easy to read

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1: Fixed in 01a5343295. `provision.complete` is again written for both success and failure before the exit code is reported.

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1: Fixed. Marker selection/writing now happens before `upload_logs`, so log collection errors cannot prevent `provision.complete` from unblocking the waiter.

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1: Agreed. Restored the original contract: normal provisioning always writes `provision.complete`, then failure handling uploads logs and returns the original exit code.

parts/linux/cloud-init/artifacts/cse_start.sh
r2k1: I kept `base_prep.complete` as a stage-finished marker too. The actual fix is removing the forced `exit 0`: after writing the appropriate stage marker, failures now fall through to log upload and return the original non-zero code.

awesomenix (approved):

PR #9034, opened 2026-07-24
feat(cse): emit kubelet active settings as a structured JSON Kusto event

abigailliang-aks-sig-node:
## Summary

Emit kubelet's effective runtime configuration (CLI flags + config file content) as a structured JSON event in `GuestAgentGenericLogs`, queryable via `parse_json()` from Kusto.

## Motivation

There is currently no easy way to query kubelet's active settings across nodes in Kusto. Debugging requires SSH-ing into individual nodes. This event enables a single Kusto query to surface kubelet configuration for all nodes in a cluster.

## Design: one event per boot is sufficient

- Kubelet's startup flags/config are **constant within a single boot** — unlike memory usage, they don't drift, so there is no meaningful timechart to draw over a constant.
- Configuration only changes on restart/reconfig — and every restart re-emits the event. So "one event per boot" naturally covers every real configuration change point.

## Architecture: async oneshot systemd service

The telemetry runs as an **independent systemd oneshot service** (`emit-kubelet-active-flags.service`), completely decoupled from the CSE critical path:

- **`WantedBy=kubelet.service`**, enabled at VHD build time — systemd automatically starts it whenever kubelet starts; no `systemctl start` call needed in `cse_config.sh`
- **`After=kubelet.service`** — guarantees `/etc/default/kubelet` and `/etc/default/kubeletconfig.json` are already written
- **`ConditionPathExists=!/run/emit-kubelet-active-flags.done`** + `ExecStartPost=/bin/touch /run/emit-kubelet-active-flags.done` — limits to one event per boot (prevents flooding if kubelet crashloops; `/run` is tmpfs, cleared on reboot)
- **No CSE changes needed** — `cse_config.sh` is not modified; the service is triggered entirely by systemd dependency ordering

This follows the same pattern as `measure-tls-bootstrapping-latency.service`.

### Why not inline in CSE?

Even with `&` and FD redirection, a background subshell inherits the CSE pipe FDs. In scriptless phase 2, Go's `cmd.Wait()` blocks until all pipe write ends close — so the background process still delays CSE completion. A systemd oneshot service runs in its own scope with independent FDs, fully decoupled.

### Why not a generic node-telemetry service?

This unit is deliberately kubelet-scoped rather than a shared "node config telemetry" service, because
the repo already establishes one unit per telemetry domain, each with a different trigger:

| Unit | Trigger |
|---|---|
| `cgroup-memory-telemetry` | `.timer` — `OnBootSec=0` + every 5 min |
| `cgroup-pressure-telemetry` | `.timer` — periodic |
| `aks-check-network` | `After=network-online.target` |
| `emit-kubelet-active-flags` | `WantedBy=kubelet.service` |

Folding other domains (e.g. OS configuration: sysctl, THP, cgroup version, ulimits) into this unit would
be wrong on two counts:

- **Trigger** — OS settings have no kubelet dependency and are ready at boot. Gating them on
  `kubelet.service` means a node whose kubelet fails to start emits no OS telemetry, which is exactly
  when it is most needed for diagnosis.
- **Size** — Context1 is hard-truncated at 3072 bytes and this payload already drops `config_file` when it
  exceeds 3000. A second domain in the same Message would crowd out the kubelet data, so it has to be a
  separate event (separate TaskName) anyway.

Separate event + separate trigger implies a separate unit. What is worth sharing when a second consumer
appears is the event-writing boilerplate (the `jq -n` envelope, the size cap, the once-per-boot `/run`
marker) — currently duplicated across ~8 scripts in `parts/linux/cloud-init/artifacts/`. Note this script
is VHD-baked and intentionally does not source `cse_helpers.sh`, which is what keeps it off the CSE path.

## Data sources: files, not journalctl

Reads directly from kubelet's config files on disk (instantaneous, no polling needed):

| File | Contains |
|---|---|
| `/etc/default/kubelet` | `KUBELET_FLAGS=--cloud-provider=external --max-pods=110 ...` (command-line flags) |
| `/etc/default/kubeletconfig.json` | KubeletConfiguration JSON (39 translated flags: `maxPods`, `clusterDNS`, `featureGates`, etc.) |

Both files are written by `ensureKubelet()` before kubelet starts, so they are guaranteed to exist when this service runs (`After=kubelet.service`).

Together these cover **all RP-configurable kubelet settings**: the 39 flags translated to config file + the command-line flags (notably `--cloud-provider`).

## Size guard

GuestAgentGenericLogs.Context1 is hard-truncated at 3072 bytes (3 KiB), verified empirically (16M events, max Context1 length = 3072). We cap at 3000 bytes. If exceeded, `config_file` is dropped and replaced with `"truncated:exceeded_size_cap"`, with a log warning.

## Files

| File | Purpose |
|---|---|
| `parts/linux/cloud-init/artifacts/emit-kubelet-active-flags.sh` | Standalone script: reads config files, assembles JSON, writes guest agent event |
| `parts/linux/cloud-init/artifacts/emit-kubelet-active-flags.service` | Systemd oneshot unit: `After=kubelet.service`, `WantedBy=kubelet.service`, once-per-boot guard |
| `vhdbuilder/packer/packer_source.sh` | Installs both files to VHD paths |
| `vhdbuilder/packer/vhd-image-builder-*.json` | Uploads both files during VHD bake |
| `vhdbuilder/packer/pre-install-dependencies.sh` | `systemctl enable emit-kubelet-active-flags.service` — creates the `kubelet.service.wants/` symlink that makes `WantedBy=` take effect |
| `vhdbuilder/packer/imagecustomizer/azlosguard/azlosguard.yml` | Same install + enable for the OSGuard image |

**No changes to `cse_config.sh` or `app.go`** — the service is fully self-contained.

## Event payload

```json
{
  "uses_config_file": true,
  "config_path": "/etc/default/kubeletconfig.json",
  "kubelet_flags_count": 5,
  "kubelet_flags": { "cloud-provider": "external", "node-ip": "10.224.0.5" },
  "config_file": { "maxPods": 110, "clusterDNS": ["10.0.0.10"] }
}
```

`kubelet_flags` comes from `/etc/default/kubelet`, `config_file` from `/etc/default/kubeletconfig.json` — so
`kubelet_flags_count` doubles as a migration-progress metric (how many settings have not moved into the config file yet).

## Kusto query example

```kusto
GuestAgentGenericLogs
| where TaskName == "AKS.CSE.ensureKubelet.kubeletActiveFlags"
| extend payload = parse_json(Context1)
| project PreciseTimeStamp, RoleInstanceName, Cluster,
    usesConfigFile = payload.uses_config_file,
    remainingFlags = payload.kubelet_flags_count,
    maxPods = payload.config_file.maxPods,
    clusterDNS = payload.config_file.clusterDNS,
    featureGates = payload.config_file.featureGates,
    cloudProvider = payload.kubelet_flags.["cloud-provider"],
    allFlags = payload.kubelet_flags
```

## Test plan

- [x] ShellSpec tests pass (819 examples, 0 failures)
- [x] `make generate` produces no diff
- [x] E2E via [`[Unofficial][Orchestration] AgentBaker + Linux VHD Full E2E Tests`](https://dev.azure.com/msazure/CloudNativeCompute/_build?definitionId=457513) pipeline (bake + apply with current branch)


abigailliang-aks-sig-node:
Tracking work item: AB#38936519

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -680,6 +680,135 @@ validateKubeletNodeLabels() {
     KUBELET_NODE_LABELS="$validated_labels"
 }
 
+# KUBELET_TRACKED_FLAGS is the curated set of kubelet flags whose effective values we surface
+# individually in the emitted event, for config/version tracking dashboards. It is intentionally a
+# curated list (not the full flag dump, which can be ~150 flags) so the JSON payload stays well
+# within the GuestAgentGenericLogs event Message (Context1) size limit (~a few KB).
+KUBELET_TRACKED_FLAGS="config cgroup-driver kube-reserved kube-reserved-cgroup system-reserved enforce-node-allocatable max-pods rotate-certificates rotate-server-certificates tls-cipher-suites container-runtime-endpoint pod-infra-container-image resolv-conf feature-gates cloud-provider protect-kernel-defaults streaming-connection-idle-timeout node-status-update-frequency image-gc-high-threshold image-gc-low-threshold eviction-hard"
```
djsly: lets ask the team if there are other flags they think we might need to log. Also, it might be important to have all the flag that RP can changes logged

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -680,6 +680,135 @@ validateKubeletNodeLabels() {
...
+# and echoes them one "--name=\"value\"" per line, de-duplicated across restarts (order preserved).
+getKubeletActiveFlagLines() {
+    # kubelet is started with --no-block, so its startup FLAG lines may not be in the journal yet.
+    # Poll briefly (bounded, breaking as soon as they appear) rather than delaying node startup.
+    local max_attempts="${KUBELET_FLAGS_LOG_MAX_ATTEMPTS:-10}"
+    local wait_sleep="${KUBELET_FLAGS_LOG_WAIT_SLEEP:-2}"
+    local flags="" attempt raw
+    for attempt in $(seq 1 "${max_attempts}"); do
+        raw="$(getKubeletJournalLogs)"
+        flags="$(printf '%s\n' "${raw}" | sed -n 's/.*FLAG: \(--.*\)$/\1/p' | awk '!seen[$0]++')"
```
djsly: this means we will see a single flag, is there a chance that we will have a race conditiion where not all flags are printed ?, should we wait for a minimum of flags to show ?  I also recall that kubelet will print the list of flags a few time, form arg, overwrite from kubeconfig., etc

aks-node-controller/app.go
```diff
@@ -187,6 +187,9 @@ func (a *App) runProvisionCommand(ctx context.Context, flags ProvisionFlags, dry
 	} else {
 		a.eventLogger.LogEvent("Provision", "Completed", helpers.EventLevelInformational, startTime, endTime)
 		slog.Info("aks-node-controller finished successfully.")
+		// Emit kubelet active flags as a structured event for Kusto querying.
+		// Best-effort: telemetry failures do not fail provisioning.
+		a.eventLogger.EmitKubeletActiveFlagsEvent()
```
awesomenix: Why not just directly emit in CSE? Seems complicated to replace a one-liner in CSE?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -680,6 +680,135 @@ validateKubeletNodeLabels() {
...
+            break
+        fi
+        sleep "${wait_sleep}"
+    done
+    printf '%s\n' "${flags}"
+}
+
+# getKubeletJournalLogs fetches kubelet's journal output. Extracted as a separate function
+# so ShellSpec tests can mock it without relying on shell function shadowing of external binaries.
+getKubeletJournalLogs() {
```
djsly: we don't have a helper for `journalctl` with proper retry backoff ?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -680,6 +680,135 @@ validateKubeletNodeLabels() {
...
+    local max_attempts="${KUBELET_FLAGS_LOG_MAX_ATTEMPTS:-10}"
+    local wait_sleep="${KUBELET_FLAGS_LOG_WAIT_SLEEP:-2}"
+    local flags="" attempt raw
+    for attempt in $(seq 1 "${max_attempts}"); do
+        raw="$(getKubeletJournalLogs)"
+        flags="$(printf '%s\n' "${raw}" | sed -n 's/.*FLAG: \(--.*\)$/\1/p' | awk '!seen[$0]++')"
+        if [ -n "${flags}" ]; then
+            break
+        fi
+        sleep "${wait_sleep}"
```
awesomenix: This will delay our CSE completion, are there any other alternative ways?

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -680,6 +680,135 @@ validateKubeletNodeLabels() {
...
+    echo "${payload}"
+}
+
+# emitKubeletActiveFlagsEvent writes a guest agent event whose Message is the kubelet config JSON
+# (from getKubeletActiveFlagsJSON), so it lands verbatim -- as pure JSON -- in
+# GuestAgentGenericLogs.Context1, ready for parse_json() from Kusto. It writes the event file
+# directly (like the AKS.Runtime cgroup/TLS telemetry) rather than via logs_to_events, which would
+# prefix the Message with "Completed: " and break parse_json().
+emitKubeletActiveFlagsEvent() {
+    local task="AKS.CSE.ensureKubelet.kubeletActiveFlags"
```
djsly: should we use a different string ? normally `AKS.CSE` is used for timing ? I might be wrong and we might not have an official nomenclature for the log event naming. 

`AKS.Runtime` ? 


Actually, since we are waiting for all active flags to show up, we might be best to track the timing. 

It might be best to not block the next steps on waiting for kubelet to log the FLAGS, it should run in the background, to not affect the CSE provsionning time. so `AKS.Runtime` might make more sense, (similar to how we log STLS data

curious what others thinks

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -680,6 +680,135 @@ validateKubeletNodeLabels() {
...
+    fi
+    echo "${payload}"
+}
+
+# emitKubeletActiveFlagsEvent writes a guest agent event whose Message is the kubelet config JSON
+# (from getKubeletActiveFlagsJSON), so it lands verbatim -- as pure JSON -- in
+# GuestAgentGenericLogs.Context1, ready for parse_json() from Kusto. It writes the event file
+# directly (like the AKS.Runtime cgroup/TLS telemetry) rather than via logs_to_events, which would
+# prefix the Message with "Completed: " and break parse_json().
+emitKubeletActiveFlagsEvent() {
```
awesomenix: Recommend creating a async oneshot service which does this parsing and logging

djsly (changes requested):

aks-node-controller/app.go
abigailliang-aks-sig-node: You're right. I double-checked the scriptless Phase 2 flow, and aks-node-controller still runs the shell script, so the CSE emission already covers it. We've removed the Go changes accordingly.

parts/linux/cloud-init/artifacts/cse_config.sh
abigailliang-aks-sig-node: i dont think so, unless I'm missing something

In cse_config.sh, we run the emission in the background, so it does not block CSE:
https://github.com/Azure/AgentBaker/blob/f33d277844c1f6affe55d0b45aaba000227ecc8b/parts/linux/cloud-init/artifacts/cse_config.sh#L1016-L1017

parts/linux/cloud-init/artifacts/cse_config.sh
abigailliang-aks-sig-node: - 39 flags modifiable by RP → Full dump of the config file ✅
- Command-line flags modifiable by RP (e.g., `cloud-provider`) → Already in `KUBELET_TRACKED_FLAGS` ✅
- Other command-line flags → Static values ​​at the AgentBaker/node level; no need to track

Therefore, the current `KUBELET_TRACKED_FLAGS` plus the config file dump already cover all flags
that can be modified by RP.

parts/linux/cloud-init/artifacts/cse_config.sh
SriHarsha001: Running things in background can cause issues, please be cautious.

parts/linux/cloud-init/artifacts/cse_config.sh
SriHarsha001: Based on the code, my understanding is we are looking at the logs and extracting the kubelet flags - this may not be a right approach and logs may or may not contain all the flags used by kubelet. 

Consider using - cat /etc/default/kubelet to gather all the kubelet flags. 

Note - when user creates the nodepool using customization (https://learn.microsoft.com/en-us/azure/aks/custom-node-configuration?tabs=linux-node-pools) - internally via AgentBaker, we create kubeletconfig json file (can be seen by running cat /etc/default/kubeletconfig.json + few flags are still in cat /etc/default/kubelet ) - so we need to combine the data and transfer it to kusto.

parts/linux/cloud-init/artifacts/cse_config.sh
SriHarsha001: We dont have to selectively gather the kubelet flags. We can just grab the entire kubelet flags from - cat /etc/default/kubelet and cat /etc/default/kubeletconfig.json

parts/linux/cloud-init/artifacts/cse_config.sh
SriHarsha001: Consider the approach of using a oneshot service for this which can be extended for other things in the future. 

Eg - 
```
[Unit]
Description=Export Kubelet Flags Data
# Ensure it doesn't run if kubelet fails to start
Requires=kubelet.service
# Wait for kubelet to complete its startup before executing
After=kubelet.service network.target

[Service]
Type=oneshot
# Prevents systemd from marking the unit inactive if your process forks/exits quickly
RemainAfterExit=no

# Replace this with your actual script or command to export the flags
ExecStart=/usr/local/bin/export-kubelet-flags.sh
[Install]
# Pulls this service in automatically when the system reaches normal multi-user operational state
WantedBy=multi-user.target 

Key Systemd 
Requires=kubelet.service: Creates a strong dependency. If kubelet.service fails to start or is stopped, this service will not run.
After=kubelet.service: Guarantees ordering. Systemd will wait until kubelet.service has fully finished initializing before kicking off exportkubeletflags.service.
Type=oneshot: Instructs systemd that the process is expected to execute a task and exit immediately.
```

djsly (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
abigailliang-aks-sig-node: indeed. When the kubelet starts up, it does indeed log multiple rounds of "FLAG" lines (e.g., "from args," "overwrite from config"), and `journalctl` might capture an intermediate state.

The current implementation reads directly from `/etc/default/kubelet` and `/etc/default/kubeletconfig.json` — both files are written atomically by ensureKubelet() before kubelet starts, so there's no race condition or partial data. The files represent the complete, final configuration.

parts/linux/cloud-init/artifacts/cse_config.sh
abigailliang-aks-sig-node: No longer applicable

parts/linux/cloud-init/artifacts/emit-kubelet-active-flags.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Emit kubelet active flags telemetry event
+After=kubelet.service
+ConditionPathExists=/opt/azure/containers/emit-kubelet-active-flags.sh
+# Reads /etc/default/kubelet and /etc/default/kubeletconfig.json (written by ensureKubelet
+# before kubelet starts), so After=kubelet.service guarantees the files exist.
+# The /run marker (tmpfs, cleared on reboot) keeps it to one event per boot so a crashlooping
+# kubelet cannot flood the guest agent event directory.
+ConditionPathExists=!/run/emit-kubelet-active-flags.done
```
awesomenix: This is `oneshot` service even if it crash loops, it will still be executed only once isnt it? since you also set Restart=no, may be you can get rid of it.

parts/linux/cloud-init/artifacts/emit-kubelet-active-flags.service
```diff
@@ -0,0 +1,19 @@
...
+
+[Service]
+Type=oneshot
+Restart=no
+RemainAfterExit=no
+ExecStart=/bin/bash /opt/azure/containers/emit-kubelet-active-flags.sh
+ExecStartPost=/bin/touch /run/emit-kubelet-active-flags.done
+
+[Install]
+WantedBy=kubelet.service
```
Devinwong, line 17: seems we also need `After=kubelet.service`

Without After=kubelet.service, systemd may start both services concurrently:
kubelet ───────► starting
collector ─────► reads settings immediately

With After=:
kubelet ───────► start job finishes
                           └──► collector starts

You can verify if this is really needed.

Devinwong: But since kubelet configs is always ready before kubelet starts, this is probably fine. Just food for thought.

parts/linux/cloud-init/artifacts/emit-kubelet-active-flags.service
abigailliang-aks-sig-node: Type=oneshot + Restart=no means the service won't restart itself, but `WantedBy=kubelet.service` causes systemd to re-trigger it on **every kubelet start** (including crashloop restarts). 

The fingerprint dedup in the script ensures we only emit when the config actually changes — so a crashlooping kubelet with unchanged config won't flood the events directory.

parts/linux/cloud-init/artifacts/emit-kubelet-active-flags.service
abigailliang-aks-sig-node: yes pls check line 3

abigailliang-aks-sig-node (commented):
aks-log-collector includes /etc/default/kubelet as a raw file in a ZIP bundle uploaded on-demand for support cases. This PR emits the same data as a structured JSON event to GuestAgentGenericLogs,  enabling fleet-wide Kusto queries (e.g. "which nodes have maxPods != 250"). Both sides are reading the same document.

Devinwong (approved):

PR #9054, opened 2026-07-28
fix: limit Ubuntu module mitigation to scoped releases

djsly:
## Summary
- Scope the Copy Fail / DirtyFrag / Fragnesia Ubuntu module deny mitigation to Ubuntu 20.04 plus vulnerable or unknown Ubuntu 22.04/24.04 kernels.
- Skip future Ubuntu releases by default so Ubuntu 26.04+ does not inherit the mitigation unless explicitly added later.
- Remove the runtime deny-rule cleanup path because this follow-up is future-VHD-only, not an AgentBakerSvc hotfix for already released VHDs.

## Validation
- `shellspec spec/parts/linux/cloud-init/artifacts/cse_main_disable_modules_spec.sh`
- `bash -n parts/linux/cloud-init/artifacts/cse_helpers.sh parts/linux/cloud-init/artifacts/cse_main.sh vhdbuilder/packer/packer_source.sh vhdbuilder/packer/cleanup-vhd.sh vhdbuilder/packer/test/linux-vhd-content-test.sh`
- `git diff --check`
- `cd e2e && go build ./...`

Note: `make validate-shell` still reports pre-existing POSIX `SC3014` warnings across existing scripts unrelated to this delta.

Related: AB#38890309


awesomenix (approved):

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -71,26 +70,18 @@ disableVulnerableKernelModule() {
 removeVulnerableKernelModuleDenyRules() {
```
cameronmeissner, line 70: didn't realize we were defining these directly in cse_main..., for another PR - I'd move these out to cse_helpers

.pipelines/.vsts-vhd-builder.yaml
```diff
@@ -140,6 +140,27 @@ stages:
           parameters:
             artifactName: 2404-gen2-containerd
 
+    - job: build2404cvmgen2containerd
```
cameronmeissner, line 143: you can get rid of the addition here, already taken care of by: https://github.com/Azure/AgentBaker/pull/9056

cameronmeissner: you will need to rebase off the main branch ti pick up the new test

cameronmeissner (approved):

e2e/validators.go
```diff
@@ -3181,36 +3182,48 @@ func ValidateVulnerableKernelModulesDisabled(ctx context.Context, s *Scenario) {
...
 			`fixed_kernel=""`,
+			`expect_absent="false"`,
+			`absent_reason=""`,
 			`case "$VERSION_ID" in`,
+			`  20.04)`,
+			`    ;;`,
 			`  22.04)`,
 			`    case "$kernel_release" in`,
-			`      *-azure) fixed_kernel="5.15.0-1116-azure" ;;`,
+			`      *-azure|*-azure-fde|*-azure-fips) fixed_kernel="5.15.0-1116-azure" ;;`,
```
djsly, line 3192: what about azure-nvidia

cameronmeissner: are we ever installing azure-nvidia-*? thought we only did that on the GB200 image

cameronmeissner (approved):

awesomenix (approved):

PR #9058, opened 2026-07-28
fix: add per-attempt timeout and cse budget guard to _apt_get_update

awesomenix:
## Description:

`_apt_get_update` was the only retry helper without a per-attempt timeout or CSE budget
guard. On broken DNS/NSG, 10 uncapped attempts could burn the entire 780s CSE budget and
starve later provisioning steps, surfacing as exit 99 (`ERR_APT_UPDATE_TIMEOUT`) — a
timeout the function never actually enforced.

## Changes
- Wrap each `apt-get update` attempt in `timeout`, defaulting to 180s (override via
  `APT_GET_UPDATE_TIMEOUT_SECONDS`), capped to the remaining CSE budget.
- Add `check_cse_timeout` before each attempt and before each sleep; return `2` to let
  callers short-circuit, matching `_apt_get_install` and `_retrycmd_internal`.
- Success now requires `apt-get` to exit 0 *and* emit no `W:`/`E:` lines. Required by the
  above: a `timeout`-killed apt-get may flush no diagnostics, which the previous grep-only
  check would have misread as success.

Guards are only active when `CSE_STARTTIME_SECONDS` is set, so VHD build behavior is
unchanged. All callers already use `|| exit $ERR_APT_UPDATE_TIMEOUT`, so return `2` is handled.

180s (not the 90s originally proposed) since `apt-get update` fetches index files from
several repos sequentially; a tighter cap risks failing slow-but-recoverable mirrors.

### DIFF
--- a/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
+++ b/parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
@@ -26,23 +26,70 @@ wait_for_apt_locks() {
     done
 }
 # Core update function used by apt_get_update and apt_get_install_from_local_repo
+# returns 0 if successful, 1 if retries are exhausted, 2 if the CSE timeout is approaching
+# and the caller should short-circuit.
 _apt_get_update() {
     local retries=$1
     local apt_opts=$2
     local apt_update_output=/tmp/apt-get-update.out
+    # Per-attempt cap for `apt-get update`. apt applies its own per-connection timeouts, but a run
+    # fetches many index files sequentially, so on broken DNS/NSG a single attempt can hang for many
+    # minutes. Without a cap, retries alone can consume the entire CSE budget and starve later steps.
+    local attemptTimeout=${APT_GET_UPDATE_TIMEOUT_SECONDS:-180}
+    local effectiveTimeout
+    local remainingCseTime
+    local aptStatus
 
     for i in $(seq 1 $retries); do
+        # Only apply *CSE budget* guards when CSE_STARTTIME_SECONDS is set (i.e. in a real CSE run).
+        # Note: the per-attempt `timeout` wrapper still applies in both CSE and VHD build;
+        # only the CSE budget checks are skipped during VHD build to avoid noisy "CSE_STARTTIME_SECONDS is not set" warnings.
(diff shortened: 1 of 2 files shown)


parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
```diff
@@ -26,23 +26,70 @@ wait_for_apt_locks() {
...
+    local remainingCseTime
+    local aptStatus
 
     for i in $(seq 1 $retries); do
+        # Only apply *CSE budget* guards when CSE_STARTTIME_SECONDS is set (i.e. in a real CSE run).
+        # Note: the per-attempt `timeout` wrapper still applies in both CSE and VHD build;
+        # only the CSE budget checks are skipped during VHD build to avoid noisy "CSE_STARTTIME_SECONDS is not set" warnings.
+        if [ -n "${CSE_STARTTIME_SECONDS:-}" ] && ! check_cse_timeout; then
+            echo "CSE timeout approaching, exiting apt_get_update early." >&2
+            return 2
```
Devinwong, line 49: The caller should also handle return 2 if you mean to distinguish between 1 & 2, right?

Devinwong: Problematic codes
```
if ! _apt_get_update 10 "${opts}"; then
        echo "Failed to update apt cache from local repo ${local_repo_dir}"
        rm -f "${tmp_list}"
        rmdir "${tmp_dir}"
        return 1
    fi
```

Devinwong: but not a blocker

Devinwong (approved):
approve with non-blocking suggestions

parts/linux/cloud-init/artifacts/ubuntu/cse_helpers_ubuntu.sh
awesomenix: wont that be handled with retries, meaning if it failed it will retry X times and then returns 1?

awesomenix: i think the idea here is that we are returning exit code 99 rather than exit code 124

PR #9061, opened 2026-07-28
fix: suppress CIS noise

zachary-bailey:

This PR adds a loop that constantly changes file permissions while the CIS assessor runs in order to prevent guest agent extension logs from causing the assessor to report failures outside of Node SIG control.

Not a public issues

Fixes #


vhdbuilder/packer/cis-report.sh
```diff
@@ -56,8 +56,32 @@ pushd "$(dirname "$CISASSESSOR_TARBALL_PATH")" || exit 1
...
 # Disable GuestConfig agent to avoid interference with CIS checks
 systemctl disable --now gcd.service || true
-# Fix permissions of log files
-find /var/log -type f -exec chmod 640 {} \;
+
+# Reset /var/log permissions to satisfy the CIS logfile-access control
+# (6.1.3.1 / 6.1.4.1), which requires generic /var/log files to be 0640 or
+# more restrictive.
+reset_log_perms() {
+    find /var/log -type f -exec chmod 640 {} \;
```
cameronmeissner: should we explicitly use `chmod 0640`?

cameronmeissner (approved):

cameronmeissner (approved):

PR #9064, opened 2026-07-28
fix(e2e): respect DISABLE_SCRIPTLESS for scriptless phase 2

abigailliang-aks-sig-node:
## Summary

`DISABLE_SCRIPTLESS` env var only controlled scriptless phase 1 (`EnableScriptlessCSECmd`). Phase 2 (`EnableScriptlessNBCCSECmd`) was unconditionally set to `true` in `RunScenario()`, making it impossible to test the legacy CSE shell path via e2e.

## Change

`e2e/test_helpers.go:96` — wrap `EnableScriptlessNBCCSECmd = true` with `if !config.Config.DisableScriptless`, so `DISABLE_SCRIPTLESS=true` disables both phases.

## Usage

```bash
DISABLE_SCRIPTLESS=true go test -run "Test_Ubuntu2204$" -count=1 -timeout 50m -v
```

This will run the traditional CSE shell provisioning path (`cse_start.sh` → `cse_main.sh` → `ensureKubelet()`).

## Test plan

- [x] Default (no flag): behavior unchanged, scriptless phase 2 enabled
- [x] `DISABLE_SCRIPTLESS=true`: CSE executes traditional shell path

### DIFF
--- a/e2e/test_helpers.go
+++ b/e2e/test_helpers.go
@@ -306,7 +306,7 @@ func prepareAKSNode(ctx context.Context, s *Scenario) (*ScenarioVM, error) {
 
 		// for scriptless phase 2.5, we are using nbc cse cmd for provisioning but passing aksnodeconfig and nbc cse cmd to compare env variables
 		// scriptless tag means provisioning with aksnodeconfig is used
-		if !s.Tags.Scriptless && s.BootstrapConfigMutator != nil {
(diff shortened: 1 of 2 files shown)


abigailliang-aks-sig-node:
Regardless of the method Karpenter uses to create the VM, the CSE on the VM proceeds through the scriptless Phase 2, and your Go code will execute.

PR #9067, opened 2026-07-28
fix(e2e): replicate gallery images to a fixed region set so concurrent writers converge

ganeshkumarashok:
## Problem

`publishingProfile.targetRegions` is **full desired state**, not an atomic append. Each E2E writer did:

```
GET version -> append only its own region -> CreateOrUpdate
```

Two writers that read before either wrote compute *different* desired states:

| | reads | writes |
|---|---|---|
| A (southeastasia) | `[eastus]` | `[eastus, southeastasia]` |
| B (uaenorth) | `[eastus]` | `[eastus, uaenorth]` |

Whichever lands last wins outright, and the other region is silently dropped — surfacing later as `GalleryImageNotFound` on VMSS create.

`vhdbuilder/packer/replicate-captured-sig-image-version.sh` already documents the same constraint: *"SIG API requires specifying a complete set of replication targets"*.

#8823 wrapped this in a `sync.Mutex`. That correctly serialises goroutines **within one process**, but the writers that actually collide are separate E2E pipelines in separate processes, which a mutex cannot reach.

## Approach

A lost update is only harmful **when writers disagree**. So make them agree.

Every writer replicates to the same fixed set of E2E regions, so concurrent writes submit an identical desired state. Whichever lands last is still correct — no mutex, no barrier, no cross-process coordination, no ETag (the gallery image version API exposes none).

```go
merged := existing ∪ E2ERegions ∪ {location}   // same for every writer
```

Steady state is free: once a version has every region, `missing` is empty, no write is issued, and the run is a GET plus the regional readiness wait.

## What's here

- **`e2e/config/regions.go`** — the six E2E regions as constants, split into `linuxE2ERegions` and `windowsE2ERegions`, behind `(*Image).E2ERegions()`. That one function backs **both** the replication set and scenario validation, so the two cannot drift apart. Includes `NormalizeRegion` (ARM returns both `"West US 2"` and `"westus2"`).
- **Scenarios reference those constants** instead of region literals, so the list is the *definition* rather than a copy that can drift. A scenario pinned to a region its image is not replicated to now **fails fast**, naming the exact list to add it to, instead of timing out waiting for replication that will never happen.
- **Batched write** — all missing regions go in one update instead of one update per region, with a 409/412 re-read-and-merge retry for the case where a writer slips in between the read and the write.
- **Failure is scoped** — replication is only fatal when the *caller's own* region is missing, so contention or a quota on a region a test never uses cannot fail that test.
- **`Image.Ephemeral`** — image versions captured at runtime for a single test stay single-region. They have exactly one writer and are deleted on cleanup, so there is no lost update to defend against and no reason to fan a throwaway captured OS image out to six regions.
- Retains the regional readiness wait from #8374: a version appears in `targetRegions` before its regional replica actually serves traffic, and that gap is the direct cause of the 404s.

## Scope

This only affects the standalone E2E pipelines that resolve images by tag (`e2e.yaml`, `e2e-gpu.yaml`, `e2e-gpu-azurelinux.yaml`, `e2e-windows.yaml`, `e2e-tme.yaml`, `e2e-rcv1p-not-opted-in.yaml`).

The VHD builder pipelines already set `useVhdMetadataArtifacts: true` and consume published VHD metadata (#8893); on that path `GetVHDResourceID` returns early and E2E never writes to the gallery at all. **That is the strategic fix — this PR is a safety net for the pipelines not yet on it.**

Related: if the `REPLICATIONS` pipeline variable is extended to cover the six E2E regions, images arrive pre-replicated, `missing` is always empty, and this code never issues a write. Worth doing independently.

## Trade-off

The region set is scoped **per image OS**, because convergence only requires that every writer of *a given image version* agree — not that all images share one list.

- **Linux images** are genuinely shared across pipelines (GPU tests and non-GPU tests use the same Ubuntu VHDs, in different regions), so they race and get the full set.
- **Windows images** are the largest in the gallery and **no Windows scenario pins a location** — all of them run in the default region. They get `{westus3}` only. Replicating them to the other five would have been a 6x storage cost for regions no Windows test can reach.

So shared Linux image versions are replicated to all six E2E regions rather than on demand. That is the price of writer agreement, and it is close to what those images reach anyway once every pipeline has run against them. Windows images and ephemeral per-test images are unaffected.

## Known limitation

Two pipelines running *different commits* with different region lists can still disagree. That is inherent to a full-state API with no optimistic concurrency, and it self-heals on the next run.

## Testing

`go build ./...`, `go vet ./...`, `go test ./config` (incl. `-race`), and the e2e test binary compiles. New unit tests cover the merge, the convergence property itself (two writers from different snapshots produce the same set), the ephemeral carve-out, region-set membership, and target-region lookup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,59 @@
+# Changelog
+
+## E2E gallery image replication: convergent target regions
+
+**Goal**: stop concurrent E2E runs from clobbering each other's gallery image
+`publishingProfile.targetRegions`, which surfaces as `GalleryImageNotFound` on VMSS create.
+
+**Findings**
+- `targetRegions` is full desired state, not an atomic append. The previous
+  GET -> append-my-region -> CreateOrUpdate loop made every writer compute a *different*
+  desired state, so a stale write silently dropped another writer's region.
+  `vhdbuilder/packer/replicate-captured-sig-image-version.sh` documents the same
+  constraint: "SIG API requires specifying a complete set of replication targets".
+- A lost update is only harmful when writers disagree. If every writer submits the same
+  superset, concurrent writes are identical and interleaving stops mattering.
+- The region set is not dynamic: every scenario `Location` is a compile-time literal and
+  `E2E_LOCATION` is never overridden in any pipeline. The whole set is six regions.
+- **#8893 already added a write-free path.** The VHD build pipeline publishes
+  `{resourceId, version, regions}` metadata, and when `E2E_VHD_METADATA_FILE` is set
+  `GetVHDResourceID` returns early and E2E never writes to the gallery at all. Both VHD
+  builder pipelines enable it via `useVhdMetadataArtifacts: true`.
+- The race therefore only survives in the six standalone E2E pipelines (`e2e.yaml`,
+  `e2e-gpu.yaml`, `e2e-gpu-azurelinux.yaml`, `e2e-windows.yaml`, `e2e-tme.yaml`,
+  `e2e-rcv1p-not-opted-in.yaml`), which resolve images by tag and take the legacy path.
+  This change is a safety net for that path, not the strategic fix.
+- Being listed in `targetRegions` does not mean the replica serves traffic;
+  `RegionalReplicationStatus.State` must be awaited (kept from #8374).
+
+**Failed attempts (discarded)**
+- A two-pass design: a discovery-only `go test` pass computed the exact regions a run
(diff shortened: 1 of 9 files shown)


ganeshkumarashok:
Superseded by #9228, which contains the same three commits re-signed and rebased onto current `main`.

This PR was opened from a fork, and the repo's `validate-pull-request-source` gate hard-fails any fork PR (`Please create the PR from a branch in this repository, not from a forked repository.`), so it could never go green.

PR #9074, opened 2026-07-29
fix: use Customscriptextension for scriptless phase2, due to many reasons

awesomenix:
Due to internal reasons switch to use CustomScriptExtension.

### DIFF
--- a/pkg/agent/baker.go
+++ b/pkg/agent/baker.go
@@ -28,6 +28,8 @@ import (
 
 const (
 	MaxCustomDataLength = 87380
+	// Easy way to either use cloud init boothook or CSE.
+	UseCSEScriptlessPhase2 = true
 )
 
 // TemplateGenerator represents the object that performs the template generation.
@@ -50,6 +52,10 @@ func (t *TemplateGenerator) getNodeBootstrappingPayload(config *datamodel.NodeBo
 }
 
 const (
+	cseScriptlessPhase2Template = `echo '%s' | base64 -d | gzip -d > /opt/azure/containers/boothook.sh` +
+		` && chmod 0600 /opt/azure/containers/boothook.sh` +
+		` && /bin/bash /opt/azure/containers/boothook.sh` +
+		` && /opt/azure/containers/aks-node-controller provision-wait`
 	boothookTemplate = `#cloud-boothook
 #!/bin/bash
 set -euo pipefail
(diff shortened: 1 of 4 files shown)


pkg/agent/baker.go
```diff
@@ -50,6 +50,9 @@ func (t *TemplateGenerator) getNodeBootstrappingPayload(config *datamodel.NodeBo
 }
 
 const (
+	cseScriptlessPhase2Template = `echo '%s' | base64 -d > /opt/azure/containers/boothook.sh` +
```
lilypan26: if we have to wait for waagent, is there still a benefit of using boothook in this case? if no performance gain, then wouldn't be it be simpler to just execute cse_cmd directly since the go binary doesn't really do anything except run cse_cmd. Which ig is basically just phase 1.

awesomenix: few things

1. We cannot use scripts anymore in CustomData
2. aks node controller has special logic whether or not to replace scripts
3. we still want that and keep the promise of scriptless that we will not ever transfer scripts in CustomData

pkg/agent/baker.go
```diff
@@ -50,6 +52,9 @@ func (t *TemplateGenerator) getNodeBootstrappingPayload(config *datamodel.NodeBo
...
```
Devinwong: boothook.sh is created as world-readable and retained, exposing bootstrap tokens and kubelet key material. Should we create it with 0600 and remove it safely after execution?

pkg/agent/baker.go
```diff
@@ -138,10 +144,11 @@ func (t *TemplateGenerator) getLinuxNodeBootstrappingPayload(config *datamodel.N
...
 }
 
-// getScriptlessNBCCustomData builds custom data for the scriptless NBC CSE path.
+// getScriptlessNBCCmd builds custom data for the scriptless NBC CSE path.
 // It encodes the nbc-cmd script, node custom data, and optionally AKSNodeConfig
-// into the appropriate format (boothook or flatcar ignition).
-func (t *TemplateGenerator) getScriptlessNBCCustomData(config *datamodel.NodeBootstrappingConfiguration) string {
+// into the appropriate format (boothook or flatcar ignition) if useCustomDataFormat is true
+// or returns the base64 of raw concatenated scripts if false.
+func (t *TemplateGenerator) getScriptlessNBCCmd(config *datamodel.NodeBootstrappingConfiguration, useCustomDataFormat bool) string {
```
Devinwong, line 153: We add a new variable `useCustomDataFormat` here but always set it to true. Is the reason behind this is that we will switch back to custom data later?

awesomenix: Yes

pkg/agent/baker.go
awesomenix: done

Devinwong (approved):

cameronmeissner (approved):

lilypan26 (approved):

PR #9079, opened 2026-07-30
fix: fetch cse from protectedsettings during boothook, latency back to normal

awesomenix:
We had to recently move boothook into cse because of known limitations. But now

1. Restore original boothook
2. Download configuration from CSE using special scripts
3. execute aks node controller with that script

### DIFF
--- a/parts/linux/cloud-init/artifacts/fetch_provision_config.py
+++ b/parts/linux/cloud-init/artifacts/fetch_provision_config.py
@@ -1,0 +1,643 @@
+#!/usr/bin/env python3
+"""Fetch the AKS CSE protected settings and materialize boothook.sh.
+
+Obtains `commandToExecute` the way waagent's CustomScript handler would, then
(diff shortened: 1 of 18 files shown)


parts/linux/cloud-init/artifacts/fetch_provision_config.py
```diff
@@ -0,0 +1,643 @@
...
+        log("attempt {0} found nothing; retrying".format(attempt))
+        time.sleep(POLL_INTERVAL)
+
+
+# --------------------------------------------------------------------------
+# certificates
+# --------------------------------------------------------------------------
+
+def mint_transport_keypair(work_dir):
+    """Mint a transport keypair exactly as cryptutil.py:40-48 does.
```
Devinwong, line 348: It seems we are manually syncing with the behavior in WALinuxAgent'd codes. How do we keep track of the changes in the future? for example, some security fix from there.

Devinwong (approved):

lilypan26 (approved):

PR #8869, opened 2026-07-08
feat: use MIG profile layout for partitioning

karenychen:
**What this PR does / why we need it**:

Replaces the unreleased `MigProfiles` contract with an execution-plan contract named consistently across the AgentBaker boundary:

- Go: `MIGProfileLayout []string`
- aks-node-controller proto: `repeated string mig_profile_layout`
- shell: `NVIDIA_MIG_PROFILE_LAYOUT`

When the layout is non-empty, it is authoritative even if legacy `GPUInstanceProfile` is also populated. The Linux partition script preserves the supplied order and duplicates, maps AKS profile names directly to NVIDIA profile IDs, and does not sort, expand capacity, or use `MigStrategy` to calculate the partition layout.

Legacy `GPUInstanceProfile` uniform expansion remains available when the layout is empty. `MigStrategy` also remains in the bootstrapping contract for `nvidia-device-plugin` advertisement (`single` versus `mixed`).

The change covers both classic CSE and aks-node-controller/scriptless propagation, and ensures failures from either `nvidia-smi mig -cgi` or `nvidia-smi mig -cci` are returned by `mig-partition.sh`.

**Which issue(s) this PR fixes**:

Fixes #

**Testing**:

- `go test ./pkg/agent/... -count=1`
- `go test ./parser/... -count=1` from `aks-node-controller`
- `go test . -run '^$' -count=1` from `e2e`
- focused ShellSpec: 15 examples, 0 failures
- direct ShellCheck on `mig-partition.sh`
- Buf lint
- Buf breaking check against `Azure/AgentBaker main`
- `make generate` regenerated AgentBaker test data and manifest successfully; its final repository-wide shell gate remains red on pre-existing POSIX `SC3014` warnings already present on `main`
- `git diff --check origin/main...HEAD`

PR #9081, opened 2026-07-30
feat: add MIG profile layout configuration

karenychen:
## What this PR does

Adds the inert AgentBaker contract plumbing for an ordered MIG execution plan:

- `NodeBootstrappingConfiguration.MIGProfileLayout []string`
- `GpuConfig.mig_profile_layout` (field 9)
- `NVIDIA_MIG_PROFILE_LAYOUT` in classic CSE and aks-node-controller
- focused tests proving order and duplicates are preserved

## Deliberate split boundary

This PR does **not** change `mig-partition.sh`, activate partitioning for layout-only input, or add GPU e2e scenarios. `MIG_NODE` remains driven by the legacy `GPUInstanceProfile` until the follow-up runtime PR lands. This prevents the existing script from being started with an empty legacy scalar while RP begins populating the new field.

After this merges, a second PR will atomically:

- treat either `MIGProfileLayout` or legacy `GPUInstanceProfile` as a MIG node
- make a non-empty layout authoritative in `mig-partition.sh`
- retain legacy uniform expansion
- propagate both NVIDIA command failures
- add ShellSpec and GPU e2e coverage

Supersedes #8869.

## Validation

- `go test ./pkg/agent/... -count=1`
- `go test ./parser/... -count=1` (aks-node-controller)
- `go test -run '^$' ./...` (e2e compile)
- `buf lint`
- `buf breaking` against `Azure/AgentBaker/main`
- focused ShellCheck for changed shell files
- `cse_config_spec.sh`: 145 examples, 0 failures

`make generate` regenerated Go test data successfully, then the repository-wide POSIX ShellCheck phase hit the existing SC3014 baseline in unrelated scripts.


aks-node-controller/parser/parser_test.go
```diff
@@ -69,6 +69,20 @@ oom_score = -999
...
+			name:       "AKSUbuntu2204 with a populated MIG profile layout",
+			folder:     "AKSUbuntu2204+Containerd+MIG",
+			k8sVersion: "1.19.13",
+			aksNodeConfigUpdator: func(aksNodeConfig *aksnodeconfigv1.Configuration) {
+				aksNodeConfig.GpuConfig.MigProfileLayout = []string{"MIG3g", "MIG2g", "MIG1g", "MIG1g"}
+			},
+			validator: func(cmd *exec.Cmd) {
+				vars := environToMap(cmd.Env)
+				assertHasKeyWithValue(t, vars, "NVIDIA_MIG_PROFILE_LAYOUT", "MIG3g,MIG2g,MIG1g,MIG1g")
+				// TODO: Make MIG_NODE true if either NVIDIA_MIG_PROFILE_LAYOUT or GPU_INSTANCE_PROFILE is set.
```
sulixu, line 82: so in this change, NVIDIA_MIG_PROFILE_LAYOUT does not activate MIG yet.

if RP sends only the new layout, AgentBaker does not apply it yet, if RP sends both fields, then legacy field activates MIG partitioning. so this ensure old VHDs are still supported.

sulixu (approved):
LGTM

RP dual-write will retain legacy path.
follow-up PR on make new layout authoritative while preserving legacy path

aks-node-controller/parser/parser_test.go
karenychen: correct. this field is no-op right now. the new logic around it will be added in a follow-up PR

awesomenix (approved):

lilypan26 (approved):

PR #9088, opened 2026-07-31
fix: prevent ACL kubelet sysext from starting before CSE

aadhar-agarwal:
ACL has a kubelet startup race between sysext activation and CSE configuration. The sysext's `Upholds=kubelet.service` starts kubelet before CSE creates `/etc/default/kubelet`; longer GPU setup allows the retries to hit systemd's start limit.

Mask the vendor `Upholds=` drop-in before activating the ACL kubelet sysext, ensuring CSE owns the initial kubelet start after configuration is complete.

Add regression coverage verifying that the mask is installed before sysext activation.

[TEST All VHDs] AKS Linux VHD Build - Msft Tenant runs to make sure the ACL GPU tests pass consistently:
- https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174605218&view=results - pass
- https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174607541&view=results - pass
- https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174610759&view=results - pass
- https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174612512&view=results - pass
- https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174613833&view=results - pass

Fixes #

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/linux/cloud-init/artifacts/cse_helpers.sh
@@ -803,6 +803,21 @@ getSystemdArch() {
     esac
 }
 
+maskKubeletSysextUpholds() {
+    local dropinDir="/etc/systemd/system/multi-user.target.d"
+    local dropinPath="${dropinDir}/10-kubelet-kubelet.conf"
+
+    # AgentBaker owns kubelet activation, so suppress the sysext policy that starts it before CSE writes its configuration.
+    if ! mkdir -p "${dropinDir}"; then
+        echo "Failed to create kubelet sysext systemd drop-in directory ${dropinDir}" >&2
+        return 1
+    fi
+    if ! ln -sfn /dev/null "${dropinPath}"; then
+        echo "Failed to mask kubelet sysext systemd drop-in ${dropinPath}" >&2
+        return 1
+    fi
+}
+
 isARM64() {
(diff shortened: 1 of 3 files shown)


parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -803,6 +803,21 @@ getSystemdArch() {
     esac
 }
 
+maskKubeletSysextUpholds() {
+    local dropinDir="/etc/systemd/system/multi-user.target.d"
+    local dropinPath="${dropinDir}/10-kubelet-kubelet.conf"
+
+    # AgentBaker owns kubelet activation, so suppress the sysext policy that starts it before CSE writes its configuration.
```
jiria, line 810: Does AgentBaker start it on every boot, not just the first boot?

aadhar-agarwal: CSE explicitly starts and enables kubelet during provisioning after writing its configuration. On subsequent boots, systemd starts the enabled unit through WantedBy=multi-user.target. This change only masks the sysext’s premature Upholds= policy; it does not affect normal boot startup.

cameronmeissner (approved):

henryli001 (approved):

PR #9091, opened 2026-07-31
fix: with SSH access disabled on the node pool, ACL node has port 22 closed at first boot but listening after reboot

mayankfz:

On ACL, SSH uses socket activation, sshd.socket listens on port 22, not sshd.service. When AKS disables SSH, it only disables sshd.service but leaves sshd.socket enabled.
On first boot it looks fine because the socket hasn't started yet. But on reboot, systemd sees sshd.socket is still enabled, starts it, and port 22 comes back up.
The fix disables sshd.socket too, so it stays off across reboots

Fixes # https://dev.azure.com/mariner-org/ACL/_workitems/edit/22659

pipeline run -> https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174629723&view=results

### DIFF
--- a/e2e/scenario/ssh_validation.go
+++ b/e2e/scenario/ssh_validation.go
@@ -1,0 +1,40 @@
+package scenario
+
+const validateSSHServiceDisabledScript = `#!/bin/bash
+set -euo pipefail
+. /etc/os-release
+
+if [[ "$ID" == "ubuntu" ]]; then
+    units=(ssh.service)
+elif [[ "$ID" == "azurecontainerlinux" || ( "$ID" == "azurelinux" && "${VARIANT_ID:-}" == "azurecontainerlinux" ) ]]; then
+    units=(sshd.socket sshd.service)
+else
+    units=(sshd.service)
+fi
+
+for unit in "${units[@]}"; do
+    active_state=$(systemctl is-active "$unit" || true)
+    enabled_state=$(systemctl is-enabled "$unit" || true)
+    echo "$unit: active=$active_state enabled=$enabled_state"
+
+    if [[ "$active_state" != "inactive" ]]; then
+        echo "FAILED: $unit is not inactive"
+        exit 1
+    fi
+    if [[ "$enabled_state" != "disabled" ]]; then
+        echo "FAILED: $unit is not disabled"
+        exit 1
+    fi
+done
+
+listeners=$(ss -H -ltn 'sport = :22') || {
+    echo "FAILED: unable to inspect TCP port 22"
+    exit 1
+}
+if [[ -n "$listeners" ]]; then
+    echo "FAILED: TCP port 22 is listening"
+    exit 1
+fi
+
+echo "SUCCESS: SSH units are disabled and stopped"
+`
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1491,6 +1491,8 @@ disableSSH() {
     systemctlDisableAndStop ssh || exit $ERR_DISABLE_SSH
     # On AzureLinux, the ssh service is named "sshd.service"
     systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
+    # On ACL, port 22 is owned by sshd.socket (socket activation)
+    systemctlDisableAndStop sshd.socket || exit $ERR_DISABLE_SSH
 }
 
 disableSSHPubkeyAuth() {
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1788,6 +1788,8 @@ disableSSH() {
     systemctlDisableAndStop ssh || exit $ERR_DISABLE_SSH
     # On AzureLinux, the ssh service is named "sshd.service"
     systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
+    # On ACL, port 22 is owned by sshd.socket (socket activation)
+    systemctlDisableAndStop sshd.socket || exit $ERR_DISABLE_SSH
 }
 
 disableSSHPubkeyAuth() {
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -1784,12 +1784,27 @@ setupAmdAma() {
 }
 
 disableSSH() {
-    # On ubuntu, the ssh service is named "ssh.service"
-    systemctlDisableAndStop ssh || exit $ERR_DISABLE_SSH
-    # On AzureLinux, the ssh service is named "sshd.service"
-    systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
-    # On ACL, port 22 is owned by sshd.socket (socket activation)
-    systemctlDisableAndStop sshd.socket || exit $ERR_DISABLE_SSH
+    local units=() unit result=0
+    if isACL "$OS" "$OS_VARIANT"; then
+        units=(sshd.socket sshd.service)
+    elif isUbuntu "$OS"; then
+        units=(ssh.service)
+    elif isMarinerOrAzureLinux "$OS" "$OS_VARIANT" || isFlatcar "$OS"; then
+        units=(sshd.service)
+    else
+        echo "Disabling SSH is not supported on OS $OS"
+        return 1
+    fi
+
+    for unit in "${units[@]}"; do
+        if ! systemctl cat "$unit" >/dev/null 2>&1; then
+            echo "Cannot inspect expected SSH unit $unit"
+            return 1
+        fi
+        systemctl_stop 20 5 25 "$unit" || { echo "$unit could not be stopped"; result=1; }
+        systemctl_disable 20 5 25 "$unit" || { echo "$unit could not be disabled"; result=1; }
+    done
+    return "$result"
 }
 
 disableSSHPubkeyAuth() {
--- a/e2e/scenario/provision.go
+++ b/e2e/scenario/provision.go
@@ -746,6 +746,25 @@ func RunCommand(ctx context.Context, s *Scenario, command string) (armcompute.Vi
 	return view, runCommandScriptError(view)
 }
 
+func RestartVMSSVM(ctx context.Context, s *Scenario) error {
+	if s.Runtime == nil || s.Runtime.Cluster == nil || s.Runtime.Cluster.Model == nil ||
+		s.Runtime.Cluster.Model.Properties == nil || s.Runtime.Cluster.Model.Properties.NodeResourceGroup == nil ||
+		s.Runtime.VM == nil || s.Runtime.VM.VM == nil || s.Runtime.VM.VM.InstanceID == nil {
+		return fmt.Errorf("scenario runtime is incomplete for VMSS VM restart")
+	}
+
+	rg := *s.Runtime.Cluster.Model.Properties.NodeResourceGroup
+	instanceID := *s.Runtime.VM.VM.InstanceID
+	poller, err := config.Azure.VMSSVM.BeginRestart(ctx, rg, s.Runtime.VMSSName, instanceID, nil)
+	if err != nil {
+		return fmt.Errorf("begin VMSS VM restart: %w", err)
+	}
+	if _, err = poller.PollUntilDone(ctx, config.PollUntilDoneOptions()); err != nil {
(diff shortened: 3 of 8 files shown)


SeanDougherty (approved):

aadhar-agarwal (approved):

djsly (approved):

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1491,6 +1491,8 @@ disableSSH() {
     systemctlDisableAndStop ssh || exit $ERR_DISABLE_SSH
     # On AzureLinux, the ssh service is named "sshd.service"
     systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
+    # On ACL, port 22 is owned by sshd.socket (socket activation)
+    systemctlDisableAndStop sshd.socket || exit $ERR_DISABLE_SSH
```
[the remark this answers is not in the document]
djsly: good insight

parts/linux/cloud-init/artifacts/cse_config.sh
djsly, line 1786: those `|| ERR_DISABLE_SSH` are useless, since the systemctlDisableAndStop call always returns true. 

This logic of disabling all services on all OS when they are not all existing or might not want to be diable ? could we make this more OS aware ?

djsly (approved):

djsly (changes requested):
can you add a e2e test to ensure that nothing is up after a reboot ?

mayankfz:
> can you add a e2e test to ensure that nothing is up after a reboot ?

Verified E2E on an ACL VHD from this branch.

ACL_DisableSSH passed in build [180699839](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=180699839&view=results). The scenario checks SSH state, reboots, then checks again, the reboot check is the actual regression.

Both times:

_sshd.socket: active=inactive enabled=disabled
sshd.service: active=inactive enabled=disabled_

Two notes

- disableSSH now returns instead of "exit $ERR_DISABLE_SS". Caller in cse_main.sh does disableSSH || exit "$ERR_DISABLE_SSH", so exit 172 is unchanged.
- Not ACL-only. The old code ran systemctlDisableAndStop on both ssh and sshd on every distro, it now dispatches per-OS and fails if an expected unit is missing. Ubuntu/AzureLinux paths change too.

parts/linux/cloud-init/artifacts/cse_config.sh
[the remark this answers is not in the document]
mayankfz: updated

parts/linux/cloud-init/artifacts/cse_config.sh
mayankfz: updated

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -1784,10 +1784,27 @@ setupAmdAma() {
 }
 
 disableSSH() {
-    # On ubuntu, the ssh service is named "ssh.service"
-    systemctlDisableAndStop ssh || exit $ERR_DISABLE_SSH
-    # On AzureLinux, the ssh service is named "sshd.service"
-    systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
+    local units=() unit result=0
```
awesomenix, line 1787: this seems complex. Why not just add

```
 systemctlDisableAndStop sshd.socket || exit $ERR_DISABLE_SSH
```

awesomenix: internally it takes care of existence of service.

mayankfz: > this seems complex. Why not just add
> 
> ```
>  systemctlDisableAndStop sshd.socket || exit $ERR_DISABLE_SSH
> ```

It started out exactly like that, "systemctlDisableAndStop ssh/sshd/sshd.socket". Problem is that helper only echoes on failure and always returns 0, so || exit $ERR_DISABLE_SSH never actually fires. A failed disable would leave port 22 open and CSE would report success. That got flagged in an earlier review, which is why it looks like this now.

The current version returns non-zero if a stop/disable fails, and fails loudly if an expected unit is missing, which is the case that caused this bug in the first place. There are ShellSpec tests covering it, including the exit-172 path.

Happy to go simpler if you'd rather, but that means giving that back up. The other option is fixing 
systemctlDisableAndStop to propagate failure, then this drops to three lines, I just didn't want to change a shared helper that nvidia/dnf also use in this PR. thanks.

djsly: I vote for: systemctlDisableAndStop. can we list which services are leveraging this call ? to see if it's worth failling or not the provsionning

mayankfz: Done, went with flipping the helper. systemctlDisableAndStop now returns non-zero if a stop/disable fails (a unit that isn't installed is still a no-op), so disableSSH is back to the three lines we discussed and the || exit $ERR_DISABLE_SSH actually fires now.

Checked the blast radius first: GPU teardown is unaffected (that branch ends in rm -f, so the function still returns 0, and there's no set -e in CSE), and the iptables call in tool_installs.sh already had || exit $ERR_DISBALE_IPTABLES as dead cod,  that's build-time, and failing there looks right given the comment about pod networking.

Updated the three cse_helpers_spec.sh cases that asserted best-effort success, and reworked the disableSSH specs around unit presence instead of OS branching.

djsly (approved):

djsly:
Added SSH-disable outcome telemetry in ff5ea5aecb.

- `AKS.CSE.disableSSH`: correlated `Started` / `Completed` events provide an actual attempt denominator and expose attempts without a completion event.
- `AKS.CSE.disableSSH.unit`: per-unit inspection, stop and disable retry results, elapsed time, and bounded post-operation systemd state checks.
- Structured JSON includes an attempt ID, OS/version/variant, and pre-provision-only context. Skipped units and unknown diagnostic state are explicit; they are not counted as successful stop operations.
- Existing retry behavior, unit ordering, and exit 172 remain unchanged. Telemetry failures produce warnings, not new provisioning failures. Other shared-helper callers and generic `logs_to_events` behavior are unchanged.

The [telemetry contract and Kusto query](https://github.com/Azure/AgentBaker/blob/ff5ea5aecb/parts/linux/cloud-init/artifacts/README.MD#ssh-disable-telemetry) document deduplication, completed-attempt failure rates, incomplete attempts, and PIS coverage limitations.

Validation: 20 targeted ShellSpec examples passed (SSH telemetry, shared helper, and GPU caller coverage); AgentBaker package tests/snapshot generation and repository-configured ShellCheck passed. The documented Kusto query also passed semantic validation. End-to-end guest-agent ingestion will still need confirmation on a node running this revision.

mayankfz:
> ### AgentBaker Linux gate detective
> **Run:** https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=182262652 **Failed job/stage/task:** `e2e` / `Run AgentBaker E2E` / `ACL_DisableSSH`; plus Test/Scan repeated the repair-linked CIS baseline drift.
> 
> **Detective summary:** E2E failed deterministically in `ACL_DisableSSH`: validation expected SSH to be disabled and inactive, but `sshd.service` was still active even though disabled. The suite reports one failed scenario and no flaky tests.
> 
> **Likely cause / signature:** $disableSig.
> 
> **Confidence:** High.
> 
> **Recommended owner/action:** PR author should inspect the DisableSSH changes and ensure both `sshd.socket` and `sshd.service` are inactive/disabled after provisioning; track the CIS drift separately under repair #39802117.
> 
> **Strongest alternative:** shared E2E infrastructure flake; less likely because the final failure is a scenario-specific SSH service state assertion matching prior PR #9091 failures.
> 
> **Evidence:** E2E final summary names `ACL_DisableSSH` and reports `sshd.socket: active=inactive enabled=disabled` but `sshd.service: active=active enabled=disabled`; timeline marks Run AgentBaker E2E failed; prior wiki row has the same signature for PR #9091.
> 
> **Wiki signature:** [$disableSig](https://dev.azure.com/msazure/09706533-03bf-4b43-9a9b-b49c75429646/_wiki/wikis/ed4a85e9-1085-4151-a39b-2753523eba2b?pagePath=%2FAKS%2FSIGs%20and%20Teams%2FAKS%20Components%2FSIG%3A%20Node%20Lifecycle%2FAI%20Agent%20Knowledge%2FAgentBaker%20Gate%20PR%20Pipeline%20Flakiness%2FNoisy%20Active%20Items)

this will fail becasue its dependent on the ACL base image changes https://github.com/microsoft/azure-container-linux/pull/73 and we have already verified the test with the fixed ACL image changes - https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=174629723&view=results

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -725,13 +725,25 @@ users:
...
+    start_time=$(date -u +"%F %T.%3N")
+    attempt_id="$(date +%s%N)-${BASHPID:-$$}-${RANDOM}"
+    emitSSHDisableEvent "$attempt_id" Started "" "$start_time" 0 null '{}' ||
+        echo "WARNING: could not emit SSH disable start event" >&2
+    # Ubuntu uses ssh, Azure Linux uses sshd, and ACL also uses sshd.socket.
+    for unit in ssh sshd sshd.socket; do
+        if ! systemctlDisableAndStop "$unit" "$attempt_id"; then
+            result=$ERR_DISABLE_SSH
+            break
+        fi
```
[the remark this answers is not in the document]
mayankfz: updated.

parts/linux/cloud-init/artifacts/README.MD
```diff
@@ -2,3 +2,71 @@ For tutorial for the descriptions of components.json, please refer to this [doc]
...
 For CA refresh logging and Linux opt-in certificate diagnostics, see
 [CA refresh diagnostics](../../../../doc/ca-refresh.md).
+
+## SSH-disable telemetry
+
+When `DISABLE_SSH=true` and `basePrep` runs, `disableSSH` emits structured
+guest-agent events. Existing stop/disable retries, unit ordering, missing-unit
+handling, and CSE exit code 172 are unchanged. Telemetry failures are warnings,
+not new provisioning failures. Other `systemctlDisableAndStop` callers and
+`logs_to_events` keep their existing telemetry behavior.
```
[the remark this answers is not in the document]
mayankfz: updated

e2e/scenario/ssh_validation.go
```diff
@@ -0,0 +1,40 @@
+package scenario
+
+const validateSSHServiceDisabledScript = `#!/bin/bash
+set -euo pipefail
+. /etc/os-release
+
+if [[ "$ID" == "ubuntu" ]]; then
+    units=(ssh.service)
+elif [[ "$ID" == "azurecontainerlinux" || ( "$ID" == "azurelinux" && "${VARIANT_ID:-}" == "azurecontainerlinux" ) ]]; then
+    units=(sshd.socket sshd.service)
```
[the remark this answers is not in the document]
mayankfz: updated

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -725,13 +725,26 @@ users:
...
-    systemctlDisableAndStop sshd || exit $ERR_DISABLE_SSH
+    local start_time attempt_id unit result=0 start_seconds=$SECONDS
+    start_time=$(date -u +"%F %T.%3N")
+    attempt_id="$(date +%s%N)-${BASHPID:-$$}-${RANDOM}"
+    emitSSHDisableEvent "$attempt_id" Started "" "$start_time" 0 null '{}' ||
+        echo "WARNING: could not emit SSH disable start event" >&2
+    # Ubuntu uses ssh, Azure Linux uses sshd, and ACL also uses sshd.socket.
+    # Keep going after a failure so a later unit still gets disabled, on ACL the
+    # socket owns port 22 and skipping it would leave SSH reachable.
+    for unit in ssh sshd sshd.socket; do
```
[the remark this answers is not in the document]
mayankfz: Skipping this one for now. After configureSSHService disables ssh.socket, the unit file is still there, so adding it to the loop means we would stop/disable it on every SSH-disabled Ubuntu node ~2,448 pools, and any failure would now fail provisioning. That's a big scope increase for an ACL-only PR, to guard against a VHD-build path that's already visible as a warning.

PR #9104, opened 2026-08-01
fix: rennovate all packages manually since some moved to v2 etc

awesomenix:
lots of changes to rennovate

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -338,22 +338,6 @@
     },
     {
       "matchPackageNames": [
-        "oss/v2/open-policy-agent/gatekeeper",
-        "azure-policy/policy-kubernetes-addon-prod"
-      ],
-      "assignees": [
-        "kevinhwangkr-msft",
-        "anlandu",
-        "nreisch"
-      ],
-      "reviewers": [
-        "kevinhwangkr-msft",
-        "anlandu",
-        "nreisch"
-      ]
-    },
-    {
-      "matchPackageNames": [
         "oss/binaries/kubernetes/azure-acr-credential-provider",
         "azure-acr-credential-provider",
         "oss/v2/kubernetes/azure-acr-credential-provider-sysext"
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -97,22 +97,6 @@
       ]
     },
     {
-      "downloadURL": "mcr.microsoft.com/oss/kubernetes/pause:*",
-      "amd64OnlyVersions": [],
-      "multiArchVersionsV2": [
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/pause",
-          "latestVersion": "3.6"
-        }
-      ],
-      "windowsVersions": [
-        {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=oss/kubernetes/pause",
-          "latestVersion": "3.9-hotfix-20230808"
-        }
-      ]
-    },
-    {
       "downloadURL": "mcr.microsoft.com/oss/v2/kubernetes/pause:*",
       "amd64OnlyVersions": [],
       "multiArchVersionsV2": [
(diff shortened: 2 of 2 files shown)


.github/renovate.json
```diff
@@ -336,22 +336,6 @@
         "ashokobulapuram7"
       ]
     },
-    {
```
awesomenix, line 339: this component is deprecated

parts/common/components.json
```diff
@@ -96,22 +96,6 @@
         }
       ]
     },
-    {
-      "downloadURL": "mcr.microsoft.com/oss/kubernetes/pause:*",
```
awesomenix, line 100: dont use v2 anymore

parts/common/components.json
```diff
@@ -205,10 +170,22 @@
...
-        },
+        }
+      ]
+    },
+    {
+      "downloadURL": "mcr.microsoft.com/v2/containernetworking/azure-cni:*",
+      "amd64OnlyVersions": [],
+      "windowsVersions": [
         {
-          "renovateTag": "registry=https://mcr.microsoft.com, name=containernetworking/azure-cni",
```
awesomenix, line 210: this has moved to v2

parts/common/components.json
```diff
@@ -1886,136 +1799,76 @@
...
                 "renovateTag": "RPM_registry=https://packages.microsoft.com/azurelinux/3.0/prod/cloud-native/x86_64/repodata, name=azure-acr-credential-provider, os=azurelinux, release=3.0",
                 "latestVersion": "1.34.12-2.azl3",
-                "previousLatestVersion": "1.34.12-1.azl3"
+                "previousLatestVersion": "1.34.11-2.azl3"
               },
               {
                 "k8sVersion": "1.35",
                 "renovateTag": "RPM_registry=https://packages.microsoft.com/azurelinux/3.0/prod/cloud-native/x86_64/repodata, name=azure-acr-credential-provider, os=azurelinux, release=3.0",
                 "latestVersion": "1.35.7-2.azl3",
-                "previousLatestVersion": "1.35.7-1.azl3"
```
awesomenix, line 1993: cant have same major minor version and different patch version

timmy-wright: Well, we can. Hopefully the script will only cache it once - and it means on the next upgrade, the "latestVersion" will move to "previousLatestVersion" and all will be good again. In any case, with the rate of new versions of the components and the rollout frequencies of aks-rp and VHDs, I don't think caching a single previous latest version is useful anymore.

.github/renovate.json
cameronmeissner: you confirmed this was @anlandu @nreisch?

parts/common/components.json
```diff
@@ -1508,6 +1455,11 @@
                 "k8sVersion": "1.35",
                 "renovateTag": "name=kubelet, repository=production, os=ubuntu, release=22.04",
                 "latestVersion": "1.35.6-ubuntu22.04u1"
+              },
+              {
+                "k8sVersion": "1.36",
+                "renovateTag": "name=kubelet, repository=production, os=ubuntu, release=22.04",
+                "latestVersion": "1.36.3-ubuntu22.04u1"
```
cameronmeissner, line 1462: way want to match what we're using for 2404/2604

parts/common/components.json
cameronmeissner, line 1419: can we also update 2604 to match the other ubuntu versions?

.github/renovate.json
awesomenix: PR from owners https://github.com/Azure/AgentBaker/pull/8327

parts/common/components.json
awesomenix: Will handle it in different PR using rennovate

parts/common/components.json
awesomenix: Will handle it in different PR using rennovate

parts/common/components.json
awesomenix: I think this causes failures while installation because we have conflicting versions, obviously our scripts are the problem, but thats a problem to solve another day, for now we are keepinglatest and latest-1 versions.

cameronmeissner (approved):

PR #9128, opened 2026-08-04
fix: update pause to 3.10.2

timmy-wright:

Updates pause image to 3.10.2

Fixes #


awesomenix (approved):

awesomenix (approved):

timmy-wright:
## CI Failure Analysis

Investigated the 3 failing checks. **None appear to be regressions caused by this PR** — this PR only bumps the cached "pause" sandbox image (Linux `3.6` → `3.10.2`, Windows `3.10.1` → `3.10.2`) plus matching comment/test/mocks updates. No logic changes.

### 1. Linux VHD Build gate — `Test_Ubuntu2604Minimal_CSE_CachedPerformance`
CSE task `AKS.CSE.ubuntuSnapshotUpdate` took 2.09s vs a 2s perf threshold. This function (`ensureSnapshotUpdate`) is untouched by this PR and unrelated to the pause image — looks like a marginal timing flake (0.09s over threshold).

### 2. GPU E2E — `nvidia-cdi-refresh.path`/`.service` entered failed state
Failed on `Test_Ubuntu2204_GPUA100` and `Test_Ubuntu2204_GPUNC`. This is a **known upstream nvidia-container-toolkit systemd race**: the CDI-refresh unit can fire before the NVIDIA driver/NVML is fully loaded at boot ([nvidia-container-toolkit#1611](https://github.com/NVIDIA/nvidia-container-toolkit/issues/1611)). This PR doesn't touch GPU driver install, containerd-toolkit version, or CDI generation — only the `ubuntu/gb` (arm64 Grace-Blackwell) containerd sandbox image, which isn't used by these test images. `Test_Ubuntu2404_GPU_H100`/`A100` additionally hit bastion-tunnel `403`s — infra flakiness.

### 3. Windows E2E — `Test_Windows2022CachingRegression`
Failed waiting for a test pod to become ready: `client rate limiter Wait returned an error: context deadline exceeded` — an API-server client-side throttling signature, not a code issue. The only Windows file touched here (`kubernetesfunc.ps1`) is a comment-only change.

**Recommendation:** rerun the failed jobs. If `nvidia-cdi-refresh` recurs on unrelated PRs it may warrant its own tracking issue, but it doesn't look tied to this change.

parts/common/components.json
```diff
@@ -102,13 +102,15 @@
       "multiArchVersionsV2": [
         {
           "renovateTag": "registry=https://mcr.microsoft.com, name=oss/v2/kubernetes/pause",
-          "latestVersion": "3.6"
+          "latestVersion": "3.10.2",
+          "previousLatestVersion": "3.6"
```
r2k1, line 106: Do you want to cache both images?

r2k1 (approved):

parts/common/components.json
djsly: yes we have to. this is critical that this image is present all the time .  NI, etc

djsly: # These images are critical for pod startup and aren't supported with private ACR since containerd won't be using azure-acr-credential to fetch them.

timmy-wright: Ouch - that would be nasty. Hmm. This probably means that I'll need to delay the change to aks-rp to use 3.10.2 for a month to make sure the VHDs that contain the new version are globally available.

PR #9129, opened 2026-08-04
feat: use MIG profile layout for partitioning

karenychen:
## What this PR does

Completes the runtime half of the MIG profile layout contract introduced by #9081.

- treats either `MIGProfileLayout` or legacy `GPUInstanceProfile` as a MIG node in classic CSE and aks-node-controller
- makes a non-empty `MIGProfileLayout` authoritative when both fields are populated
- maps the ordered layout directly to NVIDIA profile IDs without sorting, capacity expansion, or `MigStrategy` interpretation
- retains legacy uniform expansion when only `GPUInstanceProfile` is populated
- propagates failures from both `nvidia-smi mig -cgi` and `nvidia-smi mig -cci`
- keeps `MigStrategy` solely for nvidia-device-plugin advertisement behavior
- adds focused ShellSpec coverage plus Single, Mixed, and legacy GPU e2e coverage

This supersedes the runtime/test portion of closed #8869. AKS RP layout calculation remains out of scope.

## Validation

- `make generate` regenerated Go test data and the manifest without changes; its repository-wide ShellCheck tail stops on the existing POSIX `SC3014` baseline
- `go test ./pkg/agent/... -count=1`
- `go test ./parser/... -count=1` from `aks-node-controller`
- `go test . -run '^$' -count=1` from `e2e`
- focused ShellCheck for `mig-partition.sh` and its ShellSpec
- focused ShellSpec: 15 examples, 0 failures
- Buf lint
- Buf breaking check against `Azure/AgentBaker` main
- `git diff --check`


karenychen:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

aks-node-controller/parser/parser.go
```diff
@@ -46,6 +46,7 @@ func getCSEEnv(ctx context.Context, config *aksnodeconfigv1.Configuration, gpuCo
 
 	containerdVersion, _ := detectContainerdVersion(ctx)
 	cloudProviderSettings := getCloudProviderSettings(config)
+	migConfig := config.GetGpuConfig()
```
karenychen:
```suggestion
	isMigNode := getIsMIGNode(config.GetGpuConfig().GetGpuInstanceProfile(), config.GetGpuConfig().GetMigProfileLayout)
```

aks-node-controller/parser/parser_test.go
```diff
@@ -79,8 +84,7 @@ oom_score = -999
 			validator: func(cmd *exec.Cmd) {
 				vars := environToMap(cmd.Env)
 				assertHasKeyWithValue(t, vars, "NVIDIA_MIG_PROFILE_LAYOUT", "MIG3g,MIG2g,MIG1g,MIG1g")
-				// TODO: Make MIG_NODE true if either NVIDIA_MIG_PROFILE_LAYOUT or GPU_INSTANCE_PROFILE is set.
-				assertHasKeyWithValue(t, vars, "MIG_NODE", "false")
+				assertHasKeyWithValue(t, vars, "MIG_NODE", "true")
```
karenychen, line 89: also assert values for `GPU_INSTANCE_PROFILE`, `NVIDIA_MIG_PROFILE_LAYOUT` and `NVIDIA_MIG_STRATEGY`

aks-node-controller/parser/parser_test.go
```diff
@@ -497,6 +501,7 @@ func TestAKSNodeConfigCompatibilityFromJsonToCSECommand(t *testing.T) {
 				assertHasKeyWithValue(t, vars, "LOCATION", "")
 				assertHasKeyWithValue(t, vars, "GPU_NODE", "false")
 				assertHasKeyWithValue(t, vars, "GPU_INSTANCE_PROFILE", "")
+				assertHasKeyWithValue(t, vars, "NVIDIA_MIG_PROFILE_LAYOUT", "")
```
karenychen, line 539: add `assertHasKeyWithValue(t, vars, "MIG_NODE", "true")`

e2e/validators.go
```diff
@@ -2717,13 +2751,34 @@ func ValidateMIGModeEnabled(ctx context.Context, s *Scenario, gpuCountExpected i
...
+	require.Equal(s.T, expected, actual, "unexpected MIG instance geometry.\nOutput:\n%s", stdout)
+}
+
 func ValidateMIGInstancesCreated(ctx context.Context, s *Scenario, migProfile string, instanceCountExpected int) {
 	s.T.Helper()
 	s.T.Logf("validating that %d MIG instances are created with profile %s", instanceCountExpected, migProfile)
 
 	command := []string{
 		"set -ex",
-		// List MIG devices using nvidia-smi
```
karenychen, line 3315: restore

spec/parts/linux/cloud-init/artifacts/mig_partition_spec.sh
```diff
@@ -0,0 +1,121 @@
...
+    End
+
+    It 'prioritizes the layout when the legacy scalar is also populated'
+        When run env GPU_INSTANCE_PROFILE="MIG7g" NVIDIA_MIG_PROFILE_LAYOUT="MIG2g,MIG1g" bash "$SCRIPT_PATH"
+
+        The status should be success
+        The output should include "nvidia-smi mig -cgi 14,19"
+    End
+
+    It 'does not use MIG strategy to calculate the layout'
```
karenychen:
```suggestion
    It 'does not use MIG strategy to calculate the layout since it is already computed on caller site'
```

aks-node-controller/parser/parser.go
karenychen: Applied in 737b7fd0: isMigNode is computed before constructing the environment map.

aks-node-controller/parser/parser_test.go
karenychen: Applied in 737b7fd0: the layout-only test now asserts GPU_INSTANCE_PROFILE, NVIDIA_MIG_PROFILE_LAYOUT, and NVIDIA_MIG_STRATEGY.

aks-node-controller/parser/parser_test.go
karenychen: Added the MIG_NODE assertion in 737b7fd0. This fixture uses an empty configuration, so the correct default is false; asserting true would contradict getIsMIGNode and incorrectly classify every empty config as a MIG node.

e2e/validators.go
karenychen: Restored the original List MIG devices using nvidia-smi comment in 737b7fd0.

spec/parts/linux/cloud-init/artifacts/mig_partition_spec.sh
karenychen: Applied the suggested test description in 737b7fd0.

parts/linux/cloud-init/artifacts/mig-partition.sh
```diff
@@ -2,26 +2,104 @@
...
+    MIG_LAYOUT="$(parse_mig_profile_layout "${NVIDIA_MIG_PROFILE_LAYOUT}")" || exit 1
+elif [ -n "${GPU_INSTANCE_PROFILE:-}" ]; then
+    MIG_LAYOUT="$(uniform_mig_profile_layout "${GPU_INSTANCE_PROFILE}")" || exit 1
+else
+    echo "neither NVIDIA_MIG_PROFILE_LAYOUT nor GPU_INSTANCE_PROFILE is set" >&2
+    exit 1
+fi
+
+nvidia-smi mig -cgi "${MIG_LAYOUT}" || exit $?
+nvidia-smi mig -cci
```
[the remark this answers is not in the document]
karenychen: No code change is needed here. `nvidia-smi mig -cci` is the final command in the script, so Bash uses its exit status as the script's exit status automatically. The focused ShellSpec case mocks `-cci` returning nonzero and confirms the script fails; the suite passes with 16 examples and 0 failures.

karenychen:
the agentbaker GPU E2E fails because this change requires rebuilding VHDs. Here is the passed GPU E2E run with VHD built from this branch: https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=177161624&view=results

karenychen (commented):
### Code review

Found 1 issue:

1. High Risk - Backward Compatibility: `datamodel.IsMIGNode` is an exported API in current published AgentBaker module tags, and changing it from `IsMIGNode(string)` to `IsMIGNode(string, []string)` breaks existing downstream callers at compile time. Preserve the one-argument function and add a new layout-aware helper (or another backward-compatible API). This is also tracked by the unresolved inline review thread on this line.

https://github.com/Azure/AgentBaker/blob/7131e37fd2ec7f427fe098e859053bea07f5bedd/pkg/agent/datamodel/helper.go#L128-L131

parts/linux/cloud-init/artifacts/mig-partition.sh
```diff
@@ -2,26 +2,104 @@
 
 #NOTE: Currently, Nvidia library mig-parted (https://github.com/NVIDIA/mig-parted) cannot work properly because of the outdated GPU driver version
```
ganeshkumarashok, line 3: still the case?

parts/linux/cloud-init/artifacts/mig-partition.sh
```diff
@@ -2,26 +2,104 @@
...
+            echo "0"
+            ;;
+        *)
+            echo "not a valid MIG profile: $1" >&2
+            return 1
+            ;;
+    esac
+}
+
+# TODO: Support GPU models with fewer than seven total partitions.
```
ganeshkumarashok, line 30: Which GPU instances are those?

parts/linux/cloud-init/artifacts/mig-partition.sh
karenychen: I took a look at this before but not sure if it's the path we want to go: this would require us introducing a new dependency for the `nvidia-mig-parted` tool. Also I am not sure if it is as helpful with our mixed MIG model since it would require us to deliberately define all possible MIG profile combinations. I am okay with replacing this TODO with a comment

parts/linux/cloud-init/artifacts/mig-partition.sh
karenychen: Oops I meant to remove this comment since the new GPUProfileLayout should already support it (we will compute the exact layout on RP side).

There are a few GPUs that support != 7 partitions documented on the official MIG user guide https://docs.nvidia.com/datacenter/tesla/mig-user-guide/latest/supported-mig-profiles.html

In Azure, we currently don't support any of these. However, I have spoken with Sachi and we want to maintain the flexibility to support them

karenychen:
passing E2E after resolving merge conflict: https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177740936&view=results

parts/linux/cloud-init/artifacts/mig-partition.sh
karenychen: will remove that comment in a follow-up PR

ganeshkumarashok: Maintaining the list of available options manually is not ideal long-term. We should look into how we can avoid hardcoding it. That's more so for the future though.

aks-node-controller/parser/helper.go
```diff
@@ -250,8 +250,8 @@ func isContainerdV2(version string) bool {
 	return IsKubernetesVersionGe(version, "2.0.0")
 }
 
-func getIsMIGNode(gpuInstanceProfile string) bool {
-	return gpuInstanceProfile != ""
+func getIsMIGNode(gpuInstanceProfile string, migProfileLayout []string) bool {
+	return gpuInstanceProfile != "" || len(migProfileLayout) > 0
```
ganeshkumarashok, line 254: This could cause issues for new CSE + old VHD? Say RP provides migProfileLayout in the new CSE but old VHD would not have the `mig-partition.sh` so it would silently no-op?

ganeshkumarashok: Shipping `mig-partition.sh` with the VHD should fix that?

karenychen: we will gate this with a version check in the frontend validator so cx can only set this on or after a certain vhd version

parts/linux/cloud-init/artifacts/mig-partition.sh
karenychen: Yes agreed, this one will probably also be done from the validator side. Maintaining a list of vm sizes for the MIG options like you said is probably not ideal long term but it does give us the reliability + latency saves at the moment. Let's revisit in the future

ganeshkumarashok:
Approving conditional on e2es passing here and on the standalone side

ganeshkumarashok (approved):

PR #9136, opened 2026-08-05
fix: persist Azure Linux THP settings after reboot

djsly:
## Summary

Fix Azure Linux custom Linux OS config persistence after node reboot for support case 2608050010000389. AgentBaker was writing transparent huge page settings to `/etc/sysfs.conf`, but Azure Linux does not reliably apply that file at boot, so THP reverted to the distro default after reboot.

## Changes

- Add an AKS-owned Azure Linux/Mariner systemd oneshot service to reapply custom THP enabled/defrag settings on boot before kubelet starts.
- Keep the existing `/etc/sysfs.conf` behavior for compatibility with distros that apply it.
- Persist custom swap activation through an AKS systemd service with `noauto,nofail` fstab entries to avoid Azure Linux `/mnt` swap ordering cycles after reboot.
- Add an Azure Linux V3 E2E regression test that validates custom Linux OS config values before and after reboot.
- Fix E2E helpers so `DISABLE_SCRIPTLESS=true` fully exercises local CSE script changes during local validation.

## Testing

- `cd e2e && go test . -run '^Test_AzureLinuxV3_CustomLinuxOSConfigPersistsAfterReboot$' -count=0`
- `DISABLE_SCRIPTLESS=true TAGS_TO_RUN='name=Test_AzureLinuxV3_CustomLinuxOSConfigPersistsAfterReboot' ./e2e-local.sh` — passed
- `make generate` attempted, but blocked while installing ginkgo by internal Go proxy `401 Unauthorized`; fallback snapshot generation above passed.

## Related

- AB#39130468
- Support case: 2608050010000389

### DIFF
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -24,12 +24,12 @@ configureTransparentHugePage() {
     local thp_defrag_path="/sys/kernel/mm/transparent_hugepage/defrag"
 
     if [ -n "${THP_ENABLED}" ]; then
-        printf '%s\n' "${THP_ENABLED}" > "${thp_enabled_path}"
-        printf 'kernel/mm/transparent_hugepage/enabled=%s\n' "${THP_ENABLED}" >> "${etc_sysfs_conf}"
+        printf '%s\n' "${THP_ENABLED}" > "${thp_enabled_path}" || exit "$ERR_SYSCTL_RELOAD"
+        printf 'kernel/mm/transparent_hugepage/enabled=%s\n' "${THP_ENABLED}" >> "${etc_sysfs_conf}" || exit "$ERR_SYSCTL_RELOAD"
     fi
     if [ -n "${THP_DEFRAG}" ]; then
-        printf '%s\n' "${THP_DEFRAG}" > "${thp_defrag_path}"
-        printf 'kernel/mm/transparent_hugepage/defrag=%s\n' "${THP_DEFRAG}" >> "${etc_sysfs_conf}"
+        printf '%s\n' "${THP_DEFRAG}" > "${thp_defrag_path}" || exit "$ERR_SYSCTL_RELOAD"
+        printf 'kernel/mm/transparent_hugepage/defrag=%s\n' "${THP_DEFRAG}" >> "${etc_sysfs_conf}" || exit "$ERR_SYSCTL_RELOAD"
     fi
     reconcileTransparentHugePagePersistence
 }
@@ -43,17 +43,33 @@ reconcileTransparentHugePagePersistence() {
 configureTransparentHugePageSystemdService() {
     local service_name="aks-transparent-hugepage"
     local script_path="/opt/azure/containers/aks-transparent-hugepage.sh"
+    local config_dir="/opt/azure/containers/aks-transparent-hugepage"
     local service_path="/etc/systemd/system/${service_name}.service"
 
-    mkdir -p "$(dirname "${script_path}")" || exit "$ERR_SYSCTL_RELOAD"
-    if ! tee "${script_path}" > /dev/null <<EOF
+    mkdir -p "$(dirname "${script_path}")" "${config_dir}" || exit "$ERR_SYSCTL_RELOAD"
+    if [ -n "${THP_ENABLED}" ]; then
+        printf '%s\n' "${THP_ENABLED}" | tee "${config_dir}/enabled" > /dev/null || exit "$ERR_SYSCTL_RELOAD"
+    else
+        rm -f "${config_dir}/enabled" || exit "$ERR_SYSCTL_RELOAD"
+    fi
+    if [ -n "${THP_DEFRAG}" ]; then
+        printf '%s\n' "${THP_DEFRAG}" | tee "${config_dir}/defrag" > /dev/null || exit "$ERR_SYSCTL_RELOAD"
+    else
+        rm -f "${config_dir}/defrag" || exit "$ERR_SYSCTL_RELOAD"
+    fi
+
+    if ! tee "${script_path}" > /dev/null <<'EOF'
 #!/bin/bash
 set -e
-if [ -n "${THP_ENABLED}" ]; then
-    printf '%s\n' "${THP_ENABLED}" > /sys/kernel/mm/transparent_hugepage/enabled
+config_dir="/opt/azure/containers/aks-transparent-hugepage"
+thp_enabled_config="${config_dir}/enabled"
+thp_defrag_config="${config_dir}/defrag"
+
+if [ -s "${thp_enabled_config}" ]; then
+    cat "${thp_enabled_config}" > /sys/kernel/mm/transparent_hugepage/enabled
 fi
-if [ -n "${THP_DEFRAG}" ]; then
-    printf '%s\n' "${THP_DEFRAG}" > /sys/kernel/mm/transparent_hugepage/defrag
+if [ -s "${thp_defrag_config}" ]; then
+    cat "${thp_defrag_config}" > /sys/kernel/mm/transparent_hugepage/defrag
 fi
 EOF
     then
--- a/parts/linux/cloud-init/artifacts/cse_config.sh
+++ b/parts/linux/cloud-init/artifacts/cse_config.sh
@@ -17,15 +17,49 @@ Environment="KUBE_API_SERVER_NAME=${API_SERVER_NAME}"
 EOF
   systemctlEnableAndStart reconcile-private-hosts 30 || exit $ERR_SYSTEMCTL_START_FAIL
 }
+
+validateTransparentHugePageValue() {
+    local setting_name="$1"
+    local value="$2"
+    local supported_values_path="$3"
+
+    if [ -z "${value}" ]; then
+        return 0
+    fi
+
+    case "${value}" in
+        *[!abcdefghijklmnopqrstuvwxyz+]*)
+            echo "Invalid transparent huge page ${setting_name} value '${value}': only lowercase THP tokens are supported" >&2
+            return 1
+            ;;
+    esac
+
+    if [ ! -r "${supported_values_path}" ]; then
+        echo "Cannot validate transparent huge page ${setting_name}; ${supported_values_path} is not readable" >&2
+        return 1
+    fi
+
+    if ! tr ' ' '\n' < "${supported_values_path}" | tr -d '[]' | grep -Fx "${value}" > /dev/null; then
+        echo "Unsupported transparent huge page ${setting_name} value '${value}'. Supported values: $(tr -d '[]' < "${supported_values_path}")" >&2
+        return 1
+    fi
+}
+
 configureTransparentHugePage() {
-    ETC_SYSFS_CONF="/etc/sysfs.conf"
+    local etc_sysfs_conf="/etc/sysfs.conf"
+    local thp_enabled_path="/sys/kernel/mm/transparent_hugepage/enabled"
+    local thp_defrag_path="/sys/kernel/mm/transparent_hugepage/defrag"
+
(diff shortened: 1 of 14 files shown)


parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -21,12 +21,52 @@ configureTransparentHugePage() {
...
+    if { [ -n "${THP_ENABLED}" ] || [ -n "${THP_DEFRAG}" ]; } && isMarinerOrAzureLinux "$OS" "$OS_VARIANT"; then
+        configureTransparentHugePageSystemdService
+    fi
+}
+
+configureTransparentHugePageSystemdService() {
+    local service_name="aks-transparent-hugepage"
+    local script_path="/opt/azure/containers/aks-transparent-hugepage.sh"
+    local service_path="/etc/systemd/system/${service_name}.service"
+
```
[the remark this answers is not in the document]
djsly: Added validation for THP enabled/defrag values against the kernel-supported sysfs tokens before writing them or generating the boot-time helper script, with ShellSpec coverage for supported, unsupported, and metacharacter-containing values.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -86,7 +160,48 @@ configureSwapFile() {
...
+    swap_mount_path="$(dirname "${swap_location}")"
+
+    mkdir -p "$(dirname "${script_path}")"
+    tee "${script_path}" > /dev/null <<EOF
+#!/bin/bash
+set -e
+if ! swapon --show=NAME --noheadings | grep -Fxq "${swap_location}"; then
+    swapon "${swap_location}"
+fi
```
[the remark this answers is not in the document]
djsly: Fixed in `37f2e1775a`: both the generated swap oneshot and the runtime verification now normalize `swapon --show=NAME --noheadings` with `awk '{print $1}'` before exact path matching, so leading whitespace no longer causes an already-active swap file to be missed.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -49,15 +133,67 @@ configureSystemdUseDomains() {
...
+    local swap_location="${1:-}"
+
+    if [ -z "${swap_location}" ]; then
+        swap_location="$(findExistingSwapFileLocation || true)"
+    fi
+
+    if [ -z "${swap_location}" ]; then
+        echo "No existing AKS swap file found; skipping swap file persistence reconciliation"
+        return 0
+    fi
```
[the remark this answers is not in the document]
djsly: Updated `reconcileSwapFilePersistence` so a node-prep reconciliation with no existing swapfile now calls `configureSwapFile`, which creates/activates the requested swapfile or fails through the existing swap setup errors instead of silently skipping. I also added ShellSpec coverage for both the missing-swapfile create path and the existing-swapfile reconciliation path.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

spec/parts/linux/cloud-init/artifacts/cse_config_spec.sh
```diff
@@ -2683,4 +2881,138 @@ EOF
...
     End
+
+    Describe 'configureSwapFile'
+        SWAP_FILE_SIZE_MB=1
+
+        function [ {
+            if test "$1" = "-L" && test "$2" = "/dev/disk/azure/resource-part1"; then
+                return 0
+            fi
+            command [ "$@"
```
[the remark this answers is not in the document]
djsly: Updated the bracket stub to explicitly preserve the closing `]` when Bash passes it through, and append it only if absent. I reran the focused ShellSpec coverage for `configureSwapFile` and `reconcileSwapFilePersistence` successfully.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -17,18 +17,108 @@ Environment="KUBE_API_SERVER_NAME=${API_SERVER_NAME}"
 EOF
   systemctlEnableAndStart reconcile-private-hosts 30 || exit $ERR_SYSTEMCTL_START_FAIL
 }
+
+validateTransparentHugePageValue() {
```
awesomenix: Shouldnt this validation be done in baker.go where the value is provided? so in case of invalid values we throw this in validator instead of failed node bootstrapping

awesomenix:
>Azure Linux does not reliably apply that file at boot, so THP reverted to the distro default after reboot.

Did we confirm with azure linux team on why?

awesomenix:
>Persist custom swap activation through an AKS systemd service with noauto,nofail fstab entries to avoid Azure Linux /mnt swap ordering cycles after reboot.

How do we know that this is reliable?

parts/linux/cloud-init/artifacts/cse_config.sh
djsly: Moved this into AgentBaker request validation so unsupported THP values now fail before bootstrap data is generated. I kept the CSE validation as defense-in-depth for runtime/kernel-supported values.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

djsly:
> Azure Linux does not reliably apply that file at boot, so THP reverted to the distro default after reboot.
>
> Did we confirm with azure linux team on why?

Not yet with the Azure Linux team. The fix intentionally stops depending on `/etc/sysfs.conf` being replayed by the distro boot path: AKS owns the CustomLinuxOSConfig contract, so the PR now installs/reconciles an AKS oneshot service that runs after `systemd-sysctl.service` and before `kubelet.service` on Azure Linux/Mariner.

> Persist custom swap activation through an AKS systemd service with noauto,nofail fstab entries to avoid Azure Linux /mnt swap ordering cycles after reboot.
>
> How do we know that this is reliable?

The swap service is generated by CSE, reconciled again from `nodePrep` for PIS/scale-up cases, has `After=local-fs.target`, `RequiresMountsFor=<swap mount path>`, `ConditionPathExists=<swapfile>`, and `Before=kubelet.service`, then runs `swapon` only if the exact swap file is not already active. The E2E validates the full contract by checking sysctls, ulimits, swap, and THP before and after a VM reboot.

pkg/agent/baker.go
```diff
@@ -785,6 +789,46 @@ func ValidateAndSetLinuxNodeBootstrappingConfiguration(config *datamodel.NodeBoo
...
+	if err := validateTransparentHugePageConfigValue("transparentHugePageEnabled", config.TransparentHugePageEnabled, map[string]struct{}{
+		"always":  {},
+		"madvise": {},
+		"never":   {},
+	}); err != nil {
+		return err
+	}
+
+	return validateTransparentHugePageConfigValue("transparentHugePageDefrag", config.TransparentHugePageDefrag, map[string]struct{}{
+		"always":        {},
```
awesomenix: Define valid values constant
const (
AllowedTransparentHugePageEnabledOptions = []string{"always","madvise","never"}
)
AllowedTransparentHugePageDefragOptions = []string{"always","defer","defer+madvise","never"}
)
)

Then you can do 

slices.Contains(AllowedTransparentHugePageEnabledOptions, value)

pkg/agent/baker.go
```diff
@@ -785,6 +789,46 @@ func ValidateAndSetLinuxNodeBootstrappingConfiguration(config *datamodel.NodeBoo
...
+	if _, ok := allowedValues[value]; ok {
+		return nil
+	}
+
+	allowed := make([]string, 0, len(allowedValues))
+	for allowedValue := range allowedValues {
+		allowed = append(allowed, allowedValue)
+	}
+	sort.Strings(allowed)
+	return fmt.Errorf("customLinuxOSConfig.%s value %q is invalid; allowed values are: %s", fieldName, value, strings.Join(allowed, ", "))
```
awesomenix: if its a []string you can you strings.Join()

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -17,18 +17,108 @@ Environment="KUBE_API_SERVER_NAME=${API_SERVER_NAME}"
...
+    validateTransparentHugePageValue "defrag" "${THP_DEFRAG}" "/sys/kernel/mm/transparent_hugepage/defrag" || exit "$ERR_SYSCTL_RELOAD"
+}
+
 configureTransparentHugePage() {
-    ETC_SYSFS_CONF="/etc/sysfs.conf"
+    local etc_sysfs_conf="/etc/sysfs.conf"
+    local thp_enabled_path="/sys/kernel/mm/transparent_hugepage/enabled"
+    local thp_defrag_path="/sys/kernel/mm/transparent_hugepage/defrag"
+
+    validateTransparentHugePageConfig
```
awesomenix: can we remove this now since its validated in baker.go

pkg/agent/baker.go
djsly: Done — refactored the enabled/defrag allowed values into reusable package-level slices and switched validation to `slices.Contains`. I kept them as vars since Go does not support constant slices.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

pkg/agent/baker.go
djsly: Done — the invalid-value error now uses `strings.Join` directly on the allowed-value slice, so the text stays deterministic without building/sorting a map.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

parts/linux/cloud-init/artifacts/cse_config.sh
djsly: I kept this runtime validation as defense-in-depth. `baker.go` validates the static AKS API contract, but the CSE check still validates against the live node kernel-supported THP tokens and rejects malformed values before generating the reboot oneshot script; it also protects any path that bypasses API validation.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

awesomenix: this is not proper validation, all it checks is that its has right characters and not valid values, this check is meaningless and wasteful

djsly: Good point — I removed the shell-side THP validation and the matching ShellSpec coverage. The generated CSE now relies on the semantic API-level validation in pkg/agent/baker.go, so the script only applies values that AgentBaker has already accepted.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

parts/linux/cloud-init/artifacts/cse_config.sh
```diff
@@ -49,20 +103,97 @@ configureSystemdUseDomains() {
...
+    echo "${fstab_entry}" >> "${temp_fstab}" || {
+        rm -f "${temp_fstab}"
+        return 1
+    }
+    cat "${temp_fstab}" > "${fstab_file}" || {
+        rm -f "${temp_fstab}"
+        return 1
+    }
+    rm -f "${temp_fstab}"
+}
```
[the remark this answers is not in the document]
djsly: Fixed — `ensureSwapFileFstabEntry` now writes the replacement fstab in the same directory and atomically moves it into place, so a failed replace leaves the existing fstab untouched. I also added focused ShellSpec coverage for that failure path.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

awesomenix (approved):

awesomenix (approved):

awesomenix (approved):

PR #9142, opened 2026-08-06
test(e2e): add Kata Containers E2E scenario for AzureLinux V3

Bickor:

AgentBaker has no E2E coverage for Kata Containers today (`grep -ri kata e2e/` returns nothing), even though Kata VHDs are built and shipped by this repo.

Because Kata is a *runtime*, the thing that can silently break is the containerd configuration. `pkg/agent/baker.go` only emits the Kata runtime handler block when the agent pool's `Distro` satisfies `Distro.IsKataDistro()`, and nothing verified that the result is actually usable on the node.

This adds `Test_AzureLinuxV3Gen2Kata`, which asserts three increasingly strong properties:

1. **Rendered config** — `/etc/containerd/config.toml` carries the `kata` handler block, the kata v2 shim `runtime_type`, and `disable_snapshot_annotations = false`.
2. **Effective config** — `containerd config dump`, run over SSH on the node, shows the handler in the *parsed* config and emits no `level=warning` on either stdout or stderr.
3. **End-to-end** — for every handler in `kataRuntimeHandlers`, a pod scheduled through a Kata `RuntimeClass` reaches Running and is genuinely VM-isolated (guest `uname -r` differs from the host's). Reaching Running proves the handler is registered with CRI; the kernel comparison proves the pod did not silently fall back to shared-kernel runc.

Verified against a real E2E run (build 175521223): the node bootstrapped, the kata handler was present in the effective config, and the Kata pod ran with guest kernel `6.6.137.mshv1-1.azl3` vs host `6.6.137.mshv2-1.azl3`.

Selecting a Kata VHD flows through `e2e/node_config.go` -> `AgentPoolProfile.Distro` -> the `IsKata` template func, so the scenario exercises that whole path against a real node. The PR is purely additive to `e2e/`.

Additionally, `nbcToAKSNodeConfigV1` now derives `Configuration.IsKata` from the distro. The aks-node-controller path gates its Kata containerd blocks on that proto field alone, while the NBC/baker path uses the distro, and nothing connected the two. This is inert for existing scenarios, since `nbcToAKSNodeConfigV1` only runs when a scenario sets `AKSNodeConfigMutator`.

**Why step 2 exists, and what it does and does not prove**

Kata VHDs ship their own containerd (CSE skips installing one — see the `azurelinuxkata` `<SKIP>` entries in `parts/common/components.json`), so the containerd *version* is decided by the image, while the *template* is decided by the distro (`IsContainerdV2Distro` short-circuits to the v1 template for every Kata distro). These disagree today: AzureLinux V3 Kata boots containerd 2.x while being handed a containerd 1.x style config.

The E2E run shows that combination currently works, because containerd 2.x migrates the legacy `io.containerd.grpc.v1.cri` runtime handlers onto the current `io.containerd.cri.v1.runtime` paths, and raises no warnings doing so. So this is a fragility rather than a live bug. Step 2 pins the property we actually care about — after parsing, the handlers are present and containerd was not unhappy — so that if that migration behaviour ever changes, it is caught here rather than in production.

**Scope and known gaps**

- **`kata-cc`** is intentionally not covered: its handler block is templated for all Kata VHDs, but it targets a different VHD than regular Kata, so this image cannot run it.
- **`kata-preview`** is intentionally not covered, as the change adding it to the containerd config was reverted. The validators and scenario are driven by the `kataRuntimeHandlers` list, so re-adding it is a one-line change. Note that re-adding it will also need VHD-side artifacts (`configuration-clh-templating.toml`, `mkfs.erofs`) present in the Kata VHD that E2E resolves; see the note on VHD/code skew below.
- **The aks-node-controller containerd templates (`aks-node-controller/parser/templates/*.gtpl`) are not exercised.** This scenario renders from `baker.go`'s inline templates. Worth flagging for anyone planning to close that gap: `Tags.Scriptless` is referenced in exactly one place (`test_helpers.go:307`), where it only guards a redundant re-set of `EnableScriptlessNBCCSECmd`, which `RunScenario` already sets for every Linux scenario. Since the aks-node-controller prefers `--nbc-cmd` over `--provision-config` (`aks-node-controller/app.go:636`), simply adding an `AKSNodeConfigMutator` + `Tags.Scriptless` scenario does **not** appear sufficient to reach those templates. The `IsKata` mapping above is the groundwork, but the harness likely needs a way to suppress `EnableScriptlessNBCCSECmd` first.
- **VHD/code skew**: E2E compiles baker and aks-node-controller from the PR (`replace` directives in `e2e/go.mod`), but resolves node images from the SIG. With no `VHD_BUILD_ID` it uses the latest `branch=refs/heads/main` image, so config-template changes are validated immediately while VHD-content changes are not. The Linux VHD PR gate does not build a Kata VHD, and its E2E stage sets `IgnoreScenariosWithMissingVhd: true`, so this scenario skips there rather than failing.

Each `RuntimeClass` is named after its handler and pinned to the scenario's node via `Scheduling.NodeSelector`, since `RuntimeClass` is cluster-scoped and e2e scenarios run in parallel against a shared cluster.

**Notes for reviewers**

- The scenario uses `Standard_D4ds_v5` (nested virtualization plus headroom over the 2-vCPU e2e default) and deliberately leaves unattended upgrades enabled so CSE's Kata-specific opt-out branch (`cse_main.sh`) is genuinely exercised rather than passing vacuously.
- **Related pre-existing bug, documented via TODO rather than fixed here**: `ValidateContainerd2Properties` runs `containerd config dump` through `execOnVMForScenarioOnUnprivilegedPod`, i.e. inside the `debugnonhost` daemonset pod. That pod is a bare CBL-Mariner base image with no volume mounts, so `containerd` is unreachable and the command exits 127. Since that validator never checks the exit code and `attemptExecOnPod` returns no error for non-zero exits, its `level=warning` assertion has been passing vacuously at all 11 call sites.

N/A

### DIFF
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -414,7 +414,7 @@ func Test_AzureLinuxV3(t *testing.T) {
 // Test_AzureLinuxV3Gen2Kata verifies that AgentBaker correctly bootstraps a Kata-enabled node.
 //
 // Kata Containers is a runtime, so the thing that can silently break is the containerd
-// configuration: pkg/agent/baker.go only emits the `kata` and `kata-cc` runtime handler blocks
+// configuration: pkg/agent/baker.go only emits the `kata` runtime handler blocks
 // when the agent pool's Distro satisfies Distro.IsKataDistro(). Selecting a Kata VHD here flows
 // through e2e/node_config.go -> AgentPoolProfile.Distro -> the IsKata template func, so this
 // scenario exercises that whole path against a real node.
@@ -425,7 +425,7 @@ func Test_AzureLinuxV3(t *testing.T) {
 //  3. a pod scheduled via a Kata RuntimeClass runs and is genuinely VM-isolated.
 func Test_AzureLinuxV3Gen2Kata(t *testing.T) {
 	RunScenario(t, &Scenario{
-		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with working kata and kata-cc containerd runtime handlers, and can run a VM-isolated pod via a Kata RuntimeClass",
+		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with a working kata containerd runtime handler, and can run a VM-isolated pod via a Kata RuntimeClass",
 		Tags: Tags{
 			Kata: true,
 		},
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -422,11 +422,11 @@ func Test_AzureLinuxV3(t *testing.T) {
 // The scenario asserts three increasingly strong properties:
 //  1. the rendered /etc/containerd/config.toml contains the Kata runtime handlers,
 //  2. containerd actually parsed and loaded them (no warnings, handlers in `config dump`),
-//  3. a pod scheduled via a Kata RuntimeClass runs and is genuinely VM-isolated, for both the
-//     `kata` and `kata-preview` runtime handlers.
+//  3. for every handler in kataRuntimeHandlers, a pod scheduled via a Kata RuntimeClass runs
+//     and is genuinely VM-isolated.
 func Test_AzureLinuxV3Gen2Kata(t *testing.T) {
 	RunScenario(t, &Scenario{
-		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with working kata and kata-preview containerd runtime handlers, and can run VM-isolated pods via Kata RuntimeClasses",
+		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with working kata containerd runtime handlers, and can run VM-isolated pods via Kata RuntimeClasses",
 		Tags: Tags{
 			Kata: true,
 		},
@@ -449,8 +449,9 @@ func Test_AzureLinuxV3Gen2Kata(t *testing.T) {
 				ValidateKataContainerdConfig(ctx, s)
 				ValidateKataContainerdConfigDump(ctx, s)
 				ValidateKataHostReadiness(ctx, s)
-				ValidateKataPodIsIsolated(ctx, s, "kata")
-				ValidateKataPodIsIsolated(ctx, s, "kata-preview")
+				for _, handler := range kataRuntimeHandlers {
+					ValidateKataPodIsIsolated(ctx, s, handler)
+				}
 			},
 		},
 	})
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -422,10 +422,11 @@ func Test_AzureLinuxV3(t *testing.T) {
 // The scenario asserts three increasingly strong properties:
 //  1. the rendered /etc/containerd/config.toml contains the Kata runtime handlers,
 //  2. containerd actually parsed and loaded them (no warnings, handlers in `config dump`),
-//  3. a pod scheduled via a Kata RuntimeClass runs and is genuinely VM-isolated.
+//  3. a pod scheduled via a Kata RuntimeClass runs and is genuinely VM-isolated, for both the
+//     `kata` and `kata-preview` runtime handlers.
 func Test_AzureLinuxV3Gen2Kata(t *testing.T) {
 	RunScenario(t, &Scenario{
-		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with a working kata containerd runtime handler, and can run a VM-isolated pod via a Kata RuntimeClass",
+		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with working kata and kata-preview containerd runtime handlers, and can run VM-isolated pods via Kata RuntimeClasses",
 		Tags: Tags{
 			Kata: true,
 		},
@@ -448,7 +449,8 @@ func Test_AzureLinuxV3Gen2Kata(t *testing.T) {
 				ValidateKataContainerdConfig(ctx, s)
 				ValidateKataContainerdConfigDump(ctx, s)
 				ValidateKataHostReadiness(ctx, s)
-				ValidateKataPodIsIsolated(ctx, s)
+				ValidateKataPodIsIsolated(ctx, s, "kata")
+				ValidateKataPodIsIsolated(ctx, s, "kata-preview")
 			},
 		},
 	})
--- a/e2e/scenario_test.go
+++ b/e2e/scenario_test.go
@@ -411,6 +411,52 @@ func Test_AzureLinuxV3(t *testing.T) {
 	})
 }
 
+// Test_AzureLinuxV3Gen2Kata verifies that AgentBaker correctly bootstraps a Kata-enabled node.
+//
+// Kata Containers is a runtime, so the thing that can silently break is the containerd
+// configuration: pkg/agent/baker.go only emits the `kata` and `kata-cc` runtime handler blocks
+// when the agent pool's Distro satisfies Distro.IsKataDistro(). Selecting a Kata VHD here flows
+// through e2e/node_config.go -> AgentPoolProfile.Distro -> the IsKata template func, so this
+// scenario exercises that whole path against a real node.
+//
+// The scenario asserts three increasingly strong properties:
+//  1. the rendered /etc/containerd/config.toml contains the Kata runtime handlers,
+//  2. containerd actually parsed and loaded them (no warnings, handlers in `config dump`),
+//  3. a pod scheduled via a Kata RuntimeClass runs and is genuinely VM-isolated.
+func Test_AzureLinuxV3Gen2Kata(t *testing.T) {
+	RunScenario(t, &Scenario{
+		Description: "Tests that an AzureLinuxV3 Gen2 Kata node is bootstrapped with working kata and kata-cc containerd runtime handlers, and can run a VM-isolated pod via a Kata RuntimeClass",
+		Tags: Tags{
+			Kata: true,
+		},
+		Config: Config{
+			Cluster: ClusterKubenet,
+			VHD:     config.VHDAzureLinuxV3Gen2Kata,
+			BootstrapConfigMutator: func(_ *Cluster, nbc *datamodel.NodeBootstrappingConfiguration) {
+				// Kata launches each pod inside its own VM, which requires nested virtualization
+				// and more headroom than the 2-vCPU e2e default (config.Config.DefaultVMSKU).
+				nbc.ContainerService.Properties.AgentPoolProfiles[0].VMSize = kataVMSize
+				nbc.AgentPoolProfile.VMSize = kataVMSize
+				// Leave unattended upgrades on so that CSE's kata-specific opt-out branch is
+				// actually exercised, which ValidateKataHostReadiness asserts.
+				nbc.DisableUnattendedUpgrades = false
+			},
+			VMConfigMutator: func(vmss *armcompute.VirtualMachineScaleSet) {
+				vmss.SKU.Name = to.Ptr(kataVMSize)
+			},
+			Validator: func(ctx context.Context, s *Scenario) {
+				ValidateKataContainerdConfig(ctx, s)
+				ValidateKataContainerdConfigDump(ctx, s)
+				ValidateKataHostReadiness(ctx, s)
+				ValidateKataPodIsIsolated(ctx, s)
+			},
+		},
+	})
+}
+
+// kataVMSize is a nested-virtualization capable SKU with enough capacity to host a Kata guest VM.
+const kataVMSize = "Standard_D4ds_v5"
+
 func Test_AzureLinuxV3_CustomCA(t *testing.T) {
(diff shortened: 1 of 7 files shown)


Camelron (approved):
LGTM (from kata side at least). I can add on tests for VM templating and snapshot/restore in the future.

pkg/agent/datamodel/types.go
```diff
@@ -252,6 +252,7 @@ var AKSDistrosAvailableOnVHD = []Distro{
 	AKSAzureLinuxV3Gen2FIPS,
 	AKSCBLMarinerV2Gen2Kata,
 	AKSAzureLinuxV2Gen2Kata,
+	AKSAzureLinuxV3Gen2Kata,
```
cameronmeissner: bad change here

Bickor: Removed.

Bickor:
The aks-node-controller containerd templates appear to be unreachable from E2E — can someone confirm?

There are two renderers for /etc/containerd/config.toml: pkg/agent/baker.go's inline templates, and aks-node-controller/parser/templates/*.gtpl. This scenario only covers the former, and I couldn't find a way to reach the latter.

aks-node-controller/app.go:636 prefers --nbc-cmd over --provision-config ("if NBC command is provided, we prioritize it over the aks node config"). The --nbc-cmd file is written whenever EnableScriptlessNBCCSECmd is true, and RunScenario (test_helpers.go:96) sets that for every non-Windows scenario.

Tags.Scriptless doesn't change this. It's referenced in exactly one place:
// test_helpers.go:307
if !s.Tags.Scriptless && s.BootstrapConfigMutator != nil {
    nbc.EnableScriptlessNBCCSECmd = true   // already set true at line 282
}

It only guards a redundant re-set of a flag that's already true. So a Tags.Scriptless: true + AKSNodeConfigMutator scenario ships a config.json that ANC then ignores — it still provisions from the NBC path, while looking like it tested the scriptless one.

I read this as the intended "scriptless phase 2.5" state (ship both, provision from NBC), which would mean the .gtpl containerd templates currently have no E2E coverage at all. Closing that likely needs a harness switch to suppress EnableScriptlessNBCCSECmd per scenario. Am I missing an existing mechanism? If not, this is worth a separate work item — it also means the wrong-plugin-path concern in containerd_v2.toml.gtpl can't be validated today.

This PR adds the groundwork either way: nbcToAKSNodeConfigV1 now derives Configuration.IsKata from the distro, so a scriptless Kata scenario gets a correct config once the path is reachable.

Bickor:
## How to run this scenario today

Posting this so someone can independently confirm the results, and so the Kata build/test loop is written down — it's not obvious.

### 1. What already runs automatically

`Test_AzureLinuxV3Gen2Kata` is picked up by the **Agentbaker E2E** pipeline on every push to this PR. `.pipelines/e2e.yaml` has a `pr:` trigger with a path filter on `e2e/**`, and this PR is `e2e/`-only, so no extra action is needed.

That run does **not** set `VHD_BUILD_ID`, so it resolves the latest `branch=refs/heads/main`-tagged image from the staging gallery (image definition `AzureLinuxV3katagen2`). Note the split this creates:

- **baker / aks-node-controller code comes from this PR** — `e2e/go.mod` has `replace github.com/Azure/agentbaker => ../`, so `/etc/containerd/config.toml` is rendered from the PR's templates.
- **the node image comes from `main`** — so VHD-content changes are *not* covered.

Result from build [175521223](https://msazure.visualstudio.com/09706533-03bf-4b43-9a9b-b49c75429646/_build/results?buildId=175521223) (before the latest fix), which is what I'd like confirmed:

```
creating pod "...-kata-pod" under RuntimeClass "kata-pl45-..."
✓ waiting for pod ... done (18.0s)
host kernel: "6.6.137.mshv2-1.azl3", kata guest kernel: "6.6.137.mshv1-1.azl3"
```

Node bootstrapped, the `kata` handler was registered with CRI, the pod ran under the Kata RuntimeClass, and the guest kernel differs from the host — genuine VM isolation. The only failure was an assertion bug on my side (`containerd config dump` emits single-quoted TOML, the config file uses double quotes), fixed in the latest commit.

### 2. Running just this scenario locally

```bash
cd e2e
TAGS_TO_RUN="name=Test_AzureLinuxV3Gen2Kata" ./e2e-local.sh
```

To pin a specific VHD build instead of latest-from-main:

```bash
SIG_VERSION_TAG_NAME=buildId SIG_VERSION_TAG_VALUE=<buildId> \
  TAGS_TO_RUN="name=Test_AzureLinuxV3Gen2Kata" ./e2e-local.sh
```

### 3. Testing against a Kata VHD built from a branch

Needed for anything VHD-side — most immediately, re-adding `kata-preview`, which depends on `configuration-clh-templating.toml` and `mkfs.erofs` being present in the image. Those cannot be validated against a `main` VHD until they ship there.

Manually queue **`.pipelines/.vsts-vhd-builder-release.yaml`** against the branch (it is `trigger: none`, so manual-queue only) and enable **only** the Kata job:

- set parameter `buildAzureLinuxV3gen2kata` = **true**
- set every other `build*` parameter = **false**

That single run closes the loop:

1. Packer tags the resulting SIG image version with the ADO build ID (`vhdbuilder/packer/vhd-image-builder-mariner.json:49`, `"buildId": "{{user 'build_id'}}"`).
2. The pipeline's own `e2e` stage (line 938) runs with `VHD_BUILD_ID: $(Build.BuildId)`, which `.pipelines/scripts/e2e_run.sh:83` converts into `SIG_VERSION_TAG_NAME=buildId` / `SIG_VERSION_TAG_VALUE=<build>`.
3. Because only the Kata VHD carries that `buildId` and the stage sets `IgnoreScenariosWithMissingVhd: true`, every other scenario skips — you get a Kata-only E2E run against a Kata VHD built from your branch.

So a `kata-preview` change can put the VHD artifacts and the template change on one branch and validate both together, without waiting for anything to land on `main`.

> Caveat: I've verified this wiring by reading the YAML but have not queued the pipeline myself — it may have branch policies or approvals restricting non-`main` runs. If it refuses the branch, any manual VHD build works as long as the image lands in the staging gallery tagged `buildId=<n>`; pass that as `VHD_BUILD_ID`.

### 4. Gap worth knowing about

The Linux VHD PR gate (`.pipelines/.vsts-vhd-builder.yaml`) **builds no Kata VHD** (no `kata` job in it), and its E2E stage sets `IgnoreScenariosWithMissingVhd: true`. So on a PR touching `vhdbuilder/**` or `parts/**`, this scenario **skips silently** — VHD-side Kata changes get no pre-merge E2E coverage at all today, and are only exercised once a release Kata VHD reaches `main`.

Adding a Kata job to that gate, ideally behind an opt-in pipeline parameter so it doesn't add build time to every VHD PR, would close that. Happy to file it separately if people agree it's worth doing.

e2e/scenario_test.go
```diff
@@ -411,6 +411,55 @@ func Test_AzureLinuxV3(t *testing.T) {
...
+				for _, handler := range kataRuntimeHandlers {
+					ValidateKataPodIsIsolated(ctx, s, handler)
+				}
+			},
+		},
+	})
+}
+
+// kataVMSize is a nested-virtualization capable SKU with enough capacity to host a Kata guest VM.
+const kataVMSize = "Standard_D4ds_v5"
```
awesomenix, line 461: Did you try the default `Standard_D2ds_v5` without any overrides?

awesomenix (approved):

Bickor:
### Correction on my earlier "aks-node-controller is untestable" comment

I was wrong, thanks for pushing back. I traced `app.go:639` (`ProvisionConfig != "" && NBCCmd == ""`) and concluded the `.gtpl` templates never execute. I stopped one function short of `app.go:660`:

```go
// If both flags are provided, compare environments.
if flags.ProvisionConfig != "" && flags.NBCCmd != "" {
    compareEnvs(ctx, flags, a.eventLogger, a.getGPUComponentsFilePath())
}
```

`compareEnvs` builds the command from the provision-config, which renders the `.gtpl` templates and produces `CONTAINERD_CONFIG_CONTENT`, then `diffEnvMaps` compares every CSE env var against the NBC path. Only six keys are exempt (`isExpectedDiffCSEVar`) and the containerd config is not one of them; `ValidateScriptlessPhase3` (`validation.go:60`) fails the test on any difference.

So the accurate statement is narrower: aks-node-controller does not *provision* from the `.gtpl` output today, but that output **is** rendered and byte-compared against baker's for any scenario setting `AKSNodeConfigMutator`. I've dropped the "unreachable" claim from the PR description.

### The remaining Kata-specific gap

Adding `AKSNodeConfigMutator` to this scenario would pull it into that comparison for free. I'd expect it to **fail**, for a real reason: the two generators disagree on which containerd template to use for Kata.

- **baker** picks by **distro** — `IsContainerdV2Distro()` short-circuits to `false` for every Kata distro (`pkg/agent/datamodel/types.go:1200`), so Kata always renders the **v1** template.
- **aks-node-controller** picks by the **containerd version detected on the node** (`parser/helper.go:246`). AzureLinux V3 Kata ships containerd 2.x, so it renders the **v2** `.gtpl`.

Different template, different `CONTAINERD_CONFIG_CONTENT`, so the phase-3 env comparison should report a diff.

This is invisible today only because no Kata scenario sets `AKSNodeConfigMutator`, so the comparison never runs for Kata. It matters when aks-node-controller becomes the provisioning path: Kata nodes would switch to the v2 `.gtpl`, whose Kata block declares handlers under the legacy `io.containerd.grpc.v1.cri` path while the rest of that file uses `io.containerd.cri.v1.runtime`. That happens to work now because containerd 2.x migrates the legacy paths (confirmed in build 175521223 — the effective config shows the handlers under `io.containerd.cri.v1.runtime` with no warnings), but it is untested and inconsistent.

Keeping it out of this PR so it doesn't land red. Happy to open a follow-up that adds the mutator and either fixes the selection mismatch or documents it — let me know which you'd prefer.

Bickor:
### Question: is the Kata carve-out from containerd v2 still needed?

@Camelron — this came out of adding Kata E2E coverage and I'd like a sanity check from someone who knows the containerd 2.0 rollout.

**The situation.** baker and aks-node-controller each have v1 and v2 containerd templates, and they choose between them differently:

- **baker** chooses by **distro** — `IsContainerdV2Distro()` (`pkg/agent/datamodel/types.go:1200`) short-circuits to `false` for every Kata distro, so Kata always renders the **v1** template.
- **aks-node-controller** chooses by the **containerd version detected on the node** (`aks-node-controller/parser/helper.go:246`).

For a non-Kata AzureLinux V3 node both pick v2 and agree. For a **Kata** AzureLinux V3 node, baker is forced to v1 while ANC detects containerd 2.x and picks v2 — so the two generators produce different `CONTAINERD_CONFIG_CONTENT`.

That would normally be caught by the phase-3 env comparison (`compareEnvs` -> `ValidateScriptlessPhase3`), but that only runs for scenarios setting `AKSNodeConfigMutator`, and no Kata scenario does. So for Kata the comparison has never executed and the mismatch is unobserved.

**What we know empirically**, from the E2E run on this PR ([175521223](https://msazure.visualstudio.com/09706533-03bf-4b43-9a9b-b49c75429646/_build/results?buildId=175521223)) against an `AzureLinuxV3katagen2` node:

- the node runs **containerd 2.x** (`version = 3`, `io.containerd.cri.v1.*` plugins)
- it was handed baker's **v1**-style config
- containerd migrated the legacy `io.containerd.grpc.v1.cri` Kata handlers onto `io.containerd.cri.v1.runtime` automatically, with **no `level=warning`**
- a pod under a `kata` RuntimeClass ran and was VM-isolated (guest kernel `6.6.137.mshv1-1.azl3` vs host `6.6.137.mshv2-1.azl3`)

So it works today purely because containerd 2.x tolerates the legacy paths. It's also worth noting the v2 `.gtpl` is internally inconsistent on this point: its runc handlers use `io.containerd.cri.v1.runtime` while its Kata block still uses `io.containerd.grpc.v1.cri` (`containerd_v2.toml.gtpl:56`).

**The question.** The Kata short-circuit was added in #8825 alongside `IsContainerdV2Distro` itself, so I assume it was a deliberate "hold Kata back during the containerd 2.0 rollout" guard rather than an accident. Is that still required?

Concretely: **is there any harm in letting Kata distros use the v2 template, so both generators agree?** The Kata VHD already ships containerd 2.x, and containerd is already effectively running the v2 form of the config. If there's a reason Kata must stay on v1 (kata-cc, the tardev snapshotter, shim expectations, something in the VHD), I'd rather know than guess.

**Why it matters now.** As long as baker provisions, this is invisible. When aks-node-controller takes over provisioning, Kata nodes switch to the v2 `.gtpl` — a config nothing has validated for Kata — and they'd stop relying on baker's v1 output.

I've kept this out of the current PR to avoid landing red. The one-line follow-up is adding `AKSNodeConfigMutator` to `Test_AzureLinuxV3Gen2Kata`, which pulls Kata into the phase-3 comparison and would surface the mismatch. Happy to open that either as a failing test documenting the issue, or alongside a fix that aligns the two selections — whichever you prefer.

PR #9159, opened 2026-08-07
fix: improved error handling on windows scripts baked into VHD

timmy-wright:

Improved error handling on scripts baked into windows VHDs.

Fixes #


timmy-wright:
## Review: improved error handling on Windows CSE scripts

Reviewed with multiple angle passes (architecture/simplification, security, reliability, regression, performance) plus a targeted pass on idempotency risk from the new fatal error paths. Security and performance came back clean — no material findings. Two issues below should block merge; one is a non-blocking design note.

### 🔴 Blocking: new absolute dot-source breaks CI/local unit tests

`configfunc.ps1:1`, `containerdfunc.ps1:1`, `kubeletfunc.ps1:1`, and `all.ps1:3` all add `. c:\AzureData\windows\helpers.ps1` (hardcoded absolute path). That path only exists once bootstrapped on a real node — it doesn't exist when Pester loads these scripts directly from the repo.

Repro:
```
pwsh -NoProfile -NonInteractive -Command "$ErrorActionPreference='Stop'; . 'staging/cse/windows/configfunc.ps1'"
```
fails immediately at line 1 because the path is absent. Running the actual Pester suite for `configfunc.tests.ps1` produces 0/17 passing for the same reason. `containerdfunc.tests.ps1` was patched to pre-load helpers via `. $PSScriptRoot\helpers.ps1`, but that doesn't help — `containerdfunc.ps1`'s own absolute-path line still fires and throws regardless. `configfunc.tests.ps1` and `kubeletfunc.tests.ps1` weren't patched at all. This matches the PR's own CI comment (124/393 tests failing, 127 failed runs) exactly.

**Fix:** use `. $PSScriptRoot\helpers.ps1` in all four files instead of the hardcoded absolute path. `windows_package_cse.sh` copies `staging/cse/windows/*` flat into the deployed package, so `$PSScriptRoot` resolves correctly both in-repo and on a provisioned node.

### 🔴 Conditional regression: re-invocation after a partial failure can now hard-fail where it previously succeeded

`helpers.ps1`'s new `Invoke-Nssm` throws on any non-zero `nssm.exe` exit code — including `nssm.exe install <service>` when the service already exists. Previously that exit code was never checked, so a subsequent `nssm.exe set ...` would still reconfigure the already-existing service and provisioning proceeded. Affected installs (no idempotency/cleanup before `install`):
- `configfunc.ps1:550` (`New-CsiProxyService`) and `configfunc.ps1:573` (`New-HostsConfigService`)
- `kubeletfunc.ps1:367` (Kubelet) and `kubeletfunc.ps1:394` (Kubeproxy)
- `containerdfunc.ps1:48`, immediately following a *best-effort* `sc.exe delete containerd` at `:43` — the delete failing/racing can still leave the service present when `install` runs

The template has no built-in CSE retry loop, but it only guards re-entry via `$CSEResultFilePath` (`kuberneteswindowssetup.ps1.template:686-689,738-741`) — so an external re-invocation after a partial BasePrep failure will revisit these functions against a VM that already has some services registered from the prior attempt, and would now hard-fail instead of proceeding. The new Pester tests mock `Invoke-Nssm` directly, so they don't exercise this install-conflict path.

**Fix:** make each service registration idempotent — best-effort remove/reconfigure an existing service before `install` (mirroring the existing `sc.exe delete containerd` exemption), or explicitly tolerate the "service already exists" case and continue to the `set` calls. Worth adding a test that asserts behavior when `Invoke-Nssm install` hits an already-registered service.

### 🟡 Non-blocking design note: `Invoke-Nssm` duplicates `Invoke-Executable`

`windowscsehelper.ps1` already has `Invoke-Executable` (supports `AllowedExitCodes`, retries, and funnels failures through `Set-ExitCode`), and `configfunc.ps1` already depends on it today pre-PR. Worth considering reuse instead of a parallel `Invoke-Nssm`/manual `$LASTEXITCODE` pattern — but it's not a trivial substitution: `Invoke-Executable`'s failure path calls `Set-ExitCode`, which exits the process immediately rather than throwing, so call sites that currently catch `Invoke-Nssm`'s `throw` to assign a specific granular exit code (e.g. the GMSA registry-import error) would need restructuring. Not a merge blocker, just a follow-up worth scoping separately.

timmy-wright:
## Update: both blocking issues addressed

**Dot-source fix (already pushed separately):** `configfunc.ps1`, `containerdfunc.ps1`, `kubeletfunc.ps1`, and `all.ps1` now use `. $PSScriptRoot\helpers.ps1` instead of the hardcoded absolute path, resolving the CI/Pester failures.

**Idempotency regression fix (this push):** Addressed the conditional regression where `Invoke-Nssm` now throws on "service already exists" for a service left over from a prior partial provisioning attempt:

- Added a shared `Remove-ServiceIfExists` helper in `helpers.ps1`, generalizing the best-effort `sc.exe delete` cleanup pattern already used in `RegisterContainerDService` (best-effort: tolerates non-zero exit, e.g. 1072 "marked for deletion").
- Refactored `RegisterContainerDService` to use the shared helper instead of its inline duplicate.
- Applied the same best-effort pre-delete before `nssm.exe install` in `New-CsiProxyService`, `New-HostsConfigService` (configfunc.ps1), and `New-NSSMService` for Kubelet/Kubeproxy (kubeletfunc.ps1) — the four call sites that previously lacked this protection.
- Added Pester regression tests mirroring the existing `RegisterContainerDService` "service already exists" tests for all four sites plus `Remove-ServiceIfExists` directly (`helpers.tests.ps1`, new file).

Validated locally via `pwsh`/Pester (7.6.4): compared pass/fail counts before and after each commit to confirm zero regressions — all pre-existing local failures are unrelated Linux-sandbox limitations (missing Windows-only cmdlets like `Get-Service`/`Start-Service`, and a Pester 6.0.1 `Assert-MockCalled` compatibility quirk that predates this PR).

Commits: 0caf58a8a7, 4da773f0a8, a6d8e7ceca, d39c42b05a, b2eea5ba4e.

PR #9165, opened 2026-08-07
fix(e2e): reconcile IP connect on the existing shared bastion

ganeshkumarashok:
## What

`ensureSharedBastion` reconciles `enableIpConnect` on the **existing** shared bastion, not just on the one it creates.

## Why

#9149 enabled IP connect, but only in the creation path. `ensureSharedBastion` returns early when the bastion already exists:

```go
existing, err := config.Azure.BastionHosts.Get(ctx, rg, SharedBastionName, nil)
if err == nil {
    ...
    return *existing.Properties.DNSName, nil   // <- creation properties never applied
}
```

The shared bastion in the e2e subscription was created before #9149, so it kept `enableIpConnect` unset and #9149 never reached it. Confirmed on the live resource:

```
$ az network bastion show -n abe2e-shared-bastion -g abe2e-westus2 \
    --query "{sku:sku.name,tunnel:enableTunneling,ipconnect:enableIpConnect}"
{ "sku": "Standard", "tunnel": true, "ipconnect": null }
```

Every SSH attempt therefore failed, 5/5 retries, on every scenario:

```
failed to start bastion tunnel: error creating tunnel: 403
```

`bastionssh.go` builds `.../providers/Microsoft.Network/bh-hostConnect/<ip>`, which is Bastion's **IP-based** connect flow. Bastion rejects it with 403 unless IP connect is on. Bastion itself was healthy the whole time — the resource-based flow (`az network bastion tunnel --target-resource-id ...`) connected fine.

## Verification

I set the flag out of band on the shared bastion and re-ran `Test_Ubuntu2204_GPUA100` locally. SSH succeeded on **attempt 1/5** instead of failing 5/5, and validation proceeded through the kubelet, NIC, kernel and waagent validators:

```
bastionssh.go:304: Attempt 1/5 establishing SSH over bastion to 10.99.32.5
test_helpers.go:1032: SSH connectivity to 10.99.32.5 verified successfully
```

This change makes that self-healing so the fix survives without manual intervention. The update is a no-op once the flag is set, so it costs nothing on subsequent runs.


ganeshkumarashok:
Confirmed working: GPU E2E build 175689787 on this branch shows **0** bastion 403s (baseline 175585765 on `main` had 252).

Note this unmasks a pre-existing failure — all 12 GPU scenario attempts now fail on `[nvidia-cdi-refresh.path nvidia-cdi-refresh.service]` entering a failed state, with 0 `VMExtensionProvisioningError`. The validator simply could not reach the node before. That is fixed by #9150, not by this PR; the two should land together.

awesomenix (approved):

ganeshkumarashok:
Updated the branch onto `main` to re-run the E2E gates.

The three red gates were all from build set 175689775 / 175689787 / 175689780 on **2026-08-07**, which predates #9150 (merged 2026-08-11). Per my earlier note, the remaining GPU failures on that run were the `nvidia-cdi-refresh.path / nvidia-cdi-refresh.service` failed-state assertions that #9150 fixes, not anything from this change.

The other two gates on that run were unrelated infra flake, for the record:
- **Linux E2E** (175689775): 18x `proxy error from localhost:9443 while dialing <node>:10250, code 500` — apiserver→kubelet exec tunnel on `abe2e-kubenet-v5-150ee`, not a node-bootstrap failure.
- **GPU E2E** (175689787): `VMOperationFailedOnHost` — Azure host-level allocation error during VMSS create.
- **Windows E2E** (175689780): servercore test pod never reached Running, `context deadline exceeded`.

This PR itself remains validated: 0 bastion 403s on this branch vs 252 on the `main` baseline (175585765).

PR #9167, opened 2026-08-09
test(e2e): use the default VM SKU for kata and build the kata VHD in the PR gate

Bickor:

Follow-up to #9142. Both changes were requested in review there but landed on the branch after it was merged, so they missed the merge.

**1. Use the default VM SKU for the Kata scenario**

`Test_AzureLinuxV3Gen2Kata` pinned `Standard_D4ds_v5`. Reverted to `config.Config.DefaultVMSKU` to avoid consuming extra quota; the default is nested-virtualization capable, which is what Kata requires. This removes the `kataVMSize` constant along with the `VMConfigMutator` and the two `VMSize` assignments that existed only to apply it.

**2. Build the Kata VHD in the Linux VHD PR check-in gate**

`.pipelines/.vsts-vhd-builder.yaml` builds no Kata image today. That means no Kata VHD carries the gate build's `buildId` tag, and since its E2E stage sets `IgnoreScenariosWithMissingVhd: true`, `Test_AzureLinuxV3Gen2Kata` **silently skips** there. VHD-side Kata changes (kata packages, erofs tooling, kata configuration files) therefore get no pre-merge E2E coverage at all — they are only exercised once a release Kata VHD reaches `main`.

Adds `buildAzureLinuxV3gen2kata`, mirroring the job in `.vsts-vhd-builder-release.yaml` (`OS_VERSION=V3kata`, `FEATURE_FLAGS=kata`, `Standard_D16ads_v5`). With this, a Kata image carries the gate's `buildId`, the E2E stage resolves it via `VHD_BUILD_ID`, and the scenario runs against a VHD built from the PR.

**Notes for reviewers**

- This adds a job with a 360-minute timeout to every PR matching the gate's path filter (`vhdbuilder/packer`, `parts/linux/*`, `aks-node-controller/**`, `parts/common/components.json`, `go.mod`, ...). If that cost is higher than wanted, it is a small change to put it behind a pipeline parameter defaulting to false — happy to do that instead.
- On the SKU change: the Kata pod came up in 18s on 4 vCPU / 16 GB in the #9142 E2E run. The default SKU gives the guest VM less headroom, so if the Kata pod starts flaking this is the first thing to revisit.
- Kept as a draft until the new VHD job is confirmed working in this PR's own gate run, since this PR touches `.pipelines/**` and should trigger it.

N/A

### DIFF
--- a/.pipelines/.vsts-vhd-builder.yaml
+++ b/.pipelines/.vsts-vhd-builder.yaml
@@ -183,6 +183,32 @@ stages:
           parameters:
             artifactName: azurelinuxv3-gen2
 
+    # Kata VHDs are built here so that VHD-side Kata changes (kata packages, erofs tooling,
+    # kata configuration files) are covered by the E2E stage below. Without this job no Kata
+    # image carries this build's buildId tag, so Test_AzureLinuxV3Gen2Kata silently skips
+    # (the E2E stage sets IgnoreScenariosWithMissingVhd: true).
+    - job: buildAzureLinuxV3gen2kata
+      timeoutInMinutes: 360
+      steps:
+        - bash: |
+            echo '##vso[task.setvariable variable=OS_SKU]AzureLinux'
+            echo '##vso[task.setvariable variable=OS_VERSION]V3kata'
+            echo '##vso[task.setvariable variable=IMG_PUBLISHER]MicrosoftCBLMariner'
+            echo '##vso[task.setvariable variable=IMG_OFFER]azure-linux-3'
+            echo '##vso[task.setvariable variable=IMG_SKU]azure-linux-3-gen2'
+            echo '##vso[task.setvariable variable=IMG_VERSION]latest'
+            echo '##vso[task.setvariable variable=HYPERV_GENERATION]V2'
+            echo '##vso[task.setvariable variable=AZURE_VM_SIZE]Standard_D16ads_v5'
+            echo '##vso[task.setvariable variable=FEATURE_FLAGS]kata'
+            echo '##vso[task.setvariable variable=ARCHITECTURE]X86_64'
+            echo '##vso[task.setvariable variable=ENABLE_FIPS]false'
+            echo '##vso[task.setvariable variable=ENABLE_TRUSTED_LAUNCH]False'
+            echo '##vso[task.setvariable variable=ENABLE_CGROUPV2]True'
+          displayName: Setup Build Variables
+        - template: ./templates/.builder-release-template.yaml
+          parameters:
+            artifactName: azurelinuxv3-gen2-kata
+
     - job: buildAzureLinuxV3ARM64gen2fips
       timeoutInMinutes: 360
       steps:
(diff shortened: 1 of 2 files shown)


Bickor:
## Verification results

Both changes verified end to end. Summary of the runs so far.

### The new Kata VHD job works

`buildAzureLinuxV3gen2kata` built successfully in the gate (35m53s), and the E2E stage then resolved it via `VHD_BUILD_ID` and ran the scenario **against the Kata VHD built from this PR**:

```
VHD_BUILD_ID: 175846987
VHD_BUILD_ID is specified (175846987). Running tests using VHDs from that build
...
✓ Azure linux V3 gen 2 kata (257.19s)
```

Before this change no Kata image carried the gate's `buildId`, so the scenario silently skipped there (`IgnoreScenariosWithMissingVhd: true`). That gap is now closed — a `kata-preview` change carrying both VHD artifacts and template changes can be validated in a single PR.

### The default VM SKU is fine

No regression from dropping `Standard_D4ds_v5`. Kata scenario timings across every run:

| Run | Pipeline | Kata result |
|---|---|---|
| 175846990 | Agentbaker E2E (main-tagged VHD) | ✓ 245.31s |
| 175860034 | Agentbaker E2E (main-tagged VHD) | ✓ 308.31s |
| 175846987 | VHD gate e2e (PR-built VHD) | ✓ 257.19s |
| 175860033 attempt 1 | VHD gate e2e (PR-built VHD) | ✓ 253.89s |
| 175860033 attempt 2 | VHD gate e2e (PR-built VHD) | ✓ 286.50s |

Comparable to the 249.31s it took on `D4ds_v5` in #9142. Kata is 5/5 green.

### About the red gate check

`Agentbaker E2E` is green. The VHD gate has failed three times, each with a **different** unrelated test and none involving Kata (`E2E_FAILED_TESTS_RETRY_COUNT: 0`, so there is no retry budget to absorb noise):

1. `175846987` — 5 network-isolated scenarios, all failing in ~3s in shared cluster prep: `listing private DNS zones: unexpected end of JSON input` → sibling tasks `context canceled`. Infra flake.
2. `175860033` attempt 1 — `Test_Ubuntu2204ARM64`, VMSS creation exceeded the context deadline at 1022s. Capacity/provisioning flake.
3. `175860033` attempt 2 — `Test_Ubuntu2604Minimal_SecondaryNIC`, kernel-log validator flagged a DRM backtrace at boot:
   ```
   [PANIC/CRASH]:
   [    7.664900] RIP: 0010:drm_atomic_helper_wait_for_vblanks.part.0+0x23a/0x250
   ```

The third one is **not** a flake — it is deterministic image content, so further retries will not clear it. The node otherwise booted fine (kubelet Ready, NIC ethtool checks passed, pod scheduled), so it looks like a benign `WARN_ON` in the hyperv_drm path being matched by the validator's PANIC/CRASH pattern rather than a real panic.

It is also unrelated to this PR: the only changes here are adding a Kata build job to the gate and removing a VM SKU override in a Kata-only scenario, neither of which can affect the Ubuntu 26.04 Minimal kernel. If someone can confirm whether that signature is already known on `main`, that would be good to know — otherwise it likely needs its own item and a decision on whether the kernel-log validator should treat this DRM backtrace as fatal.

awesomenix (approved):

PR #9179, opened 2026-08-11
feat: add reload plugin to LocalDNS Corefile

saewoni:
## Change Summary

Adds the CoreDNS `reload` plugin to generated LocalDNS Corefiles so CoreDNS can reload configuration changes without restarting the `localdns` systemd unit.

- Adds server-block-level `reload` directives to the `aks-node-controller` LocalDNS template.
- Adds the same directives to the legacy `pkg/agent` LocalDNS template so both render paths stay consistent.
- Preserves the existing `hosts` plugin `reload 5s` directives, which watch the hosts file rather than the Corefile.
- Updates the LocalDNS Corefile golden tests and assertions for both render paths.

The CoreDNS defaults are used: a 30-second reload interval with 15 seconds of jitter. Configuration changes are therefore applied asynchronously after the file is updated; restarting `localdns.service` is not required.

## Testing

- `cd aks-node-controller && go test ./parser ./pkg/nodeconfigutils`
- `go test ./pkg/agent ./pkg/agent/datamodel`
- Standalone smoke test on LocalDNS-enabled cluster `ldnsreload2243` from run `176179054`:
  - Verified the bundled CoreDNS binary includes the `reload` plugin.
  - Verified LocalDNS readiness on `169.254.10.10:8181` and `169.254.10.11:8181`.
  - Verified cluster DNS resolution through `169.254.10.11`.
  - Restored the original Corefile and confirmed `localdns.service` remained active.

The standalone test validated the rendered configuration and service health. A follow-up test should update the Corefile while the service remains running and verify that the new configuration takes effect without a service restart.

### DIFF
--- a/aks-node-controller/parser/helper_test.go
+++ b/aks-node-controller/parser/helper_test.go
@@ -1686,6 +1686,7 @@ health-check.localdns.local:53 {
         policy sequential
         max_concurrent 1000
     }
+    reload
     ready 169.254.10.10:8181
     cache 3600 {
         success 9984
@@ -1713,6 +1714,7 @@ cluster.local:53 {
         policy sequential
         max_concurrent 1000
     }
+    reload
     ready 169.254.10.10:8181
     cache 3600 {
         success 9984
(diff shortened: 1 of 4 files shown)


saewoni:
Closing this PR to recreate it from the latest main branch with the same code changes.

PR #9101, opened 2026-07-31
feat: embed Linux hotfix scripts in ANC

Devinwong:
**What this PR does / why we need it**:

Moves selected Linux provisioning-hotfix scripts from CRP CustomData into patched aks-node-controller packages. This avoids making script size part of the long-term 64 KB CustomData constraint.

```mermaid
flowchart TD
    Script["Changed hotfixable script"] --> Assets["Generate embedded payload<br/>and retained CRP fallback"]
    Assets --> Pointer["Write version + scripts_version pointer"]
    Pointer --> Build["Build patched ANC with embedded payload"]
    Build --> Publish["Publish package to PMC"]
    Publish --> ABSvc["ABSvc emits existing pointer<br/>and CRP payload"]

    ABSvc --> CRP["VHD-baked ANC applies CRP scripts"]
    CRP --> Install["Attempt PMC install"]
    Install --> Result{"Install succeeds?"}

    Result -->|Yes| Patched["Patched ANC validates and<br/>applies embedded scripts"]
    Patched --> Provision["Run normal provisioning"]

    Result -->|No| Degraded["Degraded: patched ANC unavailable"]
    Degraded --> Legacy["Provision with CRP-applied scripts"]

    classDef owned fill:#ffedd5,stroke:#ea580c,stroke-width:3px,color:#111827
    classDef shared fill:#dbeafe,stroke:#2563eb,stroke-width:2px,color:#111827
    classDef fallback fill:#dcfce7,stroke:#16a34a,stroke-width:2px,color:#111827
    classDef degraded fill:#fee2e2,stroke:#dc2626,stroke-width:2px,color:#111827

    class Script,Assets,Pointer,Build,Patched owned
    class Publish,ABSvc,Install,Result,Provision shared
    class CRP,Legacy fallback
    class Degraded degraded
```

_Orange: changed by this PR. Blue: existing/shared flow. Green: retained CRP fallback. Red: degraded hotfix outcome._

- Generates an ANC-owned manifest and embeds only selected hotfix scripts without importing the root AgentBaker `parts` module.
  - If there is an active hotfix, `manifest.json` will look like:

    ```json
    {
      "schema_version": 1,
      "entries": [
        {
          "source": "cse_config.sh",
          "embedded_path": "payloads/cse_config.sh",
          "destination": "/opt/azure/containers/provision_configs.sh",
          "mode": "0744",
          "platforms": [
            "ubuntu"
          ]
        }
      ]
    }
    ```

[...]

PR #9182, opened 2026-08-11
feat: add CustomData/protectedSettings size guards for hotfix PRs

abigailliang-aks-sig-node:
## Summary
- Add Go-test guard (`pkg/agent/customdata_size_guard_test.go`) for the scriptless NBC CSE path
- **Guard 1** (`TestCustomDataSizeWithHotfix`): asserts slim Mode B CustomData (P + H) < 87,380
- Add CI step in `hotfix-generate.yml` to run guard after hotfix injection, before commit
- Design doc in `docs/design-customdata-size-guard.md`

**⚠️ Backward compatibility note:** The workflow and test live on `main`. New `official/*` branches fork from `main` weekly, so they inherit the guard automatically. Older `official/*` branches lack it, but those map to VHDs that either already bake the fix (post-0729) or are approaching end of 6-month support. Backporting is not necessary.

## Context
Bridge solution until #9101 (embed hotfix scripts in ANC binary) lands. See design doc for full analysis of encoding chain, Mode A/B architecture, and why only H (not R) can overflow Mode B CustomData.

## Test plan
- [x] Guard passes locally with current (no-hotfix) nodecustomdata.yml: 1,756 / 87,380 bytes
- [x] Validate with a mock hotfix injection that approaches the limit

### DIFF
--- a/.github/workflows/hotfix-generate.yml
+++ b/.github/workflows/hotfix-generate.yml
@@ -78,6 +78,18 @@ jobs:
         run: |
           python3 hotfix/hotfix_generate.py "origin/${{ github.base_ref }}"
 
+      # Runs AFTER hotfix injection (so the //go:embed'd nodecustomdata.yml carries the injected
+      # scripts H) and BEFORE the commit step (so an oversized hotfix fails the PR before its
+      # content is committed).
+      #   Guard 1 (TestCustomDataSizeWithHotfix): slim Mode B CustomData (P + H) < 87,380.
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
+          go-version: '1.25'
+      - name: Validate CustomData size
+        run: |
+          go test ./pkg/agent/ -run TestCustomDataSizeWithHotfix -v -count=1
+
       - name: Commit changes via API
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
--- a/.github/workflows/hotfix-generate.yml
+++ b/.github/workflows/hotfix-generate.yml
@@ -83,7 +83,7 @@ jobs:
       # content is committed).
       #   Guard 1 (TestCustomDataSizeWithHotfix): slim Mode B CustomData (P + H) < 87,380.
       - name: Set up Go
-        uses: actions/setup-go@v7
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
           go-version: '1.25'
       - name: Validate CustomData size
--- a/pkg/agent/customdata_size_guard_test.go
+++ b/pkg/agent/customdata_size_guard_test.go
@@ -18,12 +18,12 @@ import (
 //	In the hotfix CI it runs AFTER hotfix_generate.py injects the real H into nodecustomdata.yml,
 //	so the //go:embed'd template already carries H — no synthetic injection needed.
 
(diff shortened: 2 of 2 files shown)


.github/workflows/hotfix-generate.yml
```diff
@@ -78,6 +78,18 @@ jobs:
...
+      # scripts H) and BEFORE the commit step (so an oversized hotfix fails the PR before its
+      # content is committed). See docs/design-customdata-size-guard.md.
+      #   Guard 1 (TestCustomDataSizeWithHotfix): slim Mode B CustomData (P + H) < 87,380.
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
+          go-version: '1.25'
+      - name: Validate CustomData size
+        run: |
+          go test ./pkg/agent/ -run TestCustomDataSizeWithHotfix -v -count=1
```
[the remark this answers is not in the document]
abigailliang-aks-sig-node: The workflow and test live on main. New official/* branches fork from main weekly, so they inherit the guard automatically. Older official/* branches lack it, but those map to VHDs that either already bake the fix (post-0729) or are approaching end of 6-month support. Backporting is not necessary.

pkg/agent/customdata_size_guard_test.go
```diff
@@ -0,0 +1,73 @@
...
+func newScriptlessGuardConfig() *datamodel.NodeBootstrappingConfiguration {
+	agentPoolProfile := &datamodel.AgentPoolProfile{
+		Name:   "nodepool1",
+		OSType: datamodel.Linux,
+		Distro: datamodel.AKSUbuntuContainerd2204Gen2,
+	}
+	return &datamodel.NodeBootstrappingConfiguration{
+		ContainerService: &datamodel.ContainerService{
+			Location: "eastus",
+			Properties: &datamodel.Properties{
```
r2k1, line 31: Please add an AKS custom-cloud case. When you set `CustomCloudEnv.Name` to `akscustom`, the renderer includes compressed `init-aks-cloud.sh`, and the Mode B payload grows from 1,756 to 14,036 bytes. A near-limit hotfix can pass this public-cloud fixture and exceed `MaxCustomDataLength` on custom-cloud nodes. Include this case in the table-driven coverage with the distro branches.

pkg/agent/customdata_size_guard_test.go
```diff
@@ -0,0 +1,73 @@
...
+		KubeletConfig:             map[string]string{},
+		EnableScriptlessNBCCSECmd: true,
+	}
+}
+
+// TestCustomDataSizeWithHotfix (Guard 1): the slim Mode B CustomData (base platform content P
+// plus any hotfix-injected scripts H already embedded in nodecustomdata.yml) must stay under
+// MaxCustomDataLength. Forcing ScriptlessCSEProvisionMode makes getScriptlessBoothook return the
+// slim CustomData directly, so this measures exactly the Mode B CustomData branch.
+func TestCustomDataSizeWithHotfix(t *testing.T) {
```
r2k1, line 65: I think you should iterate over multiple distros. ubuntu, AzureLinux, ACL

pkg/agent/customdata_size_guard_test.go
abigailliang-aks-sig-node: thx for the big picture point of view

PR #9186, opened 2026-08-12
fix: pin 2404 kernel to 6.8.0-1062-azure due to perf regression

awesomenix:
>Holding at 6.8.0-1062 forgoes 507 CVE fixes from the 1063 security rebase + 1064 roll-up. ~35 are critical-class (UAF/OOB/double-free/race) in cloud-reachable paths — core net/netfilter/TCP-IP, KVM, NVMe-oF/RDMA, the Azure mana NIC driver (CVE-2026-43056), and SMB/NFS. Rollback should therefore be a temporary, targeted mitigation for affected workloads, with a fixed 1064-base kernel (regressions reverted) prioritized.

Kernel **6.8.0-1063-azure (Ubuntu 24.04 / noble)** introduces a regression affecting networking dataplane and CPU-bound workloads. 6.8.0-1062-azure and 6.8.0-1059-azure are unaffected. The regression is in the kernel only — userspace (iptables  1.8.10-3ubuntu2 , etc.) is unchanged across the affected/unaffected builds, so package versions do not explain the symptoms.

**Root cause — two independent 1063-only changes**

1. LP #2070072 —  SAUCE: Revert "netfilter: br_netfilter: skip conntrack input hook for promisc packets" 
• Reverts a perf optimization (added to fix a bridge hairpinning bug). The conntrack input hook now runs on every promisc-mode packet.
• Impact: nf_tables ruleset programming instability and elevated per-packet conntrack CPU on bridge/promisc dataplanes. Observed as a NAT ruleset that reports a successful sync but does not persist ( nft list ruleset  empty despite hundreds of programmed rules), causing loss of service DNAT → TCP retransmit stalls (65/134/285/365s) and 5xx.
2. LP #2144730 — "ITS mitigation is not enabled on affected CPUs"
• Enables ITS mitigation (aligned branch/return thunks, CVE-2024-28956) on affected Intel SKUs.
• Impact: extra indirect-branch overhead on JIT/dispatch-heavy workloads (observed ~2x CPU/request on a managed-runtime service). Confirmed active:  indirect_target_selection: Mitigation: Aligned branch/return thunks .

**Observed impact (all on kernel 6.8.0-1063-azure)**

• Dataplane: nf_tables NAT rules fail to persist on affected nodes → service routing/DNAT loss → connection stalls and 5xx.
• CPU: ~2x CPU/request + intermittent 5xx on an indirect-branch-heavy workload, immediately after upgrade from a 1062-based image.
• Mitigation confirmed in the field: rolling the affected nodes back to a pre-1063 image cleared the issue in every case.

**Repro status**

Attribution is from the  6.8.0-1059/1062 → 1063  changelog diff plus field rollback evidence. A clean same-SKU 1062-vs-1063 A/B is still pending (test nodes landed on differing CPU SKUs).

### DIFF
--- a/vhdbuilder/packer/pre-install-dependencies.sh
+++ b/vhdbuilder/packer/pre-install-dependencies.sh
@@ -185,6 +185,18 @@ if [[ ${UBUNTU_RELEASE//./} -ge 2204 && "${ENABLE_FIPS,,}" != "true" ]]; then
       "linux-headers-azure-lts-${UBUNTU_RELEASE}"
     )
     echo "Installing fde LTS kernel for CVM Ubuntu ${UBUNTU_RELEASE}"
+  elif [ "${UBUNTU_RELEASE}" = "24.04" ]; then
+    # Pin to 6.8.0-1062-azure to avoid regression in 6.8.0-1063-azure
+    # (LP #2070072 conntrack-on-promisc revert + LP #2144730 ITS mitigation).
+    KERNEL_IMAGE="linux-image-6.8.0-1062-azure"
+    KERNEL_PACKAGES=(
+      "linux-image-6.8.0-1062-azure"
+      "linux-tools-6.8.0-1062-azure"
+      "linux-cloud-tools-6.8.0-1062-azure"
+      "linux-headers-6.8.0-1062-azure"
+      "linux-modules-extra-6.8.0-1062-azure"
+    )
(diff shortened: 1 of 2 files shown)


lilypan26 (approved):

awesomenix:
Not needed anymore

PR #9204, opened 2026-08-14
test: tolerate cloud-provider-uninitialized taint on Windows validation pods

timmy-wright:
## Summary
Windows E2E validation pods (`podWindows`, used for the servercore/nanoserver checks) had no tolerations. If the target node reported `Ready=True` before the Azure cloud-controller-manager removed its `node.cloudprovider.kubernetes.io/uninitialized:NoSchedule` taint, the validation pod stayed `Pending` until the 360s pod-wait timed out, failing the test with a misleading `FailedScheduling` error rather than a real product/VHD issue.

Investigated from build 176544987 (windows-2022-containerd-gen2): node `winw06a000000` reported ready at 887.774s still carrying the `uninitialized` taint; the validation pod then failed to schedule and the test timed out at 1250.370s.

## Change
Add a toleration for `node.cloudprovider.kubernetes.io/uninitialized:NoSchedule` to `podWindows()` so the validation pod schedules onto the target node as soon as it's Ready, without waiting on CCM taint removal.

## Testing
- `go build ./...` and `go vet ./...` pass in `e2e/`.
- `gofmt -l` clean.

### DIFF
--- a/e2e/kube.go
+++ b/e2e/kube.go
@@ -793,8 +793,7 @@ func podWindows(s *Scenario, podName string, imageName string) *corev1.Pod {
 			Tolerations: []corev1.Toleration{
 				{
 					Key:      "node.cloudprovider.kubernetes.io/uninitialized",
-					Operator: corev1.TolerationOpEqual,
-					Value:    "true",
+					Operator: corev1.TolerationOpExists,
 					Effect:   corev1.TaintEffectNoSchedule,
 				},
 			},
--- a/e2e/kube.go
+++ b/e2e/kube.go
@@ -768,9 +768,9 @@ func getNodeSelectorForScenario(s *Scenario) map[string]string {
 	}
 }
 
-func debugDaemonsetWindows(s *Scenario, podName string, imageName string) *appsv1.DaemonSet {
+func debugPodWindows(s *Scenario, podName string, imageName string) *corev1.Pod {
 	deploymentName := fmt.Sprintf("%s-test-%s-pod", s.Runtime.VM.KubeName, podName)
-	return &appsv1.DaemonSet{
+	return &corev1.Pod{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "DaemonSet",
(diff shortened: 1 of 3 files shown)


sinmentis (approved):

r2k1:
I traced the recurring failure to the Windows pause image.

Build `176544987` shows the full chain in `Test_Windows2022CachingRegression`:

- The node reported `Ready=True` with `node.cloudprovider.kubernetes.io/uninitialized:NoSchedule`.
- The scheduler assigned `cloud-node-manager-windows` at `23:25:40Z`.
- At `23:25:46Z`, kubelet reported `FailedCreatePodSandBox` for `mcr.microsoft.com/oss/v2/kubernetes/pause:3.10.2-windows-ltsc2022-amd64: not found`.
- CSE package `v0.0.52` creates that tag by adding `-windows-<version>-amd64` to the supplied pause image.
- MCR contains the legacy Windows tag for `3.10.1`, but it does not contain the generated `3.10.2` tag.
- Commit `45f5643549` in #9128 changed the Windows pause image from `3.10.1` to `3.10.2`. The E2E history changes from green to red across that commit.

`cloud-node-manager` removes the taint after node initialization. Containerd could not create its pod sandbox, so the manager never ran and the taint remained. Scheduler retries repeat the same image-pull error.

This test uses `v0.0.52` on purpose. It covers Windows VHDs built before local caching. Moving the test to the current CSE removes that compatibility coverage.

Please replace the toleration with a compatibility fix:

1. Publish the missing `3.10.2-windows-<version>-amd64` aliases for supported Windows versions.
2. Until those aliases exist, keep the Windows pause image at `3.10.1` for legacy CSE packages.

`WaitUntilNodeReady` must also require removal of the cloud-provider taint. On timeout, the test can print the final node state and `cloud-node-manager` events.

The validation-pod toleration weakens the readiness check without repairing node initialization.

smiezah-msft (approved):

PR #9218, opened 2026-08-16
fix: correct CODEOWNERS handle prefix

sinmentis:
## Summary
- Replace the invalid `$fcher` entries with `@fcher`
- Restore ownership matching for snapshot and security patch release-note changes

## Test plan
- [x] Verify no `$fcher` entries remain in `CODEOWNERS`


janenotjung-hue (approved):
LGTM

PR #9229, opened 2026-08-17
chore(gpu): bump aks-gpu-grid to 570.237-20260817204535

ganeshkumarashok:
## What

Bumps the `aks-gpu-grid` container image from `570.211.01-20260522192315` to **`570.237-20260817204535`**.

This picks up the NVIDIA GRID driver bump merged in [Azure/aks-gpu#179](https://github.com/Azure/aks-gpu/pull/179), which moved the GRID runfile from `570.211.01` to `570.237` — the vGPU18.8 build for `NVadsA10_v5`, per the [Azure N-series driver docs](https://learn.microsoft.com/en-us/azure/virtual-machines/linux/n-series-driver-setup).

## ⚠️ The new version has only TWO components, and that broke things

`570.237` is **not** a truncated `570.237.xx`. It is genuinely two-component. Verified from the runfile itself:

- runfile header: `version_string=570.237`, `targetdir=NVIDIA-Linux-x86_64-570.237-grid-azure`
- the built image ships `libnvidia-ml.so.570.237`

Two places in this repo assumed a three-component version and had to be fixed:

### 1. `.github/renovate.json` — this is why Renovate never proposed it

The `aks/aks-gpu-grid` rule pinned versioning to a regex requiring **exactly** three numeric components:

```
regex:^(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)-(?<prerelease>\d{14})$
```

`570.237-20260817204535` does not match, so Renovate could not parse the new tag and would have **silently skipped this bump indefinitely** — no error, no PR. Retriggering Renovate would never have worked. The `patch` group is now optional:

```
regex:^(?<major>\d+)\.(?<minor>\d+)(\.(?<patch>\d+))?-(?<prerelease>\d{14})$
```

Renovate [treats omitted groups as `0`](https://docs.renovatebot.com/modules/versioning/#regular-expression-versioning), so ordering still works correctly — `570.237` sorts above `570.211.01` on the minor component. Empirically verified the new regex:

| input | result |
|---|---|
| `570.211.01-20260522192315` | matches, `patch=01` |
| `570.237-20260817204535` | matches, `patch` undefined |
| `bogus-tag` | no match |
| 13-digit timestamp | no match (still anchored) |

### 2. `pkg/agent/datamodel/gpu_components_test.go`

`TestLoadConfig` asserted `^\d+\.\d+\.\d+$` and failed on the new version:

```
gpu_components_test.go:48: NvidiaGridDriverVersion '570.237' does not match expected format
```

Relaxed to `^\d+(\.\d+)+$` (two or more components), which mirrors the version guard used in the upstream `Azure/aks-gpu` repo.

**No production code change was needed.** `gpu_components.go` splits `latestVersion` on `-` into version + suffix and makes no assumption about component count.

## Validation

- ✅ Confirmed the new tag is already **mirrored to MCR** (`mcr.microsoft.com/aks/aks-gpu-grid:570.237-20260817204535`)
- ✅ `go test ./pkg/...` — all green (`pkg/agent`, `pkg/agent/datamodel`, `pkg/agent/toggles`, `pkg/vhdbuilder/datamodel`)
- ✅ `go test ./pkg/gpu/... ./parser/...` in `aks-node-controller` — green
- ✅ `make generate` produces **no diff** for this change — the GRID driver version is not embedded in any testdata, so no regeneration is required
- ✅ `components.json` validated as well-formed JSON

<details>
<summary>Note on <code>make generate</code> / <code>validate-shell</code> in this repo</summary>

`make generate` currently fails at the `validate-shell` step, and `make generate-testdata` produces testdata churn — but **both reproduce identically on a pristine clone of `main`** with zero changes applied. They are pre-existing drift, unrelated to this PR, and are deliberately excluded from this diff to keep it reviewable.
</details>

## Follow-up (deliberately out of scope)

The **three sibling GPU Renovate rules carry the identical strict 3-component regex** and have the same latent silent-failure risk if NVIDIA ships a two-component build on those branches:

- `aks/aks-gpu-cuda-lts`
- `aks/aks-gpu-cuda`
- `aks/aks-gpu-grid-v20`

I left them alone to keep this PR scoped to the driver actually being bumped, but they should probably be relaxed the same way.

Also: existing Renovate PR **#8306** proposes `570.211.01-20260629214405` — an *older driver* with a newer rebuild timestamp. It is superseded by this PR and should be closed.

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -741,7 +741,7 @@
       "downloadURL": "mcr.microsoft.com/aks/aks-gpu-grid:*",
       "gpuVersion": {
         "renovateTag": "registry=https://mcr.microsoft.com, name=aks/aks-gpu-grid",
-        "latestVersion": "570.211.01-20260522192315"
+        "latestVersion": "570.237-20260817204535"
       }
     },
     {
--- a/pkg/agent/datamodel/gpu_components_test.go
+++ b/pkg/agent/datamodel/gpu_components_test.go
@@ -32,8 +32,12 @@ func TestLoadConfig(t *testing.T) {
 	}
 
 	// Define regular expressions for expected formats
-	versionRegex := `^\d+\.\d+\.\d+$` // match version strings in a format like "X.Y.Z", where each of X, Y, and Z are numbers. e.g., "550.90.12"
-	suffixRegex := `^\d{14}$`         //  match a string of exactly 14 digits, which can represent a timestamp e.g., "20241021235610"
+	// Match NVIDIA driver versions, which have two or more dot-separated numeric components.
+	// Most are three components ("550.90.12"), but NVIDIA also ships two-component vGPU builds
+	// (e.g. GRID "570.237"), so the third component must be optional. This mirrors the version
+	// guard in the upstream Azure/aks-gpu repo: ^[0-9]+(\.[0-9]+)+$
+	versionRegex := `^\d+(\.\d+)+$`
+	suffixRegex := `^\d{14}$` //  match a string of exactly 14 digits, which can represent a timestamp e.g., "20241021235610"
 
 	// Compile the regular expressions
 	versionPattern := regexp.MustCompile(versionRegex)


awesomenix (approved):

ganeshkumarashok:
## GPU E2E failure: root-caused, not caused by this change

`Agentbaker GPU E2E` ([build 176997071](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=176997071)) reported three failures. None indicate a problem with this driver bump.

### 1. `Test_Ubuntu2204_GPUA10` — VHD-vs-source skew (systemic, blocks any GPU bump)

`ValidateScriptlessPhase3` requires zero env var differences between two paths with different sources of truth:

- `provision-config` reads `/opt/azure/components.json`, which is **baked into the VHD**
- `nbc-cmd` is generated from the **source under test**

E2E pins `SIG_VERSION_TAG_VALUE` to `refs/heads/main` (`e2e/config/config.go`), so this PR's code runs against a VHD built from `main`. Changing a GPU driver version therefore *always* yields:

```
env var differences (2): differs: GPU_DRIVER_VERSION; differs: GPU_IMAGE_SHA
```

The matching VHD can only exist after merge, so this is unfixable from within this PR. Tracked and fixed in **#9244**; this PR will be rebased on it.

### 2 & 3. `Test_Ubuntu2404_GPU_H100`, `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG_MultiGPU` — infrastructure

Both failed with bastion SSH tunnel timeouts on all 5 attempts:

```
ssh: handshake failed: failed to get reader: context deadline exceeded
```

They never reached GPU validation, and both are CUDA SKUs unaffected by this GRID change.

## Positive evidence that the bump works

From the `Test_Ubuntu2204_GPUA10` VM logs:

- CSE completed with **exit code 0**
- the node pulled and ran `mcr.microsoft.com/aks/aks-gpu-grid:570.237-20260817204535`
- `nvidia-modprobe` was enabled and started successfully
- **exactly 2** env vars differed, both derived from the same components.json version string — every other env var matched, so nothing else regressed

Most conclusively, **`Test_Ubuntu2404_GPUA10` passed** on this very same new driver image. It exercises the same GRID driver but does not run the phase 3 comparison, so it was not caught by the skew.

ganeshkumarashok:
GPU E2E here (build 176997071) is blocked by #9244, not by the driver bump itself.

`Test_Ubuntu2204_GPUA10` failed 3/3 with:

```
🔴 FAIL: expected no env var differences between provision-config and nbc-cmd, but found differences:
{"msg":"env var differences (2): differs: GPU_DRIVER_VERSION; differs: GPU_IMAGE_SHA"}
```

That skew is the expected consequence of this PR: it moves `GPU_DRIVER_VERSION`/`GPU_IMAGE_SHA` forward in the provision config while the VHD under test was baked with the previous GRID driver, so `nbc-cmd` (VHD-sourced) and `provision-config` (PR-sourced) necessarily disagree. The driver itself installs fine — the failure is purely the scriptless phase-3 env comparison.

#9244 ("don't fail scriptless phase 3 on VHD-sourced GPU env skew") is exactly that fix, and its GPU E2E run confirms it: the A10 env-skew failures are gone there, leaving only an unrelated `MIG_MultiGPU` bastion-tunnel timeout.

Suggest landing #9244 first, then re-running this gate.

sulixu:
@copilot resolve the merge conflicts in this pull request

lilypan26 (approved):

lilypan26 (approved):

PR #9233, opened 2026-08-18
feat(anc): direct HTTP download for hotfix with SHA-256 verification

abigailliang-aks-sig-node:
## Summary
- Add `artifacts` field to hotfix config JSON for direct HTTP download URLs + SHA-256 checksums
- ~25x faster than apt-get/dnf path (0.87s vs 10-22s)
- URL restricted to HTTPS + PMC allowlist, cross-domain redirects rejected
- Backward compatible: falls back to apt/dnf when no artifacts field or no OS/arch match

## Fallback strategy
| Condition | Behavior |
|-----------|----------|
| `artifacts` missing or no OS/arch match | Fallback to apt/dnf |
| HTTP network error | Fallback to apt/dnf |
| SHA-256 mismatch | **Hard fail**, keep VHD-baked ANC |
| Invalid URL (non-HTTPS, non-PMC) | **Hard fail**, keep VHD-baked ANC |

## Example hotfix config with artifacts
```json
{
  "hotfixes": {
    "202607.02": "202607.02.2"
  },
  "artifacts": {
    "202607.02.2": {
      "ubuntu-22.04-amd64": {
        "url": "https://packages.microsoft.com/...deb",
        "sha256": "c5c29cd3..."
      }
    }
  }
}
```

## Future work (separate PRs)
- Descriptor signing (`schemaVersion`, `expiresAt`, `keyId`)
- LPS/ABSvc/CRP changes to preserve `artifacts` field during serialization
- Package content validation (version/arch/OS inside package)

## Test plan
- [x] All existing hotfix tests pass
- [x] New tests: artifact HTTP success, SHA mismatch hard fail, HTTP error fallback, no artifacts fallback, invalid URL hard fail
- [x] `buildArtifactKey` tests for Ubuntu/AzureLinux
- [x] `validateArtifactURL` tests for HTTPS/host allowlist
- [ ] E2E with real PMC artifact

### DIFF
--- a/aks-node-controller/hotfix.go
+++ b/aks-node-controller/hotfix.go
@@ -541,7 +541,10 @@ func (a *App) downloadAndVerify(ctx context.Context, artifactURL, expectedSHA256
 	}
 
 	// Write to temp file.
-	dir := filepath.Dir(hotfixBinaryPath)
+	dir := a.downloadDir
+	if dir == "" {
+		dir = filepath.Dir(hotfixBinaryPath)
+	}
 	tmp, err := os.CreateTemp(dir, ".aks-node-controller-download-*")
 	if err != nil {
 		return "", fmt.Errorf("creating temp file in %s: %w", dir, err)
--- a/aks-node-controller/hotfix.go
+++ b/aks-node-controller/hotfix.go
@@ -40,9 +40,6 @@ const (
 	downloadTimeout = 30 * time.Second
 )
 
-// allowedArtifactHosts is the set of hosts from which artifact downloads are permitted.
-var allowedArtifactHosts = []string{"packages.microsoft.com"}
-
 // downloadHotfix installs the requested hotfix and stages it alongside the VHD-baked binary.
 // The wrapper script decides which binary to execute after this command returns.
 func (a *App) downloadHotfix(ctx context.Context) error {
@@ -508,12 +505,12 @@ func validateArtifactURL(rawURL string) error {
 		return newIntegrityError("artifact URL must be HTTPS, got %q", u.Scheme)
 	}
 	host := strings.ToLower(u.Hostname())
-	for _, allowed := range allowedArtifactHosts {
-		if host == allowed {
-			return nil
-		}
+	switch host {
+	case "packages.microsoft.com":
+		return nil
+	default:
+		return newIntegrityError("artifact URL host %q not in allowlist", host)
 	}
-	return newIntegrityError("artifact URL host %q not in allowlist %v", host, allowedArtifactHosts)
 }
 
 // downloadAndVerify downloads the artifact from the given URL, computes its SHA-256 digest,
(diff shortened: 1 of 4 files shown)


aks-node-controller/hotfix.go
```diff
@@ -378,6 +410,256 @@ func copyBinaryAlongside(src, dst, refPath string) error {
...
+		return nil, key
+	}
+	return &ai, key
+}
+
+// buildArtifactKey constructs the OS/arch lookup key for the artifacts map.
+// Format: "ID-VERSION_ID-GOARCH" (e.g. "ubuntu-22.04-amd64").
+func (a *App) buildArtifactKey() (string, error) {
+	osReleasePath := a.osReleasePath
+	if osReleasePath == "" {
```
djsly, line 496: this is LInux only, I think we want to eventaully use this for Windows.

Also, don't we already ahve this infromation somewhere else within ANC ? I feel this info belings in a helper function and should be detected at the start of the process ?

aks-node-controller/hotfix.go
```diff
@@ -378,6 +410,256 @@ func copyBinaryAlongside(src, dst, refPath string) error {
 	return nil
 }
 
+// tryDirectDownload attempts to download the hotfix binary directly via HTTP using the
```
djsly, line 413: why are we trying t directDown over package manager, might be worth explaining

djsly (approved):

aks-node-controller/hotfix.go
abigailliang-aks-sig-node: sure. I will add it in to comment.

basically it's for saving time. 

Original package-manager workflow
```
LPS tells ANC: Install version 202608.21.1
→ `apt-get update`: Download the package index from the Microsoft repo
→ `apt`: Locate the corresponding .deb URL in the index based on the version
→ `apt`: Download the .deb file
→ `dpkg`: Unpack and install to `/usr/bin/aks-node-controller`
→ ANC: Copy the executable to the hotfix path
```

New direct-download workflow
```
LPS directly tells ANC:
version = 202608.21.1
URL = https://packages.microsoft.com/...deb
SHA-256 = abc...
→ ANC: Download from this specific URL via HTTPS
→ ANC: Calculate and verify the SHA-256 checksum
→ ANC: Stage the hotfix executable to the hotfix path
```
thereby reducing the time required from approximately 10–22 seconds to less than 1 second.

aks-node-controller/hotfix.go
abigailliang-aks-sig-node: The reason we have buildArtifactKey here is that the current function only extracts the OS ID, like Ubuntu / Azure Linux / Mariner, but not VERSION_ID. buildArtifactKey adds the additional logic we need for that.

We could refactor this into a shared utility function with a struct like:

```
type platformInfo struct {
    ID        string
    VersionID string
    Arch      string
...
}
```
Then:

detectPackageManager()
    → uses platform.ID

buildArtifactKey()
    → uses platform.ID, platform.VersionID, platform.Arch


This way we can keep the platform detection logic in one place instead of parsing the same information separately.

PR #9241, opened 2026-08-18
fix: correct LocalDNS conntrack validation regex

pdamianov-dev:
Remove spacing assumptions that caused valid iptables rules to fail E2E validation.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

Copilot-Session: a9137761-bcd6-477c-92ad-98d6f98ebf

**What this PR does / why we need it**: Remove spacing assumptions that caused valid iptables rules to fail E2E validation.

Fixes #

### DIFF
--- a/e2e/validators.go
+++ b/e2e/validators.go
@@ -1863,12 +1863,12 @@ test "$response_rule_count" = "4" || { echo "expected 4 response-direction local
 test "$prerouting_response_rule_count" = "0" || { echo "expected 0 PREROUTING response-direction localdns NOTRACK rules, got $prerouting_response_rule_count"; exit 1; }
 
 for rule in \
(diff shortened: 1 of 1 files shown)


saewoni:
Pushed a follow-up fix to avoid the remaining `set -o pipefail` false failure: changed the representative rule checks from `echo "$rules" | grep -Eq "$rule"` to `grep -Eq "$rule" <<< "$rules"`. The previous pipeline can fail when `grep -q` exits early and `echo` gets SIGPIPE even though the regex matched. I validated the regexes against representative `iptables -S` output and reran `GOPROXY=https://proxy.golang.org,direct GONOSUMDB="*" go test ./... -run "^$"` from the e2e module successfully.

awesomenix (approved):

PR #9258, opened 2026-08-19
fix(e2e): keep localdns reply-direction NOTRACK rules allowlisted

ganeshkumarashok:
## What type of PR is this?

/kind bug
/kind cleanup

## What this PR does / why we need it

`Agentbaker GPU E2E` has been red on `main` since #9247 merged. All 18 real-node GPU scenarios (A100 x2, H100, A10 x2, NC, RTXPro6000, GridDriver, NoDriver, NvidiaDevicePlugin, MIG, DRA) fail on the same assertion from `ValidateIPTablesCompatibleWithCiliumEBPF`:

```
Rule in table raw did not match any pattern:
-A OUTPUT -s 169.254.10.10/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --sport 53 -j NOTRACK
...
Rules found that do not match any of the given patterns.
```

### Root cause

#9247 reverted #9230. That revert removed the reply-direction NOTRACK rules from `localdns.sh` **and** removed the matching entry from the e2e iptables allowlist. Those two removals do not take effect at the same time:

- `localdns.sh` is baked into the VHD by the packer `file` provisioner (`vhdbuilder/packer/*.json`). It is **not** delivered through CSE custom data. A change to it only reaches a node when a **new VHD is published**.
- The e2e allowlist lives in the test code and takes effect on the **next pipeline run**.

So every VHD published while #9230 was on `main` still installs the reply-direction rules, and those VHDs stay in support for **6 months** — but the test stopped accepting them immediately.

Timeline (UTC):

| When | What |
| --- | --- |
| 08-17 22:44 | `ea63b859a0` (#9230) adds reply rules + allowlist entry |
| 08-18 08:29 | VHD `1.1787040551.32430` published — rules baked in |
| 08-18 18:44 | build 177153237 green |
| 08-19 01:51 | `7716637881` (#9247) reverts both |
| 08-19 02:00+ | builds 177223847, 177234019, 177266063 red |

`main` may partially self-heal as post-revert VHDs roll out, but any run pinned to `VHD_BUILD_ID` in the 08-18 → 08-19 window, and any release branch still on those images, stays red. The allowlist is also required again the moment #9230 is re-landed.

## Changes

**`e2e/validation.go`** — restore the reply-direction pattern in the `raw` table. The allowlist describes what a node is *permitted* to have, not what `main` currently installs, so it must cover the whole supported VHD window.

**`e2e/validators.go`** — extract the matching loop into `iptablesRuleMatchesAnyPattern` and stop discarding the `regexp.MatchString` error. Previously a malformed allowlist entry silently degraded to "no match", which surfaces as a confusing *"unsupported iptables rule"* failure on every scenario rather than pointing at the bad pattern. The loop also assigned to the outer `matched` variable; the extraction removes that hazard. Matching semantics (regex OR literal substring, blank patterns skipped) are unchanged — note in particular that a pattern which is *not* valid regex syntax still gets its literal-substring chance, since the allowlist is deliberately fed rules pasted out of `iptables -S` (a comment like `--comment "kube-proxy (v1.31"` is not a valid regex). The compile error is only surfaced when the rule went unmatched overall, i.e. exactly when the malformed pattern is the likely culprit.

**`.github/workflows/go-test.yml`** — `e2e/` is a separate Go module, so `make test` in the root module never reaches it and `check-coverage.yml` explicitly filters it out. Without this, the new tests could only fail *inside* the ADO E2E pipeline they exist to protect, and could not block the PR causing the regression. Added an `e2e-unit-test` job modelled on the existing `version-consistency` job in `validate-components.yml`, scoped to tests needing neither Azure credentials nor network.

**`e2e/validation_iptables_test.go`** (new) — pure unit tests, no Azure resources needed:

- every global and per-table allowlist pattern compiles
- both request- and reply-direction LocalDNS rules are allowlisted (the regression guard)
- the allowlist still fails closed on an unknown rule, so the guard can't be satisfied by an overly broad pattern
- table-driven coverage of the new helper, including the invalid-pattern error path

## Does this PR introduce a user-facing change?

```release-note
NONE
```

## Testing / verification

Product behaviour is unchanged — this only touches e2e test code.

```
$ cd e2e && go build ./... && go vet ./...
$ go test -run 'Test_iptables' -v ./
--- PASS: Test_iptablesAllowlist_patternsCompile
--- PASS: Test_iptablesAllowlist_localdnsRules
--- PASS: Test_iptablesAllowlist_rejectsUnknownRule
--- PASS: Test_iptablesRuleMatchesAnyPattern (5 subtests)
ok  github.com/Azure/agentbaker/e2e
```

Negative check — deleting the restored pattern reproduces the pipeline failure exactly:

```
--- FAIL: Test_iptablesAllowlist_localdnsRules
    LocalDNS rule is not allowlisted for eBPF host routing: -A OUTPUT -s 169.254.10.10/32 -p tcp ... --sport 53 -j NOTRACK
    (x4, matching the 4 unmatched rules per node in build 177266063)
```

Refs #9230, #9247. ADO build 177266063 (definition 443878, `Agentbaker GPU E2E`).

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -21,3 +21,23 @@ jobs:
         set -ex
         make test
       name: Run unit tests for go code in the repository
+
+  e2e-unit-test:
+    # e2e is a separate Go module, so `make test` in the root module does not reach it. Run the
+    # e2e tests that need neither Azure credentials nor network so they gate pull requests.
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        persist-credentials: false
+    - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
+      with:
+        go-version-file: e2e/go.mod
+        cache-dependency-path: e2e/go.sum
+    - name: Run iptables allowlist unit tests
+      working-directory: ./e2e
+      run: |
+        set -ex
+        go test -v -run '^Test_iptables' .
--- a/e2e/validation.go
+++ b/e2e/validation.go
@@ -303,6 +303,12 @@ func getIPTablesRulesCompatibleWithEBPFHostRouting() (map[string][]string, []str
 		},
 		"raw": {
 			`^-A (PREROUTING|OUTPUT) -d 169\.254\.10\.(10|11)\/32 -p (tcp|udp) -m comment --comment "localdns: skip conntrack" -m (tcp|udp) --dport 53 -j NOTRACK$`,
+			// Reply-direction localdns NOTRACK rules. localdns.sh is baked into the VHD (packer file
+			// provisioner) rather than delivered via CSE, so VHDs published while #9230 was on main
+			// still install these rules and remain in support for 6 months. This entry must stay
+			// even though #9230 was reverted in #9247, otherwise every scenario booting one of those
+			// VHDs fails this validator.
+			`^-A OUTPUT -s 169\.254\.10\.(10|11)\/32 -p (tcp|udp) -m comment --comment "localdns: skip conntrack" -m (tcp|udp) --sport 53 -j NOTRACK$`,
 		},
 		"security": {
 			`-A OUTPUT -d 168\.63\.129\.16/32 -p tcp -m tcp --dport 53 -j ACCEPT`,
--- a/e2e/validation_iptables_test.go
+++ b/e2e/validation_iptables_test.go
@@ -120,10 +120,25 @@ func Test_iptablesRuleMatchesAnyPattern(t *testing.T) {
 			wantMatch: false,
 		},
 		{
-			name:     "invalid pattern is reported",
+			name:     "invalid pattern is reported when nothing matches",
 			rule:     `-A OUTPUT -j DROP`,
 			patterns: []string{`^-A (OUTPUT`},
 			wantErr:  true,
+		},
+		{
+			// An entry pasted verbatim out of `iptables -S` need not be a valid regex. It must
+			// still get its literal-substring chance, and must not be reported as an error once
+			// the rule is accounted for.
+			name:      "invalid pattern still matches literally",
+			rule:      `-A OUTPUT -m comment --comment "kube-proxy (v1.31" -j ACCEPT`,
+			patterns:  []string{`--comment "kube-proxy (v1.31"`},
+			wantMatch: true,
+		},
+		{
+			name:      "invalid pattern is not reported when a later pattern matches",
+			rule:      `-A OUTPUT -j DROP`,
+			patterns:  []string{`^-A (OUTPUT`, `^-A OUTPUT -j DROP$`},
+			wantMatch: true,
 		},
 	}
 
--- a/e2e/validation_iptables_test.go
+++ b/e2e/validation_iptables_test.go
@@ -1,0 +1,147 @@
+package e2e
+
+import (
+	"regexp"
+	"testing"
+)
+
+// localdnsRawRules are raw-table rules exactly as `iptables -t raw -S` prints them on a node.
+//
+// The request-direction rules are produced by build_localdns_iptable_rules in
+// parts/linux/cloud-init/artifacts/localdns.sh. The reply-direction rules are not produced by
+// localdns.sh on main today, but they are baked into VHDs published while #9230 was on main,
+// and those VHDs stay in support for 6 months. Both directions must therefore stay allowlisted.
+var localdnsRawRules = []string{
+	`-A PREROUTING -d 169.254.10.10/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK`,
+	`-A PREROUTING -d 169.254.10.10/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK`,
+	`-A PREROUTING -d 169.254.10.11/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK`,
+	`-A PREROUTING -d 169.254.10.11/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK`,
+	`-A OUTPUT -d 169.254.10.10/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK`,
+	`-A OUTPUT -d 169.254.10.10/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK`,
+	`-A OUTPUT -d 169.254.10.11/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --dport 53 -j NOTRACK`,
+	`-A OUTPUT -d 169.254.10.11/32 -p udp -m comment --comment "localdns: skip conntrack" -m udp --dport 53 -j NOTRACK`,
+	`-A OUTPUT -s 169.254.10.10/32 -p tcp -m comment --comment "localdns: skip conntrack" -m tcp --sport 53 -j NOTRACK`,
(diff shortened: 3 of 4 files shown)


ganeshkumarashok:
## E2E evidence: the fix works

`Agentbaker GPU E2E` build [177368786](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177368786) ran against this PR. Comparing it to the last `main` run before the fix, build [177266063](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177266063):

| | main (177266063) | this PR (177368786) |
| --- | --- | --- |
| Failed tests | **18** | **2** |
| Occurrences of `Rules found that do not match any of the given patterns` | every failure | **0** |
| `Test_Ubuntu2204_GPUA100` | Failed | **Passed** |
| `Test_Ubuntu2404_GPU_H100` | Failed | **Passed** |

Every one of the 18 failures on `main` was the iptables allowlist assertion. That signature is **completely absent** from the PR run — `grep -c "do not match any of the given patterns"` over all three task logs returns `0`.

### The 2 remaining failures are unrelated infrastructure flakes

Both are `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG{,_MultiGPU}`, and neither reaches validation — they fail during VM creation:

```
bastionssh.go:323: Attempt 1/5 SSH handshake failed: ssh: handshake failed: failed to get reader: context deadline exceeded
...  (5/5 attempts, across 3 re-runs)
test_helpers.go:102: FAIL: create vmss "...": failed to start bastion tunnel
```

`Test_Ubuntu2404_GPU_A100` shows as skipped for an unrelated reason too — `vmss.go:683: skipping scenario SKU quota exceeded`.

The other two pipelines are likewise unrelated to this change: `Agentbaker E2E` failed on `Test_RCV1P_ACL` and `Test_Ubuntu2204_ArtifactStreaming_TrustedLaunch`; `Agentbaker Windows E2E` failed on `Test_Windows2022_VHDCaching_LegacyTLSBootstrap` — this PR only touches Linux e2e validation code, so it cannot affect Windows.

All three pipelines have been re-queued (GPU 177378215, E2E 177378257, Windows 177378268).

ganeshkumarashok:
## Re-run results — the two non-GPU pipelines are now explained

| Pipeline | Re-run | Result |
| --- | --- | --- |
| `Agentbaker E2E` | [177378257](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177378257) | ✅ **succeeded** |
| `Agentbaker GPU E2E` | [177378215](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177378215) | 1 failure (was 18 on `main`) |
| `Agentbaker Windows E2E` | [177378268](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177378268) | pre-existing `main` failure |

### GPU E2E: 18 → 1, and the iptables signature is still gone

`grep -c "do not match any of the given patterns"` over all task logs of the re-run: **0**.

The single remaining failure is `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG_MultiGPU`, which never reaches validation — it fails during VM creation:

```
FAIL: create vmss ...: handshake failed: failed to get reader: context deadline exceeded
```

`Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG` — which failed in the first run for the same bastion reason — **passed** on the re-run, confirming this is an infrastructure flake rather than a code issue.

### Windows E2E failure is pre-existing on `main`

`Test_Windows2022_VHDCaching_LegacyTLSBootstrap` also fails on `main` in build [177273329](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177273329). This PR changes only Linux e2e validation code (`e2e/validation.go`, `e2e/validators.go`) plus a test file and a GitHub workflow, so it cannot affect Windows provisioning.

GPU E2E queued once more as 177384709 to try to clear the bastion flake.

ganeshkumarashok:
## Final CI evidence across three GPU E2E runs

The iptables regression this PR fixes is gone in **every** run. The residual failures are an Azure Bastion infrastructure problem that is **also present on `main`**, so GPU E2E cannot go fully green right now for reasons unrelated to this change.

| Build | Branch | Failing tests | `do not match any of the given patterns` | `handshake failed: failed to get reader` |
| --- | --- | --- | --- | --- |
| [177266063](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177266063) | `main` (pre-fix) | **18** | present in all 18 | **10** |
| [177368786](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177368786) | this PR | 2 | **0** | 21 |
| [177378215](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177378215) | this PR | 1 | **0** | 18 |
| [177384709](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177384709) | this PR | 4 | **0** | 6 |

Two independent signals:

1. **The allowlist assertion never fires again.** 18 → 0, sustained across three runs.
2. **The bastion timeout is not new.** It occurs 10 times in the pre-fix `main` build, before this branch existed. Every residual failure dies at `create vmss ... failed to start bastion tunnel`, i.e. before the node is ever provisioned, so `ValidateCommonLinux` — and therefore the code this PR touches — is never reached.

The set of affected tests also shuffles between runs (`MIG` failed then passed; `NvidiaDevicePluginRunning` passed twice then failed), which is the signature of an environment flake rather than a deterministic code fault.

Other pipelines:
- `Agentbaker E2E` [177378257](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177378257) — ✅ **succeeded**
- `Agentbaker Windows E2E` — `Test_Windows2022_VHDCaching_LegacyTLSBootstrap` fails identically on `main` ([177273329](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177273329)); this PR touches no Windows code
- All 23 GitHub checks pass, including the new `e2e-unit-test` job, `actionlint` and `zizmor`

ganeshkumarashok:
## Does this re-introduce what #9247 reverted? No — and here is the proof

Reasonable objection, so stating it explicitly. #9247 removed **four** things. This PR restores exactly **one** of them, and it is the only one with permissive rather than prescriptive semantics.

| Removed by #9247 (`7716637881`) | Effect | Restored here? |
| --- | --- | --- |
| `localdns.sh` — emission of `-s <ip> --sport 53 -j NOTRACK` | **product behaviour** | ❌ no |
| `localdns_spec.sh` — shellspec assertions for reply rules | test | ❌ no |
| `ValidateLocalDNSConntrackRules` (`e2e/validators.go`) | **requires** reply rules exist (`test "$response_rule_count" = "4"`) | ❌ no |
| allowlist regex in `getIPTablesRulesCompatibleWithEBPFHostRouting` | **permits** reply rules if present | ✅ **yes, only this** |

`git diff --name-only origin/main...HEAD` touches `e2e/validation.go`, `e2e/validators.go`, `e2e/validation_iptables_test.go`, `.github/workflows/go-test.yml`. **`localdns.sh` is not among them** — `grep -c "sport 53" parts/linux/cloud-init/artifacts/localdns.sh` returns `0` on this branch, exactly as on `main`. `ValidateLocalDNSConntrackRules` does not exist anywhere in this branch.

### The polarity is the whole point

An allowlist entry cannot cause a node to install a rule. It only says "if a node has this rule, do not fail it." Restoring it is incapable of re-enabling #9230's product behaviour.

### What actually triggered the revert was the *other* validator

Build [177145051](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=177145051) (`main`, inside the #9230 window, 18 failures):

```
localdns should install request and response direction NOTRACK rules   → 85 occurrences
expected 4 response-direction localdns NOTRACK rules
"do not match any of the given patterns"                                →  0 occurrences
```

So #9230 broke CI via `ValidateLocalDNSConntrackRules`, which **required** rules that VHDs did not yet have. The revert correctly removed that requirement — and this PR does not bring it back.

This PR fixes the **mirror image** of that same VHD-skew bug: the test now **forbids** rules that VHDs still *do* have. The two fixes are complementary, not opposed:

- #9247 removed a check that failed on VHDs **older** than #9230
- #9258 restores a permission so validation passes on VHDs **published during** #9230

Both are needed because `localdns.sh` ships only in the VHD image, so test-side expectations must tolerate every VHD in the 6-month support window — in both directions.

PR #9280, opened 2026-08-21
refactor(e2e): log through a Logger interface instead of testing.TB

r2k1:

Remove explicit calls to *testing.T for logging. Preparation to move away from go testing for E2E.

### DIFF
--- a/e2e/artifact_streaming.go
+++ b/e2e/artifact_streaming.go
@@ -74,9 +74,9 @@ func ValidateArtifactStreamingImagePull(ctx context.Context, s *Scenario) error 
 	// rootfs is mounted, so the assertion must run with the pod still alive.
 	kube := s.Runtime.Kube
 	pod := podStreamingImageLinux(s, image)
-	truncatePodName(s.T, pod)
+	truncatePodName(s.Logger, pod)
 
-	s.T.Logf("launching pod %q from artifact-streaming image %q", pod.Name, image)
+	s.Logger.Logf("launching pod %q from artifact-streaming image %q", pod.Name, image)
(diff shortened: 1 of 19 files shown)


r2k1:
CI triage: **Agentbaker GPU E2E** (build `177588270`) is red for a pre-existing reason, not this PR.

Only `Test_Ubuntu2404_NvidiaDevicePluginRunning_MIG_MultiGPU` fails (3 runs, 3 failures). Root cause is `StandardNCADSA100v4Family` core quota exhaustion in WestUS3: the A100 scenarios run in parallel and need 144 cores against a limit of 100. When the VMSS PUT is rejected synchronously the scenario skips and the build is green; when the PUT is accepted but the instance never allocates, the VMSS exists with no ID or tags, the VM never boots, and the failure surfaces as five bastion SSH handshake timeouts. Boot diagnostics for the failing VM return `BlobNotFound`, while every passing GPU scenario in the same run has a full serial console.

The same scenario fails identically on `main`: builds `177574607`, `177446225`, `177400259`, `177383459`, `177365480`, `177174586` — 6 of 6, same `bastionssh.go:304/:323` retry ladder and same `vmss.go:525` no-tags line.

This PR cannot be the cause. `bastionssh.go` and `cluster.go` are not in the diff, the failing test at `scenario_gpu_managed_experience_test.go:736` is untouched (the only hunks in that file are inside `Test_DCGM_Exporter_Compatibility`, lines 305-390), and no product code changed. Agentbaker E2E (`177588266`) and Agentbaker Windows E2E (`177588268`) pass on the same commit.

Worth a separate repair item: raise the WestUS3 A100 quota or serialize the A100 scenarios, and fail fast when a VMSS comes back with no ID instead of burning ~200s on doomed bastion handshakes.

timmy-wright (approved):

PR #9313, opened 2026-08-26
fix: use base64 encoding for SSH key literals in PowerShell template

benjamin-brady:
Switch `getSSHPublicKeysPowerShell` from rendering key data inside double-quoted PowerShell strings to base64-encoding each value and emitting a decode expression. The base64 alphabet (`A-Za-z0-9+/=`) contains no PowerShell delimiters or interpolation characters, ensuring key comments are preserved as literal data regardless of content.

### Changes
- `pkg/agent/utils.go` — new `encodePowerShellBase64Literal` helper, called from `getSSHPublicKeysPowerShell`
- `pkg/agent/utils_test.go` — unit tests covering various key comment contents, whitespace trimming, multiple keys, and round-trip decoding

### DIFF
--- a/pkg/agent/utils_test.go
+++ b/pkg/agent/utils_test.go
@@ -1671,3 +1671,107 @@ func TestValidateAndSetLinuxNodeBootstrappingConfiguration_TransparentHugePageVa
 		})
 	}
 }
+
+func TestEncodePowerShellBase64Literal(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{"plain key", "ssh-rsa AAAAB3 user@host",
+			"[System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String('c3NoLXJzYSBBQUFBQjMgdXNlckBob3N0'))"},
+		{"embedded apostrophe", "ssh-rsa AAAAB3 user's key",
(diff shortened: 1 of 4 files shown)


pkg/agent/utils_test.go
```diff
@@ -1671,3 +1671,107 @@ func TestValidateAndSetLinuxNodeBootstrappingConfiguration_TransparentHugePageVa
 		})
 	}
 }
+
+func TestEncodePowerShellBase64Literal(t *testing.T) {
```
timmy-wright, line 1675: It would be good to have an agentbaker e2e test that tries to use a certificate with bad characters in it so we can check the full e2e flow.

timmy-wright (dismissed):
I think for this one we need a full e2e test using a cert with invalid characters in it.

pkg/agent/utils_test.go
benjamin-brady: Added `Test_Windows2022_SSHKeyWithSpecialCharComment` — injects an SSH key with apostrophe and semicolons in the comment (`user's test;key`), then validates it appears verbatim in `C:\ProgramData\ssh\administrators_authorized_keys`.

E2E result from [build 178263924](https://msazure.visualstudio.com/09706533-03bf-4b43-9a9b-b49c75429646/_build/results?buildId=178263924):
```
✓ Windows2022 SSH key with special char comment (246.97s)
```

r2k1 (approved):

sinmentis (approved):

r2k1 (approved):

PR #9324, opened 2026-08-27
refactor(e2e): decouple scenarios from test control

r2k1:

Stack layer 1, on top of #9323 (base branch `r2k1-ci-fix-azure-login-comment`). It decouples the E2E scenario implementation from `testing.TB` and fixes two shared-state races, without changing how scenarios are defined or run.

`Scenario` no longer carries a `testing.TB`. Implementation code logs through `toolkit.Logger` and records named check results (`Scenario.recordCheck`) instead of creating sub-tests, so no file under `e2e/` outside `_test.go` depends on test control any more.

`RunScenario(t, s)` stays as a compatibility adapter, moved to `e2e/scenario_runner_test.go`. It:
- attaches a test-backed logger and the test context,
- owns the cleanup stack and its `t.Cleanup` registration, keeping the existing `VHDCreation` / `VMProvision` staging and cleanup ordering,
- maps recorded checks back into `t.Run` child results, so ADO keeps per-check CSE timing history under the same names (`TotalCSEDuration`, `Task_<name>`),
- turns a returned `*skipError` into `t.Skip`, preserving skip semantics that previously came from `t.Skipf` inside implementation code.

Every current scenario `_test.go` definition, scenario name, tag filter, retry, pipeline, `TestMain`, environment parsing, config initialization, signal handling and gotestsum behavior are unchanged.

**Race fixes**

1. Effective VM size is now per scenario. The Gen1 and NVMe→SCSI fallbacks used to assign `config.Config.DefaultVMSKU`, a process-global read by every concurrently running scenario, so one image's fallback silently changed the VM size of unrelated scenarios. The effective size now lives in `ScenarioRuntime.VMSize`, and VMSS construction reads it through `scenarioVMSize(s)`.
2. ARM poll options are per operation. `config.DefaultPollUntilDoneOptions` was a single mutable `*runtime.PollUntilDoneOptions` handed to every long-running ARM call across all scenarios. It is replaced by `config.PollUntilDoneOptions()`, which returns a fresh value per call; all call sites are updated. The 15s frequency is unchanged.

**Out of scope for this layer**: standalone CLI, scenario registry, executor, JUnit writer, urfave config migration, scenario file moves, pipeline changes, docs.

**Validation**

- `gofmt` clean on all changed Go files.
- `go build ./...` and `go vet ./...` clean in `e2e/`.
- `go test` and `go test -race` pass for every unit test in `e2e/` and in `e2e/assert`, `e2e/components`, `e2e/config`, `e2e/nodeexporter`, `e2e/toolkit`. The live scenario `Test_*` functions need real Azure infrastructure and were not run locally; the E2E gate covers them.
- New unit tests cover `recordCheck`, `markScenarioOutcome` skip-vs-failure classification, `skipIfSKUNotAvailableErr`, scenario-local VM sizing, and that `PollUntilDoneOptions()` returns independent values.
- Verified repo-wide that no `e2e` implementation file imports `testing` and that no `s.T` reference remains; `testing` survives only in `_test.go` files and in `e2e/toolkit/log.go`, which provides the runner-boundary test logger.

**Behavior caveats**

- A scenario that both declares `ExpectedError` and hits an Azure SKU capacity skip now skips instead of asserting the expected error. Previously `t.Skip` ended the goroutine before the assertion; the new code checks for `*skipError` explicitly to keep that outcome.
- `extractLogsFromVMWindows` now branches on `Scenario.failed`, set when the scenario flow returns, rather than on `t.Failed()`. Failures reported outside the scenario flow (for example a cleanup error) no longer flip it.
- The dead `toolkit.UnwrapTestingT` helper and its test are removed; its only caller was the CSE timing sub-test code replaced here.

None.

### DIFF
--- a/e2e/aks_model.go
+++ b/e2e/aks_model.go
@@ -359,7 +359,7 @@ func addFirewallRules(
 		if err != nil {
 			return fmt.Errorf("failed to start adding route %q: %w", *route.Name, err)
 		}
-		_, err = poller.PollUntilDone(ctx, config.DefaultPollUntilDoneOptions)
+		_, err = poller.PollUntilDone(ctx, config.PollUntilDoneOptions())
 		if err != nil {
 			return fmt.Errorf("failed to add route %q to AKS route table: %w", *route.Name, err)
 		}
@@ -409,7 +409,7 @@ func ensureFirewallRouteTable(
 		if err != nil {
 			return fmt.Errorf("failed to start creating firewall route table %q: %w", routeTableName, err)
 		}
-		routeTableResp, err := poller.PollUntilDone(ctx, config.DefaultPollUntilDoneOptions)
+		routeTableResp, err := poller.PollUntilDone(ctx, config.PollUntilDoneOptions())
 		if err != nil {
 			return fmt.Errorf("failed to create firewall route table %q: %w", routeTableName, err)
 		}
@@ -1150,7 +1150,7 @@ func createNetworkIsolatedSecurityGroup(ctx context.Context, cluster *armcontain
 	if err != nil {
 		return nil, err
 	}
-	nsg, err := poller.PollUntilDone(ctx, config.DefaultPollUntilDoneOptions)
+	nsg, err := poller.PollUntilDone(ctx, config.PollUntilDoneOptions())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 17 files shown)


r2k1:
Closing because the stacked-PR split was canceled. Work continues in #9321.

PR #9325, opened 2026-08-27
feat: cache azure iptables monitor

QxBytes:
iptables blocker binary inside the monitor image is used during startup so should be cached

azure-iptables-monitor was not cached, leading to startup latency

NA

### DIFF
--- a/parts/common/components.json
+++ b/parts/common/components.json
@@ -300,6 +300,17 @@
       "windowsVersions": []
     },
     {
+      "downloadURL": "mcr.microsoft.com/containernetworking/azure-iptables-monitor:*",
+      "amd64OnlyVersions": [],
+      "multiArchVersionsV2": [
+        {
+          "renovateTag": "registry=https://mcr.microsoft.com, name=containernetworking/azure-iptables-monitor",
+          "latestVersion": "v0.0.5-0"
+        }
+      ],
+      "windowsVersions": []
+    },
+    {
       "downloadURL": "mcr.microsoft.com/containernetworking/azure-npm:*",
       "amd64OnlyVersions": [],
       "multiArchVersionsV2": [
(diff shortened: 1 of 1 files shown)


benjamin-brady (approved):
LGTM. It should speed up node provisioning plus it's useful to have a cache if MCR has any problems.

ttevatia:
@QxBytes @benjamin-brady Thanks for pushing this. I see the Linux VHD builds are failing, could you please take a look and get this PR merged? Thanks!

ttevatia:
I traced the reported exit `170` against the exact PR revision. This does **not**
appear to be an `azure-iptables-monitor` pull failure.

Source-level evidence:

- `ERR_BCC_INSTALL_TIMEOUT=170` is defined for the BCC installation path.
- `startEBPFToolsInstallation` starts `installBcc` as a background subshell and
  records its PID in `BCC_PID`.
- `cacheContainerImageComponents` then runs concurrently and drains jobs using
  unscoped `wait -n` while testing `jobs -p`; that can reap the BCC subshell,
  not only PIDs in `image_pids`.
- When that happens, the loop prints
  `A background job pullContainerImage failed: 170`, even though status 170
  came from `installBcc`.
- The image-fetcher/retry path does not define or return 170. The observation
  that every started image pull logged success is consistent with this.

The likely primary failure is therefore a BCC apt install or purge timeout
recorded in `/var/log/bcc_installation.log`. The missing published BCC log and
`disk-usage.txt` are downstream effects after Packer exits early. The image
entry itself is low-risk: all other Linux variants, including Ubuntu 22.04
ARM64, passed in the latest gate.

Recommended unblock: rerun the failed Ubuntu 22.04 Gen2 job. If it repeats,
capture `/var/log/bcc_installation.log`. Separately, the latent diagnostic bug
should be fixed by waiting only on the image PIDs rather than any background
job, so a BCC failure cannot be attributed to `pullContainerImage`.

PR #9326, opened 2026-08-27
fix: preserve CIS module denylist (incl. SCTP) on AzureLinux 3.0

djsly:
## Summary

Customer report ("AKS STCP Support" email thread) found that SCTP works on AzureLinux V3 node pools but is blocked on Ubuntu node pools. Traced to `/etc/modprobe.d/CIS.conf` containing `install sctp /bin/true`.

## Root Cause

`parts/linux/cloud-init/artifacts/modprobe-CIS.conf` is the single shared source for both Ubuntu and AzureLinux and blocks SCTP/DCCP/RDS/TIPC/cramfs/freevxfs/jffs2/hfs/hfsplus/usb-storage per CIS 3.5.x/1.1.1.x.

In `vhdbuilder/packer/packer_source.sh` (`copyPackerFiles`):
- **Ubuntu:** the file is always baked in. Only 4 unrelated lines (`algif_aead`/`esp4`/`esp6`/`rxrpc`, a 2026 LPE CVE mitigation) are surgically stripped on fixed 22.04/24.04 kernels (#9054).
- **AzureLinux 3.0 (non-OSGuard):** the *entire file* was skipped (#8546), not just the 4 CVE lines — silently unblocking SCTP and the rest of the CIS baseline as a side effect.

#8546 intended only to unblock the 4 CVE-related modules on AzL3 once its kernel had the upstream fix, but took the blunt "skip whole file" approach instead of the surgical per-line removal later implemented correctly for Ubuntu in #9054.

## Changes

- `vhdbuilder/packer/packer_source.sh`: extracted the Ubuntu surgical-stripping logic into a reusable `bakeModprobeCISWithoutVulnerableModules()` helper and applied it to the AzureLinux 3.0 path instead of skipping the whole `modprobe-CIS.conf` file. Only `algif_aead`/`esp4`/`esp6`/`rxrpc` are stripped; the rest of the CIS denylist (`dccp`/`sctp`/`rds`/`tipc`/`cramfs`/etc.) is now baked in consistently across OS streams.
- `vhdbuilder/packer/test/linux-vhd-content-test.sh`: extended `testVulnerableKernelModulesDisabled` to assert the CIS baseline denylist (including `sctp`) is present on AzureLinux 3.0, not just Ubuntu.
- `e2e/validators.go`: updated `ValidateVulnerableKernelModulesDisabled` comments/assertions for AzureLinux 3.0 to also assert `dccp`/`sctp`/`rds`/`tipc` remain blocked, alongside the existing CVE-module absence checks.

## Testing

- `bash -n` + `shellcheck` on both modified shell scripts (no new warnings).
- `go build ./e2e/...` compiles cleanly.
- Simulated the `sed` transformation against the real `modprobe-CIS.conf` locally — confirmed the 4 CVE lines are removed while `dccp`/`sctp`/`rds`/`tipc`/`cramfs`/etc. all remain (see PR discussion/commit for output).

## Related

- AB#39442583
- Follows up on #8546 / #9054

### DIFF
--- a/vhdbuilder/packer/test/linux-vhd-content-test.sh
+++ b/vhdbuilder/packer/test/linux-vhd-content-test.sh
@@ -1480,9 +1480,10 @@ testNfsServerService() {
 # To add a new CVE mitigation, append the module to BOTH loops below — the
 # absence loop AND the default presence + load-refusal loop.
 #
-# AzureLinux 3.0 is descoped: kernel 6.6.139.1-1.azl3+ fixes the CVEs upstream and
-# the modprobe blacklist is NOT baked into newly-built AzL3 VHDs (customer workloads
-# require those modules). Ubuntu 22.04 linux-azure 5.15.0-1116-azure and Ubuntu
+# AzureLinux 3.0 is descoped: kernel 6.6.139.1-1.azl3+ fixes the CVEs upstream, so only the
+# algif_aead/esp4/esp6/rxrpc lines are stripped from newly-built AzL3 VHDs (customer workloads
+# require those modules); the rest of the CIS module denylist (dccp/sctp/rds/tipc/cramfs/etc.)
+# is still baked in and asserted below. Ubuntu 22.04 linux-azure 5.15.0-1116-azure and Ubuntu
 # 24.04 linux-azure 6.8.0-1058-azure include the fixes, so newly-built Ubuntu
 # 22.04/24.04 VHDs with a fixed running kernel also stop baking the vulnerable-module
 # blacklist while keeping the baseline CIS module deny list. Ubuntu 20.04 and vulnerable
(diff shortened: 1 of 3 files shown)


vhdbuilder/packer/test/linux-vhd-content-test.sh
```diff
@@ -1553,19 +1554,29 @@ testVulnerableKernelModulesDisabled() {
...
-          failed=1
-        else
-          echo "$test: CIS modprobe config correctly blocks ${mod}"
-        fi
-      done
-    fi
+    # Only the algif_aead/esp4/esp6/rxrpc lines above are stripped for the CVE mitigation;
+    # the rest of the CIS 3.5.x / 1.1.1.x module denylist must remain intact on every OS
+    # stream (including AzureLinux 3.0, which used to skip the whole modprobe-CIS.conf file).
+    for mod in dccp sctp rds tipc; do
```
djsly, line 1560: this hardcoded list creates a hard copling with the contain of the CSI file, we should ensure maybe that what is in the file is blocked ?

awesomenix (approved):

PR #9328, opened 2026-08-27
fix(gpu): remove prebake before PIS customization

sulixu:
## Summary

- remove the AKS CUDA driver prebake during Ubuntu `basePrep` when `PreProvisionOnly=true`
- run cleanup before PIS customer customization scripts can install a customer-owned driver
- fail the PIS bake when cleanup leaves the AKS prebake marker in place
- leave normal node provisioning and the existing node-time cleanup behavior unchanged

## Why

PIS starts from the shared AKS VHD, which can contain an AKS-prebaked NVIDIA driver and `/opt/azure/aks-gpu/dkms-marker`. Customer customization runs after `basePrep`. If the marker is captured with a customer driver, node-time cleanup cannot reliably distinguish AKS-owned files from customer-owned files, especially when both drivers use the same version.

Removing the AKS prebake during PIS `basePrep` establishes a clean ownership boundary before customer scripts run.

## Behavior

- PIS with a customer driver captures only the customer installation and no AKS marker.
- PIS without a customer driver captures a driverless image.
- A managed GPU node installs the AKS driver during `nodePrep` because the prebake marker is absent.
- Normal non-PIS nodes retain the shared-VHD prebake fast path.
- Existing cached PIS images require a rebake to receive this behavior.

## RequiresCurrentSourceVHD: true
Developers can add `RequiresCurrentSourceVHD: true` to E2E, so it skips PR E2E run. example: https://github.com/Azure/AgentBaker/pull/9328/changes#diff-54b233c475f80f6f594b754ac778c41e99b79533f6e7259aeefbf2a2ccb140a2R3396

But PR VHD build gate is not skipped: runs against the PR-built VHD.

Prove:
Build: [179179245](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=179179245&view=results)
 
  Proof from the E2E log:

  VHD_BUILD_ID: 179179245
  SKIP_CURRENT_SOURCE_VHD_SCENARIOS: false
  VHD_BUILD_ID is specified (179179245). Running tests using VHDs from that build

  The test ran and passed:

  ✓ Ubuntu2404 VHD caching GPU prebake removed before customization
  ✓ Ubuntu2404 VHD caching GPU prebake removed before customization VHD creation
  ✓ Ubuntu2404 VHD caching GPU prebake removed before customization VHD creation VM provision

  The E2E job succeeded, and the GitHub AKS Linux VHD Build - PR check-in gate check passed.


xuexu6666:
### Code review

No issues found. Checked for bugs and CLAUDE.md/AGENTS.md compliance against commit `1230a24e902c7ed9f888829ddd0aa9d0258c5260`.

Verified the key correctness points of the change:
- `logs_to_events` propagates the wrapped function's non-zero exit status, so `|| exit "$ERR_GPU_PREBAKE_CLEANUP_FAIL"` in `cse_main.sh` fires correctly and aborts the bake (fail-closed, intended).
- The new wrapper `cleanUpPrebakedGPUDriverForImageCustomization` is marker-guarded, re-checks the marker after cleanup, and returns 1 if it survives — matching the "prove complete removal or fail" pattern established in prior PRs (#8786).
- `PRE_PROVISION_ONLY` and the `OS == UBUNTU_OS_NAME` guard are in scope during `basePrep`; the Ubuntu guard is load-bearing since the function is only defined in the Ubuntu-templated install script.
- `ERR_GPU_PREBAKE_CLEANUP_FAIL=236` is unique; ShellSpec coverage exercises the no-op, success, and failure paths plus basePrep-only placement.

Non-blocking nits only (not posting inline): the new error code is inserted out of numeric order in `cse_helpers.sh`, and the success path relies on `echo`'s exit status rather than an explicit `return 0`.

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -378,6 +376,24 @@ cleanUpPrebakedGPUDriver() {
...
+# customer-owned software. The function is a no-op without an AKS marker and fails if cleanup
+# leaves the marker in place, which prevents capture of an image with ambiguous driver ownership.
+cleanUpPrebakedGPUDriverForImageCustomization() {
+    local marker="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
+    if [ ! -f "${marker}" ]; then
+        return 0
+    fi
+
+    cleanUpPrebakedGPUDriver
+    if [ -f "${marker}" ]; then
```
xuexu6666, line 389: **[blocking] Spurious bake failure on a GPU-SKU PIS bake.** A PIS bake for a customer GPU driver runs on the target GPU SKU, and on cuda/cuda-lts SKUs `nvidia.ko` auto-loads at boot. The wrapped `cleanUpPrebakedGPUDriver` only `rmmod`s when the module is idle (`refcnt==0` and no `/dev/nvidia*`); if a device node is present or `refcnt>0` it skips the unload → `module_after=true` → `status=incomplete` → the marker is kept. This re-check then returns 1 → `cse_main.sh` turns it into `exit 236` and hard-fails the whole bake — on exactly the customer-owned-driver case this PR targets. Pre-PR an incomplete teardown was only a logged signal; here it becomes a fatal, non-deterministic failure (depends on transient module-busy state).

Suggested: gate the failure on the *ambiguity-causing on-disk residue* (marker + DKMS registration + `nvidia-modprobe` binary), not on a merely-resident-but-idle module; or settle/retry the unload (a reboot is available between PIS phases) before failing. At minimum, surface the inner `status=incomplete` (`dkms_after`/`module_after`) in the failure message so the cause is diagnosable.

xuexu6666 (commented):
Reviewed the PIS prebake-removal fix. The approach — establish a clean ownership boundary at **bake** time instead of disambiguating AKS-vs-customer driver ownership at node time — is the right call, and the return-code plumbing / arch gating check out (marker is x86-Ubuntu-only; `PRE_PROVISION_ONLY` is set only on the pre-provision path; `logs_to_events` propagates the exit code). One blocking item on the strict-failure path, plus test/observability follow-ups inline.

**Missing coverage worth adding here:** there's no PIS-on-GPU e2e guarding this. AgentBaker e2e already has both ingredients — `nbc.PreProvisionOnly=true` bakes and GPU SKUs — and `runScenarioWithPreProvision` bakes → captures the VHD → boots a fresh node. A GPU scenario on the shared gen2 Ubuntu (prebake) VHD run through that flow would (a) guard the blocking finding for free (a bake that `exit 236`s fails Stage 1) and (b) let Stage 2 assert the marker/DKMS are gone and a customer-baked driver survives. That's the natural home for the regression guard since it sits next to this code.

parts/linux/cloud-init/artifacts/cse_main.sh
```diff
@@ -410,6 +410,13 @@ EOF
         createManDbAutoUpdateFlagFile
         /usr/bin/mandb && echo "man-db finished updates at $(date)" &
     fi
+
+    # PIS captures basePrep output before customer customization and skips nodePrep on the bake VM.
+    # Release the shared-VHD GPU prebake here, while every marked NVIDIA asset is still AKS-owned.
+    # The strict wrapper keeps the marker and fails the bake if complete removal cannot be proven.
+    if [ "${PRE_PROVISION_ONLY}" = "true" ] && [ "$OS" = "$UBUNTU_OS_NAME" ]; then
```
xuexu6666: **[discussion] Managed-GPU PIS pays a nodePrep driver recompile — worth documenting.** The strip here is unconditional (`PRE_PROVISION_ONLY && Ubuntu`), so a managed-GPU pool built from this PIS image loses the prebake fast path and rebuilds the DKMS module at `nodePrep` (the work #8786/#8803 added the prebake to skip). This is *correct and unavoidable*: a PIS image is reusable across pools, so at bake time you can't know whether it'll be attached to a `--gpu-driver None` or a managed pool — gating the strip on driver mode would bake pool-specific behavior into a shared artifact. Recommend just calling out the managed-GPU nodePrep recompile cost explicitly in the PR/commit so that segment isn't surprised. (Only the DKMS module + marker are removed; the aks-gpu driver bits stay cached in the VHD, so it's a recompile, not a network pull.)

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -378,6 +376,24 @@ cleanUpPrebakedGPUDriver() {
...
+        echo "Failed to remove the AKS GPU driver prebake before image customization"
+        return 1
+    fi
+
+    echo "AKS_GPU_PREBAKE event=image_customization status=released"
```
xuexu6666, line 394: [nit] The `event=image_customization status=released` line doesn't carry the `dkms_after`/`module_after` fields that the `event=teardown` line does, so a fleet query can't distinguish 'released cleanly' from 'released but a module was still resident' without correlating two log lines. Consider echoing those fields here too.

spec/parts/linux/cloud-init/artifacts/cse_install_ubuntu_spec.sh
```diff
@@ -87,6 +87,32 @@ Describe 'cse_install_ubuntu.sh'
         End
     End
 
+    Describe 'cleanUpPrebakedGPUDriverForImageCustomization'
```
xuexu6666, line 90: [medium] All three cases stub `cleanUpPrebakedGPUDriver` (via redefinition), so they exercise only the thin wrapper's three branches — never the real idle-check/rmmod/marker-retention logic that decides whether the marker survives. That's exactly the path that can spuriously fail the bake on a GPU SKU (see the blocking comment on `cse_install_ubuntu.sh`). Please add a case that runs the *real* `cleanUpPrebakedGPUDriver` with a faked resident-but-idle module (marker kept → wrapper returns 1), plus an idle-module success case.

spec/parts/linux/cloud-init/artifacts/cse_main_pis_reconcile_spec.sh
```diff
@@ -114,5 +114,19 @@ Describe 'cse_main.sh PIS-safe configuration'
             The line 4 of output should equal '    nodePrep'
             The lines of output should equal 4
         End
+
+        It 'removes the GPU prebake during image-creation basePrep only'
+            gpu_cleanup_counts() {
+                phase_count "basePrep" "cleanUpPrebakedGPUDriverForImageCustomization"
+                phase_count "nodePrep" "cleanUpPrebakedGPUDriverForImageCustomization"
+                phase_body "basePrep" | code_lines | grep -F -c 'if [ "${PRE_PROVISION_ONLY}" = "true" ] && [ "$OS" = "$UBUNTU_OS_NAME" ]; then' || true
```
xuexu6666: [medium] This asserts the guard by `grep`-ing the source text (the `if [ "${PRE_PROVISION_ONLY}"...` string and the `ERR_` token), so it only proves the lines *exist* in `basePrep` — it executes nothing and is fragile to any whitespace/refactor of the condition. A behavioral test (run `basePrep` with `PRE_PROVISION_ONLY=true` on a mocked Ubuntu env, assert the cleanup ran; plus the non-Ubuntu / non-PIS no-op) would be far more robust.

parts/linux/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -149,6 +149,7 @@ ERR_NVIDIA_GPG_KEY_DOWNLOAD_TIMEOUT=225 # Timeout waiting for NVIDIA GPG key dow
 ERR_NVIDIA_AZURELINUX_REPO_FILE_DOWNLOAD_TIMEOUT=226 # Timeout waiting for NVIDIA AzureLinux repo file download
 ERR_MANAGED_NVIDIA_EXP_INSTALL_FAIL=227 # Error installing Managed NVIDIA GPU experience packages
 ERR_NVIDIA_DCGM_INSTALL=235 # Error installing Managed NVIDIA GPU experience packages from cache
+ERR_GPU_PREBAKE_CLEANUP_FAIL=236 # Error removing the AKS GPU driver prebake before image customization
 ERR_NVIDIA_DCGM_FAIL=228 # Error starting or enabling NVIDIA DCGM service
```
xuexu6666, line 153: [nit] `236` is free (no collision), but it's inserted between `235` and `228` — out of numeric order in this block. Minor; since it's a new code, consider placing it at the numeric tail for readability.

spec/parts/linux/cloud-init/artifacts/cse_main_pis_reconcile_spec.sh
```diff
@@ -114,5 +114,19 @@ Describe 'cse_main.sh PIS-safe configuration'
             The line 4 of output should equal '    nodePrep'
             The lines of output should equal 4
         End
+
+        It 'removes the GPU prebake during image-creation basePrep only'
```
xuexu6666: [medium] **Add a PIS-on-GPU e2e to guard this path.** ShellSpec can't exercise the real bake, and the change's one realistic failure mode (the blocking comment on `cse_install_ubuntu.sh` — an auto-loaded prebaked module leaving the marker → `exit 236`) only shows up on a real GPU-SKU bake. AgentBaker e2e already has both ingredients: `nbc.PreProvisionOnly=true` bakes and GPU SKUs (`Test_ACL_GPUNC` etc.), and `runScenarioWithPreProvision` (`e2e/test_helpers.go`) bakes → `CreateImage` captures the VHD → boots a fresh node in Stage 2. A GPU scenario on the shared gen2 Ubuntu (prebake) VHD run through that flow would:
- guard the blocking finding for free — a bake that `exit 236`s fails Stage 1;
- let Stage 2 assert `/opt/azure/aks-gpu/dkms-marker` and `/var/lib/dkms/nvidia` are gone; and
- with a `PreProvisionBootstrapConfigMutator` that installs a customer driver on a `--gpu-driver None` pool, assert that driver survives on the Stage-2 node (the actual end-to-end fix).

This is the natural home for the regression guard since it lives next to the code being changed. (The full customer-flow-through-the-RP variant can follow later in aks-rp e2ev3's existing GPU+PIS suite.)

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
```diff
@@ -378,6 +376,24 @@ cleanUpPrebakedGPUDriver() {
...
+# leaves the marker in place, which prevents capture of an image with ambiguous driver ownership.
+cleanUpPrebakedGPUDriverForImageCustomization() {
+    local marker="${GPU_DKMS_MARKER_FILE:-/opt/azure/aks-gpu/dkms-marker}"
+    if [ ! -f "${marker}" ]; then
+        return 0
+    fi
+
+    cleanUpPrebakedGPUDriver
+    if [ -f "${marker}" ]; then
+        echo "Failed to remove the AKS GPU driver prebake before image customization"
```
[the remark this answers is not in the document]
sulixu: cleanUpPrebakedGPUDriver is a best-effort cleanup, and this PR does not try to harden or refine cleanUpPrebakedGPUDriver func.

This PR only changes when the existing cleanup runs: it runs before PIS customer customization.  The E2E test validates this ordering and ownership transition; it does not need validate exhaustive artifact removal

parts/linux/cloud-init/artifacts/ubuntu/cse_install_ubuntu.sh
sulixu: cleanUpPrebakedGPUDriver is a best-effort cleanup, and this PR does not try to harden or refine cleanUpPrebakedGPUDriver func.

This PR only changes when the existing cleanup runs: it runs before PIS customer customization.  The E2E test validates this ordering and ownership transition; it does not need validate exhaustive artifact removal

ganeshkumarashok:
Question for future design and for PIS team: what's a good way to prevent such issues in the future? vs depending on one-off logic and always considering PIS edge cases for every PR?

sulixu:
close it, implementation will change after PIS+GPU change

PR #9344, opened 2026-08-31
fix(windows): retry kubelet start check after NodeResetScriptTask

xuexu6666:
## What

`Start-NodeResetScriptTask` (in `parts/windows/windowscsehelper.ps1`) asserted the `kubelet` service was `Running` **exactly once**, immediately after the `k8s-restart-job` reset task reported success. `windowsnodereset.ps1` issues `Start-Service kubelet`, but on slower first boots — notably **sysprepped/cached (golden) images** provisioned in the Windows VHDCaching e2e scenario — kubelet can still be in `StartPending` at that instant, so CSE failed intermittently with `WINDOWS_CSE_ERROR_START_NODE_RESET_SCRIPT_TASK` ("kubelet service is not running after NodeResetScriptTask completed").

## Fix

Poll for kubelet to reach `Running` with bounded retries (6 × ~5s) and re-nudge `Start-Service` on each attempt, mirroring the existing `Start-Containerd` pattern in `staging/cse/windows/containerdfunc.ps1`. A genuinely failed kubelet still exits with the same code/message after the timeout, so the intent of #8970 (validate the reset result) is preserved — this only absorbs start-latency.

## Evidence

`Agentbaker Windows E2E` is flaky on this exact scenario on **unchanged `main`** (green some days, red others). The failing case is always `Test_Windows2022_VHDCaching_LegacyTLSBootstrap/VHDCreation/VMProvision`: the base node comes up and validates, but the node provisioned from the sysprepped/captured image fails the reset kubelet check. Within a run, the faster base node passes while the slower cached node fails — consistent with a start-latency race, not a deterministic break.

## Testing

Adds a Pester case in `windowscsehelper.tests.ps1` covering kubelet reaching `Running` after a retry; existing cases (including "fails when kubelet is not running") still hold. Windows E2E triggered on this PR.

### DIFF
--- a/parts/windows/windowscsehelper.ps1
+++ b/parts/windows/windowscsehelper.ps1
@@ -333,8 +333,45 @@ function Start-NodeResetScriptTask {
         Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_START_NODE_RESET_SCRIPT_TASK -ErrorMessage "NodeResetScriptTask failed with result $($taskInfo.LastTaskResult)"
     }
 
+    # windowsnodereset.ps1 (run by NodeResetScriptTask) issues Start-Service kubelet, but the
+    # service can still be in StartPending immediately after the task reports success, especially
+    # on slower first boots such as sysprepped/cached (golden) images. Asserting Running exactly
+    # once here caused intermittent CSE failures, so poll with bounded retries and re-nudge the
+    # service, matching the containerd start pattern in containerdfunc.ps1 (Start-Containerd).
     $kubeletService=Get-Service -Name "kubelet" -ErrorAction SilentlyContinue
-    if ($null -eq $kubeletService -or $kubeletService.Status -ne "Running") {
+    if ($null -eq $kubeletService) {
+        Set-ExitCode -ExitCode $global:WINDOWS_CSE_ERROR_START_NODE_RESET_SCRIPT_TASK -ErrorMessage "kubelet service is not installed after NodeResetScriptTask completed"
+    }
+
+    $retryCount=0
+    $maxRetryCount=6 # 1 minute
+    do {
+        $kubeletService=Get-Service -Name "kubelet" -ErrorAction SilentlyContinue
+        if ($kubeletService.Status -eq "Running") {
+            Write-Log -Message "kubelet service is running after NodeResetScriptTask"
+            break
+        }
+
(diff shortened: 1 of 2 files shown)


parts/windows/windowscsehelper.tests.ps1
```diff
@@ -666,4 +667,20 @@ Describe "Start-NodeResetScriptTask" {
...
+
+  It "recovers when kubelet reaches running after a retry" {
+    $script:kubeletCallCount = 0
+    Mock Get-Service -MockWith {
+      $script:kubeletCallCount++
+      if ($script:kubeletCallCount -eq 1) {
+        return [pscustomobject]@{ Status = "StartPending" }
+      }
+      return [pscustomobject]@{ Status = "Running" }
+    }
```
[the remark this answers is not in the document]
xuexu6666: Good catch — fixed in ec7171ff84. The service now stays `StartPending` through the presence check and the loop's first check, with a no-op `WaitForStatus`, so the recovery branch runs `Start-Service` exactly once and then exits `Running`.

parts/windows/windowscsehelper.ps1
```diff
@@ -333,8 +333,45 @@ function Start-NodeResetScriptTask {
...
+            $kubeletService.WaitForStatus('Running', [timespan]::FromSeconds(5))
+            Write-Log -Message "kubelet service started successfully after NodeResetScriptTask"
+            break
+        }
+        catch {
+            Write-Log -Message "Failed to start kubelet service after NodeResetScriptTask: $_"
+        }
+
+        if ($retryCount -lt $maxRetryCount) {
+            Start-Sleep -Seconds 5
```
r2k1, line 370: This is large number. We don't need to protect any downstream service. You probably can retry every few ms without negative consequences
But it may frequently delay VM provision time by extra 5 seconds

parts/windows/windowscsehelper.ps1
```diff
@@ -333,8 +333,45 @@ function Start-NodeResetScriptTask {
...
+        }
+    } while ($retryCount -lt $maxRetryCount)
```
djsly, line 372: those type of fixes always make me nervous. 
 
why is this a new problem ? as if something regressed ? kubelet is having a harder time to start ?
 
or stuck in a retry loop until eventually working ?

